PR #98, opened 2013-01-29
Added flag to databaseChangeLog to allow skipping of checksum validation...

damienbiggs:
....

Currently, there's no way to make improvements to old changesets without breaking users who have already ran those change sets. This approach allows the ability to skip validation for a change log.

This is a rework of my first attempt at this.
My first approach was too coarse grained, it removed the validation of the actual
changes in a changeset. This approach just prevents the validation of checksums for
changesets that have already ran.

Added integration test for the new functionality.
Tested using H2IntegrationTests.

┆Issue is synchronized with this [Jira Story](https://datical.atlassian.net/browse/LB-127) by [Unito](https://www.unito.io/learn-more)


ro-rah:
Hi @damienbiggs, I am looking to process the backlog of PRs and will bring this up with @nvoxland

PR #167, opened 2013-09-05
Fixed bugs with foreign keys

dmorari:
[1] Composite foreign key generation now works
(ForeignKeySnaphotGenerator.java)

https://liquibase.jira.com/browse/CORE-548

[2] - Crash since NullPointerException on generateChangeLog in
ForeignKeyComparator.java.


BarrieP:
I seem to be having a similar issue.

I'm new to MySQL and Liquibase, and have inherited a schema with compound keys ( based on 3 columns ), and a 4th column which is auto-increment ( but not primary key ).

MySQL is quite happy, we've then got liquibase to generate a changelog ( YML ), but when we then try to use the changelog, we get the same error "Incorrect table definition; there can be only one auto column and it must be defined as a key."

PR #238, opened 2014-02-07
make the check sum generation of sql and sqlFile changes independent of the default encoding which is different on different OS

Vampire:
If you have some non-ascii characters in your sql or sqlFile change that are checksum relevant, the generated checksum depends on the default encoding. This means you get different checksums if you run liquibase on windows (iso-8859-1) or linux (utf-8).
I even got different results on windows when running from the commandline and from within idea as idea seems to set the default charset of the VM it runs to utf-8.
With this change UTF-8 is always used to translate the characters to bytes and thus the checksum is os independent.

### DIFF
--- a/liquibase-core/src/main/java/liquibase/change/core/CreateProcedureChange.java
+++ b/liquibase-core/src/main/java/liquibase/change/core/CreateProcedureChange.java
@@ -14,6 +14,7 @@ import liquibase.util.StringUtils;
 import java.io.ByteArrayInputStream;
 import java.io.IOException;
 import java.io.InputStream;
+import java.io.UnsupportedEncodingException;
 
 @DatabaseChange(name = "createProcedure",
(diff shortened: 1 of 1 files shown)


lumiru:
Hi!

It seams that this was never sent to master, so we still meet this problem now.
Do we need a new PR? Or is there any other thing to do?

Vampire:
It was sent to `master`, it might have been broken again though of course:
```
$ git branch -a --contains 6eaf4f2 | grep upstream/master
  remotes/upstream/HEAD -> upstream/master
  remotes/upstream/master
$ git tag --contains 6eaf4f2 | grep '^v'
v3.10.0
v3.10.1
v3.10.2
v3.10.3
v3.8.1
v3.8.2
v3.8.3
v3.8.4
v3.8.5
v3.8.6
v3.8.7
v3.8.8
v3.8.9
v3.9.0
v4.0.0
v4.0.0-beta1
v4.0.0-beta2
v4.1.0
v4.1.1
v4.2.0
v4.2.1
v4.2.2
v4.3.0
v4.3.1
v4.3.2
v4.3.3
v4.3.4
v4.3.5
v4.4.0
v4.4.1
v4.4.2
v4.4.3
```

PR #277, opened 2014-06-09
MySQL specific features and changes

ghost:
Fulltext Index -- When changeset is created for text column with fulltext index a sql error is thrown when trying to migrate, i.e 
`event` text NOT NULL,
KEY `event` (`event`)
instead of
FULLTEXT KEY `event` (`event`)

SQL escape -- Not properly escaping single quotes for MySQL

Table Engine -- ignored or defaulted to innodb even though it should be something else.

Column Datatype -- Datatypes weren't being created properly. i.e DOUBLE as DOUBLE(22), INT maxed to INT(10) when it had been created as INT(11), etc.

Changelog specific changes and features:
Creating multiple changelog files at once -- using arg --changeLogFiles={0001.xml,0002.xml,0003.xml} and specifying in the liquibase.properties file which file gets which tables i.e
exclude_tables: { "0001.xml":["Table1","Table2"], "0003.xml":["Table3","Table4"] }
include_tables: { "0002.xml":["Table1","Table2"] }
Include takes priority over Exclude, i.e Only those tables will be dumped.


k2s:
> -- I don't remember the exact error, but I got an error reading views after I made the changes. I think it is in > liquibase-core/src/main/java/liquibase/snapshot/JdbcDatabaseSnapshot.java -- the getTables section. I >haven't had time to investigate yet.

please have look at this ticket https://liquibase.jira.com/browse/CORE-2400

ro-rah:
Hi @k2s , I see the submitter's account has been deleted, did you author this PR or know its history?

molivasdat:
Hi @k2s .  Are you still using Liquibase and need these changes?  Have they made their way into the current base or some version of them?

kataggart:
closing due to age; can reopen/open new issue as needed

PR #327, opened 2014-11-16
Merged code from the slf4j extension by Matt Bertolini, so that now it i...

cbotiza:
Merged Matt Bertolini's code from liquibase-slf4j so that slf4j logging is the default in the core.
Added logback classic as optional dependency, along with the JUL and JCL bridges.


mattbertolini:
Just found this thread. I'm with you @nvoxland. If you are going to be switching the logging setup in the 4.x line I would just use SLF4J directly.

nvoxland:
Yeah, it makes more sense. I'll do that.

PR #331, opened 2014-11-25
CORE-2041 HsqlDatabase should emit uppercase names when quoting reserved...

seanstephenson:
... words.  Otherwise, the database will end up not being case insensitive on queries.


xavier1985:
Does this piece of code is going to be in release of newer version?

PR #338, opened 2014-12-20
Fix CORE-2178 Use correct database collation

rhuffman:
Fixes CORE-2178: https://liquibase.jira.com/browse/CORE-2178

MSSQLDatabase.isCaseSensitive() uses the system collation rather than the database collation to determine whether or not the database is case-sensitive. That is fixed by checking DATABASEPROPERTYEX(<catalog>, 'collation').

Also, MSSQLDatabase.getdbcSchemaName() always converts to lower case. That causes a false negative on a case-sensitive database when checking to see if the lock table exists.

I am not nearly familiar enough to understand the full impact of this change, but it does seem to fix the problem we are having on SQL Server 2012.

### DIFF
--- a/liquibase-core/src/main/java/liquibase/database/core/MSSQLDatabase.java
+++ b/liquibase-core/src/main/java/liquibase/database/core/MSSQLDatabase.java
@@ -337,7 +337,7 @@ public class MSSQLDatabase extends AbstractJdbcDatabase {
     @Override
     public String getJdbcSchemaName(CatalogAndSchema schema) {
         String schemaName = super.getJdbcSchemaName(schema);
-        if (schemaName != null) {
+        if (schemaName != null && ! isCaseSensitive()) {
             schemaName = schemaName.toLowerCase();
         }
         return schemaName;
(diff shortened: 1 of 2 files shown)


liquibase-core/src/main/java/liquibase/database/core/MSSQLDatabase.java
```diff
@@ -349,8 +349,10 @@ public boolean isCaseSensitive() {
         if (caseSensitive == null) {
             try {
                 if (getConnection() != null) {
-                    String collation = ExecutorService.getInstance().getExecutor(this).queryForObject(new RawSqlStatement("SELECT CONVERT(varchar(100), SERVERPROPERTY('COLLATION'))"), String.class);
-                    caseSensitive = ! collation.contains("_CI_");
+                  String catalog = getConnection().getCatalog();
+                  String sql = String.format("SELECT CONVERT(varchar(100), DATABASEPROPERTYEX('%s', 'COLLATION'))", catalog);
```
nvoxland: Since the goal of Liquibase is to run arbitrary SQL I tend to not worry too much about sql injection problems so I think it is fine.

PR #339, opened 2015-01-05
Chnages needed for extension development for NoSQL databases

sbonde:
In order to develop extension for NoSQL databases like Cassandra, few private members needs to be make protected so I can extend LockService.


sbonde:
CORE-2185:Few fields needs to be changed as protected and need additional field in RanChangeSet

PR #340, opened 2015-01-06
4.0 action

nvoxland:
Looking for a code review on what I am thinking for the "Action" logic. 

The general idea is that Action implementations logically describe something that is wanted to be done like CreateSequence or QuerySql or DropTable. Actions should not just relate to database commands but describe any external system interaction like MakeApiCall or DeleteFile. ActionLogic implementations are where the logic is contained to actually perform the Action. They use a getPriority() method to allow extensibility so the when running against an Oracle database the CreateSequenceLogic may be different than the standard CreateSequence logic.

This action logic will take the place of the Change, Statement, and Generator logic in 3.x with Action taking the place of Change and Statement and ActionLogic taking the place of the *Generator classes. I wasn't seeing the value of having the separation of Change and Statement which is why they are being unified. I also removed the use of get/set wrappers in favor of basically a Map lookup to better support extensions of Actions which may add additional attributes. I'm using an "Attr" nested enum to provide some compile-support for standard attributes, but any key can be used as an attribute name.

The Action classes will also be used for metadata lookup in favor of the SnapshotGenerator classes in 3.x. This simplifies things by not having two abstractions around how you access the data.

You will also see use of the new Scope object both for managing singletons and storing execution state like the current database. 

Once we feel confident in this pattern, we can do the work of moving the existing Change/Statement/Generator/SnapshotGenerator logic over. For now, I only created a couple example Action classes and did not touch the existing code.


pmendelson:
Not sure if this is the right place tom comment on this concern.

A common code pattern in lq 3 can be found in

liquibase.change.coreAddColumnChange. generateStatements(Database)

...
            if ((database instanceof MySQLDatabase) && (column.getAfterColumn() != null)) {
                addColumnStatement.setAddAfterColumn(column.getAfterColumn());
            } else if (((database instanceof HsqlDatabase) || (database instanceof H2Database))
                       && (column.getBeforeColumn() != null)) {
                addColumnStatement.setAddBeforeColumn(column.getBeforeColumn());
            } else if ((database instanceof FirebirdDatabase) && (column.getPosition() != null)) {
                addColumnStatement.setAddAtPosition(column.getPosition());
            }

            addColumnStatements.add(addColumnStatement);

            if (database instanceof DB2Database) {
                sql.add(new ReorganizeTableStatement(getCatalogName(), getSchemaName(), getTableName()));
            }            

            if (column.getValueObject() != null) {
                UpdateStatement updateStatement = new UpdateStatement(getCatalogName(), getSchemaName(), getTableName());
                updateStatement.addNewColumnValue(column.getName(), column.getValueObject());
                sql.add(updateStatement);
            }
...

My home for lq 4 is that there is enough abstraction in the Database interface so that a change or action would not be aware of any specific Database implementation.

Of course this is much easier said than done.

one thought is lots of db methods like isAfterColumnClauseOk();

another thought is to give database modify options on (mostly) generated sql.

nvoxland:
Yes, I'm changing that pattern in 4.0. Each database has it's own sub-module now so it is even enforced by the compiler. I haven't pushed the 4.0 branch for a while because it's very torn apart, but I think it is significantly improved by removing all "database instanceof " logic out of liquibase-core. 

I was able to handle most of the isAfterColumnClauseOk() style logic by allowing database-specific logic to more easily reorder SQL clauses, introducing some Database-level methods like "requiresExplicitNullDefinitions", and  a more general AlterColumnAction that handle database-specific differences like "ALTER COLUMN"  vs "MODIFY COLUMN"

nvoxland:
Thanks, grammar has been fixed in master and will make it's way back to this branch.

nvoxland:
Added a targetted question on testing at http://forum.liquibase.org/#Topic/49382000001343003

PR #341, opened 2015-01-10
Handle timestamp precision

ancoron:
The current handling for TIMESTAMP precision handling really doesn't do a good job.

In the case of MySQL, the code simply ignores any precision specified and the version check is simply wrong (would fail for 6.x.x or 5.7.x).

In case of PostgreSQL, also the precision is simply ignored.

Related at least to the following issues:
* https://liquibase.jira.com/browse/CORE-1424


nvoxland:
Thanks!

PR #342, opened 2015-01-13
Core 2008

fm8:
First cut of changes for CORE-2008 (https://liquibase.jira.com/browse/CORE-2008).


fm8:
It's just the changeset 09225d8 that contains the changes, the rest are from an earlier merge. Happy to pull that out into a separate branch if that would be helpful...

nvoxland:
If you want to that woudl be great. Otherwise I can manually apply it but that loses some history and can cause more merge issues on your side.

nvoxland:
Resubmitted as #351

PR #345, opened 2015-01-17
CORE-2196 Adding root cause message to the build exception

mattbertolini:
I am adding the output of the root cause exception to the corresponding Ant BuildException to give more context about the error. Before the Ant task refactoring, there were no custom messages when BuildException was through which resulted in the root cause message being the only message. Once more messages were added, the root cause message was lost.
This should fix CORE-2196.


arunr307:
from where i get liquibase.jar with these correction? Now iam using jar that downloaded from this link.
http://sourceforge.net/projects/liquibase/files/Liquibase%20Core/liquibase-3.3.2-bin.zip/download
but this was not effected in this jar.

nvoxland:
The fix was just pulled into master, which will be in the Liquibase 3.4.0 release. I'm hoping to have 3.4.0 out in a couple weeks, but you can get a snapshot master build from https://liquibase.jira.com/builds/browse/CORE-LB which should include the fix in build 62 or above

PR #346, opened 2015-01-17
Remove 'Liquibase Home is not set.' from stdout

foxel:
Remove 'Liquibase Home is not set.' from stdout.
Output of updateSQL should not contain this line


nvoxland:
Thanks

PR #348, opened 2015-01-21
Core 2025

ssuehs-shango:
CORE-2025 listUnrunChangeSets should only list what would
be run.  It was returning all changesets

┆Issue is synchronized with this [Jira Story](https://datical.atlassian.net/browse/LB-80) by [Unito](https://www.unito.io/learn-more)

### DIFF
--- a/liquibase-core/pom.xml
+++ b/liquibase-core/pom.xml
@@ -281,6 +281,42 @@
                 </dependencies>
             </plugin>
         </plugins>
+        <pluginManagement>
+        	<plugins>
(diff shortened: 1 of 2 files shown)


RichardBradley:
The bug https://liquibase.jira.com/browse/CORE-2025 has been closed with the following comment:

> It is only listing the unrun changes. The seenChanges() is just returning the changeSets that are found by the visitor and the visitor is only seeing unrun changes

ro-rah:
It sounds like this is not an issue per @RichardBradley's comments (thanks for looking!), recommending PR close.

PR #352, opened 2015-01-27
When creating view definitions, trim leading and trailing whitespace

SteveDonie:


SteveDonie:
I tightened this up some - now looks specifically for zero at the end of the definition, to avoid changing other whitespace.

PR #354, opened 2015-02-03
CORE-2190: Proposed fix for not applying "modifySql" to remarks statements.

dusiema:
I looked into https://liquibase.jira.com/browse/CORE-2190 and also [commented](https://liquibase.jira.com/browse/CORE-2190?focusedCommentId=23995&page=com.atlassian.jira.plugin.system.issuetabpanels:comment-tabpanel#comment-23995) on the issue for what I think could be a solution. 
Here is the pull request with the changes of the description.

### DIFF
--- a/liquibase-core/src/main/java/liquibase/executor/AbstractExecutor.java
+++ b/liquibase-core/src/main/java/liquibase/executor/AbstractExecutor.java
@@ -31,7 +31,7 @@ public abstract class AbstractExecutor {
                 continue;
             }
             returnSql[i] = sql[i].toSql();
-            if (sqlVisitors != null) {
+            if (sqlVisitors != null && !statement.skipVisitors()) {
                 for (SqlVisitor visitor : sqlVisitors) {
                     returnSql[i] = visitor.modifySql(returnSql[i], database);
                 }
--- a/liquibase-core/src/main/java/liquibase/statement/AbstractSqlStatement.java
+++ b/liquibase-core/src/main/java/liquibase/statement/AbstractSqlStatement.java
@@ -6,4 +6,8 @@ public abstract class AbstractSqlStatement implements SqlStatement {
     public boolean skipOnUnsupported() {
         return false;
     }
+    
+    public boolean skipVisitors() {
+    	return false;
+    }
 }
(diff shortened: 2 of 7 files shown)


devilgate:
Hi Jens, This fixes the problem for me as far as the `COMMENT ON TABLE...` statements go. But then I hit the same problem with `COMMENT ON COLUMN...` 

The solution, of course, is to add your `skipVisitors` method to `SetColumnRemarks` as well.

I've made that change in my own fork, but I'm pretty new to Git and GitHub, so I'm not quite sure how to get that change to you. And in any case it might be easiest for you to apply the change and include it in this pull request.

dusiema:
Hey, good to hear it's working for you ! 
I'll add this to the `SetColumnRemaks` as well. Should go automatically into the pull request then.

dusiema:
I've added `skipVisitors` to `SetColumnRemarks`as well.

nvoxland:
The fix in this pull request definitely works fine, but I am concerned with how broad the change is. There can more more uses for sqlVisitor than just modifySql, and there are probably times people want the remarks modified as well.

I agree that the main problem is that the modifySql should be in the createTable etc. tags. It was mainly put in changeSet for simplicity of implementation and XSD, but then it does also mean it needs to apply to the whole changeset SQL string, not just some (as it does).

I think the best approach will be to look at a larger improvement to how modifySql works with 4.0 including adding more targetted support in all the change tags.

ro-rah:
Hi @dusiema,

Would you resolve branch conflict for this PR (thanks for adding the tests)?  Here's some guidance on that below:

Thanks for your pull request!

Here’s what happens next:

AlexUmarov:
Wow,  i have same problem in 2021 O_o
Fix some one this bug

nvoxland:
On further review, I decided this approach isn't going to work. 

I agree that modifySql/append is a bit of blunt instrument, but it's also doing what it says it does and in general we can't safely know what statements the user expects to be modify and which not. 

There is modifySql/replace and modifySql/regExReplace which gives you more control, and perhaps there are better modifySql logic statements we can create as well. But I don't think we can be building logic into some statements to simply not allow modification of them like this.

PR #355, opened 2015-02-05
Output a delimiter command for MySQL client.

adangel:
This behavior can be triggered with `outputDelimiter="true"` <strike>if "delimiter//" is used as endDelimiter</strike>
for SQL changes.
This allows for generating SQL via outputSQL that can be used
directly with the MySQL command line tool.

The following example changeset:

    <changeSet id="1" author="a">
        <sql endDelimiter="//" splitStatements="false" outputDelimiter="true">
            CREATE PROCEDURE x (a int)
            BEGIN
              DECLARE b int;
            END//
        </sql>
    </changeSet>

Will produce the following SQL:

    --  Changeset test-changelog.xml::1::a
    delimiter //
    CREATE PROCEDURE x (a int)
            BEGIN
              DECLARE b int;
            END//
    delimiter ;


nvoxland:
Thanks for the pull request, it looks like a very useful fix.

However, the "delimiter" prefix on the endDelimiter value is a little too magic for me. I'd prefer an additional boolean attribute to control whether the delimiter should be redefined before and after the statement.

Could you also rebase the change to master? It is more than I like to make in a maintenance release.

adangel:
Thanks for looking at my pull request. I agree, using the endDelimiter is magical :)

I'll try adding an additional attribute to the `sql` tag and rebase to master.

adangel:
Hm... seems like github still compares this pull request against the 3.3.x branch... 
I actually want to merge just the last two commits b30c37a and b7976fa into master

Should I open an new PR?

If chosen now the name `outputDelimiter`, so that the example will look now like this:

    <changeSet id="1" author="a">
        <sql endDelimiter="//" splitStatements="false" outputDelimiter="true">
            CREATE PROCEDURE x (a int)
            BEGIN
              DECLARE b int;
            END//
        </sql>
    </changeSet>
 
which creates this output:

    --  Changeset test-changelog.xml::1::a
    delimiter //
    CREATE PROCEDURE x (a int)
    BEGIN
        DECLARE b int;
    END//
    delimiter ;

Let me know, if the name `outputDelimiter` is ok or whether I should change it.

Thanks!

adangel:
Just created a new pr: #377

nvoxland:
Thanks for resubmitting it. I'll look at the new one

PR #356, opened 2015-02-07
Add a "tagExists" command to the liquibase commandline

adangel:
This allows for checking whether a given tag already exists. It is
needed in order to write a idempotent tag/update script and to prevent
accidentally overriding an already existing tag.

It also enhances the offline database to support tagging.


nvoxland:
Thanks. I merged this into master for 3.4 since it's more than I like to include in a maintenance release.

PR #358, opened 2015-02-13
Make createProcedure Change honor the procedureText Attribute

Brar:
The DatabaseChangeProperty metadata for
CreateProcedureChange.getProcedureText contained 'isChangeProperty =
false' which effectively made it unusable.

The problem is so obvious, that it looks like beeing intentional.
On the other hand it effectively disables a documented feature, gives a nonsensical error message, and there is no hint (neither a comment in the code nor something that I colud find in the issue tracker or via google), so it might as well be a simple oversight.

### DIFF
--- a/liquibase-core/src/main/java/liquibase/change/core/CreateProcedureChange.java
+++ b/liquibase-core/src/main/java/liquibase/change/core/CreateProcedureChange.java
@@ -121,7 +121,7 @@ public class CreateProcedureChange extends AbstractChange implements DbmsTargete
         this.procedureText = procedureText;
     }
 
-    @DatabaseChangeProperty(isChangeProperty = false)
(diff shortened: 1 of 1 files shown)


nvoxland:
The reason it is that way is because both procedureText and procedureBody return the same value and only procedureBody should be serialized/loaded from changelogs. The method to use in the process of changing to getProcedureText but we need to get procedureBody as the changeProperty for checksum backwards compatibility for the time being.

Brar:
Did you recognize that this effectively renders the documented (http://www.liquibase.org/documentation/changes/create_procedure.html#tab-yaml) use of procedureText unusable (at least for YAML changelogs - haven't tested XML) without giving the user any hint?

PR #359, opened 2015-02-17
CORE-2232

findepi:
https://liquibase.jira.com/browse/CORE-2232

### DIFF
--- a/liquibase-core/src/main/java/liquibase/database/core/PostgresDatabase.java
+++ b/liquibase-core/src/main/java/liquibase/database/core/PostgresDatabase.java
@@ -9,6 +9,7 @@ import liquibase.structure.DatabaseObject;
 import liquibase.exception.DatabaseException;
 import liquibase.executor.ExecutorService;
 import liquibase.logging.LogFactory;
+import liquibase.statement.core.RawCallStatement;
 import liquibase.statement.core.RawSqlStatement;
 import liquibase.structure.core.Table;
 import liquibase.util.StringUtils;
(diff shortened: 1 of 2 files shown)


nvoxland:
Thanks!

liquibase-core/src/main/java/liquibase/database/core/PostgresDatabase.java
```diff
@@ -235,7 +236,14 @@ public boolean isReservedWord(String tableName) {
 
     @Override
     protected String getConnectionSchemaName() {
-        return "public";
+        try {
+            String currentSchema = ExecutorService.getInstance().getExecutor(this)
+                    .queryForObject(new RawCallStatement("select current_schema"), String.class);
```
danielc576: Hi, 
This query should be: "select current_schema()"
Please see: http://www.postgresql.org/docs/8.3/static/functions-info.html
Can you please fix this ?

Thanks,
Daniel

nvoxland: I pushed a change to master for this change

shrz:
@nvoxland 
It seems that current_schema without () doesn't work for PostgreSQL < 8.4
http://www.postgresql.org/docs/8.3/static/functions-info.html
http://www.postgresql.org/docs/8.4/static/functions-info.html

So Current Liquibase Core Release: 3.4.1 (July 28, 2015) doesn't work for PostgreSQL < 8.4.

PR #362, opened 2015-02-24
CORE-2192: Can't generate SQL scripts from Offline Oracle databases. In ...

joaomc:
Fix for CORE-2192, which prevents me from using liquibase:updateSQL from the Maven plugin. OracleDatabase's setConnection is always calling getWrappedConnection for the DatabaseConnection, but OfflineConnection doesn't implement it.


nvoxland:
Thanks!

PR #363, opened 2015-02-24
Fixed bug that Servlet listener wrongly handle LiquibaseServletListener.LIQUIBASE_PARAMETER

michaldo:
There is a bug that a dot remains as parameter name prefix


nvoxland:
Thanks

PR #364, opened 2015-02-24
CORE-2217 Add DataTypeFactory support for delimited data type names, improve resolution of MSSQL data types

mches:
[CORE-2217 Add DataTypeFactory support for delimited data type names, improve resolution of MSSQL data types](https://liquibase.jira.com/browse/CORE-2217)

### DIFF
--- a/liquibase-core/src/main/java/liquibase/database/AbstractJdbcDatabase.java
+++ b/liquibase-core/src/main/java/liquibase/database/AbstractJdbcDatabase.java
@@ -1480,4 +1480,9 @@ public abstract class AbstractJdbcDatabase implements Database {
     public String escapeDataTypeName(String dataTypeName) {
         return dataTypeName;
     }
+
+    @Override
+    public String unescapeDataTypeName(String dataTypeName) {
+        return dataTypeName;
+    }
 }
--- a/liquibase-core/src/main/java/liquibase/database/AbstractJdbcDatabase.java
+++ b/liquibase-core/src/main/java/liquibase/database/AbstractJdbcDatabase.java
@@ -69,6 +69,7 @@ public abstract class AbstractJdbcDatabase implements Database {
     protected String sequenceCurrentValueFunction;
     protected String quotingStartCharacter = "\"";
     protected String quotingEndCharacter = "\"";
+    protected String quotingEndReplacement = "\"\"";
 
     // List of Database native functions.
     protected List<DatabaseFunction> dateFunctions = new ArrayList<DatabaseFunction>();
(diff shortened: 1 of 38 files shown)


mches:
Resolved conflict with CORE-2287

nvoxland:
It is good to support that style, but I think it would be good to not return delimited type names by default  in the generated SQL. Is there a reason it makes sense to delimite the type names in sqlserver?

mches:
Thanks for taking the time to review this request. My goal with this change is to make Liquibase support for SQL Server data types truly first-class.

To answer your question, in SQL Server, unlike PostgreSQL, delimiting entirely lowercase identifiers is always allowed. Delimiting with square brackets is always allowed regardless of the SET QUOTED_IDENTIFIER ON/OFF setting. Professional tools like Microsoft SQL Server Management Studio and Red Gate SQL Toolbelt always use delimited identifiers in generated SQL. Built-in types would never require quoting, however, user-defined types can use identifiers that require quoting. For these reasons I believe it is a sensible default to canoncialize the names to their quoted equivalent.

mches:
Here is an example of other professional tools:

3-10-2015 3-44-47 pm

mches:
I made a few changes that improve integration with the core Database and DatabaseObject abstractions as well as handle user-defined types and other unknown types a lot better. I think you'll find it more palatable. Let me know what you think.

nvoxland:
If brackets are needed to support custom types, I think it makes sense to include it for sqlserver, and if other tools just include them for everything it probably makes sense to follow that pattern.

I had to make a couple changes to better support snapshot logic. In particular, I removed the "extends AbstractDatabaseObject" because it will take a bit of work to integrate handling of datatypes with snapshot. That meant the call to escapeObjectName() in escapeTypeName() no longer worked, but I put in a work-around just passing DatabaseObject.class as the second parameter.

I also put in a change to MissingTableChangeGenerator to remove the brackets from the type name so they aren't included in generateChangeLog/diffChangeLog. That also is a bit hack-y but I don't want the brackets in the output changeSets since it hurts portability between databases. I'm working on larger changes to how all that logic works in Liquibase 4.0 and will clean it up at that point.

I pushed the branch with my changes to NSIT-CORE-2217 in the main liquibase repo. Can you test that my changes still handle what you need them to?

mches:
I agree we wouldn't want to see any braces in any generated change logs. I cherry picked your commit into this PR. I added an abstraction for the unenscaping to make it a little less hacky. That will allow it to be easily reused, modified, or overridden as needed. How do you like it?

mches:
I noticed my change to MissingTableChangeGenerator didn't account for data types with parameters. I will see what I can do about it.

mches:
This is ready for review again.

mches:
Integrated [CORE-2304](https://liquibase.jira.com/browse/CORE-2304) from [3.3.x](https://github.com/liquibase/liquibase/commits/3.3.x) branch and resolved conflict in DataTypeFactoryTest

mches:
I propose to push this to 3.4.x rather than 3.3.x. Please see the rebased version at [PR#398](https://github.com/liquibase/liquibase/pull/398).

PR #365, opened 2015-02-25
CORE-1857: Wrong column size detection on varchar2 fields with char as d...

wardvijf:
See https://liquibase.jira.com/browse/CORE-1857


nvoxland:
Thanks!

PR #367, opened 2015-02-27
Core-2262 3.3.2 ant task dies on NPE in ChangeLogParameters

bsandiford:
CORE-2262 3.3.2 ant task dies on NPE in ChangeLogParameters - check for "currentLabelExpression == null" before using it.


nvoxland:
Thanks!

PR #368, opened 2015-02-27
CORE-2263 Index Snapshot - doesn't include upper cased name indexes when db is NOT case sensitive

bsandiford:
CORE-2263 Index Snapshot - doesn't include upper cased name indexes when db is NOT case sensitive - Make key to foundIndexes map 'corrected' to match 'corrected' exampleName.


nvoxland:
Thanks!

PR #369, opened 2015-03-01
CORE-1411 - MariaDB support

smasset:
Added a MariaDBDatabase class extending MySQLDatabase class and updated tests


nvoxland:
Merged into master rather than 3.3.x since the change is larger than I like to do in a point release. It will go out with 3.4.

smasset:
Thanks. Any estimate as to when 3.4 will be released ?
I have another pending PR to include MariaDB support for a project using liquibase.

PR #370, opened 2015-03-03
reset of quoting strategy is wrong

bjamet:
reset should be set to previous strategy, not LEGACY.
A better place to reset should be in changeSet.execute(...), in case of exception...

        this.database.setObjectQuotingStrategy(ObjectQuotingStrategy.LEGACY);


nvoxland:
Thanks!

PR #371, opened 2015-03-07
CORE-2274 Changing change log parameter from FileResource to String

mattbertolini:
When the Ant tasks were refactored the change log parameter because an Ant ```FileResource``` object. This means change logs can only be located on the file system. This negates the usefulness of the ```ResourceAccessor``` interface in Liquibase.
I have converted the change log parameter back to a ```String``` and am letting the ```ResourceAccessor``` handle the locating and reading of the file.
This fixes CORE-2274.


nvoxland:
Thanks

PR #373, opened 2015-03-11
Added option "updateable" to the column tag.

SigmaIt:
Added option "updateable" to the column tag. This option allows you to not include a column in the update script. Default value = "true".

┆Issue is synchronized with this [Jira Story](https://datical.atlassian.net/browse/LB-84) by [Unito](https://www.unito.io/learn-more)
┆Fix Versions: Liquibase 4.2.0

### DIFF
--- a/liquibase-core/src/main/java/liquibase/change/core/LoadDataChange.java
+++ b/liquibase-core/src/main/java/liquibase/change/core/LoadDataChange.java
@@ -174,7 +174,7 @@ public class LoadDataChange extends AbstractChange implements ChangeWithColumns<
 
                     Object value = line[i];
 
-                    ColumnConfig columnConfig = getColumnConfig(i, headers[i].trim());
+                    LoadDataColumnConfig columnConfig = getColumnConfig(i, headers[i].trim());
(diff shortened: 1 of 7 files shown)


ro-rah:
@SigmaIt ,

Sorry for the very long delay.

Thanks for your pull request!

Here’s what happens next:

ro-rah:
@SigmaIt I am moving the target to 3.10.x since that is the latest release we are working on.  Please let me know if this was intended only for 3.2.x releases.

nvoxland:
If I'm understanding the intent, the idea is that `<loadData>` has logic to be smart about whether a row is an update or an insert. 

But, sometimes you don't want to update specific fields, even if they are different. This fix makes it so you can mark some of the columns in `loadData` as not ones to update.

It seems like a useful change, but we should spend a few minutes thinking about if "updatable" is the right term to be adding as a flag, and making sure it still works in the current codebase.

It changes the xsd, so it will have to go out in a fill release, not a patch release.

nvoxland:
I merged the change in, but renamed the field from "updateable" to "allowUpdate"

PR #375, opened 2015-03-11
CORE-2234: columnExists precondition speedup by relying on SQL parser in DB

findepi:
Speed up columnExists precondition by delegating work to SQL parser of the
database. Instead of inspecting database metadata (which can be slow when
many tables/columns exist in the db), check that the column can be selected
from the table.


nvoxland:
That is a good approach, I may look into implementing something similar for other objects with the liquibase 4.0 refactoring. Thanks for the idea

PR #378, opened 2015-03-12
CORE-1943 - Handle Error: InetAddress.getLocalHost().getHostName() UnknownHostException

bsandiford:
CORE-1943 - Handle Error: InetAddress.getLocalHost().getHostName() UnknownHostException results in NoClassDefFoundError - Use 'Smarter' way to get localhost for local host name as well as local host address.  (Basically same approach as 2.0.2 functionality, which worked)


sbuettner:
Would be great if this could be merged. Current release has issues on AWS.

nvoxland:
Thanks!

PR #379, opened 2015-03-13
CORE-2296 Upgrade Groovy and Spock to maintained versions

mches:
[CORE-2296 Upgrade Groovy and Spock to maintained versions](https://liquibase.jira.com/browse/CORE-2296)


nvoxland:
thanks

PR #380, opened 2015-03-16
CORE-1866 - Filtering changelog list by includeAll tag is not working

bsandiford:
CORE-1866 - Filtering changelog list by includeAll tag is not working - fix bad "filter" entry in xsd's to be "resourceFilter" to match Java code.


nvoxland:
Thanks!

PR #382, opened 2015-03-16
CORE-2299 - Add capability to ignore missing or empty folder with includeAll

bsandiford:
CORE-2299 - Add capability to ignore missing or empty folder with includeAll - Add new 'ignoreMissingOrEmptyDirectory' attribute to 'includeAll'.  Default false.


nvoxland:
Thanks!

I merged to master rather than 3.3.x due to the xsd change. I updated the change to use the 3.4 XSD.

I also changed the flag to be errorIfMissingOrEmpty with a default of true

PR #383, opened 2015-03-16
CORE-2298: HSQL not allowed to add defaultValueComputed for datetime columns.  Added method to HsqlDatabase to check for conditions under which computed values are allowed.

nthorn552:
According to the HSQL documentation (http://www.hsqldb.org/doc/guide/ch09.html#columnDef-entry), there are several types of computed default values allowed for datetime columns.  This update adds a funciton to the HsqlDatabase class to check if a computed value function is allowed given the column data type, and uses this function in the AddDefaultValueGenerator to allow certain computed values.


nvoxland:
Thanks!

PR #387, opened 2015-03-20
CORE-2307 ChangeLog table name option in command line tool

gquintana:
Add databaseChangeLogTableName and databaseChangeLogLockTableName options to command line


nvoxland:
Thanks for the fix. I merged it into master rather than 3.3.x since it's a larger change than I like to make in a patch release.

bruce-szalwinski:
Thanks for the fix.  I see that it is available in 3.4-SNAPSHOT.  Maven Central Repo has 3.3.5 as the latest.  When could I expect to find 3.4 in Maven Central Repo?

nvoxland:
I'm hoping to wrap up the 3.4.0 final release later this week

PR #388, opened 2015-03-24
CORE-2316 Don't overwrite datatypes if they have the same priority

sabineheider:
Commit message: 
*TreeSet with Comparator means that per priority only one class
will survive. If you have for example both MaxDB and HANA
extensions on your classpath only one of them will be used,
not both as they both return PRIORITY_DATABASE for DateTime
types.*

What do you think of our approach to switch from TreeSet to ArrayList? Is that feasible, or would you prefer a different solution?

It's a bit ugly that the registry is exposed by the getRegistry method. We commented it to see whether that method is used at all in our scenario, apparently it isn't. Should we include it again with an adapted return type?

### DIFF
--- a/liquibase-core/src/main/java/liquibase/datatype/DataTypeFactory.java
+++ b/liquibase-core/src/main/java/liquibase/datatype/DataTypeFactory.java
@@ -18,7 +18,7 @@ public class DataTypeFactory {
 
     private static DataTypeFactory instance;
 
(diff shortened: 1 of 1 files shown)


nvoxland:
I think the approach works for now. I'm doing a general refactoring with 4.0 including improvements to how all the Registry pattern extensions work. 

I did merge the change into master for 3.4 rather than the 3.3.x bugfix branch due to the API/logic change, though.

Thanks for the fix

PR #389, opened 2015-03-25
CORE-2318 Add support for converting BigDecimal objects to a SQL string via DataTypeFactory

mches:
[CORE-2318](https://liquibase.jira.com/browse/CORE-2318) Add support for converting BigDecimal objects to a SQL string via DataTypeFactory


nvoxland:
Yes, looks fine for 3.3.x. Merged in

PR #393, opened 2015-03-30
Core-2206

carter1982:
fix ConcurrentModificationException when running org.liquibase:liquibase-maven-plugin:diff using hibernate:classic configuration


nvoxland:
Thanks!

PR #395, opened 2015-04-02
CORE-2325 Liquibase - New versions break DB create

mches:
[CORE-2325](https://liquibase.jira.com/browse/CORE-2325) Liquibase - New versions break DB create

Fixed UUIDType.objectToSql(), added a test case for UUIDType, added infrastructure for piecewise override of LiquibaseDataType.objectToSql(), added support for converting java.util.UUID


nvoxland:
Thanks!

PR #397, opened 2015-04-07
CORE-2321 Update MSSQL query to use UPDATE statement rather than a MERGE

mches:
[CORE-2321](https://liquibase.jira.com/browse/CORE-2321) Update MSSQL query to use a safer, backward-compatible UPDATE statement rather than a MERGE


mches:
Unless you can think of a case where ID, AUTHOR, FILENAME is not unique, the updated should guarantee updating exactly one record. If that's not the case, we'll have to take into account ORDEREXECUTED and DATEEXECUTED in the JOIN.

PR #400, opened 2015-04-09
CORE-2322 Fix fall through issue

mches:
[CORE-2322](https://liquibase.jira.com/browse/CORE-2322) Add ability to update to a tag

Fixed a fall through issue that is causing a test failure on master


nvoxland:
Already saw and made that change. Thanks for the quick fix though

mches:
Great, thanks!

PR #401, opened 2015-04-09
CORE-2300 Unsigned Int / Bigint cannot be created

mches:
[CORE-2300](https://liquibase.jira.com/browse/CORE-2300) Unsigned Int / Bigint cannot be created


nvoxland:
Thanks, looks good

PR #404, opened 2015-04-09
CORE-2325 Improve the merge conflict resolution between CORE-2325 and CORE-2318

mches:
I had a slightly different idea of how to handle the merge conflict between CORE-2325 and CORE-2318. This should make behavior more predictable if someone later overrides LiquibaseDataType#numberToSql(Number, Database). The code in UUIDType#otherToSql(Object, Database) was also able to be simplified due to the inheriting the necessary code to handle a DatabaseFunction from LiquibaseDataType#objectToSql(Object, Database).


nvoxland:
Thanks

PR #407, opened 2015-04-15
CORE-419 Add support for descending key columns on indexes, unique constraints and primary keys

mches:
Adds support for descending key columns on indexes, unique constraints and primary keys

For example:
```
<createIndex tableName="my_table" indexName="my_index">
    <column name="col1"/>
    <column name="col2" descending="true"/>
</createIndex>

<addPrimaryKey tableName="my_table" columnNames="col1, col2 DESC"/>

<addUniqueConstraint tableName="my_table" columnNames="col1, col2 DESC"/>
```

~~Snapshots aren't addressed, but~~ This should resolve all the following issues:

Liquibase JIRA:
[CORE-419](https://liquibase.jira.com/browse/CORE-419) Allowing ASC and DESC in index definitions
[CORE-1232](https://liquibase.jira.com/browse/CORE-1232) Ordered index creation

Liquibase Forum:
http://forum.liquibase.org/topic/create-index-and-order

Stack Overflow:
http://stackoverflow.com/questions/11955313/how-to-create-ordered-index-with-liquibase

### DIFF
--- a/liquibase-core/src/main/java/liquibase/change/ColumnConfig.java
+++ b/liquibase-core/src/main/java/liquibase/change/ColumnConfig.java
@@ -54,6 +54,7 @@ public class ColumnConfig extends AbstractLiquibaseSerializable {
     private BigInteger startWith;
     private BigInteger incrementBy;
     private String remarks;
+    private Boolean descending;
 
     /**
      * Create a ColumnConfig object based on a {@link Column} snapshot.
@@ -728,6 +729,15 @@ public class ColumnConfig extends AbstractLiquibaseSerializable {
         return this;
     }
 
+    public Boolean getDescending() {
+        return descending;
+    }
+
+    public ColumnConfig setDescending(Boolean descending) {
+        this.descending = descending;
+        return this;
+    }
+
     @Override
     public String getSerializedObjectName() {
         return "column";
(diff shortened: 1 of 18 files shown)


mathieu-pousse:
+1 for the effort !

mches:
I started implementing the necessary snapshot logic, however I have observed that the Index#getColumns() are essentially being replaced with Table#getColumns() by DatabaseSnapshot#includeNestedObjects(DatabaseObject). The table columns don't have the "descending" attribute set. Therefore, when the MissingIndexChangeGenerator runs, it won't have access to the "descending" attribute and all the index key columns are treated as ascending.

mches:
Development Discussion: http://forum.liquibase.org/#Topic/49382000001294001

mches:
Snapshot and diff work as expected with the following [workaround code](https://github.com/NSIT/liquibase/commit/dcd70fe13a384907402b2be40368a06f4fbda08a) added to `DatabaseSnapshot#includeNestedObjects(DatabaseObject)`. 

    if (object.getClass() == Index.class && field.equals("columns")) {
        continue;
    }
    if (object.getClass() == PrimaryKey.class && field.equals("columns")) {
        continue;
    }
    if (object.getClass() == UniqueConstraint.class && field.equals("columns")) {
        continue;
    }

But, what possible negative consequences am I overlooking?

mches:
Rebased to resolve conflicts

nvoxland:
I think the changes look good, and I like how you accounted for people using the name="address desc" work-around. 

I think your snapshot work-around is fine, I can't think of issues it will cause off hand. I'm looking to simplify and re-architect how the snapshot process works with Liquibase 4.0 which will hopefully avoid having to do something similar at that point.

PR #408, opened 2015-04-15
CORE-2052 : dropAll failure on Postgres

bonnetb:
On a Postgres database, dropAll tries to drop even views not owned by the
user, which causes errors.


jmvivo:
Great. Thanks!!

nvoxland:
The documentation is not quite right. DropAll shouldn't drop object based on the owner but instead drops everything in the configured schema(s). I updated the javadoc for 3.4.0.

Closing the pull request since it would change the functionality

PR #410, opened 2015-04-17
Fixed invalid YAML serialized changelog output

mdehoog:
...by representing changeSets as a list instead of an associative array


nvoxland:
Merged, thanks!

PR #413, opened 2015-04-30
Fixed the classpath handling to make DbDoc work

mrumpf:
This is a hack as the concept of storing the path seems to be broken or at least I do not understand it yet.


nvoxland:
It is a bit of a hack, but a good change for now. I'm doing a lot of cleanup with 4.0 that should avoid this type of duplication, but it works for now. Thanks!

PR #416, opened 2015-05-05
Oracle12c autoincrement identity

20mithrandir:
Because Oracle 12c supports 'Identity' in a similar way like MS SQL Server and we are currently migrating a big product stack to Oracle we need this feature enabled.

The change basically returns true for the OracleDatabase if the major version is 12 or greater. Otherwise it continues to deliver false (like before).

I included following example in the comment (which maybe can be removed). The dataset:

	<changeSet author="p.taucher" id="test-pk-identity-autoincrement-01">
		<createTable tableName="AutoIncTest">
			<column autoIncrement="true" name="IDPrimaryKey" type="${bigint}">
				<constraints nullable="false" primaryKey="true" primaryKeyName="PK_AutoIncTest"/>
			</column>
			<column name="TypeID" type="${tiny}">
			</column>
			<column name="Description" type="${description.type}">
				<constraints nullable="true"/>
			</column>
		</createTable>
	</changeSet>

This resulted in following SQL script prior to my change:

CREATE TABLE AutoIncTest (IDPrimaryKey NUMBER(19) NOT NULL, TypeID NUMBER(3) NOT NULL, Description NVARCHAR2(50), CONSTRAINT PK_AutoIncTest PRIMARY KEY (IDPrimaryKey));

After my change the script now includes the identity clause (GENERATED BY DEFAULT AS IDENTIY)

CREATE TABLE AutoIncTest (IDPrimaryKey NUMBER(19) GENERATED BY DEFAULT AS IDENTITY NOT NULL, TypeID NUMBER(3) NOT NULL, Description NVARCHAR2(50), CONSTRAINT PK_AutoIncTest PRIMARY KEY (IDPrimaryKey));


nvoxland:
Thanks! Closed issue https://liquibase.jira.com/browse/CORE-1731

deepu105:
@nvoxland is this in the latest release? if not when will be the latest release?

khorchanov:
Hello,

I am using liquibase for my database migrations, with the yml format I set autoIncrement: true on a bigint primary key, 

The generated script still : id NUMBER(38, 0) NOT NULL

PR #418, opened 2015-05-07
Update AbstractLiquibaseMojo.java

jkutner:
A default value of "true" creates many problems on CI and in automated deployments. Many users have difficulty figuring out why errors such as "No X11 DISPLAY variable was set, but this program performed an operation which requires it" occur.


nvoxland:
I'm going to keep the promptOnNonLocalDatabase parameter as true by default because that is the safest route and there are people who expect it to be working like that.

I may revisit how to handle it with 4.0 since it does cause confusion

jkutner:
Thanks for taking a look.

PR #419, opened 2015-05-07
Added support for constraint name in addNotNullConstraint change. Cur…

mrmaus:
Hi
Unfortunately at the moment Liquibase doesn't support custom constraint names for NOT NULL constraints in Oracle, so I've added it into addNotNullConstraint change. This is very important for us, since we want to maintain same constraint names across all our database installations. 
Please take a look and let me know what you think. It would be great if you could put the change into next 3.3.x release.
The change is done based on 3.3.x branch
Tested on oracle instance, works fine

Thanks a lot!
Andrei


nvoxland:
The change looks good, although with the XSD change I had to merge it into master for 3.4 instead. I try not to make any xsd changes in patch releases such as 3.3.6

nvoxland:
Fixes issue https://liquibase.jira.com/browse/CORE-562

nvoxland:
Merged into master

PR #420, opened 2015-05-07
CORE-2358 Improve data types for Liquibase tables in MSSQL

mches:
[CORE-2358](https://liquibase.jira.com/browse/CORE-2358) Improve data types for Liquibase tables in MSSQL

On SQL Server, for the Liquibase tables, DATABASECHANGELOG and DATABASECHANGELOGLOCK, use nvarchar, which supports Unicode, rather than varchar, which supports single-byte encodings on a single code page only, if the JDBC driver, e.g. Microsoft or jTDS, is configured to send all String parameters as Unicode. On SQL Server 2008 and later, use datetime2(3), which can accurately represent milliseconds, rather than datetime, which rounds the milliseconds. Make it easy for a trivial extension, for any database, not just SQL Server, to override only the data types.


nvoxland:
Merged in, thanks!

PR #422, opened 2015-05-11
CORE-2360 Skip is active by default

mches:
[CORE-2360](https://liquibase.jira.com/browse/CORE-2360) Skip is active by default

Liquibase should probably continue to run by default when run using Maven. The default behavior changed in 3.3.3 to not run by default.


mches:
The default behavior also changed on master for the unreleased 3.4.0.

PR #423, opened 2015-05-11
CORE-2361 preConditions, rollback, property, include, includeAll cannot be serialized

mches:
[CORE-2361](https://liquibase.jira.com/browse/CORE-2361) preConditions, rollback, property, include, includeAll cannot be serialized

Makes it possible to serialize the `preConditions`, `rollback` children of a `changeSet` and the `property`, `preConditions`, `include` and `includeAll` children of a `databaseChangeLog`.

Note that the changes to `ChangeLogSerializer` interface are backwards compatible to callers, but not implementors. I think that is acceptable for 3.4.x. It's not possible to keep or deprecate the former `write` method due to duplicate method signatures caused by type erasure.

### DIFF
--- a/liquibase-core/src/main/java/liquibase/changelog/ChangeLogChild.java
+++ b/liquibase-core/src/main/java/liquibase/changelog/ChangeLogChild.java
@@ -1,0 +1,6 @@
+package liquibase.changelog;
+
+import liquibase.serializer.LiquibaseSerializable;
+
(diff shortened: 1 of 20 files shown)


nvoxland:
It looks like a lot of good changes. I agree that it is ok to require changes to implementations, there should be few of them and it is a required change.

PR #427, opened 2015-05-12
CORE-2357 alterSequence does not work as expected when you need to change the cache size

sanastasiadis:
Added annotation on cacheSize property getter in AlterSequenceChange class and include it in checkStatus checks.
Updated alter sequence unit test in XMLChangeLogSerializerTest and added AlterSequenceGeneratorTest.


nvoxland:
Thanks!

PR #428, opened 2015-05-13
CORE-2366   shouldRun defaulted false

eino-makitalo:
I found out that system does not work with Derby Client Driver in my project.  3.3.3 seems to work but it was false positive, because maven plugin gave me shouldRun=false and warning ....

I think shouldRun should be true by default.


nvoxland:
Thanks!

PR #430, opened 2015-05-19
CORE-2019: Rollback Fails When Containing Only a Comment Tag

bradhandy:
When parsing a changelog where one or more changeSets have a rollback containing a single "comment" tag, the rollback will fail.  Liquibase will attempt converting the comment into a change.  The conversion fails with a null Change object which is not checked prior to adding to the rollbackChanges collection.

Note:  I just realized by commit message does not include the summary of CORE-2019.  My apologies.


nvoxland:
Thanks! I merged it into master for 3.4.0 since I'm not planning on another 3.2 patch release.

bradhandy:
Thanks!

PR #431, opened 2015-05-21
CORE-864 ability for LoadData to emit multi row insert statements.

AartBluestoke:
the InsertSet is a container for a set of Insert statements, which would allow batching the updates rather than issuing RBAR (row by agonizing row) insert statments as is currently done (see CORE-864)

== code compiles and passes all unit tests in LoadDataChangeTest.groovy==

i'm proposing this PR in order to a) get some feedback, b) provide a codebase to assist resolving CORE-864.

todo:
- InsertSetGenerator.validate
- better logic for controlling the batch size; 

i suspect there are dbms specific row, packet, and sql string length limits that would be exceeded here. This is why i selected 50 as the batch size; reduces the overhead to 2% of what it was, without having a significant chance of excessively long statements.

as neither the change nor sqlgenerator folders have changed since the last merge of the 3.3.x branch, this pr should be able to be applied to all branches (3.3.x,3.4.x,4.x)

Compatibility of current batch approach:

sql server: since 2008
postgres: since 8.2
mysql: since 3.22.5

oracle: not in this format, force batch size to 1, or perform an alternate code block removing the keyword "insert" then adding a header of "insert all" and a footer of "SELECT * FROM dual;" for each batch.

### DIFF
--- a/liquibase-core/src/main/java/liquibase/change/core/LoadDataChange.java
+++ b/liquibase-core/src/main/java/liquibase/change/core/LoadDataChange.java
@@ -11,6 +11,7 @@ import liquibase.resource.ResourceAccessor;
 import liquibase.resource.UtfBomAwareReader;
 import liquibase.statement.SqlStatement;
 import liquibase.statement.core.InsertStatement;
+import liquibase.statement.core.InsertSetStatement;
 import liquibase.structure.core.Column;
 import liquibase.util.StreamUtil;
 import liquibase.util.StringUtils;
@@ -162,7 +163,7 @@ public class LoadDataChange extends AbstractChange implements ChangeWithColumns<
                 throw new UnexpectedLiquibaseException("Data file "+getFile()+" was empty");
             }
 
-            List<SqlStatement> statements = new ArrayList<SqlStatement>();
+            InsertSetStatement statements = this.createStatementSet(getCatalogName(), getSchemaName(), getTableName());
             String[] line;
             int lineNumber = 0;
 
@@ -222,10 +223,10 @@ public class LoadDataChange extends AbstractChange implements ChangeWithColumns<
 
                     insertStatement.addColumnValue(columnName, value);
                 }
-                statements.add(insertStatement);
+                statements.addInsertStatement(insertStatement);
             }
 
-            return statements.toArray(new SqlStatement[statements.size()]);
+            return new SqlStatement[]{statements}; // we only return a single "statement" - it's capable of emitting multiple sub-statements, should the need arrise, on generation.
         } catch (IOException e) {
             throw new RuntimeException(e);
         } catch (UnexpectedLiquibaseException ule) {
@@ -285,6 +286,10 @@ public class LoadDataChange extends AbstractChange implements ChangeWithColumns<
 
     protected InsertStatement createStatement(String catalogName, String schemaName, String tableName){
         return new InsertStatement(catalogName, schemaName,tableName);
+    }
+
+    protected InsertSetStatement createStatementSet(String catalogName, String schemaName, String tableName){
+        return new InsertSetStatement(catalogName, schemaName,tableName);
     }
 
     protected ColumnConfig getColumnConfig(int index, String header) {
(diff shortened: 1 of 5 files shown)


nvoxland:
Interesting to see what a performance improvement you get, even batching at 50 rows. What does the resulting SQL look like and what databases would it support?

AartBluestoke:
currently generatated code (without this change) looks like:
INSERT INTO test (a, b) VALUES ('0', '--')
GO
INSERT INTO test (a, b) VALUES ('1', 't1')
GO
INSERT INTO test (a, b) VALUES ('2', 't2')
GO

this PR aims to emit code like:
INSERT INTO test (a, b) VALUES ('0', '--')
,('1', 't1')
,('2', 't2');

---
sql server uses this syntax.
( https://technet.microsoft.com/en-us/library/dd776381(v=sql.105).aspx )

oracle has the same syntax:
INSERT INTO tbl_name (a,b,c) VALUES(1,2,3),(4,5,6),(7,8,9);
(https://docs.oracle.com/cd/E17952_01/refman-5.1-en/insert.html)
(various other hints at solving this problem at
https://docs.oracle.com/cd/E17952_01/refman-5.1-en/insert-speed.html
)

mysql can use this syntax:
mysql has the same insert values syntax:
http://www.electrictoolbox.com/mysql-insert-multiple-records/

postgres supports this syntax:
http://www.postgresql.org/docs/8.2/static/sql-insert.html

--
testing batch inserts on my laptop ( code generated by sql server, not liquibase, but in the format of the statement above)

running on a hdd: 6400 rows, 
batch size 1: minimum time to insert all rows: 4.9 seconds (over 6 seconds sometimes)
batch size 100: time is more variable but is between 1 and 2.5 seconds

ssd:
batch size 1: 3.5 seconds
batch size 100: down to 0.5 seconds, but around 1 typically

with this small number of rows, re-running the insert with a delete * from table gives a varied time, 

overall, on fast or slow disks (or if you would be running against a remote server and be latency constrained) approx 80% time reduction at batch size 100.

table spec (has a primary key and a foreign key, enforced on insert, inserts are in ascending primary key order):
CREATE TABLE [dbo].[Nodes](
	[ID] [int] IDENTITY(1,1) NOT NULL,
	[Parent] [int] NOT NULL,
	[Type] [int] NOT NULL,
	[PermissionGroup] [int] NOT NULL,
	[Name] [varchar](max) NOT NULL,
	[Status] [int] NULL,
 CONSTRAINT [PK_nodes] PRIMARY KEY CLUSTERED 
([ID] ASC))
ALTER TABLE [dbo].[Nodes]  WITH CHECK ADD  CONSTRAINT [FK_nodes_nodeTypes] FOREIGN KEY([Type])
REFERENCES [dbo].[NodeTypes] ([ID])
GO
ALTER TABLE [dbo].[Nodes] CHECK CONSTRAINT [FK_nodes_nodeTypes]

AartBluestoke:
also i fixed a couple of bugs in the PR, it now compiles. One of the unit tests fails in my code branch, in "LoadUpdateDataChange", but i don't think my code interacts with that, as i have only altered the loadData, and not touched the loadUpdate.

AartBluestoke:
sql server has a very strong performance penalty if you try to insert more than 1000 values in a single statement - (ie, 100 rows with 10 columns should be max in a single statement)
http://stackoverflow.com/questions/8635818/multiple-insert-statements-vs-single-insert-with-multiple-values

AartBluestoke:
further comments on performance vs batch size: 
https://www.simple-talk.com/sql/performance/comparing-multiple-rows-insert-vs-single-row-insert-with-three-data-load-methods/

performance gain for batches and optimal batch size shrinks as table width grows, but the 5x speedup reported above seems resonable; (was tested on sql server 2008 express)

nvoxland:
I merged the change into master since it is more of a change than I like to make in a patch release.

I added logic to LoadDataChange to only generate the InsertSetStatement with mssql, postgresql and mysql since those are known to work. I'll look into expanding the usage with the liquibase 4.0 changes (and put the compatibility check in InsertSetGenerator like it should be)

PR #432, opened 2015-05-21
Disable  DATABASECHANGELOG table creation when running offline

gquintana:
https://liquibase.jira.com/browse/CORE-2334


nvoxland:
Thanks! I think I'm going to change the flag passed around a bit, though. I'm thinking either making outputLiquibaseSql take values of "true", "false", or "data_only" or making an outputLiquibaseUpdateSql and an outputLiquibaseCreateSql with the old outputLiquibaseSql flag just set both of those to true. Leaning toward the different options for outputLiquibaseSql, though. I'm trying to wrap up a 3.3.4 release and then I'll look at it more closely.

gquintana:
I can change this PR if you wish, tell me what solution you prefer.

nvoxland:
Yes, those look like good changes. Thanks!

PR #434, opened 2015-05-21
include defaultFile arg for liquibase properties in $LIQUIBASE_HOME

colinlieberman:
This is a change I use locally for managing single properties for multiple lqb schema projects to keep command line commands shorter. Might be useful to others.


nvoxland:
I'm going to close this without merging because Liquibase should already be looking for ./liquibase.properties as a default defaultsFile and I don't want to break compatibility.

PR #435, opened 2015-05-25
Added reading of content of gzipped source files

padcom:
I created this pull request because the source input file that I am dealing with is over a gigabyte in size and it would be very good for us not to be storing it in an uncompressed form. This is the initial snapshot of the database and we would like it to be the first changset.


nvoxland:
Merged in, thanks!

PR #439, opened 2015-06-03
CORE-2236 Update test case to be aware of support for the yml extension

mches:
[CORE-2236](https://liquibase.jira.com/browse/CORE-2236) 

There are now 6 supported file extensions including "yml", so the test case needs to be updated


nvoxland:
Thanks

PR #441, opened 2015-06-08
Patch copy rows for edge case with no copy columns

zeddever:
If the copy rows function is called with no copy columns, you'll end up with an invalid SQL statement.

This occurs when you try to add a single column to a table in an SQLITE database. This occurs because to alter table with an SQLite database you create a temporary table and copy over the rows.

Note: In the future, you might want to update the alter table implementation for SQLite since SQLIte now supports the "ALTER TABLE Query"

This patch fixes the `liquibase: [SQLITE_ERROR] SQL error or missing database (near ")": syntax error)` bug


nvoxland:
Thanks!

PR #442, opened 2015-06-10
Put back echo for $LIQUIBASE_HOME

adriens:
Since removal of the echo, the liquibase command does not work properly at all : it is even not possible to get liquibase version

liquibase --version
/usr/bin/liquibase: 5: /usr/bin/liquibase: Syntax error: "else" unexpected 

This affects the debian version.

### DIFF
--- a/liquibase-debian/src/main/resources/liquibase
+++ b/liquibase-debian/src/main/resources/liquibase
@@ -1,7 +1,7 @@
 #! /bin/sh
 
 if [ -n "${LIQUIBASE_HOME+x}" ]; then
-# echo "Liquibase Home: $LIQUIBASE_HOME"
+  echo "Liquibase Home: $LIQUIBASE_HOME"
 else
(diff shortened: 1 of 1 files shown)


juancarloscruz:
Also found it ... [9e31fbe]

1:1 duplicate of [4629dc7]?

adriens:
yes, i think it is... it looks like a regresssion i thought was fixed a while.

adriens:
the same error on both debian and rpm shell script indeed.

juancarloscruz:
I see :+1:

adriens:
Thanks a lot juan.

nvoxland:
I fixed it instead by inverting the if logic and removing the else. I don't want the echo output when liquibase_home is set because people pipe updateSql out directly to a file and the echo message messes that up. 

The syntax error should be gone, now

adriens:
That makes sense. Thank you for patching.

PR #443, opened 2015-06-10
Put back echo for $LIQUIBASE_HOME

adriens:
Since removal of the echo, the liquibase command does not work properly at all : it is even not possible to get liquibase version

liquibase --version
/usr/bin/liquibase: 5: /usr/bin/liquibase: Syntax error: "else" unexpected

This affects the debian version.


nvoxland:
I fixed it instead by inverting the if logic and removing the else. I don't want the echo output when liquibase_home is set because people pipe updateSql out directly to a file and the echo message messes that up. 

The syntax error should be gone, now

adriens:
That makes sense. Thank you for patching.

PR #445, opened 2015-06-23
CORE-2397 Improve robustness of SQL generation for obtaining a view definition on MSSQL

mches:
[CORE-2397](https://liquibase.jira.com/browse/CORE-2397) MSSQL View Snapshot should not use sp_helptext

Improves robustness of SQL generation for obtaining a view definition on SQL Server. Supports SQL Server 2000 and later without using sp_helptext.


nvoxland:
Thanks for the improvement

PR #449, opened 2015-06-24
CORE-2405, CORE-2406 MSSQL data type bugs

mches:
[CORE-2405](https://liquibase.jira.com/browse/CORE-2405) Collation not preserved, depending on configuration
[CORE-2406](https://liquibase.jira.com/browse/CORE-2406) Escaped built-in data types should be lower case

Preserves collation and ensures compatibility with case-sensitive database collations


nvoxland:
Thanks!

PR #451, opened 2015-06-30
Avoid error message under Kubuntu

MHWallaby:
I got the message "./liquibase//liquibase: 39: ./liquibase//liquibase: [[: not found" under Kubuntu


nvoxland:
I think this should be fixed by https://github.com/liquibase/liquibase/pull/457

PR #452, opened 2015-07-02
CORE-2419 Support fluent/builder-style change properties

mches:
[CORE-2419](https://liquibase.jira.com/browse/CORE-2419) Support fluent/builder-style change properties


nvoxland:
Seem like good changes. I do like the fluent style better than void return types.

PR #453, opened 2015-07-06
Informix: Return null for connection schema name

gregjones60:
This fixes a problem with using Informix with Liquibase. The base class method causes a call to a non-existent stored procedure.


nvoxland:
Thanks

PR #454, opened 2015-07-06
Added support for commenting lines in csv files (by starting line wit…

rgrebski:
I needed to add some comments into CSV file, but there is no support for that. This pull request contains solution which makes it possible to add comment into CSV file.

All lines starting with # (hash sign) will be treated as commented lines.


nvoxland:
I like being able to have comments in the csv, but I worry that introducing a comment will break existing CSVs that have a line that happens to start with #. I think a safer approach would be to add a "commentDelimiter" attribute to loadData which is used instead of hard coding # 

Also, should not add org.apache.commons.lang.StringUtils as a runtime dependency.

rgrebski:
Breaking existing csv files was my only concern, but hash seems to be a common comment sign.
We could change comment sign ("#") to a different one or introduce maven plugin switch -DenableCsvComments (false by default - not to break existing csv files), whats your preference ?

Regarding StringUtils I found it on a classpath and i thought its ok to use it, I will re-write the code not to use StringUtils

nvoxland:
# seems common enough. I tried looking to see if there was any sort of standard with CSV but it looks like not.

StringUtils is in the classpath, but only used for the SDK currently. Dependency management will be improved with liquibase 4.0.

I think since the logic to include commenting or not is pretty specific to loadData, we should add it is a tag on loadData instead of a system property. You can do that by adding a new private field to LoadDataChange with get/set attributes. That should automatically be populated when the XML/YAML/JSON changelog is parsed. [Saw your edit after writing this, kept it in case it helps]

nvoxland:
It looks like the checksum value should not be affected by if the attribute is there or not, which is good.

I think the checksum value should depend only on the contents of the CSV, so:
- adding or removing the attribute when there are no comments in the file will not affect the checksum
- changing the attribute to make line(s) commented out should affect the checksum

With the checksum logic, I think it would be ok to set the default value of the new attribute to be "#" so it removes lines that start with # by default. That should be very rare, and if it happens to happen, the checksum will detect it and it can be handled correctly. I can add a note to the release announcement. Defaulting it to "#" will make it much easier in the standard case since you will not need to constantly include the attribute.

rgrebski:
Sounds good, I will implement it your way this weekend

rgrebski:
Ok updated, please review, especially:

1) LoadDataChange.setCommentLineStartsWith (I am ignoring null values, because null was passed when there was no commentLineStartsWith tag in loadData)
2) EmptyLineAndCommentSkippingInputStream - I created such input stream to fulfill checksum contract you mentioned in previous comment

Let me know if i missed something or if I should do it in a different way

nvoxland:
The fix looks good. I did have to merge it into master vs. 3.4.x because of the change to the XSD, that is something I try to not do in patch releases.

I also changed the setCommentLineStartsWith to set the "#" default rather than relying on the XSD default value because then it works as expected with json or yaml changelogs.

PR #455, opened 2015-07-07
CORE-2410 Snapshot should not include paramaters for MSSQL geometry, geography or sql_variant types

mches:
Reverts 8713973 (unneeded due to [CORE-2217](https://liquibase.jira.com/browse/CORE-2217), [CORE-2345](https://liquibase.jira.com/browse/CORE-2345))  while retaining cd69419 (needed for [CORE-2410](https://liquibase.jira.com/browse/CORE-2410))

Adds a test case that shows CORE-2410 is fixed


nvoxland:
Yes, I think it was duplicated work from an independent branch. Thanks

PR #457, opened 2015-07-15
Fix [[ not found error -> [[ is bash-builtin

EVODelavega:
Suggested fix: use /usr/bin/env bash instead of /bin/sh

using /bin/sh yields error output:

```
liquibase: 39: liquibase: [[: not found
```

### DIFF
--- a/liquibase-core/src/main/resources/dist/liquibase
+++ b/liquibase-core/src/main/resources/dist/liquibase
@@ -1,4 +1,4 @@
-#! /bin/sh
+#!/usr/bin/env bash
 
 if [ ! -n "${LIQUIBASE_HOME+x}" ]; then
   # echo "LIQUIBASE_HOME is not set."


EVODelavega:
A POSIX-compliant (`#!/usr/bin/env sh`) compliant approach would be this:

```sh
pattern_match () {
    case "$1" in
        $2)
            return 0
            ;;
        *)
            return 1
            ;;
    esac;
}
if pattern_match $(uname) 'MINGW*'; then
    CP_SEPARATOR=";"
else
    CP_SEPARATOR=":"
fi
```

If needs must, I'll change the PR to keep the POSIX shell subset, and use an inline `case`, or the function above instead.

nvoxland:
I think the simple env bash option will work well enough, so I merged that in. If people run into problems I can look at the more complex POSIX approach. Thanks for writing both up.

PR #458, opened 2015-07-16
CORE-2438 DeleteGenerator does not handle parameter names and values with $ or \ properly

mches:
[CORE-2438](https://liquibase.jira.com/browse/CORE-2438) Quote special characters \ and $ in regex replacement strings


mches:
[PR#461](https://github.com/liquibase/liquibase/pull/461) is related.

nvoxland:
I think it seems like a good fix, thanks.

PR #459, opened 2015-07-22
CORE-2385 fix handling of null basePath in FilenameUtils.concat

pkozelka:
When concatenating *something* to `null` base path, the method `FilenameUtils.concat` should return that *something*, rather than `null`.
At least it is used that way in [SpringLiquibase.SpringResourceOpener.list](https://github.com/liquibase/liquibase/blob/master/liquibase-core/src/main/java/liquibase/integration/spring/SpringLiquibase.java#L138), as the [issue CORE-2385](https://liquibase.jira.com/browse/CORE-2385) indicates.


nvoxland:
Merged https://github.com/liquibase/liquibase/pull/460 into 3.4.x, will get this fix in master after merge from 3.4

PR #460, opened 2015-07-22
CORE-2385 fix handling of null basePath in FilenameUtils.concat - backport to 3.4.x

pkozelka:
we need the same fix like #459 in the `3.4.x` branch so that we can use next minor release.

BTW I created a local release for the meantime and published it here: https://github.com/pkozelka/liquibase/releases/tag/liquibase-parent-3.4.0-pkozelka-1


o-orand:
@pkozelka Thanks for the fix.

PR #461, opened 2015-07-22
CORE-2442 Creating MD5 checksum fails if changeSet id contains the character "?"

mches:
[CORE-2442](https://liquibase.jira.com/browse/CORE-2442) Make UpdateGenerator more resilent to parameter names and values that contain the pattern to match


mches:
[PR#458](https://github.com/liquibase/liquibase/pull/458) is related.

PR #464, opened 2015-07-23
CORE-2452 Quote index name on SQL Server

mches:
[CORE-2452](https://liquibase.jira.com/browse/CORE-2452) Quote index name on SQL Server


mches:
Merge to wrong branch

PR #467, opened 2015-07-25
CORE-2456 includeObjects/excludeObjects not work for UniqueConstraint…

huygun:
….  Change filter change to check for UniqueConstraints.


nvoxland:
looks good, thanks

PR #468, opened 2015-07-28
Liquibase-Hibernate-Extension

klopfdreh:
This PR provides a new functionality to start a liquibase changelog update before the actual hibernate validation (hibernate.hbm2ddl.auto -&gt; validate) by using org.hibernate.integrator.spi.Integrator.

Usage is (see javadoc in class):
- Create a File in META-INF and add the Implementation as a Service
- Create a class nearly the same way as in liquibase-cdi and provide the datasource, etc.

Additional features:
- If you have a multi module project with several subprojects updating the same database schema use LiquibaseHibernateIntegratorConfig.setMultiProjectSetup(boolean)
- If you want to drop the database schema afterwards the hibernate tear down use LiquibaseHibernateIntegratorConfig.setDropAtShutdown(boolean)


liquibase-hibernate/pom.xml
```diff
@@ -0,0 +1,87 @@
...
+                <artifactId>maven-javadoc-plugin</artifactId>
+                <executions>
+                    <execution>
+                        <id>api</id>
+                        <phase>prepare-package</phase>
+                        <goals>
+                            <goal>javadoc</goal>
+                        </goals>
+                        <configuration>
+                            <sourcepath>../liquibase-hibernate/src/main/java</sourcepath>
```
martin-g: Maybe you should use `${project.sourceDirectory}` (see https://maven.apache.org/pom.html)

liquibase-hibernate/src/main/java/liquibase/integration/hibernate/AbstractHibernateLiquibaseIntegrator.java
```diff
@@ -0,0 +1,392 @@
...
+ * Example:
+ * 
+ * <pre>
+ * <code>
+ * <b>Java-Class:</b>
+ * public class LiquibaseIntegrator extends AbstractHibernateLiquibaseIntegrator {
+ * 
+ *     {@literal @}Override
+ *     public ResourceAccessor create() {
+ * 	return new ClassLoaderResourceAccessor(getClass().getClassLoader());
```
martin-g: add some whitespace in the start

klopfdreh:
Thank you Martin! I also noticed that I used the wrong parent version number. I fixed that, too.

dmitrygusev:
FYI: We did something similar with Liquibase & Hibernate/JPA, but instead of implementing a `Integrator` we've build a wrapper around `LiquibaseServletListener`:
https://github.com/anjlab/anjlab-tapestry-commons/blob/133bc99958f506ea620c4096d093b7e5ab72eae6/anjlab-tapestry-liquibase/src/main/java/com/anjlab/tapestry5/services/liquibase/AutoConfigureLiquibaseDataSourceFromHibernateJPAModule.java#L71

There we have access to contents of `persistence.xml` and we're building a datasource manually based on settings from that file.

klopfdreh:
Good to see that there is need for such an integration. Following points I saw in your code:
* The data source is going to be bound in the ServletListener which should be only received from the JEE container which is providing and binding it.
* There are a lot of depdencies to tapestry and I think if you would choose my solution tapestry would not be involved into the hibernate startup process, but only receive the datasource lateron.
* I build up a multi module startup, so if you have 2 wars starting up and trying to write into the same schema it is handled and will not lead to some errors. (like exception while creating the tables DATABASECHANGELOG / DATABASECHANGELOGLOCK)
* The database can be cleared (maybe wanted for developer environments)

dmitrygusev:
@klopfdreh if the datasource can be provided by container then we may simply use `LiquibaseServletListener ` as is. And Hibernate in this case will likely to use the same datasource.

However you may want not to use container-provided datasource and configure DB connection in `persistence.xml`, this is where that code should fit. Since `LiquibaseServletListener` can only work with JNDI datasources that code temporarily binds one just for this.

This is true, that there are a lot of tapestry dependencies, but that's a tapestry integration, that's the point :) Just wanted to show the approach, it should work for others too.

This is true that multi-module use-case isn't covered there, we didn't need it.

And for development environments we're using this plugin:
https://github.com/tlberglund/gradle-liquibase-plugin

klopfdreh:
Okay, maybe we can change the createConfig / createDataSource / create methods and add the arguments of org.hibernate.cfg.Configuration / org.hibernate.metamodel.source.MetadataImplementor so that we have access to the hibernate configuration / implementor.

klopfdreh:
Now the datasource can be created by receiving the properties from the hibernate configuration. Example:

http://www.tutorialspoint.com/hibernate/hibernate_configuration.htm

```code
hibernate.connection.url
hibernate.connection.username
...
```

and then create a datasource with it.

http://commons.apache.org/proper/commons-dbcp/api-1.4/org/apache/commons/dbcp/BasicDataSource.html

```java
BasicDataSource datasource = new BasicDataSource();
datasource.setUsername(configuration.getProperty("hibernate.connection.username"));
...
```

dmitrygusev:
@klopfdreh right, another example would be to use the same connection provider as in hibernate properties, this is in case if datasource is already configured there which is usually the case:

``` java
    String connectionProviderClassName =
        configuration.getProperty(
            org.hibernate.cfg.AvailableSettings.CONNECTION_PROVIDER);

    Class<?> connectionProviderClass = Class.forName(connectionProviderClassName);

    Object connectionProvider = connectionProviderClass.newInstance();

    ((org.hibernate.service.spi.Configurable) connectionProvider)
        .configure(configuration.getProperties());

    DataSource dataSource =
        ((org.hibernate.service.spi.Wrapped) connectionProvider)
            .unwrap(javax.sql.DataSource.class);
```

klopfdreh:
@dmitrygusev thank you for the example :+1:

klopfdreh:
Liquibase updates can be activated / deactivated by overriding "isAutoUpdateActive", now:

```java
protected boolean isAutoUpdateActive() {
   return Boolean.parseBoolean(System.getProperty("liquibase.update"));
}
```

klopfdreh:
@nvoxland Anything I can do to get this PR integrated?

klopfdreh:
I am going to close this PR, because I see not activity of the devs here.

PR #470, opened 2015-07-30
CORE-2466 Rollback referencing a change set in another file cannot be parsed

mches:
[CORE-2466](https://liquibase.jira.com/browse/CORE-2466) Set the change log of the included change sets to the referencing change log


mches:
Wrong branch

PR #471, opened 2015-07-30
CORE-2466 Rollback referencing a change set in another file cannot be parsed

mches:
[CORE-2466](https://liquibase.jira.com/browse/CORE-2466) Rollback referencing a change set in another file cannot be parsed

Makes it possible to resolve the root change log from within a nested change log. Then the root change log can be used to resolve a change set referenced in a rollback node.

Maintains the original change set reference for rollback changes that are included by reference, so that paths relative to the referenced change log file can be resolved correctly.


nvoxland:
The fix looks good, thanks. I was concerned the fix would be harder, glad it wasn't.

PR #481, opened 2015-08-13
Fix and unify where parameters handling

Vampire:
I would have refactored out the where parameters handling from all three occurrences into a method of a helper class, but I was not sure which helper class is suitable for this.


mches:
Thanks for finding and fixing the bug. I have some test case improvements and additions [here](https://github.com/Vampire/liquibase/pull/1) that verify where parameter handling is both fixed and unified across the the 2 SqlGenerator implementations and 1 ExecutablePreparedStatement implementation.

nvoxland:
Merged, thanks for the fix and the additional test cases in the other pull request

PR #482, opened 2015-08-13
MSSQL snapshot did not include primary keys with default schema

evigeant:
In SQLServer, I had a precondition checking for the existence of a primary key which did not behave correctly. I traced it down to the snapshot which did not find any primary keys.

When using the default schema (null), the query to fetch primary keys had a where clause like this:

    WHERE [s].[name] = N'null'

so it failed as the default schema is 'dbo'


nvoxland:
Thanks!

PR #483, opened 2015-08-13
Fixed: Absolute filename as valueLobFile would never be found by ClasspathResourceAccessor

evigeant:
There were checks that were based on java.io.File class which always failed when the resource was on the classpath.


nvoxland:
Thanks

PR #484, opened 2015-08-13
Fix error with postgres throwing an exception with an offline connection and Refactor getConnectionSchemaName to prevent code errors

viftodi:
By default getConnectionSchemaName calls an sql statement that obtains
the current schema. Not only does this force a custom database to copy
the code just so it can modify the call statement, but it is also error
prone as when the base implementation evolves the changes don't occur in
overridden classes. Such error occurred when the base
getConnectionSchemaName has changes so it would return null in the case
of an OfflineConnection but PostgresDatabse still used old code and
would fail on an offline connection.
This evolution adds the getconnectionSchemaNameCallStatement that can be
overridden by databases that only need to customise the statement, that
way the logic is shared with the abstract class.

The following databases have been refactored to use the new method:
- MSSQLDatabase
- PostgresDatabase
- SybaseDatabase


nvoxland:
Looks like a good change, thanks

PR #485, opened 2015-08-17
Fix XML for diffChangeLog if empty element

talklittle:
A self-closing `<databaseChangeLog/>` XML element,
such as the empty one generated by `generateChangelog`,
would cause malformed XML when `diffChangeLog` was run.

To reproduce the original bug, run `generateChangelog` immediately followed by `diffChangeLog`.
`generateChangelog` will correctly produce an empty `<databaseChangeLog/>` XML element. However, prior to this commit, `diffChangeLog` fails to overwrite the self-closing `<databaseChangeLog/>` with an opening tag.

(Java 1.8.0_51 on Windows.)


nvoxland:
Thanks! Created https://liquibase.jira.com/browse/CORE-2502 to track the fix

talklittle:
@nvoxland Great, thank you.

PR #486, opened 2015-08-20
Only consider the current file when listing unexpected changes

gidoca:
When running the unexpectedChangeSets command, currently all changesets not in the current file are shown. Instead, we propose to show only change sets
whose recorded filename matches the current change log file's logical
filename. This is a fix for CORE-2496.

### DIFF
--- a/liquibase-core/src/main/java/liquibase/Liquibase.java
+++ b/liquibase-core/src/main/java/liquibase/Liquibase.java
@@ -2,12 +2,7 @@ package liquibase;
 
 import java.io.*;
 import java.text.DateFormat;
-import java.util.Arrays;
-import java.util.Collection;
-import java.util.Date;
-import java.util.HashSet;
-import java.util.List;
-import java.util.Set;
+import java.util.*;
 
 import javax.xml.parsers.ParserConfigurationException;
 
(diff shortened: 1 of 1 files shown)


nvoxland:
Are you running multiple independent changelogs that all use the same schema/databasechangelog table?  

The problem with only listing unexpected changeSets in the current changelog file is that you can have include-ed changelog files which report themselves as different paths but are really part of the same changelog and I think should be included. 

Are you able to determine the complete list of include-ed paths and only exclude paths that are not in that list?

gidoca:
Thank you for looking at this. Our setup is such that we have three domain schemas and a liquibase schema, and we use a separate changelog for each domain schema, but the changes are all stored in a single schema called liquibase. In other words, we use essentially `--defaultSchemaName=domainschema<n> --liquibaseSchemaName=liquibase --changeLogFile=domainschema<n>.xml` We do this because we didn't want to mix the domain and liquibase tables. 

What we would like to do is to verify that the database schema is not newer than the application. To do this, we are trying to check that there are no unexpected changes for any of our changelog files. However, reading your message, it occured to me that we could just have a changelog which includes the other three, which should solve our immediate problem. However, I still feel that the current behaviour is less-than-intuitive, but I see that a solution which takes into account includes would be quite a bit more complex.

gidoca:
...actually, no, having a changelog which includes the others doesn't help. One of the reasons to have a separate changelog file for each schema was so that we wouldn't have to explicitly name the schema in each change. However, unfortunately, you can't set defaultSchemaName in the changelog file, only on the command line.

filipelautert:
Hello @gidoca  - does this still a problem for you? I see that this PR is from pre-liquibase 3.5.0 and the current code has been moved into UnexpectedChangesetsCommandStep class. The output for this command seems to have changed a lot within time too. 
I'll close this PR and if you still need this feature please open a new issue so we could think about it and implement it (maybe adding a new flag to this command so we don't follow includes?) 

Thanks

PR #489, opened 2015-08-25
Sequence Support for LoadData

lazaronixon:
Today is possible load csv into tables with sequence columns this way:
```csv
id_pais|cod_bacen|descricao
SEQNAME.NEXVAL()|1234|BRASIL
SEQNAME.NEXVAL()|1234|ARGENTINA
```
```xml
<loadData tableName="pais" file="liquibase/seeds/pais.csv" separator="|" >
    <column type="computed" name="id_pais"/>
    <column type="numeric" name="cod_bacen" />
    <column type="string" name="descricao" />
</loadData>
```
but it is not multidatabase compatible.. because this I propose this change:
```csv
id_pais|cod_bacen|descricao
0|1234|BRASIL
0|1234|ARGENTINA
```
```xml
<loadData tableName="pais" file="liquibase/seeds/pais.csv" separator="|" >
    <column type="sequence" name="id_pais" defaultValue="pais_seq"/>
    <column type="numeric" name="cod_bacen" />
    <column type="string" name="descricao" />
</loadData>
```

### DIFF
--- a/liquibase-core/src/main/java/liquibase/change/core/LoadDataChange.java
+++ b/liquibase-core/src/main/java/liquibase/change/core/LoadDataChange.java
@@ -234,6 +234,9 @@ public class LoadDataChange extends AbstractChange implements ChangeWithColumns<
                             } else if (columnConfig.getType().equalsIgnoreCase("COMPUTED")) {
                                 liquibase.statement.DatabaseFunction function = new liquibase.statement.DatabaseFunction(value.toString());
                                 valueConfig.setValueComputed(function);
+                            } else if (columnConfig.getType().equalsIgnoreCase("SEQUENCE")) {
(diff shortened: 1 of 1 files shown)


nvoxland:
Related issue: https://liquibase.jira.com/browse/CORE-2260

nvoxland:
Made some minor changes, see https://liquibase.jira.com/browse/CORE-2540

d3ns0n:
I tried to use Sequences in CSV files but didn't manage to get it to work. As long as I use "updateSQL" it generates correct SQL which looks like this: 
`
INSERT INTO principal (id, name, enabled, comment) VALUES (nextval('hibernate_sequence'), 'admin', TRUE, 'Superuser');
`
but when I use "update" there will be NULL instead of the correct value from the sequence
`
Batch entry 0 INSERT INTO principal(id, name, enabled, comment) VALUES(NULL, 'admin', 'TRUE', 'Superuser') was aborted: ERROR: null value in column "id" violates not-null constraint
`

PR #493, opened 2015-08-31
onUpdate="RESTRICT" is to be ignored on MS SQL Server like it is done for onDelete already.

t-beckmann:
MS SQL Server does not understand the syntax but behaves as intended. A restrict-on-update means one must not violate the foreign key constraint which is what is enforced by the constraint anyways. The RESTRICT keyword is required for some databases, I think MySQL, which support different alternatives. Without the change <addForeignKeyConstraint onUpdate="RESTRICT"/> is doomed to fail.


nvoxland:
Thanks, fixes https://liquibase.jira.com/browse/CORE-575

PR #501, opened 2015-09-14
CORE-2059 escapeObjectName has no respect for quotingStrategy on Postgres

vfpfafrf:


nvoxland:
Thanks. I merged the change into master rather than 3.4 since quoting changes is something I try to avoid in patch releases.

PR #505, opened 2015-09-17
CORE-2541 - added support for registering a change exec listener class on command…

pellcorp:
… line, as well as injecting Database and/or Properties into the listener


nvoxland:
Thanks. I merged in the change, but into master for 3.5 since I try to keep the 3.4.x releases as minor patch releases.

I also changed ChangeExecListenerAdapter to AbstractChangeExecListener and made it abstract to better match the naming of similar patterns in the codebase.

PR #506, opened 2015-09-18
CORE-2537 - add support for 'TODAY' literal in date fields in CSV data files

bsandiford:


nvoxland:
I like the idea but I'd like it to be a bit more generic, so instead of just TODAY+number of days, it's NOW + # units like "NOW + 1 day" or "NOW + 4 hour"  

If you don't get a chance to make the changes, I'll look at doing it with Liquibase 4

bsandiford:
Hi, Nathan.  What about supporting both?  A 'TODAY' base would have no 'time' component, but a 'NOW' would?  i.e. the way we use this is for just Date columns.

Anyways - I wouldn't have time in the next little while to make that type of change, sorry. :(

Bob.

bsandiford:
Nathan - I'm back and will have a number of enhancements to offer.  Meanwhile - this one is still pending.  Are you OK with the "TODAY{+/-n}" approach, in addition to a "NOW{+/-n u" where "n" is number and "u" is units - initially I'm thinking units could be "m" / "minute" / "minutes", "h" / "hour" / "hours", or "d" / "day" / "days"?  And - I'll probably shoot this PR and create another one based of current master...  Does that sound OK?  Thanks!

nvoxland:
I like your proposal of both TODAY+int and NOW+int unit. I'll add that comment to the pull request. If you have the today part working and don't want to implement the NOW at this point, that can always be a feature feature.

bsandiford:
Closing since it's several years old, and not complete (only handles TODAY, not NOW).  Will create a new on on current code base, with full implementation.

PR #507, opened 2015-09-18
CORE-2299 - bugfix - apply 'errorIfMissingOrEmpty' attribute for 'includeAll' to dbchangelog-3.5.xsd.

bsandiford:
(Was in dbchangelog-3.4, but not dbchangelog-3.5)


nvoxland:
Thanks

PR #508, opened 2015-09-21
CORE-2539 - CSV # of elements must exactly match # of headers.

bsandiford:
- Also - fix the line number display when they don't match (was off by 1).


nvoxland:
Thanks!

PR #509, opened 2015-09-21
CORE-2540 - Database Independent 'Sequence' assignment

bsandiford:
CORE-2260 - Ability to specify which sequence to use with loadData

CORE-2540 has the info about how it works from the user's point of view.


lazaronixon:
#489 Same here...

nvoxland:
Fixed by #489

PR #510, opened 2015-09-22
CORE-2542 added support for specifying clob / blob in loadData

pellcorp:


nvoxland:
Merged into 3.5 since it's more than I'd like to do in a 3.4 patch release. Thanks!

PR #511, opened 2015-09-23
Ear changelog order

jimmycd:
pom in oarent and core: easyMock is missing (added for scope test)

includeAll Function:
by ignoring "WEB-INF/classes/" in path of all found (and 3evtl. already filtered) changelog Files independent whether they are in a WAR or in a JAR are ordered following the same rule. So in case of using an EAR as source for db-management and when habing distributed changelogs all over the ear the disired alphanumerical order is given for all files as desired. Without the fix first of all files found in WEB-INF/classes/ are processed and then all jar's. But that is not allwayss desirable.
In case somebody relieas on the legacy order logic, he has to change the naming of the changelog files.

**Further solution is in work:** providng an additional tag in analogy of Interface *IncludeAllFilter* for defining the Comparator Class for custom ordering.


nvoxland:
Merged pull request. Thanks!

PR #512, opened 2015-09-23
Filter include all configurable comparator

jimmycd:
New Attribute *resourceComparator* in **includeAll**-Tag for passing a class implementing Comparator<String> for free controlling of ordering of changelog files in bigger bundles. If no comparator is given, default implentation is taken.


nvoxland:
Thanks!

nvoxland:
I moved the the XSD change to 3.5 from the 3.4 xsd

mches:
For completeness, `liquibase.changelog.ChangeLogInclude` and `liquibase.changelog.ChangeLogIncludeAll` should be updated with the new attribute

PR #513, opened 2015-09-24
Fix schema name generation (again)

gregjones60:
I have finally had a chance to revisit this problem. The original fix made it easy to generate the tables for the first time but it still did not support upgrading. I have fixed the getConnectionSchemaName method now to get the schema for the current session.

This has enabled us to upgrade our current database for Keycloak with the changes for the most recent version. 

There is still an issue whereby the Informix database cannot be in ANSI mode, due to the various combinations of upper- and lower-case conversion and case sensitivity (not helped by the JDBC driver). I think, though, that this will fix the problem for the majority of cases.


nvoxland:
Thanks!

PR #519, opened 2015-10-05
CORE-910: Handle MSSQL returning rowcounts of -1 when NOCOUNT enabled

ahus1:
This is trying a pragmatic fix for this MSSQL NOCOUNT issue as no row count is available here. 

In my installation with MS SQL Server 2008 R2 the update always returns -1 when NOCOUNT is enabled.

This leads to the situation that the lock of the row succeeds, and the unlock fails. 

This fix makes both the lock and unlock succeed and will print a warning message.

As this code still accepts a "-1" when locking the row, this might lead to a race condition when locking the database for a migration. But this is already the case with the current setup.

The alternative would be to issue a "set nocount on" once a "-1" is received and to retry the statement. But this could be a second ticket.

Please let me know if this works for you.

Best regards,
Alexander

https://liquibase.jira.com/browse/CORE-910

### DIFF
--- a/liquibase-core/src/main/java/liquibase/lockservice/StandardLockService.java
+++ b/liquibase-core/src/main/java/liquibase/lockservice/StandardLockService.java
@@ -196,6 +196,13 @@ public class StandardLockService implements LockService {
 
                 executor.comment("Lock Database");
                 int rowsUpdated = executor.update(new LockDatabaseChangeLogStatement());
+                if (rowsUpdated == -1) {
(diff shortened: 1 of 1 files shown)


ahus1:
I decided to do a safe implementation and to re-issue statement with "set nocount off" once a "-1" is received. The code is updated and merged into a single commit.

nvoxland:
I merged this in with a change to use @mches 's sp_executesql block instead of calling nocount on then off. Seemed safer and less likely to cause side effects.

PR #521, opened 2015-10-09
logging fix prevents stack trace in user definable log

logemann:
logging fix prevents stack trace in user definable log. The exception which reaches the appServer will stacktrace to stdout anyway.


nvoxland:
Thanks. 

Yes, it should be fine to remove that since the actual stacktrace is re-thrown and logged later.

PR #522, opened 2015-10-12
CORE-155 Add ability for a change set to inherit context expressions from parent containers (WIP)

mches:
Adds ability for a change set to inherit context expressions via databaseChangeLog, include or includeAll parent containers

[CORE-155](https://liquibase.jira.com/browse/CORE-155) Default-context attribute in databaseChangeLog tag

The context expressions from the parent containers are inheritable and ANDed together logically.

Example usage:

_changelog-master.xml:_
```xml
<databaseChangeLog>
    <include file="changelog-included.xml" context="d and e"/>
    <includeAll ... context="f or g"/>
</databaseChangeLog>
```
_changelog-included.xml:_
```xml
<databaseChangeLog ... context="a and (b or c)">
    <!-- logically, the context expression for this change set is 
         "(d and e) and (a and (b or c)) and (h or j or k)", 
         assuming changelog-master.xml is the root change log
    -->
    <changeSet ... context="h or j or k">...</changeSet>
    <changeSet ...>...</changeSet>
    <changeSet ...>...</changeSet>
</databaseChangeLog>
```

### DIFF
--- a/liquibase-core/src/main/java/liquibase/ContextExpression.java
+++ b/liquibase-core/src/main/java/liquibase/ContextExpression.java
@@ -162,4 +162,19 @@ public class ContextExpression {
     public boolean isEmpty() {
         return this.contexts == null || this.contexts.size() == 0;
     }
+
+    public static boolean matchesAll(Collection<ContextExpression> expressions, Contexts contexts) {
+        if (expressions == null || expressions.isEmpty()) {
+            return true;
+        }
+        if (contexts == null || contexts.isEmpty()) {
+            return true;
+        }
+        for (ContextExpression expression : expressions) {
+            if (!expression.matches(contexts)) {
+                return false;
+            }
+        }
+        return true;
+    }
 }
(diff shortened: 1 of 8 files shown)


mches:
Rebased to resolve merge conflicts

nvoxland:
This does look like a good feature and implementation, I assume you haven't ran into any other issues yet?

mches:
@nvoxland Yes, we've been using it in Production for awhile. The only limitation is that the `CONTEXTS` column on the `DATABASECHANGELOG` table is not incorporating inherited contexts. As far as I know, that's just an informational column, but users might expect it to have the inherited contexts. In a complex case where there are multiple context expressions to combine, I would expect it to have all context expressions in parenthesis and ANDed together. See my XML comment in the above example. I haven't had time to implement that and hadn't started. I also didn't have time yet to implement the same functionality for `labels` or `dbms` which I thought would be appropriate, but nothing is broken and the implemented functionality stands on its own.

PR #523, opened 2015-10-14
CORE-1167 Changed validCheckSum behavior to make more sense...

instantdelay:
[CORE-1167](https://liquibase.jira.com/browse/CORE-1167) Changed validCheckSum behavior to make more sense and match
the original documentation. By default, each validCheckSum now specifies
a stored checkSum value that will be considered valid. This allows a
changeSet to be altered and still pass validation.

The old behavior can still be used by setting the legacyCheckSumCompare
configuration property to true.

----

See the reasoning for this change here: https://liquibase.jira.com/browse/CORE-1167
I will try to give another example below.

Say I create a changeSet to update the value of a column

    <changeSet id="update-some-value" author="svanhoose">
        <update tableName="WIDGET">
            <column name="PRICE" valueComputed="special sql" />
        </update>
    </changeSet>

This goes into production and it works for 90% of my users on Oracle but fails for one user on Postgres. I look into the problem and find that my update SQL was bad so I fix it to be DB-agnostic.

Now I want to fix the changeSet for the next release. It already ran successfully for 90% of my customers and they should not have to do anything. I want the updated changeSet to run for the 10% of customers where it previously failed.

Essentially, "**If this changeSet already ran successfully and the old checksum matches the provided value, everything is okay. Otherwise run the changeSet or report a validation error.**"

    <changeSet id="update-some-value" author="svanhoose">
        <validCheckSum>7:2cad55a6ddef7f531bc0bdb867929d89</validCheckSum>
        <update tableName="WIDGET">
            <column name="PRICE" valueComputed="improved special sql" />
        </update>
    </changeSet>

Except that this doesn't work. The current behavior of Liquibase is such that validCheckSum is checked against the **newly computed checksum.** This behavior is logically equivalent to:

    <changeSet id="update-some-value" author="svanhoose">
        <ignoreDatabaseChecksum>true</ignoreDatabaseChecksum>
        <update tableName="WIDGET">
            <column name="PRICE" valueComputed="improved special sql" />
        </update>
    </changeSet>

I don't want to ignore the stored database checksum entirely. I only want to ignore one specific value.

----

Apologies for the long message... I'm really trying to make sure this is clear since the previous discussion of this issue in [CORE-1112](https://liquibase.jira.com/browse/CORE-1112) ended with the documentation being updated instead of the bug being fixed.

Let me know if you think I should submit this against a different branch.

### DIFF
--- a/liquibase-core/src/main/java/liquibase/changelog/ChangeSet.java
+++ b/liquibase-core/src/main/java/liquibase/changelog/ChangeSet.java
@@ -9,6 +9,8 @@ import liquibase.change.DbmsTargetedChange;
 import liquibase.change.core.EmptyChange;
 import liquibase.change.core.RawSQLChange;
 import liquibase.changelog.visitor.ChangeExecListener;
+import liquibase.configuration.GlobalConfiguration;
+import liquibase.configuration.LiquibaseConfiguration;
 import liquibase.database.Database;
 import liquibase.database.DatabaseList;
 import liquibase.database.ObjectQuotingStrategy;
@@ -832,6 +834,7 @@ public class ChangeSet implements Conditional, ChangeLogChild {
                 return true;
             }
         }
+
         CheckSum currentMd5Sum = generateCheckSum();
         if (currentMd5Sum == null) {
             return true;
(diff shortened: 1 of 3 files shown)


nvoxland:
Instead of merging this in, I decided to just accept both currentMD5Sum and storedCheckSum.

PR #524, opened 2015-10-14
fix escaping of column, table and schema on ColumnExistsPrecondition

hoegertn:


axelhodler:
:+1:

without the above the `<columnExists>` Precondition will not work on Oracle tables and columns that are lowercase. These have to be escaped if they are not uppercase.

nvoxland:
Thanks!

PR #526, opened 2015-10-17
Added databaseChecksumsCompatible variable.

Willem1987:
When the checksums are incompatible upgradeChecksums should be ran (in case of updateSQL command this gets skipped because the set MD5SUM null command is only written to sql file)
Because upgradeChecksums is not directly executed for updateSQL command, ValidatingVisitor should consider checksum valid if databaseChecksumsCompatible is false.


nvoxland:
Fixes CORE-2554  Thanks!

PR #527, opened 2015-10-30
CORE-2579 don't drop Oracle system sequences

magnusr:
In Oracle 12c, creating a table with an autoincrement column makes Oracle also
create a system-generated sequence for that column. Such sequences cannot be
dropped, but only purged, since undropping the table would otherwise fail.

Attempting to drop a system-generated sequence results in an ORA-32794 error,
which would make the dropAll command fail. This change makes dropAll skip all
sequences that are system objects.


liquibase-core/src/main/java/liquibase/database/AbstractJdbcDatabase.java
```diff
@@ -764,6 +765,13 @@ public void dropDatabaseObjects(final CatalogAndSchema schemaToDrop) throws Liqu
                         if (change instanceof DropTableChange) {
                             ((DropTableChange) change).setCascadeConstraints(true);
                         }
+                        if (change instanceof DropSequenceChange) {
```
magnusr: Actually, this does not seem to be needed.

magnusr:
Squashed the fixup

PR #529, opened 2015-11-16
Run on change fix, binary and nclob types resolving

sergseven:


bsnisar:
+1 to this functional!

brucemcbruce:
Awesome! We've just discovered an issue where "runOnChange=true" will re-run the changeset every time, if it has been changed at least once. After investigating we found that it was because liquibase was comparing the checksum of the oldest matching change, rather than the most recent. Hopefully this commit will fix it!

sergseven:
I hope developers will approve this pull request, even created issue in Jira, please vote for https://liquibase.jira.com/browse/CORE-2641

Tradunsky:
:thumbsup: +1

tomurka:
+1 definitely!

iherasymenko:
+1

nvoxland:
The big question is why changeSets are being marked as executed twice. That shouldn't be happening, does anyone have an example of when they are getting a changeSet in the databasechangelog table twice?

brucemcbruce:
I've added an example to the jira ticket [here](https://liquibase.jira.com/browse/CORE-2641?focusedCommentId=28795&page=com.atlassian.jira.plugin.system.issuetabpanels:comment-tabpanel#comment-28795)

nvoxland:
I'm not sure how we can get duplicate rows int he databasechangelog, but merged in this pull request to handle the case when it somehow happens.

I did change the implementation so it uses dateexecuted then orderexecuted since that is the correct sort order. 

I kept in the fix for mysql nclob -> longtext but took out the oracle binary -> raw since there is an actual binary type in oracle. Why were you wanting the binary -> raw change?

iherasymenko:
@nvoxland I'm using Oracle 12c and can't find BINARY type.
`Error report:
SQL Error: ORA-00907: missing right parenthesis
00907. 00000 -  "missing right parenthesis"
*Cause:    
*Action:
Error starting at line 1 in command:
CREATE TABLE "TEST_TABLE2" 
   (	"test_id" BINARY(16)
	
   ) 
Error at Command Line:2 Column:22
Error report:
SQL Error: ORA-00907: missing right parenthesis
00907. 00000 -  "missing right parenthesis"
*Cause:    
*Action:
(here "binary" was replaced to RAW)
table "TEST_TABLE2" created.
`

iherasymenko:
@nvoxland please take a look at http://docs.oracle.com/database/121/SQLRF/sql_elements001.htm#SQLRF0021
there is no BINARY datatype.

nvoxland:
OK, thanks. Not sure what list I was looking at that listed binary as a type. RAW does seem to be the closest match. I pushed that change up as well.

savinov:
The patch from @sergseven has been replaced after merge: https://github.com/liquibase/liquibase/pull/529/commits/a272d9ffd6a796237fad1465ed13cd726a3be617

Current version of the filter initializes a map with oldest changeset row instead of the most recent which is still incorrect.

Maybe condition should be changed to:
```
if (comparedDates > 0) {
    addToSet = true;
} 
```

PR #530, opened 2015-12-03
Adding support to the 3.4.x branch

TooDizzy:
Hi,

Please add this feature set to the core project.
It contains support for for Oracles Clob and Blob datatypes. Test cases are updated and I have tested both on a Oracle XE.


TooDizzy:
Just noticed that a test case is failing. Will look into this and create another pull request.

PR #532, opened 2015-12-05
Fix for CORE-2610

mozvip:
Fix for the CORE-2610 issue related to loadData : https://liquibase.jira.com/browse/CORE-2610


nvoxland:
Thanks!

PR #535, opened 2015-12-12
3.4.x

iihero:
When mapping in the changeset definition, it should be mapped
into "datetime" column type.
Fix it here.


nvoxland:
Thanks!

PR #538, opened 2015-12-22
Update RanChangeSet to handle classpath prefix

jdenly:
Currently, when running Liquibase as part of Spring Boot, having changesets set to runOnChange and loading configuration from the classpath causes the changeset to be run every time once it has changed once and creates a new EXECUTED line in the DATABASECHANGELOG, rather than a RERAN line. This is caused by the RanChangeSet.isSameAs method not stripping classpath: from both entries, meaning they are always identified as not the same when they are.


nvoxland:
Thanks!

PR #540, opened 2015-12-22
FIxing NPE when object name is null

Panthro:
Happens in this case:

https://github.com/liquibase/liquibase-hibernate/issues/101


nvoxland:
Thanks!

PR #541, opened 2015-12-22
Fix formatted sql precondition not expanding changelog parameters.

gudmundur-heimisson:
Formatted sql precondition does not expand changelog parameters.


nvoxland:
Thanks. Created issue CORE-2652 for change

PR #542, opened 2016-01-01
Fix the Maven artifact name for EasyMock.

mattbertolini:
The proper Maven artifact name for EasyMock is `easymock` not the camel-case `easyMock`. The camel-case spelling causes issues on case-sensitive filesystems (like those in use on Linux systems).


nvoxland:
Thanks!

PR #543, opened 2016-01-07
(Derby) If error when determine Driver Version set a bigger

lazaronixon:
When I use liquibase with maven plugin it can't get the Driver Version, so it is set to -1 and I can't use sequences or other resource available on bigger versions of database. I think would better set it to bigger version and not restrict the features. If you could find why it doesn't work on maven plugin would be a solution too.


nvoxland:
I committed my own version that sets it to 10.6 instead of 99. That seemed to be a version old enough to support everything while still being released long enough ago that most installs should be at least that.

nvoxland:
Fixed under CORE-2651

PR #547, opened 2016-02-04
CORE-2594 replace FileUtil.unzip with in-memory scanners

RichardBradley:
1. change `ClassLoaderResourceAccessor` to use `JarFile` instead of `FileUtil.unzip` when scanning JARs
2. replace `FileUtil.unzip` with `loader.getResourceAsStream` when scanning nested JARs

This fixes:
 * [CORE-2595](https://liquibase.jira.com/browse/CORE-2595) DefaultPackageScanClassResolver fails if "fat" jar has dirs and files with same name
 * [CORE-2483](https://liquibase.jira.com/browse/CORE-2483) Liquibase does not delete unzip directories from temporary directory
 * [CORE-2115](https://liquibase.jira.com/browse/CORE-2115) Really slow when using fat jars
 * Some problems I've been having where `/tmp` on my CI server runs out of inodes

By eliminating `FileUtil.unzip` which scanned JARs on the classpath by extracting them to `/tmp/`

As discussed on [CORE-2594](https://liquibase.jira.com/browse/CORE-2594), the best long-term fix for all these problems is to stop fighting against the JVM with `PackageScanClassResolver` and switch to using ` java.util.ServiceLoader`.

I tried quite hard to make these changes totally backwards compatible, and I think they most likely are, but I haven't been able to test that exhaustively, as I don't know when nested JARs are needed by Liquibase, and there's no test coverage of the feature.


RichardBradley:
This change reduces my liquibase run time by about 30 seconds on Windows (my fat JAR has 41,000 entries for a smallish enterprise web app, including one nested JAR imported from {{net/sf/ehcache}}, which was triggering a full JAR unzip to `tmp` by Liquibase on every run).

PR #550, opened 2016-02-11
Reduced fetch size. 

mbocek:
Original fetch size causing SQLWarning exception with Oracle TimesTen jdbc driver.

There is no comment in oracle docu regarding setFetchSize but i have found note in squirrel forum: http://sourceforge.net/p/squirrel-sql/mailman/message/9012007/

Exception from timesten jdbc driver:
org.liquibase:liquibase-maven-plugin:3.4.2:update failed: liquibase.exception.DatabaseException: com.timesten.jdbc.JdbcOdbcSQLWarning: [TimesTen][TimesTen 11.2.2.7.0 CLIENT]Option value changed

### DIFF
--- a/liquibase-core/src/main/java/liquibase/database/AbstractJdbcDatabase.java
+++ b/liquibase-core/src/main/java/liquibase/database/AbstractJdbcDatabase.java
@@ -55,6 +55,7 @@ import java.util.regex.Pattern;
 public abstract class AbstractJdbcDatabase implements Database {
 
     private static final Pattern startsWithNumberPattern = Pattern.compile("^[0-9].*");
+    private final static int FETCH_SIZE = 1000;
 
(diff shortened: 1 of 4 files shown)


nvoxland:
I'd like this to only reduce the fetchsize for TimesTen. Are you able to better target the setting change?

mbocek:
We can pull this "fetchSize" (get/set) to Database interface. Each database implementation should be able to override this default value. Are you OK with modification one of core interface?

mbocek:
Please check modification. Fetch size can be now database specific.

mbocek:
@nvoxland Please update pull request. I'm close to have first implementation for timesten DB.

nvoxland:
Looks good now, thanks

PR #551, opened 2016-02-11
Update MANIFEST.MF

seamusmac:
In Websphere Liberty, on startup I get the following.
[WARNING ] SRVE9967W: The manifest class path lib/snakeyaml-1.13.jar can not be found in jar file file:/C:/Users/Mac/.m2/repository/org/liquibase/liquibase-core/3.4.2/liquibase-core-3.4.2.jar or its parent.


nvoxland:
The problem is that if I remove it, then the command line "java -jar" calls will no longer work. Since it's just a warning on your side I'm going to not bring this change in.

PR #555, opened 2016-02-28
CORE-2680 Fixed the classpath prefix not being removed when comparing…

SquirrelGrip:
… RanChangeSets and ChangeSet


nvoxland:
Thanks!

PR #557, opened 2016-03-08
Fix build, remove defaults

umcodemonkey:
This main point of this is to fix the build of liquibase-debian, which does not build in a clean environment where 3.4.2-SNAPSHOT isn't present. While fixing that, I reviewed the other poms for cleanups and removed some defaults that should be left to Maven to get right.


adriens:
Did you see same kind of thing on liquibase-rpm ?

umcodemonkey:
The pom.xml look fine. I'm on a Mac, and so I can't test the RPM build.

adriens:
great, thanks @umcodemonkey for feedback.

nvoxland:
Thanks!

PR #559, opened 2016-03-08
maven poms clean up

mbocek:
I have cleaned up maven poms. More focused on using jars dependency management and also plugins. No new functionality only maintenance pull request. 

Fixed warning with duplicate javacc-maven-plugin plugin declaration.


nvoxland:
Thanks

PR #561, opened 2016-03-13
Wrote a manpage

bceverly:
Hi,

I added a manpage for liquibase.  I'm in the process of submitting liquibase to the OpenBSD ports/packages system to be included in the list of available binary packages.

Thanks!

### DIFF
--- a/liquibase-core/src/main/resources/liquibase.1
+++ b/liquibase-core/src/main/resources/liquibase.1
@@ -1,0 +1,267 @@
+.Dd $Mdocdate$ 
+.Dt LIQUIBASE 1
+.Os 
+.Sh NAME 
+.Nm liquibase 
+.Nd A tool to manage sql database schemas via config files 
+.\" .Sh LIBRARY 
+.\" For sections 2, 3, and 9 only. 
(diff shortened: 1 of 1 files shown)


adriens:
That's great @bceverly . Do yo intend to port it in the deb and rpm packages or shall we have to do it by ourselves ?

bceverly:
Sorry, I'm not understanding the question.  Do I need to overwrite the liquibase.1 files in the rpm and core subdirectories and include that in the commit?

ro-rah:
Sizing Med to allow time for Eng, Docs and QA to review.

nvoxland:
This PR is just bringing in an older version of the help docs into a file which isn't being included in the artifacts.

I'm also not sure how different the openbsd man format is from what we will be shipping with rpm/deb. We really want to just be managing that man page content one time, and have it used across all the packaging formats.

Since this particular PR is out of date and would not be used anywhere, rather than bringing it in directly we should use it as a basis for the man page portion of the deb/rpm/openbsd package creation logic.

kataggart:
closing this due to aging; we can revisit the idea of a MAN page in the future as needed. Thanks.

PR #562, opened 2016-03-24
Added validateOnly flag to make liquibase just check the database is up-to-date

nigelsim:
The change means that when you use the spring integration you can have Liquibase check the DB is in a valid state and throw an exception if it isn't, instead of trying to modify the state.

### DIFF
--- a/liquibase-core/src/main/java/liquibase/integration/spring/SpringLiquibase.java
+++ b/liquibase-core/src/main/java/liquibase/integration/spring/SpringLiquibase.java
@@ -257,6 +257,9 @@ public class SpringLiquibase implements InitializingBean, BeanNameAware, Resourc
      */
     private boolean ignoreClasspathPrefix = true;
 
+    private boolean validateOnly = false;
+
+
 	public SpringLiquibase() {
 		super();
 	}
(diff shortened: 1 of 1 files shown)


nvoxland:
I can see some uses for this, but I'm not sure this has a wide-enough use to include in core Liquibase. What use case are you using it for?

You can always subclass SpringLiquibase yourself and use that bean with your logic in it, even if it's not in the main liquibase code.

nigelsim:
We require that an DDL changes to production machines be done separately and deliberately, so when we are using Hibernate we set <property name="generateDdl" value="false" /> so that the application won't start unless the database is in the correct state, but it won't actually make any changes. I am trying to achieve the same thing with Liquibase, and for the time being I've done exactly what you suggested. 

I've actually flipped the logic in our latest build to match the generateDdl option in Spring/JPA, so I can just use the same configuration parameter. I'd suggest not accepting this pull request on that basis, but I'm curious to understand your reservations on this in general?

nvoxland:
That makes sense as a use case, and probably worth adding. 

My concern was around keeping an application from starting vs. just updating the database but I see the value of using it to ensure an updateSql script has been ran separately against a production database.

I created https://liquibase.jira.com/browse/CORE-2726 to look at adding a similar feature in the 4.0 codebase.

tompson:
stumbling over this 8 years later having the exact same use case

we want to use liquibase to check in PROD if all the changes were applied already before starting the Spring Boot application

currently we are using Hibernate DDL validate but that does not check enough

PR #563, opened 2016-03-25
CORE-2704 DB2 - multiple reorgs created

smrutimo:
Hi,

I made a fix to move the addition of reorg table statement outside of the add column loop so it creates a single reorg at the end.

Thanks & Regards,
Smruti..


nvoxland:
Thanks!

PR #566, opened 2016-04-19
* fixed the mapping of TINYINT type for DB2 database

rglow:
TINYINT should be mapped to SMALLINT in DB2.
See Table 1. http://www.ibm.com/developerworks/data/library/techarticle/dm-0807patel/


nvoxland:
Thanks!

PR #569, opened 2016-04-26
Escaped schema names for case when schema name can contain hyphen ( f…

snorochevskiy:
Hello.
On our legacy project we have Oracle schemas with names that contain hyphen. For these schemas liquibase fails with following message:

Unexpected error running Liquibase: ORA-00922: missing or invalid option
 [Failed SQL: ALTER SESSION SET CURRENT_SCHEMA=BUR-QA1-AGT]

That's because hyphens are not allowed for schema names. Such names should be surrounded with quotes.

With my changes "liquibase update" works fine. I hope I did everything right.


nvoxland:
Looks good, thanks!

PR #571, opened 2016-04-29
CORE-2735 test rollback on update added to SpringLiquibase

chexauen:
SpringLiquibase has been enhanced with an additional configuration field that allows users to test rollback during updating


nvoxland:
Thanks!

PR #573, opened 2016-05-03
CORE-2734 Fix regression in newline handling in sqlFile endDelimiter

dpovey:
This fix substantially rewrites the endDelimiter handling to support matching the tokenized output from SqlParser.  endDelimiters are treated differently depending on whether they are a regex or a standard expression.  Also improved handling of escape sequences other than newlines.


nvoxland:
Went with a simpler version that seems to work as well. Thanks, though.

PR #575, opened 2016-05-09
SQLServer: Fix to getConnectionSchema with jTDS driver

evigeant:
Using SQLServer with the jTDS driver, I get errors on `AbstractJdbcDatabase.getConnectionSchemaName()`.

Specifically, the error is related to the fact that `select schema_name()` is run in a `RawCallStatement` instead of `RawSqlStatement`. Looking at the commit which refactored this code, it was executed in a RawSqlStatement before.

Also, `PostgresDatabase` seems to have been overwritten in ab4a642eb8c866fb9d23a7aefe76f03313814abd

This restores the refactored structure along with the changes that were lost in
5e77041293ccd6b07eb96436366f29031a3d7367
ccae26f42e79d69c593fdf7315081837fbb8f2f6

It is possible that Postgres should use a `RawSqlStatement` instead of `RawCallStatement` to execute the `select schema_name()` query, but I don't have a Postgres database to test and I left the statements as they were before the refactor.


liquibase-core/src/main/java/liquibase/database/core/PostgresDatabase.java
```diff
@@ -236,18 +235,8 @@ public boolean isReservedWord(String tableName) {
...
-        try {
-            String currentSchema = ExecutorService.getInstance().getExecutor(this)
-                    .queryForObject(new RawCallStatement("select current_schema"), String.class);
-            return currentSchema;
-
-        } catch (Exception e) {
-            throw new RuntimeException("Failed to get current schema", e);
-        }
+    protected SqlStatement getConnectionSchemaNameCallStatement() {
+        return new RawCallStatement("select current_schema()");
```
evigeant: Should it be `RawSqlStatement`?

nvoxland:
Merged to 3.5.x branch for 3.5.2 release

PR #581, opened 2016-05-26
Extract MSSQL sql query from SnapshotGenerator

MarcelScheeler:
In my oppinion the database specific query must not in ithe snapshot generator.

Also change the query for mssql2000


nvoxland:
For 3.x I'm going to keep the code structured as it is but I'm working on a larger cleanup for 4.0 and will look at this for that version.

MarcelScheeler:
It is 3 years ago.

PR #582, opened 2016-05-27
Better support for DB2 z/OS and fix of DB2 data server type detection

MarkusBernhardt:


MarkusBernhardt:
Have to clean up first

PR #583, opened 2016-06-01
Update for DB2 iSeries support

bhat-man:
Update to support DB2 for the iSeries.  With this version many of the commands which were failing before are able to complete succesfully.  A new referential constraint query for DB2i was created and executed in the appropriate places.  The DB2i query was built based on the documentation for version 7.3 of DB2 for i.

https://www.ibm.com/support/knowledgecenter/ssw_ibm_i_73/db2/rbafzcatalog.htm

As indicated in the link, it is preferred to use SYSIBM for querying instead of QSYS2 for performance reasons.  This was tested and verified on a DB2 for i database running version 7.2.


nvoxland:
I merged the fix in and refactored out the AS400 check like we do with the isZOS() check. Let me know if you run into any problems.

bhat-man:
Thanks much!

ingoschaefer:
@nvoxland 
Seems this fix was merged out somewhere in the 3.6 series, it is not there anymore so generateChangeLog is failing again using the released version von liquibase and the db2i-addon.

PR #585, opened 2016-06-10
CORE-2788 support VARBINARY in Oracle and h2

ChristophGr:


ChristophGr:
any news here?
is something missing?

nvoxland:
Thanks for the fix!

PR #591, opened 2016-06-19
Add reverse engineering support for DB2 z/OS

MarkusBernhardt:
In DB2 z/OS the database definition is stored in different tables than in a standard DB2 LUW. With this patch LiquiBase will use the right SQLs to determine the database layout on a DB2 z/OS.


nvoxland:
Thanks!

PR #593, opened 2016-06-27
Close input stream after parsing JSON snapshot

bartblacha:
CORE-2806 If the input stream parsing the snapshot file does not get
closed, then the process cannot delete the snapshot file after it is done
using it.


bartblacha:
Nathan @nvoxland , please review.

nvoxland:
Your pull request is targetting master. Do you want it to actually target ddb?

nvoxland:
Merged ddb pull request instead

PR #594, opened 2016-06-27
Close input stream after parsing JSON snapshot

bartblacha:
CORE-2806 If the input stream parsing the snapshot file does not get
closed, then the process cannot delete the snapshot file after it is done
using it


bartblacha:
@nvoxland Thanks, yes, I want it to target ddb.  Github allowed me to change the target here.

nvoxland:
thanks

PR #600, opened 2016-08-07
CORE-1863 Fix Postgres BLOB type mappings

henryptung:
Postgres does not actually support the BINARY type.
Instead, use OID for streamable javax.sql.Blobs, and
BYTEA for everything else.


nvoxland:
Thanks. I'm going to cherry pick it to the 3.5.x branch to go in 3.5.2

candrews:
@nvoxland this issue caused a pretty significant regression - https://github.com/liquibase/liquibase/issues/605 - can you please merge the fix for it?

PR #602, opened 2016-08-25
Add support for Informix using IBM DB2 Universal JDBC (jcc) driver

BastiaanProot:
When using the IBM DB2 Universal JDBC (jcc) driver to connect to an Informix database the "product name" will be something like "IDS/UNIX64" or "IDS/UNIX32". Liquibase fails to recognise these databases as Informix databases. With this fix these "IDS" product names will be recognised as Informix databases.

For further  explanation:
http://forum.liquibase.org/#Topic/49382000001534007


nvoxland:
Thanks!

PR #604, opened 2016-09-01
Pretty durations for changesets that take longer than 999 ms

firehooper:
Some of our bulk inserts take longer than 999 ms.  Didn't want to add another dependency (joda-time) so used: http://stackoverflow.com/questions/625433/how-to-convert-milliseconds-to-x-mins-x-seconds-in-java


nvoxland:
Incorporated change with other ongoing work. Thanks!

PR #608, opened 2016-10-03
fix() This fixes the includeAll directive when the liquibase changelo…

maxdidato:
When using the directive includeAll in changelog.xml, if the classpath is within a JAR file, the class ClassLoaderResourceAccessor will try to find the path passed as input, in the format of jar:file:/file_path!/folder/with/changelogs in the list of paths inside the jar, not finding it eventually. 
It needs to compare only the classpath with the list of paths inside the jar. The input needs to be stripped to folder/with/changelogs and try to find it inside the jar

### DIFF
--- a/liquibase-core/src/main/java/liquibase/resource/ClassLoaderResourceAccessor.java
+++ b/liquibase-core/src/main/java/liquibase/resource/ClassLoaderResourceAccessor.java
@@ -72,11 +72,12 @@ public class ClassLoaderResourceAccessor extends AbstractResourceAccessor {
             URL fileUrl = fileUrls.nextElement();
 
             if (fileUrl.toExternalForm().startsWith("jar:file:")
-                    || fileUrl.toExternalForm().startsWith("wsjar:file:")
-                    || fileUrl.toExternalForm().startsWith("zip:")) {
+                || fileUrl.toExternalForm().startsWith("wsjar:file:")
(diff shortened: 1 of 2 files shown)


liquibase-core/src/main/java/liquibase/resource/ClassLoaderResourceAccessor.java
```diff
@@ -72,24 +72,29 @@ public ClassLoaderResourceAccessor(ClassLoader classLoader) {
...
 
             if (fileUrl.toExternalForm().startsWith("jar:file:")
-                    || fileUrl.toExternalForm().startsWith("wsjar:file:")
-                    || fileUrl.toExternalForm().startsWith("zip:")) {
+                || fileUrl.toExternalForm().startsWith("wsjar:file:")
+                || fileUrl.toExternalForm().startsWith("zip:")) {
 
                 String[] zipAndFile = fileUrl.getFile().split("!");
                 String zipFilePath = zipAndFile[0];
+                path = zipAndFile[1];
```
bilak: this won't work for implementations like spring-boot. There you have one fat-jar which contains other jars inside -> spring-boot-fat.jar!/BOOT-INF/libs/implementation-lib.jar!/db/change-log/change-log.xml

nvoxland:
Fixed by similar pull requests

PR #611, opened 2016-10-08
Changes the stream output of liquibase when success message

Alexandre-Machado:
On Octopus Deploy, the stream is used for the status of the deploy:

System.err.println
image

System.out.println
image


hkdennis2k:
I think System.err fit for most cases.

If the patch was just for Octopus Deploy, may be a configuration or integration module does better than change everyone else?

kaltinril:
I agree with @Alexandre-Machado , the only thing that should go to stderr are errors.

We have a bamboo process that I was investigating when I stumbled across main.java and this issue.  Bamboo reports every line from the liquibase deployment as an "error" because the output is on the stderr stream.

Why would you want success messages going to the error stream?

Alexandre-Machado:
### Related issues
- [CORE-2913](https://liquibase.jira.com/browse/CORE-2913): Shell script prints to stderr on success

gytisgreitai:
this just hit us too. Stackdriver shows these as errors, so we cannot create an alert if liquibase actually failed.

Since this has been sitting here since October, as I understand there is no intention to merge this?

Alexandre-Machado:
closing commit because now it does not make sense anymore

PR #614, opened 2016-10-21
BUG FIX: PostgreSQL does not support DISABLE

isaki-x:
I attempted to create Jira account, but Atlassian stated that Jira is out of licenses (which is odd, because I have an Atlassian Cloud account already). Anyway, there is a bug in liquibase 3.5.x that causes certain automated table migrations to fail.

Basically, commit ddd903a added support for PostgreSQL DEFER. It did this by piggybacking existing Oracle code (as Oracle supports this as well). However, the way this change was made introduced "DISABLE" to PostgreSQL as well (which is not supported).

I have corrected this mistake by splitting disable into its own 'if' test.

This should also be merged into master as master has the same bug.

Here is a relevant Exception text:

Reason: liquibase.exception.DatabaseException: ERROR: syntax error at or near "DISABLE"
  Position: 94 [Failed SQL: ALTER TABLE public.some_table ADD CONSTRAINT some_table_id UNIQUE (id) DEFERRABLE INITIALLY DEFERRED DISABLE]

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,0 +1,10 @@
+language: java
+sudo: required
+jdk:
+  - openjdk6
+
+before_install:
+  - sudo apt-get -qq update
+  - sudo apt-get install -y rpm
+
+script: mvn clean package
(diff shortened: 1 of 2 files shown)


isaki-x:
This pull request has been updated as follows:

1) Minor performance/code style fix to store repeated instanceof checks as booleans.
2) Fixed compiler warnings for the class I touched as part of this pull.
3) Branches have not been updated to support your Travis-CI stuff. As such, I have included the .travis.yml from HEAD of master in this pull so that it passed your CI requirement.

nvoxland:
There were enough other changes to the existing code that I ended up just applying your DISABLE change directly.

isaki-x:
Awesome thanks!

PR #615, opened 2016-10-29
Set the unique constraint name in AddColumnChange

amiel-marqeta:
Seems like a small but annoying bug that we can't set the unique constraint name when adding a new column.


nvoxland:
Applied change

PR #618, opened 2016-11-02
fix the escaping of object name on the MS SQL

pvojtechovsky:
There is a bug that AbstractJdbcDatabase.quotingStrategy is ignored when 
```java
class MSSQLDatabase  {
public String escapeObjectName(String objectName, Class<? extends DatabaseObject> objectType)
```
calls directly 
```
quoteObject(objectName, objectType)
```
I think it should call 
```java
super.escapeObjectName(objectName, objectType)
```
then quotingStrategy can be used to influence quoting of objects in MS SQL scripts. It works for me on branch 3.5.x. I did not tested on master, but it seems to be same there too


hkdennis2k:
MSSQLDatabase also ignored `setOutputDefaultSchema(false)`

PR #624, opened 2016-11-10
Don't hardcode version dependencies using 'Class-Path:' in MANIFEST.MF - 4.0.x

steffenyount:


nvoxland:
4.0.0 branch will be going away

PR #626, opened 2016-11-15
Postgres does not support parameters for type int

felixbarny:
See https://www.postgresql.org/docs/current/static/datatype-numeric.html


felixbarny:
Anything wrong with this pull request?

felixbarny:
@nvoxland would you mind having a look at this pull request? I've also encountered a similar set of issues for HSQLDB. Is it ok for you if issue more pull request for that?

PR #628, opened 2016-11-28
StringUtils.isDelimiter() bug fix

MirekPluta:
Delimiter was not detected if there was a white space just before that.
In example following string "statement 1 \nGO\nstatement 2" failed when searching for \nGO delimiter within.

I've made a fix in isDelimiter() method to handle this case and also update unit test to cover it.


nvoxland:
Thanks!

PR #629, opened 2016-11-29
CORE-2944 Honor `outputDefaultSchema=false` and `outputDefaultCatalog=false` when generating SQL code.

pbzdyl:
I believe that the condition checked when escaping object name in `AbstractJdbcDatabase` was incorrect. The condition to omit the schema part of the DB name should be:
- if the default schema has been set
- and the current DB object name's schema is the same as the default schema
- and `outputDefaultSchema` has been set to false

The above can be translated to:

`defaultSchemaSet && isDefaultSchema(catalogName, schemaName) && !getOutputDefaultSchema()`

The same applies to default catalog name condition.


pbzdyl:
Is there any chance to get this PR reviewed and merged?

nvoxland:
Looking at it soon, sorry for the delay

nvoxland:
Similar fix has been applied for 3.6.0

PR #630, opened 2016-12-05
Trigger an exception on a RAISERROR

jdenly:
Iterate through result sets and update counts returned from statement execute to trigger exception to be thrown if there is a RAISERROR.


nvoxland:
Same change applied in #632

PR #639, opened 2017-01-09
Fix for incorrect serialization of Timestamps

sebbes:
When generating changeLogs I stumbled over a serialization bug. The fractional seconds (millieseconds) in java.sql.Timestamp are not serialized/deserialized correctly; This is due to not using the default DateTimeFormats. This should fix the problem.

Here is a test that outlines the problem with current implementation:

`
import org.junit.Assert;
import org.junit.Test;

import java.text.ParseException;
import java.text.SimpleDateFormat;
import java.util.Date;

public class SimpleTimestampTest {

    private static final String DATE_TIME_FORMAT_STRING_WITH_SPACE_AND_DECIMAL = "yyyy-MM-dd HH:mm:ss.SS";
    private SimpleDateFormat dateTimeFormat = new SimpleDateFormat(DATE_TIME_FORMAT_STRING_WITH_SPACE_AND_DECIMAL);

    @Test
    public void testFormat() throws ParseException {
        String dateAsString = "2017-01-06 11:10:48.75";

        Date date = dateTimeFormat.parse(dateAsString);

        System.out.println(new java.sql.Timestamp(dateTimeFormat.parse(dateAsString).getTime()));
        System.out.println(java.sql.Timestamp.valueOf(dateAsString));

        Assert.assertEquals(java.sql.Timestamp.valueOf(dateAsString), new java.sql.Timestamp(dateTimeFormat.parse(dateAsString).getTime()));
    }
}
`

### DIFF
--- a/liquibase-core/src/main/java/liquibase/change/ColumnConfig.java
+++ b/liquibase-core/src/main/java/liquibase/change/ColumnConfig.java
@@ -4,6 +4,7 @@ import liquibase.parser.core.ParsedNode;
 import liquibase.parser.core.ParsedNodeException;
 import liquibase.resource.ResourceAccessor;
 import liquibase.serializer.AbstractLiquibaseSerializable;
+import liquibase.serializer.ReflectionSerializer;
 import liquibase.statement.DatabaseFunction;
 import liquibase.statement.SequenceCurrentValueFunction;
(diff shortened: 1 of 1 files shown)


sebbes:
Closed, as I see in the current master snapshot the timeformat handling has already changed

PR #640, opened 2017-01-10
fix for Timestamp/Date handling

sebbes:
I want to share my fix for the timestamp handling whenever the change log created via liquibase contains values with fractional seconds less than 3 digits; in this case the current implementation in the corresponding "insert" always add leading zeros (e.g. 2017-01-10 10:34:12.34 -> 2017-01-10 10:34:12.[**0**]34.

My solution is to override the getSerializableFieldValue in the ColumnConfig class, using liquibase's ISODateFormat. ISODateFormat now uses java.util.Date as a fallback in the format method.

As a side effect that changes the output format in the change logs which works well for me, but I am not sure about backwards compatibility. I also had to fix the timezone for the JsonChangeLogSerializerTest.e 

I would kindly ask you to review this PR and let me know whether that this can be merged.

### DIFF
--- a/liquibase-core/src/main/java/liquibase/change/ColumnConfig.java
+++ b/liquibase-core/src/main/java/liquibase/change/ColumnConfig.java
@@ -917,7 +917,7 @@ public class ColumnConfig extends AbstractLiquibaseSerializable {
     @Override
     public Object getSerializableFieldValue(String field) {
         Object o = ReflectionSerializer.getInstance().getValue(this, field);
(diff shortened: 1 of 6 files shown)


stupiddog1979:
Hi Sebbes, thanks for your fix. I think I had the same problem and I also made a fix for it in https://github.com/liquibase/liquibase/pull/637 - maybe we can merge our work?

nvoxland:
I think the change should be good. Merging it in, thanks!

PR #642, opened 2017-01-16
CORE-2977 Use the same name as Postgres for primary key constraints

esneider:
If no primary key constraint name is specified when creating a table, Postgres will use the name `{table}_pkey` (see http://stackoverflow.com/a/4108266/530680). However, liquibase uses `pk_{table}`. It'd be nice if liquibase naming was the same as postgres' when using a postgres database.

Postgres naming example:
```
testuser=> create table tests (id int primary key);
CREATE TABLE
testuser=> \d tests
     Table "public.tests"
 Column |  Type   | Modifiers
--------+---------+-----------
 id     | integer | not null
Indexes:
    "tests_pkey" PRIMARY KEY, btree (id)
```


nvoxland:
Code already merged

PR #643, opened 2017-01-17
Bugfix for JIRA issue CORE-2978

drivera73:
This change fixes the bug described https://liquibase.jira.com/browse/CORE-2978, while also supporting the intended schema functionality from earlier changes.


drivera73:
Evidently, the POM file changes are beyond scope.  I'm newish to git so bear with me :)

PR #645, opened 2017-01-31
CORE-2773 Detection of DB2/AS400 improved

maberle:


maberle:
Branch contained more than one fix

PR #648, opened 2017-02-03
CORE-1852 checksum depends on line endings and startCommentSymbol

melnikov-ivan:
Line endings are not processed correct in comment strings. 
Wrote test with different line endings for windows/unix while parsing csv.


melnikov-ivan:
Actually I am not sure if we should skip comments in csv-files and empty lines to calculate checksum. I would suggest just to hack line endings.

PR #651, opened 2017-02-15
Fixed FindForeignKeyConstraintsGeneratorDerby

klaffke:
I ran a changeset with dropAllForeignKeyConstraints against a derby database and soon realized that the foreign keys weren't actually dropped. The reason is the following line in the FindForeignKeyConstraintsStatementGeneratorDerby:

sb.append("AND sc.schemaname = '").append(schema.getCatalogName()).append("' ");

If the default catalog name is used, AND sc.schemaname = '' gets appended to the SQL string, which is wrong. If no catalog name is provided, the same faulty SQL is generated:

SELECT co.constraintname AS CONSTRAINT_NAME, t.tablename AS TABLE_NAME, t2.tablename AS REFERENCED_TABLE_NAME FROM sys.sysconstraints co JOIN sys.sysschemas sc ON co.schemaid = sc.schemaid JOIN sys.systables t ON co.tableid = t.tableid JOIN sys.sysforeignkeys f ON co.constraintid = f.constraintid JOIN sys.sysconglomerates cg ON f.conglomerateid = cg.conglomerateid JOIN sys.sysconstraints co2 ON f.keyconstraintid = co2.constraintid JOIN sys.systables t2 ON co2.tableid = t2.tableid JOIN sys.syskeys k ON co2.constraintid = k.constraintid JOIN sys.sysconglomerates cg2 ON k.conglomerateid = cg2.conglomerateid WHERE co.type = 'F' AND sc.schemaname = '' AND t.tablename = 'my_table_name'

Only appending    AND sc.schemaname = ''    if a catalog name is present does the trick.


nvoxland:
Changes to code avoids need for fix.

PR #653, opened 2017-02-16
CORE-3003: Index-name without schemaName

andipohl:
SchemaName is not allowed in Index names


nvoxland:
Already merged in. Thanks!

PR #667, opened 2017-03-31
Adding logic to persist full hierarchy of context expressions into databasechangelog table

yuranos:
You can also take a look at: http://stackoverflow.com/questions/43032168/context-attribute-in-include-includeall-is-not-persisted-in-databasechangelog

@mches, @nvoxland, your review will be highly appreciated.

For several reasons, we decided to stick to a formatted SQL format and since it doesn't support 'tag' attribute, we want to at least be able to track every context applied to a changeSet(fine-grained use for different environments including Production).
I know that adding something similar for 'label' is also under consideration, but it is more critical for 'context', I believe, simply because the hierarchy of 'contexts' is more compler and more difficult to trace.

### DIFF
--- a/liquibase-core/src/main/java/liquibase/ContextExpression.java
+++ b/liquibase-core/src/main/java/liquibase/ContextExpression.java
@@ -3,8 +3,11 @@ package liquibase;
 import liquibase.exception.UnexpectedLiquibaseException;
 import liquibase.util.StringUtils;
 
-import java.text.ParseException;
-import java.util.*;
+import java.util.Collection;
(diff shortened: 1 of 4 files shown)


yuranos87:
Another thing I noticed while testing the functionality:
Brackets are not supported with commas.
@nvoxland , tell me if you are planning to add it or it's avoided for a reason. I can take a look and add it if you think it's a good idea.

mches:
@yuranos @nvoxland Neat! I look forward to trying this out.

PR #669, opened 2017-04-01
Sets class name in CustomWrapper classes must be normalized.

najor:
To set the class name in the custom wrapper, you must call 'setClass' or 'setClassName' depending on the custom wrapper class. This pull request normalized this so both custom wrapper can be called with 'setClass' or 'setClassName'.


nvoxland:
Thanks!

PR #671, opened 2017-04-11
CORE-3051 ASANY supports no CASCADE in DROP TABLE

andipohl:


nvoxland:
Change already brought into master for 3.6.0.  Thanks!

PR #679, opened 2017-06-08
Add changeLogDirectory property in the maven plugin

lhauspie:
On our project, before using the maven plugin, we ran our liquibase scripts by using the command line directly from the "src/main/changelog" folder, so, the databasechangelog.filename was of type "36.0.0/master-changelog.xml".
But after using the maven plugin, the databasechangelog.filename content became "src/main/changelog/36.0.0/master-changelog.xml" so we lost the synchronization and the already executed scripts started up again.

I didn't find any way to ask to liquibase to remove the prefix folder "src/main/changelog" and I didn't want to move our changelogs (because of implying too much changes to our packaging chain, our production deployment processes, and so on). So, I thought it would be a good idea to be able to configure the location of the main changelog to remove the prefix folder in the databasechangelog.filename field.

So now we can configure liquibase-maven-plugin like that to specify the main changelog file folder :
```
<plugin>
	<configuration>
		<changeLogDirectory>${project.basedir}/src/main/changelog</changeLogDirectory>
		<changeLogFile>update-db-changelog.xml</changeLogFile>
	</configuration>
</plugin>
```
So, in this particular case, the databasechangelog.filename field will only contain "update-db-changelog.xml" and not "${project.basedir}/src/main/changelog/update-db-changelog.xml" or "src/main/changelog/update-db-changelog.xml".

This new changeLogDirectory configuration property is not mandatory and it can be an absolute or a relative path to the main changelog file folder (based on the project.basedir in relative case).

Regards.
Logan HAUSPIE.

### DIFF
--- a/liquibase-core/src/main/java/liquibase/integration/ant/AbstractChangeLogBasedTask.java
+++ b/liquibase-core/src/main/java/liquibase/integration/ant/AbstractChangeLogBasedTask.java
@@ -10,6 +10,7 @@ import java.io.OutputStreamWriter;
 import java.io.Writer;
 
 public abstract class AbstractChangeLogBasedTask extends BaseLiquibaseTask {
+    private String changeLogDirectory;
     private String changeLogFile;
     private String contexts;
     private LabelExpression labels;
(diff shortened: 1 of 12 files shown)


lhauspie:
Don't understand why Travis CI failed one rake command.
> No Rakefile found (looking for: rakefile, Rakefile, rakefile.rb, Rakefile.rb)
> /home/travis/.rvm/gems/ruby-1.9.3-p551/bin/ruby_executable_hooks:15:in `eval`
> /home/travis/.rvm/gems/ruby-1.9.3-p551/bin/ruby_executable_hooks:15:in `<main>`
> (See full trace by running task with --trace)
> 
> 
> The command "rake" exited with 1.

nvoxland:
Merged into master for 3.6.0

PR #682, opened 2017-06-19
CORE-3040 - onlyUpdate="true" flag generates empty statements for MySQL DB

caladyon:
Modified InsertOrUpdateGeneratorMySQL.getUpdateStatement to print an
update query when onlyUpdate is true. (With tests).


nvoxland:
Merged into master

PR #688, opened 2017-07-11
Compatibility with java 9 where version can be just "9". 

cocorossello:
I just copied StringUtils.substring(str, start, end) from apache commons lang3 to prevent the IndexOutofBoundsException thrown when attempting to parse jdk 9 version

### DIFF
--- a/liquibase-core/src/main/java/liquibase/util/SystemUtils.java
+++ b/liquibase-core/src/main/java/liquibase/util/SystemUtils.java
@@ -1151,7 +1151,7 @@ public class SystemUtils {
         if (JAVA_VERSION_TRIMMED == null) {
             return 0f;
         }
-        String str = JAVA_VERSION_TRIMMED.substring(0, 3);
+        String str = StringUtils.substring(JAVA_VERSION_TRIMMED, 0, 3);
         if (JAVA_VERSION_TRIMMED.length() >= 5) {
             str = str + JAVA_VERSION_TRIMMED.substring(4, 5);
         }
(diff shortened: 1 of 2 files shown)


eodonohoe:
I tried this out locally and encountered the same problem int the getVersionAsInt method as well.  

Referenced [here](https://github.com/cocorossello/liquibase/blob/dc319377b317eb8a47bd21930ba46c0a6ee11420/liquibase-core/src/main/java/liquibase/util/SystemUtils.java#L1179-L1184)

nkonev:
https://github.com/liquibase/liquibase/pull/710 another %)

esiqveland:
Please consider fixing this :-)

SteveDonie:
I spoke with @nvoxland and we definitely want to get this pull request merged in quickly. He's been pretty swamped at work.

SteveDonie:
See also pull request #710

nvoxland:
Merged into master and 3.5.x for the next patch release

ramsrib:
Is this fix available in 3.5.3 or any other version (to public) of liquidbase?

liquibase-core/src/main/java/liquibase/util/SystemUtils.java
```diff
@@ -1151,7 +1151,7 @@ private static float getJavaVersionAsFloat() {
         if (JAVA_VERSION_TRIMMED == null) {
             return 0f;
         }
-        String str = JAVA_VERSION_TRIMMED.substring(0, 3);
+        String str = StringUtils.substring(JAVA_VERSION_TRIMMED, 0, 3);
```
karlgrund: This needs to be used on line 1183 and 1184 too. 👍

PR #689, opened 2017-07-13
CORE-1609 Command Prompt: Can't connect to database with a special character in pwd

billwashere:
Special characters passed in via command prompt are being lost.
This is due to delayedexpansion is enabled.

Solution:
Turn off delayedexpansion when it is no longer required.

How to test: connect to the database using liquibase.bat and pass in the password as an argument.
Password must contain an exclamation mark (!).


nvoxland:
Thanks!

PR #690, opened 2017-07-13
Support loadUpdateData with informix database. (Bug fix / feature)

simonlamb:
Found that loadUpdateData wasn't operating correctly against an Informix database.  

Added Informix implementation of 'InsertOrUpdateGenerator'. The implementation utilizes Informix's version of the MERGE statement feature.


nvoxland:
Thanks!

PR #694, opened 2017-07-31
CORE-3033 Postgres: missing reserved word (American English)

rdziadosz:
https://liquibase.jira.com/browse/CORE-3033

British and American English support. Probably made synonyms to avoid confusion. Both should be reserved words.
https://doxygen.postgresql.org/kwlist_8h_source.html


nvoxland:
Merged

PR #696, opened 2017-08-08
Resolve bug in includeAll with Spring Boot

claude:
Why did you make this change?

The includeAll feature doesn't work with spring-boot packaged jar files.

Something chnaged in the path format that liquibase isn't able to
handle. It attempts to load the resources in the given path, but isn't
able to find the files because the path isn't pointing to the correct
place in the jar file.

I created a simple utility class to check for the signature of
the spring boot fat jar, which is that it includes 2 exclamation points,
and modified the path to correctly load the resources in the given path.

Resolves:
https://liquibase.jira.com/browse/CORE-2863
https://liquibase.jira.com/browse/CORE-2948


okineadev:
bruh PR from Claude in 2017

PR #697, opened 2017-08-09
Hotfix 3.5.3 rollback loadUpdateData tag using empty csv

simpletasks:
It should be possible to rollback chnageSet with loadUpdateData tag when using empty CSV file. File have correct column names (first row) but no data rows.


nvoxland:
Thanks!

PR #698, opened 2017-08-10
Resolve bug in includeAll with Spring Boot

claude:
Why did you make this change?

The includeAll feature doesn't work with spring-boot packaged jar files.

Something chnaged in the path format that liquibase isn't able to
handle. It attempts to load the resources in the given path, but isn't
able to find the files because the path isn't pointing to the correct
place in the jar file.

I created a simple utility class to check for the signature of
the spring boot fat jar, which is that it includes 2 exclamation points,
and modified the path to correctly load the resources in the given path.

Resolves:
https://liquibase.jira.com/browse/CORE-2863
https://liquibase.jira.com/browse/CORE-2948

### DIFF
--- a/liquibase-core/src/main/java/liquibase/util/SpringBootFatJar.java
+++ b/liquibase-core/src/main/java/liquibase/util/SpringBootFatJar.java
@@ -1,0 +1,12 @@
+package liquibase.util;
+
+public class SpringBootFatJar {
+    public static String getPathForResource(String path) {
+        String[] components = path.split("!");
+        if (components.length == 3) {
+            return String.format("%s%s", components[1].substring(1), components[2]);
+        } else {
+            return path;
+        }
+    }
+}
--- a/liquibase-core/src/main/java/liquibase/resource/ClassLoaderResourceAccessor.java
+++ b/liquibase-core/src/main/java/liquibase/resource/ClassLoaderResourceAccessor.java
@@ -4,6 +4,7 @@ import liquibase.configuration.GlobalConfiguration;
 import liquibase.configuration.LiquibaseConfiguration;
 import liquibase.logging.LogFactory;
 import liquibase.util.StringUtils;
+import liquibase.util.SpringBootFatJar;
 
 import java.io.File;
 import java.io.IOException;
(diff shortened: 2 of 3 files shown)


nvoxland:
Thanks, cherry picked into 3.5.x branch

liquibase-core/src/main/java/liquibase/util/SpringBootFatJar.java
```diff
@@ -0,0 +1,12 @@
+package liquibase.util;
+
+public class SpringBootFatJar {
+    public static String getPathForResource(String path) {
+        String[] components = path.split("!");
+        if (components.length == 3) {
+            return String.format("%s%s", components[1].substring(1), components[2]);
```
jevgeniv: Why is the components[1] part required?
This creates following issue for me at the moment. Let's say I have my changeSet file `main/resources/db/changes/foo.sql`. The changeLog looks like this:
```
<databaseChangeLog
...
    <includeAll path="db/changes" />
...
</databaseChangeLog>
```

If I run the application as a Spring Boot packaged fat-jar, then filename of the aforemention name is `BOOT-INF/classes/db/foo.sql`, but when I run Boot application using `gradle bootRun` or using IDE run-configuration, then the path for the same changeSet file becomes `db/changes/foo.sql`.

This inconsistency makes Liquibase treat that same file as different changeSets, depending on the path they are accessed by, and re-run the SQL script.

coolgang123:
Is the issue fixed ? I can still it in liquibase 3.5.5

longwa:
Agreed, same here. Can't seem to use includeAll in 3.5.5

liquibase-core/src/main/java/liquibase/util/SpringBootFatJar.java
longwa: When the `ClassLoaderResourceAccessor` iterates over the entries in the zip file and compares their start, it needs to include both component[1] and component[2] in order to find them in this check:

`                        if (entry.getName().startsWith(path)) {`

The entries in the zip look like `/WEB-INF/classes/mypackage/foo` but with just component[2] it will only check against `/foo` and not actually find anything. At least that's what I'm seeing.

My full resolved path is:
`jar:file:/Users/username/sandbox/projectName/project/build/libs/myproject-7.5.0.war!/WEB-INF/classes!/migrations/`

longwa:
Actually, looking at this more, it seems like the call to the `SpringBootFatJar` is wrong and needs to be the full path, not the path that is passed into the `list` method.

Changing to:
`path = SpringBootFatJar.getPathForResource(fileUrl.getFile());`

and keeping the original:
`String.format("%s%s", components[1].substring(1), components[2]);`

seems to work for me.

liquibase-core/src/main/java/liquibase/util/SpringBootFatJar.java
longwa: You are right though, the name needs to be considered without the middle part from a changeSet name perspective so perhaps the right change is to filter that part out of the entry name when looking for a match in the Jar file.

PR #646, opened 2017-02-01
CORE-3006 Oracle CSV-Import: "String index out of range: -1"

maberle:
Prevention of error "String index out of range: -1" for timestamps with 0 nanoseconds, e.g. "2011-01-01T00:00:00.0"

With Liquibase 3.6.0-SNP the CSV import aborts with an error. This error was caused by commit 1f20d13ecb6c0aa66a0e66d9db0cfe38fb2d28c0

Unexpected error running Liquibase: String index out of range: -1

2017-02-01 14:25:06,465 FATAL String index out of range: -1
liquibase.exception.MigrationFailedException: Migration failed for change set latest/data.xml::fill-c_messg::latest:
     Reason: java.lang.StringIndexOutOfBoundsException: String index out of range: -1
	at liquibase.changelog.ChangeSet.execute(ChangeSet.java:626)
	at liquibase.changelog.visitor.UpdateVisitor.visit(UpdateVisitor.java:51)
	at liquibase.changelog.ChangeLogIterator.run(ChangeLogIterator.java:79)
	at liquibase.Liquibase.update(Liquibase.java:218)
	at liquibase.Liquibase.update(Liquibase.java:196)
	at liquibase.integration.commandline.Main.doMigration(Main.java:1135)
	at liquibase.integration.commandline.Main.run(Main.java:190)
	at liquibase.integration.commandline.Main.main(Main.java:105)
	at de.hpa.bpe.env.tools.BpeDatabaseUpdater.runLiquibaseCommand(BpeDatabaseUpdater.java:266)
	at de.hpa.bpe.env.tools.BpeDatabaseUpdater.run(BpeDatabaseUpdater.java:254)
	at de.hpa.bpe.env.tools.BpeDatabaseUpdater.main(BpeDatabaseUpdater.java:99)
Caused by: java.lang.StringIndexOutOfBoundsException: String index out of range: -1
	at java.lang.String.substring(String.java:1967)
	at liquibase.database.core.OracleDatabase.getDateLiteral(OracleDatabase.java:262)
	at liquibase.database.AbstractJdbcDatabase.getDateTimeLiteral(AbstractJdbcDatabase.java:443)
	at liquibase.database.AbstractJdbcDatabase.getDateLiteral(AbstractJdbcDatabase.java:463)
	at liquibase.sqlgenerator.core.InsertGenerator.generateValues(InsertGenerator.java:76)
	at liquibase.sqlgenerator.core.InsertGenerator.generateSql(InsertGenerator.java:42)
[...]

PR #700, opened 2017-08-16
CORE-3006: Improve handling of invalid date literals

marchof:
In case of invalid date literals OracleDatabase.getDateLiteral() throws
an StringIndexOutOfBoundsException. This isn't helpful as the user can't
see the problematic input. With this change a "UNSUPPORTED:xxx" message
is created.


marchof:
There is another PR #646 for the same issue. The proposed change here has the following differences:

- A "INVALID" message is created when the format is not recognized
- Unit tests included

PR #701, opened 2017-08-18
DVONE 5036 changed removeTrailingDelimiter to remove delimiter from the end even when they are followed by comments and whitespaces

neupsh:
changed CreateProcedureGenerator.removeTrailingDelimiter to remove the end delimiter string, if any, from the text along with the comments (line and block) and whitespaces following them. The text will be returned as it is if the end delimiter is not found.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,0 +1,3 @@
+language: java
+sudo: false
+script: mvn clean package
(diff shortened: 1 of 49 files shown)


neupsh:
Oops!!

PR #703, opened 2017-08-24
Adding hook to support liquibase hibernate extension

bernerbrau:
Liquibase-Hibernate fails in Maven due to a classloading issue. I've added this small interface change to the main Liquibase project to allow a fix for this issue in Liquibase-Hibernate.


nvoxland:
Thanks!

PR #707, opened 2017-09-22
Support replaceIfExists in HsqlDatabase

eperez:


nvoxland:
It doesn't look like hsql supports "create view if exists" syntax http://www.hsqldb.org/doc/guide/ch09.html#create_view-section

eperez:
I think you confused "create view if exists" with "drop view if exists"
Did you test my change?
Did you get any error?

PR #709, opened 2017-09-24
Fix for file system case sensitivity in FileSystemResourceAccessor.list

gitfool:
TL;DR: Fixes a regression in databaseChangeLog includeAll behavior on Windows after upgrading from Liquibase 3.5.0 to 3.5.3.

Note: I currently can't log in to [liquibase.jira.com](https://liquibase.jira.com) for some reason, so I'm referring to related JIRA issues here: this should fix [CORE-2851](https://liquibase.jira.com/browse/CORE-2851) (and [CORE-2898](https://liquibase.jira.com/browse/CORE-2898)).

### DIFF
--- a/liquibase-core/src/main/java/liquibase/resource/FileSystemResourceAccessor.java
+++ b/liquibase-core/src/main/java/liquibase/resource/FileSystemResourceAccessor.java
@@ -134,7 +134,13 @@ public class FileSystemResourceAccessor extends AbstractResourceAccessor {
             for (String returnPath : returnSet) {
                 returnPath = returnPath.replace("\\", "/");
                 for (String rootPath : rootPaths) {
-                    if (returnPath.startsWith(rootPath)) {
+                    boolean matches = false;
+                    if (isCaseSensitive()) {
+                        matches = returnPath.startsWith(rootPath);
+                    } else {
+                        matches = returnPath.toLowerCase().startsWith(rootPath.toLowerCase());
+                    }
+                    if (matches) {
                         returnPath = returnPath.substring(rootPath.length());
                         break;
                     }


gitfool:
This was tricky to track down so I'm not surprised it was initially noted as "unable to reproduce".

- if Liquibase is run from the command prompt in the working directory it would _succeed_
- however, if Liquibase is run via [CreateProcess](https://msdn.microsoft.com/en-us/library/windows/desktop/ms682425.aspx) and the working directory is specified
  - it would _fail_ **if** the case of the working directory name did not match what Liquibase expects

Indeed I hit this issue because my process "shells out" to Liquibase via CreateProcess. After trial and error I managed to guess it was due to file system case sensitivity and reproduced the issue with the following PowerShell script:
```
$processStartInfo = New-Object System.Diagnostics.ProcessStartInfo
$processStartInfo.UseShellExecute = $false
$processStartInfo.CreateNoWindow = $false
$processStartInfo.WorkingDirectory = 'C:\TempWorkingFolder\Test'
$processStartInfo.FileName = 'java'
$processStartInfo.Arguments = '-cp ".;c:\Liquibase\liquibase.jar;c:\Liquibase\lib\snakeyaml-1.17.jar;c:\Liquibase\lib\sqljdbc42.jar" liquibase.integration.commandline.Main --driver="com.microsoft.sqlserver.jdbc.SQLServerDriver" --changeLogFile="Test/main.xml" --url="jdbc:sqlserver://mssql:1433;databaseName=Test" --username="username" --password="password" --defaultSchemaName="tx00000001" --logLevel="debug" update'
$process = [System.Diagnostics.Process]::Start($processStartInfo)
$process.WaitForExit()
```
The above will succeed as is, but will fail when the working directory is specified with different case, e.g.:
```
$processStartInfo.WorkingDirectory = 'c:\TempWorkingFolder\Test'
```
I had a look at the source changes between 3.5.0 and 3.5.3 and it seemed to be a regression in [FileSystemResourceAccessor](https://github.com/liquibase/liquibase/blob/master/liquibase-core/src/main/java/liquibase/resource/FileSystemResourceAccessor.java#L103). I based my change on its base class [AbstractResourceAccessor](https://github.com/liquibase/liquibase/blob/master/liquibase-core/src/main/java/liquibase/resource/AbstractResourceAccessor.java#L108-L118), which takes into account file system case sensitivity when stripping root paths.

stefanoscerra:
Any news about this PR?

gitfool:
Not that I know of; this repo doesn't seem to be very active. 😞 

( ... and I still can't log in to [liquibase.jira.com](https://liquibase.jira.com) to try other channels.)

nvoxland:
Sorry, I've been far too inactive lately. Back at it, though. Thanks for tracking down the fix!

PR #711, opened 2017-10-02
Weblogic throws empty directory exception when deploying

zeatful:
Scenario:
- Multiple Module maven project deployed as a WAR with liquibase files in the module that packages the war
- Deploying the unexploded war against Weblogic 12.1.3.2 (works fine with an exploded war)
- Using Spring to execute liquibase at deployment time

While debugging I noticed that there is a generated jar thrown into `\Users\me\Oracle\Middleware\Oracle_Home\user_projects\domains\mydomain\servers\AdminServer\tmp\_WL_user\myartifactname\randompathname\_WL_gen_cs.jar` which contains the liquibase files.  However in the list method of the 'ClassLoaderResourceAccessor' it has a file type (not extension) of 'zip:', and therefore skips the logic for populating the changeset files within the directory for the jar.  Deep debugging was pointing me back to some ClassLoader code within the java source.

My fix is a hack fix, but seems harmless and alleviates the deeper issue.  Seems to be a bigger issue with how the .jar file metadata is created when it is generated.  I think it maybe easier to look for a .jar extension instead of relying on the file type itself.


zeatful:
Sorry absolutely did this wrong...

PR #713, opened 2017-10-04
fix for CORE-3117 : H2 now supports TIMESTAMP WITH TIME ZONE data type

cbxp:
This is a fix for https://liquibase.jira.com/browse/CORE-3117


nvoxland:
Thanks!

PR #716, opened 2017-10-10
[CORE-3119] setFieldValue correctly if field type = File

ghaase:
This resolves an issue where driverPropertiesFile wasn't being consumed and applied correctly.


nvoxland:
Thanks!

PR #717, opened 2017-10-12
Fix includeAll resourceComparator

solaslis:
Plz consider includeAll fix


solaslis:
Missed the correct branch, will re-do

PR #721, opened 2017-10-23
Core-3115: add trim in field splitting

flyhard:
Added trim() before splitting the type field


nvoxland:
This has been merged in separately. Thanks!

PR #726, opened 2017-11-16
CORE-3140: Invalid function original_db_name in MSSQL 2005

andipohl:
built-in function original_db_name was introduced in MSSQL 2008


nvoxland:
Thanks!

PR #728, opened 2017-12-06
Fix initializeDatabase for MSSQL

gitfool:
Note: As with PR #709 I can't create an issue in GH and I _still_ can't log in to [liquibase.jira.com](https://liquibase.jira.com).

This PR removes "alter user" used for MSSQL databases in [CommandLineUtils.initializeDatabase](https://github.com/gitfool/liquibase/blob/3.5.x/liquibase-core/src/main/java/liquibase/integration/commandline/CommandLineUtils.java#L141) since unlike all other databases this is at the database level and not the session level.

The problem is that it's causing **deadlock** issues when a database schema is being provisioned via Liquibase while data is concurrently being processed in another schema in the same database!

Given the assumption that the user exists, it's reasonable to assume the user already has the default schema; Liquibase should not intervene in this way!

### DIFF
--- a/liquibase-core/src/main/java/liquibase/integration/commandline/CommandLineUtils.java
+++ b/liquibase-core/src/main/java/liquibase/integration/commandline/CommandLineUtils.java
@@ -92,7 +92,7 @@ public class CommandLineUtils {
                     liquibaseCatalogName = liquibaseSchemaName;
                 }
             }
-            
+
             defaultCatalogName = StringUtils.trimToNull(defaultCatalogName);
             defaultSchemaName = StringUtils.trimToNull(defaultSchemaName);
 
@@ -110,7 +110,7 @@ public class CommandLineUtils {
                     database.setDatabaseChangeLogLockTableName(databaseChangeLogTableName+"LOCK");
                 }
             }
-            
+
             //Todo: move to database object methods in 4.0
             initializeDatabase(username, defaultCatalogName, defaultSchemaName, database);
 
@@ -130,9 +130,9 @@ public class CommandLineUtils {
 
     /**
      * Executes RawSqlStatements particular to each database engine to set the default schema for the given Database
-     *  
+     *
      * @param username            The username used for the connection. Used with MSSQL databases
-     * @param defaultCatalogName  Catalog name and schema name are similar concepts. Used if defaultCatalogName is null. 
+     * @param defaultCatalogName  Catalog name and schema name are similar concepts. Used if defaultCatalogName is null.
      * @param defaultSchemaName   Catalog name and schema name are similar concepts. Catalog is used with Oracle, DB2 and MySQL, and takes
      *                             precedence over the schema name.
      * @param database            Which Database object is affected by the initialization.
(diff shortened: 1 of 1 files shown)


gitfool:
This PR would also fix this Liquibase Forums [post](http://forum.liquibase.org/#Topic/49382000001733005).

stupiddog1979:
You are correct, this really looks like a bug. I looked at what the code for the other RDBMS does. I think the reason this code exists is that there is a command line option

 --defaultSchemaName=<name>                 Overwrites the default schema name
                                            (returned by the RDBMS) with a
                                            different database schema

To make this work, we need to set the default schema for the current connection/session that we have with the RDBMS server. In Oracle, we would use ALTER SESSION SET current_schema=the_given_schema_name; while in PostgreSQL, we would use SET SEARCH_PATH TO=the_given_schema_name;

You are correct in that ALTER USER ... is the wrong statement to do it in MSSQL, but I fear the feature might break if we would delete the block completely. So my question is:

Can you give us the SQL statement that has the same effect on MSSQL, but only works on the current session?  I tried searching through the documentation, but the best thing I could come up with was https://docs.microsoft.com/en-us/sql/t-sql/statements/set-statements-transact-sql , which did not help.

stupiddog1979:
I wrote a test case for that today and it is as I feared: I ran the integration test suite against a newly-created SQL Server login who had a different schema than the default one (dbo). As the integration test suite does not actually execute CommandLineUtils.initializeDatabase, it bypassed that code, and then it broke horribly. The problem is not only that some code in Liquibase core is not prepared to have a defaultSchema that is not the session's default schema (that would be fixable):

Every execution of "native" SQL (by using sql and sqlFile changes, and includeAll when it includes native SQL files) will break, and there is nothing Liquibase can do to prevent that (except parsing and re-writing every native SQL statement, which is extremely dangerous and error-prone -> no chance).

As the current method (ALTER USER...) is a no-go as well (the change is permanent beyond the current session) and there does not seem to be any possibility to change the default schema for the current session, the only feasable method I could think of would be:

- Apply gitfool's patch, but:
- Abort the LB run if (after parsing all options) defaultSchemaName is != null AND different from the  default schema of the user login (SELECT schema_name() immediately after connecting to MSSQL).

gitfool:
@stupiddog1979 Sorry for the late reply as I've been away. Thanks for merging my change!

There's one problem with your addition to check the default schema though - we're using two different users: (1) an extra privileged user used only at database creation time and whose default schema is dbo, and (2) an application level user used by an application at run time and whose default schema is different for every customer. So this addition breaks us! 😞 

FWIW, it's complicated so here are more details: we have to pass the customer's default schema to Liquibase in order for Liquibase to create the changelog in the customer specific schema, which already exists but isn't the default schema of the executing user - we specify the privileged user to Liquibase and the script creates a customer specific user _without login_ but with the customer specific default schema; we then impersonate the user (using `execute as user`) before executing the rest of the script and `revert` afterwards.

Would you please be able to add a command-line override to disable this check?

gitfool:
Oh I just noticed you merged into a fork of Liquibase. Have you tried pushing your fork back to Liquibase?

gitfool:
@nvoxland I rebased this just now, given your recent activity, and would really appreciate it if you could merge this too. It would be great to discuss it, but I don't think there's a way to do this as intended, so removing it is safest as it avoids incorrect behavior.

PS. Will fixes for 3.5.x also get cherry-picked or merged into master?

nvoxland:
Thanks for the fix. Yes, the change will make it into master as well, it actually already is.

PR #729, opened 2017-12-07
Uses SafeConstructor to enhance security

gresham-Carling:
YAML files can contain custom object definitions, containing method
definitions, which can be executed. This can provide a means of remote
code execution.

By passing a SafeConstructor into SnakeYAML, we limit the objects it
will construct and deny this potential exploit.

Unfortunately this makes yaml.loadAs(stream, Map.class) not work because
Map is not recognised as a valid tag. Instead we can 'load' and cast to
Map. If a ClassCastException is thrown it will be caught by the catch-all
block already in place.


nvoxland:
Thanks!

PR #730, opened 2017-12-23
fix includeall on WebSphere and Weblogic

Pazus:
Wheblogic path to the includeall folder starts with zip:, not file/jar. It is taken into account several lines further but not on the main check.


nvoxland:
Thanks!

PR #732, opened 2018-01-02
Fix: generateChangeLog handles AUTO_INCREMENT UNIQUE column incorrectly.

ravd:
When column is the only part of unique constraint mark it unique in column specification rather than creating addUniqueConstraint after table creation.

This allows the following from MySQL to be correctly migrated from DB to change log via generateChangelog command:

CREATE TABLE AAA (
	c1_auto_increment MEDIUMINT NOT NULL AUTO_INCREMENT UNIQUE,
	c2_pk MEDIUMINT NOT NULL,
	c3_pk MEDIUMINT NOT NULL,
	PRIMARY KEY (c2_pk, c3_pk)
);

Before the fix generateChangeLog output had 2 parts:
1. create table AAA(
c1_auto_increment[auto_increment:true],
c2_pk[pk:true],
c3_pk[pk:true],
)
2. addUniqueConstraint
Step 1 fails because you can't create auto increment column that is not part of any key.

After the fix the output has 1 change:
1. create table AAA(
c1_auto_increment[auto_increment:true, unique:true],
c2_pk[pk:true],
c3_pk[pk:true],
)

For now marked as MySQL only, but it seems the above table is can be created in H2 also.

### DIFF
--- a/liquibase-core/src/main/java/liquibase/diff/output/changelog/core/MissingTableChangeGenerator.java
+++ b/liquibase-core/src/main/java/liquibase/diff/output/changelog/core/MissingTableChangeGenerator.java
@@ -22,10 +22,13 @@ import liquibase.structure.DatabaseObject;
 import liquibase.structure.core.Column;
 import liquibase.structure.core.PrimaryKey;
 import liquibase.structure.core.Table;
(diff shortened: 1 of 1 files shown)


nvoxland:
Looks good, thanks!

PR #733, opened 2018-01-15
CORE-3155 Let CSV line content behind inline comment character contri…

helgoboss:
…bute to checksum

See https://liquibase.jira.com/browse/CORE-3155


nvoxland:
Thanks!

PR #734, opened 2018-01-23
Fix for includeAll which duplicates relative path with absolute in da…

Serenitas:
…tabase on macOS

Fix for ValidatingVisitor: there was a bug (NullPointerException) when foundErrors == null

### DIFF
--- a/liquibase-core/src/main/java/liquibase/changelog/visitor/ValidatingVisitor.java
+++ b/liquibase-core/src/main/java/liquibase/changelog/visitor/ValidatingVisitor.java
@@ -107,22 +107,23 @@ public class ValidatingVisitor implements ChangeSetVisitor {
             
             if(shouldValidate){
                 warnings.addAll(change.warn(database));
-            
-                try {                
-                    ValidationErrors foundErrors = change.validate(database);
 
-                    if ((foundErrors != null) && foundErrors.hasErrors() && (changeSet.getOnValidationFail().equals
-                        (ChangeSet.ValidationFailOption.MARK_RAN))) {
+                try {
+                    ValidationErrors foundErrors = change.validate(database);
+                    if ((foundErrors != null)) {
+                        if (foundErrors.hasErrors() && (changeSet.getOnValidationFail().equals
+                                (ChangeSet.ValidationFailOption.MARK_RAN))) {
                             LogService.getLog(getClass()).info(
                                     LogType.LOG, "Skipping change set " + changeSet + " due to validation error(s): " +
                                             StringUtils.join(foundErrors.getErrorMessages(), ", "));
                             changeSet.setValidationFailed(true);
-                    } else {
-                        if (!foundErrors.getWarningMessages().isEmpty())
-                            LogService.getLog(getClass()).warning(
-                                    LogType.LOG, "Change set " + changeSet + ": " +
-                                            StringUtils.join(foundErrors.getWarningMessages(), ", "));
-                        validationErrors.addAll(foundErrors, changeSet);
+                        } else {
+                            if (!foundErrors.getWarningMessages().isEmpty())
+                                LogService.getLog(getClass()).warning(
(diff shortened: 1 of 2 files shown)


Serenitas:
We wanted to update from 3.4.2 to 3.5.3 liquibase version. After update we noticed that some patches started to execute although they was executed a months ago. It was happened because patches with includeAll was duplicated in database with absolute file path. 
Also I found the NullPointerException caused by foundErrors == null in ValidatingVisitor and there is no check for it.

I tried snapshot with https://github.com/liquibase/liquibase/pull/709 , but it didn't work for me. 

I tried to delete `/` from `rootPaths.add(rootPath.replaceFirst("^file:", "").replace("\\", "/"));` and it works for me (I use macOS), but I'm not really sure it will be ok for everyone.

Also I added simple NPE check for ValidatingVisitor.

vbauer:
+1 It looks like, that we have the same problem with `includeAll` in the last available version (3.5.3) 

@nvoxland Excuse me,
- Is it possible to review this PR in the nearest future?
- Do you have any plans for 3.5.4 (when it will be released)?

avetokhin:
Plus from me. We faced with the same problem. Waiting for update.

homer-j:
+1 we faced the same issue

nvoxland:
Hoping to have a 3.5.4 release out in the next few days

vbauer:
Thank you for information!

Serenitas:
@nvoxland Excuse me, can you please tell something about next release date? Have your plans been changed or 3.5.4 will release soon?

gitfool:
@nvoxland I just tried 3.5.4 and hit this very issue again. Since there have been lots of changes in the "includeAll" area, perhaps the merge clobbered something? Having a look now...

gitfool:
@nvoxland Found the regression - if I revert the change from https://github.com/liquibase/liquibase/commit/3ff751ea47565995d7e7447805244c705529589d then includeAll works again.

gitfool:
I must say the file path handling in Liquibase feels way too complicated so it's hard to grok the best way to fix this. Anyway, looking around I see something similar in [ClassLoaderResourceAccessor.java#L81](https://github.com/liquibase/liquibase/blob/3.5.x/liquibase-core/src/main/java/liquibase/resource/ClassLoaderResourceAccessor.java#L81) where it seems to differentiate a file path that you might get on Windows, e.g. `file:/C:/Temp/File.sql` to strip the leading forward slash, otherwise it does not strip the forward slash.

I applied the same trick to FileSystemResourceAccessor.java in https://github.com/gitfool/liquibase/commit/1241256cc77d9b037f30cbb588e618782e3d3530 and it works for me on Windows. @Serenitas does this work for you on Mac?

(Note: I didn't escape the forward slashes in the regex as that shouldn't be needed in this context.)

Serenitas:
@gitfool Yes, it seems to work properly without duplicates.

nvoxland:
Thanks for finding the problem and the fix.

Yes, the file handling is far to complex, it mainly grew out of old Java 6 classloader-based support. I had been working on a re-implementation as part of a larger 4.0 release, but since that has been too difficult to manage along with maintaining 3.x I'm going to be backporting it for the 3.7 release.

amssaravanan:
Hello,
   This is regarding CORE-3008 - includeAll uses full file path. We are suffering from this issue ever since we tried to upgrade to 3.5.3. The issue occurs in the latest release (3.6.2) as well.

   If we have relative path in includeAll (eg: <includeAll path="../y"/>) then absolute path is entered in FILENAME column of DATABASECHANGELOG table as of 3.5.3 version and this issue is blocking us from uptaking this version.

   We are holding the upgrade for almost 2 years now. Can you please let me know if and when this issue will be resolved?

   Please find attached the reproducer in http://forum.liquibase.org/#Topic/49382000001728019

PR #735, opened 2018-01-27
Do not re-build reservedWords collection on each instantiation

vbauer:
It looks like a minor issue for `MySQLDatabase` & `MariaDBDatabase` classes


nvoxland:
Thanks

PR #736, opened 2018-01-27
3.5.x - CORE-3163 dbdocs Link text case mismatch

OsBlaineOra:
CORE-3163 dbdocs Link text case mismatch

Found a couple broken links in the dbdoc generated documentation.

Made 3 changes to fix the links.
* Changed the names of the changelog html files to be lowercase.
* Changed the href to changelog file in the authors page to be lowercase.
* The html files for columns included the schema name but the link in the table page did not.  Added the schema name to the href for the columns in the table page.


OsBlaineOra:
I fixed the conflict in liquibase-core/src/main/java/liquibase/dbdoc/TableWriter.java.

My changed were not in liquibase-core/src/main/java/liquibase/integration/commandline/CommandLineUtils.java so I left that conflict un-resolved.

nvoxland:
Thanks!

PR #738, opened 2018-02-07
CORE-3171: loadUpdateData for SASQL

andipohl:
SQL Anywhere need the following syntax:

`INSERT INTO <tablename> ON EXISTING UPDATE VALUES (...)`
[Docu](http://dcx.sap.com/index.html#sqla170/de/html/81712f3a6ce21014ada883e2b235cf26.html)


nvoxland:
Thanks!

PR #739, opened 2018-02-07
CORE-3010 MySQL Add Default Value failed

andipohl:
Following Table:
CREATE TABLE table1 (
ID int(11) NOT NULL AUTO_INCREMENT,
CREATED timestamp NULL,
PRIMARY KEY (ID));

Now I want to give CREATED a default value:

ALTER TABLE table1 ALTER CREATED SET DEFAULT NOW()
brings me:
Error Code: 1064. You have an error in your SQL syntax; check the manual that corresponds to your MySQL server version for the right syntax to use near 'NOW()' at line 1

But

ALTER TABLE table1 CHANGE COLUMN CREATED CREATED TIMESTAMP DEFAULT NOW()

works correctly

### DIFF
--- a/liquibase-core/src/main/java/liquibase/sqlgenerator/core/AddDefaultValueGeneratorMySQL.java
+++ b/liquibase-core/src/main/java/liquibase/sqlgenerator/core/AddDefaultValueGeneratorMySQL.java
@@ -3,9 +3,11 @@ package liquibase.sqlgenerator.core;
 import liquibase.database.Database;
 import liquibase.structure.core.Schema;
 import liquibase.datatype.DataTypeFactory;
+import liquibase.exception.ValidationErrors;
 import liquibase.database.core.MySQLDatabase;
 import liquibase.structure.core.Column;
 import liquibase.structure.core.Table;
+import liquibase.util.StringUtils;
 import liquibase.sql.Sql;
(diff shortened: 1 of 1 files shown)


nvoxland:
See comments in https://liquibase.jira.com/browse/CORE-3010

ro-rah:
Hi @andipohl ,

Thanks for your pull request!

Here’s what happens next:

ro-rah:
Hi again @andipohl , Please resolve branch conflicts and consider adding tests.  Thanks in advance!

Ronak

kataggart:
comments from @nvoxland in the old Jira ticket

"Nathan Voxland 
February 13, 2018 at 12:19 AM

The problem with this change is that it requires the correct datatype, nullability, etc. to be part of the change tag or they are changed as well. With the "SET DEFAULT" syntax, it keeps the rest the same.

Is there a way to support a default value function with the SET DEFAULT syntax?"

nvoxland:
The original poster found he could no longer reproduce the error.

The SQL currently being used for add default should be valid except perhaps for old verisons of mysql (5.6 and before). The suggested change in this PR would make mysql loose other metadata on the column so it's best to keep the generated SQL as is.

Closing this PR

PR #710, opened 2017-09-27
Make Java version number parsing compatible with JDK-9.

aborshik:

PR #740, opened 2018-02-12
Fixes Java9 compatibility in SystemUtils::getJavaVersionAsInt

izrobin:
Re-introduces the fix made in #710 **but overseen** in #688 .


nvoxland:
Thanks

PR #742, opened 2018-02-13
CORE-1888: Add BigInt for Sybase (Can't specify a length)

equeinne:
Fix error: liquibase.exception.DatabaseException: Can't specify a length, scale or storage property on type 'bigint' during table creation with type="BIGINT(19)"


equeinne:
Thanks

PR #743, opened 2018-02-13
Fix root path to work on Mac and Windows

gitfool:
Fixes https://github.com/liquibase/liquibase/pull/734


gitfool:
@nvoxland FYI: I downloaded 3.5.5 from [releases](https://github.com/liquibase/liquibase/releases) to confirm this fix was included and `liquibase version` returns `3.5.4`. 😕

PR #745, opened 2018-02-27
CORE-3180: Skip change sets that don't match on DBMS during change log parsing

mches:
A DBMS-specific change set referencing a DBMS-specific rollback can't be parsed on a different DBMS

e.g. MSSQL change-set referencing MSSQL rollback can't be parsed on H2.

```
<databaseChangeLog
        xsi:schemaLocation="http://www.liquibase.org/xml/ns/dbchangelog http://www.liquibase.org/xml/ns/dbchangelog/dbchangelog-3.5.xsd">

    <changeSet id="1" author="mches" dbms="mssql">
        <createTable tableName="a">
            <column name="b" type="int"/>
        </createTable>
    </changeSet>

    <changeSet id="2" author="mches" dbms="mssql">
        <dropTable tableName="a"/>
        <rollback changeSetId="1" changeSetAuthor="mches"/>
    </changeSet>

</databaseChangeLog>
```
Having Liquibase skip the non-matching change sets avoids the exception parsing the rollback node, with no side unwanted effects that I can observe.

Closes: [CORE-3180](https://liquibase.jira.com/browse/CORE-3180)

### DIFF
--- a/liquibase-core/src/main/java/liquibase/changelog/ChangeSet.java
+++ b/liquibase-core/src/main/java/liquibase/changelog/ChangeSet.java
@@ -227,13 +227,7 @@ public class ChangeSet implements Conditional, ChangeLogChild {
     }
 
     protected void setDbms(String dbmsList) {
-        if (StringUtils.trimToNull(dbmsList) != null) {
-            String[] strings = dbmsList.toLowerCase().split(",");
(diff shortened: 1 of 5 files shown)


nvoxland:
Switched merge to master for 3.6.0

PR #747, opened 2018-03-05
[CORE-3185] New Refactoring "dropAllIndexes" should be available to drop all indices of a table

Wallaby1999:
Proposal for implementation of [CORE-3185]


Wallaby1999:
Messed up with GIT.
Will create a new pull request for it. Sorry for the confusion

PR #748, opened 2018-03-10
Resource validation feature

artemptushkin:
Hello!

I'd like to request PR for the feature to add maven mojo task to run none-database connected process of validation resources for the future liquibase-update run.

1. Problem:
Resources validation such as sql files, not existed included changelog files at the maven build phase to prevent liquibase exceptions like 'sql file not found' while running post-deploy liquibas-update process.
1. Solution:
Added method at Liquibase class to run start validation.
Added method at DatabaseChangelog class to prepare classes for validation process.
Added setter for 'includeMD5Sum' at ChangelogIterator to expect loop for everyone changeset relevant to current databasechangelog.
Added ResourceValidationChangeSetVisitor to visit changeset with validate logic.
Added liquibase mojo.
UT included.

Please, let me know to redirect PR and i'll be glad to discuss code for the future merge.

##### Setup
* Create a changelog with a change set that references an external SQL file
* Do NOT create the SQL file

##### Validations
_Verify mvn compile liquibase:validateResources returns an error._
ASSERT ::

* validateResources returns an error that a file is missing
* validateResources does NOT make a connection to the live database.

_Verify mvn compile:validateResources returns success if the missing file is created._

* validateResources returns success
* validateResources does NOT make a connection to the live database.

_Verify mvn compile:update applies changes to the live database._
ASSERTS ::

* update is successful
* objects are created on the database.

## QA Functional Automated 
* None

[CORE-3341](https://liquibase.jira.com/browse/CORE-3341)

┆Issue is synchronized with this [Jira Story](https://datical.atlassian.net/browse/LB-114) by [Unito](https://www.unito.io/learn-more)
┆Fix Versions: Community 4.2.x


ro-rah:
Hi @artemptushkin 

Thanks for your pull request!

Here’s what happens next:

ro-rah:
Thanks for adding tests @artemptushkin , I am moving the merge target to 3.10.x, please let me know if this was truly meant only for 3.6.x

ro-rah:
Tagged as riskMed, due to time need to assess wide reaching change but not high since it looks like tests have been added.  I believe this is ImpactHigh.

artemptushkin:
Hi @ro-rah
I'm glad that your team are finally back to this PR. I'll be ready to change things.

ro-rah:
Hi @artemptushkin , sorry for the long delay.  Our community team is just a few months old and we are committed to connecting with our hard working Liquibase community.  Before that, the hardworking founder of our open source project @nvoxland was the only person working on....really all aspects Liquibase.  

Thanks for being a part of our community!

artemptushkin:
It is nice to hear that liquibase got new team! Nathan made a great job, I talked with him a few :)

ro-rah:
Oh cool @artemptushkin , may I schedule a short call with you?  i was hoping to get feedback from you as a developer on our PR process.  It would help us to improve here :)

nvoxland:
Still part of the team. Now just more of us :)

We do have an existing validate() function in Liquibase and that is exposed via the CLI. It's not on maven like it should be, but how is your validation different than that? I'm wondering any checking you are doing should be merged into that logic so we can have a single validate call used by both maven and cli and elsewhere.

artemptushkin:
Nice to see you @nvoxland!

I've never tried CLI validate. To compare this feature with maven `validate`: `validate` run resource validation indirectly cause it builds changeset in memory. But `validateResources` validated only and all the resource that will be needed for the further update.

I remember when was writing, that it could be a part of validate one somehow.

ValidateResources is useful in cases you don't have access from your build machine to production env but want to verify your files.

artemptushkin:
@ro-rah sorry for missing your message. Your PR process looks better for sure. Ask me after this feature release :)

ro-rah:
@artemptushkin , I will surely follow up with you after this has been merged!

nvoxland:
I can see where what you are looking to check can be different than what gets ran in the current validate code, but I think it would be best to have a single validate command that can check all the things we can, but add support to handle cases when you don't have access to the target database.

I've not looked enough at the code yet, but perhaps we can just make the connection information optional and we update the validate logic to output a message for the parts we can't validate without the connection?

Otherwise, it could be a flag to validate to specify what to check.

Would you be able to see about combining your code into validate and adding that logic?

molivasdat:
Monitoring for closure.

artemptushkin:
@nvoxland If someone describes at plugin database connection he expects that it would be used, for example for validation.

I mean in this way you've just described it is possible to get passed [green] validation but actually there hadn't been validation against database cause no database connection.
It confuses.

The only way to put:
```xml
<validateDatabase>true</validateDatabase> //default is true to keep BC
<validateResources>true</validateResources>
```

I can try to refactor this way.
and this specific underhood will be: _**If (validateDatabase) then `validateResources` is off underhood**_

What do you think?

artemptushkin:
@nvoxland kind reminder :) need your opinion

artemptushkin:
@molivasdat @ro-rah hey! I saw removing of 3.10.3 label from this PR, could you please update the status for me. Shall we resolve the question above or go as is? I prefer the second one because another way requires rework.

r2liquibase:
@molivasdat @ro-rah @artemptushkin Artem was at the Jenkins presentation yesterday. He's pretty cool. Please make this a priority.

nvoxland:
I'm taking a look at it.

r2liquibase:
Thanks, @nvoxland !

molivasdat:
Hi @artemptushkin.  Nathan is looking into it.  Thanks @nvoxland

nvoxland:
I have a local branch with the conflicts resolved. So don't have to worry about that. 
It's enough of a functionality fix/change that it should go out in one of the the 4.x releases vs. the 3.10.3 patch-only release.

@artemptushkin Looking at the code, it surprised me that the existing "validate" function requires a url and live database. I was thinking it was more focused on the local file structure (xml is valid, referenced files exist, etc.)

But, most of the reason for needing a url for validate is to be able to do database-specific validation. Like if a change attribute is valid or invalid for your particular database, it can let you know. It's not really validating the database, just validating that the resources are correct FOR that database.

For consistency with users, I would like to keep the "validate" method in maven and the CLI running the same logic. 

I tried running the CLI validate command with an invalid reference, and it did correctly fail with "file does not exist", so it looks like at least part of the logic you are looking to add should already exist in the liquibase.validate() function.

I'd be OK with making URL optional for the validate method for both maven and the CLI. If they specify a url, we use it to know the database and can perform better resource validation. If they don't specify the URL, we do what we can without a specific database.

An alternative would be to keep requiring the url, but have users leverage the "offline" urls when they don't have an actual database. They can specify `--url offline:mysql`, and it will not actually connect to a database, but still acts like a mysql database. You can even control the database version with parameters.

The offline url is a bit less obvious, but does give more control.

Thoughts on the options? Or what I'm missing?

artemptushkin:
@nvoxland thanks for open comment. You're right, I thought the same when was working on this feature. I'm going to change this PR to the way of reusing `validate` and let you know to review then. I should do it within a week

nvoxland:
Great. 

I pushed the merge work with the current master branch to https://github.com/liquibase/liquibase/tree/artemptushkin-resource-validation-4 so you can check out those changes and pull that branch into this PR.

nvoxland:
We are planning some larger refactoring work around the CLI and how to define what is required or not for the higher-level commands like "verify". So rather than making url no longer required, I'm thinking that for now it will work best to leverage the offline url structure when you don't have an actual database.

But, one thing I noticed when trying with the offline url syntax with this, is that it does require a csv file for the changelog history. In this case, I don't really need the history and it was an easy fix in the OfflineChangeLogHistory service to return an empty array from getRanChangeSets if the csv file does not exist. But, I'm not sure if that is what we want to do in general, since other uses of offline do need the changelog file and in that case it is good to get an error if the file does not exist. 

Maybe we want a special value for changeLogFile meaning "don't look for the file", like `offline:mysql?changeLogFile=none` ?

artemptushkin:
> Great.
> 
> I pushed the merge work with the current master branch to https://github.com/liquibase/liquibase/tree/artemptushkin-resource-validation-4 so you can check out those changes and pull that branch into this PR.

I've got more changes against `4.1.x` after the sync with `artemptushkin-resource-validation-4` :( looks like brought something that wasn't merged, @nvoxland if you don't what it is I can rebuild branch against `4.1.x` only with my commits or just reopen PR, because it felt like it will be done other way

artemptushkin:
> We are planning some larger refactoring work around the CLI..

I'm not so familiar with CLI and the absent changelog file looks special case but another feature. I'd focus here on validation and then resolves it with absent changelog file like: _validation will be ok if it is said to ignore changelog file_

nvoxland:
> I've got more changes against `4.1.x` after the sync with `artemptushkin-resource-validation-4` :( looks like brought something that wasn't merged, @nvoxland if you don't what it is I can rebuild branch against `4.1.x` only with my commits or just reopen PR, because it felt like it will be done other way

If it's easier to just rebuild against 4.1.x with a different PR, that is just fine too. Whatever works best for you.

artemptushkin:
@nvoxland sorry for the late reply. I checked what liquibase has and our options.

* There is no more `validate` maven goal as for `4.1.1` but it existed for `3.4.1` and it exists in CLI now
`Could not find goal 'validate' in plugin org.liquibase:liquibase-maven-plugin:4.1.1`
* `Offline` feature really covers most part of the problem I wanted to cover with this PR but has additional dependencies

@nvoxland why `validate` was removed as Maven Mojo?

Frankly, if there is no reason to get back `validate` to maven task I see no reasons to keep this idea as part of the Liquibase system, maybe as an external plugin.

Or only if you want it this way - as `validateResources`. It helps to run faster the validation of files than offline and doesn't need [changeLogFile](https://docs.liquibase.com/workflows/liquibase-community/offline.html) as the `offline` connection property.

nvoxland:
I don't remember why validate might have been removed. In the past, I tended to just let the community manage the maven plugin without inspecting the PRs closely enough. We are working on synchronizing all the interfaces to provide similar functionality, and that should include bring the validate command back.

What additional dependencies do you mean with the Offline support? The snakeyaml dependency?

artemptushkin:
Got it, thanks for the clarification.

I have to have `.csv` table inside the file system in order to work with offline database, this dependency

PR #749, opened 2018-03-10
BugFix: mvn liquibase:generateChangeLog throws IllegalAccessException in AbstractLiquibaseMojo.java

tlachy:
If you run "mvn liquibase:generateChangeLog" with properties file including "diffTypes=data" the method bellow in AbstractLiquibaseMojo.java will throw IllegalAccessException because the field is private. The same for other private fields.

        private boolean isCurrentFieldValueSpecified(Field f) throws IllegalAccessException {
                Object currentValue = f.get(this);
                ...


nvoxland:
Thanks!

PR #756, opened 2018-04-07
AddColumnGeneratorSQLite should only support SQLite database.

xiaodong-xie:
AddColumnGeneratorSQLite should only `support` SQLite database, so that this generator won't apply to other databases. 

Otherwise, I have to apply this hack `SqlGeneratorFactory.getInstance().unregister(AddColumnGeneratorSQLite.class);` to make Dropwizard build, see here: https://github.com/dropwizard/dropwizard/pull/2325/files

Thanks.


nvoxland:
Merged similar pull request

PR #759, opened 2018-04-13
CORE-3206 incompatibility with 3.5

haster:
https://liquibase.jira.com/browse/CORE-3206
Note: this is a PR for 3.6.x since that seems to be the most recent codebase. The relevant commits (all but the POM changes) should be easy to apply to master as well.

- Modified get/setTable methods in UniqueConstraint and Index to once again concern only Table-objects so the code would be backwards compatible with code written or compiled against 3.5.x versions of liquibase.
- Added a get/setRelation method that concerned Relation-objects but uses the same attributename as get/setTable because this seems to be part of the public API (at least, the get/setAttribute methods are public and I couldn't be sure who would be depending upon the exact attributename being used).
- Deprecated get/setTable, made sure it called the respective Relation method.
- Rewrote all internal code to work with get/setRelation

- Fixed the version numbers in the POMs for the submodules so maven would want to build this project.
- Fixed a bunch of unrelated tests that failed due to earlier changes in the code base, either merge errors that restored old code or changes in the code base that hadn't updated the related tests.

### DIFF
--- a/liquibase-core/src/main/java/liquibase/database/AbstractJdbcDatabase.java
+++ b/liquibase-core/src/main/java/liquibase/database/AbstractJdbcDatabase.java
@@ -815,7 +815,7 @@ public abstract class AbstractJdbcDatabase implements Database {
         } else if (object instanceof Column) {
             return isLiquibaseObject(((Column) object).getRelation());
         } else if (object instanceof Index) {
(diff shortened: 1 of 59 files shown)


nvoxland:
Merged in pull request. Thanks!

PR #763, opened 2018-04-22
[CORE-3213] includeAll will not find child changelogs in multi-modues Spring Boot's executable JAR

devnied:
Fixes https://liquibase.jira.com/browse/CORE-3213

Fixes IncludeAll in multi-modues Spring Boot's executable JAR
Remove duplicate URL to improve speed of liquibase
Avoid reuse path variable


davidyscott:
Yes please.

longwa:
+1

PR #765, opened 2018-04-22
print full path of changelog in case it's not found for better troubl…

mailtosmyk:
print full path of changelog in case it's not found for better troubleshooting


nvoxland:
Thanks!

PR #769, opened 2018-05-15
CORE-3229 Oracle 11g doesn't support TIMESTAMP WITHOUT TIME ZONE datatype

Alemorato:
Oracle 11g doesn't support WITHOUT clause in TIMESTAMP data type (https://docs.oracle.com/cd/B19306_01/server.102/b14225/ch4datetime.htm#sthref389).
For this reason replace TIMESTAMP WITHOUT TIME ZONE with TIMESTAMP.


Alemorato:
Thank, the same should be cherry-pick on master.

PR #770, opened 2018-05-16
CORE-3220 print detailed output when using logLevel INFO or greater

litpho:
In the current 3.6.x code, logLevel is not used to configure the root logger and the outputLogs field of the consoleLogFilter is never changed.

This change will keep the current behaviour if --logLevel is not used.
If --logLevel is used and it is INFO or greater, the root logger is set to this level and outputLogs will be set to true as well.

This will probably be of interest to people watching https://liquibase.jira.com/browse/CORE-3220 as well.

┆Issue is synchronized with this [Jira Story](https://datical.atlassian.net/browse/LB-117) by [Unito](https://www.unito.io/learn-more)

### DIFF
--- a/liquibase-core/src/main/java/liquibase/integration/commandline/Main.java
+++ b/liquibase-core/src/main/java/liquibase/integration/commandline/Main.java
@@ -187,6 +187,7 @@ public class Main {
             }
 
             main.applyDefaults();
+            main.changeLogLevel();
             main.configureClassLoader();
             main.doMigration();
 
@@ -246,6 +247,25 @@ public class Main {
             CommandLineOutputAppender appender = new CommandLineOutputAppender(LoggerFactory.getILoggerFactory(), target);
             root.addAppender(appender);
             appender.start();
+        }
+    }
+
+    /**
+     * Change the logging based on the parsed value of logLevel
+     */
+    protected void changeLogLevel() {
+        if ("off".equals(logLevel)) {
+            return;
+        }
+
+        Level newLevel = Level.toLevel(logLevel);
+
+        ch.qos.logback.classic.Logger root =
+            (ch.qos.logback.classic.Logger) LoggerFactory.getLogger(ch.qos.logback.classic.Logger.ROOT_LOGGER_NAME);
+        root.setLevel(newLevel);
+
+        if (Level.INFO.isGreaterOrEqual(newLevel)) {
+            consoleLogFilter.outputLogs = true;
         }
     }
 


stevesaliman:
Out of curiosity, why limit the log level change to INFO or greater?  Why not set the level to whatever the user specified?

litpho:
The root logger is, in fact, set to the level of whatever the user specified.

What may not have been clear is that the "INFO or greater" threshold is used to activate the separate "outputLogs" switch. That could have been an "Anything but OFF" switch as well, which may have been less confusing.

stevesaliman:
It looks to me like the `Level.INFO.isGreaterOrEqual(newLevel)` code will prevent debug level output from going to the console.  I actually like the "anything but OFF" idea. I think all we'd need to do is remove the INFO comparison.

prerakalw:
You can use https://github.com/prerakalw/liquibase-with-logs version as a temporary fix.

jimzucker:
> You can use https://github.com/prerakalw/liquibase-with-logs version as a temporary fix.

@prerakalw Is there a GitHub branch with this changes that we can review and build ourselves?

hajdukd:
Why it takes so long to merge such a crucial fix ?

huytmb:
This issue has not been resolved :(

prerakalw:
@huytmb try the attached jar.
@jimzucker did you try the attached jar? is it solving your problem..

asdfklgash:
> @huytmb try the attached jar.
> @jimzucker did you try the attached jar? is it solving your problem..

For me it didn't work.

huytmb:
> @huytmb try the attached jar.
> @jimzucker did you try the attached jar? is it solving your problem..

Thank @prerakalw, i'll try it.
At the moment, I decided to downgrade to use liquibase 3.5.5.

jhartotr:
Will liquibase 3.5.5 work with spring boot 2.1.5?

darioseidl:
For anyone using Gradle, the liquibase-gradle-plugin also has a workaround for this: `OutputEnablingLiquibaseRunner` (see their readme).

Is there anything stopping this pull request from being merged? It looks like a very simple fix and a lot of people are waiting for this. The only thing it doesn't address is that the SQL statements are also logged on info level now, which may be excessive, but that is less of a concern and could be addressed separately. Just having a way to turn logging on at all would be important.

ro-rah:
Hi @litpho , could I move this PR to be merged into our upcoming 3.10.x release?  Or does it have to be on the 3.6.x?  The 3.10.x would be the faster way to go.

Sorry for the long delay.  Up until a couple of months ago, @nvoxland was the only one working on the community side but, we have a community team tasked with getting PRs evaluated.  Thanks for the patience.

stevesaliman:
I'm not sure this PR is needed in 3.10.x.  The way logging is initialized seems to be one of the things that tends to change from version to version, and it looks like it works properly in the 3.10.x branch.

At least when I tested it with the Liquibase Gradle plugin, the `logLevel` switch had the desired effect.

ro-rah:
@stevesaliman !  Thanks for weighing in, looks like it is then a version specific fix.  @litpho do you still needed this PR merged to the 3.6.x for a fix to get made?  Just for full disclosure, we are prioritizing latest release (3.10.x at this writing) and master merges (for 4.0 Beta).

molivasdat:
Closing as not going to merge since already fixed in 3.10.x release

PR #751, opened 2018-03-15
Turkish upper/lower case conversion bug fix for database types

talepweb:
There is a bug related with Turkish dotless i and capital I with dot causing problems with Metabase open source project using liquibase. There are some simular bug reports & fixes related with this problem such as;
[Liquibase JIRA Bug report 1](https://liquibase.jira.com/browse/CORE-2772)
[Metabase Forum](https://discourse.metabase.com/t/metabase-centos-install-error-please-help-turkish-locale/2943)

I tried to fix by making case changes according to en-US locale for database types and it should be for such SQL language-related strings, I am not aware if there are more in the project. But the case related with database types seems fixed within my Metabase build.

Information about Turkish locale differences:
[Blog article](https://haacked.com/archive/2012/07/05/turkish-i-problem-and-why-you-should-care.aspx/)

PR #772, opened 2018-05-18
Fix case sensitivity issues in MSSQL (n)text (n)varchar conversion

ibauersachs:
Without these fixes column definitions of text/ntext where not all characters are lowercase end up with duplicate datatypes (e.g. `COL1 varchar (max) TEXT`) or are not recognized at all.
One part of the problem was introduced in #751, but even without the toUpperCase bug from that PR there are problems in the regexes.


nvoxland:
Thanks!

PR #773, opened 2018-05-21
Update MySQL's reserved words list for 8.0.x

senior:
The source of the reserve words list can be found here:
https://dev.mysql.com/doc/refman/8.0/en/keywords.html. This commit
adds the new MySQL 8 words to the reserved words list.

Fixes CORE-3235


nvoxland:
Also merged to 3.6.x branch. Thanks!

senior:
@nvoxland any idea when the fix will make it into a released version of Liquibase?

PR #775, opened 2018-06-05
Core 3159 log binding classcast

nniesen:
1. Changed Main to check the root logger name before attempting to cast to the Logback implementation.
2. Moved all the Logback implementation specific classes into the CommandLineOutputAppender.

I looked at making a CommandLineLog4jOutputAppender but log4j is really painful to configure programmatically. I don't think using the logging framework for program output is the best choice but at this point not easily changed. For now, this should at least allow the Main to be used by the gradle-liquibase plugin.

### DIFF
--- a/changelog.txt
+++ b/changelog.txt
@@ -1,6 +1,12 @@
 Liquibase Core Changelog
 ===========================================
 
+Changes in version 3.6.1 (2018.04.11)
+
+- [CORE-3200] - Wrong SQL generator is selected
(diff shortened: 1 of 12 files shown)


nniesen:
Do I need to redo this PR so that it merges to 3.6.x instead of master? I only changed Main and CommandLineOutputAppender.

nniesen:
Made a new PR that merges back to 3.6.x.

PR #778, opened 2018-06-07
Bugfix. Directories included in resource list even if includeDirectories is false

keddok:
returnSet of method ClassLoaderResourceAccessor.list contains not only files but also their directories despite includeDirectories is false. This can lead to minor performance loss.


nvoxland:
Thanks

PR #780, opened 2018-06-08
Load data is unable to find column configuration when using camel or snake case identifiers

manuelsanchezortiz:
If you normalize only LoadDataConlumnConfig's name; they may not match against tableColumns given a NullPointerException afterwards.


nvoxland:
Thanks!

PR #782, opened 2018-06-13
CORE-2377: PostgreSQL supports CREATE OR REPLACE VIEW

eperez:


nvoxland:
Thanks!

VitalyKolesnikov:
@nvoxland I faced the same issue: upgraded from liquibase:3.5.3 to the latest version and got lots of errors like "cannot drop columns from view" and "cannot rename columns in view". I can't change my old changelogs due to checksums change, so is there any workaround?

PR #785, opened 2018-07-04
Fixing DatabaseFunction as default value.

nexsandro:
Liquibase was ignoring DatabaseFunction value and processing only SequenceNextValueFunction when creating tables and generating column default value.


nvoxland:
Thanks!

PR #791, opened 2018-07-17
add MariaDB precision for TIME, DATETIME, and TIMESTAMP types for old MariaDB releases

sviatoslav1989:
MariaDb since version 5.3  has time fractional support (https://mariadb.com/kb/en/library/microseconds-in-mariadb/), but liquibase doesn't support it for MariaDB, because MySQLDatabase (superclass of MariaDBDatabase) has method getMaxFractionalDigitsForTimestamp(), which return 0 for all MariaDB 5.3.x-5.5.x versions.


nvoxland:
Thanks!

PR #792, opened 2018-07-18
Adds clearCheckSums property in spring integration

anuragagarwal561994:
# Motivation
If the database is restored from a dump on another system, if the spring-boot-liquibase tries to run migration it fails as the md5 check sum validation fails on startup. With liquibase-maven-plugin, clearCheckSums command line option is available but it is not possible to do the same with spring boot liquibase integration.

# Solution
Adds `clearCheckSums` property, which if set to true will call `liquibase.clearCheckSums` method before performing the update.


anuragagarwal561994:
@nvoxland what is the process of PR review and merge for collaboration?

nvoxland:
Sorry for the slow response, but have merged it in now. Thanks!

PR #795, opened 2018-07-27
Fix CORE-3231: NPE in LogService.pushContext for MDCs

tholu:
Overwrite `toString()` method to return an empty string for the empty anonymous implementation of DatabaseChangeLog.


nvoxland:
Thanks!

PR #796, opened 2018-07-27
Tag should remain after rollback [CORE-2946]

philomory:
Fixes https://liquibase.jira.com/browse/CORE-2946, making behavior of liquibase more consistent and predictable - rollback to tag behavior does not depend on whether the tag is isolated in it's own changeset or not. This also mirrors the behavior of updateToTag, so that updating to a tag, and rolling back to that tag, both have identical results.

### DIFF
--- a/liquibase-core/src/main/java/liquibase/changelog/filter/AfterTagChangeSetFilter.java
+++ b/liquibase-core/src/main/java/liquibase/changelog/filter/AfterTagChangeSetFilter.java
@@ -24,9 +24,6 @@ public class AfterTagChangeSetFilter implements ChangeSetFilter {
 
             if (!seenTag && tag.equalsIgnoreCase(ranChangeSet.getTag())) {
                 seenTag = true;
-                if ("tagDatabase".equals(StringUtils.trimToEmpty(ranChangeSet.getDescription()))) { //changeSet is just tagging the database. Also remove it.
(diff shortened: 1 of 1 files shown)


ro-rah:
@philomory 
Thanks for your pull request!

Here’s what happens next:

ro-rah:
Hi @philomory ,

Would you be up for adding some tests for this change?  I will size at a medium for now but can drop it down to a Low to get merged in faster if I see you commit tests before we start on 4.0 Beta 2.

Thanks!

Ronak

nvoxland:
This is handling the case where someone has a changelog with:

```
<changeSet id="1" author="test">
   <tagDatabase tag="my_tag">
</changeSet>
```

which is a different way of tagging the database than with `liquibase tag my_tag`

The difference in behavior happens because `liquibase tag` tags the last changeSet in the database and is done, whereas the tagDatabase inside a changeSet will tag the last changeset in the database and THEN create a new datbasechangelog entry saying changeset 1::test was executed.

The difference ends up being that what is tagged is different in the two cases and people want different things. https://liquibase.jira.com/browse/CORE-2815 was opened and fixed based on one way people see it working and this PR undoes that fix based on a different use case.

We need to put a bit more thought into exactly how the tagDatabase command works and whether (or if) it should be tagging the changeSet it is in or the changeSet after or if it should be a configurable thing in the tagDatabase tag or something.

filipelautert:
Fixed by PR https://github.com/liquibase/liquibase/pull/6835 .

PR #797, opened 2018-07-31
Added support for Oracle datatype.  If you could refer to

grieverbs:
https://docs.oracle.com/cd/E12151_01/doc.150/e12155/oracle_mysql_compared.htm#i1027526 on section 2.3.3.1, mediumint would be suggested as NUMBER(7,0) and mediumint does not exists in Oracle.  INT gets tricky, as liquibase translate it as INTEGER.  Oracle treats INTEGER as NUMBER(38,0) which is like a BIGINT or the maximum size support.  Following the same documentation, it's probably best to be used as NUMBER(10,0).
#3292


nvoxland:
Thanks!

PR #798, opened 2018-07-31
Fix #CORE-3293 Non global change log parameters are wrongly resolved …

dannnnn:
…in inner files

Fixed changelog comparison
added a test


nvoxland:
Thanks! Cherry picked to 3.6.x

PR #799, opened 2018-08-02
CORE-3296 Fix NPE when missing manifest entries

wesbiggs:
Handle possible null values for manifest entries.


pmihalcin:
I've hit this issue as well. This should be merged straight away.

PR #800, opened 2018-08-03
CORE-3221 datetime string is valid, if no millis part is found

TimGoedde:
The fix is done in the oracle specific class to reduce complications with other databases.
The generated sql part is also correct, if the oracle column datatype is a timestamp.


zayats103:
When it could be merged? Have a big need for it

PR #802, opened 2018-08-09
CORE-3242 fix version check for MS SQL Server 2008

joggeli34:
Fix version comparison to allow MS SQL 2008 as intended by the log entry below.


nvoxland:
Merged duplicate #850 so it could go into the 3.6 branch. Thanks for the fix!

PR #803, opened 2018-08-10
Fix TIMESTAP WITHOUT TIME ZONE issue for H2 database

wstrzalka:
H2 also doesn't support TIMESTAMP WITHOUT TIME ZONE syntax


nvoxland:
Thanks! Also cherry picked into 3.6.x

PR #804, opened 2018-08-13
Add ability to disable running of Liquibase via CDILiquibaseConfig

adamldoyle:
Currently the only way to disable Liquibase from running in a CDI environment is to set the system property "liquibase.should.run" to false.  It would be more useful for me to be able to provide this via a configuration setting that I can toggle via my producer for CDILiquibaseConfig.  This way I can set appropriate behavior in my application via config files instead of relying on system properties to be set. Passivity is maintained through all of this by still supporting the existing process and by defaulting the new process to enable by default.

### DIFF
--- a/liquibase-cdi/src/main/java/liquibase/integration/cdi/CDILiquibase.java
+++ b/liquibase-cdi/src/main/java/liquibase/integration/cdi/CDILiquibase.java
@@ -115,6 +115,10 @@ public class CDILiquibase implements Extension {
             ));
             return;
         }
+        if (!config.getShouldRun()) {
+            log.info(LogType.LOG, String.format("Liquibase did not run on %s because CDILiquibaseConfig.shouldRun was set to false.", hostName));
+            return;
+        }
         initialized = true;
         try {
             performUpdate();
--- a/liquibase-cdi/src/main/java/liquibase/integration/cdi/CDILiquibaseConfig.java
+++ b/liquibase-cdi/src/main/java/liquibase/integration/cdi/CDILiquibaseConfig.java
@@ -15,6 +15,7 @@ public class CDILiquibaseConfig {
     private Map<String,String> parameters;
     private boolean dropFirst;
     private String defaultSchema;
+    private boolean shouldRun = true;
 
     public String getContexts() {
         return contexts;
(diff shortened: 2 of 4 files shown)


nvoxland:
It's a good change. What is the reason for checking the liquibase.config.shouldRun property in addition to the normal liquibase.shouldRun property?

adamldoyle:
We're really just using that liquibase.config.shouldRun property as a way to toggle on/off the config setting inside of the CDITestProducer.  Normally in this situation we would be using something like Arquillian to manage the CDI test environment or drive it via configuration with microprofile-config, but I didn't want to rock the boat too much and went with a solution that would mostly fit within the current testing environment.

nvoxland:
Could you just use the normal liquibase.shouldRun? That's the global run/not-run system property that all the ways of running liquibase should use.

adamldoyle:
I need to test the scenario where the liquibase.shouldRun system property is set to true but the CDILiquibaseConfig.shouldRun value is false (and also the case where liquibase.shouldRun is false but CDILiquibaseConfig.shouldRun is true - both of which should result in liquibase not running).  This different property is only used from the tests, whereas our actual production code would look something like this (when using microprofile-config):

`
@Inject
@ConfigProperty(name = "liquibase.should.run.local.config.setting", defaultValue = "true")
private boolean shouldRun;

@Produces
@LiquibaseType
public CDILiquibaseConfig createConfig() {
    CDILiquibaseConfig config = new CDILiquibaseConfig();     
    config.setChangeLog("db/liquibase/changelog-master.xml");
    config.setShouldRun(shouldRun);
    return config;
}
`

derekm:
@nvoxland — Are you talking about the if statement added to the CDI Extension? Or the confusingly-named pinhole system property that @adamldoyle uses in the tests to toggle his provided CDILiquibaseConfig instance?

If you’re talking about the if statement added to the CDI Extension, I think Adam added an additional if statement instead of reusing the existing if statement for backwards compatibility with the config file and to avoid having to adapt CDILiquibaseConfig to optionally populate the other Configuration system.

nvoxland:
Yeah, I was not noticing that the liquibase.config.liquibaseShouldRun was just in the tests. That does not bother me as much, then. I'll merge it in. Thanks!

PR #805, opened 2018-08-17
Fix: MariaDB supports fractional digits for timestamp since v5.3.0

leandrocosta:


nvoxland:
Thanks!

nvoxland:
Also cherry picked into 3.6.x

PR #814, opened 2018-09-10
Fixing CORE-3121  Getting a java.io.FileNotFoundException using Liquibase with Spring reactive web

ShaManHFel:
As per Cabot's comment in the issue, "resource loader generates FilteredReactiveWebContextResource and that implementation returns false for exists()." To fix this issue an additional call for exists() is added before trying to call getURL() for the resource. If the resource does not exist, it does not make sense to try to get its URL.


maratynsky:
👍

gahissy:
Hey there. What's your policy on pull-requests? Is there a reason this one wasn't merged?
Thank you.

nvoxland:
Thanks. I've been working on some changes to how the resourceAccessor works in general which is what was holding up looking at this pull request. Sorry for the delay.

PR #815, opened 2018-09-12
Fix for bug: Creation of parent directories when they don't exist, wh…

desiderati:
The plugin does not create the parent directory for changelog file if it does not exist. Suppose you set the changeLogFile property with src/main/resources/db/changelog.xml and the db directory does not exist, an error will occur: FileNotFoundException: src/main/resources/db/changelog.xml (No such file or directory) -> [Help 1]

I know that will only happen during the plugin's first execution. But why not just create the directory when the file does not exist?

https://liquibase.jira.com/browse/CORE-3174

### DIFF
--- a/liquibase-core/src/main/java/liquibase/diff/output/changelog/DiffToChangeLog.java
+++ b/liquibase-core/src/main/java/liquibase/diff/output/changelog/DiffToChangeLog.java
@@ -82,6 +82,7 @@ public class DiffToChangeLog {
         this.changeSetPath = changeLogFile;
         File file = new File(changeLogFile);
         if (!file.exists()) {
+            file.getParentFile().mkdirs();
             LogService.getLog(getClass()).info(LogType.LOG, file + " does not exist, creating");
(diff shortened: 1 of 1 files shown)


desiderati:
Hi thanks.

desiderati:
Hey guys, error still happening on version 3.8.1!!! Directory **db/changelog** does not exists.

Failed to execute goal org.liquibase:liquibase-maven-plugin:3.8.1:generateChangeLog (default-cli) on project treasy-tree-editor-server: Error setting up or running Liquibase: liquibase.command.CommandExecutionException: java.io.FileNotFoundException: src/main/resources/db/changelog/0-initial-schema.xml (No such file or directory)

SteveDonie:
Thanks - we will take a look at this and get this fixed in an upcoming 3.8.x version. Datical, the company I work for, has recently started investing more time and effort into Liquibase and we will be responding much more quickly to issues.

desiderati:
Thanks ;)

PR #816, opened 2018-09-12
Fix for bug: Error while including resources with includeAll directiv…

desiderati:
If we try to access a resource present in the Application's Classpath, when we are using a JBoss / WildFly server, this resource is made available through the server's own file system, called VFS.

There are two basic ways we may deploy a Web Application on a JBoss / WildFly server. Or via exploded War, or via War artifact itself.

In both cases, if you try to access for example the db/changelog/20180223-12-45-changelog.xml resource in the my-app.war application, the JBoss / WildFly server will generate the following path for the resource:

1) War Exploded
vfs:/<my-local-projects-path>/my-app/target/my-app/WEB-INF/classes/db/changelog/20180223-12-45-changelog.xml

2) War Artifact
vfs:/content/my-app.war/WEB-INF/classes/db/changelog/20180223-12-45-changelog.xml

Even worse, if the resource is present in a Jar file and not inside the War application itself, the path for it will be:

1) War Exploded
vfs:/<my-local-projects-path>/my-app/target/my-app/WEB-INF/lib/my-lib.jar/db/changelog/20180223-12-45-changelog.xml

2) War Artifact
vfs:/content/my-app.war/WEB-INF/lib/my-lib.jar/db/changelog/20180223-12-45-changelog.xml

As you can imagine, liquibase until now does not have support for VFS resources inside a Jar file or directly in a War file. Basically what we need to do is to change the ClassLoaderResourceAccessor class so when it is listing the resources defined by the includeAll directive, it knows how to handle VFS resources.

I've tried to NOT use Google Reflections Lib to patch this error, but I couldn't find an easy and simple way. So if you guys are interested in a solution using Google Reflections, I can send a PR.

Here, as attachment (Lines 137:178), is the code to solve the problem using Google Reflections.

https://liquibase.jira.com/browse/CORE-3177

### DIFF
--- a/liquibase-cdi/pom.xml
+++ b/liquibase-cdi/pom.xml
@@ -33,6 +33,7 @@
 		<dependency>
 			<groupId>org.jboss.weld.se</groupId>
 			<artifactId>weld-se</artifactId>
+			<version>1.1.12.Final</version>
 		</dependency>
 
 		<!-- provided scope -->
--- a/liquibase-core/pom.xml
+++ b/liquibase-core/pom.xml
@@ -144,6 +144,11 @@
             <artifactId>logback-classic</artifactId>
         </dependency>
 
+		<!-- Google Reflections -->
+		<dependency>
+			<groupId>org.reflections</groupId>
+			<artifactId>reflections</artifactId>
+		</dependency>
     </dependencies>
 
 	<build>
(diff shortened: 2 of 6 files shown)


nvoxland:
I ended up reverting this after re-considering on the google reflections dependency. 

I'm working on some larger changes with 3.7 that will hopefully sidestep the issue.

PR #817, opened 2018-09-12
Fix for bug: Error while including resources with includeAll directive.

desiderati:
https://liquibase.jira.com/browse/CORE-3192


nvoxland:
Thanks!

PR #821, opened 2018-10-02
CORE-3100 fix

nflferreira:
Change-Id: I675d62ae759793e0656fa549fac747def53689ba

Fix on diff databases. Besides the description on core-3100, this was also claiming inexistent differences.


nvoxland:
Thanks!

PR #822, opened 2018-10-03
Fix bug: can't set some parameters for snapshot & generate changelog

li-on:
else if was in wrong place
some fields was missing in class definition


nvoxland:
Thanks!

PR #823, opened 2018-10-06
[Fix Bug] fix nullPointerException when migrating RedShift

tyrelltle:
in ColumnSnapshotGenerator, it is trying to get "NULLABLE" column metadata which is Integer type, and assign to an int variable. 
When running migration for RedShift, the column metadata "NULLABLE" is null, and when assigning null to int variable it raises NullPointerExeception because can not unbox null value


nvoxland:
Thanks!

PR #826, opened 2018-10-16
CORE-3316 Add 'SELECT' statements for populating column data.

bsandiford:
Also:
- Remove some Copyright statements that should never have been committed (from a previous change I submitted)

### DIFF
--- a/liquibase-core/src/main/java/liquibase/change/AbstractChange.java
+++ b/liquibase-core/src/main/java/liquibase/change/AbstractChange.java
@@ -1,9 +1,16 @@
 package liquibase.change;
 
 import liquibase.Scope;
+import liquibase.change.core.LoadDataChange;
+import liquibase.change.core.VariableConfig;
 import liquibase.changelog.ChangeSet;
 import liquibase.database.Database;
-import liquibase.exception.*;
+import liquibase.exception.LiquibaseException;
+import liquibase.exception.RollbackImpossibleException;
+import liquibase.exception.SetupException;
+import liquibase.exception.UnexpectedLiquibaseException;
+import liquibase.exception.ValidationErrors;
+import liquibase.exception.Warnings;
 import liquibase.parser.core.ParsedNode;
 import liquibase.parser.core.ParsedNodeException;
 import liquibase.plugin.AbstractPlugin;
@@ -21,7 +28,16 @@ import java.beans.PropertyDescriptor;
 import java.lang.reflect.Method;
 import java.lang.reflect.Modifier;
 import java.lang.reflect.Type;
-import java.util.*;
+import java.util.ArrayList;
+import java.util.Arrays;
+import java.util.Collection;
+import java.util.HashMap;
+import java.util.HashSet;
+import java.util.List;
+import java.util.Map;
+import java.util.Set;
+import java.util.SortedSet;
+import java.util.TreeSet;
 
 /**
  * Standard superclass to simplify {@link Change } implementations. You can implement Change directly, this class is
@@ -34,6 +50,7 @@ import java.util.*;
 public abstract class AbstractChange extends AbstractPlugin implements Change {
 
     protected static final String NODENAME_COLUMN = "column";
+    protected static final String NODENAME_VARIABLE = "variable";
     private ResourceAccessor resourceAccessor;
 
     private ChangeSet changeSet;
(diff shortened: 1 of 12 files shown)


bsandiford:
Nathan - Updated this PR to match current master branch.  thanks!  Bob.

ro-rah:
Hi @bsandiford , would you resolve the conflicts on this PR please?

For now I will size at a Med to:
- allow Eng time to resolve conflict
- thanks for adding tests!

rpsandiford:
I'll try to get to it this week...

ro-rah:
Thanks for the quick reply @rpsandiford !

bsandiford:
Ronak - I'm having problems building.  I did a clean clone of liquibase/liquibase, and master is failing on a couple of unit tests:
[ERROR] Errors:
[ERROR]   H2IntegrationTest>AbstractIntegrationTest.testAbsolutePathChangeLog:907 » ChangeLogParse
[ERROR]   HsqlIntegrationTest>AbstractIntegrationTest.testAbsolutePathChangeLog:907 » ChangeLogParse
[INFO]
[ERROR] Tests run: 396, Failures: 0, Errors: 2, Skipped: 312

I'm trying to build with:
  Maven 3.6.2
  Oracle JDK 1.8.1_221

Thoughts?

ro-rah:
@bsandiford good to hear!

bsandiford:
Conflicts resolved.  Builds locally for me (both 'package' and 'install' targets).

bsandiford:
OK - I'm gonna need help understanding what the problems are here.  Looks like it is complaining about test coverage - but I'm not sure what the numbers mean or what the resolution will be.

bsandiford:
OK - looks like we're good.  Over to you, Ronak!

ro-rah:
@bsandiford I have sent it to conditioning!  Thanks again!

bsandiford:
Hi, Erzsebet.  As far as I can tell by double checking the code against the description in the JIRA story, they match up OK.

One thing I did notice - the JIRA has fix version 3.9.0.  I've updated the PR to be against 4.0, so the JIRA needs updating for fix version - I don't have permissions to do that.

nvoxland:
( copied the jira description into this ticket since we're moving away from jira)

bsandiford:
Sorry - didn't realize the JIRA stories were now obsolete.  Anything future I'll do through Github Issues.

nvoxland:
No problem, it's still a transition in progress :)

nvoxland:
This is a great feature, and see the value of it. There are a couple things I'm thinking about:

1. It's using the same ${} syntax we use for changelog property substitution. I like that it doesn't introduce new syntax, but will we run into conflicts in usage? Is it better to think about these as changeSet-scoped properties? If we think about changeSet scoped properties as a thing, how would those work for not just this but other use cases?

2. We do have valueComputed already on <column> which is used for computed values in general. Should this select behavior be a different attribute? Or part of the same? The "selectNoResults" feature is good, but how does that fit with the more generic valueComputed?

3. The new "<variable>" name wasn't immediately obvious to me what it was for. Maybe something like <setProperty> or something depending on where we go with question 1?

bsandiford:
Re 1;  Our initial custom implementation used a %xxx% syntax.  The ${} syntax was your suggestion :) but I'm happy to change it back, or to something else.  Back works for me - less changes to make when we move off our custom Liquibase version.

bsandiford:
Re 2: I'm not really familiar with the valueComputed usage.  I'll dig a bit deeper and see about it.  The advantage to the SELECT implementation is it's DB Version independent - and we use Liquibase against multiple DB vendors for the same change sets.  But - if valueComputed can also be vendor independent, that won't be an issue.

bsandiford:
Re 3:  Do you mean the empty value position in the CSV file?  I'm not sure what you're asking on this one, sorry.

bsandiford:
Sorry - my response for number 2 is nonsense.  Of course the SELECT implementation is potentially db vendor specific - have to be careful to write generic code.  I'll still dig into the valueComputed.

bsandiford:
Re: number 2.  How about instead of adding the "select" attribute to the loadData column def, we instead just use the existing defaultValueComputed attribute as appropriate.  And, update the handling of valueComputed and defaultValueComputed to handle the replaceable parameters (in whatever syntax they end up being).  Using the current mechanism, we'd have something like:

blitzenzzz:
I'm inheriting this work from Bob, who is retiring. I know I'm waking the dead a bit with this, but can I get someone to pick this up for a moment and describe what needs to be done? I see that there are merge conflicts, which I can handle, but what, if anything, needs to change in the design of the feature?

filipelautert:
Hey @blitzenzzz - It's a lot of conflicts, but from the comments above I would suggest you to pick up from Bob last comment .
I'll move this PR back to visibility, thanks!

kevin-atx:
@blitzenzzz - checking to see if you've had a chance to resolve the conflicts. I'm happy to get this scheduled into the next release, when you have it updated. Thanks

filipelautert:
Closing this PR after 7+ years. Thank you to @bsandiford for the original work and the thoughtful implementation, and to @blitzenzzz for attempting to pick it up.

The feature idea (SELECT-based column population for loadData) is still valid and useful, but this PR is no longer viable:

- The codebase has changed substantially since 2018, making a rebase essentially a rewrite
- The design questions raised by @nvoxland in May 2020 (variable syntax conflicts, `select` vs `valueComputed`, `<variable>` naming) were never resolved
- The original author has retired and the successor went inactive

If there's still interest in this feature, I'd recommend opening a fresh issue to capture the requirements and design decisions so a clean implementation can be done against the current codebase.

PR #827, opened 2018-10-16
Fixed broken Integration tests.

bsandiford:
- At least one issue was caused by a recent change to the MySQL reserved words (adding 'description'), which also affected the MariaDB, which inherits from MySQL.
- The others are H2 related - as far as I could tell from the on-line docs, the H2 database handling is correct, and the tests are wrong.  So, I modified the tests to match the current db handling.


nvoxland:
Thanks!

PR #829, opened 2018-10-17
WIP: CORE-2375 actively prolong lock 3.5.x

BernhardBln:
Working code, needs some fine-tuning and cleaning.


BernhardBln:
See https://liquibase.jira.com/browse/CORE-2375
Would like to get some feedback here.

razioyo:
SET lock_timeout TO '2s'
can we set lock timeouts while firing queries just curious to know.

BernhardBln:
> SET lock_timeout TO '2s'
> can we set lock time outs while acquiring lock just curious to know.

@Razi007 - not sure if I'm getting your question. If you can change the property? 

At the moment, the writer decides until what time the lock goes stale, so that different readers with (in the worst case) different configuration cannot steal the lock.

Also I think 2s is way too low, there is a chance that then somebody would steal the lock from a service that thinks it still holds it... why you want to set it to 2s?

Razi007:
okey!! my bad , basically you are trying to preserve lock here.There is also an issue with liquibase that is if  liquibase exits unexpectedly it is not releasing lock  on database and when we try to acquire lock again(re-run liquibase migrations) system keep waiting.We can discuss same at different ticket/pr  I guess.

BernhardBln:
Hey @Razi007 - my ticket solves exactly the problem you described :) 

When you use this version, the system will not wait forever in case the instance that created it died, but after a certain time (default: 85 seconds) removes the lock and continue.

liquibase-core/src/main/java/liquibase/Liquibase.java
```diff
@@ -148,6 +157,16 @@ public Database getDatabase() {
         return database;
     }
 
+    /**
+     * Returns the Database used by this Liquibase instance for locking.
+     * <p>
+     * Should be a separate one as some lock services might commit transactions asynchronously,
+     * and hence should not rollback or commit what the actual updates is doing.
```
RichardBradley: This is a separate *connection* to the primary database, not a separate database (in the normal sense of the term) for locking, right?

Are you sure this is needed? All the changelogs operate in their own transactions, so the current `master` version of Liquibase is able to use the same connection for the lock operations as for the main operations. What has changed in that regard to make this necessary?

The "database" terminology here is confusing. It would be a lot clearer to use a separate connection to the primary database (if indeed it is necessary), rather than have it labelled as a separate "database" all over the place. Your hands might be tied here if the Liquibase "database" object actually represents a connection rather than a datbase, but I think Liquibase's wrappers do distinguish between a database and a connection.

It also seems like a bug magnet: if any future changes forget to use the correct "database", they will introduce some rather subtle bugs.

liquibase-core/src/main/java/liquibase/lockservice/ProlongingLockService.java
```diff
@@ -0,0 +1,879 @@
...
+     * in LOCKEXPIRES in the database.
+     */
+    private volatile Date lockExpires = null;
+
+    private volatile ScheduledFuture<?> lockProlonger = null;
+
+    /**
+     * How long shall we wait for a lock before we give up?
+     * <p>
+     * Unit: minutes
```
RichardBradley: Better to use a `java.time.Duration`? 
Time fields without units in their names are conventionally seconds (but I prefer putting units in the name, e.g. `changeLogWaitTimeMinutes`)

liquibase-core/src/main/java/liquibase/lockservice/ProlongingLockService.java
```diff
@@ -0,0 +1,879 @@
...
+ * <p>
+ * Locks older than 100 seconds are considered stale and will get removed.
+ * <p>
+ * Since locks get actively prolonged, do use a very high lock wait time (1 day).
+ * <p>
+ * Must <b>not</b> be used in junction with the {@link StandardLockService}!
+ * <p>
+ * Note: This is in alpha state. We do not use the time of the database server right now,
+ * but a local date, so this might not work properly when several services with very
+ * un-synchronised locks connect to the same database.
```
RichardBradley: It seems like it would be better to always use database time; is that not possible here?

liquibase-core/src/main/java/liquibase/lockservice/ProlongingLockService.java
```diff
@@ -0,0 +1,879 @@
...
+     * Unit: minutes
+     */
+    private Long changeLogLockWaitTime;
+
+    /**
+     * How often shall we recheck
+     * <p>
+     * Should be &gt; {@link #changeLogLockProlongingRateInSeconds},
+     * <p>
+     * Units: seconds
```
RichardBradley: I would definitely use `java.time.Duration` or add units to these names; having different units here but not putting them in the names is a bug magnet

liquibase-core/src/main/java/liquibase/lockservice/ProlongingLockService.java
```diff
@@ -0,0 +1,879 @@
...
+                    return "datetime2(3)";
+                }
+            } catch (DatabaseException e) {
+                // ignore
+            }
+        }
+        return "datetime";
+    }
+
+    private boolean newColumnsExist(Executor executor) throws DatabaseException {
```
RichardBradley: "newColumns" isn't a great method name here: the columns are new now, but the next time this class is modified they won't be, and this method name will be obsolete and confusing. Better to name more like `lockProlongColumnsExist` or similar IMO.

liquibase-core/src/main/java/liquibase/lockservice/ProlongingLockService.java
```diff
@@ -0,0 +1,879 @@
...
+     * <p>
+     * Unit: seconds
+     */
+    private Long staleChangeLogLockRemovalTimeInSeconds;
+
+    private Boolean hasDatabaseChangeLogLockTable;
+    private boolean isDatabaseChangeLogLockTableInitialized;
+    private ObjectQuotingStrategy quotingStrategy;
+
+    private final ScheduledExecutorService executorService = Executors.newScheduledThreadPool(1);
```
RichardBradley: I feel like there are now thread-safety issues in this class that have not been fully thought through.
I think that there need to be a few more synchronised blocks where the main thread interacts with the prolonging thread, but I have not analysed it in detail.

liquibase-core/src/main/java/liquibase/lockservice/ProlongingLockService.java
```diff
@@ -0,0 +1,879 @@
...
+ * <p>
+ * Since locks get actively prolonged, do use a very high lock wait time (1 day).
+ * <p>
+ * Must <b>not</b> be used in junction with the {@link StandardLockService}!
+ * <p>
+ * Note: This is in alpha state. We do not use the time of the database server right now,
+ * but a local date, so this might not work properly when several services with very
+ * un-synchronised locks connect to the same database.
+ */
+public class ProlongingLockService implements LockService {
```
RichardBradley: There is quite a bit of code in this class which is duplicated with StandardLockService, e.g. edge cases around database boolean handling, but there is no mechanism to keep future bug fixes in sync between the two classes.

It would be better from a maintainability point of view if the common code was commonised (perhaps this class could extend StandardLockService?) or if there were code comments advising future developers to make any changes in both places.

Or perhaps all this functionality should go directly into StandardLockService, to avoid this duplication? The old behaviour could be achieved by setting prolong rate to zero, or similar?

RichardBradley:
Note: I am not a Liquibase maintainer; I do not have write access to Liquibase.

This looks like a working solution to a real problem that lots of real world Liquibase users (including me) have: orphaned locks. Nice work!

In my opinion, this seems like not the best possible solution to the problem. I hope that something more like #872 can be made to work: I think that version has fewer moving parts and is simpler and more likely to be correct in all cases. The approach here depends on a few timeout parameters which will need careful tuning and complicates the codebase. (See also my comment inline about naming issues.)

I also don't think this patch is ready to merge as-is. I have put some comments inline with my thoughts on that. Please note that I am not suggesting that you should make a load of changes to satisfy me, and I am certainly not suggesting that Liquibase will merge your PR if you do! That would be up to them. You should probably wait for feedback from the Liquibase maintainers before deciding whether or not to put more work into this PR.

You mentioned on CORE-3044 that it hasn't been merged or received any attention. I hope I won't be speaking out of turn if I say: don't lose heart! These changes are obviously useful to you, and it has been useful to me that you shared them here, thank you.

liquibase-core/src/main/java/liquibase/Liquibase.java
BernhardBln: Yeah. You are totally right about the wording, but if I remember correctly, the liquibase "Database" is what I would call a "Connection" in general. I did not find a "Connection" Object in Liquibase that can be used for this purpose, do you know if that exists?

In fact, the separate connection is needed. My lock service actively prolongs the lock in the background (writing a new timestamp every x seconds), and in order to make this change visible to other clients, it needs to commit the transaction every x seconds, but of course you would not want the actual updates to the database schema (that are still in progress and incomplete) get committed when this happens.

BernhardBln: True this could be a bug magnet, do you have a suggestion how to improve that?

liquibase-core/src/main/java/liquibase/lockservice/ProlongingLockService.java
BernhardBln: Can't use duration here, it came with java 1.8 but we have 1.6 here.
Changing the name sounds like a good idea.

liquibase-core/src/main/java/liquibase/lockservice/ProlongingLockService.java
BernhardBln: As above, can't use duration here, as we have java 1.6 here.
Changing the name sounds like a good idea.

liquibase-core/src/main/java/liquibase/lockservice/ProlongingLockService.java
BernhardBln: It's been quite a while that I wrote this, but I think it works this way: In the lock on the database, we use database time (so that all clients, no matter what local time they have, can verify if the lock is stale), but locally, we use java.util.Date to measure if the lock we just created is stale. 
Probably the comment is also obsolete now, could be I put this when I wrote the first version.
Anyways, should be analysed before integrating!

liquibase-core/src/main/java/liquibase/lockservice/ProlongingLockService.java
BernhardBln: Totally agreed, I saw this more as a POC and did not want to spend time refactoring everything out and then learning it's not getting merged ;-) I also think this could replace the other service, after some time of testing.

liquibase-core/src/main/java/liquibase/lockservice/ProlongingLockService.java
BernhardBln: you think about memory effects? For this I have added volatile to some fields. Or you mean execution order?

liquibase-core/src/main/java/liquibase/lockservice/ProlongingLockService.java
BernhardBln: True, agree. Again, was just a POC, we would have to think how liquibase manages their own database schema...
maybe with liquibase? :D

BernhardBln:
Hey, thanks a lot for your comments, this is the first real feedback I got for this ;-)

Totally agree #872 would be very elegant, but seems it would only work for postgres? 

The advantage of my solution (which also makes it so cumbersome) is that all synchronisation happens through the database, you don't run into the problem that client A holds a lock and has configured a lock timeout of 1 minute, and prolongs every 30 seconds, and client B has a timeout of 15 seconds configured and steals the lock, while A still assumes it still has it. 

Also there are no problems with time zones, as this is already using database time for verification if a lock of another service is stale (RemoveStaleLocksGenerator.java: "... AND LOCKEXPIRES < " + database.getCurrentDateTimeFunction())

Since this hasn't received any official feedback from the maintainers, I decided not to put more effort into this at least until I hear back from them.

However, we were using this for some time and never ran into a problem (but we didn't have a high load on the system anyways).

Again, thank you very much, and let's see if this gets picked up some time :)

RichardBradley:
> Totally agree #872 would be very elegant, but seems it would only work for postgres?

I think there is an equivalent for most DBs, e.g. here is the same thing on MySQL:
https://dev.mysql.com/doc/refman/8.0/en/locking-functions.html#function_get-lock

ro-rah:
Hey @BernhardBln,

Thanks for your pull request!

Here’s what happens next:

BernhardBln:
That's great, it seems this is a problem still open today and annoying :)

Regarding the failed build - for me it's a bit awkward, it was passing and then I created a version by removing the snapshot and then it failed... no code changes involved...

ro-rah:
I updated the target branch to 3.10.x, as we are now having a big push to get last PRs in there.  Updated to RiskHigh since build is not passing but passing it to eng to see if they want to include this.

BernhardBln:
Sweet. I’d also be interested in some feedback whether my approach was ok, as I was totally new to the liquibase code base when I implemented this. It also should be tested on different systems. However, what I implemented was turned off by default when I remember correctly.

ro-rah:
Hi @BernhardBln , glad you are still active on this PR!  Would you like to help and take a look at the merge conflict listed here?  

I have this conditioned and going to our internal core eng. team but any work we can do before then (like getting rid of the merge conflict) would greatly help the team :)

ro-rah:
@BernhardBln we will also provide feedback on the approach :)

BernhardBln:
This was a quite a merge hell, well, it's been two years...

At least it's compiling now, and tests were also running on my machine ;-)

It's been ages I was on it, I think AbstractIntegrationTest should be given some attention (that it also properly tests my new stuff), I reverted a change of mine in there because it didn't work well together with the output that is generated (since you are supposed to use two different db connections with my change, to prevent that a rollback of changes also rolls back changes to the lock table, the create DATABASECHANGELOGLOCK was missing in the output).

It should be separated again and the output be unified somehow.

Also, I copy-pasted the Standard Lock Strategy back then (hoping the standard one could be dumped at some point), hope I'm not missing changes now made to that one...

Anyways, would be nice if someone can take a look at this, also check the discussion that already started back then but was not really continued above, and tag me if you need any help.

BernhardBln:
Hi, agreed, but please also consider that the main change went into a new class that is only ever used when explicitly activating it via a property. Otherwise it will not be used. Like a feature switch.

You could focus first if my changes to the existing code is fine and expose the new property as experimental, and then take more time to test the new code. 

I'm writing this because the merge was quite big (this PR is 2 years old), and I would prefer if this gets integrated sooner than later to not have another big merge ;-)

ro-rah:
@BernhardBln , thanks for staying engaged on this ticket.  Would you be willing to speak with me during a 30 min max call?  Would love to know what keeps you motivated and how we can spread that enthusiasm to the rest of the liquibase community.  Would you care to email me at ronak@liquibase.org please :)

-Ronak

ro-rah:
@nvoxland , Mike, Caroline and I just spoke with @BernhardBln, and he enlightened us on the use case for this within his SpringBoot / hibernate / microservices world this leftover lock is a common occurrence in almost every project he has worked on.

Many reasons for this, one example given: a container can spin up, migration that takes too long gets killed an restarted (which can occur often) and then a db lock on the tracking table is left.

nvoxland:
Yes, that is a common issue that we need a solution for. I do tend to worry that a solution based around auto-timing out locks will either still be too long to wait or will unknowningly release the lock for unexpectedly long updates. 

There are a few other open PRs that are trying to use more database-specific options to remove the lock when the connection closes. The final feature change will probably need to be a combination of that as well as this. But, we need some time to collect together the use cases we're wanting to make sure we cover as well as the different ways we have to solve it.

In the mean time, I've solved it through a couple mechanisms:
1. Move the liquibase update logic from the "startup" phase into an "init" phase. How exactly you do that depends on your PAAS, but you can usually define much longer-lived init phases without the system killing the process.

2. Use the https://github.com/liquibase/liquibase-nochangeloglock extension which makes liquibase just not use the databasechangeloglock functionality. It is still important that you don't have multiple instances of liquibase running against the same database at the same time, but if you are already doing that through some other platform-specific mechanism there is really no need for liquibase to also be watching that for you.

ro-rah:
Thanks for the response @nvoxland !  @BernhardBln would those workarounds be helpful till we get a chance to look at lock removal at a more db platform specific level?

BernhardBln:
Hey, thanks for your reply. 

Option 2 is not really an option in micro service environments. It might be technically possible, but I’d really not prefer to rely on it. 

Option one we already considered, but in the cloud the timeout is not the only reason why a service can crash or disappear. 

Regarding „will unknowningly release the lock for unexpectedly long updates“ - I took that into consideration, the service that is updating is actively prolonging the timeout for the lock. 

So for example you can configure that the lock gets removed after one minute, which I think is an ok time to wait after the service crashed, but the updating service will refresh the timeout every ten seconds. So even if the update takes much longer, the lock would not be removed. 

That is the reason why I introduced a second database connection, as I cannot refresh the lock with the one used for migration.

BernhardBln:
Regarding other solutions checking closed database connections - if that works reliably it would be more elegant and simpler of course.

RichardBradley:
I'm glad to see some activity on this and on liquibase more widely.
I would like to re-iterate [my comment from May 2019](https://github.com/liquibase/liquibase/pull/829#issuecomment-494730863): orphaned locks are the most pressing bug for Liquibase, but I don't think this PR is the best way to fix it. (Lots more details about why in that comment)

In my opinion, the best path to fix orphaned locks is:

 1. Switch to a model where Liqiubase only locks when needed (i.e. "check - lock - check" or "double checked locking"). This will be an easy fix and will reduce the number of liquibase locks in production by 90%+, which will make orphaned locks a lot rarer. See https://liquibase.jira.com/browse/CORE-3044 and https://github.com/liquibase/liquibase/pull/846

 2. Implement, for each database, a mechanism to reliably release a lock server-side when a client dies. This looks like https://github.com/liquibase/liquibase/pull/872 for postgres, but there are similar mechanisms for each other RDBMS. This will be a fair amount of work.

 3. Then, I think the next most urgent work for Liquibase is getting the integration test suite working again, so that users are able to submit pull requests for small bug fixes with confidence. I worked on this a bit and corresponded with @nvoxland briefly about this in June 2019 and he said that fixing the integration tests was on hold pending a rewrite of the core functionality. Personally, I think a rewrite/refactor is not the most useful way forward for users of the project, but I'm not the one putting in the work here, so I have a limited say in that! :-)

(I think 3 would help with 2.)

Thanks for Liquibase, I think it's a great project.

molivasdat:
Hi @BernhardBln Is this code complete?  We are looking at merging this into the 3.10.x release.

BernhardBln:
Hi,

functionality-wise it is, also the feature is off by default, but there were also some changes in other parts of the software, so please do a good review 😉 but happy to read this gets finally merged 🙃

ziolkow:
Hallo @BernhardBln I just want to ask if there is any progress on it? Is there anything I could do to help there?

BernhardBln:
Hey @ziolkow I'm a bit confused now, please check my comment just above your comment ;-)

Basically the functionality I wanted is in there, it is hidden away behind a feature toggle (and off by default), but I also had to change existing code for this to work, and these changes would be active already.

And I had to merge this a couple of times, so I would need someone from your side to do a good review on my changes, but functinality-wise, it should be done.

BernhardBln:
I resolved the merge conflicts. The target branch has failing tests, I fixed one. 

However, this should be reviewed and merged soon, it's already lying around since 2 years and every new merge conflict introduces lots of work.

molivasdat:
Hi @BernhardBln Thank you for fixing the merge conflicts in 3.10.x.  It's really appreciated.  We been reviewing the code for a couple of days as well and we are concerned with the scope of the changes involved.  It runs across the entire locking service of Liquibase.  We are discussing how to best move forward to address your issue and do it safely and completely for the entire community.  We will talk with @nvoxland this week to come up with options on how to move forward.

RichardBradley:
> In my opinion, the best path to fix orphaned locks is:
> ...
> 1. Switch to ... "double checked locking"
> ...
> 2. Implement, for each database, a mechanism to reliably release a lock server-side when a client dies. ...

2 is fixed very well by the plugin https://github.com/blagerweij/liquibase-sessionlock . It would be nice to bring that into core as it is strictly better than the standard lock. I think that plugin obsoletes this Pull Request

1 is still open, see my comment above for a PR which works towards a fix but which wasn't taken up by the core maintainers.

BernhardBln:
Agreed, also I wouldn't be willing to resolve merge conflicts for a third time :) Thanks @RichardBradley for mentioning the plugin!

RichardBradley:
Looks like #2105 is a newer version of a fix for 1 "double checked locking" (but Spring only)

RichardBradley:
Looks like #1453 is an official ticket from nvoxland tracking locking issues. I have commented there

PR #830, opened 2018-10-18
CORE-3262 fix: FileSystemResourceAccessor fails with non-hierarchical…

iorjfeiofjseoirf:
… URLs on class path

Multi-release Jars on the classpath caused a java.lang.IllegalArgumentException "URI is not hierarchical" in FileSystemResourceAccessor::addRootPath()

### DIFF
--- a/liquibase-core/src/main/java/liquibase/resource/FileSystemResourceAccessor.java
+++ b/liquibase-core/src/main/java/liquibase/resource/FileSystemResourceAccessor.java
@@ -3,6 +3,7 @@ package liquibase.resource;
 import liquibase.exception.UnexpectedLiquibaseException;
 
 import java.io.*;
+import java.net.URI;
 import java.net.URISyntaxException;
 import java.net.URL;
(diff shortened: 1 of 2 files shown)


flozano:
Any chance of this being merged? we can't upgrade JDK without this.

nvoxland:
Merged in, thanks. Was this an issue you saw in 3.6 as well? Or just in master?

vbauer:
[FYI] We faced with it in version 3.6

iorjfeiofjseoirf:
[FYI] I was able to reproduce the issue in versions 3.5.5, 3.6.1 and master.

flozano:
Thanks for merging. When is release planned including this change? a 3.6.x hotfix would be enough probably...

iorjfeiofjseoirf:
This fix is part of Liquibase Release [3.6.3](https://github.com/liquibase/liquibase/releases/tag/liquibase-parent-3.6.3).

PR #833, opened 2018-10-26
prints the error code on Exception

ayacub:
SQLException includes an integer errorCode that can be very useful in debugging when something goes wrong.

In general, the more information available when a mistake happens the better.

The particular case that made me create this patch was calling a stored procedure where the message was just `Unhandled user-defined exception condition` (in mysql). But seeing the error code `1051` that can be looked up to mean "Unknown table" let us quickly identify the problem.


nvoxland:
Thanks!

PR #839, opened 2018-12-12
Do not use absolute path when loading command line properties from re…

tgyurci:
…source

### DIFF
--- a/liquibase-core/src/main/java/liquibase/integration/commandline/Main.java
+++ b/liquibase-core/src/main/java/liquibase/integration/commandline/Main.java
@@ -441,7 +441,7 @@ public class Main {
                 if (potentialPropertyFile.exists()) {
                     parseDefaultPropertyFileFromFile(potentialPropertyFile);
                 } else {
-                    parseDefaultPropertyFileFromResource(potentialPropertyFile);
+                    parseDefaultPropertyFileFromResource(potentialPropertyFile.getPath());
                 }
             } catch (IOException e) {
                 throw new CommandLineParsingException(e);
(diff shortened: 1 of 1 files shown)


liquibase-core/src/main/java/liquibase/integration/commandline/Main.java
```diff
@@ -441,7 +441,7 @@ protected void parseDefaultPropertyFiles() throws CommandLineParsingException {
                 if (potentialPropertyFile.exists()) {
                     parseDefaultPropertyFileFromFile(potentialPropertyFile);
                 } else {
-                    parseDefaultPropertyFileFromResource(potentialPropertyFile);
+                    parseDefaultPropertyFileFromResource(potentialPropertyFile.getPath());
```
gerbrand: (Reviewing as bystandard) File is a more specific type, I'd still use File as type. File instances can be relative too, no need to convert to a string-path.

liquibase-core/src/main/java/liquibase/integration/commandline/Main.java
```diff
@@ -455,10 +455,10 @@ protected void parseDefaultPropertyFiles() throws CommandLineParsingException {
      * @throws IOException if the file cannot be opened
      * @throws CommandLineParsingException if an error occurs during parsing
      */
-    private void parseDefaultPropertyFileFromResource(File potentialPropertyFile) throws IOException,
```
gerbrand: Same here, seems the method's goal is to read a (local) file, not a resource-reference.

nvoxland:
I've redone how the file access works in master and I think the changes should resolve this issue.

PR #841, opened 2018-12-14
Call Class.getConstructor.newInstance instead of newInstance

ArloL:
First of all thanks for all the great work and providing it as open source code :)

I would like to report a problem that @Nikos410 and I have come across while testing JDK 11 with our applications:

`Class.newInstance` is deprecated as of Java 9. Usage should be replaced by calling `Class.get(Declared)Constructor.newInstance` which is backwards compatible. See [java.lang.Class.newInstance](https://docs.oracle.com/en/java/javase/11/docs/api/java.base/java/lang/Class.html#newInstance()) for more information.

Not only is newInstance deprecated but using it also creates a problem testing our applications with JDK11. There even is a comment in the source code of newInstance about possible problems with the current memory model: http://hg.openjdk.java.net/jdk/jdk/file/76072a077ee1/src/java.base/share/classes/java/lang/Class.java#l543
We are experiencing such a problem and can't start using JDK 11 due to this issue. I can explain it in more detail if necessary but don't feel like it is exactly necessary to argue for the removal of a deprecated call.

### DIFF
--- a/liquibase-core/src/main/java/liquibase/AbstractExtensibleObject.java
+++ b/liquibase-core/src/main/java/liquibase/AbstractExtensibleObject.java
@@ -264,7 +264,7 @@ public class AbstractExtensibleObject implements ExtensibleObject {
                 Object value = this.get(attr, Object.class);
                 if (value instanceof Collection) {
                     try {
-                        Collection valueClone = (Collection) value.getClass().newInstance();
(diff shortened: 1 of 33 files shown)


nvoxland:
Thanks! Merged in the pull request and will cherry pick it into 3.6.x

PR #843, opened 2018-12-19
CORE-2621: Adding tagDatabase support to formatted SQL

lucasmorano:
Adding support to tag changeset, it can be used such as:

```
--changeset morano:8 tagDatabase:”1.2.4”
create table table5 (
    id int primary key
);
--rollback drop table table5;
```

or

```
--changeset morano:8 tagDatabase:1.2.4
create table table5 (
    id int primary key
);
--rollback drop table table5;
```

Without quotes.

Few people who also missed this feature:
https://stackoverflow.com/questions/34376604/tagdatabase-and-include-sql-syntax-for-liquibase/
http://forum.liquibase.org/topic/tag-changset-in-sql-format
Task:
https://liquibase.jira.com/browse/CORE-2621


nvoxland:
I'm thinking the format for this should be a bit different. 

The "tag" command is just one example of other "higher level" sql changes and/or non-sql commands that would be good to allow to be ran in formatted SQL. For example, introduceLookupTable or potential future commands like "backupDatabase".

To handle that, I think the tag or whatever command should be another line in the changeSet rather than part of the changeSet.

More like:
```
--changeset morano:8 
--action tagDatabase tag:1.2.4
create table table5 (
    id int primary key
);
--rollback drop table table5;
```

or even command-only changeSets like:
```
--changeset morano:8 
--action tagDatabase tag:1.2.4
--action backup path:/tmp/backup.zip

--changeset morano:9
create table table5 (
    id int primary key
);
--rollback drop table table5;
```

ro-rah:
Hi @lucasmorano !

The PR now has conflicts, would you be willing to resolve?

For now, I will size this at a RiskHigh to allow for Eng time to sort through @nvoxland and the submitted PR.  Thanks for including tests!

lucasmorano:
Hello @ro-rah !
As you've requested I just solved the conflicts. Thanks for bringing this back up!

ro-rah:
@lucasmorano thanks for responding so quickly!  reducing to a RiskMed

PR #844, opened 2018-12-30
CORE-3367: Clean up Yaml ChangeLog file InputStream

scottsteen:
Parsing a yaml ChangeLog left an open InputStream from the file. If the process parsing the ChangeLog was long-running (such as a Gradle daemon) the file would be locked until the process finished.


nvoxland:
Thanks!

PR #846, opened 2019-01-18
Springliquibase lock only if needed

dancek:
The current policy of always running update has the side effect of
always acquiring the database changelock lock. This has many downsides,
for example:
 - if the application crashes or is shut down while it has the changelog
   lock, consecutive restarts will fail (this is surprisingly common in
   cloud environments)
 - multiple instances of an application cannot start simultaneously as they
   all need to get the lock.

This commit adds logic to first check if there are any unrun changesets.
If no changesets would be applied, update is not called at all.

---

I'm happy to receive any review comments and willing to make reasonable edits.

### DIFF
--- a/liquibase-core/src/main/java/liquibase/integration/spring/SpringLiquibase.java
+++ b/liquibase-core/src/main/java/liquibase/integration/spring/SpringLiquibase.java
@@ -3,6 +3,7 @@ package liquibase.integration.spring;
 import liquibase.Contexts;
 import liquibase.LabelExpression;
 import liquibase.Liquibase;
+import liquibase.changelog.ChangeSet;
 import liquibase.configuration.ConfigurationProperty;
 import liquibase.configuration.GlobalConfiguration;
 import liquibase.configuration.LiquibaseConfiguration;
@@ -33,6 +34,7 @@ import java.net.URLConnection;
 import java.sql.Connection;
 import java.sql.SQLException;
 import java.util.HashSet;
+import java.util.List;
 import java.util.Map;
 import java.util.Set;
 import java.util.jar.Attributes;
@@ -302,7 +304,7 @@ public class SpringLiquibase implements InitializingBean, BeanNameAware, Resourc
             c = getDataSource().getConnection();
             liquibase = createLiquibase(c);
             generateRollbackFile(liquibase);
-            performUpdate(liquibase);
+            performUpdateIfNeeded(liquibase);
         } catch (SQLException e) {
             throw new DatabaseException(e);
         } finally {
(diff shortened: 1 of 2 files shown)


nvoxland:
I like the idea, but could you move the code to StandardLockService instead so it works across all ways LIquibase is used rather than just SpringLiquibase?

dancek:
@nvoxland it looks (and feels) like a LockService knows nothing about changesets (either the available or already run ones). I can't think of a clean way to have this logic inside StandardLockService.

The logic could probably exist in the Liquibase class, though. Should I reimplement the check there?

dancek:
@nvoxland how should I continue with this PR?

BernhardBln:
@dancek - I tried solving the problem with stale locks in cloud environments by prolonging locks actively, and having them expire when the application that created the lock crashed.
My PR didn't get much attention until now though ;-) 
https://github.com/liquibase/liquibase/pull/829

BernhardBln:
PS: we're already using my fix in production, no issues so far

dancek:
We're also using a wrapping class in production so the issue doesn't affect us, but I felt like this would be useful to other people.

@nvoxland I'm still willing to rework the fix if you want it in some form.

RichardBradley:
This would be a great optimisation. See also https://liquibase.jira.com/browse/CORE-3044

liquibase-core/src/main/java/liquibase/integration/spring/SpringLiquibase.java
```diff
@@ -340,47 +342,58 @@ private void generateRollbackFile(Liquibase liquibase) throws LiquibaseException
...
+            if (tag != null) {
+                liquibase.update(tag, new Contexts(getContexts()), new LabelExpression(getLabels()));
+            } else {
+                liquibase.update(new Contexts(getContexts()), new LabelExpression(getLabels()));
+            }
+        }
+    }
+
+    protected void performUpdateIfNeeded(final Liquibase liquibase) throws LiquibaseException {
+        if (updateNeeded(liquibase, new Contexts(getContexts()), new LabelExpression(getLabels()))) {
```
admigator: @dancek thanks for this solution! We tried to apply it and find out that adding 

`ChangeLogHistoryServiceFactory.getInstance().getChangeLogService(liquibase.getDatabase()).reset();`

 before performUpdate(liquibase);  helps to avoid unnecessary DB update in case of multiple server nodes simultaneous startup.

RichardBradley:
It looks like the code currently in this PR has a race condition when changes are needed. See discussion on https://liquibase.jira.com/browse/CORE-3044

dancek:
That's unfortunate.

We use a workaround in our codebase to avoid building a custom version of the library. Since I have no need for this PR personally and I estimate the chances of it being merged to be near zero, I'm closing it.

RichardBradley:
Now moved to #2105

PR #857, opened 2019-02-09
Fix Turkish uppercase column keys bug

camsaul:
[`Liquibase.listUnrunChangeSets` is broken when the user's Locale is Turkish](https://github.com/liquibase/liquibase.github.com/issues/125). (See also related bugfix PR: #751)

If your locale is set to Turkish `String.toUpperCase()` surprisingly will putting dots on things for you. 

e.g. 

```clojure
;; with Locale set to Turkish
(.toUpperCase "filename") ;; -> "FİLENAME"
```

The right solution is to always use `Locale.US` for things like 

```clojure
(.toUpperCase "filename" Locale/US) ;; -> "FILENAME"
```

which are used as keys elsewhere in the codebase. 

Fixes https://github.com/liquibase/liquibase.github.com/issues/125


nvoxland:
Thanks!

PR #858, opened 2019-02-11
Generate diff with onDelete and onUpdate

Uraty:
When I am generating diff for database that has onDelete = CASCADE, liquibase sees it, but doesn't add to changelog. It just creates drop constraint and add constraint without onDelete statement.

To solve this, I've added
```
addFkChange.setOnDelete(fk.getDeleteRule());
addFkChange.setOnUpdate(fk.getUpdateRule());
```
to `ChangedForeignKeyChangeGenerator#fixChanged`


nvoxland:
Thanks!

PR #859, opened 2019-02-11
Handle java.sql.Time when setting prepared statement parameter values

bgK:
When loading data from a CSV file using `<loadData>`, `ISODateFormat` constructs a `java.sql.Time` value when the column format is `HH:mm:ss`.

`PreparedStatement.setTime()` needs to be called for those values for the PostgreSQL JDBC driver to properly understand the parameter.

This is a regression we noticed when upgrading to liquibase 3.6 from 3.5. Liquibase 3.5 did not use prepared statements in that case.


nvoxland:
Thanks!
Also cherry picked to 3.6.x branch

PR #861, opened 2019-02-15
Update versions for debian and rpm

adamtnagy:
Hello,

When building 3.6.x branch locally, the debian and rpm versions need to be bumped too to 3.6.4-SNAPSHOT so Maven can resolve the parent POMs.

Best regards,
Adam


nvoxland:
Thanks

PR #863, opened 2019-02-20
Fix file replace

n3utrino:
The autoclosable does not close the streams, therefore the file cant be deleted


n3utrino:
Maybe we should cherrypick this to the 3.6.x branch.
Because the feature is not working.

nvoxland:
Thanks. Also cherry picked into 3.6.x

PR #864, opened 2019-02-20
Can't locate relative path if master file is in root folder [CORE-3386]

regrog:
fixed https://liquibase.jira.com/browse/CORE-3386

I also created few tests to ensure the expected behavior.


nvoxland:
I cherry picked this to 3.6.x since the file access code is completely rewritten in master.

regrog:
Thanks,
I'll try my case in master to check if it's still there.

PR #866, opened 2019-02-26
Use computeIfAbsent to avoid concurrency issues

Olard:
computeIfAbsent avoids concurrency issues when multiple upgrades run simultaneously


nvoxland:
Thanks!

tandr:
No, unfortunately it does not. The collection - HashMap - is not concurrency safe. This postpones it, and makes it efficient, but it does not fix the problem.

PR #867, opened 2019-03-07
Cleanup: deleted SystemUtils copy-pasted from commons-lang

feoktant:
Big utility file SystemUtils was copied from commons-lang 1.4, however it is barely used. Lets remove it make code cleaner.


nvoxland:
Thanks!

PR #870, opened 2019-03-11
Changed for-loop for addAll method on collections

feoktant:
Such change should increase performance and make code cleaner.


nvoxland:
Thanks!

PR #871, opened 2019-03-14
CORE-3400 added "manifoldclass:" to list of skipped URL protocols in …

ugrepel:
DefaultPackageScanClassResolver is present but unused in current master (due to switch to standard java ServiceLoader usage). I then created this pull request for the 3.6.x branch, currently 3.6.4-SNAPSHOT.

The 3.6.4-SNAPSHOT doesn't build cleanly (some tests fail), but the same tests fail before and after applying this change.

skipping tests for testing my change allowed creating a local snapshot, which fixes the issue I reported in CORE-3400.

### DIFF
--- a/liquibase-core/src/main/java/liquibase/servicelocator/DefaultPackageScanClassResolver.java
+++ b/liquibase-core/src/main/java/liquibase/servicelocator/DefaultPackageScanClassResolver.java
@@ -186,6 +186,13 @@ public class DefaultPackageScanClassResolver implements PackageScanClassResolver
                     continue;
                 }
 
+                // manifold classes should be skipped (no physical jar file present)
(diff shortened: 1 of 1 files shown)


nvoxland:
Thanks!

nvoxland:
Did you test to see if the new ServiceLoader code in master works with your setup?

ugrepel:
Yes, I did test it. I had to run a build with "ignore tests" however since some unittests currently fail (with or without my patch - the same failures). I then used the patched liquibase-core.jar in my setup with manifold, and the expected debug log output was there: no more stacktraces, therefore the "skipping" log output.

ugrepel:
And - thanks to you too!

PR #872, opened 2019-03-15
Use postgres advisory lock as changelog lock service

berardino:
This commit introduces a new lock service that specifically
targets Postgres >= 9.1 and uses transaction advisory lock
(pg_try_advisory_xact_lock). This fixes (at least for pg)
the issue of DATABASECHANGELOGLOCK getting stuck in lock
when Liquibase does not exit cleanly.

Transaction advisory locks have the nice property that don't need
to be explicitly released. In fact, they are automatically released
by pg when the transaction is completed or the connection closed.

### DIFF
--- a/liquibase-core/src/main/java/liquibase/lockservice/PostgresAdvisoryLockService.java
+++ b/liquibase-core/src/main/java/liquibase/lockservice/PostgresAdvisoryLockService.java
@@ -1,0 +1,173 @@
+package liquibase.lockservice;
+
+import liquibase.configuration.GlobalConfiguration;
+import liquibase.configuration.LiquibaseConfiguration;
+import liquibase.database.Database;
+import liquibase.database.core.PostgresDatabase;
+import liquibase.exception.DatabaseException;
+import liquibase.exception.LockException;
+import liquibase.executor.Executor;
+import liquibase.executor.ExecutorService;
+import liquibase.servicelocator.PrioritizedService;
+import liquibase.statement.core.RawSqlStatement;
+
+import java.time.Duration;
+import java.time.Instant;
+import java.util.logging.Logger;
+
+public class PostgresAdvisoryLockService implements LockService {
+
+    private static final Logger LOGGER = Logger.getLogger(PostgresAdvisoryLockService.class.getName());
+
+    private static final int LOCK_NAMESPACE_ID = "liquibase".hashCode();
+    private static final int LOCK_ID = "lockservice".hashCode();
+
+    static final RawSqlStatement TRY_ACQUIRE_LOCK_SQL =
+        new RawSqlStatement(String.format("select pg_try_advisory_lock(%d,%d)", LOCK_NAMESPACE_ID, LOCK_ID));
+
+    static final RawSqlStatement RELEASE_LOCK_SQL =
+        new RawSqlStatement(String.format("select pg_advisory_unlock(%d,%d)", LOCK_NAMESPACE_ID, LOCK_ID));
+
+    private Database database;
+    private boolean hasChangeLogLock;
+    private long changeLogLockWaitTime;
+    private long changeLogLocRecheckTime;
+
+    public PostgresAdvisoryLockService() {
+        this.changeLogLockWaitTime = LiquibaseConfiguration.getInstance()
+            .getConfiguration(GlobalConfiguration.class)
+            .getDatabaseChangeLogLockWaitTime();
+        this.changeLogLocRecheckTime = LiquibaseConfiguration.getInstance()
+            .getConfiguration(GlobalConfiguration.class)
+            .getDatabaseChangeLogLockPollRate();
+    }
+
+    boolean tryAcquireLock(Executor executor) throws LockException {
+        try {
+            final Boolean tryLock = executor.queryForObject(TRY_ACQUIRE_LOCK_SQL, Boolean.class);
+            if (tryLock) {
+                LOGGER.info("Successfully acquired change log lock");
+            }
+            return tryLock;
+        } catch (DatabaseException e) {
+            throw new LockException(e);
+        }
+    }
+
+    @Override
+    public boolean supports(final Database database) {
+        final boolean isPostgres = database instanceof PostgresDatabase;
+        if (!isPostgres) {
+            return false;
+        }
+        final Boolean useDbLock = LiquibaseConfiguration.getInstance()
+            .getConfiguration(GlobalConfiguration.class)
+            .getUseDbLock();
+        try {
+            // Only works with Postgres version >= 9.1
+            return useDbLock && isAtLeastPostgres91(database);
+        } catch (DatabaseException e) {
+            return false;
+        }
+    }
+
+    boolean isAtLeastPostgres91(final Database database) throws DatabaseException {
+        return (database.getDatabaseMajorVersion() > 9) ||
+            (database.getDatabaseMajorVersion() == 9 && database.getDatabaseMinorVersion() >= 1);
+    }
+
+    @Override
+    public void setDatabase(final Database database) {
+        this.database = database;
+    }
+
+    @Override
+    public void setChangeLogLockWaitTime(final long changeLogLockWaitTime) {
+        this.changeLogLockWaitTime = changeLogLockWaitTime;
+    }
+
+    public long getChangeLogLockWaitTime() {
+        return changeLogLockWaitTime;
+    }
+
+    @Override
+    public void setChangeLogLockRecheckTime(final long changeLogLocRecheckTime) {
+        this.changeLogLocRecheckTime = changeLogLocRecheckTime;
+    }
+
+    public long getChangeLogLocRecheckTime() {
+        return changeLogLocRecheckTime;
+    }
+
+    @Override
+    public boolean hasChangeLogLock() {
+        return hasChangeLogLock;
+    }
+
+    @Override
+    public void waitForLock() throws LockException {
+        boolean locked = false;
+        long timeToGiveUp = Instant.now().plus(Duration.ofMinutes(changeLogLockWaitTime)).toEpochMilli();
+        while (!locked && System.currentTimeMillis() < timeToGiveUp) {
+            locked = acquireLock();
+            if (!locked) {
+                LOGGER.info("Waiting for changelog lock....");
+                try {
+                    Thread.sleep(Duration.ofSeconds(changeLogLocRecheckTime).toMillis());
+                } catch (InterruptedException e) {
+                }
+            }
+        }
+
+        if (!locked) {
+            throw new LockException("Could not acquire change log lock.");
+        }
+    }
+
+    @Override
+    public boolean acquireLock() throws LockException {
+        return hasChangeLogLock = tryAcquireLock(getExecutor());
+    }
+
+    Executor getExecutor() {
+        return ExecutorService.getInstance().getExecutor(database);
+    }
+
+    @Override
+    public void releaseLock() throws LockException {
+        try {
+            getExecutor().queryForObject(RELEASE_LOCK_SQL, Boolean.class);
+        } catch (DatabaseException e) {
+            throw new LockException(e);
+        }
+    }
+
+    @Override
+    public DatabaseChangeLogLock[] listLocks() {
+        return new DatabaseChangeLogLock[0];
+    }
+
+    @Override
+    public void forceReleaseLock() {
+    }
+
+    @Override
+    public void reset() {
+    }
+
+    @Override
+    public void init() {
+    }
+
+    @Override
+    public void destroy() {
+    }
+
+    /**
+     * Prioritize this implementation over {@link StandardLockService}.
+     */
+    @Override
+    public int getPriority() {
+        return PrioritizedService.PRIORITY_DEFAULT + 1;
+    }
+}
--- a/liquibase-core/src/main/java/liquibase/configuration/GlobalConfiguration.java
+++ b/liquibase-core/src/main/java/liquibase/configuration/GlobalConfiguration.java
@@ -21,6 +21,7 @@ public class GlobalConfiguration extends AbstractConfigurationContainer {
     public static final String DIFF_COLUMN_ORDER = "diffColumnOrder";
     public static final String ALWAYS_OVERRIDE_STORED_LOGIC_SCHEMA = "alwaysOverrideStoredLogicSchema";
     public static final String GENERATED_CHANGESET_IDS_INCLUDE_DESCRIPTION = "generatedChangeSetIdsContainsDescription";
+    public static final String USE_DB_LOCK = "useDbLock";
 
     public GlobalConfiguration() {
         super("liquibase");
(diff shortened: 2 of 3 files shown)


liquibase-core/src/main/java/liquibase/lockservice/PostgresAdvisoryLockService.java
```diff
@@ -0,0 +1,165 @@
...
+
+import java.time.Duration;
+import java.time.Instant;
+import java.util.logging.Logger;
+
+public class PostgresAdvisoryLockService implements LockService {
+
+    private static final Logger LOGGER = Logger.getLogger(PostgresAdvisoryLockService.class.getName());
+
+    private static final int LOCK_NAMESPACE_ID = "liquibase".hashCode();
```
gerbrand: Nice idea to use just a hashCode of the string

liquibase-core/src/main/java/liquibase/lockservice/PostgresAdvisoryLockService.java
```diff
@@ -0,0 +1,165 @@
...
+    private static final int LOCK_ID = "lockservice".hashCode();
+
+    static final RawSqlStatement TRY_ACQUIRE_LOCK_SQL =
+        new RawSqlStatement(String.format("select pg_try_advisory_xact_lock(%d,%d)", LOCK_NAMESPACE_ID, LOCK_ID));
```
gerbrand: Liquebase creates a lock per changeset, and a changeset could consist multiple changes.
Afaik since each change is run in it's own transaction, that would mean the lock would be released after the first change.

gerbrand (commented):
Like the change, but I wonder if autoreleasing the lock wouldn't cause problems if there are multiple changes in one changeset (might try this out myself)..

liquibase-core/src/main/java/liquibase/lockservice/PostgresAdvisoryLockService.java
berardino: thank you @gerbrand , I was not aware of one-tx per change. I had assumed that all changeset would be run in the same transaction. 

Can you point me to the entry point where all changeSet are applied? So that I can acquire one lock for all migrations, eg:

berardino: If I read this correctly, Liquibase seems to be acquiring one lock per ChangeLog https://github.com/liquibase/liquibase/blob/master/liquibase-core/src/main/java/liquibase/Liquibase.java#L941 but I am not 100% sure on the tx scope because I don't see that being explicitly created anywhere.

RichardBradley: > Liquibase seems to be acquiring one lock per ChangeLog 

It acquires the global lock once for the changeset, which may contain multiple Changelogs.
The method you linked to, `StandardLockService.waitForLock()` will acquire the lock when first called, and return true if called a second or subsequent time.

RichardBradley:
This would be a great optimisation. We have recurring problems with Liquibase global locks being orphaned if a VM is killed while holding the lock. If the lock were a db-native lock, instead of a row in CHANGELOGLOCK, then the lock would be released in that case.

I was thinking of trying to do the same thing on MySQL, using something like "SELECT FOR UPDATE" or "LOCK TABLES". The transaction-per-changeset problem that you have found here might scupper that idea as well though.

liquibase-core/src/main/java/liquibase/lockservice/PostgresAdvisoryLockService.java
berardino: uhm, so I can probably change this implementation to use advisory session lock instead of tx locks. The lock will be bounded to the connection and it is automatically released when the connection is closed (so in the event of the VM crashing for instance).
What do you think ?

RichardBradley: Sounds good to me, yes.

RichardBradley:
> Transaction advisory locks have the nice property that don't need
> to be explicitly released. In fact, they are automatically released
> by pg when the transaction is completed or the connection closed.

If a machine holding this lock is suddenly killed (or disconnects from the network), will its psql database connection be promptly closed server-side, or might it take a long time to "time out"? If the latter, what settings control this timeout and what are their defaults?

berardino:
@RichardBradley all advisory locks are automatically released when the connection is closed.

You can easily test it by opening two psql shells :

On the first one you acquire the lock :
```sql
select pg_advisory_lock(10);

select * from pg_locks where objid = 10;
```

On the second session you can check that the lock still exists:
```sql
select * from pg_locks where objid = 10;
```

Now exit the first session and re-run `select * from pg_locks where objid = 10;` on the second session the lock is gone lol.

RichardBradley:
Thanks, but that's not what I was asking about.
What will happen in the case where the client connection is not closed, but dropped?
TCP connections where the client has disconnected will remain "half open" indefinitely, unless there is application-level staleness checking.
Do you know whether psql will close the connection in that case with a timeout, and after how long and according to which setting(s)?

This is a key use-case for this enhancement IMO, because orphaned Liquibase locks due to servers being killed is a big problem for Liquibase at present.

I might do some tests on this later if I get time.

berardino:
I re-ran the same test as before but this time I `kill -9` the psql session holding the lock the result was the same, the lock is immediately released.  Let me know if you run additional tests, I should have the changes ready by eod.

RichardBradley:
I found the same when killing a process or "terminating" a VM on AWS: the connection is closed immediately. I believe that the O/S or VM host is explicitly closing the connection in that case.

This seemed too good to be true to me, so I found a second physical machine and retried the experiment. I found that if a client that is holding a `pg_advisory_lock` is disconnected from the network (by unplugging the network cable), then the connection will remain open at the server end (and the lock held by the orphaned connection) for at least 5 minutes.

I have not been able to establish what settings control this timeout. I suspect that the O/S TCP keepalive settings on the server will play some part, and there may also be some settings in `psql` that affect this.

This isn't a deal breaker for the approach proposed here, but it does need documenting in the new lock service and it would be great to be able to document which settings affect this and how to change them.

berardino:
Good point! I have a couple of raspberry pi at home :) I will try to plug the cable and see what happens. See https://www.postgresql.org/docs/9.6/runtime-config-client.html#GUC-IDLE-IN-TRANSACTION-SESSION-TIMEOUT

```


    Terminate any session with an open transaction that has been idle for longer than the specified duration in milliseconds. **This allows any locks held by that session to be released** and the connection slot to be reused; it also allows tuples visible only to this transaction to be vacuumed. See Section 24.1 for more details about this.

    The default value of 0 disables this feature.
```

liquibase-core/src/main/java/liquibase/lockservice/PostgresAdvisoryLockService.java
```diff
@@ -0,0 +1,173 @@
...
+    private static final Logger LOGGER = Logger.getLogger(PostgresAdvisoryLockService.class.getName());
+
+    private static final int LOCK_NAMESPACE_ID = "liquibase".hashCode();
+    private static final int LOCK_ID = "lockservice".hashCode();
+
+    static final RawSqlStatement TRY_ACQUIRE_LOCK_SQL =
+        new RawSqlStatement(String.format("select pg_try_advisory_lock(%d,%d)", LOCK_NAMESPACE_ID, LOCK_ID));
+
+    static final RawSqlStatement RELEASE_LOCK_SQL =
+        new RawSqlStatement(String.format("select pg_advisory_unlock(%d,%d)", LOCK_NAMESPACE_ID, LOCK_ID));
```
berardino: @RichardBradley I have updated the PR to use session advisory lock. 

I will run some more tests to confirm how session locks are released in case of an half-closed connection and document the behavior. 

Notice also that I have introduced a configuration flag `USE_DB_LOCK` that by default keep this lock strategy disabled so that if you want to use advisory locks instead of the standard lock you should explicitly enabled it.

RichardBradley: >  I have introduced a configuration flag USE_DB_LOCK that by default keep this lock strategy disabled

I think this should be on by default, as it is strictly better than the current locking system, if it is tested and working. It might be more likely to get merged if it is off by default, I suppose.

sic2:
This seems like a great PR! 
Do you know if there is a plan to merge it to master?
cheers

jeremykrueger:
This MR would be resolving lots of problems consisting of locks.. Why it wouldn't be merged in the past?

timja:
@SteveDonie any thoughts on this? this causes quite a lot of pain, is there anything we can do to help?

liquibase-core/src/main/java/liquibase/lockservice/PostgresAdvisoryLockService.java
```diff
@@ -0,0 +1,173 @@
...
+import java.time.Duration;
+import java.time.Instant;
+import java.util.logging.Logger;
+
+public class PostgresAdvisoryLockService implements LockService {
+
+    private static final Logger LOGGER = Logger.getLogger(PostgresAdvisoryLockService.class.getName());
+
+    private static final int LOCK_NAMESPACE_ID = "liquibase".hashCode();
+    private static final int LOCK_ID = "lockservice".hashCode();
```
stanio: Shouldn't the lock ID (or namespace ID) be specific to the schema used, or possibly be made configurable?  Wouldn't this clash if there are multiple schemas supporting different application environments/stages in the same DB instance?

stanio: For anyone interested, here's my take on this:
https://gist.github.com/stanio/b7843cc18a80b42812ae128aba212245
(I find the Liquibase' own Executor/SqlStatement overly complicated for the task at hand.)

stanio: > _"Beware of bugs in the above code; I have only proved it correct, not tried it." ⁠—Donald Knuth_

verkhoro:
Hey! Any progress on this PR?

dbahatSAP:
Joining in on the question. 
@nvoxland @SteveDonie what are your thoughts about resurrecting this pull request?

liquibase-core/src/main/java/liquibase/lockservice/PostgresAdvisoryLockService.java
metacubed: There's definitely a use case for multiple Liquibase-managed schemas in the same database.

@berardino, you might be able to use the `oid` of each `DATABASECHANGELOG` table for the advisory lock, instead of a constant hash.

oridool:
Hi, 
Looks great.
I took a slight different approach, and used the database PID as the identifier to be stored in the LOCKEDBY column of lock table.
See my code here:
https://github.com/oridool/liquibase-locking

Please review my solution.

In any case, any solution that becomes part of master Liquibase and resolve this issue automatically is a blessed one :)

ro-rah:
Hi @berardino , sorry for the long wait.  Liquibase has been streamlining their community outreach efforts and for my part as Liquibase Developer Evangelist I have been focusing on making PRs easier to submit and moving them along for evaluation by Eng and QA.

We will take a look.  It is an interesting PR, but want to make sure design-wise the issues brought up here and the proposed changes in this PR are consistent with our roadmap.

Sizing this a RiskLarge:
- Mainly due to Eng time to sort through
- Thanks for updating tests!

Here's the stock PR submission response that contains some helpful info:
Thanks for your pull request!

Here’s what happens next:

liquibase-core/src/main/java/liquibase/lockservice/PostgresAdvisoryLockService.java
stanio: > For anyone interested, here's my take on this:
> https://gist.github.com/stanio/b7843cc18a80b42812ae128aba212245
> (I find the Liquibase' own Executor/SqlStatement overly complicated for the task at hand.)

I have now set up a simple project: [liquibase-sessionlock](/stanio/liquibase-sessionlock).

RichardBradley: That looks great, @stanio, I'll definitely be trying that on my projects. Thanks!

nvoxland:
As part of merging this in, I'll also take a look at https://github.com/oridool/liquibase-locking and https://gist.github.com/stanio/b7843cc18a80b42812ae128aba212245 and other linked ideas

dbahatSAP:
@nvoxland Since the change is only in the LockService, perhaps it make more sense to release this as an official extension instead of a part of liquibase-core? (like https://github.com/liquibase/liquibase-nochangeloglock)

RichardBradley:
> Since the change is only in the LockService, perhaps it make more sense to release this as an official extension instead of a part of liquibase-core

Personally, I think it’s quite important to get this fix into core liquibase and to turn it on by default (and create equivalent fixes for other rdbms). There are already some user-contributed plugins listed in this thread, so an official plugin wouldn’t really advance things much.

I have been using liquibase in commercial projects for 10+ years. It’s a really important tool and in my opinion is pretty much complete and working, except for the perennial problem of orphaned locks. If you look on the advice sites for what people are talking about re liquibase it’s pretty much all people trading tips on how to clear an orphaned lock by deleting the CHANGELOG tables. (e.g. this is the top result on stackoverflow for liquibase: https://stackoverflow.com/questions/15528795/liquibase-lock-reasons )

I understand that there is a pressure to avoid risky changes, but the current lock approach is broken, because servers can and do die in production while they are holding the lock.

Steven-M-Brown:
Any update on this? We encounter this issue pretty often in our k8s cluster where pods have a tendency to get reaped and moved as other higher QoS workloads scale to the point where we have set up specific alerting around it.

capricube:
@berardino if you allow my user to push changes to your PR branch i'll push the resolved conflict. I also rebased your change onto the latest liquibase master

liquibase-core/src/main/java/liquibase/lockservice/PostgresAdvisoryLockService.java
stanio: FWIW, @blagerweij has made readily available [public packages](https://search.maven.org/artifact/com.github.blagerweij/liquibase-sessionlock), and the project could be found at:

* https://github.com/blagerweij/liquibase-sessionlock

filipelautert:
If someone wants to update this PR I can get it reviewed and merged now.

RichardBradley:
The https://github.com/blagerweij/liquibase-sessionlock plugin implements this feature on pqsl and 3 other databases and is in widespread use. It might be better & safer to merge in that battle-tested and updated code, rather than revive this now quite old PR.

elmodeer:
do we have any updates on this change? I think this problem is causing a lot of pain. 
I can update the PR if you are willing to merge it afterwards @filipelautert . also, do you consider just using the code from the plugin https://github.com/blagerweij/liquibase-sessionlock  if this offers an easier way to implement this.

filipelautert:
@elmodeer looks interesting! And yes, if we can use the code from session-lock this would be awesome! I would just start with postgresql in one PR instead of bringing everything from there in a single PR.

elmodeer:
@filipelautert thanks for checking. Do you want me to prepare this PR, or would you take over here?

elmodeer:
@filipelautert I know this might not be on your priority list, but I would really appreciate an answer for my last question 🙏🏻

filipelautert:
> @filipelautert I know this might not be on your priority list, but I would really appreciate an answer for my last question 🙏🏻

@elmodeer sorry, I answered it in my mind and forgot to write it here.
Could you open the PR? Thanks!

PR #877, opened 2019-04-15
ignoreLines command in formatted sql log

li-on:
Added ability to ignore some lines in formatted SQL. 
Now it's posible to leave commands not supported by liquibase but required by other software - like 
SQLPLus.

Example:
```
--liquibase formatted sql logicalFilePath:st
--changeset test:1
CREATE OR REPLACE PACKAGE test as
END	test;
--ignoreLines:1
/
--changeset test:2
CREATE OR REPLACE PACKAGE BODY test as
END	test;
--ignoreLines:1
/
```


nvoxland:
I like the idea, but worry that a count like that will be easy to accidentally break. Would it be better to have
something like `--ignore:start` and `--ignore-end` blocks?

li-on:
Reason for count was: less typing = more free time :D
But start-end may be better.
I've added support for syntax:
--ignoreLines:start
--ignoreLines:end

Feel free to modify/remove code as you like.

nvoxland:
Both options are always nicer! Thanks

PR #878, opened 2019-04-22
ChangeLogPropertyDefinedPrecondition support addded in formatted sql

shubham4194:
Added support for the ChangeLogPropertyDefinedPrecondition in formatted sql format
**Format:**
--precondition-defined-property property:abc value:xyz

### DIFF
--- a/liquibase-core/src/main/java/liquibase/parser/core/formattedsql/FormattedSqlChangeLogParser.java
+++ b/liquibase-core/src/main/java/liquibase/parser/core/formattedsql/FormattedSqlChangeLogParser.java
@@ -1,5 +1,12 @@
 package liquibase.parser.core.formattedsql;
 
+import java.io.BufferedReader;
+import java.io.File;
+import java.io.IOException;
+import java.io.InputStream;
+import java.util.regex.Matcher;
+import java.util.regex.Pattern;
+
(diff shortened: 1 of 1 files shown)


shubham4194:
@nvoxland  Please review this. Earlier only SQLCheck precondition was supported in formatted sql. Now I have added ChangeLogPropertyDefinedPrecondition support

nvoxland:
It would probably be good to look at a more generic precondition format to avoid having to keep adding precondition regular expressions and to support whatever preconditions may be available through plug-ins.

Something like:
-- precondition changeLogPropertyDefined property:abc value:xyz

ro-rah:
Hi @shubham4194 ,

Thanks for the PR.  Would you consider adding tests to the PR so I could size it lower and get it into Conditioning for the upcoming Beta release?

Here's some guidance on testing, till then I will size this large (RiskLarge) since it needs tests and possible design considerations by Eng.

- Liquibase tests ([unit ](https://liquibase.jira.com/wiki/spaces/LB/pages/1274937609/How+to+Write+Liquibase+Core+Unit+Tests)and/or [integration](https://liquibase.jira.com/wiki/spaces/LB/pages/1276608569/How+to+Write+Liquibase+Core+Integration+Tests)) 

Ronak

PR #879, opened 2019-05-14
Java-side date generation

kazetsukaimiko:
This contains a new feature allowing one to override the existing currentDateTimeFunction (now renamed currentDateTimePlaceholder) that uses a SQL function call, with a Java-generated static date strings. Some SQL datastores (Azure Datawarehouse) do not support functions on INSERT. This makes using Liquibase with these datastores impossible for trivial reasons. 

See NamedDateTimeFormatter. This is an enum of named common instances for date time formats that might be used. When one cannot be found, it falls back on DateTimeFormatter.parse() for the input argument. When using a named formatter, it encapsulates the resulting date string with single quotes (may need to be configurable). There's an accompanying command line option to specify the formatter.

We started building this functionality off the 3.6.x branch, as it looked more stable. This PR is against master, for 3.7.0-SNAPSHOT. master as it is, does not currently seem to be able to load changeset xml files, be it from the classpath or the local filesystem. There are two issues present:

* The classloader configured for CommandLineResourceAccessor by Main.configureClassLoader() seems incorrect. I may be missing configuration and/or other settings/changes needed. FileSystemResourceAccessors' rootPaths is always empty regardless of my testing without changing the classloader.
* InputStreams for our chngeset xml files are getting closed prematurely (prior to read) because their parent zip file(s) get closed in the try-with-resources block. I worked around this by changing InputStreamList to take Supplier<InputStream> instead, but really you should use Java 8 Stream<InputStream> and Stream.onClose().

I've "fixed" these bugs in this branch by making ad-hoc changes that ignore the intent of the current developers. I'm looking forward to finding the real fixes to these issues, one endorsed by liquibase development. Unfortunately, I am limited in my knowledge and understanding of the conventions at play and may have missed some of this intent in the process.

### DIFF
--- a/liquibase-cdi/pom.xml
+++ b/liquibase-cdi/pom.xml
@@ -3,7 +3,7 @@
 	<parent>
         <groupId>org.liquibase</groupId>
         <artifactId>liquibase-parent</artifactId>
-        <version>3.7.0-SNAPSHOT</version>
+        <version>3.7.0-localdates-SNAPSHOT</version>
 	</parent>
 	<artifactId>liquibase-cdi</artifactId>
 	<name>Liquibase CDI</name>
--- a/liquibase-core/pom.xml
+++ b/liquibase-core/pom.xml
@@ -5,7 +5,7 @@
 	<parent>
 		<groupId>org.liquibase</groupId>
 		<artifactId>liquibase-parent</artifactId>
-		<version>3.7.0-SNAPSHOT</version>
+		<version>3.7.0-localdates-SNAPSHOT</version>
 	</parent>
 
 	<artifactId>liquibase-core</artifactId>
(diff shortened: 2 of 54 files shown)


ro-rah:
Hi @kazetsukaimiko !

Thanks for your pull request!

Here’s what happens next:

ro-rah:
Hi again @kazetsukaimiko, 
There appears to be some conflicts to resolve, would you mind doing that so I can move the PR into Conditioning?  Thanks for including tests!

ro-rah:
Made risk high, looks like it needs time to resolve conflicts and description states it is a work around fix.

nvoxland:
I'm going to close this PR for now. The overall fix of creating the time from the client is something we actively try to avoid because the client machines running liquibase are often wrong in their time configuration. That's why we lean on the database to be the stable/correct time supplier. It also breaks functionality like "update-sql" giving SQL that will be correclty inserting the current time whenever it's ran vs. when it is generated.

If there are databases like Azure Datawarehouse which doesn't have a "current time "function, we will have to address that as we add official support in for that database somehow.

PR #880, opened 2019-05-19
CORE-3124 Indexes with DESC sorting are not created for PostgreSQL

ianmcderp:
Fixed bug where descending sort order of column is not added to generated statement in CreateIndexGeneratorPostgres.


nvoxland:
Thanks! I'll cherry pick it into the 3.6.x branch as well

PR #881, opened 2019-05-20
Core 3063 resourceaccessor fixes

RichardBradley:
This fixes a few bugs that were causing `MySQLIntegrationTest` to fail, I think caused by recent work to `ResourceAccessor`.

See my comments on https://liquibase.jira.com/browse/CORE-3063 -- what is the current plan for the integration tests?

There is still one failing test: [testAbsolutePathChangeLog](https://github.com/liquibase/liquibase/blob/658e17bb5b415cf975e63277ddc84bc8e6ed6cb2/liquibase-integration-tests/src/test/java/liquibase/dbtest/AbstractIntegrationTest.java#L882) which checks that Liquibase will accept an absolute file path to a changelog. After recent changes to `ResourceAccessor`, this now fails as `new FileSystemResourceAccessor().openStream(x)` will not resolve an absolute filename `x`.


nvoxland:
Thanks!

PR #884, opened 2019-05-29
Recursive resources lookup does not work in springboot environment

andreybpanfilov:
SpringResourceOpener#list method ignores recursive parameter, whist double asterisk is required for recursive search due to the nature of ant path matchers.


nvoxland:
Thanks!

PR #883, opened 2019-05-29
Print message for liquibase tag and tagExists command to system.out

eugenekainara:
I've found that after upgrade from Liquibase 3.5.5 to 3.6.3 - there is no message when databased _liquibase tag <someTag>_ executes and also there is no message when try to execute _liquibase tagExists <someTag>_

e.g. for liquibase 3.5.5 
`java -cp ./:/somedir/postgresql-42.2.5.jar:/somedir/liquibase-core-3.5.5.jar liquibase.integration.commandline.Main  --changeLogFile=liquibase/master.xml --url jdbc:postgresql://192.168.1.1:5432/postgres --username postgres --password password tagExists test`
The **tag** test does not exist in postgres@jdbc:postgresql://192.168.1.1:5432/postgres
Liquibase 'tagExists' Successful
for liquibase 3.6.3
`java -cp ./:/somedir/postgresql-42.2.5.jar:/somedir/liquibase-core-3.6.3.jar:/somedir/logback-core-1.2.3.jar:/somedir/slf4j-api-1.7.26.jar:/somedir/logback-classic-1.2.3.jar liquibase.integration.commandline.Main  --changeLogFile=liquibase/master.xml --url jdbc:postgresql://192.168.1.1:5432/postgres --username postgres --password password tagExists test`
Starting Liquibase at Wed, 29 May 2019 13:10:19 EEST (version 3.6.3 built at 2019-01-29 11:34:48)
Liquibase command 'tagExists' was executed successfully.

PR #890, opened 2019-06-18
Print message for liquibase tag and liquibase tagExists to system.out

eugenekainara:
Same cli issue fix as https://github.com/liquibase/liquibase/pull/883 but for 3.7.0-SNAPSHOT version.


nvoxland:
Thanks, but I'm going to do a merge from 3.6 to master and the change will come in along with that. Don't want to have to have conflicts

PR #891, opened 2019-06-18
CORE-2641

savinov:
Correction of the patch:
ShouldRunChangeSetFilter should initialize recent changesets instead of latest.

commit de0a05fff4e45edb004eddff4ab00f96050e6160
Author: Nathan Voxland <nathan@voxland.net>
Date:   Tue Mar 15 15:31:36 2016 -0500
CORE-2641 runOnChange change set runs every time even if there wasn't changed
Changed implementation to use dateExecuted before order since that is the main sort order


savinov:
Please consider this pull-request as a correction of https://github.com/liquibase/liquibase/pull/529

PR #894, opened 2019-06-21
Using proxy user for Oracle Connections

lifaguiar:
Support for Oracle Proxy User through jdbc url sintax:
--url=jdbc:oracle:thin:PROXY_USER/@host:port/servicename


nvoxland:
Thanks!

PR #897, opened 2019-07-02
fix issue of creating empty changesets

SpiReCZ:
This proposed fix should prevent creation of empty changesets.


nvoxland:
Thanks!

PR #900, opened 2019-07-30
Allow false on Disallowed Values

HelenS52:
Liquibase validate changesets before they are applied to a DBMS.
For example, following changeset for PostgreSQL is rejected because
PostgreSQL does not support ordered sequences:

```
<changeSet id="1" author="author">
   <createSequence cycle="false"
                   cacheSize="20"
                   incrementBy="50"
                   sequenceName="SOME_SEQUENCE"
                   ordered="true"
                   startValue="1"/>
</changeSet>
```

However, if someone writes the following changeset it will be
rejected by liquibase as well:

```
<changeSet id="1" author="author">
   <createSequence cycle="false"
                   cacheSize="20"
                   incrementBy="50"
                   sequenceName="SOME_SEQUENCE"
                   ordered="false"
                   startValue="1"/>
</changeSet>
```

This commit improves the validation in order to write ordered=false
explicitly.

By the way, liquibase documentation claims to support ordered for postgresql: https://www.liquibase.org/documentation/changes/create_sequence.html

┆Issue is synchronized with this [Jira Bug](https://datical.atlassian.net/browse/LB-65) by [Unito](https://www.unito.io/learn-more)
┆Fix Versions: Community 4.x,Community 4.4.0,Liquibase 4.4.0

### DIFF
--- a/liquibase-core/src/main/java/liquibase/exception/ValidationErrors.java
+++ b/liquibase-core/src/main/java/liquibase/exception/ValidationErrors.java
@@ -57,7 +57,7 @@ public class ValidationErrors {
             }
         }
 
-        if (isDisallowed && (value != null)) {
+        if (isDisallowed && (value != null && value instanceof Boolean && ((Boolean) value).booleanValue())) {
             addError(disallowedFieldName + " is not allowed on "+(database == null?"unknown":database.getShortName()));
         }
     }


ro-rah:
Hi @HelenS52 ,

So sorry for the long wait in getting to this PR.  We are a few weeks into our new process for accepting PRs and very much thank you for your time and patience.

Also, thanks for the PR!  Here's the canned stuff below that contains some useful info:
Thanks for your pull request!

Here’s what happens next:

ro-rah:
I believe this could be sized to RiskLow but will leave at a Med if we still need:

@HelenS52 if you are so inclined to write the test(s) we would welcome your help :)

-Ronak

nvoxland (changes requested):
This is in that code that handles the value checking for everything in general. The goal of the existing code is "if the field is not allowed, but any value is set then fail the validation".

So the proposed fix breaks all the existing validation because it would now only care about validating fields are booleans. I'm also not sure that a generic assumption that `x="false"` should pass validation when x is not an allowed field for the database because there are likely times where "false" isn't the same as "not set".

I understand the problem that we'd like to have a setting like ordered="false" for on a changeset that is used by both oracle and postgresql and have that be used by oracle and ignored by postgresql. 

The implementation for that is going to have to be more complex. 

Maybe taking into account default values vs just "is not set"? Although the defaults can be different depending on the databases. 

Perhaps need to rethink how the "disallowed" is defined in general in favor of allowed values that get specified for all the databases?

PR #901, opened 2019-07-31
Microsoft SQL Server doesn't support ORDER in sequences

danielthegray:
The CREATE SEQUENCE command of Microsoft SQL Server doesn't support the `ORDER` clause (see at https://docs.microsoft.com/en-us/sql/t-sql/statements/create-sequence-transact-sql?view=sql-server-2017).

Therefore, it should be ignored for this DB as well. I have tested this change on an SQL Server 2017 (I had an error with the CREATE SEQUENCE command before, and after applying this and installing to my local Maven repo, the problem was gone).

### DIFF
--- a/liquibase-standard/src/main/java/liquibase/sqlgenerator/core/CreateSequenceGenerator.java
+++ b/liquibase-standard/src/main/java/liquibase/sqlgenerator/core/CreateSequenceGenerator.java
@@ -112,14 +112,15 @@ public class CreateSequenceGenerator extends AbstractSqlGenerator<CreateSequence
             }
         }
 
-        if (!(database instanceof MariaDBDatabase) && statement.getOrdered() != null) {
-            if (!(database instanceof SybaseASADatabase)) {
-                if (statement.getOrdered()) {
-                    queryStringBuilder.append(" ORDER");
-                } else {
-                   if (database instanceof OracleDatabase) {
-                       queryStringBuilder.append(" NOORDER");
-                   }
+        boolean databaseSupportsOrderedSequences = !(database instanceof MariaDBDatabase
+                        || database instanceof SybaseASADatabase
+                        || database instanceof MSSQLDatabase);
+        if (databaseSupportsOrderedSequences && statement.getOrdered() != null) {
+            if (statement.getOrdered()) {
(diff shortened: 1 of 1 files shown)


schrieveslaach:
Somewhere liqiubase calls `ValidationErrors#checkDisallowedField` for PostgreSQL and statest that `ORDER` is not supported. Is it required to call this for MSSQL as well?

danielthegray:
It would depend on the focus you want to have, I guess. Should it ignore the parameter that it doesn't understand, or should it fail explicitly if there is one? If PostgreSQL fails/checks explicitly, why do MySQL/MariaDB and Sysbase get to ignore it? (there might be a valid reason, I'm just asking 😃)

I'm guessing the implications of having an ordered sequence on Oracle become a non-ordered sequence on another DB will be up to each app developer, but I feel that the decision to either fail loudly or ignore the `ORDER` parameter is an architectural decision that shouldn't be up to me to make.

schrieveslaach:
I think, we need some guidance through @nvoxland what is the correct approach.

ro-rah:
Hello @danielthegray , I will send this through conditioning.  Size at a Med.
- Contains no related tests so sizing some QA time
- increase MSSQL compatibility indicates mitigating risk since it is already not working.

nvoxland:
Yes, there are conflicting examples of what to do within liquibase. 

We are starting a usability push, and coming up with a standard way of handling things like this will be a big part of that. When do we try to be smart about guessing what the user wants? 

You say you want it ordered, but the database doesn't support that so do we figure you only care for databases that do support it? Or fail validation saying we can't do what you want.

I'll move this ticket into our bucket to talk about with regard to that overall discussion before merging it in.

danielthegray:
Thank you! For what it's worth, from my point of view, it should only take it into account if it is a supported option. One potential use case of Liquibase, as I see it, is to be able to apply the changelog to different DBMS engines without needing to change it or maintain a different one for each DBMS.

filipelautert (approved):

tati-qalified:
This is working as expected for MSSQL, PostgreSQL, H2 and MySQL. I think it's fair to say that it will work in general.
CC @rberezen

rberezen (approved):

PR #902, opened 2019-08-01
CORE-2947 Make Load*DataChange aware of ChangeLogParameters

lindback:
Fix for: https://liquibase.jira.com/browse/CORE-2947

Enable dynamic substitution of parameters in CSV files if ChangeLog property or JVM parameter enableCsvParameterSubstitution is set to true, if not explicitly set to true no substitution is done for backwards compatibility.

I've also added some unit tests in LoadDataChangeTest to test behaviour both with and without enableCsvParameterSubstitution set.

Not sure using a property is the best way to enable this feature, but wasn't sure how to propagate a flag from command line down to code in question otherwise.

No new test failures have been observed due to this fix.

### DIFF
--- a/liquibase-core/src/main/java/liquibase/change/core/LoadDataChange.java
+++ b/liquibase-core/src/main/java/liquibase/change/core/LoadDataChange.java
@@ -295,6 +295,14 @@ public class LoadDataChange extends AbstractChange implements ChangeWithColumns<
 
             boolean isCommentingEnabled = StringUtil.isNotEmpty(commentLineStartsWith);
 
+            boolean isParameterSubstitutionEnabled = false;
+            if (getChangeSet() != null && getChangeSet().getChangeLogParameters() != null) {
+                Object value = getChangeSet().getChangeLogParameters().getValue("enableCsvParameterSubstitution", getChangeSet().getChangeLog());
+                if (value != null && "true".equalsIgnoreCase(value.toString())) {
+                    isParameterSubstitutionEnabled = true;
+                }
+            }
+
             List<SqlStatement> statements = new ArrayList<>();
             while ((line = reader.readNext()) != null) {
                 lineNumber++;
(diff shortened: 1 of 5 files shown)


AlexMattern:
I read through the pull request since I would also like to see this feature. Change Log Parameter replacement is something that could be turned on and off at two levels. The global level and the change log level.

At the change log level:
1. update documentation https://www.liquibase.org/documentation/changes/load_update_data.html
2. update XSD
3. Add variable, getter/setter to liquibase.change.core.LoadDataChange
It looks like the liquibase.change.AbstractChange has a method for populating properties so hopefully you would not need your pull request code lines 298-395.
4. You already have the if statement on line 335, but you could move it to liquibase.changelog.ChangeLogParameters expandExpressions() line 203. That way all change logs would get this new on/off feature and not just csv handling.

At the global level:
1. update documentation https://www.liquibase.org/documentation/command_line.html
2. add variable to liquibase.configuration.GlobalConfiguration
3. In liquibase.changelog.ChangeLogParameters ExpressionExpander() add a line like this.replaceChangeLogParameters = LiquibaseConfiguration.getInstance().getConfiguration(GlobalConfiguration.class).getReplaceChangeLogParameters();

I hope the analysis is clear and it helps.

lindback:
I also would like to see this feature, obviously, but since I got no reaction from the developers, not even a comment (now it has soon been 9 month since I created the CR) I don't really want to spend any more time on this...

ro-rah:
Hi @lindback ,

Sorry for the long delay.  We have been working on our new PR process and only within the last couple of weeks have we rolled this out, I am taking a look at this now to get it into the Conditioning state.

Here's the canned response that contains some info on our updated processes (I noticed you included tests so thanks for that too!):

Thanks for your pull request!

Here’s what happens next:

ro-rah:
Risk Med/High:
- includes tests
- looks reasonable but not sure how the proposed change lines up with our roadmap

famod:
@ro-rah 
> looks reasonable but not sure how the proposed change lines up with our roadmap

Has this assessment changed since April?

filipelautert:
Hi @lindback, thanks for this contribution and for including thorough tests — that's always appreciated.

After reviewing this in the context of our current codebase, we're going to close this PR:

- The codebase has changed significantly since 2019 (module restructuring, method refactoring), so this would need a complete rewrite to apply cleanly.
- The approach of using a magic changelog parameter (`enableCsvParameterSubstitution`) to control the behavior was flagged in the comments as suboptimal — a proper `loadData` attribute or global configuration setting would be a better fit.
- The CLA was never signed.

The feature idea itself — parameter substitution in CSV data files — is valid and something we'd consider if someone wanted to take a fresh crack at it with a changeset-level attribute approach.

Thanks again for the effort, and sorry it took this long to get a definitive answer.

PR #905, opened 2019-08-16
Changes for liquibase 3.8.x, including liquibase pro enhancements

SteveDonie:


maksymgendin:
> This can lead to NPE. Null chekcer should be implemented

Fixed in https://github.com/liquibase/liquibase/pull/913

PR #906, opened 2019-08-19
CORE-3464 Add Oracle Foreign Key defs. to dependency graph

danielthegray:
Since a dependency graph already exists, I simply added mappings for the
foreign key relationships on the DB. I also renamed an Oracle-specific
method for more clarity.

The goal of this PR is to ensure that the generated changelog for an Oracle database has the table definitions in the right order (sometimes it would create a table with a foreign key pointing to a table that hadn't been created/defined yet).

### DIFF
--- a/liquibase-core/src/main/java/liquibase/diff/output/changelog/DiffToChangeLog.java
+++ b/liquibase-core/src/main/java/liquibase/diff/output/changelog/DiffToChangeLog.java
@@ -377,7 +377,7 @@ public class DiffToChangeLog {
         return new ArrayList<>(objects);
     }
 
-    private List<Map<String, ?>> queryForDependencies(Executor executor, List<String> schemas)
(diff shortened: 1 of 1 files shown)


danielthegray:
I might also add that the FOREIGN KEY dependencies are already added to the dependency graph for MS SQL Server (so I think it makes perfect sense to add it for Oracle as well!)

danielthegray:
@SteveDonie is there a chance for this to make it into the upcoming 3.8.x release? I saw in a comment on another PR that you guys would be releasing 3.8.1 soon?

If it doesn't fit into you're schedule, should I just point the PR at the master branch instead?

danielthegray:
I can see that the new changes on branch 3.8.x make this PR obsolete (since the re-ordering of the Oracle database objects has been re-written). Therefore, I will now close this PR.

SteveDonie:
Thanks Daniel - I appreciate the effort!

PR #911, opened 2019-08-28
fix CORE-3379: escaping ? for postgresql

AlexSchuetz:
Replacing the postgres-operators '?', '?&' and '?|' with escaped operators '??', '??&' and '??|' if not already escaped.

If further databases use a '?' in sql-dialect I can add some more escaping rules.


mueihn:
Simple replacing all single question marks will be too much.

Consider the example of a query in which a question marks appears in a string literal:
select * from table where slogan = 'What should I say?'

For this case the query transformed as above would return rows where the slogan is 'What should I say??', which is not intended.

So, definitely, question marks with string literals should be untouched. Maybe, there are more situations in which a question mark is meant as literal text question mark instead of an operator?

AlexSchuetz:
I refactored my regex to not match these cases. 😄

AlexSchuetz:
moved this to the patch-release branch 3.8.x #912

PR #914, opened 2019-09-04
add British pound sign to list of unicode letters

ermeaney:
so, for reasons not 100% known to me, we use the British pound sign £ as a place holder for a comma in postgresql hstore. This caused our migrations to fail with the following exception:

`Exception in thread "main" liquibase.util.grammar.TokenMgrError: Lexical error at line 130, column 115.  Encountered: "\u00a3" (163), after : ""`

This PR adds that character to the list of unicode letters in the parser

### DIFF
--- a/liquibase-core/src/main/javacc/liquibase/util/grammar/SimpleSqlGrammar.jj
+++ b/liquibase-core/src/main/javacc/liquibase/util/grammar/SimpleSqlGrammar.jj
@@ -80,6 +80,7 @@ TOKEN:
 Built list from http://stackoverflow.com/a/37668315/45756
 */
 | < #UNICODE_LETTERS: ["\u00AA",
+    "\u00A3",
     "\u00B5",
     "\u00BA",
     "\u00C0"-"\u00D6",
--- a/liquibase-core/src/main/javacc/liquibase/util/grammar/SimpleSqlGrammar.jj
+++ b/liquibase-core/src/main/javacc/liquibase/util/grammar/SimpleSqlGrammar.jj
@@ -69,7 +69,7 @@ TOKEN : /* Numeric Constants */
 TOKEN:
 {
     < COMPLEX_IDENTIFIER: (<S_IDENTIFIER> | <S_QUOTED_IDENTIFIER>) ((["\r","\n"," "])* "." (["\r","\n"," "])* (<S_IDENTIFIER> | <S_QUOTED_IDENTIFIER>))+ >
-|	< S_IDENTIFIER: ( <LETTER> | <UNICODE_LETTERS> )+ ( <DIGIT> | <LETTER> | <UNICODE_LETTERS> | <SPECIAL_CHARS> )* >
+|	< S_IDENTIFIER: ( <LETTER> | <UNICODE_LETTERS> | <UNICODE_CURRENCIES> )+ ( <DIGIT> | <LETTER> | <UNICODE_CURRENCIES> | <UNICODE_LETTERS> | <SPECIAL_CHARS> )* >
 | 	< #LETTER: ["a"-"z", "A"-"Z", "_", "$"] >
 |   < #SPECIAL_CHARS: "$" | "_" | "#" | "@" >
 |   < S_CHAR_LITERAL: "'" (~["'"]|"\\'")* "'" ("'" (~["'"]|"\\'")* "'")*>
--- a/liquibase-core/src/main/javacc/liquibase/util/grammar/SimpleSqlGrammar.jj
+++ b/liquibase-core/src/main/javacc/liquibase/util/grammar/SimpleSqlGrammar.jj
@@ -75,12 +75,31 @@ TOKEN:
 |   < S_CHAR_LITERAL: "'" (~["'"]|"\\'")* "'" ("'" (~["'"]|"\\'")* "'")*>
 |   < S_QUOTED_IDENTIFIER: "\"" (~["\n","\r","\""])+ "\"" | ("`" (~["\n","\r","`"])+ "`") | ( "[" ~["0"-"9","]"] (~["\n","\r","]"])* "]" ) >
 |    <EMPTY_QUOTE: "\"" "\"">
+/*
+built from http://unicode.org/charts/PDF/U20A0.pdf
+*/
+| < #UNICODE_CURRENCIES: ["\u0024",
+    "\u00A2"-"\u00A5",
+    "\u058F",
+    "\u060B",
+    "\u09F2"-"\u09F3",
+    "\u0AF1",
+    "\u0BF9",
+    "\u0E3F",
+    "\u17DB",
+    "\u2133",
+    "\u5143",
+    "\u5186",
+    "\u5706",
+    "\u5173",
+    "\uFDFC",
+    "\u20A0"-"\u20BF"
+    ]>
 
 /*
 Built list from http://stackoverflow.com/a/37668315/45756
 */
 | < #UNICODE_LETTERS: ["\u00AA",
-    "\u00A3",
     "\u00B5",
     "\u00BA",
     "\u00C0"-"\u00D6",
--- a/liquibase-core/src/test/groovy/liquibase/util/SqlParserTest.groovy
+++ b/liquibase-core/src/test/groovy/liquibase/util/SqlParserTest.groovy
@@ -52,6 +52,7 @@ class SqlParserTest extends Specification {
     @Unroll
     def "parse with unicode"() {
         expect:
+        def p_out = SqlParser.parse(input)
         SqlParser.parse(input).toArray(true) == output
 
         where:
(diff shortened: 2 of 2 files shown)


RichardBradley:
This is the wrong fix for your problem.
"£" is not a letter, but is in the category "Symbol, Currency": https://www.fileformat.info/info/unicode/char/00a3/index.htm

If you have a liquibase migration which is failing when you are using a "£" symbol where Liquibase thinks it is invalid, but it is in fact valid, the correct fix would be to update the incorrect validation rules, not to break Liquibase's unicode lookup tables.

It might also help if you could add a test case demonstrating a valid changeset which fails before this fix and succeeds after it.

ermeaney:
Fair enough, first here is a changeset that would break it:
`<databaseChangeLog
		xsi:schemaLocation="http://www.liquibase.org/xml/ns/dbchangelog http://www.liquibase.org/xml/ns/dbchangelog/dbchangelog-3.0.xsd">
    <changeSet author="me" id="create-test-tables">
        <sql>
            CREATE TABLE "mock_names"
            (
                contact_id int,
                first_name character varying(100),
                last_name character varying(100),
                custom_fields hstore,
                PRIMARY KEY(contact_id)
            );
		</sql>
        <sql>
            INSERT INTO mock_names (contact_id, first_name, last_name, custom_fields) VALUES (1, 'John', 'Doe', '"Region"=>"US£££ West", "Job Title"=>"", "Cost Center"=>""');
        </sql>    
    </changeSet>
</databaseChangeLog>`

And the reason that I went the direction I did was https://liquibase.jira.com/browse/CORE-2843 has the same error and updated the lexer code. See https://github.com/liquibase/liquibase/commit/b61e7fb75180cd7e260c1b73509d1187ab8f23e2

And this PR for sure made my problem go away.

So, I looked all over for the validation rules that might be causing the problem, and have not found anything. So that and the fact that the exact error I was was solving was fixed by adding to the file I did, /shrug?

That said, I took to heart what you were saying about the "£" is not a letter but a currency, and have updated the code on my side to have something more like:

`| < #UNICODE_CURRENCIES: ["\u0024",
    "\u00A2"-"\u00A5",
    "\u058F",
    "\u060B",
    "\u09F2"-"\u09F3",
    "\u0AF1",
    "\u0BF9",
    "\u0E3F",
    "\u17DB",
    "\u2133",
    "\u5143",
    "\u5186",
    "\u5706",
    "\u5173",
    "\uFDFC",
    "\u20A0"-"\u20BF"
    ]>
`

But I have not tested locally yet and have not pushed it until I get a chance to validate it

Now, if you still feel like I am totally going the wrong direction, if you could point me in the right direction I would be happy to try a different approach. The first pass at this was me between a rock and hard place where the team need a feature in updated versions, and it was broke for me.

RichardBradley:
> CORE-2843 has the same error and updated the lexer code. See b61e7fb

That commit correctly updates the definition of "`S_CHAR_LITERAL`" (i.e. a string literal) to include escaped quotes. I think that commit is correct.

> I am going to have to find what is actually a valid change set that is breaking it, since I am not sure what is doing it

I take it the changset you posted then removed is accepted by Liquibase? Looking at the grammar, any char is allowed inside a quoted string, so "£" should be fine there, as you have seen.

Perhaps you are using "£" in a SQL identifier? It looks like `S_IDENTIFIER` would disallow that before your change but allow it afterwards? If so, I think the correct fix would be to add either "£" alone or all of `UNICODE_CURRENCIES` to the `S_IDENTIFIER` expression? (Along with a citation from some relevant source about why that is correct. )

I have had a quick skim of the psql hstore docs, and I don't see anything that looks like it would cause trouble here, so an example would be very helpful

ermeaney:
you are correct, what I had and removed was fine, and I spent a good part of today tracking it down. Sorry it took me a while, getting a database in a place where I can run them and then figuring out where it broke was a pain, since I don't manage the change sets and migrations.

Anyway, here is where I got to. The problem gets triggered via a changeset that in including a slq file via something like this:

`<databaseChangeLog
        xsi:schemaLocation="http://www.liquibase.org/xml/ns/dbchangelog http://www.liquibase.org/xml/ns/dbchangelog/dbchangelog-3.0.xsd">
  <changeSet id="test" author="me">
  <sqlFile path="file_of_sql.sql" splitStatements="false" relativeToChangelogFile="true" dbms="postgresql" stripComments="false"/>
  </changeSet>
</databaseChangeLog>`

Which inside that file is a HUGE function that is getting updated or replaced in the database. I am not going to post the whole function, but I tracked it down to this part of the function:

`    , reimbursement_hist_var_rate_v1( --> remember to add \\ when escaping
		X.unit
		, (('['||(REPLACE(REPLACE(REPLACE(REPLACE((X.adj_descr_list)::text, '\\',''),'"[','['),']"',']'),'£££',','))||']')::json)->0 --> need to cast/fix this JSON
	) 
	AS variable_rate`

The has a little context, but really this:
`		, (('['||(REPLACE(REPLACE(REPLACE(REPLACE((X.adj_descr_list)::text, '\\',''),'"[','['),']"',']'),'£££',','))||']')::json)->0 --> need to cast/fix this JSON`

Now I 100% know this is where we ran into trouble, because I changed it to:
`		, (('['||(REPLACE(REPLACE(REPLACE(REPLACE((X.adj_descr_list)::text, '\\',''),'"[','['),']"',']'),'ttt',','))||']')::json)->0 --> need to cast/fix this JSON
`

And it ran fine then.

Also, and shame on me for not putting this in the initial PR, but here is the whole stack trace from when I triggered this:

`Exception in thread "main" liquibase.util.grammar.TokenMgrError: Lexical error at line 130, column 115.  Encountered: "\u00a3" (163), after : ""
	at liquibase.util.grammar.SimpleSqlGrammarTokenManager.getNextToken(SimpleSqlGrammarTokenManager.java:1513)
	at liquibase.util.grammar.SimpleSqlGrammar.getNextToken(SimpleSqlGrammar.java:109)
	at liquibase.util.SqlParser.parse(SqlParser.java:40)
	at liquibase.util.StringUtils.processMutliLineSQL(StringUtils.java:55)
	at liquibase.change.AbstractSQLChange.generateStatements(AbstractSQLChange.java:225)
	at liquibase.database.AbstractJdbcDatabase.executeStatements(AbstractJdbcDatabase.java:1253)
	at liquibase.changelog.ChangeSet.execute(ChangeSet.java:609)
	at liquibase.changelog.visitor.UpdateVisitor.visit(UpdateVisitor.java:53)
	at liquibase.changelog.ChangeLogIterator.run(ChangeLogIterator.java:83)
	at liquibase.Liquibase.update(Liquibase.java:202)
	at liquibase.Liquibase.update(Liquibase.java:274)
	at liquibase.Liquibase.update(Liquibase.java:251)
	at liquibase.integration.commandline.Main.doMigration(Main.java:1257)
	at liquibase.integration.commandline.Main.run(Main.java:209)
	at liquibase.integration.commandline.Main.main(Main.java:132)`

I can't really drop the whole function here, there is to much information about us that I would rather not have out there, but if you need more information I can try to help.

P.S. Sorry I didn't have all of this at the start, I get you burn time without enough information.

ermeaney:
I asked one of our data guys to write a function that you can reproduce with, I will post it once I get it

RichardBradley:
Interesting. It looks to me very much like that `£` is inside a quoted string, which should get parsed fine by Liquibase, as you have seen above. Perhaps you have triggered some other bug which has put the lexer into the wrong state?

ermeaney:
@RichardBradley I updated the PR with the changes I mentioned, plus added some unit tests

ermeaney:
[testChangeset.tar.gz](https://github.com/liquibase/liquibase/files/3602836/testChangeset.tar.gz)
attached is a change set which should trigger this bug

ro-rah:
Hi @ermeaney, thanks for the PR and sorry for the long wait in response!
@RichardBradley thanks for collaborating.

I will put this in Conditioning, and RiskMed (it has tests, thanks, sized it up from a Low for the time it takes internal Eng. to weigh in on how this should be accomplished).

Here's the canned response for PR submission, it contains helpful guidance:
Thanks for your pull request!

Here’s what happens next:

RichardBradley:
The change doesn't look right to me.

1. I think the change to the grammar looks wrong

Your test case SQL is as follows:

```
    final_data:=((SELECT (REPLACE(REPLACE(REPLACE(REPLACE((final_data), '\\',''),'"[','['),']"',']'),'£££',','))||']')); --> need to cast/fix this JSON
```

The problem "£" symbol is inside a quoted string here.
However, you are changing the parser to allow "`S_IDENTIFIER`" to start with or contain "£".

I wonder if the bug here is that something else about the expression is confusing the parser and it ends up in state `S_IDENTIFIER` when it gets to the quoted string, but it should be in the quoted string state? Maybe the sql parser is confused by the `'\\'` string? My IDE seems to think this is OK sql though and colour-codes it correctly.


2. Your test case is parsing a string that doesn't look like valid SQL

The input to the test looks like:
```
   ", (('['||(REPLACE(REPLACE(REPLACE(REPLACE((X.adj_descr_list)::text, '\\\\',''),'\"[','['),']\"',']'),'£££',','))||']')::json)->0 --> need to cast/fix this JSON"
```
... which you are passing to "`SqlParser.parse`"

I wouldn't expect valid SQL to start with ",". Perhaps the test case should look more like the changelog you attached earlier?

kataggart:
Closing this due to aging. Please re-open or re-submit if it pops up again and interferes with your work. Thanks.

PR #916, opened 2019-09-06
Add missing @Override annotations

InfiniteLoop90:
Added missing Override annotations so that if/when methods in superclasses change
there will be a compile-time error if subclasses haven't been changed accordingly.

### DIFF
--- a/liquibase-cdi/src/main/java/liquibase/integration/cdi/SchemesCDIConfigBuilder.java
+++ b/liquibase-cdi/src/main/java/liquibase/integration/cdi/SchemesCDIConfigBuilder.java
@@ -69,6 +69,7 @@ public class SchemesCDIConfigBuilder {
         final InputStream is = SchemesCDIConfigBuilder.class.getResourceAsStream(SCHEMA_NAME);
         try {
             return jvmLocked(id, new Callable<CDILiquibaseConfig>() {
+                @Override
                 public CDILiquibaseConfig call() throws Exception {
                     return createCDILiquibaseConfig(id, is);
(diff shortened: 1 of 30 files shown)


ro-rah:
Hi @InfiniteLoop90 ,

Thanks for your pull request!

Here’s what happens next:

ro-rah:
Hi @InfiniteLoop90,

Would you help me come up with how to test this change?  I could pass that on to QA, and it would speed up the acceptance of the PR.

Thanks,

Ronak

ro-rah:
I've sized it currently for RiskHigh until further guidance on how to test.  Changes sound reasonable but wide spread and need to confirm if we want to do it this way.

InfiniteLoop90:
Hi @ro-rah,
The `@Override` annotation is kind of unique in that it as a Retention Policy of `SOURCE`, meaning that the annotation is discarded by the compiler. That also means that the annotation has no impact on the resulting bytecode, so I don't think there's much of anything that can be tested specifically other than verifying that the build happens successfully.

For context, here's the definition of the Override annotation in Java
```
@Target(ElementType.METHOD)
@Retention(RetentionPolicy.SOURCE)
public @interface Override {
}
```

Let me know if you have other questions.

Thanks

ro-rah:
Thanks @InfiniteLoop90 !  I reduced the risk to Medium down from Hard.  I believe the internal team has enough to evaluate.

filipelautert:
Most of those changes have been implemented under PR #3558 . Thanks anyway for the contribution!

PR #919, opened 2019-09-23
Fixed issued with bundling in shaded jars

isaki:
The implementation in 3.8.0 that searches for build information in the MANIFEST is not compatible with shaded builds as not all shaded builds have the required information. Additionally, there is an unclosed input stream in the current implementation.

The issue manifests as an NPE when `buildInfoValue.equals("UNKNOWN")` after this value was assigned `null` after reading MANIFEST.MF.

This PR addresses these two issues in the following way:

- Wrapped manifest input read with a try-with-resources block.
- Changed default build info to be null and change all checks to be against null; these are null safe.
- Added final null check and returned a default value ("UNKNOWN", set now as a constant).

┆Issue is synchronized with this [Jira Story](https://datical.atlassian.net/browse/LB-51) by [Unito](https://www.unito.io/learn-more)
┆Fix Versions: Liquibase 3.10.2

### DIFF
--- a/liquibase-core/src/main/java/liquibase/util/LiquibaseUtil.java
+++ b/liquibase-core/src/main/java/liquibase/util/LiquibaseUtil.java
@@ -1,7 +1,6 @@
 package liquibase.util;
 
 import liquibase.exception.UnexpectedLiquibaseException;
-import liquibase.integration.commandline.CommandLineUtils;
 
 import java.io.IOException;
 import java.io.InputStream;
@@ -13,8 +12,10 @@ import java.util.jar.Manifest;
 
 public class LiquibaseUtil {
 
+    private static final String DEFAULT_BUILD_INFO_VALUE = "UNKNOWN";
+
     public static String getBuildVersion() {
-        return getBuildInfo("build.version","Bundle-Version");
+        return getBuildInfo("build.version", "Bundle-Version");
     }
 
     public static String getBuildTime() {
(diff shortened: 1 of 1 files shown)


Hronom:
Devs please merge this and release

mpe85:
@nvoxland Is it possible to have this in an upcoming 3.8.1 patch release?

SteveDonie:
This will be fixed in 3.8.1

ro-rah:
Hi @isaki ,

Looks like there is a merge conflict here, would you please resolve and push to PR?

Please also consider adding tests.

Tests can move this to the next state (Conditioning) easier if you include tests ([unit ](https://liquibase.jira.com/wiki/spaces/LB/pages/1274937609/How+to+Write+Liquibase+Core+Unit+Tests)and/or [integration](https://liquibase.jira.com/wiki/spaces/LB/pages/1276608569/How+to+Write+Liquibase+Core+Integration+Tests)).  Would you be willing to add tests as well?

Thanks,

Ronak

isaki:
I thought this was fixed in 3.8.1 using a different approach. Has this problem resurfaced?

ro-rah:
@isaki that definitely could be the case, I can check and close if that is the case.

ro-rah:
Putting at risklow for the time it takes to confirm this is no longer an issue and close out.

ro-rah:
Closing per @isaki's comments.

PR #922, opened 2019-09-30
Updated to properly handle symbolic links and paths with spaces.

nwertzberger:
Because spaces are converted to '%20' in uri's, paths that should have matched as relative paths were failing, and instead being loaded as absolute paths.

This change also makes the default path pull from a canonical path, which makes sure that things are resolved correctly when running from a directory with symlinks, and that "thefolder/../thefile" resolves to the same thing as "thefile".

┆Issue is synchronized with this [Jira Story](https://datical.atlassian.net/browse/LB-48) by [Unito](https://www.unito.io/learn-more)

### DIFF
--- a/liquibase-core/src/main/java/liquibase/resource/AbstractResourceAccessor.java
+++ b/liquibase-core/src/main/java/liquibase/resource/AbstractResourceAccessor.java
@@ -87,17 +87,28 @@ public abstract class AbstractResourceAccessor implements ResourceAccessor {
         for (File file : files) {
             if (file.isDirectory()) {
                 if (includeDirectories) {
-                    returnSet.add(convertToPath(file.getAbsolutePath()));
+                    returnSet.add(convertToPath(extractPath(file)));
                 }
                 if (recursive) {
                     getContents(file, recursive, includeFiles, includeDirectories, basePath, returnSet);
                 }
             } else {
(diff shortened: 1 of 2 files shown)


nwertzberger:
This will break people who had their relative paths accidentally load as absolute paths because of this issue. It does require the paths in the databasechangelog to be changed in order for stuff to start working as expected.

nwertzberger:
Any update here? I can’t upgrade without this pull request

nwertzberger:
@nvoxland sorry to prod, just need to know what the plan is for this issue.

ro-rah:
Hi @nwertzberger ,

Sorry for the wait, we are in the midst of our PR process submission streamlining and I am attempting to work through the backlog.

Thank you for including [unit test ](https://liquibase.jira.com/wiki/spaces/LB/pages/1274937609/How+to+Write+Liquibase+Core+Unit+Tests).

Since this has a potential to break existing changelogs but does include unit tests, I am adding Risk Med here.

Ronak

nwertzberger:
Did this make it into 3.10.1?

nvoxland:
This issue has been resolved with the larger ResourceAccessor refactoring we did in 4.0.0. I added a test case for a space in a filename to 4.1.x and the test passes

PR #926, opened 2019-10-05
changing HashMap to LinkedHashMap for deterministic iterations

contextshuffling:
Test `liquibase.sqlgenerator.core.DropColumnGeneratorTest#testDropMultipleColumnsMySQL` has this following test assertion 
```
Assert.assertEquals("[TEST_TABLE.col1, TEST_TABLE.col2, TEST_TABLE, DEFAULT]", String.valueOf(sql[0].getAffectedDatabaseObjects()))
``` 
that asserts the `toString()` value of `affectedDatabaseObjects`. 

`affectedDatabaseObjects` is a HashSet and the `toString()` function depends on the iteration order of the hash set. However, `HashSet` does not guarantee any specific order of entries. Therefore, the assertion will fail if the order is different.

This PR proposes to use a LinkedHashSet for a deterministic order (insertion order of elements) and modifies the corresponding test assertions. 

Please let me know if you want to discuss the changes more.

### DIFF
--- a/liquibase-core/src/test/java/liquibase/sqlgenerator/core/AddColumnGeneratorTest.java
+++ b/liquibase-core/src/test/java/liquibase/sqlgenerator/core/AddColumnGeneratorTest.java
@@ -11,6 +11,10 @@ import liquibase.statement.PrimaryKeyConstraint;
 import liquibase.statement.core.AddColumnStatement;
 import org.junit.Test;
 
+import java.util.Arrays;
+import java.util.List;
+import java.util.stream.Collectors;
+
 import static org.junit.Assert.*;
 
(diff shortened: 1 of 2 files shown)


contextshuffling:
@ianmcderp After a more thorough search, I realize there is one more test that are affected by this change and I have modifies it in the new changes.

Feel free to discuss more if you have questions!

contextshuffling:
@danielthegray I totally agree. I updated the PR to propose using the `containsAll()` method in two ways to do a "Set"-ish equivalence.

contextshuffling:
@ianmcderp any new feedbacks for this PR?

PR #933, opened 2019-11-09
3.5.x

wuttke:
MaxDB Liquibase Diff does not work for sequences as this database (liquibase-maxdb repo) is not included in the SequenceSnapshotGenerator.
(Query should be moved to the Database successor, anyways...)
This is the backport to the 3.5.x branch.


wuttke:
Could this PR be merged please?

nvoxland:
We won't be releasing any more 3.5 patch releases, so not bringing this in

PR #941, opened 2019-11-27
[CORE-3520] Add support for ALWAYS generation type with PostgreSQL auto-increment column.

suilven:
When using Liquibase to create a table with an auto-incrementing integer, the corresponding identity column in PostgreSQL is always created with the constraint `BY DEFAULT`, irrespective of the value set for the generation type.

Add support to Liquibase for the `ALWAYS` generation type when creating an auto-incrementing column in PostgreSQL. This will allow the generation type to be set to `BY DEFAULT` or `ALWAYS`. The default value if the `generationType` attribute is not specified in the column tag, remains `BY DEFAULT`.

┆Issue is synchronized with this [Jira Story](https://datical.atlassian.net/browse/LB-39) by [Unito](https://www.unito.io/learn-more)

### DIFF
--- a/liquibase-core/src/main/java/liquibase/database/core/PostgresDatabase.java
+++ b/liquibase-core/src/main/java/liquibase/database/core/PostgresDatabase.java
@@ -238,6 +238,24 @@ public class PostgresDatabase extends AbstractJdbcDatabase {
     }
 
     @Override
+    protected String getAutoIncrementClause(final String generationType, final Boolean defaultOnNull) {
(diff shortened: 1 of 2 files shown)


ro-rah:
Hi @suilven , another good complete PR requests, thanks again for the [unit test](https://github.com/liquibase/liquibase/pull/944).

ro-rah:
tagged as RiskLow due to @SteveDonie comment and the inclusion of tests.

PR #942, opened 2019-11-28
Modify SetColumnRemarks support for mysql. The original code does not support mysql modification notes

zyzyzy523:
It turns out that the sql statement generated by setColumnRemarks of the mysql database is ALTER TABLE cat.person COMMENT = 'A String';
Now modify it to ALTER TABLE cat.person MODIFY COLUMN id int COMMENT 'A String';

┆Issue is synchronized with this [Jira Bug](https://datical.atlassian.net/browse/LB-38) by [Unito](https://www.unito.io/learn-more)
┆Fix Versions: Community 4.x


SteveDonie:
Thanks for the pull request. Can you please edit it so that all the comments are in English? 
We will need to review this because of the change to the xsd and decide when it can be included.

SteveDonie:
Thanks for the update!

liquibase-core/src/main/resources/www.liquibase.org/xml/ns/dbchangelog/dbchangelog-3.8.xsd
```diff
@@ -920,6 +920,7 @@
             <xsd:attribute name="tableName" type="xsd:string" use="required"/>
             <xsd:attribute name="columnName" type="xsd:string" use="required"/>
             <xsd:attribute name="remarks" type="xsd:string"/>
+            <xsd:attribute name="columnDataType" type="xsd:string"/>
```
nvoxland: Since the 3.8 XSD is released and frozen, this will have to go into a new 3.9 xsd

ro-rah:
Hi @zyzyzy523 ,

Thanks for your PR submission.  Its been awhile since we heard from you.  Did you need any help or further guidance around the changes @nvoxland requested?

Thanks!

Ronak

ro-rah:
@zyzyzy523 I will move this into Conditioning but size it RiskMed.  Eng time will be needed to make the change requested by @nvoxland.  Thanks for including tests and translations!

nvoxland:
What the "modification notes" that the old version did not support?

The other problem we'll have is that the new code requires dataType, which will cause everyone with existing changeSets on mysql that didn't need to specify the dataType before to fail.

We'll need to either use one SQL or the other depending on if dataType is set, or find a way to support the modification notes without requiring dataType.

php-coder:
I've just faced it on 3.8.9

Is there plans to backport the fix to 3.x ?

ro-rah:
@php-coder, we are concentrating on 3.10.1 and 4 currently, and operational patch fixes to the existing branches is being hashed out. Hang tight and we'll update this PR shortly.

liquibase-core/src/main/java/liquibase/statement/core/SetColumnRemarksStatement.java
```diff
@@ -9,13 +9,27 @@
     private String tableName;
     private String columnName;
     private String remarks;
-
-    public SetColumnRemarksStatement(String catalogName, String schemaName, String tableName, String columnName, String remarks) {
+    private String columnDataType;
+    private Object defaultValue;
+
+    public SetColumnRemarksStatement(String catalogName,
```
nvoxland, line 15: We need to keep the existing constructor as well for extension compatibility.

liquibase-core/src/main/resources/www.liquibase.org/xml/ns/dbchangelog/dbchangelog-4.0.xsd
```diff
@@ -924,6 +924,8 @@
             <xsd:attribute name="tableName" type="xsd:string" use="required"/>
             <xsd:attribute name="columnName" type="xsd:string" use="required"/>
             <xsd:attribute name="remarks" type="xsd:string"/>
+            <xsd:attribute name="columnDataType" type="xsd:string" use="required"/>
```
nvoxland, line 927: Because this adds new attributes, we'll have to release this as part of an x.y release, and we will have to move these to the correct version based on when it gets merged.

nvoxland (changes requested):

nvoxland:
I merged this into master after moving the new xsd attributes to 4.3.xsd.

I did take out the new defaultValue setting. I know setRemarks is going to cause that to be lost, but lots of other things get lots too (primary keys, constraints, etc.). It's a general problem with mysql we need to find a better solution to. But the defaultValue attribute is just a part-there answer so at least trying to be consistent in our handling of it.

I also cherry picked the changes to 4.3.x for the 4.3.0 release.

PR #945, opened 2019-12-03
[CORE-3523] Consistently map Liquibase integer types to Oracle integer types

suilven:
All integer types are represented in Oracle as number(precision, scale). The `INTEGER` keyword is converted to a number(*) by Oracle and the `MEDIUMINT` keyword is not supported.

Adopt a consistent mapping from Liquibase to Oracle numbers:

* tinyint => number(3,0)
* smallint => number(5,0)
* mediumint => number(7,0)
* integer => number(10,0)
* bigint => number(19,0)

┆Issue is synchronized with this [Jira Story](https://datical.atlassian.net/browse/LB-36) by [Unito](https://www.unito.io/learn-more)
┆Fix Versions: Liquibase 4.1.0

### DIFF
--- a/liquibase-core/src/main/java/liquibase/datatype/core/BigIntType.java
+++ b/liquibase-core/src/main/java/liquibase/datatype/core/BigIntType.java
@@ -37,7 +37,7 @@ public class BigIntType extends LiquibaseDataType {
             }
         }
         if (database instanceof OracleDatabase) {
-            return new DatabaseDataType("NUMBER", 38,0);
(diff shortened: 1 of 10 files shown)


nvoxland:
We'll have to move this to master, since it's a larger change than we'd want to do in a patch release. I'm going to wait until I get master updated from 3.8.x first, though

ro-rah:
Added Part of a larger change tag since it requires master to be updated.

PR #947, opened 2019-12-04
[CORE-3525] Add support for java.sql.Types.TIMESTAMP_WITH_TIMEZONE.

suilven:
Liquibase does not currently support the `java.sql.Types.TIMESTAMP_WITH_TIMEZONE` type. For example, using `java.sql.Types.TIMESTAMP_WITH_TIMEZONE` as the column type in Liquibase generates a column with the `TIMESTAMP WITHOUT TIME ZONE` type in PostgreSQL.

Add support for the `java.sql.Types.TIMESTAMP_WITH_TIMEZONE` type to Liquibase so the correct SQL is generated for databases supporting timestamps with timezones.

## Repo Steps
1. Create a createTable changeSet with a column having `type="java.sql.Types.TIMESTAMP_WITH_TIMEZONE"` 
1. Run changeSet against postgresql or another database that supports timestamp with timezone types
1. Check that the table column was created as timestamp with timezone information

## QA Manual Test Requirements
* Execute the steps as described in repro and assert the correct dataype is created on the database.

## QA Automated Test Requirements
* Write a Postgres regression test following the repro steps.

┆Issue is synchronized with this [Jira Story](https://datical.atlassian.net/browse/LB-34) by [Unito](https://www.unito.io/learn-more)
┆Fix Versions: Liquibase 3.10.2

### DIFF
--- a/liquibase-core/src/main/java/liquibase/datatype/core/TimestampType.java
+++ b/liquibase-core/src/main/java/liquibase/datatype/core/TimestampType.java
@@ -125,7 +125,7 @@ public class TimestampType extends DateTimeType {
             || database instanceof H2Database
             || database instanceof HsqlDatabase)) {
 
-            if ((database instanceof PostgresDatabase)) {
+            if (database instanceof PostgresDatabase || database instanceof H2Database) {
                 type.addAdditionalInformation("WITH TIME ZONE");
             } else {
                 type.addAdditionalInformation("WITH TIMEZONE");
@@ -143,7 +143,7 @@ public class TimestampType extends DateTimeType {
 
             if (additionalInformation != null) {
                 String additionInformation = additionalInformation.toUpperCase(Locale.US);
-                if ((database instanceof PostgresDatabase) && additionInformation.toUpperCase(Locale.US).contains("TIMEZONE")) {
+                if ((database instanceof PostgresDatabase || database instanceof H2Database) && additionInformation.toUpperCase(Locale.US).contains("TIMEZONE")) {
                     additionalInformation = additionInformation.toUpperCase(Locale.US).replace("TIMEZONE", "TIME ZONE");
                 }
                 // CORE-3229 Oracle 11g doesn't support WITHOUT clause in TIMESTAMP data type
--- a/liquibase-core/src/main/java/liquibase/datatype/core/TimestampType.java
+++ b/liquibase-core/src/main/java/liquibase/datatype/core/TimestampType.java
@@ -16,14 +16,12 @@ import liquibase.logging.LogType;
 import liquibase.util.StringUtils;
 import liquibase.util.grammar.ParseException;
 
-import java.util.Locale;
-
 /**
  * Data type support for TIMESTAMP data types in various DBMS. All DBMS are at least expected to support the
  * year, month, day, hour, minute and second parts. Optionally, fractional seconds and time zone information can be
  * specified as well.
  */
-@DataTypeInfo(name = "timestamp", aliases = {"java.sql.Types.TIMESTAMP", "java.sql.Timestamp", "timestamptz"}, minParameters = 0, maxParameters = 1, priority = LiquibaseDataType.PRIORITY_DEFAULT)
+@DataTypeInfo(name = "timestamp", aliases = {"java.sql.Types.TIMESTAMP", "java.sql.Types.TIMESTAMP_WITH_TIMEZONE", "java.sql.Timestamp", "timestamptz"}, minParameters = 0, maxParameters = 1, priority = LiquibaseDataType.PRIORITY_DEFAULT)
 public class TimestampType extends DateTimeType {
 
     /**
@@ -119,6 +117,21 @@ public class TimestampType extends DateTimeType {
             type =  new DatabaseDataType("TIMESTAMP", fractionalDigits);
         } else {
             type = new DatabaseDataType("TIMESTAMP");
+        }
+
+        if (originalDefinition.startsWith("java.sql.Types.TIMESTAMP_WITH_TIMEZONE")
+            && (database instanceof PostgresDatabase
+            || database instanceof OracleDatabase
+            || database instanceof H2Database
+            || database instanceof HsqlDatabase)) {
+
+            if ((database instanceof PostgresDatabase)) {
+                type.addAdditionalInformation("WITH TIME ZONE");
+            } else {
+                type.addAdditionalInformation("WITH TIMEZONE");
+            }
+
+            return type;
         }
 
         if (getAdditionalInformation() != null
--- a/liquibase-core/src/test/groovy/liquibase/datatype/DataTypeFactoryTest.groovy
+++ b/liquibase-core/src/test/groovy/liquibase/datatype/DataTypeFactoryTest.groovy
@@ -210,8 +210,50 @@ class DataTypeFactoryTest extends Specification {
         "nclob"                                        | new OracleDatabase()   | "NCLOB"                                        | ClobType      | false
         "xml"                                          | new OracleDatabase()   | "XMLTYPE"                                      | XMLType       | false
         "xmltype"                                      | new OracleDatabase()   | "XMLTYPE"                                      | XMLType       | false
+        "timestamp"                                    | new OracleDatabase()   | "TIMESTAMP"                                    | TimestampType | false
+        "timestamp(6)"                                 | new OracleDatabase()   | "TIMESTAMP(6)"                                 | TimestampType | false
+        "TIMESTAMP WITH TIMEZONE"                      | new OracleDatabase()   | "TIMESTAMP WITH TIMEZONE"                      | TimestampType | false
+        "TIMESTAMP(6) WITH TIMEZONE"                   | new OracleDatabase()   | "TIMESTAMP(6) WITH TIMEZONE"                   | TimestampType | false
+        "timestamp without timezone"                   | new OracleDatabase()   | "TIMESTAMP"                                    | TimestampType | false
+        "timestamp(6) without timezone"                | new OracleDatabase()   | "TIMESTAMP(6)"                                 | TimestampType | false
+        "timestamptz"                                  | new OracleDatabase()   | "TIMESTAMP"                                    | TimestampType | false
+        "timestamptz(6)"                               | new OracleDatabase()   | "TIMESTAMP(6)"                                 | TimestampType | false
+        "java.sql.Timestamp"                           | new OracleDatabase()   | "TIMESTAMP"                                    | TimestampType | false
+        "java.sql.Timestamp(6)"                        | new OracleDatabase()   | "TIMESTAMP(6)"                                 | TimestampType | false
+        "java.sql.Types.TIMESTAMP"                     | new OracleDatabase()   | "TIMESTAMP"                                    | TimestampType | false
+        "java.sql.Types.TIMESTAMP(6)"                  | new OracleDatabase()   | "TIMESTAMP(6)"                                 | TimestampType | false
+        "java.sql.Types.TIMESTAMP_WITH_TIMEZONE"       | new OracleDatabase()   | "TIMESTAMP WITH TIMEZONE"                      | TimestampType | false
+        "java.sql.Types.TIMESTAMP_WITH_TIMEZONE(6)"    | new OracleDatabase()   | "TIMESTAMP(6) WITH TIMEZONE"                   | TimestampType | false
         "xml"                                          | new PostgresDatabase() | "XML"                                          | XMLType       | false
+        "timestamp"                                    | new PostgresDatabase() | "TIMESTAMP WITHOUT TIME ZONE"                  | TimestampType | false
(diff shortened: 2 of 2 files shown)


liquibase-core/src/main/java/liquibase/datatype/core/TimestampType.java
```diff
@@ -16,14 +16,12 @@
...
 
-import java.util.Locale;
-
 /**
  * Data type support for TIMESTAMP data types in various DBMS. All DBMS are at least expected to support the
  * year, month, day, hour, minute and second parts. Optionally, fractional seconds and time zone information can be
  * specified as well.
  */
-@DataTypeInfo(name = "timestamp", aliases = {"java.sql.Types.TIMESTAMP", "java.sql.Timestamp", "timestamptz"}, minParameters = 0, maxParameters = 1, priority = LiquibaseDataType.PRIORITY_DEFAULT)
+@DataTypeInfo(name = "timestamp", aliases = {"java.sql.Types.TIMESTAMP", "java.sql.Types.TIMESTAMP_WITH_TIMEZONE", "java.sql.Timestamp", "timestamptz"}, minParameters = 0, maxParameters = 1, priority = LiquibaseDataType.PRIORITY_DEFAULT)
```
katzyn: In JDBC the `java.time.OffsetDateTime` is used for `Types.TIMESTAMP_WITH_TIMEZONE`.

liquibase-core/src/main/java/liquibase/datatype/core/TimestampType.java
```diff
@@ -121,6 +119,21 @@ public DatabaseDataType toDatabaseDataType(Database database) {
...
+        if (originalDefinition.startsWith("java.sql.Types.TIMESTAMP_WITH_TIMEZONE")
+            && (database instanceof PostgresDatabase
+            || database instanceof OracleDatabase
+            || database instanceof H2Database
+            || database instanceof HsqlDatabase)) {
+
+            if ((database instanceof PostgresDatabase)) {
+                type.addAdditionalInformation("WITH TIME ZONE");
+            } else {
+                type.addAdditionalInformation("WITH TIMEZONE");
```
katzyn: What `WITH TIMEZONE` is?

liquibase-core/src/test/groovy/liquibase/datatype/DataTypeFactoryTest.groovy
```diff
@@ -210,8 +210,50 @@ class DataTypeFactoryTest extends Specification {
...
+        "java.sql.Types.TIMESTAMP_WITH_TIMEZONE"       | new PostgresDatabase() | "TIMESTAMP WITH TIME ZONE"                     | TimestampType | false
+        "java.sql.Types.TIMESTAMP_WITH_TIMEZONE(6)"    | new PostgresDatabase() | "TIMESTAMP(6) WITH TIME ZONE"                  | TimestampType | false
         "BINARY(16)"                                   | new H2Database()       | "BINARY(16)"                                   | BlobType      | false
+        "timestamp"                                    | new H2Database()       | "TIMESTAMP"                                    | TimestampType | false
+        "timestamp(6)"                                 | new H2Database()       | "TIMESTAMP(6)"                                 | TimestampType | false
+        "TIMESTAMP WITH TIMEZONE"                      | new H2Database()       | "TIMESTAMP WITH TIMEZONE"                      | TimestampType | false
+        "TIMESTAMP(6) WITH TIMEZONE"                   | new H2Database()       | "TIMESTAMP(6) WITH TIMEZONE"                   | TimestampType | false
```
katzyn: H2 cannot parse this due to missing space between `TIME` and `ZONE`.

liquibase-core/src/test/groovy/liquibase/datatype/DataTypeFactoryTest.groovy
```diff
@@ -210,8 +210,50 @@ class DataTypeFactoryTest extends Specification {
...
+        "timestamptz(6)"                               | new H2Database()       | "TIMESTAMP(6)"                                 | TimestampType | false
+        "java.sql.Timestamp"                           | new H2Database()       | "TIMESTAMP"                                    | TimestampType | false
+        "java.sql.Timestamp(6)"                        | new H2Database()       | "TIMESTAMP(6)"                                 | TimestampType | false
+        "java.sql.Types.TIMESTAMP"                     | new H2Database()       | "TIMESTAMP"                                    | TimestampType | false
+        "java.sql.Types.TIMESTAMP(6)"                  | new H2Database()       | "TIMESTAMP(6)"                                 | TimestampType | false
+        "java.sql.Types.TIMESTAMP_WITH_TIMEZONE"       | new H2Database()       | "TIMESTAMP WITH TIMEZONE"                      | TimestampType | false
+        "java.sql.Types.TIMESTAMP_WITH_TIMEZONE(6)"    | new H2Database()       | "TIMESTAMP(6) WITH TIMEZONE"                   | TimestampType | false
```
katzyn: The same problem here.

Note that H2 supports `TIMESTAMP WITH TIME ZONE` only since 1.4.197, in older versions this data type was experimental only and was not really usable.

liquibase-core/src/main/java/liquibase/datatype/core/TimestampType.java
suilven: Why is that relevant here? The `Timestamp` type in Liquibase represents a TIMESTAMP in SQL, with or without a Timezone. The code in this class is used for creating a database schema, not reading data out of the table. Or am I missing something?

liquibase-core/src/main/java/liquibase/datatype/core/TimestampType.java
suilven: I don't understand this comment. Could you add some extra details please?

liquibase-core/src/test/groovy/liquibase/datatype/DataTypeFactoryTest.groovy
suilven: See reply for next comment.

liquibase-core/src/test/groovy/liquibase/datatype/DataTypeFactoryTest.groovy
suilven: Thanks, I wasn't aware of that change. I've pushed a new commit to correct the generated SQL.

liquibase-core/src/main/java/liquibase/datatype/core/TimestampType.java
katzyn: `TIMESTAMP [ WITHOUT TIME ZONE ]` and `TIMESTAMP WITH TIME ZONE` should not have the same data type, they are different. The only one database that uses the similar internal representation for them is PostgreSQL and it looks like the violation of the SQL Standard by this database.

AFAIK, other databases not directly related with this one really hold the time zone offset component (Oracle also can hold the name of the time zone, that is not supported by the Standard). Legacy, buggy and defective by design `java.sql.Timestamp` is not fully reliable for the plain `TIMESTAMP`, but it is also completely unsuitable for `TIMESTAMP WITH TIME ZONE` for all databases with exception for PostgreSQL, it can only lead to data loss.

I think you need to request a review from somebody from Luquibase community. Such changes should be carefully designed.

liquibase-core/src/main/java/liquibase/datatype/core/TimestampType.java
katzyn: What database accepts `TIMESTAMP WITH TIMEZONE`? PostgreSQL, Oracle, H2, and HSQLDB from your PR use `TIMESTAMP [(fractionalSecondsPrecision)] WITH TIME ZONE` with the space between `TIME` and `ZONE` and this data type is a part of the SQL Standard.
https://www.postgresql.org/docs/12/datatype-datetime.html
https://docs.oracle.com/en/database/oracle/oracle-database/19/sqlrf/Data-Types.html#GUID-BE23545B-469A-4A57-8D13-505F2F5DB706
https://h2database.com/html/datatypes.html#timestamp_with_time_zone_type
http://hsqldb.org/doc/2.0/guide/sqlgeneral-chapt.html#sgc_datetime_types
And there are others, such as DB2:
https://www.ibm.com/support/knowledgecenter/en/SSEPEK_12.0.0/sqlref/src/tpc/db2z_timestampvalues.html

liquibase-core/src/main/java/liquibase/datatype/core/TimestampType.java
katzyn: > The code in this class is used for creating a database schema, not reading data out of the table. Or am I missing something?

If it is not really used anywhere it could be OK, but in that case you also need to check for `TIMESTAMPTZ` from PostgreSQL below
```Java
        if (originalDefinition.startsWith("java.sql.Types.TIMESTAMP_WITH_TIMEZONE")
            && (database instanceof PostgresDatabase
            || database instanceof OracleDatabase
            || database instanceof H2Database
            || database instanceof HsqlDatabase)) {
```

katzyn: Or maybe not and `TIMESTAMPTZ` will be used as is. Such things should be tested somewhere.

liquibase-core/src/main/java/liquibase/datatype/core/TimestampType.java
```diff
@@ -130,7 +143,7 @@ public DatabaseDataType toDatabaseDataType(Database database) {
 
             if (additionalInformation != null) {
                 String additionInformation = additionalInformation.toUpperCase(Locale.US);
-                if ((database instanceof PostgresDatabase) && additionInformation.toUpperCase(Locale.US).contains("TIMEZONE")) {
+                if ((database instanceof PostgresDatabase || database instanceof H2Database) && additionInformation.toUpperCase(Locale.US).contains("TIMEZONE")) {
                     additionalInformation = additionInformation.toUpperCase(Locale.US).replace("TIMEZONE", "TIME ZONE");
                 }
```
katzyn: I don't understand where `TIMEZONE` may came from, but other databases don't support `TIMEZONE` too, not only the PostgreSQL and H2.

katzyn: If it comes from the code above, it should be adjusted instead to produce a correct SQL. But maybe there is some other source of incorrect name? It that case this fix should be unconditional for all databases.

liquibase-core/src/main/java/liquibase/datatype/core/TimestampType.java
suilven: This code is about creating timestamps in a database schema. It has nothing to do with the representation of that timestamp in Java code. If I wanted to store a timestamp with a timezone in a **Java** class, I would use the `java.time.OffsetDateTime` type, exactly as you suggest. But I would map that field to a `TIMESTAMP WITH TIMEZONE` **SQL** type. This is the [approach used in Hibernate](https://docs.jboss.org/hibernate/orm/5.4/userguide/html_single/Hibernate_User_Guide.html#basic-datetime-java8).

The Liquibase docs state that ['..specifying a java.sql.Types.* type will be converted to the correct type..'](https://www.liquibase.org/documentation/column.html). However, `java.sql.Types.TIMESTAMP_WITH_TIMEZONE` does not work correctly, which is the point of this PR. I tried this approach the first time I wanted to create a `TIMESTAMP WITH TIMEZONE` SQL type. From studying the Liquibase code, I now know I could have used the raw SQL type `TIMESTAMP WITH TIMEZONE` instead. I created this PR to avoid others falling into the same trap. If support for this type isn't added, the Liquibase docs should be updated to reflect that position.

nvoxland:
I was going to change the base to master vs. 3.8.x because this is a large enough change that it will need to go in a major release, not a patch release. But, there were a lot of conflicts because I've not gotten 3.8.x merged back to master for a while so I changed the base back for now.

I added the "part of a larger change" label because there is a bigger piece we are looking at around improving data type support and how we can test and manage new datatypes across all our supported databases. We'll want to make sure we're considering that when we merge this in.

ro-rah:
RiskHigh since the change looks big and may have issues to sort through, attempting to get this into 3.10.1

nvoxland:
This change itself is small and low-risk and fits with the current design. @suilven is right that the change is just fixing up our support for specifying `type="java.sql.Types.TIMESTAMP_WITH_TIMEZONE"` like you can specify `type="java.sql.Types.INTEGER"` instead of the sql version of the type. That is a type we should try to handle.

This fix may not be handling all the timezone specifics that all the different databases support, but it's improving what's there anyway and this is just a convenience/alternative to specifying `type="TIMESTAMP WITH TIMEZONE"` or whatever. This change doesn't affect any other ways people can specify their types.

@katzyn is right that the overall design/split between generic timestamp types and timezoned timestamps need to be better modeled in liquibase, and that is something we are planning on addressing. But, no reason to block this PR until that larger problem gets addressed.

nvoxland:
I merged this into a LB-34 branch for easier local testing

PR #948, opened 2019-12-05
[CORE-3526] TABLE is a reserved keyword in H2database since 1.4.197

tholu:
- Added test case to root.changelog.xml, added "TABLE" to keywords in H2 Database, updated h2 to latest version 1.4.200

### DIFF
--- a/liquibase-core/src/main/java/liquibase/database/core/H2Database.java
+++ b/liquibase-core/src/main/java/liquibase/database/core/H2Database.java
@@ -68,6 +68,7 @@ public class H2Database extends AbstractJdbcDatabase {
             "SYSDATE",
             "SYSTIME",
             "SYSTIMESTAMP",
+            "TABLE",
             "TODAY",
             "TRUE",
             "UNION",
--- a/liquibase-core/src/main/java/liquibase/database/core/H2Database.java
+++ b/liquibase-core/src/main/java/liquibase/database/core/H2Database.java
@@ -35,45 +35,87 @@ public class H2Database extends AbstractJdbcDatabase {
     private static String END_CONCAT = ")";
     private static String SEP_CONCAT = ", ";
     private static List keywords = Arrays.asList(
+            "ALL",
+            "AND",
+            "ARRAY",
+            "AS",
+            "BETWEEN",
+            "BOTH",
+            "CASE",
+            "CHECK",
+            "CONSTRAINT",
             "CROSS",
+            "CURRENT_CATALOG",
             "CURRENT_DATE",
+            "CURRENT_SCHEMA",
             "CURRENT_TIME",
             "CURRENT_TIMESTAMP",
+            "CURRENT_USER",
             "DISTINCT",
             "EXCEPT",
             "EXISTS",
             "FALSE",
             "FETCH",
+            "FILTER",
             "FOR",
+            "FOREIGN",
             "FROM",
             "FULL",
             "GROUP",
+            "GROUPS",
             "HAVING",
+            "IF",
+            "ILIKE",
+            "IN",
             "INNER",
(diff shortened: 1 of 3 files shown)


tholu:
Should this rather be merged in to the 3.8.x branch? Feel free to edit or let me know if I should recreate it.

katzyn:
The list of keywords of H2 1.4.200 is a little bit longer:
https://h2database.com/html/advanced.html#keywords

In 1.4.201 `YEAR`, `MONTH`, `DAY`, `HOUR`, `MINUTE`, `SECOND`, and `VALUE` will be keywords too.

Such lists are always outdated. H2 reports the actual keywords, excluding reserved words from SQL:2003 via `DatabaseMetaData.getSQLKeywords()`.

The list of data types is not complete:
https://h2database.com/html/datatypes.html

Standard `LOCALTIME` and `LOCALTIMESTAMP` functions aren't listed, but different weird non-standard compatibility only functions are listed.

`getDateLiteral()` actually produces a `VARCHAR` literal, the `DATE` literal in H2 is the same as in the SQL Standard:
https://h2database.com/html/grammar.html#date

`isSafeToRunUpdate()` has incomplete set of patterns, H2 has many other subsystems that work with regular files. Even the simple `jdbc:h2:./db`, `jdbc:h2:/path/to/db`, etc. aren't recognized.

This list of problems may be incomplete, I'm not familiar with Liquibase.

tholu:
@katzyn Thanks for your input! I'll try to update the Pull Request to include as much fixes for the mentioned problems as possible.

tholu:
I extended the list of reserved keywords and updated the set of patters for `isSafeToRunUpdate()`. I guess that the other issues you mentioned are worth their own issue/pull requests.

SteveDonie:
Thanks for the pull request! Editing this to be able to merge with the 3.8.x branch would make it more likely to get merged into a release sooner. We are working on a roadmap for Liquibase that we will share with everyone soon.

liquibase-core/src/main/java/liquibase/database/core/H2Database.java
```diff
@@ -174,6 +216,8 @@ public boolean isSafeToRunUpdate() throws DatabaseException {
                         || url.startsWith("jdbc:h2:mem:")
                         || url.startsWith("jdbc:h2:zip:")
                         || url.startsWith("jdbc:h2:~")
+                        || url.startsWith("jdbc:h2:./db")
+                        || url.startsWith("jdbc:h2:/")
```
katzyn: `db` should be removed, it's just a sample name of the database in the current working directory (`./`).

`jdbc:h2:C:/path/to/database` with any drive letter is also a valid local URL on Windows (similar to `jdbc:h2:/path/to/database` on other systems).

Actually the whole implementation is not reliable at all. It the intention is to exclude remote urls, the `jdbc:h2:tcp:` and `jdbc:h2:ssl:` should be excluded and all other URLs should be allowed.

katzyn: Something like
```Java
super.isSafeToRunUpdate() // <- I don't know what is this
   || (!url.startsWith("jdbc:h2:tcp:") && (!url.startsWith("jdbc:h2:ssl:");
```

tholu: @katzyn Thanks for your input! I'll include that in the updated pull request against the 3.8.x branch.

tholu:
@SteveDonie Thanks for your input! Just changing the base branch obviously does not work, so I recreated the pull request based on the 3.8.x branch and included the feedback by @katzyn regarding `isSafeToRunUpdate()` in #949.

PR #949, opened 2019-12-06
[CORE-3526] TABLE is a reserved keyword in H2database since 1.4.197

tholu:
- Added test case to root.changelog.xml, added all current keywords in H2 Database, updated H2 to latest version 1.4.200
- Improved isSafeToRunUpdate() for H2

Based on #948.

┆Issue is synchronized with this [Jira Story](https://datical.atlassian.net/browse/LB-33) by [Unito](https://www.unito.io/learn-more)


ro-rah:
Hi @tholu 

Thanks for your pull request!

Here’s what happens next:

Thanks!

Ronak

ro-rah:
closes #948

ro-rah:
Risk Low, contains unit tests and it is a bug fix.

ro-rah:
Slated for Q2

camsaul:
This change actually broke Liquibase + H2 for us, since it now quotes the newly-reserved keywords as lowercase strings.

In older versions of Liquibase 3.6.3 + H2 1.4.197 we had a table with a `values` column; using the `LEGACY` quoting strategy, the column that got created was `VALUES` (uppercased). When upgrading to Liquibase 4.4.3 the column is now created as `values` (lowercased) meaning all of our custom SQL that uses the quoted identifier `"VALUES"` no longer works

nvoxland:
Yeah, it's hard to manage the combination of exactly when reserved words become invalid across database versions and within liquibase. 

Since I'd guess there isn't a lot of cases where you're hitting this, I'd suggest adding a `<modifySql>` section to the problem chagneSets that replace `"table"` with either `table` or `"TABLE"` depending on what direction you are going.

PR #953, opened 2019-12-25
StandardLockService: Ensure waitForLock() invokes acquireLock() at least once

stanio:
Unless this is considered a feature, when `changeLogLockWaitTimeInMinutes=0`, `acquireLock()` doesn't get invoked.

### DIFF
--- a/liquibase-core/src/main/java/liquibase/lockservice/StandardLockService.java
+++ b/liquibase-core/src/main/java/liquibase/lockservice/StandardLockService.java
@@ -207,19 +207,18 @@ public class StandardLockService implements LockService {
     @Override
     public void waitForLock() throws LockException {
 
-        boolean locked = false;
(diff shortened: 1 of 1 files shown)


nvoxland:
Having a way to timeout locks is useful, but since execution of a changelog can take an arbitrarily long time, I worry about hard-coding a time. 

We are looking at ways to improve the lock system in general to make it clean up locks better when possible, so that may go a long way to reducing the need for something like this. 

But, something like this may still be needed as long as it has configurable timeouts.

stanio:
My use case is a micro-service environment where the application lifecycle is controlled by a cluster manager.  If a node fails to start the manager will retry depending on its setup.

And generally, I think it makes the behavior more predictable.  In the current implementation, it exhibits one more quirk in my opinion – it always performs one extra wait cycle after `aquireLock : false` even if the timeout has been reached while trying the `aquireLock` already.

PR #955, opened 2019-12-27
DAT 3464

wwillard7800:


madorb:
I'm curious if it's intentional that there's no description provided with these DAT merges? I assume al the details are stored in an internal issue tracker. 

It seems like it would be useful to provide a publicly facing description for each of these merge requests so that potential community committers have a better understanding of why particular code changes were made?

nvoxland:
Yes, sorry. We are making some internal process changes lately, and changes to the pull request descriptions and issue tracking is things we are still working on. New pull requests we make going forward will be more descriptive and useful.

madorb:
Thanks @nvoxland! Appreciate all the work you & the Datical team do for the liquibase community!

PR #958, opened 2020-01-02
Speed up the Change factory initialization

stanio:
Turns out `Change.createChangeMetaData()` is rather slow – `AbstractChange.createChangeParameterMetadata()` to be more precise.  Creating this metadata for all `Change` types upfront unnecessary wastes time and causes slower feedback, considering not all `Change` types may be actually used.  So just do the minimal work necessary for the initial registration.

  \+ Avoid some copying in `SqlGeneratorFactory.getGenerators()`

### DIFF
--- a/liquibase-core/src/main/java/liquibase/change/ChangeFactory.java
+++ b/liquibase-core/src/main/java/liquibase/change/ChangeFactory.java
@@ -68,15 +68,13 @@ public class ChangeFactory implements SingletonObject {
      */
     public void register(Class<? extends Change> changeClass) {
         try {
-            Change instance = changeClass.newInstance();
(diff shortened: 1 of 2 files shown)


stanio:
I see this bugs some tests which rely `Change.createChangeMetaData()` may return different info than supplied by the `@DatabaseChange` annotation.

stanio:
> I see this bugs some tests which rely `Change.createChangeMetaData()` may return different info than supplied by the `@DatabaseChange` annotation.

I don't see this feature actually used in the core and it is a significant performance hit for use-cases where Liquibase always runs at the start of an application and generally does nothing (no DB updates to apply).

PR #961, opened 2020-01-06
merge DAT-3538 into master

jnewton03:
add jacoco plugin to pom


jnewton03:
based off wrong branch (master) when I should have branched of 3.8.x

PR #967, opened 2020-01-15
Add mac/windows installers

nvoxland:


liquibase-dist/src/main/install4j/liquibase.install4j
```diff
@@ -0,0 +1,510 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<install4j version="8.0.3" transformSequenceNumber="8">
+  <directoryPresets config="../../../target/liquibase-3.8.6-local-SNAPSHOT/liquibase" />
```
SteveDonie: Would be nice if this did not have a hardcoded version number (if possible).

nvoxland: Good noticing. I had that fixed before, but the install4j ide swapped it out. I set it back.

SteveDonie:
I see that there is a new profile for the installer, and it appears that to build the installer you would need to specify that profile. Not sure when this will get addressed, but we will need to update the  build scripts to activate that profile. Documentation for developers should be updated to describe that as well.

PR #968, opened 2020-01-16
Update CreateSequenceGenerator.java

karishmajain:
I executed below changeset(1001) using liquibase 3.6.3 version.
Precondition to check existence of sequence and create sequence was working absolutely fine.

<changeSet author="Karishma" id="1001"
 <preConditions onFail="MARK_RAN"
      <not
        <sequenceExists sequenceName="sec_uid_sq" schemaName="${schema-name}"/
      </not
    </preConditions
      <createSequence sequenceName="sec_uid_sq"/
  </changeSet

I upgraded liquibase to version 3.8.4 and the same changeset started failing with below mentioned error:
Reason: liquibase.exception.DatabaseException: ERROR: syntax error at or near "NOT"
  Position: 21 [Failed SQL: (0) CREATE SEQUENCE  IF NOT EXISTS liquidb.sec_uid_sq]

IF NOT EXISTS condition doesnt work in all version of postgres. 
Why IF NOT EXISTS is appended to sql query in 3.8.* version of liquibase.

Liquibase query should work on old (9.4) as well as new (9.5+)version to avoid BWC break. 
note: IF NOT EXISTS work only for version higher than 9.4 version of postgresql.

Thanks,
Karishma Jain

┆Issue is synchronized with this [Jira Story](https://datical.atlassian.net/browse/LB-27) by [Unito](https://www.unito.io/learn-more)

### DIFF
--- a/liquibase-core/src/main/java/liquibase/sqlgenerator/core/CreateSequenceGenerator.java
+++ b/liquibase-core/src/main/java/liquibase/sqlgenerator/core/CreateSequenceGenerator.java
@@ -47,9 +47,9 @@ public class CreateSequenceGenerator extends AbstractSqlGenerator<CreateSequence
     public Sql[] generateSql(CreateSequenceStatement statement, Database database, SqlGeneratorChain sqlGeneratorChain) {
         StringBuilder queryStringBuilder = new StringBuilder();
         queryStringBuilder.append("CREATE SEQUENCE ");
-        if (database instanceof PostgresDatabase) {
-            queryStringBuilder.append(" IF NOT EXISTS ");
-        }
+       // if (database instanceof PostgresDatabase) {
+       //     queryStringBuilder.append(" IF NOT EXISTS ");
+       // }
         queryStringBuilder.append(database.escapeSequenceName(statement.getCatalogName(), statement.getSchemaName(), statement.getSequenceName()));
         if (database instanceof HsqlDatabase || database instanceof Db2zDatabase) {
             queryStringBuilder.append(" AS BIGINT ");


karishmajain:
I removed as i dont see any need of adding "IF NOT EXISTS" and creating BWC where existing changelog stopped working.
Because of this change, below code which checks if a sequence exists or not ,doesnt make any sense

-      <not>
-        <sequenceExists sequenceName="sec_uid_sq"/>
-      </not>
becuase liquibase is always appending"IF NOT EXISTS" in the sql query of create sequence.
 
Liquibase has not mentioned anywhere why there is a need of adding "IF NOT EXISTS" in 3.8.x versions.

Creating query based on postgres version doesnt seem a good approach.
This issue can be handled without "IF NOT EXISTS" using pre-condition.

ro-rah:
Hi @karishmajain!

Thanks for your submission.  We are in the process of streamlining our PR requests.  Usually low risk changes are merged faster than higher risk.  While you wait, if you are inclined, please also provide a unit test.  One of the items that influence risk level is providing unit tests. 

Below I will post a snippet of the unit test guideline which will be public soon but I thought I would share to real life contributors such as you (thanks!).

-Ronak

ro-rah:
Naming and Location of the Unit Test

Unit tests are written at the Java class level. There is a 1-1 mapping between the unit test class and the Java class being tested. The unit test class has the same package and name as the Java class, but with Test.groovy appended. All test classes are stored using the src/test/groovy structure expected by Maven.

Here is the example for liquibase-core/src/main/java/liquibase/util/StringUtil.java

    Replace main/java in the file path with the new substring src/test/groovy 

        liquibase-core/src/main/java/liquibase/util/

        liquibase-core/src/test/groovy/liquibase/util/

    If a unit test doesn’t already exist, create a new unit test class with the same base filename as the Java class and append Test.groovy

        liquibase-core/src/test/groovy/liquibase/util/StringUtilTest.groovy

    The next step is to create the actual test

nfalco79:
I fall in the issue #1039 I would propose a better solution for that based on PostgreSQL documentation:
```
if (database instanceof PostgresDatabase && database.getDatabaseMajorVersion() <= 9 && database.getDatabaseMinorVersion() <= 4) {
    queryStringBuilder.append(" IF NOT EXISTS ");
}
```

ro-rah:
Hi @karishmajain , I am inclined to agree with @nfalco79.

Please check out how @nfalco79 PR solved this issue and added unit test #1039 .  I believe that would supersede this PR.  Do you concur?  I was planning to close this PR otherwise.

ro-rah:
Closing for another PR that supersedes this.

PR #969, opened 2020-01-16
Targeted rollback implementation

wwillard7800:


liquibase-maven-plugin/src/main/java/org/liquibase/maven/plugins/LiquibaseRollbackOneChangeSetMojo.java
```diff
@@ -0,0 +1,134 @@
...
+
+    @Override
+    protected void checkRequiredParametersAreSpecified() throws MojoFailureException {
+      super.checkRequiredParametersAreSpecified();
+      checkRequiredRollbackParameters();
+    }
+
+    protected void checkRequiredRollbackParameters() throws MojoFailureException {
+
+      String message = "Cannot specify multiple rollbackXXX options, please select only"
```
nvoxland: This is just copied code that can be removed, correct?

liquibase-maven-plugin/src/main/java/org/liquibase/maven/plugins/LiquibaseRollbackOneChangeSetMojo.java
```diff
@@ -0,0 +1,134 @@
...
+        if (! hasProLicense()) {
+            return;
+        }
+        Database database = liquibase.getDatabase();
+        LiquibaseCommand liquibaseCommand = (CommandFactory.getInstance().getCommand("rollbackOneChangeSet"));
+        AbstractSelfConfiguratingCommand configuratingCommand = (AbstractSelfConfiguratingCommand)liquibaseCommand;
+        Map<String, Object> argsMap = getCommandArgsObjectMap(liquibase);
+        ChangeLogParameters clp = new ChangeLogParameters(database);
+        argsMap.put("changeLogParameters", clp);
+        if (force != null) {
```
nvoxland: Since they have to pass force=true as an argument, we should make sure they specified "true" not just anything

liquibase-core/src/main/java/liquibase/integration/commandline/Main.java
```diff
@@ -1783,6 +1909,12 @@ public boolean isWindows() {
         private static final String CHANGELOG_FILE = "changeLogFile";
         private static final String DATA_OUTPUT_DIRECTORY = "dataOutputDirectory";
         private static final String DIFF_TYPES = "diffTypes";
+        private static final String CHANGE_SET_ID = "changeSetId";
+        private static final String CHANGE_SET_AUTHOR = "changeSetAuthor";
+        private static final String CHANGE_SET_PATH = "changeSetPath";
+        private static final String OUTPUT_FILE = "outputFile";
```
nvoxland: outputFile is a global parameter and so already handled with the outputFile property. It should not be added to the command-level options

liquibase-core/src/main/java/liquibase/integration/commandline/Main.java
```diff
@@ -1333,7 +1379,6 @@ protected void doMigration() throws Exception {
                 String result = snapshotCommand.execute().print();
                 outputWriter.write(result);
                 outputWriter.flush();
-                outputWriter.close();
```
nvoxland: Why did you remove the outputWriter.close() in snapshot?

liquibase-maven-plugin/src/main/java/org/liquibase/maven/plugins/LiquibaseRollbackOneChangeSetMojo.java
```diff
@@ -0,0 +1,123 @@
...
+            return;
+        }
+        Database database = liquibase.getDatabase();
+        LiquibaseCommand liquibaseCommand = (CommandFactory.getInstance().getCommand("rollbackOneChangeSet"));
+        AbstractSelfConfiguratingCommand configuratingCommand = (AbstractSelfConfiguratingCommand)liquibaseCommand;
+        Map<String, Object> argsMap = getCommandArgsObjectMap(liquibase);
+        ChangeLogParameters clp = new ChangeLogParameters(database);
+        argsMap.put("changeLogParameters", clp);
+        if (force != null && ! Boolean.parseBoolean(force)) {
+            throw new LiquibaseException("Invalid value for --force.  You must specify --force=true to use rollbackOneChangeSet.");
```
nvoxland: there is no --force argument with maven. Better to say "you must specify liquibase.force=true" ?

liquibase-maven-plugin/src/main/java/org/liquibase/maven/plugins/LiquibaseRollbackOneChangeSetSQL.java
```diff
@@ -0,0 +1,191 @@
...
+     *
+     * The path to an output file
+     *
+     * @parameter property="liquibase.outputFile"
+     *
+     */
+    protected String outputFile;
+    private static ResourceBundle coreBundle = getBundle("liquibase/i18n/liquibase-core");
+    @Override
+    protected void checkRequiredParametersAreSpecified() throws MojoFailureException {
```
nvoxland: These copy/pasted methods are not needed in this class either

liquibase-core/src/main/java/liquibase/integration/commandline/Main.java
wwillard7800: I fat-fingered that one.  Not intentional.

liquibase-maven-plugin/src/main/java/org/liquibase/maven/plugins/LiquibaseRollbackOneChangeSetMojo.java
wwillard7800: Yes.

PR #970, opened 2020-01-19
Fix: Java 1.9+ versions require jaxb as separate dependency

b-gyula:
Without explicit jaxb dependency the maven plugin cannot be built with java 1.9+

┆Issue is synchronized with this [Jira Story](https://datical.atlassian.net/browse/LB-26) by [Unito](https://www.unito.io/learn-more)

### DIFF
--- a/liquibase-maven-plugin/pom.xml
+++ b/liquibase-maven-plugin/pom.xml
@@ -15,6 +15,11 @@
 
 	<dependencies>
 		<dependency>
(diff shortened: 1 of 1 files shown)


ro-rah:
Hi @b-gyula !

I just conditioned another PR you submitted, happy to see this submission as well.  I will try and get this into the same sprint.

Ronak

ro-rah:
Assessed as a Low risk, adding the dependency should be easy to test.

PR #983, opened 2020-01-31
Fixes for loadData to support Postgres enums with java.sql.Types.OTHER

sudiptosarkar:
Pull Request for Jira Issue - https://liquibase.jira.com/browse/CORE-3576

As part of this pull request, we're adding a data type - `OTHER` for allowing `loadData` for Enum types in Postgres, as well as several other datatypes supported by `java.sql.Types.OTHER`.

┆Issue is synchronized with this [Jira Story](https://datical.atlassian.net/browse/LB-20) by [Unito](https://www.unito.io/learn-more)


SteveDonie:
Thanks so much for your contribution! Is there any way you could add a test case to show how this would work? If you cannot, we can probably still merge this in, but it may take longer to get approved.

sudiptosarkar:
> Thanks so much for your contribution! Is there any way you could add a test case to show how this would work? If you cannot, we can probably still merge this in, but it may take longer to get approved.

@SteveDonie I wrote a Unit Test for my changes in `LoadDataChange.java`. But for the other change, the return type is void and I couldn't figure out where to begin writing the Unit Test for that (I didn't see any other unit test mocking `java.sql.Connection` either).

So, instead, for that one, I wrote an integration test. However, I see a rollback changelog there as well. I hope we're using a fresh database every time in travis. If not, I'll need your help and some more time to understand how the rollback changelog which is already there is working.

sudiptosarkar:
@SteveDonie Is the integration test okay, or do you need me to write a Unit Test for that one too? I could try to figure it out again, if we need it.

ro-rah:
Putting this at RiskMed

- Is OTHER the word we want to expose for this data type?  Internal time to suss it out if this is the way Postgres should have it done.  What other things come thru (design wise) for Postgres and other DBs?
- Changes functionality so prob not a patch but next major release.
- The current tests is fine, and thanks!

sudiptosarkar:
@ro-rah We could change that `OTHER` to `java.sql.Types.OTHER`. That's what I was trying in the beginning, as is stated in [the documentation](https://www.liquibase.org/documentation/column.html), but isn't actually implemented. But adding java.sql.Types for the other types (other than `OTHER`) would be extra work, and I didn't need those types when I wrote this, so I kept it simply as `OTHER` to avoid confusion.

ro-rah:
@sudiptosarkar , let's see what our internal team says since it is already in Conditioning.  Thanks for the insight!

liquibase-core/src/main/java/liquibase/statement/ExecutablePreparedStatementBase.java
```diff
@@ -130,6 +131,8 @@ private void applyColumnParameter(PreparedStatement stmt, int i, ColumnConfig co
             LOG.debug(LogType.LOG, "value is string = " + col.getValue());
             if (col.getType() != null && col.getType().equalsIgnoreCase(LoadDataChange.LOAD_DATA_TYPE.UUID.name())) {
                 stmt.setObject(i, UUID.fromString(col.getValue()));
+            } else if (col.getType() != null && col.getType().equalsIgnoreCase(LoadDataChange.LOAD_DATA_TYPE.OTHER.name())) {
+                stmt.setObject(i, col.getValue(), Types.OTHER);
```
abhirathod, line 135: Is this code take care of JSON objects as well ?

sudiptosarkar: I'm not sure if we need this for Json columns. If I remember correctly (I'm not associated with the organization for which I wrote this), I was already using STRING for the json columns and it was working just fine.

PR #985, opened 2020-02-01
Add missing "labels" field to serializable ChangeLogInclude

stanio:
I need this in the context of [CORE-3549][], but it appears a general omission.

[CORE-3549]: https://liquibase.jira.com/browse/CORE-3549 (Implement "compileChangeLogXml" command)

┆Issue is synchronized with this [Jira Story](https://datical.atlassian.net/browse/LB-18) by [Unito](https://www.unito.io/learn-more)

### DIFF
--- a/liquibase-core/src/main/java/liquibase/changelog/ChangeLogInclude.java
+++ b/liquibase-core/src/main/java/liquibase/changelog/ChangeLogInclude.java
@@ -1,6 +1,7 @@
 package liquibase.changelog;
 
 import liquibase.ContextExpression;
+import liquibase.LabelExpression;
 import liquibase.serializer.AbstractLiquibaseSerializable;
 
 import java.util.Arrays;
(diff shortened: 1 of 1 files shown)


stanio:
Actually I'm not seeing this documented (having `labels` attribute to `include` elements):

https://www.liquibase.org/documentation/include.html

although the XML Schema declares it (along with a bunch of others missing from `ChangeLogInclude`):

http://www.liquibase.org/xml/ns/dbchangelog/dbchangelog-3.8.xsd

`includeAll` doesn't appear to have `labels`, also.

The type of `DatabaseChangeLog.includeLabels` is `LabelExpression`, and not `Labels` as is `ChangeSet.labels` – I thought a labels expression is only given as runtime argument, and the changelog contains only label lists to match against... so I'm not sure this is all currently necessary.

ro-rah:
Hi @stanio , I will bring this PR up in review with the team.  Thanks for the submission! -Ronak

ro-rah:
Tagging this as:
RiskLow:
- bug fix
- isolated (just changing what @stanio needs)
- change is additive and not changing the signature or changing behavior.

stanio:
It is that I'm not sure this is supposed to be in there, at least currently.  I would rather close this one and leave it to the Liquibase developers to resolve.  I'm currently using a workaround for my purposes.

ro-rah:
@stanio, thanks for the reply.  I put this into conditioning where the LB devs can close and / or resolve as needed.  Appreciate your submission!

ro-rah:
Discussed with @nvoxland .  This is a general problem that we would want to solve for more than just labels, we'll close this for now and take it up internally.  Glad you found a workaround.

PR #991, opened 2020-02-04
Update to Get_Started.txt

mccormickc:
Fixed code examples

### DIFF
--- a/liquibase-dist/src/main/archive/GETTING_STARTED.txt
+++ b/liquibase-dist/src/main/archive/GETTING_STARTED.txt
@@ -78,12 +78,13 @@ About the console:
 
 NOTE: If you run formatted SQL or XML on the H2 database, then want to attempt using the other method, you must restart the H2 database before running it again or you will receive an update failure due to objects already existing on the database.
 
+
 Running Your First Update
 -------------------------
 The main Liquibase command you will use is "update", which applies any changes not yet run in your changelog to your database.
 
 First,
-1. Open a command-line or Terminal app.
+1. Open a different command-line or Terminal app.
 2. Navigate to your ...examples/sql directory (or xml)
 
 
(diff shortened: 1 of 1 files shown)


liquibase-dist/src/main/archive/GETTING_STARTED.txt
```diff
@@ -117,29 +117,18 @@ Running the update command allowed your developer database to match the defined
...
   SQL changelog example code
   ---------------------------
-
-    --changeset your.name:3
-    ALTER TABLE person ADD worksfor_company_id INT;
-
-    --changeset your.name:4
-    ALTER TABLE person ADD CONSTRAINT fk_person_worksfor FOREIGN KEY (worksfor_company_id) REFERENCES company(id);
+      --changeset your.name:4
+    ALTER TABLE person ADD FOREIGN KEY (fk_person_worksfor) REFERENCES company(id) ;
```
nvoxland: I think we need this fully left-justified because if people simply copy/paste it without removing the leading spaces it may cause problems.

liquibase-dist/src/main/archive/GETTING_STARTED.txt
```diff
@@ -117,29 +117,18 @@ Running the update command allowed your developer database to match the defined
...
-        <addColumn tableName="person">
-            <column name="worksfor_company_id" type="int"/>
-        </addColumn>
-    </changeSet>
     <changeSet id="4" author="your.name">
-        <addForeignKeyConstraint constraintName="fk_person_worksfor"
-                                 baseTableName="person" baseColumnNames="worksfor_company_id"
-                                 referencedTableName="company" referencedColumnNames="id"/>
+        <addForeignKeyConstraint constraintName="fk_person_worksfor" baseTableName="person" baseColumnNames="worksfor_company_id" referencedTableName="company" referencedColumnNames="id"></addForeignKeyConstraint>
     </changeSet>
```
nvoxland: We can't just take out these previous changeSets because the ones you left rely on them. 

If you want them to only add one changeSet, you'll have to get rid of 4 not 3

liquibase-dist/src/main/archive/GETTING_STARTED.txt
mccormickc: Justified :)

liquibase-dist/src/main/archive/GETTING_STARTED.txt
mccormickc: Kept both, and renumbered to 4 and 5 per Mario.

PR #992, opened 2020-02-05
Documentation update changes

b-gyula:
- Fix: Remove duplicated code in `AbstractChange.load` to avoid loading `update/whereParams/param` into the `columns`
- New separate type for `update`, `delete` / `whereParams/param`
- `ExecuteCommandChange.os` changed to String to match the xml / yml field type
- Added description for
  + setTableRemarks / remarks
  + setColumnRemarks / remarks
  + createTable / remarks + tablespace
  + createView / remarks
  + addForeignKeyConstraint / referencedTableCatalogName + referencedTableSchemaName
  + addAutoIncrement / generationType, defaultOnNull, startWith, incrementBy
  + dropTable / cascadeConstraints
- Fix: `XMLChangeLogSerializer` was not able to create tag with mixed content (`sql`)
  + executeCommand / args was not serialized
- XSD fixes:
  + addAutoIncrement @ `generationType`, `defaultOnNull` was missing
  + tableIsEmpty @ tableName was optional
  + dropNotNullConstraint @ columnName was optional
- Common table related properties extracted into `AbstractTableChange` to reduce code multiplication
- Common `add|removeColumn` methods moved into `ChangeWithColumn`
- Minor typo fixes


SteveDonie:
Thanks so much for these changes! There is quite a bit to review here, so please be patient with us as we review these.

b-gyula:
> 
> 
> Thanks so much for these changes! There is quite a bit to review here, so please be patient with us as we review these.

Shall I split it? Would it help?

SteveDonie:
@b-gyula a question - what testing has been done on this?

b-gyula:
I've verified + extended relevant test classes and successfully executed all of them

SteveDonie:
After review by the team, we feel that this PR is not something that we can merge into the 3.8.x branch. 

* There are changes to the XSD which are mainly structural/cleanup but also some that make logical changes. Liquibase has a policy of only making changes to the XSD for minor version changes (i.e. 3.9.0 can have xsd changes, but we don't allow changes from 3.8.6 to 3.8.7). 
* There are other API-breaking changes to a couple public classes such as ChangeWithColumns, ExecuteShellChange, and AbstractModifyDataChange which should not be done in a patch release. These will need some more thought in how they tie into the upcoming 3.9/4.0+ changes.

We are very appreciative of your efforts and plan to incorporate them in some form into a future release.

b-gyula:
- The code changes I understand. (I'll create another PR without code changes)
- The only change in the XSD was to __allow users using the XML format to use the `generationType` and `defaultOnNull` on `addAutoIncrement` change, which is a long existing feature.__ Why allow to use them only with other formats? 

The primary UI for liquibase is the XSD (or the JSON schema would be). One editing the changelog XML in a schema aware editor could get all the required information from the XSD if would describe the exact schema, the code is using. E.g. 
```xml
<changeSet id="" author="">
    <createIndex tableName="" indexName="">
        <column name="" valueClobFile="">
            <constraints foreignKeyName=""/>
        </column>
    </createIndex>
</changeSet>
```
Is a valid changeSet definition according to the XSD. Liquibase even generates and executes SQL for it: `CREATE INDEX PUBLIC. ON PUBLIC.()` resulting ugly SQL errors, whereas
- `tableName` should not be allowed to be empty
- `valueClobFile` attribute and `constraints` node is not (and most probably never will be) used

With some small changes in the XSD _these probably typo errors could be avoided while one creating the XML already in the editor!_ In our company we are using a more strict XSD allowing only implemented and used properties. You can find it in my `xsd` branch. (I've created a full code review to see which properties are used and changed the XSD accordingly)

If the XSD would contain the documentation (which could be generated from the code the same way as the change documentation is for the website) the user would not need to go the the website the get the information for every change / attribute one by one.

The same in YAML format is not accepted and a nice error is given @ validation from Liquibase. With a few lines of code the same could be achieved for XML also.

By the way this YAML is valid for Liquibase. Generates and executes SQL for it
```yaml
databaseChangeLog:
  - changeSet:
      changes:
      - createIndex:
          tableName: t
          columns:
          - column:
              name: c
              position: 2
              afterColumn: 4
```
Such a great product would deserve a user friendly UI and accurate documentation. I'm happy to help in that if you want me.

PR #1000, opened 2020-02-13
DAT 3917

wwillard7800:


liquibase-core/src/main/java/liquibase/integration/commandline/Main.java
```diff
@@ -136,6 +135,7 @@
     protected String sqlFile;
     protected String delimiter;
     protected String rollbackScript;
+    protected String deploymentId;
```
nvoxland: This is not a global argument, so shouldn't be added as a field like this

liquibase-maven-plugin/src/main/java/org/liquibase/maven/plugins/LiquibaseRollbackOneUpdateSQL.java
```diff
@@ -0,0 +1,148 @@
+// Version:   $Id: $
+// Copyright: Copyright(c) 2007 Trace Financial Limited
```
nvoxland: You shouldn't be copying this copyright info

PR #1001, opened 2020-02-19
Quoted field names for PostgreSQL query

Raibaz:
To retrieve sequence information from the database, Liquibase issues a SQL query and then retrieves the result expecting uppercase field names.

However, PostgreSQL converts all column names to lowercase, so the sequence information retrieval fails and the output of the `diff` command is something like this on a PostgreSQL:

```
Changed Sequence(s):
     catalog_item_history_sequence
          cacheSize changed from 'null' to '1'
          dataType changed from 'null' to 'bigint'
          maxValue changed from 'null' to '9223372036854775807'
          minValue changed from 'null' to '1'
          willCycle changed from 'null' to 'false'
```

This in turn results in [CORE-2495](https://liquibase.jira.com/browse/CORE-2495), because the changelog being created contains no actual changes and is invalid.

To preserve the case in the column names returned by PostgreSQL, the names in the query need to be quoted.

## Steps to Reproduce
1. Create a sequence on the url connection with uppercase attribute names (as given in the [PostgresSQL Documentation](https://www.postgresql.org/docs/9.5/sql-createsequence.html)):
{CODE}
CREATE SEQUENCE  IF NOT EXISTS  catalog_item_history_sequence 
     MINVALUE minvalue 1  MAXVALUE 100  CYCLE;
{CODE}
2. Create a sequence on the referenceUrl connection with changes to the catalog_item_history_sequence sequence.
{CODE}
CREATE SEQUENCE  IF NOT EXISTS  catalog_item_history_sequence 
     MINVALUE minvalue 2  MAXVALUE 200  NO CYCLE;
{CODE}
3. Execute a Liquibase diff.

## Expected Results
The diff returns:
minvalue changed from 2 to 1
maxvalue changed from 200 to 100
cycle changed from false to true.

## Automated Test Requirements
* Write an automated regression test for Postgres 
  * Test Name: Sequence diffs return expected differences in attributes.
  * Description: Jira ticket number

## Manual Test Requirements
* Execute the Steps to Reproduce with the old/broken Liquibase version to verify those steps are correct.
* Do a desk check with the developer
* Execute the Steps to Reproduce with the new build to verify diff is fixed.

┆Issue is synchronized with this [Jira Story](https://datical.atlassian.net/browse/LB-13) by [Unito](https://www.unito.io/learn-more)
┆Fix Versions: Liquibase 4.1.1

### DIFF
--- a/liquibase-core/src/main/java/liquibase/snapshot/jvm/SequenceSnapshotGenerator.java
+++ b/liquibase-core/src/main/java/liquibase/snapshot/jvm/SequenceSnapshotGenerator.java
@@ -207,16 +207,16 @@ public class SequenceSnapshotGenerator extends JdbcSnapshotGenerator {
                 Scope.getCurrentScope().getLog(getClass()).warning("Failed to retrieve database version: " + ignore);
             }
             if (version < 10) { // 'pg_sequence' view does not exists yet
-                return "SELECT c.relname AS SEQUENCE_NAME FROM pg_class c " +
+                return "SELECT c.relname AS \"SEQUENCE_NAME\" FROM pg_class c " +
                         "join pg_namespace on c.relnamespace = pg_namespace.oid " +
                         "WHERE c.relkind='S' " +
                         "AND nspname = '" + schema.getName() + "' " +
                         "AND c.oid not in (select d.objid FROM pg_depend d where d.refobjsubid > 0)";
             } else {
-                return "SELECT c.relname AS SEQUENCE_NAME, " +
-                        "  s.seqmin AS MIN_VALUE, s.seqmax AS MAX_VALUE, s.seqincrement AS INCREMENT_BY, " +
-                        "  s.seqcycle AS WILL_CYCLE, s.seqstart AS START_VALUE, s.seqcache AS CACHE_SIZE, " +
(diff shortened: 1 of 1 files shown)


SteveDonie:
Thank you for your PR!

Here’s what happens next:

- A member of the Liquibase team will take a look at your contribution and may suggest changes, additional tests, or provide other feedback. The PR will be prioritized according to our internal development and testing capacity.

- We’ll let you know when it’s ready to move to the next step or if any changes are needed.

Thanks again!

nvoxland:
We are going to do a bit of internal testing on this before merging it in. Just to make sure it doesn't cause unexpected problems.

molivasdat:
Hi @Raibaz Which version of Liquibase were you using when you tried this?  I just tried on v3.9.0 and it returned the expected results.  Thanks a bunch and great to be working with you.

Raibaz:
I am indeed experiencing this with v3.9.0, while it's not happening with v3.8.0. In fact, I temporarily downgraded to make this work.

molivasdat:
Hi @Raibaz I can't get this to repro but I am going to merge this anyway.  It will be in the 3.10.x code under the #1372 PR

nvoxland:
This change has been applied to master

PR #1005, opened 2020-02-25
CORE-3596  Use LinkedHashMap for a deterministic iteration order in DependencyUtil.java

cpugputpu:
This PR aims to fix https://liquibase.jira.com/browse/CORE-3596

The fix is to change HashMap to LinkedHashMap so that the iteration order remains stable and the assertion failure in `liquibase.util.DependencyUtilTest#testIndependentBranchesCase` will not occur any more. In this way, the test will be more stable.

### DIFF
--- a/liquibase-core/src/main/java/liquibase/util/DependencyUtil.java
+++ b/liquibase-core/src/main/java/liquibase/util/DependencyUtil.java
@@ -4,7 +4,7 @@ import liquibase.Scope;
 import liquibase.logging.LogType;
 
 import java.util.ArrayList;
-import java.util.HashMap;
+import java.util.LinkedHashMap;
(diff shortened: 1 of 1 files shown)


SteveDonie:
Thanks for the pull request! This looks like a low risk fix, I am recommending that this get merged in for our next release.

SteveDonie:
Did a test merge of this into a branch DAT-4459 that we are using for Low Risk merges into master, which will be version 4.0.0. That branch will be merged at the start of Sprint 7.2, which will be Thursday April 2.

nvoxland:
Merged into master for the 4.0 release. It may not make it into the 4.0 beta1, but will be part of the 4.0 final release.

PR #1010, opened 2020-02-26
Potential fix for generated columns on postgres

### LINKED ISSUE #1009: Postgres SERIAL and BIGSERIAL are created as INT / BIGINT without auto increment
The following liquibase changeset does not work as expected

```xml
        <createTable tableName="test">
            <column name="id" type="SERIAL">
            </column>
        </createTable>
```

This changeset results in sql

```sql
CREATE TABLE test (id INTEGER NOT NULL, CONSTRAINT TEST_PKEY PRIMARY KEY (id));
```

In liquibase 3.6.3 and earlier the type `SERIAL` was used for the column which means that a sequence is automatically included.

This issue seems to be caused by #874
It appears that SERIAL columns are replaced by INT, however liquibase does reconize that this column type is an auto increment column.

It would be acceptible if the generated sql would be

```sql
CREATE TABLE test (id INTEGER GENERATED BY DEFAULT AS IDENTITY NOT NULL, CONSTRAINT TEST_PKEY PRIMARY KEY (id));
```

WellingR:
This hopefully fixes #1009
Some notes:

* This fix only applies to the `CreateTableChange` possible we need to include it in other places too
* The postgres tests don't seem, to run fine on my laptop, so it is hard to test this for me (some pointers would be appreciated).

┆Issue is synchronized with this [Jira Story](https://datical.atlassian.net/browse/LB-4) by [Unito](https://www.unito.io/learn-more)

### DIFF
--- a/liquibase-core/src/main/java/liquibase/change/core/CreateTableChange.java
+++ b/liquibase-core/src/main/java/liquibase/change/core/CreateTableChange.java
@@ -68,6 +68,7 @@ public class CreateTableChange extends AbstractChange implements ChangeWithColum
             Object defaultValue = column.getDefaultValueObject();
 
             LiquibaseDataType columnType = DataTypeFactory.getInstance().fromDescription(column.getType() + (isAutoIncrement ? "{autoIncrement:true}" : ""), database);
+            isAutoIncrement |= columnType.isAutoIncrement();
             if ((constraints != null) && (constraints.isPrimaryKey() != null) && constraints.isPrimaryKey()) {
                 statement.addPrimaryKeyColumn(column.getName(), columnType, defaultValue, constraints.shouldValidatePrimaryKey(),
                     constraints.getPrimaryKeyName(),constraints.getPrimaryKeyTablespace());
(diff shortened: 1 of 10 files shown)


SteveDonie:
Thanks for the PR, I've asked Liquibase architect Nathan to review this.

ro-rah:
Hi @WellingR ,

Sorry for the delay, we are working on tweaking our PR process.  I would love to put this fix sooner rather than later.  To influence that, would you be willing to write [unit test(s)](https://liquibase.jira.com/wiki/spaces/LB/pages/1274937609/How+to+Write+Liquibase+Core+Unit+Tests) for this PR?  This link contains guidance on how to write matching liquibase-core/src/test/* files.

If you are up for it great, let us know if you need more any guidance.

-Ronak

WellingR:
I noticed that #1010 has been merged, but it has the effect that SERIAL columns are now generated as identity columns.

Back when I created this PR I was hoping that some discussion would eventually help us with finding a proper fix for this issue, however unfortunately the issue was merged even without tests have been added at all. (In the issue description I mentioned having trouble running the tests).

This change has the downside that the SQL generated by liquibase changes (again) which causes some changelogs to break...

See also the discussion in #1009

PR #1024, opened 2020-03-06
Work for CORE-3457 to allow MariaDB Sequences

mykelalvis:
Sequences appeared in MariaDB 10.3.
Existing logic does not reflect that, as a major version > 10 would also need a minor version >= 3

Reflects logic change pointed out by [Lijun Liao](https://liquibase.jira.com/jira/people/557058:563e7647-7f5f-43c8-a410-159804f19377) in issue [CORE-3457](https://liquibase.jira.com/browse/CORE-3457)

This would require some help with testing.

### DIFF
--- a/liquibase-core/src/main/java/liquibase/snapshot/jvm/SequenceSnapshotGenerator.java
+++ b/liquibase-core/src/main/java/liquibase/snapshot/jvm/SequenceSnapshotGenerator.java
@@ -250,6 +250,7 @@ public class SequenceSnapshotGenerator extends JdbcSnapshotGenerator {
             .queryForList(new RawSqlStatement("select SUBSTRING( NAME FROM LOCATE( '/' ,NAME ) + 1 ) AS SEQUENCE_NAME "
                 + "from information_schema.INNODB_SYS_TABLES "
                 + "where NAME like '" + schema.getName() +"/%' AND FLAG & 12288 ;"));
+            // see https://mariadb.com/kb/en/information-schema-innodb_sys_tables-table/#flag for 12288 value of (3 << 12)
             if (res.size() == 0) {
               return "SELECT name AS SEQUENCE_NAME WHERE 1=0";
             }
--- a/liquibase-core/src/main/java/liquibase/snapshot/jvm/SequenceSnapshotGenerator.java
+++ b/liquibase-core/src/main/java/liquibase/snapshot/jvm/SequenceSnapshotGenerator.java
@@ -20,6 +20,7 @@ import liquibase.structure.core.Sequence;
 import java.math.BigInteger;
 import java.util.List;
 import java.util.Map;
+import java.util.StringJoiner;
 
 /**
  * Snapshot generator for a SEQUENCE object in a JDBC-accessible database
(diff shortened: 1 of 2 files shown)


sbruhse:
Whats the status of this?

ro-rah:
Hi @mykelalvis , 

Sorry, it looks like no one responded to you.  

The status label of Discovery is the start of the process and as a part of our process is automatically added to new PRs now.  From here it moves to conditioning where we size and put it into an sprint to complete the work to be included in the sprint end release.

We're streamlining our PR process to encourage what you are doing, so thanks and we'll keep ya posted.

-Ronak

ro-rah:
Hey @mykelalvis!  

I can help with some testing, because that helps us merge these PRs faster as well.  Here is our guidance on [unit tests](https://liquibase.jira.com/wiki/spaces/LB/pages/1274937609/How+to+Write+Liquibase+Core+Unit+Tests) and [integration tests](https://liquibase.jira.com/wiki/spaces/LB/pages/1276608569/How+to+Write+Liquibase+Core+Integration+Tests).

Up for taking a try at writing the tests?  Feel free to post any questions, if I don't know the answer I will find someone internally that can help :)

Thanks!

Ronak

liquibase-core/src/main/java/liquibase/snapshot/jvm/SequenceSnapshotGenerator.java
```diff
@@ -242,6 +243,32 @@ protected String getSelectSequenceSql(Schema schema, Database database) {
...
+        } else if (database instanceof MariaDBDatabase) {
+          StringJoiner j = new StringJoiner(" \n UNION\n");
+          try {
+            List<Map<String, ?>> res = ExecutorService.getInstance().getExecutor(database)
+            .queryForList(new RawSqlStatement("select SUBSTRING( NAME FROM LOCATE( '/' ,NAME ) + 1 ) AS SEQUENCE_NAME "
+                + "from information_schema.INNODB_SYS_TABLES "
+                + "where NAME like '" + schema.getName() +"/%' AND FLAG & 12288 ;"));
+            // see https://mariadb.com/kb/en/information-schema-innodb_sys_tables-table/#flag for 12288 value of (3 << 12)
+            if (res.size() == 0) {
+              return "SELECT name AS SEQUENCE_NAME WHERE 1=0";
```
Caltalys, line 255: I think this line shoulb be `return "SELECT '" + schema.getName() +"' AS SEQUENCE_NAME WHERE 1=0";`

stevesaliman:
It might be worth noting that almost nothing in the method that was changed was tested before - only the H2 database path.  

Since Liquibase > 3.6 is practically unusable with MariaDB > 10.3, I'd argue that the normal testing requirements should be waived for this PR.  

In the future, someone should probably see what happens when there are a lot of sequences in the database - the UNION query this PR uses could get large, but for now, it is an improvement over the exception that we get now.

ro-rah:
Well articulated @stevesaliman, I did move it to Conditioning already.  We are on the same page!

My earlier guidance on adding testing is just me doing my utmost diligence to save time for analysis and execution not only as it passes through Development but QA too :)

The conversation and feedback is much appreciated as it now understand we have a bug fix for something already broken and that does reduce the risk and help push the PR along quicker :)

mykelalvis:
I can look at the testing deficiency (with some help, probably), sometime this week.  My apologies for lack of response.  GMail filtered this for me.

liquibase-core/src/main/java/liquibase/snapshot/jvm/SequenceSnapshotGenerator.java
mykelalvis: I thought that, too, but was reluctant to make the change.  It seems like it was that same literal in other sequence code.  I'll check again.

ro-rah:
I know our internal QA would appreciate that the most probably @mykelalvis :)

mykelalvis:
@stevesaliman Do you perhaps have a test DB (that you could share) that might trigger the effect you describe above with the union query?

stevesaliman:
@mykelalvis I don't have one I can easily share, but if you happen to be using the Groovy DSL for your changesets, the following changelog can create one easily enough:
```groovy
databaseChangeLog {
  for (int i=1; i<=1000; i++) {
    changeSet(id: "make_seq_${i}", author: 'steve') {
      createSequence(sequenceName: "my_seq_${i}")
    }
  }
}
```
However, I think we have a couple of larger issues in the patch.

The first problem I ran into was a permission problem.  My user didn't have privileges on innodb_sys tables.  Fortunately, there is a way to get sequence names from information_schema.tables, and I think all users can query this table by default.  The following query should do the trick:
```java
            .queryForList(new RawSqlStatement("select table_name AS SEQUENCE_NAME "
                + "from information_schema.TABLES "
                + "where TABLE_SCHEMA = '" + schema.getName() +"' "
                + "and TABLE_TYPE = 'SEQUENCE';"));
```
The other problem I had is that the query for when there are no sequences didn't seem to work on my database.  I had to change that to `return "SELECT 'name' AS SEQUENCE_NAME from dual WHERE 1=0";`

The good news is that MariaDB must have a pretty large buffer for queries, because everything seemed to work fine with 1000 sequences.  We can play with the number in the above changelog to see how big we can get before we have issues.

stevesaliman:
I played around with this some more, and it looks like we run out of memory at around 4000 sequences.  It would happen sooner or later than that depending on the length of the sequence names, but we definitely have a limit.  

That said, I don't know that I'm too worried about that limit at the moment - I don't think there will be many schemas that have thousands of sequences, and without a patch, we can't handle any sequences at all.

I think if we could address the 2 query quirks I mentioned before, we'd be good to apply the patch.

mykelalvis:
I obvs didn't get to this last week (paying work is the bane of the OSS class).  I hope to spend a few on it this week.

ro-rah:
Thanks for the update @mykelalvis !

stevesaliman:
I created #1176 To address this issue in the 3.10.x branch.  I don't think this should wait for 4.0, since Liquibase after about 3.6 is currently unusable in a MariaDB 10.3 environment.

mykelalvis:
I am going to try and give this PR some test time this weekend.  Is that still useful?
I _think_ I should rebase to HEAD for 4.x.x+, but per @stevesaliman 's merged #1176 perhaps I should continue and let more delicate hands do the rebasing?

molivasdat:
Hi @mykelalvis Yes additional testing time is definitely useful.  If you have any unit tests or integration tests to add that would be great!

molivasdat:
@mykelalvis To answer your second question around 4.0, at this point in time I would stay with the 3.10.x code and add more testing.

nvoxland:
The fix for mariadb support should be in 3.10.2 already. That was released before 4.0.0 so it's not in that release, but the fix for that was merged into the 4.0.x branch and master for the 4.0.1 release.

Are there problems with the 3.10.2 fix and/or people not seeing it work with current master or 4.0.x?

kataggart:
I am going to close this, assuming it was resolved by https://github.com/liquibase/liquibase/pull/1176.

Please re-open/correct me if this is still an issue for folks. Many thanks.

PR #1033, opened 2020-03-12
Enable snapshotByDefault for Data if output directory specified DAT-4129

wwillard7800:
THIS IS A “DUPLICATE” ticket that was automatically created when Wes created the PR to fix DAT-4129, which has been moved to LB-185.

Please allow this to move along as a zero point ticket that tracks LB-185

As a user who wants to export ALL the data from my db when i run generatechangelog, i want to use the
--dataOutputDirectory=<mydatadir> flag and by default have all the data in csv files in a named directory.
Previously, if no --diffTypes argument is specified, the output directory was not created.  Now, the output directory is
always created if the --dataOutputDirectory is specified.


SteveDonie:
@wwillard7800 can you give some description (for those outside of Datical) of what DAT-4129 is?

PR #1034, opened 2020-03-12
Change Liquibase#close to throw LiquibaseException

andreyakostov:
There is no need to force callers to handle Exception.
The implementation of the close method can currently throw only a DatabaseException. But since the library has a parent exception class - LiquibaseException - it is used instead to not force callers to change their catch clauses if additional logic is added in the future.

### DIFF
--- a/liquibase-core/src/main/java/liquibase/Liquibase.java
+++ b/liquibase-core/src/main/java/liquibase/Liquibase.java
@@ -1678,7 +1678,7 @@ public class Liquibase implements AutoCloseable {
     }
 
     @Override
-    public void close() throws Exception {
+    public void close() throws LiquibaseException {
         if (database != null) {
             database.close();
         }


SteveDonie:
Thanks for the PR! We will review this and get back to you. While this seems like a simple change, we have to be careful of unintended side-effects.

nvoxland:
Does this change in returned exceptions to a lower type break users who have existing code compiled against the version with the wider exception type?

andreyakostov:
It should not. `LiquibaseException` is-a `Exception`, so existing catch clauses or throws declarations will still work and compile successfully.

SteveDonie:
Did a test merge of this into a branch DAT-4459 that we are using for Low Risk merges into master, which will be version 4.0.0. That branch will be merged at the start of Sprint 7.2, which will be Thursday April 2.

nvoxland:
Merged it into master for the 4.0 release. It may not make it into 4.0 beta1, but it will be part of the 4.0 final release for sure.

andreyakostov:
> Merged it into master for the 4.0 release. It may not make it into 4.0 beta1, but it will be part of the 4.0 final release for sure.

Hello, @nvoxland, it seems this change was not part of the 4.0.0 release.

kevin-at-datical:
@andreyakostov I just checked the Liquibase.java source code, and it looks like this PR is in the [4.0.0 release](https://github.com/liquibase/liquibase/releases/tag/v4.0.0). Please respond and let us know if you are experiencing something that makes you believe otherwise.

andreyakostov:
Hello, @kevin-at-datical, I am referencing the liquibase-core library in my project from Maven Central. My IDE downloads the library source and there I see the old signature of `close()`, which throws `Exception`.

nvoxland:
Yeah, sorry. We have a difference in how the buildserver setup can work in master vs. 4.0 so can't do an easy merge like we're working on and it has to be cherry picked. I must have missed getting that done after merging the PR.

We'll have to get this into 4.0.1 instead.

andreyakostov:
Hello @nvoxland, just a reminder that this change is still not released.

nvoxland:
Thanks for the reminder, @andreyakostov. I cherry picked it into 4.2.x for a 4.2.1 release we are planning in the next few days.

We're working on fixing up the issues that are making it too easy for things like this to get missed.

PR #1040, opened 2020-03-17
Added NPE check in DiffChangeLog to handle Postgres dblink DAT-4253

wwillard7800:
At this point in time, Liquibase does not handle dblink objects in Postgres.  This fix is to prevent this fact from stopping operations.


wwillard7800:
Fix was made in DAT-3674

PR #1041, opened 2020-03-17
Create CODE_OF_CONDUCT.md

SteveDonie:
Looking at the project settings, I noticed a 'community' link and explored that. This seems like a good addition. Not sure of the contact email is the one we want to use.


SteveDonie:
@mariochampion can you please review the wording of this. It came from a GitHub example. Please check that the email address mentioned is correct.

PR #1044, opened 2020-03-19
AlterSequence: include NOORDER clause ordered="false" is specified

LeBezout:
* Fix the `isOrdered`description
* Fix the `ordered` attribute in the statement generator


SteveDonie:
@nvoxland please review. If risk is low, please re-label.

liquibase-core/src/main/java/liquibase/sqlgenerator/core/AlterSequenceGenerator.java
```diff
@@ -68,7 +68,9 @@ public ValidationErrors validate(AlterSequenceStatement alterSequenceStatement,
 
         if (statement.getOrdered() != null) {
             if (statement.getOrdered()) {
-                buffer.append(" ORDER");
+                buffer.append(" ORDER ");
+            } else {
```
SteveDonie, line 72: It appears that there is at least one platform (MS SQL Server) that does not allow an ORDER clause, which could cause a problem. Could be protected by statement.getOrdered() always returning null for certain platforms?

LeBezout: Hi, thanks for review.

You're right, so the liquibase documentation is wrong (says that "mssql" supports this) in:

* https://www.liquibase.org/documentation/changes/create_sequence.html
* https://www.liquibase.org/documentation/changes/alter_sequence.html

But my fix is just on NOORDER. Alter the sequence to switch from ORDER to NOORDER like it's already done for CYCLE/NOCYCLE or CACHE/NOCACHE.

ro-rah:
Hi @LeBezout!  I reviewed this internally and I think this is a low risk PR.  So, I am going to move it along the queue.  Thanks so much for your contributions, targeting Q2 release.

nvoxland (approved):
The change seems good to me. We have the separate validation check for whether "ordered" is supported in general and what that list should be can be out of the scope of this ticket. Looking a bit, I wonder if only Oracle really supports order/noorder in general? 

Given that maybe only oracle can really take advantage of the ordered attribute regardless, I'm good with asssuming that NOORDER is the correct syntax for ordered="false" when the database supports it. It is an improvment either way over ordered="false" doing nothing.

We still continue to not set anything when ordered is not set, so this only impacts people explicitly asking for ordered="false"

suryaaki2 (approved):

FBurguer:
For this PR, I made 2 tests:
1) Create a sequence with ordered="true" attribute and then use an alter sequence to change ordered="false": `PASS`
Changsets:
```
<changeSet  author="Create-Seq"  id="1">  
    <createSequence    
            incrementBy="2"  
            maxValue="1000"  
            minValue="10"  
            ordered="true"   
            sequenceName="seq_id"  
            startValue="11"/>  
</changeSet>
<changeSet  author="Alter-Seq"  id="2">  
    <alterSequence   
            ordered="false"  
            sequenceName="seq_id"/> 
</changeSet>
```
UpdateSQL output:
```
-- Changeset LeBezout/changelog.xml::1::Create-Seq
CREATE SEQUENCE LBUSER2.seq_id START WITH 11 INCREMENT BY 2 MINVALUE 10 MAXVALUE 1000 ORDER;

-- Changeset LeBezout/changelog.xml::2::Alter-Seq
ALTER SEQUENCE LBUSER2.seq_id NOORDER;
```

2) Create a sequence with ordered="false" attribute and then use an alter sequence to change ordered="true": `PASS`
Changsets:
```
<changeSet  author="Create-Seq2"  id="3">  
    <createSequence    
            incrementBy="2"  
            maxValue="1000"  
            minValue="10"  
            ordered="false"   
            sequenceName="seq_id2"  
            startValue="11"/>  
</changeSet>
<changeSet  author="Alter-Seq2"  id="4">  
    <alterSequence   
            ordered="true"  
            sequenceName="seq_id2"/> 
</changeSet>
```
UpdateSQL output:
```
-- Changeset LeBezout/changelog.xml::3::Create-Seq2
CREATE SEQUENCE LBUSER2.seq_id2 START WITH 11 INCREMENT BY 2 MINVALUE 10 MAXVALUE 1000 NOORDER;

-- Changeset LeBezout/changelog.xml::4::Alter-Seq2
ALTER SEQUENCE LBUSER2.seq_id2 ORDER;
```

TEST Environment:
Windows 10
Java 8
oracle container, oracle image: gvenzl/oracle-xe:18.4.0-full

tabbyf00:
Thanks for your PR submission! Your contribution has been included in [Liquibase 4.17.0](https://github.com/liquibase/liquibase/releases/tag/v4.17.0) released on October 10, 2022.

PR #1049, opened 2020-03-23
Support --schemas flag both before and after the "snapshot" command

wwillard7800:
##Overall: 
`--schemas` will only work for snapshot when to the right of the command, but for other liquibase commands, like `generatechangelog`, it works and is documented as working to the left (in the "global commands" space). we need to make sure --schemas works to the standard usage (to the left of the command).

## Context
A Liquibase client was having issue running a diff between a snapshot and multiple schemas.
The user workflow was as follows:
1. Snapshot the target database with multiple schemas using the flags: "--schema=<schema1,schema2,schema3>"
2. The user conducts a few changes to the target database.
3. The user then runs a diffChangeLog between the snapshot and the target database.
Unfortunately the resulting diffChangeLog did not carry all the changes that were made in all of the schemas specified while doing the original snapshot. 
After support conducted a few tests, it appears that when running the snapshot command, Liquibase does NOT include all the schemas specified with the "--schema<schema1,schema2,schema3>" flag in the snapshot file.
Instead, it only contains the default schema.

### DIFF
--- a/liquibase-core/src/main/java/liquibase/integration/commandline/Main.java
+++ b/liquibase-core/src/main/java/liquibase/integration/commandline/Main.java
@@ -1417,7 +1417,7 @@ public class Main {
                 SnapshotCommand snapshotCommand = (SnapshotCommand) CommandFactory.getInstance()
                         .getCommand(COMMANDS.SNAPSHOT);
                 snapshotCommand.setDatabase(database);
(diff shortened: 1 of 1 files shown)


nvoxland:
This change breaks anyone currently using --schemas after the snapshot command which we cannot do.

We also need to make sure it is used for snapshotReference and dropAll commands.

Issue #1056, opened 2020-03-26
Allow adding multiple unique column constraints on table creation.


rhpvorderman:
Hi!

Currently I am working on porting an existing project to SQLite. SQLite does not allow adding constraints to a table later. Adding foreign key constraints work really well on table creation, liquibase does a wonderful job there.

I run into a problem when I try to add a unique constraint across multiple columns.
The XML
```
        <createTable tableName="MY_EXAMPLE_TABLE">
            <column autoIncrement="true" name="MY_EXAMPLE_TABLE_ENTRY_ID" type="INTEGER">
                <constraints primaryKey="true" primaryKeyName="PK_MY_EXAMPLE_TABLE_ENTRY"/>
            </column>
            <column name="TABLE_KEY" type="TEXT">
                <constraints nullable="false"
                             uniqueConstraintName="UC_MY_EXAMPLE_TABLE_ENTRY_CCEI_DK" unique="true"/>
            </column>
            <column name="TABLE_VALUE" type="CLOB"/>
            <column name="SOME_OTHER_TABLE_ENTRY_ID" type="INTEGER">
                <constraints nullable="false"
                             foreignKeyName="FK_MY_EXAMPLE_TABLE_SOME_OTHER_TABLE_ENTRY_ID"
                             referencedColumnNames="OTHER_TABLE_ENTRY_ID"
                             referencedTableName="OTHER_TABLE_ENTRY"
                             uniqueConstraintName="UC_MY_EXAMPLE_TABLE_ENTRY_CCEI_DK"
                             unique="true"/>
            </column>
        </createTable>
```
Note that the `uniqueConstraintName` for the columns `SOME_OTHER_TABLE_ENTRY_ID` and `PK_MY_EXAMPLE_TABLE_ENTRY` is the same.

expected sql
```
CREATE TABLE "MY_EXAMPLE_TABLE" (
	"MY_EXAMPLE_TABLE_ENTRY_ID"	INTEGER NOT NULL,
	"TABLE_KEY"	TEXT NOT NULL,
	"TABLE_VALUE"	TEXT,
	"SOME_OTHER_TABLE_ENTRY_ID"	INTEGER NOT NULL,
	CONSTRAINT "UC_MY_EXAMPLE_TABLE_ENTRY_CCEI_DK" UNIQUE("TABLE_KEY","SOME_OTHER_TABLE_ENTRY_ID"),
[...]

PR #1058, opened 2020-03-27
Add code to allow for multiple column unique constraints on table creation.

rhpvorderman:
Hi! SQLite does not allow adding unique constraints after table creation. So the addUniqueConstraint clause can not be used. The createTable syntax only allows single-column unique constraints even if these have the same name: see #1056 . 

My proposed solution here is to combine all the columns for uniqueConstraints that have the same name. These should be the same constraint. I do not think there is a use case for defining single-column unique constraints with the exact same constraint name. I have added a test as well.

I am not a Java programmer (experience in Scala and Python) so I apologize for the clunky Java. There is probably some syntactic sugar that can be used to make the code less verbose and more readable.

Also, wow! Such readable code in this project! It was quite easy to implement something in the right place.

EDIT: I have a question: to which branch should I add this? Given that this adds a new feature it should be on 3.9.x which is master. Or if this is seen as a bugfix then it can be on 3.8.x. I'd prefer the latter, since this is blocking for SQLite on the project I am working on, but I defer to your judgement.

### DIFF
--- a/liquibase-core/src/main/java/liquibase/sqlgenerator/core/CreateTableGenerator.java
+++ b/liquibase-core/src/main/java/liquibase/sqlgenerator/core/CreateTableGenerator.java
@@ -33,7 +33,13 @@ import liquibase.structure.core.Table;
 import liquibase.util.StringUtil;
 
 import java.math.BigInteger;
-import java.util.*;
+import java.util.ArrayList;
+import java.util.HashMap;
+import java.util.HashSet;
+import java.util.Iterator;
+import java.util.LinkedList;
+import java.util.List;
+import java.util.Set;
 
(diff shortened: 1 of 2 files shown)


SteveDonie:
Thanks for your pull request! The team will review this and add comments with a status update soon. Thanks for the feedback on the code structure! 
We are currently working on both the 3.8.x line and the next release from master, which we will call 4.0.0. So if you could re-target this to 3.8.x that would be great.

SteveDonie:
I appreciate the test you added!

rhpvorderman:
@SteveDonie Thank you for your quick reaction! I will reopen the request for the 3.8 branch.

PR #1059, opened 2020-03-27
Add code to allow for multiple column unique constraints on table creation.

rhpvorderman:
This is the same as #1058 but now for the 3.8.x branch.

Hi! SQLite does not allow adding unique constraints after table creation. So the addUniqueConstraint clause can not be used. The createTable syntax only allows single-column unique constraints even if these have the same name: see #1056 .

My proposed solution here is to combine all the columns for uniqueConstraints that have the same name. These should be the same constraint. I do not think there is a use case for defining single-column unique constraints with the exact same constraint name. I have added a test as well.

I am not a Java programmer (experience in Scala and Python) so I apologize for the clunky Java. There is probably some syntactic sugar that can be used to make the code less verbose and more readable.

Also, wow! Such readable code in this project! It was quite easy to implement something in the right place.

┆Issue is synchronized with this [Jira Story](https://datical.atlassian.net/browse/LB-170) by [Unito](https://www.unito.io/learn-more)

### DIFF
--- a/liquibase-core/src/main/java/liquibase/sqlgenerator/core/CreateTableGenerator.java
+++ b/liquibase-core/src/main/java/liquibase/sqlgenerator/core/CreateTableGenerator.java
@@ -17,7 +17,13 @@ import liquibase.structure.core.*;
 import liquibase.util.StringUtils;
 
 import java.math.BigInteger;
-import java.util.*;
+import java.util.ArrayList;
+import java.util.HashMap;
+import java.util.HashSet;
+import java.util.Iterator;
+import java.util.LinkedList;
+import java.util.List;
+import java.util.Set;
 
 public class CreateTableGenerator extends AbstractSqlGenerator<CreateTableStatement> {
 
--- a/liquibase-core/src/main/java/liquibase/sqlgenerator/core/CreateTableGenerator.java
+++ b/liquibase-core/src/main/java/liquibase/sqlgenerator/core/CreateTableGenerator.java
@@ -17,9 +17,7 @@ import liquibase.structure.core.*;
 import liquibase.util.StringUtils;
 
 import java.math.BigInteger;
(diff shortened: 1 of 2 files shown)


SteveDonie:
Thanks for your pull request! The team will review this and add comments with a status update soon.

liquibase-core/src/main/java/liquibase/sqlgenerator/core/CreateTableGenerator.java
```diff
@@ -279,7 +283,38 @@ public ValidationErrors validate(CreateTableStatement createTableStatement, Data
...
         for (UniqueConstraint uniqueConstraint : statement.getUniqueConstraints()) {
+            if (uniqueConstraint.getConstraintName() == null) {  // Only combine uniqueConstraints that have a name.
+                aggregatedUniqueConstraints.add(uniqueConstraint);}
+            else {
+                String constraintName = uniqueConstraint.getConstraintName();
+                if (uniqueConstraintMap.containsKey(constraintName)) { // if we have seen the uniqueContraint before ..
+                    UniqueConstraint oldUniqueConstraint = uniqueConstraintMap.get(constraintName);
+                    // check if it is really the same constraint
+                    if (oldUniqueConstraint.shouldValidateUnique() != uniqueConstraint.shouldValidateUnique()) {
+                        aggregatedUniqueConstraints.add(uniqueConstraint);  // if not, add it to the aggregated constraints
```
rhpvorderman, line 301: Maybe it is better to throw an error here about conflicting column information?

nvoxland: Maybe at least a warning. I'll add that in as I merge it

rhpvorderman:
Thanks @nvoxland !

Issue #1077, opened 2020-04-03
Separate Firebird v3 from earlier Firebird versions


Wallaby1999:
**Description**
Firebird 3 has some differences with respect to the created SQL statements. Currently there is no way to distinguish versions prio v3 from v3 and onwards

**Expected behavior**
The Firebird 3 behaviour should get explictily available by using its own short name "firebird3" and by creating the necessary test cases.

**Additional context**
The Firebird3Database class should be extended and all test cases should be implemented for it as well.

ro-rah:
Added Risk Medium Tag due the volume of code changes but Firebird v3 changes look pretty well isolated.  Nice use of integration tests!

lbadam:
Closing this as this is not work we plan to implement. For anyone wanting to use Firebird with Liquibase, please [read our documentation](https://contribute.liquibase.com/extensions-integrations/directory/database-tutorials/firebird/).

PR #1079, opened 2020-04-03
Separate Firebird v3 from earlier Firebird versions

Wallaby1999:
Proposal for implementing 'firebird3' together with tests


ro-rah:
Thanks @Wallaby1999 for your pull request!

Here’s what happens next:

-Ronak

ro-rah:
closes #1077

ro-rah:
This has been targeted for release in Q2, updates as work progresses. :)

ro-rah:
Hey @Wallaby1999,

I was looking at https://github.com/liquibase/liquibase/issues/1093, were you latest changes differentiating the shortname for Firebird vs Firebird3?

Thanks,

Ronak

Wallaby1999:
> were you latest changes differentiating the shortname for Firebird vs Firebird3?

Yes, it does. If you take a look into commit [2a651ce6ba26a7dab9d9cd14872c203c3023fa5d](https://github.com/liquibase/liquibase/pull/1079/commits/2a651ce6ba26a7dab9d9cd14872c203c3023fa5d) it is introducing firebird3 as shortname for Firebird Databases >= 3.

ro-rah:
Brilliant @Wallaby1999!  

I thought those commits covered that case but since I'm still a liquibase newb, I thought it better to ask than to hold up this PR for a lack of my understanding.  Thanks for the response!

-Ronak

nvoxland:
I think we'd want to keep the shortname the same between 3 vs. not. 

As a user, it would be surprising they'd need to use a "firebirdv3" vs. just "firebird" in dbms attributes, and it would require updating old changeSets when a database is upgraded from 2 to 3.

I see splitting the V3Database class off vs. adding "if version > 3" blocks in the code as an implementation detail of how we want to manage the differences vs. something that should be exposed to the users.

Wallaby1999:
@nvoxland Just for clarification: it would be enough to use the same shortname for both Database classes and adjust the tests accordingly?

Wallaby1999:
So I just added a commit to remove the "firebird3" shortname again, but now I am not able to differentiate the Firebird versions in the tests, when the generated SQL is different:

- ./liquibase-core/src/test/java/liquibase/verify/saved_state/compareGeneratedSqlWithExpectedSqlForMinimalChangesets/addLookupTable/firebird.sql
- ./liquibase-core/src/test/java/liquibase/verify/saved_state/compareGeneratedSqlWithExpectedSqlForMinimalChangesets/addNotNullConstraint/firebird.sql
- ./liquibase-core/src/test/java/liquibase/verify/saved_state/compareGeneratedSqlWithExpectedSqlForMinimalChangesets/dropNotNullConstraint/firebird.sql

@nvoxland: How could I solve that issue and test both cases Firebird >= 3 and < 3?

nvoxland:
Different SQL with different versions is a general problem with the current test setup for many databases, not just firebird 2 vs. 3. 

We're working on improving that in a separate effort, but don't have anything to give you with it right now besides just have the tests at least consistent between 2 or 3. I'm guessing v3 is most common, so that is probably the best version to commit.

nvoxland:
Prior to merging this in, I will add firebird support to the build server

grauds:
Hi, could we please bump it up? Is this planned to be resolved? Or there is a well known workaround?

MalloD12:
Hi @Wallaby1999,

Are you still keen to get this PR merged? Would you mind resolving the existing merge conflicts we have at the moment? In the same line as what Nathan mentioned, I think we should align this PR to use the existing `FirebirdDatabase` class and not have a new one just to differentiate the database version number. I think if in some specific test, we need to run some Firebird tests with a specific version number we can create an instance of it by overriding its `getDatabaseMajorVersion()` method. Something like this:

```
 FirebirdDatabase h2DatabaseV1 = new FirebirdDatabase() {
            @Override
            public int getDatabaseMajorVersion() {
                return 3;
            }
 };
```
If you have any questions/concerns please do not hesitate to reach out to us. I apologize for the time this PR has been waiting, but it's now on our radar and we will try to provide an answer as soon as possible.

Thanks,
Daniel.

Wallaby1999:
It is more than three years since I started this. Unfortunately it is totally out of interest now. Do whatever you want with this, I do not care any more about it.

Issue #1039, opened 2020-03-17
[3.8.x] CreateSequence fails on Postgres <9.4 due to IF NOT PRESENT


jpmaas:
**Description**
Due to recent changes in liquibase 3.8.x, the statement to create a sequence is as follows:
`CREATE SEQUENCE IF NOT EXISTS <schema>.<sequenceName> <further configuration>`.

This is fine for versions >= 9.5. In versions below that, the IF NOT EXISTS syntax is not supported.

**To Reproduce**
Add changelog to create a sequence and execute it.

```
<changeSet author="jpmaas" id="identifier">
        <preConditions onFail="MARK_RAN">
            <not>
                <sequenceExists sequenceName="sequence"/>
            </not>
        </preConditions>
        <createSequence sequenceName="sequence" startValue="1"/>
    </changeSet>
```

In my case, this changeset is executed via application startup using spring boot 2.

During startup, the following exception is thrown (starting from cause because spring exception is not necessary):

```
Caused by: liquibase.exception.MigrationFailedException: Migration failed for change set db/changelog.xml::identifier::jpmaas:
     Reason: liquibase.exception.DatabaseException: ERROR: syntax error at or near "NOT"
  Position: 21 [Failed SQL: (0) CREATE SEQUENCE  IF NOT EXISTS schema.sequence START WITH 1]
        at liquibase.changelog.ChangeSet.execute(ChangeSet.java:646)
        at liquibase.changelog.visitor.UpdateVisitor.visit(UpdateVisitor.java:53)
        at liquibase.changelog.ChangeLogIterator.run(ChangeLogIterator.java:83)
        at liquibase.Liquibase.update(Liquibase.java:202)
        at liquibase.Liquibase.update(Liquibase.java:179)
        at liquibase.integration.spring.SpringLiquibase.performUpdate(SpringLiquibase.java:366)
        at liquibase.integration.spring.SpringLiquibase.afterPropertiesSet(SpringLiquibase.java:314)
        at liquibase.integration.spring.MultiTenantSpringLiquibase.runOnAllSchemas(MultiTenantSpringLiquibase.java:147)
        at liquibase.integration.spring.MultiTenantSpringLiquibase.afterPropertiesSet(MultiTenantSpringLiquibase.java:94)
        at org.springframework.beans.factory.support.AbstractAutowireCapableBeanFactory.invokeInitMethods(AbstractAutowireCapableBeanFactory.java:1855)
[...]

PR #1096, opened 2020-04-16
Fix 1039 sequence statement generation for PostgreSQL <= 9.4

nfalco79:
## Environment

**Liquibase Version**: 3.8.9

**Liquibase Integration & Version**: CLI, maven

**Database Vendor & Version**: PostgreSQL 9.4

## Pull Request Type

## Description

This fix the issue #1039 New liquibase version introduce a IF NOT EXIST in the sequence statement that is not supported for db version <= 9.4 as documented [here](https://www.postgresql.org/docs/9.5/sql-createsequence.html)

- [ ] Added Unit Test(s)
- [ ] Added Integration Test(s)
- [ ] Documentation Updated

This was tested against my local db and newer version using docker image of postgresql 10.16 and 12

┆Issue is synchronized with this [Jira Story](https://datical.atlassian.net/browse/LB-207) by [Unito](https://www.unito.io/learn-more)

### DIFF
--- a/base.pom.xml
+++ b/base.pom.xml
@@ -111,6 +111,13 @@
             <scope>test</scope>
         </dependency>
 
+        <dependency>
+            <groupId>org.mockito</groupId>
+            <artifactId>mockito-core</artifactId>
+            <version>3.3.3</version>
+            <scope>test</scope>
+        </dependency>
+
 
         <dependency> <!-- use a specific Groovy version rather than the one specified by spock-core -->
             <groupId>org.codehaus.groovy</groupId>
--- a/liquibase-core/src/main/java/liquibase/sqlgenerator/core/CreateSequenceGenerator.java
+++ b/liquibase-core/src/main/java/liquibase/sqlgenerator/core/CreateSequenceGenerator.java
@@ -2,6 +2,7 @@ package liquibase.sqlgenerator.core;
 
 import liquibase.database.Database;
 import liquibase.database.core.*;
+import liquibase.exception.DatabaseException;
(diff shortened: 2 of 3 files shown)


nfalco79:
I add a test cases for these changes

ro-rah:
@nfalco79,

Thanks so much for adding the unit test, love to see a complete PR.  Will be reviewing, will update shortly.

-Ronak

ro-rah:
Added RiskLow tag, as user @nfalco79 added unit tests to the PR (awesome!)

ro-rah:
@nfalco79 , I added to the next sprint for now, its reasonable to expect this out in Q2.  Thanks so much for your submission! -Ronak

nvoxland:
While it can be nice to auto-add an "IF EXISTS" to object creation to avoid errors, I also worry about magically adding that when the person writing the changeSet did not explicitly ask for it. 

We are looking hard at usability, and one of the factors in that is how much we try to guess at what a user really wants to do, and how much we do specifically what they told us to do. 

Using this as an example, the user did not tell us to continue without an error if the objects exist. Do we guess that is what they actually wanted? Or do we just do what they said? 

I'm not sure the correct answer yet, but also want to make sure that whatever the answer is we apply it consistently across all the databases and change types. 

Anyone have thoughts on that?

nvoxland:
Nevermind, I see that the IF EXISTS is currently in the code, and this fix is just making sure that is not used for pre-9.5 postgreql.

We'll bring this fix in, but will continue to have the discussion on how we should handle smart logic like "if exists" .

PR #1150, opened 2020-05-13
Support `onFailMessage` and `onErrorMessage` in formatted SQL

12345ieee:
Introduce two new attributes of the `preconditions` element:
--preconditions onFailMessage:fail\ message onErrorMessage:error_message
to set the respective xml attributes.

Whitespace can be used in messages if escaped with a backslash.

---
name: Pull Request
about: Create a report to help us improve
title: ''
labels: Status:Discovery
assignees: ''

---

## Environment

**Liquibase Version**: 3.8.9 (applies to lower versions as well)

**Database Vendor & Version**: Any (tested on Oracle)

**Operating System Type & Version**: Any (tested on Linux)

## Pull Request Type

- [x] Enhancement/New feature (non-breaking change which adds functionality)

## Description

From the SO question: https://stackoverflow.com/questions/61748020/on-fail-message-for-sql-formatted-changesets
we find that it is currently impossible to set the `onFailMessage` and `onErrorMessage` attributes from formatted SQL changesets. This PR introduces said support, via:
`--preconditions onFailMessage:fail\ message onErrorMessage:error_message`.

## Steps To Reproduce
Try to use `onFailMessage` and `onErrorMessage` attributes from formatted SQL changesets

## Actual Behavior
No matter what you write, you can't set those attributes (as no code path that sets them exists)

## Expected/Desired Behavior
By using `--preconditions onFailMessage:fail\ message onErrorMessage:error_message` you can set the fail message to `fail message` and the error message to `error_message`.

## Additional Context
The patch is really basic, it would be more efficient to write a proper subparser instead of throwing the same string under a regex per every tag, but I did the minimum here, can refactor if asked to.
I wanted to allow whitespace, so I decided to allow escaping it with a `\`. If other whitespace systems are preferred (i.e. quoting the whole message), just ask.

I have based this on the `3.10.x` branch, but I can rebase this over `master` if it's more appropriate (clearly I'd like it to get in the first release possible for my own use).

- [x] Build is successful and all new and existing tests pass
- [ ] Added Unit Test(s) -> AFAICT there is no dedicated formatted SQL unit testing, beside reparsing a generated SQL file
- [ ] Added Integration Test(s)
- [ ] Documentation Updated -> the documentation at https://www.liquibase.org/documentation/sql_format.html should include this change, but I couldn't find where the site source is stored (if accessible).

┆Issue is synchronized with this [Jira Story](https://datical.atlassian.net/browse/LB-261) by [Unito](https://www.unito.io/learn-more)


ro-rah:
Hi @12345ieee, would you like to make these fixes against 3.10.x?  If it is not specific to that version then we would recommend to do the pull request to mater.

12345ieee:
I'll reopen it against the right branch then.

ro-rah:
@12345ieee much appreciated!

PR #1417, opened 2020-09-20
Use postgres advisory lock as changelog lock service

capricube:
This commit introduces a new lock service that specifically
targets Postgres >= 9.1 and uses transaction advisory lock
(pg_try_advisory_xact_lock). This fixes (at least for pg)
the issue of DATABASECHANGELOGLOCK getting stuck in lock
when Liquibase does not exit cleanly.

Transaction advisory locks have the nice property that don't need
to be explicitly released. In fact, they are automatically released
by pg when the transaction is completed or the connection closed.

This pull request is a duplicate of https://github.com/liquibase/liquibase/pull/872 and contains the resolved merged conflice.

┆Issue is synchronized with this [Jira Bug](https://datical.atlassian.net/browse/LB-690) by [Unito](https://www.unito.io)


molivasdat:
Hi @capricube Thanks for creating this PR. If you could fix any build failures that would be great too.
Here’s what happens next:

according to our development and testing capacity.

capricube:
Hi @molivasdat i've now fixed the failing unit tests except for test StandardServiceLocatorTest which i do not fully understand yet.

capricube:
@molivasdat finally the build succeeded!

nvoxland:
This makes the databasechangeloglock do some postgresql-specific things when on postgresql. I think we shouldn’t just take this as-is, but use it as fodder in a larger refactoring of the databasechangeloglock logic as we think about whether to include database-specific functionality or keep it more generic. And if we want to have database-specific features, how to redesign the service to support that.

filipelautert:
This PR duplicates https://github.com/liquibase/liquibase/pull/872 , so I'll close it. If someone wants to update it (or PR #872) I can get it reviewed and merged now.

PR #1479, opened 2020-10-14
Fixes #1477 missing uniqueconstraints in snapshot

### LINKED ISSUE #1477: Liquibase 3.10+/4.1+ missing UniqueConstraints in snapshot
## Environment
**Liquibase Version**: 
4.1.0

<none>

liquibase-hibernate5, version 4.1.0

Postgres, 12 (running in a docker container based on the postgres:12 container)

Linux, Ubuntu 20.04/CentOS 7+  (local/buildserver)

## Description
When building a snapshot of a Postgres Database (seemingly any non-Oracle and non-MSSQL database) only the first four uniqueconstraints get loaded. All others are ignored because liquibase switches to a bulk "get-columns-for-uniqueconstraint" query but saves the results under the wrong cachekey, leading to no columns for the uniqueconstraints when the cached entry is then read, which in turn leads to the uniqueconstraint getting ignored.

This is especially noticeable if this snapshot gets compared/diffed against a snapshot that does read in all uniqueconstraints, since it will report a mismatch (provided the databases have more than 4 uniqyueconstraints).

In our specific case we use liquibase-hibernate5 to build a reference snapshot based on our JPA annotations, and compare this against a postgres database that has run all liquibase changelogs. If those database match, our liquibase changelogs is complete, if not, we need to add changesets. 
However, due to this bug liquibase now incorrectly reports 33 out of 37 uniqueconstraints as missing in the postgres database.

This seems to be the caused by https://github.com/liquibase/liquibase/commit/bebbb6eeec29f857b321704f3705eb8e683b4079 (#1199 ), where the logic to batch select the columns for uniqueconstraints gets expanded to all database (instead of just oracle and MSSQL) but the "CONSTRAINT_CONTAINER" field didn't get included in the select of the actual queries, leading to results being saved under the cachekey "null_uq_mytableconstraintname" instead of "myschema_uq_mytableconstraintname".

Upon loading the results, liquibase does look for the correct "myschema_uq_mytableconstraintname", gets a result of null despite and deduces there are no columns therefore the uniqueconstraint should not be added to the snapshot.

## Steps To Reproduce
Create a DatabaseSnapshot based on a fully initialized (ie, all changelog run) postgres database, create a DatabaseSnapshot based on a hibernate/jpa annotations "database" (or a fully initilized oracle or mssql database, as long as it is guaranteed to include _all_ uniqueconstraints). (Make sure their should be >4 uniqueconstraints present).

Run a diff on the two snapshot using the liquibase DiffGenerator. 

## Actual Behavior
The diff reports a difference in uniqueconstraints caused by the postgres database snapshot having fewer unique constraints than the other snapshot.

## Expected/Desired Behavior
Both databasesnapshots should have the same uniqueconstraints.

## Additional Context

┆Issue is synchronized with this [Jira Bug](https://datical.atlassian.net/browse/LB-820) by [Unito](https://www.unito.io/learn-more)
┆Fix Versions: Liquibase 4.2.1

haster:
- - -
name: Fix #1477 missing constraints in snapshot
about: Fixes the bug as described in #1477
title: ''
labels: Status:Discovery
assignees: ''

- - -

## Environment
**Liquibase Version**: 4.1.0

liquibase-hibernate5 4.1.0

Postgres 12

Linux

## Pull Request Type

* [x] Bug fix (non-breaking change which fixes an issue.)  closes #1477 
* [ ] Enhancement/New feature (non-breaking change which adds functionality)

## Description
As described in #1477, this PR fixes an issue where due to incorrect caching of batch-query results every uniqueconstraint after the first four was ignored for a number of databases when building a snapshot, leading to incorrect snapshots/diff/changelog-generation.

This PR:

* Adds a check with explicit exception when trying to save a batch-query result while the CONSTRAINT_CONTAINER and/or CONSTRAINT_NAME are not present in the query result (which would lead to an incorrect cachekey).
* Adds said fields to a number of queries for specific databases where they were missing and I felt sure enough of what I was doing to add it.

Note that those fields are still missing from some of the queries and therefore this code will probably start throwing exceptions on some platforms! Though these are all cases where the code would previously generate incorrect results, and fail-fast is probably a better option. 

In cases where the query can't (easily) provide a CONSTRAINT_CONTAINER you might be able to workaround that by "hardcoding" it in the query set to the current schema name. Something like 

```java
query = "select " + database.correctObjectName(schema.getName(), Schema.class) + " AS CONSTRAINT_CONTAINER, "
```

In cases where the CONSTRAINT_NAME is missing the query has to be fixed to provide that or the UniqueConstraintSnapshotGenerator has to be modified to never use (and cache) batch-queries for those RDBMSs. 

## Steps To Reproduce
See #1477 

## Actual Behavior
For any RDBMS, if the "get columns for unique constraint ( UniqueConstraintSnapshotGenerator.listColumns ) query has been executed more than 4 times, it will execute without filtering for a specific constraint, ie it will try to get the columns for all constraints at once, and then cache these results per constraint. However, if the CONSTRAINT_CONTAINER or CONSTRAINT_NAME fields are missing from the query-results it will cache these under an incorrect (and potentially non-unique) cachekey. 

When the generator then tries to retrieve the columns for a UniqueConstraint, it will see there are results cached, but will be unable to retrieve results for the specific constraint, leading it to conclude there are no columns for that constraint and it wil ignore the constraint and not add it to the snapshot (it gets added to the "always null" collection).

This then leads to invalid/incorrect results when using this snapshot to diff against another snapshot or generate a changelog.

## Expected/Desired Behavior
When creating a database snapshot, the snapshot will either succeed, meaning it gathered the correct number of unique constraints, or it will fail explicitly with an error if this is not possible. It will not create invalid data, leading to difficult-to-track-down errors further down the line.

## Additional Context
Confirmed to solve the problem for us, when building a snapshot of a postgres 12 database to diff against another database (that wasn't affected by this bug, nor by the solution).

* [x] Build is successful and all new and existing tests pass   (locally, most integration tests auto-skipped)
* [ ] Added [Integration Test(s)](https://liquibase.jira.com/wiki/spaces/LB/pages/1276608569/How+to+Write+Liquibase+Core+Integration+Tests)
* [ ] Documentation Updated

## Need Help?

┆Issue is synchronized with this [Jira Bug](https://datical.atlassian.net/browse/LB-822) by [Unito](https://www.unito.io/learn-more)
┆Fix Versions: Community 4.x

### DIFF
--- a/liquibase-core/src/main/java/liquibase/snapshot/jvm/UniqueConstraintSnapshotGenerator.java
+++ b/liquibase-core/src/main/java/liquibase/snapshot/jvm/UniqueConstraintSnapshotGenerator.java
@@ -342,6 +342,14 @@ public class UniqueConstraintSnapshotGenerator extends JdbcSnapshotGenerator {
                 columnCache = new HashMap<>();
                 snapshot.setScratchData(cacheKey, columnCache);
                 for (Map<String, ?> row : rows) {
+                    if (!row.containsKey("CONSTRAINT_CONTAINER"))
+                    {
+                        throw new IllegalArgumentException("Missing 'CONSTRAINT_CONTAINER' for bulk query!");
+                    }
+                    if (!row.containsKey("CONSTRAINT_NAME"))
+                    {
+                        throw new IllegalArgumentException("Missing 'CONSTRAINT_NAME' for bulk query!");
+                    }
                     String key = row.get("CONSTRAINT_CONTAINER") + "_" + row.get("CONSTRAINT_NAME");
                     List<Map<String, ?>> constraintRows = columnCache.get(key);
                     if (constraintRows == null) {
--- a/liquibase-core/src/main/java/liquibase/snapshot/jvm/UniqueConstraintSnapshotGenerator.java
+++ b/liquibase-core/src/main/java/liquibase/snapshot/jvm/UniqueConstraintSnapshotGenerator.java
@@ -170,7 +170,7 @@ public class UniqueConstraintSnapshotGenerator extends JdbcSnapshotGenerator {
             snapshot.setScratchData(queryCountKey, columnQueryCount + 1);
 
             if ((database instanceof MySQLDatabase) || (database instanceof HsqlDatabase)) {
-                sql = "select const.CONSTRAINT_NAME, COLUMN_NAME "
+                sql = "select const.CONSTRAINT_NAME, COLUMN_NAME, const.constraint_schema as CONSTRAINT_CONTAINER "
                         + "from " + database.getSystemSchema() + ".table_constraints const "
                         + "join " + database.getSystemSchema() + ".key_column_usage col "
                         + "on const.constraint_schema=col.constraint_schema "
@@ -183,7 +183,7 @@ public class UniqueConstraintSnapshotGenerator extends JdbcSnapshotGenerator {
                 }
                 sql += "order by ordinal_position";
             } else if (database instanceof PostgresDatabase) {
-                sql = "select const.CONSTRAINT_NAME, COLUMN_NAME "
+                sql = "select const.CONSTRAINT_NAME, COLUMN_NAME, const.constraint_schema as CONSTRAINT_CONTAINER "
                         + "from " + database.getSystemSchema() + ".table_constraints const "
                         + "join " + database.getSystemSchema() + ".key_column_usage col "
                         + "on const.constraint_schema=col.constraint_schema "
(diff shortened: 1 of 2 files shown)


liquibase-core/src/main/java/liquibase/snapshot/jvm/UniqueConstraintSnapshotGenerator.java
```diff
@@ -342,6 +342,14 @@ protected void addTo(DatabaseObject foundObject, DatabaseSnapshot snapshot) thro
                 columnCache = new HashMap<>();
                 snapshot.setScratchData(cacheKey, columnCache);
                 for (Map<String, ?> row : rows) {
+                    if (!row.containsKey("CONSTRAINT_CONTAINER"))
```
nvoxland, line 345: I get the worry that without the constraint_container columns, other databases will hit the same bug and it will go un-noticed. But, this check will end up breaking snapshot support completely vs. being wrong. 

We should find a way to either know what databases can support the "bulk" queries either dynamically (by searching the sql for a constraint_contaier string) or hard-coding the list better.

nvoxland (changes requested):

nvoxland:
Playing with the code a bit, I think the best thing is to undo the change that defaulted the "preferBulk" change since we really need that extra constraint_container column in there. 

I can revert that change and then incorporate your extra constraint_container columns.

molivasdat:
This was fixed by #1548 in 4.2.1

PR #1502, opened 2020-10-23
Fix CockroachDB Tests

rafiss:
* Create test database and schemas in integration test setup
* Make AddPrimaryKeyStatement account for DESC primary key
* Use DROP INDEX to drop unique constraints in CockroachDB
* Fix CREATE VIEW test changeset to not reference a different schema
* Skip rollback PK test for CockroachDB
* Stop referencing DB explicitly in CockroachDB test changeset
* Skip CockroachDB testRunUpdateOnOldChangelogTableFormat

┆Issue is synchronized with this [Jira Story](https://datical.atlassian.net/browse/LB-856) by [Unito](https://www.unito.io/learn-more)

### DIFF
--- a/.travis/databases.docker-compose.yml
+++ b/.travis/databases.docker-compose.yml
@@ -26,7 +26,7 @@ services:
       - "./postgres-init.sh:/docker-entrypoint-initdb.d/postgres-init.sh"
 
   cockroachdb:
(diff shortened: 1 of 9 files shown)


rafiss:
I'm not sure of the correct way to do testing setup, so please let me know if I need to change that or anything else.

Issue #1507, opened 2020-10-26
Regression: StackOverflowError processing large Postgres SQL changeset


goostleek:
## Environment
**Liquibase Version**: `4.1.1`

**Liquibase Integration & Version**: Spring Boot

**Database Vendor & Version**: Postgres 12.3

**Operating System Type & Version**: Ubuntu 20.04

## Description
There's a regression after upgrading from `4.1.0` to `4.1.1` that is causing `StackOverflowError` while generating Postgres statement during processing a large SQL file-based changeset.  It seems to be a regression
introduced in #912  while trying to fix [CORE-3379](https://liquibase.jira.com/browse/CORE-3379).

## Steps To Reproduce
Create a changeset:

```xml
<changeSet id="StackOverflowError" author="goostleek">
    <sqlFile path="large.sql" relativeToChangelogFile="true"/>
</changeSet>
```

The `sqlFile` is ~200kb. It's been included in [the file](https://github.com/liquibase/liquibase/blob/2bb6acad174723e770f43411fbeae2a68e2af691/liquibase-core/src/test/resources/liquibase/change/long-statement.sql) in the PR #1508 which introduces a failing unit test presenting the error.

## Actual Behavior
`StackOverflowError` is thrown during SpringBoot application startup while applying the changest.

## Expected/Desired Behavior
The changeset is executed successfully like it was before upgrading to Liquibase 4.1.1.

## Additional Context
See #912, #1508

molivasdat:
Hi @goostleek Thanks for bringing up this issue to our attention.  And thanks for recreation scenario and unit test.  It definitely helps.  We will add it to the list of issues that we are processing.

chtibob69:
Hi,
I'm experiencing it too... while updating SpringBoot from 2.3.x (liquibase-core 3.8.9) to 2.4.7 (liquibase-core 3.10.3) or 2.5.1 (liquibase-core 4.3.5)

NOTE : If I downgrade liquibase-core to 3.10.2 with springboot 2.4.7 it works fine
It seem to impact liquibase-core versions **greather than 3.10.2** ... july 2020... 

Thanks fixing it... and I hope Spring Team will publish fix quickly too

maksudsharif:
Any ETA on when this will be pushed out to repositories?

molivasdat:
Hi @maksudsharif We are moving it through getting it into a release.  I don't have an ETA yet.  In the meantime, I would suggest using the workaround that @chtibob69 mentions.

[...]

PR #1508, opened 2020-10-26
🐞 Add unit test presenting StackOverflowError while generating Postgres statement for SQL changeset

goostleek:
Add test case presenting StackOverFlow error while generating Postgres
statement for relatively large SQL changeset. This relates to #1507.

┆Issue is synchronized with this [Jira Story](https://datical.atlassian.net/browse/LB-864) by [Unito](https://www.unito.io/learn-more)


molivasdat:
Hi @goostleek 
Thanks for adding this test case.  We will take a look and process appropriately.  If you have a fix as well for this issue that would be great.

nvoxland:
Thanks, I used this test to see what line was causing the stack overflow. 

I ended up taking out that line completely so I'm not going to merge this test in for long-term testing. The problem was a result of the specific code which should not be coming back.

Issue #1483, opened 2020-10-16
Error handling long Postgres table names


jamey-clari:
## Environment
**Liquibase Version**: 4.0.0

**Liquibase Integration & Version**: Java API

**Liquibase Extension(s) & Version**: 

**Database Vendor & Version**: Postgres 10.12

**Operating System Type & Version**: Mac + Linux

## Description
If I create a simple changeset with a single createTable directive with a 63-character length tableName Liquibase fails with a relation already exists error.  Not only is this error misleading, it also is unnecessary since Postgres will not truncate table names until they are MORE THAN 63 characters.  We are looking at upgrading from Liquibase 3.5.4 to 4.0.0 and we have a programmatic table name generator that currently banks on being able to create table names up to and including 63 characters in length so this is a problem for us as we have existing 63 character tables.

## Steps To Reproduce
Create a simple changeset:

```
    <changeSet id="1" author="Me">
        <createTable tableName="my_63_char________________________________________________table">
            <column name="id" type="SERIAL"><constraints primaryKey="true"/></column>
            <column name="value" type="TEXT"><constraints nullable="false" /></column>
        </createTable>
    </changeSet>
```

Execute by some means (we do it programmatically via the Java API).

The operation will fail and the nested Exception root cause is 

```
liquibase.exception.DatabaseException: ERROR: relation "my_63_char________________________________________________table" already exists [Failed SQL: (0) CREATE TABLE public.my_63_char________________________________________________table (id INTEGER GENERATED BY DEFAULT AS IDENTITY NOT NULL, value TEXT NOT NULL, CONSTRAINT MY_63_CHAR________________________________________________TABLE_PKEY PRIMARY KEY (id))]
```

This is interesting so I executed that directly in psql and what is happening is that the table creation is fine but the name of the derived primary key is automatically truncated by Postgres to 63 characters which makes it identical to the table name (the `_PKEY` suffix is shaved off) and causes the confusing error message about the relation already existing.

For whatever reason this wasn't a problem in Liquibase 3.5.4 and, if I just execute a simple create table via psql:

`create table my_63_char________________________________________________table ( id serial primary key, value text not null );`

this works fine and Postgres creates the associated PK constraint and sequence using the names `my_63_char_________________________________________________pkey` and `my_63_char_______________________________________________id_seq`.

[...]

PR #1514, opened 2020-10-28
Issue-1483: Fixed wrongly generated PKs for PostgreSQL.

alexey-semenov:
- - -
name: Pull Request. Fixed wrongly generated PKs for PostgreSQL.
about: In case of long PostgreSQL table names (more that 63 bytes) generates PK names are truncated with '_pkey' suffixes.
title: ''
labels: Status:Discovery
assignees: ''

- - -

## Environment
**Liquibase Version**:
4.0.0 + 

## Pull Request Type

* [ ] Bug fix (non-breaking change which fixes an issue.)
* [ ] Enhancement/New feature (non-breaking change which adds functionality)

## Description

##### Full bug description is in the GH issue: [https://github.com/liquibase/liquibase/issues/1483](https://github.com/liquibase/liquibase/issues/1483)
* Describe the actual problematic behavior.
* Ensure private information is redacted.

## Steps To Reproduce

* Please be precise and ensure private information is redacted
* Include things like
  * Files used - sql scripts, changelog file(s), property file(s), config files, POM Files
  * Exact commands used - CLI, maven, gradle, spring boot, servlet, etc.

## Actual Behavior

* Include console output if relevant
* Include log files if available.

## Expected/Desired Behavior

* [ ] Build is successful and all new and existing tests pass
* [ ] Added [Integration Test(s)](https://liquibase.jira.com/wiki/spaces/LB/pages/1276608569/How+to+Write+Liquibase+Core+Integration+Tests)
* [ ] Documentation Updated

## Need Help?

┆Issue is synchronized with this [Jira Bug](https://datical.atlassian.net/browse/LB-873) by [Unito](https://www.unito.io/learn-more)
┆Fix Versions: Community 4.4.0

### DIFF
--- a/liquibase-core/src/main/java/liquibase/database/core/PostgresDatabase.java
+++ b/liquibase-core/src/main/java/liquibase/database/core/PostgresDatabase.java
@@ -33,16 +33,17 @@ public class PostgresDatabase extends AbstractJdbcDatabase {
     public static final String PRODUCT_NAME = "PostgreSQL";
     public static final int MINIMUM_DBMS_MAJOR_VERSION = 9;
     public static final int MINIMUM_DBMS_MINOR_VERSION = 2;
-    private static final int PGSQL_DEFAULT_TCP_PORT_NUMBER = 5432;
-    private static final Logger LOG = Scope.getCurrentScope().getLog(PostgresDatabase.class);
-    private static final Charset CHARSET = Scope.getCurrentScope().getFileEncoding();
 
     /**
      * Maximum length of PostgresSQL identifier.
      * For details see https://www.postgresql.org/docs/current/sql-syntax-lexical.html#SQL-SYNTAX-IDENTIFIERS.
      */
-    private static final int PGSQL_PK_BYTES_LIMIT = 63;
-    private static final String PGSQL_PK_SUFFIX = "_pkey";
+    static final int PGSQL_PK_BYTES_LIMIT = 63;
+    static final String PGSQL_PK_SUFFIX = "_pkey";
+    static final Charset CHARSET = Scope.getCurrentScope().getFileEncoding();
+
+    private static final int PGSQL_DEFAULT_TCP_PORT_NUMBER = 5432;
+    private static final Logger LOG = Scope.getCurrentScope().getLog(PostgresDatabase.class);
 
     /**
      * Represents Postgres DB types.
--- a/liquibase-core/src/main/java/liquibase/database/core/PostgresDatabase.java
+++ b/liquibase-core/src/main/java/liquibase/database/core/PostgresDatabase.java
@@ -7,18 +7,19 @@ import liquibase.database.AbstractJdbcDatabase;
 import liquibase.database.DatabaseConnection;
 import liquibase.database.ObjectQuotingStrategy;
 import liquibase.database.jvm.JdbcConnection;
-import liquibase.logging.Logger;
-import liquibase.structure.DatabaseObject;
 import liquibase.exception.DatabaseException;
+import liquibase.executor.ExecutorService;
+import liquibase.logging.Logger;
 import liquibase.statement.SqlStatement;
 import liquibase.statement.core.RawCallStatement;
+import liquibase.statement.core.RawSqlStatement;
+import liquibase.structure.DatabaseObject;
 import liquibase.structure.core.Table;
 import liquibase.util.JdbcUtils;
 import liquibase.util.StringUtil;
-import liquibase.statement.core.RawSqlStatement;
-import liquibase.executor.ExecutorService;
 
 import java.math.BigInteger;
+import java.nio.charset.StandardCharsets;
 import java.sql.ResultSet;
 import java.sql.SQLException;
 import java.sql.Statement;
@@ -34,6 +35,13 @@ public class PostgresDatabase extends AbstractJdbcDatabase {
     public static final int MINIMUM_DBMS_MINOR_VERSION = 2;
     private static final int PGSQL_DEFAULT_TCP_PORT_NUMBER = 5432;
     private static final Logger LOG = Scope.getCurrentScope().getLog(PostgresDatabase.class);
+
+    /**
+     * Maximum length of PostgresSQL identifier.
+     * For details see https://www.postgresql.org/docs/current/sql-syntax-lexical.html#SQL-SYNTAX-IDENTIFIERS.
+     */
+    private static final int PGSQL_PK_BYTES_LIMIT = 63;
+    private static final String PGSQL_PK_SUFFIX = "_pkey";
 
     /**
      * Represents Postgres DB types.
@@ -330,9 +338,22 @@ public class PostgresDatabase extends AbstractJdbcDatabase {
         return new RawCallStatement("select current_schema()");
     }
 
+    /**
+     * Generates PK following {@code PostgreSQL} conventions:
+     * <li>Postgres PK size is limited with 63 bytes.
+     * <li>Postgres PK is suffixed with '_pkey'.
+     *
+     * @param tableName Table name as the base name for the generated PK.
+     * @return PK name.
+     */
     @Override
     public String generatePrimaryKeyName(final String tableName) {
-        return tableName.toUpperCase(Locale.US) + "_PKEY";
+        final byte[] bytes = tableName.getBytes(StandardCharsets.UTF_8);
+        final int suffixBytes = PGSQL_PK_SUFFIX.length(); // All characters are ASCII symbols => 1 byte per symbol.
+
+        return bytes.length <= PGSQL_PK_BYTES_LIMIT - suffixBytes
+            ? tableName + PGSQL_PK_SUFFIX
+            : new String(bytes, 0, PGSQL_PK_BYTES_LIMIT - suffixBytes, StandardCharsets.UTF_8) + PGSQL_PK_SUFFIX;
     }
 
     @Override
--- a/liquibase-core/src/main/java/liquibase/database/core/PostgresDatabase.java
+++ b/liquibase-core/src/main/java/liquibase/database/core/PostgresDatabase.java
@@ -19,7 +19,7 @@ import liquibase.util.JdbcUtils;
 import liquibase.util.StringUtil;
 
 import java.math.BigInteger;
-import java.nio.charset.StandardCharsets;
+import java.nio.charset.Charset;
 import java.sql.ResultSet;
 import java.sql.SQLException;
 import java.sql.Statement;
@@ -35,6 +35,7 @@ public class PostgresDatabase extends AbstractJdbcDatabase {
     public static final int MINIMUM_DBMS_MINOR_VERSION = 2;
     private static final int PGSQL_DEFAULT_TCP_PORT_NUMBER = 5432;
     private static final Logger LOG = Scope.getCurrentScope().getLog(PostgresDatabase.class);
+    private static final Charset CHARSET = Scope.getCurrentScope().getFileEncoding();
 
     /**
      * Maximum length of PostgresSQL identifier.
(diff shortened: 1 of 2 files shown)


liquibase-core/src/main/java/liquibase/database/core/PostgresDatabase.java
```diff
@@ -330,9 +338,22 @@ protected SqlStatement getConnectionSchemaNameCallStatement() {
...
+     * <li>Postgres PK is suffixed with '_pkey'.
+     *
+     * @param tableName Table name as the base name for the generated PK.
+     * @return PK name.
+     */
     @Override
     public String generatePrimaryKeyName(final String tableName) {
-        return tableName.toUpperCase(Locale.US) + "_PKEY";
+        final byte[] bytes = tableName.getBytes(StandardCharsets.UTF_8);
+        final int suffixBytes = PGSQL_PK_SUFFIX.length(); // All characters are ASCII symbols => 1 byte per symbol.
```
jamey-clari, line 352: If the assumption here is to use UTF-8 then that is fine but I would still probably define a private method parameterized by Charset that is driven by that so any future enhancement to make the code gracefully handle other charsets is done easily and with little opportunity for error.

jamey-clari: If not then I'd just declare a static final byte[] for the UTF-8 encoded "_pkey" suffix and use that here.

liquibase-core/src/main/java/liquibase/database/core/PostgresDatabase.java
```diff
@@ -330,9 +338,22 @@ protected SqlStatement getConnectionSchemaNameCallStatement() {
...
+
+        return bytes.length <= PGSQL_PK_BYTES_LIMIT - suffixBytes
```
jamey-clari, line 354: I'd declare a variable for the max base name length here:

`final int maxBaseLen = PGSQL_PK_BYTES_LIMIT - suffixBytes;`

And use it in the two operations below.

liquibase-core/src/main/java/liquibase/database/core/PostgresDatabase.java
```diff
@@ -330,9 +338,22 @@ protected SqlStatement getConnectionSchemaNameCallStatement() {
...
+            ? tableName + PGSQL_PK_SUFFIX
+            : new String(bytes, 0, PGSQL_PK_BYTES_LIMIT - suffixBytes, StandardCharsets.UTF_8) + PGSQL_PK_SUFFIX;
```
jamey-clari, line 356: Ah, definitely cleaner than my proposed hack at this in the JIRA - nice!

liquibase-core/src/test/java/liquibase/database/core/PostgresDatabaseTest.java
```diff
@@ -123,4 +126,33 @@ public void test_getConcatSql() {
...
+
+        // Act
+        final String pk = this.database.generatePrimaryKeyName(tableName);
+
+        // Assert
+        assertEquals(pk, expectedPrimaryKeyName);
+    }
+
+    @Test
+    public void generatePrimaryKeyName_tableSizeNameMoreThan63Bytes_nameIsBuiltCorrectly() {
```
jamey-clari, line 144: I think it's worth adding some testing for non-ASCII characters.

jamey-clari (commented):
@alexey-semenov - thanks so much for putting this PR together.  A couple of minor suggestions in my review.  Also, are there any plans for applying this bug-fix to versions of Liquibase prior to 4.x?  The buggy naming change was made in version 3.6 and, speaking selfishly, I don't know if I will end up upgrading us to 4.x just yet or an intermediate 3.x version.

liquibase-core/src/main/java/liquibase/database/core/PostgresDatabase.java
molivasdat: Good suggestions.

liquibase-core/src/main/java/liquibase/database/core/PostgresDatabase.java
molivasdat: Not required, but still a good suggestion.

liquibase-core/src/test/java/liquibase/database/core/PostgresDatabaseTest.java
molivasdat: Some non-ASCII characters in tests would be good.

liquibase-core/src/main/java/liquibase/database/core/PostgresDatabase.java
```diff
@@ -34,6 +35,14 @@
     public static final int MINIMUM_DBMS_MINOR_VERSION = 2;
     private static final int PGSQL_DEFAULT_TCP_PORT_NUMBER = 5432;
     private static final Logger LOG = Scope.getCurrentScope().getLog(PostgresDatabase.class);
+    private static final Charset CHARSET = Scope.getCurrentScope().getFileEncoding();
+
+    /**
+     * Maximum length of PostgresSQL identifier.
+     * For details see https://www.postgresql.org/docs/current/sql-syntax-lexical.html#SQL-SYNTAX-IDENTIFIERS.
+     */
+    private static final int PGSQL_PK_BYTES_LIMIT = 63;
```
jamey-clari, line 44: make this package private so you can use them in your unit test

liquibase-core/src/test/java/liquibase/database/core/PostgresDatabaseTest.java
```diff
@@ -1,20 +1,27 @@
...
+import java.nio.charset.Charset;
+
 import static org.junit.Assert.*;
 
 /**
  * Tests for {@link PostgresDatabase}
  */
 public class PostgresDatabaseTest extends AbstractJdbcDatabaseTest {
 
+    private static final long PGSQL_PK_BYTES_LIMIT = 63;
```
jamey-clari, line 22: Use the values from `PostgresDatabase` here to avoid the potential for inconsistencies over time.

liquibase-core/src/test/java/liquibase/database/core/PostgresDatabaseTest.java
```diff
@@ -123,4 +131,64 @@ public void test_getConcatSql() {
         assertEquals("one || two || | three", database.getConcatSql("one", "two", "| three"));
     }
 
+    @Test
+    public void generatePrimaryKeyName_tableSizeNameLessThan63Bytes_nameIsBuiltCorrectly() {
```
jamey-clari, line 135: Let's create a private helper method:

```
    @Test
    public void assertPrimaryKeyName(final String tableName, final String expectedPrimaryKeyName) {
        // Act
        final String pk = this.database.generatePrimaryKeyName(tableName);

        // Assert
        assertEquals(expectedPrimaryKeyName, pk);
        assertTrue(expectedPrimaryKeyName.getBytes(CHARSET).length <= PGSQL_PK_BYTES_LIMIT);
    }
```

Now you can call this from all of your tests.

liquibase-core/src/main/java/liquibase/database/core/PostgresDatabase.java
```diff
@@ -34,6 +35,14 @@
     public static final int MINIMUM_DBMS_MINOR_VERSION = 2;
     private static final int PGSQL_DEFAULT_TCP_PORT_NUMBER = 5432;
     private static final Logger LOG = Scope.getCurrentScope().getLog(PostgresDatabase.class);
+    private static final Charset CHARSET = Scope.getCurrentScope().getFileEncoding();
```
jamey-clari, line 38: make this package private so you can use them in your unit test

jamey-clari (commented):
Thanks for the updates @alexey-semenov!  A couple of more minor things and I'll shut up:)

liquibase-core/src/test/java/liquibase/database/core/PostgresDatabaseTest.java
alexey-semenov: Basically agree except `assertPrimaryKeyName` should accept actual PK instead of table name.
That's clearer to pass both actual and expected names instead of table name.
And it's clearly shows the tested method call inside the test. 
I'd vote for explicitness instead of shortness here.

jamey-clari (approved):

jamey-clari (approved):

jamey-clari:
> ➤ Erzsebet Carmean commented:
> 
> PRIVATE COMMENT I cannot size this ticket without addition information on exactly what the bug is and how to reproduce it. SQL to set up objects and the changelog that results are also useful to include. Can you follow up on this, Mike Olivas?
> 
> CC :: Mario Champion, Kristyl Gomes

I feel like this was pretty well articulated in the original issue filed: https://github.com/liquibase/liquibase/issues/1483

I'm happy to help provide more detail if there is still confusion as to the nature of the bug and how to reproduce.

Thanks to all for working to resolve it!

molivasdat:
Hi @alexey-semenov One issue we found in our testing was that the primary key generated from this code is 62 bytes and not 63 bytes.  Everything else worked as described.  Are you good with that for us to include or do you want to make additional changes to have the PK 63 bytes in this scenario?

alexey-semenov:
Hi @molivasdat ! Please take a look https://github.com/liquibase/liquibase/pull/1514/files#diff-251a26e86b87566e522b403e6719fcd0b16e1408ff91029c37ad37eabcfc80d0R360-R363
That's probably an answer to your question.

molivasdat:
Thanks @alexey-semenov .  We did see that in code review and it does that as a safety net just in case someone is using double byte characters.  Which basically means that all PKs for tables that are at least 63 bytes long will have 62 byte primary key names.  As long as you are ok with that as the truncation routine for all PKs with tables that have 63 bytes or longer names, we are as well.

jamey-clari:
Many thanks to both @alexey-semenov and @molivasdat for driving this fix forward.  Since we are effectively emulating PG behavior here, have we attempted to test a >63-byte multi-byte character name with PG and observed the truncation behavior?

jamey-clari:
Testing the behavior of PG 10 I see the following:

Truncate behavior for >63 byte table name, note table name truncates to 62 bytes and PK similarly nets to 62 bytes due to truncation of 2-byte trailing characters:
```
clarius_default_test=# create table "table_ΩΩΩΩΩΩΩΩΩΩΩΩΩΩΩΩΩΩΩΩΩΩΩΩΩΩΩΩΩΩΩΩΩΩΩΩΩΩΩΩΩΩΩΩ" ( id SERIAL PRIMARY KEY, value TEXT );
NOTICE:  identifier "table_ΩΩΩΩΩΩΩΩΩΩΩΩΩΩΩΩΩΩΩΩΩΩΩΩΩΩΩΩΩΩΩΩΩΩΩΩΩΩΩΩΩΩΩΩ" will be truncated to "table_ΩΩΩΩΩΩΩΩΩΩΩΩΩΩΩΩΩΩΩΩΩΩΩΩΩΩΩΩ"
CREATE TABLE
clarius_default_test=# \d+ "table_ΩΩΩΩΩΩΩΩΩΩΩΩΩΩΩΩΩΩΩΩΩΩΩΩΩΩΩΩ"
                                                Table "public.table_ΩΩΩΩΩΩΩΩΩΩΩΩΩΩΩΩΩΩΩΩΩΩΩΩΩΩΩΩ"
 Column |  Type   | Collation | Nullable |                            Default                            | Storage  | Stats target | Description 
--------+---------+-----------+----------+---------------------------------------------------------------+----------+--------------+-------------
 id     | integer |           | not null | nextval('"table_ΩΩΩΩΩΩΩΩΩΩΩΩΩΩΩΩΩΩΩΩΩΩΩΩΩ_id_seq"'::regclass) | plain    |              | 
 value  | text    |           |          |                                                               | extended |              | 
Indexes:
    "table_ΩΩΩΩΩΩΩΩΩΩΩΩΩΩΩΩΩΩΩΩΩΩΩΩΩΩ_pkey" PRIMARY KEY, btree (id)
```

Truncate behavior for 63 byte table name, note PK nets to 62 bytes due to truncation of 2-byte trailing characters:
```
clarius_default_test=# create table "table__ΩΩΩΩΩΩΩΩΩΩΩΩΩΩΩΩΩΩΩΩΩΩΩΩΩΩΩΩ" ( id SERIAL PRIMARY KEY, value TEXT );
CREATE TABLE
clarius_default_test=# \d+ "table__ΩΩΩΩΩΩΩΩΩΩΩΩΩΩΩΩΩΩΩΩΩΩΩΩΩΩΩΩ"
                                               Table "public.table__ΩΩΩΩΩΩΩΩΩΩΩΩΩΩΩΩΩΩΩΩΩΩΩΩΩΩΩΩ"
 Column |  Type   | Collation | Nullable |                            Default                            | Storage  | Stats target | Description 
--------+---------+-----------+----------+---------------------------------------------------------------+----------+--------------+-------------
 id     | integer |           | not null | nextval('"table__ΩΩΩΩΩΩΩΩΩΩΩΩΩΩΩΩΩΩΩΩΩΩΩΩ_id_seq"'::regclass) | plain    |              | 
 value  | text    |           |          |                                                               | extended |              | 
Indexes:
    "table__ΩΩΩΩΩΩΩΩΩΩΩΩΩΩΩΩΩΩΩΩΩΩΩΩΩ_pkey" PRIMARY KEY, btree (id)
```

nvoxland:
Merged for 4.4.0. Thanks everyone!

jamey-clari:
Oooh, this makes me very happy, thanks Nathan, Alexey, Mike, et al!  Now I think we can upgrade our Liquibase dependency to 4.x!

alexey-semenov:
Nice to hear guys! Thanks to all for assistance! Keep moving forward!

PR #1548, opened 2020-11-18
Fixed UniqueConstraintSnapshotGenerator

### LINKED ISSUE #1477: Liquibase 3.10+/4.1+ missing UniqueConstraints in snapshot

nvoxland:
Reverted a previous commit that assumed all databases support bulk-querying constraint columns. Then, incorporated changes from #1479 to increase the number of databases that do support them.

Fixes #1477

┆Issue is synchronized with this [Jira Story](https://datical.atlassian.net/browse/LB-952) by [Unito](https://www.unito.io/learn-more)

### DIFF
--- a/liquibase-core/src/main/java/liquibase/snapshot/jvm/UniqueConstraintSnapshotGenerator.java
+++ b/liquibase-core/src/main/java/liquibase/snapshot/jvm/UniqueConstraintSnapshotGenerator.java
@@ -155,47 +155,46 @@ public class UniqueConstraintSnapshotGenerator extends JdbcSnapshotGenerator {
         }
 
         if (columnCache == null) {
+            bulkQuery = false;
+            if (columnQueryCount > 3) {
+                bulkQuery = supportsBulkQuery(database);
+            }
 
-            bulkQuery = ((database instanceof OracleDatabase) || (database instanceof MSSQLDatabase)) &&
-                (columnQueryCount > 3);
             snapshot.setScratchData(queryCountKey, columnQueryCount + 1);
 
             if ((database instanceof MySQLDatabase) || (database instanceof HsqlDatabase)) {
-                sql = "select const.CONSTRAINT_NAME, COLUMN_NAME "
+                sql = "select const.CONSTRAINT_NAME, COLUMN_NAME, const.constraint_schema as CONSTRAINT_CONTAINER "
                         + "from " + database.getSystemSchema() + ".table_constraints const "
                         + "join " + database.getSystemSchema() + ".key_column_usage col "
                         + "on const.constraint_schema=col.constraint_schema "
                         + "and const.table_name=col.table_name "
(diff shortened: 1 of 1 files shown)


wwillard7800 (approved):
Looks good

PR #1575, opened 2020-12-07
Add standard file writing support to Liquibase

nvoxland:
Currently, Liquibase has the ResourceAccessor which is designed around providing a standard way to READ files, regardless of the integration but no way to WRITE files.

The write support that we have is custom per way we try to write the files, and doesn't have any opportunity take into account differences between integrations. For example, the CLI may want to write files into the current working directory, but Maven may want to write files in src/main/resources.

### DIFF
--- a/liquibase-core/src/main/java/liquibase/Liquibase.java
+++ b/liquibase-core/src/main/java/liquibase/Liquibase.java
@@ -2125,7 +2125,7 @@ public class Liquibase implements AutoCloseable {
                     DBDocVisitor visitor = new DBDocVisitor(database);
                     logIterator.run(visitor, new RuntimeEnvironment(database, contexts, labelExpression));
 
-                    visitor.writeHTML(new File(outputDirectory), resourceAccessor);
+                    visitor.writeHTML(outputDirectory, resourceAccessor);
                 } catch (IOException e) {
                     throw new LiquibaseException(e);
                 } finally {
--- a/liquibase-core/src/main/java/liquibase/Scope.java
+++ b/liquibase-core/src/main/java/liquibase/Scope.java
@@ -11,7 +11,9 @@ import liquibase.logging.Logger;
 import liquibase.logging.core.JavaLogService;
 import liquibase.logging.core.LogServiceFactory;
 import liquibase.resource.ClassLoaderResourceAccessor;
+import liquibase.resource.FileSystemResouceWriter;
 import liquibase.resource.ResourceAccessor;
+import liquibase.resource.ResourceWriter;
 import liquibase.servicelocator.ServiceLocator;
 import liquibase.servicelocator.StandardServiceLocator;
 import liquibase.ui.ConsoleUIService;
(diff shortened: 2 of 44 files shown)


nvoxland:
When I was first thinking through this, it seemed to make sense to add writing support directly into the ResourceAccessor, which the above commit does.

It adds the following methods to the ResourceAccessor interface:
- `OutputStream openOutputStream(String relativeTo, String path, boolean append) throws IOException;`
- `boolean exists(String relativeTo, String path);`

and updates the existing ResourceAccessor implementatations to provide those as they can. It also updates the Diff/GenerateChangeLog logic to use that outputStream rather than create a custom FileOutputStream. 

This allowed me to have the MavenResourceAccessor write files to src/main/resources even though it reads them from target/classes which fixes one issue we had with maven writing files to target which doesn't work like one would like.

The fix does not handle all cases of file usage in liquibase. It does not cover file writing for:
- OfflineHistoryService
- DBDoc
- MissingDataExternalFileChangeGenerator
- SpringLiquibase

Having worked though it, though, I'm not sure I'm happy with the change.

It does break the ResourceAccessor interface. I do add default implementations to AbstractResourceAccessor, and we are OK with breaking interfaces as we must in the 4.x versions so I'm not super worried about it.

My bigger change of heart is from seeing the interaction/differences between reading and writing files. I knew there was the bug in MavenResourceAccessor where we have the "same" file we want to read from target/classes but write to src/main/resources, so I was thinking a lot in terms of "ResourceAccessor decides how we read files, it should also decide how to write those same files". But, there is maybe more of a disconnect than I was thinking at first.

1. At the generic level, it's hard for the ResourceAccesssor to know where to write. If the ClassLoaderResourceAccessor has 20 jars and 5 directories in which to read, it could write to any of those 5 directories but from a user's standpoint, it would seem random which one actually gets written to

2. Things like DBDoc need much more than just "write a file at path X"

So, I'm starting to think the file write API should be pulled out to a completely separate interface like `ResourceWriter` that is separately managed on the Scope from ResourceAccessor. We can ship a `FileResourceWriter` which can be given a single directory base to write to, but it is up to each integration to correctly set up the ResourceWriter they should use. I'd even say the default is a NotSupportedResourceAccessor that throws an error vs. a FileResourceWriter that writes to the current working directory since I'm not sure that is something to be relied on.

The new ResourceWriter interface can be a more complex interface as well that supports more operations so we can move DBDoc over to it as well.

Thoughts?

nvoxland:
So, I split out the ResourceWriter interface, but now the problem is that it is a little too separate from the ResourceAccessor :) 

It works well from the standpoint of being able to write files however we want, but when we want to update a file that is directly in the ResourceAccessor it is too separate. 

Example of the problem: when you run registerChangeLog from the CLI you want to update the actual changelog file, but when you run registerChangeLog from maven, you want to update the file in src/main/resources instead.

I used the java.nio.file.Path interface for exposing more direct file interaciton which seemed to work well. I think I just need to bring that back into ResourceAccessor with a separate setting for where to write unknown files

kataggart:
@nvoxland pretty sure you recently added this functionality as part of the ResourceAccessor improvements?

nvoxland:
Yeah, this ended up being implemented as #3064 instead so I'll close this PR.

Thanks for finding it.

PR #1589, opened 2020-12-15
Do not split SQL on delimiters within BEGIN/END blocks

### LINKED ISSUE #1553: Do not split SQL on delimiters within BEGIN/END blocks
## Environment
**Liquibase Version**:  4.2.0

**Liquibase Integration & Version**: All

## Description
Currently the Sql split logic used throughout liquibase splits on delimiters. That works fine, as long as the delimiter characters are not used inside statements. 

The main time when delimiters can be in statements are in stored procedures such as:

```
CREATE PROCEDURE country_hos(IN con CHAR(20))
BEGIN
  SELECT Name, HeadOfState FROM Country
  WHERE Continent = con;
END
```

were the ; after `WHERE Continent = con` makes liquibase split that one statement into two.

This causes the SQL to fail with `unexpected end of statement` type errors.

## Change
Most databases use a BEGIN / END keyword pair to delimit the points where we should ignore delimiters, so we should add logic to check for those tokens and NOT split on the delimiters inside them. This will make the SQL parsing significantly smarter.

A common split function is used anywhere we are splitting sql:

* In formatted sql by default
* In <sql> tags with splitStatements=true
* In the executeSql command

## Requirements
* Formatted SQL plus <sql> and <sqlFile> based SQL with `splitStatements=true` now correctly run stored procedure, function, trigger, etc. definitions with ;'s in the middle of it correctly as long as the statement uses BEGIN/END syntax. Even without changing the endDelimiter setting.

┆Issue is synchronized with this [Jira Bug](https://datical.atlassian.net/browse/LB-958) by [Unito](https://www.unito.io)

nvoxland:
## Description

Currently we apply the endDelimiter logic in parsing sql throughout the SQL. But stored logic, views, and other SQL will often have otherwise-delimiters inside their "body" which is delineated with BEGIN/END statements.

This updates the parsing logic to not look for delimiters while inside a BEGIN/END statement, thereby removing a lot of the need to specify "endDelimiter" in the first place. 

NOTE: It will not count `BEGIN/END TRANSACTION` blocks as "begin/end" blocks

Fixes LB-958
Fixes #1553

## Things to be aware of:
- Added unit tests for the new types of sql to look for

## Things to worry about:
- Nothing


molivasdat:
@nvoxland Thanks for making the updates for nested BEGIN END statements.  At least one more use case to consider that I missed earlier.  What about BEGIN TRANS or BEGIN TRANSACTION statements?
[https://www.mssqltips.com/sqlservertutorial/3305/what-does-begin-tran-rollback-tran-and-commit-tran-mean/](https://www.mssqltips.com/sqlservertutorial/3305/what-does-begin-tran-rollback-tran-and-commit-tran-mean/)
Not sure if folks still use them or those types of SQL statements will flow through this changed code.

nvoxland:
Good point on `Begin/end transaction` statements, @molivasdat . It doesn't seem like something people should generally do, but I added a check for them in case.

XDelphiGrl:
@nvoxland, thank you for this improvement! Stored logic delimiters are one of the most frustrating bits of changeset configuration to get right. I think this will be helpful.

XDelphiGrl (approved):
* Enhancement improves handling of stored logic SQL that contains internally-delimited statements wrapped in BEGIN...END. 
  * Liquibase no longer attempts to parse the statements within BEGIN...END keywords using the endDelimiter defined on the changeset. Statements with in BEGIN...END are taken as is for execution against the database.
* New unit tests added to validate several BEGIN...END parsing use cases.

APPROVED

[Passing Functional Tests](https://github.com/liquibase/liquibase-pro-tests/actions/runs/2749214665)
[Passing Test Harness Execution](https://github.com/liquibase/liquibase-pro-tests/actions/runs/2749214665)

suryaaki2 (approved):

kataggart:
@nvoxland Do we have docs that need to be updated/improved with this change? Thanks!

dozsa:
This created a regression for files containing _begin_ and _commit_ . Those now fail as their statements are not properly separated anymore. And unfortunately the way this was implemented it is not possible to know the closing statement for a begin statement, to only apply for begin-end and not other closing statements.

PR #1622, opened 2021-01-06
Filename + Id + Author is not being treated as unique 

### LINKED ISSUE #1603: Filename + Id + Author is not being treated as unique in ValidatingVisitor.findChangeSet  method
In version 3.1.1 in the ValidatingVisitor.java class the visit method correctly finds the right changeset matching an already executed one in the databasechangelog table with the one that its validating against from the xml file.  However after version 3.4.0 the new findChangeSet method in this class now uses endsWith to see if the files are the same and that is wrong.  For example I have these files for last 5+ years running in 3.1.1:

esi-olap.xml
olap.xml

These are 2 different files.  They happen to have changesets that have the same id and author in each but since they are in different filenames they aren't found to be the same using 3.1.1 and everything works.  But now the new findChangeSet method is:

```
  private RanChangeSet findChangeSet(ChangeSet changeSet) {
    RanChangeSet result = (RanChangeSet)this.ranIndex.get(changeSet.toString(false));
    if (result == null) {
      for (RanChangeSet ranChangeSet : this.ranIndex.values()) {
        if ((ranChangeSet.getId().equalsIgnoreCase(changeSet.getId())) && 
          (ranChangeSet.getAuthor().equalsIgnoreCase(changeSet.getAuthor()))) {
          String changeSetPath = normalizePath(changeSet.getFilePath());
          String ranChangeSetPath = normalizePath(ranChangeSet.getChangeLog());
          if ((ranChangeSetPath.equalsIgnoreCase(changeSetPath)) || (ranChangeSetPath.endsWith(changeSetPath)) || (changeSetPath.endsWith(ranChangeSetPath)))
          {
            result = ranChangeSet;
          }
        }
      }
    }

    return result;
  }
```

And the 2 endsWith checks causes olap.xml and esi-olap.xml to result in it thinking they are the same file and returning a changeset that isn't the same as the one in the other file and then the checksum validation fails.  So it ends up comparing the changeset in esi-olap.xml with a changeset with the same id + author in the olap.xml.

Why would this method use endsWith against the filenames?  These 2 files are not the same and I can't think of a reason why it would be done this way.  Using the above endsWith logic I could have these 2 files that it would think are the same;

ThisSchemaFileHasARealLongName.xml
e.xml

It would think the above 2 files are the same.

Note according to documentation - https://docs.liquibase.com/concepts/basic/databasechangelog-table.html

> The table tracks each changeset as a row, identified by a combination of the “id”, “author”, and “filename” columns.
and

> The composite of “id”, “author”, and “filename” is unique across all rows of the table.

But the code in findChangeSet breaks the documentation above with id + author + filename as being the unique identifier.

Thanks,
Mike

nvoxland:
ValidatingVisitor.findChangeSet method

Fixes #1603

┆Issue is synchronized with this [Jira Story](https://datical.atlassian.net/browse/LB-1101) by [Unito](https://www.unito.io/learn-more)

### DIFF
--- a/liquibase-core/src/main/java/liquibase/changelog/visitor/ValidatingVisitor.java
+++ b/liquibase-core/src/main/java/liquibase/changelog/visitor/ValidatingVisitor.java
@@ -72,15 +72,9 @@ public class ValidatingVisitor implements ChangeSetVisitor {
         RanChangeSet result = ranIndex.get(changeSet.toString(false));
         if (result == null) {
             for (RanChangeSet ranChangeSet : ranIndex.values()) {
-                if (ranChangeSet.getId().equalsIgnoreCase(changeSet.getId())) {
-                    if (ranChangeSet.getAuthor().equalsIgnoreCase(changeSet.getAuthor())) {
-                        String changeSetPath = DatabaseChangeLog.normalizePath(changeSet.getFilePath());
-                        String ranChangeSetPath = DatabaseChangeLog.normalizePath(ranChangeSet.getChangeLog());
-                        if (ranChangeSetPath.equalsIgnoreCase(changeSetPath)
-                            || ranChangeSetPath.endsWith(changeSetPath) || changeSetPath.endsWith(ranChangeSetPath)) {
-                            result = ranChangeSet;
-                        }
-                    }
+                if (ranChangeSet.isSameAs(changeSet)) {
+                    result = ranChangeSet;
+                    break;
                 }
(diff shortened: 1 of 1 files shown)


molivasdat (approved):
This looks like it's doing the same check as before in a different order and in a different way.  May switch the order so that path is the last check done in isSameAs because if the other two are not the same, you will not take the normalize path hit everytime.

PR #1685, opened 2021-02-11
Handle situation where HubService did not load

### LINKED ISSUE #1682: Liquibase 4.3.0 throws NullPointerExceptions without snakeyaml in the classpath
## Description
Snakeyaml is the 3rd party library we use for JSON and YAML parsing. 

It has always been an optional component, traditionally only needed for json/yaml changelogs and offline snapshots (which are json-formatted).

As we are adding hub.liquibase.com based functionality, that also requires snakeyaml for the json rest API calls.

With 4.3.0, there is an issue where the "get the HubService" implementation call returns null if we cannot instantiate the standard hubservice class, and then the code that expects a service to be returned throws the null pointer exception.

## Impact
This should NOT impact:

* CLI users (because snakeyaml is included in the CLI distribution)
* Liquibase-Maven-Plugin users, because Maven appears to include snakeyaml in the liquibase classpath with the `<optional>true</optional>` setting we have in the plugin's pom.xml
* Any spring boot or other embedded users, where snakeyaml is not explicitly included as a dependency

This SHOULD impact:

* Any spring boot or other embedded setups, where snakeyaml is not explicitly included as a dependency. We have it listed as <optional>true</option> in our pom.xml, so it should not be automatically included
* Ant users
* Anyone else running liquibase without snakeyaml.jar in their classpath.

## Resolution Options
1. We can make snakeyaml a required dependency. As we continue to add hub-based functionality, the ability to use json will become more and more central that it's not worth keeping it as an optional dependency.
1. Continue to keep snakeyaml optional, and improve test coverage around how to handle cases where the library is not available

NOTE: Using a different library than snakeyaml for json parsing is a different issue.

## Liquibase Internal QA Test Requirements
* Verify Liquibase CLI update does not throw a null pointer if snakeyaml is not in path.
* Verify Liquibase Ant update does not throw a null pointer if snakeyaml is not in the path.
* Verify Liquibase Servlet update does not throw a null pointer if snakeyaml is not in the path.
* Verify Liquibase Maven update does not throw a null pointer.
  * Fix should have no impact on Maven for reasons described in the Impact section.

* Verify all automated functional CLI tests pass on Jenkins.
##### Jenkins is an internal build server and results are not visible to the wider community.

┆Issue is synchronized with this [Jira Bug](https://datical.atlassian.net/browse/LB-1212) by [Unito](https://www.unito.io/learn-more)
┆Fix Versions: Community 4.3.1,Liquibase 4.3.1

nvoxland:
Fixes #1682

┆Issue is synchronized with this [Jira Story](https://datical.atlassian.net/browse/LB-1215) by [Unito](https://www.unito.io/learn-more)

### DIFF
--- a/liquibase-core/src/main/java/liquibase/Liquibase.java
+++ b/liquibase-core/src/main/java/liquibase/Liquibase.java
@@ -312,6 +312,15 @@ public class Liquibase implements AutoCloseable {
         if (executor instanceof LoggingExecutor) {
             return null;
         }
+
+        //
+        // If we do not have a Hub service then return
+        //
+        final HubService hubService = Scope.getCurrentScope().getSingleton(HubServiceFactory.class).getService();
+        if (hubService == null) {
+            return null;
+        }
+
         HubConfiguration hubConfiguration = LiquibaseConfiguration.getInstance().getConfiguration(HubConfiguration.class);
         String changeLogId = changeLog.getChangeLogId();
         HubUpdater hubUpdater = new HubUpdater(new Date(), changeLog, database);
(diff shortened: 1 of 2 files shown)


nvoxland:
Closing this in favor of #1686

Issue #1701, opened 2021-02-15
Refactor & Finalize LiquibaseCommand API


nvoxland:
## Description

Currently, liquibase.Liquibase is the primary "porcelain" façade for working with Liquibase (https://git-scm.com/book/en/v2/Git-Internals-Plumbing-and-Porcelain). However, as Liquibase's functionality has grown over the years, we have outgrown a single facade class.

A few years ago, we started the liquibase.command package as the replacement for the liquibase.Liquibase class and have been experimenting using it for a few commands. Now is the time to take what we have learned to finalize the design and apply it to all our existing commands.

## Backwards Compatibility

The liquibase.Liquibase facade methods should be backwards compatible, even as their implementations are replaced with calls to the new LiquibaseCommand classes.

The calls to the existing LiquibaseCommand classes will not necessarily remain backwards.

## Requirements / Changes

NOTE: This section will evolve as we do the actual implementation and see the actual API changes in action. Follow along with the PR for the full details.

- LiquibaseCommands can be grouped by zero or more nested parent names
- LiquibaseCommand objects are pluggable by command group + name
- Method to execute a command will be moved from LiquibaseCommand to CommandFactory
- LiquibaseCommand instances provide metadata of name, description, available parameters, required parameters, etc. which can be used to populate CLI etc. help.
- LiquibaseCommand parameters are exposed as getter/setters on the command that can be set programatically
- Unconfigured LiquibaseCommand parameters will be set with values the LiquibaseConfiguration system as part of the execute() function
- LiquibaseCommands return an object containing output data, allowing commands to call others programmatically.

## Questions

**Replace or enhance?**

Is the simple "replacement" strategy above what we want? 

An extension can replace the standard "update" command with a different version that either enhances or replaces the functionality, but only one extension can do that. The extension can create instances of the UpdateCommand class and call it if they like, but they don't have to and how they do that is completely up to them. 

An alternative would be to not replace the commands but to more register as a "chain" like ServletFllters or register pre/post hooks. That would allow an extension to not replace the standard UpdateCommand logic, but to surround it with additional logic.

nvoxland:
**Replace or Enhance discussion**

There have been a couple requests for a hook-style/enhance approach vs. the replace one.

My main worry is that it adds more complexity than we need. 

More complex:
- Documentation etc. is more difficult, since it's not a simple "command X is class com.example.X, so help outputs its description of what it does"

## Thoughts:

[...]

PR #1708, opened 2021-02-18
Command Framework : Hooks Edition

nvoxland:
## Overview 
This is an an attempt to refactor #1701 using a pattern adding logic to a command rather than using a single, replaceable command class.

### Patterns:
1. There is no longer a 1:1 correspondence between the commands to run (`liquibase update`) and a LiquibaseCommand (UpdateCommand). Instead, LiquibaseCommand register commands they are a part of and give an order to run in, and Liquibase will execute the logic in all LiquibaseCommand instances that have registered for the command string asked.
2. LiquibaseCommand objects no longer have getter/setter fields on them for properties, but instead define CommandArgumentDefinitions which can be used to type-safely get and set the stored string/object argument values
3. There is a new CommandScope class (no subclasses) which is what the code trying to call a command uses. They construct that object with the name of the command, add argument values to it, pass it to CommandFactory.execute, and get the results off it. 

### Questions (see below):
1. Thoughts so far?
1. Do we like the "crosscut" pattern for making database connections?
1. How do we manage output?

┆Issue is synchronized with this [Jira Story](https://datical.atlassian.net/browse/LB-1260) by [Unito](https://www.unito.io/learn-more)


nvoxland:
**Thinking out loud: Do we manage database connections from "cross cutting" command logic?**

With this "hook" pattern, one thing I've been considering is being able to have a LiquibaseCommand instance not be tied to a particular command name, but handle a variety of them. For example, someone could make an AuditCommand that adds itself to the command chain for every command and sends a record of what was executed off somewhere. That allows cross-cutting logic to be defined in one place and add itself to all the commands, without the other commands needing to know or care about that particular logic. 

That cross-cutting support is one of the bigger advantages I see to this "hook" style. It's not in the ticket's "patterns" section yet, since it's not yet implement.

So where I'm at now is I have the HistoryCommand, which takes a "database" argument. Partly that is because that is how the existing code works, but I think that is also the pattern we'd want to follow: command like HistoryService shouldn't be in charge of making the database connections, they should just focus on their logic. Keeps away the duplication, and also because commands will be composed together, like `updateCheckingRollback` may call the update and rollback commands and we don't want to be re-opening connections.

NOTE: I have some ideas for a new ConnectionManager class vs. the Database we have now, but am keeping the discussion with what we have for code now vs. adding unrelated changes...

**Option 1: Cross-Cut**

I could leverage the cross-cutting support to make some sort of a `OpenConnectionCrosscutCommand` which takes arguments of "url, username, password" and creates a database parameter and adds it to the commandScope for the later commands to use. And if the database parameters is already on the commandScope, it would be a no-op. I have some ideas on framework sugar to make this sort of cross-cut easy to write. 

The advantage to this is that it would make the command metadata correct. Something like "liquibase update" would show arguments of `url`, `username`, and `password` with `datatabase` marked as "taken care of". For integrations like the CLI which dynamically expose those arguments and need to have something make the database connections it works well. For integrations like Spring which already has the database connection and doesn't need the url, username, password then the metadata will look a bit off.

**Option 2: Integration managed**

Alternately, we could have the connection management hard-coded into the integration. Spring would get its connection object passed to it, CLI and maven would add url/username/password arguments. But, the fact that the CLI needs url/username/password would be in CLI-specific code. Same with maven. The Commands themselves would just say they need a database parameter.

The issue will be that for dyanamic integrations like the CLI, it needs to know what commands need a url/username/password argument and which do not and build up the configuration accordingly. Looking for a `database` argument on the command could be one way to do that, but is that the best approach? We'd also have to figure out how to handle commands that take multiple connections like diff which take a `database` and a `referenceDatabase`. Is it just "find all Database arguments, and make url/username/password arguments prefixed with the same prefix on 'database' " ?

**Initial Thoughts**

While I like the cross-cutting option's ability to leverage the command framework and a more common path, I'm thinking that the fact that the connection management is very much an integration-specific thing makes that code really belong in the integration layer. 

For security reasons, it's also probably best to continue to not save the passwords on the commandScope where other parts of the code can grab it.

Thoughts?

mcred:
I like the idea of being able to create a database connection and make it available to the commandScope. In the titan works, databases are running in containers and the titan context knows how to connect to them. For example, some pseudo workflow like:
```
liquibase update --sourceDatabase=example

beforeUpdate:
Get connection information for `example` database and construct a connection object that satisfies the requirements for update.
Prepare connection with any additional logic as required by the extension.

runUpdate:
runs update as expected based on connection

afterUpdate:
Inspect or Clean up as needed by the exension. 

```
The above before/after hook example is not titan specific, but a generic approach to satisfy the requirements of whichever command is being extended. As long as the commands exposed their requirements and the extension satisfies them the before and after logic can be very flexible.

nvoxland:
**Thinking out loud: How do we handle output?**

In the previous version, we had a CommandResponse which contained a `message` field which was supposed to be what gets sent back to the user. It also could contain other data that could be used programmatically when the command is being used in embedded or composed ways. For example, the SnapshotCommand would include the snapshot object on the CommandResponse.

What I was finding with that old style was that:
1. The commandResult.successful was too easy to forget to check (so moved failures to an exception)
2. The commandResult.message was too easy to forget to print out
3. Not all commands have a string-based output
4. Printing the consolidated message at the end of the command makes the interleaving with log messages wrong.
5. Some commands have multiple output streams, not a single "message" output
6. Commands with large output would take up a lot of memory storing it all in the string

I completely removed the CommandResult object in favor of storing programmatic data on the commandScope object. 

I've also switched the commands that output data to use the UIService directly rather than the message variable.

The trouble is that it's not as simple as "just use UIService":
1. updateSQL sometimes needs to send SQL to stdout and messages to stderr, sometimes send both to stdout
2. When a command is called in an embedded scenario, the command output may not want to be shown at all

We could add a getOutputStream() and getDataStream() type methods to commandScope which the integration can set up as needed. Often times, the output is generated further down in the code where there isn't the commandScope but that could be refactored. Or, maybe a more complex CommandOutput object that goes on commandScope that does more than just those two separate functions?

We could enhance the UIService to handle those cases better, but I'm not sure if that is too generic to have the command-level logic we sometimes need?

I'm also not sure if we should just drop the UIService.sendMessage() type functions anyway in favor of just using the LoggingSystem as the way everything outputs messages. If we did that, maybe it's less that we disable output and more change the log level? Does that have the same "too global" problems that UIService has?

wwillard7800:
I agree that Option 2 seems like the best choice.  The integrations are just too different in their connection needs.

As for the output, I would vote for keeping the UIService.  I prefer to manage the console output separately from the logging.

mcred:
I like the idea of an output stream where updates can be sent. I agree that logging should be separate. It makes sense to me that there would be 3 potential outputs: an output stream, a logging stream, or an exception. It's probably that each extension method can throw an exception that can be caught and handled upstream.

nvoxland:
I added pseudo-support into the NewMain class to see how a "find Database arguments and make connections for them in the integration" would work. 

It seems to be just fine, so going to stick with that approach for now and see how it goes.

nvoxland:
Where I struggle with the separated UIService vs. LoggingService is that for most types of ways Liquibase is ran, there is no UI besides the logs. Spring, Maven, Ant, Gradle, and even "cli with --logLevel set" all use the logs as their primary output. It's really only one mode of the CLI that has any sort of distinction.

Because we are managing the UIService and the LogService separately, we are dealing with duplicated messages in the output and/or duplicate lines of code that send something to both the UI and the logs. 

Part of our log level definition of "INFO" is "messages that would normally be shown to the user in a console application" which makes me ask "Instead of UIService, what if we just log at INFO level, and configure the CLI by default to show an unformatted INFO level log?"

nvoxland:
Trying to add the outputStream, it ended up making the most sense to add it to the CommandScope itself. I tried adding it to the regular Scope and the the UIService but both ended up feeling harder to use and too "global" for something that really is command-specific. Having it on the CommandScope makes it feel like the special-case `CommandScope.result` that it is.

Having it there will mean that any lower-level functions called by the command will need to take the stream as a parameter, but that is manageable.

PR #1738, opened 2021-03-02
Allow disallowed field value to be false. Added unit tests LB 65

wwillard7800:
Liquibase validate changesets before they are applied to a DBMS.
For example, following changeset for PostgreSQL is rejected because
PostgreSQL does not support ordered sequences:

```
<changeSet id="1" author="author">
   <createSequence cycle="false"
                   cacheSize="20"
                   incrementBy="50"
                   sequenceName="SOME_SEQUENCE"
                   ordered="true"
                   startValue="1"/>
</changeSet>
```

However, if someone writes the following changeset it will be
rejected by liquibase as well:

```
<changeSet id="1" author="author">
   <createSequence cycle="false"
                   cacheSize="20"
                   incrementBy="50"
                   sequenceName="SOME_SEQUENCE"
                   ordered="false"
                   startValue="1"/>
</changeSet>
```

This commit improves the validation in order to write ordered=false
explicitly.


nvoxland:
This is a duplicate of #900 which was closed as "should not bring in"

PR #1776, opened 2021-03-23
Refactor LiquibaseConfiguration API

### LINKED ISSUE #1691: Refactor LiquibaseConfiguration API
## Description
The existing `liquibase.configuration.LiquibaseConfiguration` API manages global configuration settings. However, it is a bit clunky to use and needs some improvements before we start recommending its usage in the LiquibaseCommand changes and beyond.

In particular:

* The current lookup style of `LiquibaseConfiguration.getInstance().getConfiguration(GlobalConfiguration.class).getOutputEncoding()` is long to use and boilerplate to write
* The ConfigurationContainer interface is trying to handle both configuration definition and value lookup
* Poor tracking of where configured values come from
* Difficult to support fuzzy matching of configuration keys in different lookup sources
* Inconsistent terminology usage
* Not following current code patterns
* Not utilizing Scope

Much of the overall design comes from Spring's configuration management system

## API Requirements
1. Split ConfigurationContainer into ConfigurationDefinition and ConfigurationValueProvider
1. Focus LiquibaseConfiguration on generic key/value lookup, with ConfigurationDefinitions being the optional "type safe" layer on top of that
1. ConfigurationValueProvider is a pluggable system of where/how to lookup configuration keys from different systems (system properties, environment variables, Scope, etc.)
1. Expose static fields for ConfigurationDefintions off GlobalConfiguration etc. to make lookups easier to code
1. Use generics and builder pattern to make ConfigurationDefinitions off GlobalConfiguration etc. easier to write
1. Improve metadata available for the ConfigurationDefinitions, including runtime lookup information
1. Move ConfigurationDefinition holders like GlobalConfiguration to be alongside the code they configure, not the configuration code.

## API Backwards Compatibility 
Tracked in #1732 

## Acceptance Criteria
* Ability to define config values that can be read from different locations in a defined hierarchy as follows:
  * Our standard PropertySource order follows spring boot's order. Properties are considered in the following order (with values from later/lower items overriding earlier/higher ones)

**# Config data (such as liquibase.properties file and piped stdin)
**# OS environment variables
**# Java System Properties
**# ServletContext init parameters
**# ServletConfig init parameters
**# Command line arguments
**# Code-level, direct setting of properties

* API changes are reviewed and support what we need in the short-term and should not need breaking changes in the future
* API is well javadoc documented and unit tested
* Existing user functionality that leverages system properties works as before

┆Issue is synchronized with this [Jira Story](https://datical.atlassian.net/browse/LB-1222) by [Unito](https://www.unito.io/learn-more)
┆Fix Versions: Liquibase 4.4.0

nvoxland:
Improve the liquibase.configuration package to support needed upcoming functionality

Fixes #1691

┆Issue is synchronized with this [Jira Story](https://datical.atlassian.net/browse/LB-1377) by [Unito](https://www.unito.io/learn-more)


liquibase-core/src/main/java/liquibase/configuration/AbstractMapConfigurationValueProvider.java
```diff
@@ -0,0 +1,82 @@
+package liquibase.configuration;
```
wwillard7800, line 1: This seems like a good class for a synopsis at the top.  Is it the top-level Map for all properties?

wwillard7800 (approved):
Unless I missed it, I think it would be good to have a listing of the precedence rules somewhere in the JavaDocs.

PR #1839, opened 2021-05-12
Implement new CLI

### LINKED ISSUE #1702: Replace existing CLI integration with one that dynamically leverages LiquibaseCommand classes
# Context
As part of the config API refactor and command object refactor in prep for new capabilities we are implementing a new CLI library. This is required to make available to the user the fruit of these previous refactors and to modernize some of the user experience. While there are many things we can do, we will focus on a couple which will deliver a better (eg, simpler, more tolerant) experience to users.

# Requirements
## Argument Locations
There are two types of settings a user can specify: "global configurations" and "command arguments". The global arguments go BEFORE the command group + name, the command arguments go AFTER the command name.

**The global settings** are not a part of command, but instead a way to change overall default behavior.
Examples of global settings include outputEncoding, logLevel, shouldRun etc.

There are never any required global settings. They are only used to change behavior from the defaults.

**The command arguments** are the settings for the particular command you are running. Examples of command settings include "changelogFile", "password", "username", "referenceUrl", etc.

Some command arguments are required, some are optional.

One thing to notice is that almost all the 4.3.0 global arguments will now be command arguments. Arguments like url, username, password, etc. apply to many of the commands but not all the commands. Starting with 4.4.0+ we will be more consistent in how we determine what is a global configuration vs. an input for a command with a very high bar for what is truly a global setting.

As an estimate, these existing global arguments would make sense to leave as global:
classloader
driver
databaseClass
defaultSchemaName
liquibaseSchemaName
defaultCatalogName
defaultSchemaName
databseChangelogTableName
databaseChangelogLockTableName
overwriteOUtputFile
driverPropertiesFile
changeExecListenerClass
changeExecListenerPropertiesFile
promptForNonLocalDatabase
includeSystemClasspath
defaultsFile
logLevel
logFile
outputFile

For end-user backwards compatibility, we will silently move known flags that used to be global arguments to be parsed as command arguments. No **existing** command line should stop working, **But, for any other arguments that are in the wrong spot we will fail with an error.**

One argument style we will **NOT support** are "positional arguments". Positional arguments are unnamed arguments that come after the command, like `git pull origin`. With our style, that would have to be `git pull --remote=origin`

### The user experience of this requirement:
* as an organization we will guide, teach, --help, and document the “right” way to supply global (left) and command args (right) on the command line. If the user puts a flag on the wrong side, we will exit and tell them.
* but, we will silently allow left-right tolerance for global flags that used to be on the other side. For example, since --url used to be a global flag we will still allow it to the left even though it is "supposed" to be on the right now.
* New commands will have their command arguments accepted only to the right of the command, and will be taught as such.

## Multi-word arguments
There are arguments which in plain english would present as multiple words. “reference URL”, “reference username” and the like. We will document the “right” style for using these, but will be tolerant to remove user frictions.

### The user experience of this requirement:
* as an organization we will guide, teach, `--help`, and document multi-word arguments as
  * “-” separated (aka kabob-case)
  * all lower case
  * For example, --reference-url
* but, we will silently allow:
  * no dash
  * mixed case
  * upper case
  * so long as the “repaired” argument matches a legitimate argument
    * “repairing” is basically string manipulations between `--` (two dashes) and the next “ “ (space)
  * for example, these would work: `--referenceUrl` , `--re-Fer-Ence-URL` , `--referenceurl`, `--REFERENCEURL`, `--REFERENCE-URL`
  * we do not allow – (two dashes) or a space in the middle of the arg.
---fails: ~~reference~~url=blah

## Argument key/value separators
We will accept and document “=” for separators between arguments and values. For example --url=jdbc:h2:/mem.

We will support “ “ (space) for separators, like `--url jdbc:h2:/mem` for backwards compatibility reasons.

We will not support “:' or anything else for separators

### The user experience of this requirement:
* as an organization we will guide, teach, --help, and document multi-word arguments as
  * using the “=“ (the equal sign) for separators between arg and value
  * values are not to be wrapped in quotes url=this/url:mem
  * so long as there is no – or “ “ (space) in the value
* but we will be tolerant and allow
  * values wrapped in quotes url=”this/url:mem”
We will not support
  * the “ “ (a space) as a separator
  * the “:“ (colon) as a separators

## Cascading Arguments
The command line arguments users specify get rolled into the overall configuration management system **at the highest level of priority.**

Therefore, arguments specified at the command line can also (still) be provided via the liquibase.properties file, in addition to other locations checked by the configuration system: system properties, environment variables,  as follows:

1. Config data (such as liquibase.properties file and piped stdin)
1. OS environment variables
1. Java System Properties
1. ServletContext init parameters
1. ServletConfig init parameters
1. Command line arguments

Each argument will be looked up in the following order with the following keys:

1. liquibase.command.commandGroup.commandName.argumentName
1. liquibase.command.argumentName
1. argumentName

from top to bottom, where "commandGroup.commandName" == the command group.name and "argumentName" is the argument.

We **do not check** "liquibase.argumentName" to avoid conflicting with other configuration settings. The non-namespaced "argumentName" version is there for backwards compatibility reasons and because some of the configuration sources (like liquibase.properties) should be able to treated as a simple arguments file.

Global parameters are looked up purely from their configured names.

### The user experience of this requirement:
* we will in docs, guidance, --help, and the like teach users that arguments provided to the command line are the “winning” arguments, over-riding the same args passed in other locations, such as defaults files (properties, POMs, etc) ENV vars and the like.
* we will not **in the console** inform users that multiple locations have set the same property and that args passed on the command line will be used

## Global Argument Format
Global arguments have namespaces associated with them to avoid conflicts between extensions. To prevent extensions from conflicting with our settings, we use the 'liquibase' namespace. Example: liquibase.logLevel is one of our configuration settings. Namespaces can be multiple levels deep, such as `liquibase.hub.url`

When exposing global settings through the CLI as arguments, we include the namespacing in the argument, coverting camel case to kabob-case as above and also replacing .s with -s.

* `liquibase.logLevel - > --liquibase-log-level`
* `liquibase.hub.url - > --liquibase-hub-url`

Using the full name on arguments continues to prevent collisions and also makes conversion to other places they can be specified such as liquibase properties, system properties, environment variables, etc. more understandable.

### The user experience of this requirement:
* we will in docs, guidance, --help, and the like teach users that arguments provided to the command line are the “winning” arguments, over-riding the same args passed in other locations, such as defaults files (properties, POMs, etc) ENV vars and the like.
* we will not in the console inform users that multiple locations have set the same property and that args passed on the command line will be used

## Tolerance
As mentioned above, except for transparently supporting backwards compatibility and case insensitivity in key names, we do not attempt to guess at what a user is asking us to do. We only run the commands/arguments they specifically put and that we specifically expect.

If a user misspells a command or argument that is close to something we support, we will try to output "did you mean X" messages if the new CLI library supports this for “free”

The user experience of this requirement:

* STAGE 1 – we will not implement “did you mean” mappings to try to offer guidance for closely misspelled commands. If PICO comes with these out of the box, then great.
* STAGE 2 - (date TBD), we will spend effort to support “did you mean” suggestions for misspelled commands.

We will continue to fail commands which use unexpected command args.

## Online Help
We will be able to generate --help which includes the description etc. from the commands and arguments stored in the LiquibaseCommand classes.

### The user experience of this requirement:
* generating new help will be a long-running epic in which help is updated to be clearer, less java-centric and takes advantage of formatting and other options made possible by PICO.

┆Issue is synchronized with this [Jira Bug](https://datical.atlassian.net/browse/LB-1250) by [Unito](https://www.unito.io)
┆Fix Versions: Liquibase 4.4.0

nvoxland:
Fixes #1702

┆Issue is synchronized with this [Jira Story](https://datical.atlassian.net/browse/LB-1663) by [Unito](https://www.unito.io)


liquibase-cli/src/main/java/liquibase/integration/commandline/CommandRunner.java
```diff
@@ -0,0 +1,75 @@
...
+        final String[] commandName = LiquibaseCommandLine.getCommandNames(spec.commandLine().getParseResult());
+        for (int i=0; i<commandName.length; i++) {
+            commandName[i] = StringUtil.toCamelCase(commandName[i]);
+        }
+
+        final CommandScope commandScope = new CommandScope(commandName);
+        final File outputFile = IntegrationConfiguration.OUTPUT_FILE.getCurrentValue();
+        OutputStream outputStream = null;
+
+        try {
```
suryaaki2, line 43: Should we be using try with resources here?

liquibase-cli/src/main/java/liquibase/integration/commandline/LiquibaseCommandLine.java
```diff
@@ -0,0 +1,789 @@
...
+import java.security.PrivilegedAction;
+import java.util.*;
+import java.util.logging.*;
+import java.util.stream.Collectors;
+import java.util.stream.Stream;
+
+import static java.util.ResourceBundle.getBundle;
+import static liquibase.util.SystemUtil.isWindows;
+
+
```
suryaaki2, line 45: Java doc comments for class definition

liquibase-cli/src/main/java/liquibase/integration/commandline/LiquibaseCommandLine.java
```diff
@@ -0,0 +1,789 @@
...
+            }
+        } catch (Throwable e) {
+            handleException(e);
+            return 1;
+        } finally {
+            cleanup();
+        }
+    }
+
+    private boolean wasHelpOrVersionRequeted() {
```
suryaaki2, line 330: typo in method name "Requeted"

liquibase-cli/src/main/java/liquibase/integration/commandline/LiquibaseCommandLine.java
```diff
@@ -0,0 +1,789 @@
...
+    private List<ConfigurationValueProvider> registerValueProviders(String[] args) {
+        final LiquibaseConfiguration liquibaseConfiguration = Scope.getCurrentScope().getSingleton(LiquibaseConfiguration.class);
+        List<ConfigurationValueProvider> returnList = new ArrayList<>();
+
+        final CommandLineArgumentValueProvider argumentProvider = new CommandLineArgumentValueProvider(commandLine.parseArgs(args), legacyNoLongerCommandArguments);
+        liquibaseConfiguration.registerProvider(argumentProvider);
+        returnList.add(argumentProvider);
+
+        final File defaultsFile = new File(IntegrationConfiguration.DEFAULTS_FILE.getCurrentValue());
+        if (defaultsFile.exists()) {
```
suryaaki2, line 405: Can defaultsFile be null?

liquibase-cli/src/main/java/liquibase/integration/commandline/LiquibaseCommandLine.java
```diff
@@ -0,0 +1,789 @@
...
+        } else {
+            classLoader = AccessController.doPrivileged((PrivilegedAction<URLClassLoader>) () -> new URLClassLoader(urls.toArray(new URL[0]), null));
+        }
+
+        Thread.currentThread().setContextClassLoader(classLoader);
+
+        return classLoader;
+    }
+
+    private void addSubcommand(CommandDefinition commandDefinition, CommandLine commandLine) {
```
suryaaki2, line 571: camelCasing subCommand

suryaaki2 (changes requested):

suryaaki2 (commented):
Can we add javadoc to Class and Methods?

liquibase-cli/src/main/java/liquibase/integration/commandline/CommandRunner.java
wwillard7800: The issue with doing a try-with here is that the OutputStream need to exist for the scope of the command, so it will need to be close after the command finishes.

liquibase-cli/src/main/java/liquibase/integration/commandline/LiquibaseCommandLine.java
```diff
@@ -0,0 +1,789 @@
...
+    }
+
+    private void cleanup() {
+        if (fileHandler != null) {
+            fileHandler.flush();
+            fileHandler.close();
+        }
+    }
+
+    public LiquibaseCommandLine() {
```
wwillard7800, line 95: Do the updateCount/Tag and changelogSyncToTag commands need to be in the legacyPositionalArguments?

liquibase-cli/src/main/java/liquibase/integration/commandline/LiquibaseCommandLine.java
wwillard7800: defaultsFile has a default value of "liquibase.properties"

suryaaki2 (approved):

liquibase-cli/src/main/java/liquibase/integration/commandline/CommandRunner.java
nvoxland: I think it's the "if" statement part that doesn't work with try with resources.

liquibase-cli/src/main/java/liquibase/integration/commandline/LiquibaseCommandLine.java
nvoxland: we're counting "subcommand" as one word, not two so shouldn't be camel cased

liquibase-cli/src/main/java/liquibase/integration/commandline/LiquibaseCommandLine.java
nvoxland: fixed

PR #1861, opened 2021-05-21
Only snapshot tables/columns in loadData

### LINKED ISSUE #1699: loaddata is slow when csv file doesn't specify type
## Environment
**Liquibase Version**: 4.1.1, 4.2.0,  4.3.1

**Liquibase Extension(s) & Version**:  hana-db (same version as liquibase-core)

* MySQL 5.7 to 8
* Postgresql 9.5 to 12.3
* Oracle 
  * 11.2.0.2
  * 12.1.0.2
  * 12.2.0.1
  * 18.3.0
  * 19.3.0
 

## Description
In our CI/CD environment we try to apply changelogs on all database that we support. 
To provide such a feature, we generate a snapshot of our database at release - 1 and apply changelogs to currently developped release.
Historycally we used 3.5.2 version of liquibase and custom hanadb extension. We try to switch to new version of liquibase and official hanadb extension. However our CI/CD pipeline just fails because of the running times.

After analysis of the logs we found out that the issue is maily due to loaddata function that takes 4 to 6 seconds to load files that just have headers.

## Steps To Reproduce
Apply this changeset

```xml
    <changeSet author="SYSTEM" id="1610033379074-154">
        <createTable tableName="invalidbobjevent">
            <column name="task_id" type="java.sql.Types.VARCHAR(255)">
            </column>
            <column name="invalidevent_id" type="java.sql.Types.VARCHAR(255)">
            </column>
        </createTable>
    </changeSet>
    <changeSet author="SYSTEM" id="1610033379074-452">
        <loadData usePreparedStatements="true" commentLineStartsWith="#" encoding="UTF-8" file="data/invalidbobjevent.csv" quotchar="&quot;" relativeToChangelogFile="true" separator="," tableName="invalidbobjevent">
            <column header="task_id" name="task_id"/>
            <column header="invalidevent_id" name="invalidevent_id"/>
        </loadData>
    </changeSet>
```

## Actual Behavior
Here is the log we get : 

```
INFO: ChangeSet database_snapshot/changelog-export.xml::1610033379074-452::SYSTEM ran successfully in 5147ms
Feb 15, 2021 9:04:59 AM liquibase.changelog
INFO: Data loaded from data/invalidevententity.csv into invalidevententity
```

## Expected/Desired Behavior
Reduce the time to import (like 1ms) such a file. Like it was on 3.5.2 version of liquibase-core.

Example of acceptable logs

```
INFO: ChangeSet database_snapshot/changelog-export.xml::1610033379074-452::SYSTEM ran successfully in 548ms
Feb 15, 2021 8:57:00 AM liquibase.changelog
INFO: Data loaded from data/invalidevententity.csv into invalidevententity
```

### time of execution of same changesets in 3.5.2
results in 3.5.2

Nota: "create" test runs the developped changeset to simulate the installation of new platform, "update" installs the snapshot then run the newly developped changelang, simulating an update to release -1 to current developpe release

### time of execution with 4.1.1](same order of magnitude for 4.3.1)
results in 4.1.1

## Additional Context
It seems that sqlsever 2017 and 2019 are less affected by the bug

### CI/CD informations
Our software supports many dbms and dialects, in order to bring a better control 
on our schema migrations we have implemented an integration test allowing to test 
both the generation of an empty schema and a schema existing in the previous version 
on all dbms and dialects. In the captures below you can observe the execution of the 
test with liquibase 3.5.2 and 4.1.

* 3.5.2 with all dataset

* 4.1.1 with all dataset (oracle hits its limit, hana has only a little issue with one changeset now the extension is official)
image
-4.1.1 with only dataset that have at least one data row (env 90 CL less)

#### Creation test
During creation there is a regression on the execution time. It will therefore be more 
expensive to generate our schema in production, which can potentially cause problems when 
starting our services (on-delivery).

#### Update test
The test of the update is done in two phases:

* Import database in previous versions with liquibase
* Apply changelog on the database to test the new changesets (since last version)

To import database in previous versions, we generate our database on postgres, 
then we use our application to insert data to finally use the generateDatabaseChangelog 
feature.

We have also implemented an internal tool to automate the processus and parsing 
some changet outputed by generateDatabaseChangelog feature to support all DBMS 
(generic type for example). The tool export also the databasechangelog table to recreate 
them during import. 

Before making the liquibase update, the update was even more efficient than the creation 
because not benefiting from the history the changelog is better orchestrated. 
As you can see in the captures above, it is now much more expensive to simulate the 
update in production. It should be noted that the export was done with liquibase 3.5.2. 
Does liquibase 4.1 use new changeset in order to insert more data faster?

nvoxland:
Fixes #1699


suryaaki2 (approved):

nvoxland:
Cherry picked into 4.4.x

PR #1864, opened 2021-05-25
Postgresql: Preserve existing search_path when setting the defaultSchemaName

### LINKED ISSUE #1821: PostgreSQL: cannot use an extension from the public schema without prefix (search_path is overriden by liquibase)
## Environment
**Liquibase Version**: 4.3.4

**Operating System Type & Version**: Ubuntu 20.04

## Description
Hello,

I'm facing the same problem that's described in [this forum thread](https://forum.liquibase.org/t/access-postgis-from-different-schema/4708/3) (I did not reply in the forum because the number of links in a comment is restricted for new users). I'm using postgis as well but I guess it's the same for any extension that you would install on the public schema. The problem is that liquibase sets the search_path to only the default schema.

In the code, this is done in [DatabaseUtils#initializeDatabase](https://github.com/liquibase/liquibase/blob/master/liquibase-core/src/main/java/liquibase/database/core/DatabaseUtils.java#L36), which is called from [PostgresDatabase#rollback](https://github.com/liquibase/liquibase/blob/master/liquibase-core/src/main/java/liquibase/database/core/PostgresDatabase.java#L421) and the rollback function seems to be called for each changeset, after the preconditions are checked, in [ChangeSet#execute](https://github.com/liquibase/liquibase/blob/master/liquibase-core/src/main/java/liquibase/changelog/ChangeSet.java#L614).

I think that the problem was introduced with this commit: https://github.com/liquibase/liquibase/commit/feb2412. Before that, the search_path was not forced for each changeset, and there was no problem using an extension from the public schema without using prefixes.

I guess that there was a good reason to force the search_path, but I think that there should be an option to override the default behavior and to allow the user to set a custom search_path.

## Steps To Reproduce
With postgis:

* Add the postgis extension to the public schema: `CREATE EXTENSION postgis;`
* Run liquibase on another schema (`my_schema`) with the following changeset:

```
create table my_schema.test_table (
     a_box        Box2D
)
```

Liquibase will fail with the following error:

```
ERROR: type "box2d" does not exist
```

Thanks.

### LINKED ISSUE #1823: Cannot call extension function from migration
## Environment
**Liquibase Version**: 4.3.3

**Liquibase Integration & Version**: Spring Boot 2.4.5

**Database Vendor & Version**: PostgresQL 12 w/ Timescale Extension

**Operating System Type & Version**: Linux 5.11

## Description
I am attempting to call an extension function that is added by the Timescale extension via a migration using a \<sql\> element. When the migration runs, I get an error indicating that the function doesn't existing with the given arguments. However, I have confirmed that the function is present by running the exact same SQL against the database directly.

I have used a similar call in the past and had success, so I eventually discovered that this sort of thing used to work in 3.7.0, but no longer works in 4.3.3. It seems like something is now configured in Liquibase such that it can't find these extension functions (schema related perhaps?).

This seems like it might be related to: https://github.com/liquibase/liquibase/issues/1821, but I wasn't sure.

## Steps To Reproduce
1) Run Timescale docker image (timescale/timescaledb:2.0.1-pg12)
2) Attempt to run a migration calling the create_hypertable extension function, or any extension function I suspect ```<sql>select create_hypertable('schema.table', 'timestamp');</sql>```
3) Run migration

## Actual Behavior
Migration fails with:

ERROR: function create_hypertable(unknown, unknown) does not exist
Hint: No function matches the given name and argument types. You might need to add explicit type casts.

## Expected/Desired Behavior
Migration succeeds.

┆Issue is synchronized with this [Jira Bug](https://datical.atlassian.net/browse/LB-1581) by [Unito](https://www.unito.io)

nvoxland:
Currently liquibase blindly sets the search_path to be equal to the defaultSchemaName. It also resets it in every rollback() call rather than relying on change types that modify it to set it back themselves.

This changes the behavior to first pull the existing search_path, and modify it as `defaultSchemaName + existing_search_path`. 
If the existing search_path already starts with the defaultSchemaName, nothing is done.

It also removes the call to DatabaseUtil.initializeDatabase() on every rollback call. 

Fixes #1823
Fixes #1821

#### Links
* Fixed Issues: 
  * [https://github.com/liquibase/liquibase/issues/1823](https://github.com/liquibase/liquibase/issues/1823)
  * [https://github.com/liquibase/liquibase/issues/1821](https://github.com/liquibase/liquibase/issues/1821)
* Local Branch: [https://github.com/liquibase/liquibase/tree/LB-1581](https://github.com/liquibase/liquibase/tree/LB-1581)
* Unit Tests: [https://github.com/liquibase/liquibase/pull/1864/checks](https://github.com/liquibase/liquibase/pull/1864/checks)
* Integration Tests: (shown in Unit Tests link above)
* Functional Tests: [https://jenkins.datical.net/job/liquibase-pro/job/LB-1581/](https://jenkins.datical.net/job/liquibase-pro/job/LB-1581/)

#### Testing
* Steps to Reproduce:
  * [https://github.com/liquibase/liquibase/issues/1823](https://github.com/liquibase/liquibase/issues/1823)
  * [https://github.com/liquibase/liquibase/issues/1821](https://github.com/liquibase/liquibase/issues/1821)
* Guidance: 
  * Impacts only postgresql, but changes code that runs before all postgresql operations
  * Changes code that changes the default schema for createProcedure type changes where we don't modify the underlying SQL but rely on a different default schema.

#### Dev Verification
Tested with a variety of schema settings and change types.

(Did not save output at the time)

## Test Requirements (Liquibase Internal QA)
This test requires a version of postgres with postgis installed. You may need to login to Docker Hub to pull the images. The login command is `docker login -u <yourname> -p <yourpassword>` .

Get the Docker Image:

`docker pull postgis/postgis`

Start the Postgres with PostGis Database :

`docker run --name=postgis  -p5439:5432 -e POSTGRES_PASSWORD=<YOURPASSWORD>-d postgis/postgis`

Create a schema on the database:

```
--DROP SCHEMA IF EXISTS  "schema1" CASCADE;
CREATE SCHEMA "schema1" AUTHORIZATION "postgres";

GRANT CREATE, USAGE ON SCHEMA "schema1" TO "postgres"; 
ALTER DEFAULT PRIVILEGES IN SCHEMA "schema1" GRANT ALL ON TABLES TO "postgres";
ALTER DEFAULT PRIVILEGES IN SCHEMA "schema1" GRANT EXECUTE ON FUNCTIONS TO "postgres";
ALTER DEFAULT PRIVILEGES IN SCHEMA "schema1" GRANT USAGE ON SEQUENCES TO "postgres";
```

Configure the database connection in liquibase.properties

```
url: jdbc:postgresql://localhost:5439/
username: postgres
password: YOURPASSWORD

# If your postgres driver is located outside of the liquibase/lib directory configure the classpath.
# classpath:postgresql-42.2.18.jar;C://
```

**Manual Tests**

* Use the attached changelog for these tests.

_Verify successful update for tables, indexes and inert statements with spatial types and spatial methods.__

* Box2D **PASS**
* GEOMETRY **PASS**
* GIST **PASS**
* ST_GeomFromText **PASS**

_Verify a successful update for a procedure with spatial parameters (BOX2D, BOX3D, GEOMETRY and GEOMETRY_DUMP)._ **_PASS_**

_Verify a successful rollback of a procedure with spatial parameters (BOX2D, BOX3D, GEOMETRY and GEOMETRY_DUMP)._ **_PASS_**

_Verify a successful generate-changelog for tables, indexes and insert statements with spatial types and spatial methods._ **_PASS_**

**Automated Tests**

No new functional tests required for this fix; unit tests were added.

┆Issue is synchronized with this [Jira Story](https://datical.atlassian.net/browse/LB-1747) by [Unito](https://www.unito.io)


suryaaki2 (changes requested):
@nvoxland Can we write integration tests?

nvoxland:
Re-reading the original issues and tracing through the code, I realized a big part of the fix needs to be pulling the search_path updating out of rollback() and relying on the underlying change classes to deal with the search_path correctly. I pushed a change for that.

We don't have a great way to add integration tests at that code spot right now. Need a way to do that, but it's a larger project. I did add better unit tests as part of the last change, though.

suryaaki2 (approved):

nvoxland:
It may impact edge cases enough I marked this with "ReleaseMajor"

jeffrey-anderson:
I'm also having this problem. Is anyone assigned to merge this into an actual release?

abedwardsw:
i'm having the same issue with liquibase 4.27.0, did this issue get rolled out to a release?

Issue #1741, opened 2021-03-03
Liquibase pollutes System.out in a Spring Boot application


monrealis:
Liquibase writes directly to System.out instead of writing through a configured logger in a Spring Boot Application. This is unexpected to the consumer of System.out, which expects, for example, JSON in System.out.

Liquibase version: 4.3.1
Spring boot version: 2.4.3
Message: Skipping auto-registration

Stack trace:

```
	at liquibase.ui.ConsoleUIService.sendMessage(ConsoleUIService.java:45)
	at liquibase.hub.HubUpdater.register(HubUpdater.java:372)
	at liquibase.Liquibase.lambda$update$1(Liquibase.java:242)
	at liquibase.Scope.lambda$child$0(Scope.java:160)
	at liquibase.Scope.child(Scope.java:169)
	at liquibase.Scope.child(Scope.java:159)
	at liquibase.Scope.child(Scope.java:138)
	at liquibase.Liquibase.runInScope(Liquibase.java:2322)
	at liquibase.Liquibase.update(Liquibase.java:216)
	at liquibase.Liquibase.update(Liquibase.java:202)
	at liquibase.integration.spring.SpringLiquibase.performUpdate(SpringLiquibase.java:322)
	at liquibase.integration.spring.SpringLiquibase.afterPropertiesSet(SpringLiquibase.java:275)
	at org.springframework.beans.factory.support.AbstractAutowireCapableBeanFactory.invokeInitMethods(AbstractAutowireCapableBeanFactory.java:1845)
	at org.springframework.beans.factory.support.AbstractAutowireCapableBeanFactory.initializeBean(AbstractAutowireCapableBeanFactory.java:1782)
	at org.springframework.beans.factory.support.AbstractAutowireCapableBeanFactory.doCreateBean(AbstractAutowireCapableBeanFactory.java:602)
	at org.springframework.beans.factory.support.AbstractAutowireCapableBeanFactory.createBean(AbstractAutowireCapableBeanFactory.java:524)
	at org.springframework.beans.factory.support.AbstractBeanFactory.lambda$doGetBean$0(AbstractBeanFactory.java:335)
	at org.springframework.beans.factory.support.DefaultSingletonBeanRegistry.getSingleton(DefaultSingletonBeanRegistry.java:234)
	at org.springframework.beans.factory.support.AbstractBeanFactory.doGetBean(AbstractBeanFactory.java:333)
[...]

PR #1869, opened 2021-05-26
HubUpdater releases lock even though registration is skipped

### LINKED ISSUE #1865: HubUpdater releases lock even though registration is skipped
This is slightly linked #1741. 

In 
https://github.com/liquibase/liquibase/blob/10a0c509c59401e9351cf7d134b4bc671b5d500f/liquibase-core/src/main/java/liquibase/hub/HubUpdater.java#L364-L383

The currently held lock is released and then reacquired again. However, when using Liquibase programatically as a library the prompt is always disabled and therefore releasing and then acquiring the lock is entirely not needed.

Can I propose a change that will do

```
if (!Scope.getCurrentScope().getUI().getAllowPrompt()) {
    return;
}
```

This will solve the problem with the polluting of the logging with "Skipping auto-registration" since registering is anyways not possible when using it as a library.

┆Issue is synchronized with this [Jira Story](https://datical.atlassian.net/browse/LB-1748) by [Unito](https://www.unito.io)

nvoxland:
Fixes #1865

LB-1748

#### Links

- Fixed Issue: https://github.com/liquibase/liquibase/issues/1865
- Local Branch: https://github.com/liquibase/liquibase/tree/LB-1748
- Unit Tests: https://github.com/liquibase/liquibase/pull/1869/checks
- Integration Tests: (shown in Unit Tests link above)
- Functional Tests: https://jenkins.datical.net/job/liquibase-pro/job/LB-1748/

#### Testing

- Steps to Reproduce: https://github.com/liquibase/liquibase/issues/1865
- Guidance: 
  - Refactors code to handle locking/unlocking better in embedded/non-CLI usage
  
#### Dev Verification

Tested with some custom, direct calls to ensure they manage the lock better. Tested that the CLI usage isn't impacted.

┆Issue is synchronized with this [Jira Story](https://datical.atlassian.net/browse/LB-1753) by [Unito](https://www.unito.io)
┆Fix Versions: Liquibase 4.4.1


filiphr:
I think that this will also fix #1741 since the logging of the skip message will never be done.

nvoxland:
It will probably help with #1741, but I'd like to keep that one separate still to set up the UIService for spring boot better in general.

filiphr:
That makes sense. I'd like to add that there is really not a big difference whether the UIService is used for Spring Boot or in another projects as a library. We at [Flowable](https://github.com/flowable/flowable-engine) are heavy users of Liquibase, it has helped us a lot to bring support in an easy way. We use Liquibase as a library. We use Liquibase outside of Spring Boot in a standalone fashion by creating an instance of `Liquibase` and working with it.

We started looking into upgrading to the latest version of Liquibase (we were on 3.8.0, due to different issues we faced earlier). That's why there were some issues created by me recently. Our current proof of concept in working with Liquibase 4.3.5 without jaxb can be seen in https://github.com/filiphr/flowable-engine/commit/d6822b6ff59f44bbab51934084ec387f3ddf69bc (in case you are interested)

suryaaki2 (approved):

PR #1901, opened 2021-06-10
Fix Race Condititons During Liquibase Initialization

### LINKED ISSUE #1584: Race condition and possible lock bypass if two nodes run on clean schema
## Environment
**Liquibase Version**:
3.10.x, 4.0.x, 4.1.x
All affected
Some parts is specific for Oracle, DB2 or Derby, others are vendor independent
OS independent

## Description
When liquibase run on clean schema it creates and initializes DatabaseChangeLogLock table before it can use it's standard locking mechanism. In clustered environment, applications can be deployed for the first time to multiple nodes at same time. This initialization process is not protected against race conditions. 

Table creation part of this issue is already mentioned here: https://liquibase.jira.com/browse/CORE-2596
It tried to fix the problem by parsing the database error message for table already exists condition.

```
catch (DatabaseException e) {
  if ((e.getMessage() != null) && e.getMessage().contains("exists")) {
  //hit a race condition where the table got created by another node.
}
```

This solution does not work with all database and/or non English environment. A bug already reported here: https://github.com/liquibase/liquibase/issues/1036

DatabaseChangeLogLock table content initialization (insert) can fail with primary key violated error if another node inserts this record between selecting and inserting operations.
On Derby and DB2 database a DatabaseChangeLogLock table recreation can happen if it has wrong column type. It can delete an already acquired lock so can bypass locking.

Possible solutions:

* Forbid running 'update' on schemas without DatabaseChangeLogLock table
* Make liquibase table initialization process protected from race conditions

For the latter, i would suggest:

* Generalizing race condition handling in DatabaseChangeLogLock table initialization:
	* check table/record existence
  * create/insert if table/record does not exist
  * Retry (GOTO 1) once if table creation or record insertion fails 
* Change DatabaseChangeLogLock table modification to table alter instead of drop/recreate and move it to the synchronized section (do it while having the lock)

## Steps To Reproduce
Run multiple instances of liquibase update on an empty database schema at same time

## Actual Behavior
* Sometimes create table DATABASECHANGELOGLOCK fails with 'table already exists' error (On oracle database)
* Sometimes insert into DATABASECHANGELOGLOCK fails with 'primary key violated' error
* Sometimes multiple nodes start updating the schema (On Derby and Db2)
* Sometimes liquibase creates and initializes it's tables successfully

## Expected/Desired Behavior
* Liquibase always creates and initializes it's tables successfully

schrieveslaach:
## Environment
**Liquibase Version**: 4.5.0

## Pull Request Type
* Bug fix (non-breaking change which fixes an issue.)

## Description
When Liquibase tries to initialize the changelog lock table, multiple race conditions can arise. This commit fixes the following issues:

- [x] During table creation the exception is already handled when two concurrent JVM processes try to create the changelog lock table. However, the database transaction has to be aborted when the exception arises.
- [x] During initialization of the table a unique constraint exception can arise and the corresponding exception handling has been added.
- [x] The database lock could be temporarily handed over to another Liquibase instance on another JVM and then the cache of the changelog service would be invalid. Therefore, the cache will be reset before applying the changeset.
- [x] For certain DBMSs the SQL state will be evaluated so that Liquibase is able to detect these race conditions reliable

Fixes #1584 (steps to reproduce are provided there)

### DIFF
--- a/liquibase-core/src/main/java/liquibase/Liquibase.java
+++ b/liquibase-core/src/main/java/liquibase/Liquibase.java
@@ -226,7 +226,9 @@ public class Liquibase implements AutoCloseable {
                     checkLiquibaseTables(true, changeLog, contexts, labelExpression);
                 }
 
-                ChangeLogHistoryServiceFactory.getInstance().getChangeLogService(database).generateDeploymentId();
+                ChangeLogHistoryService changeLogService = ChangeLogHistoryServiceFactory.getInstance()
+                   .getChangeLogService( database );
+                changeLogService.generateDeploymentId();
 
                 changeLog.validate(database, contexts, labelExpression);
 
@@ -235,6 +237,12 @@ public class Liquibase implements AutoCloseable {
                 //
                 hubUpdater = new HubUpdater(new Date(), changeLog, database);
                 hubUpdater.register(changeLogFile);
+
+                // Before continuing the changelog service has to be invalidated due to the fact that the
+                // hubUpdater releases the lock temporarily. In this time span another JVM instance might have
+                // acquired the database lock and could have applied further changesets to prevent that
+                // liquibase works with an outdated changelog.
+                changeLogService.reset();
 
                 //
                 // Create or retrieve the Connection if this is not SQL generation
@@ -2317,10 +2325,6 @@ public class Liquibase implements AutoCloseable {
         setChangeLogParameter("database.supportsSchemas", database.supportsSchemas());
         setChangeLogParameter("database.supportsSequences", database.supportsSequences());
         setChangeLogParameter("database.supportsTablespaces", database.supportsTablespaces());
-    }
-
-    private LockService getLockService() {
-        return LockServiceFactory.getInstance().getLockService(database);
     }
 
     public void setChangeExecListener(ChangeExecListener listener) {
--- a/liquibase-core/src/main/java/liquibase/database/Database.java
+++ b/liquibase-core/src/main/java/liquibase/database/Database.java
@@ -453,5 +453,13 @@ public interface Database extends PrioritizedService {
     String unescapeDataTypeString(String dataTypeString);
 
     ValidationErrors validate();
+
+    default boolean isTableExistsException(DatabaseException e) {
+        return false;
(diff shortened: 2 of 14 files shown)


aixigo-soffermann:
The pull request looks good to me. I built the modified sources locally on my Mac (thanks for adding the `<tarLongFileMode>posix</tarLongFileMode>` configuration!) and used the result in a project to perform some database migrations with it. The migration works, and I did not encounter any issues. Would be nice to see this in the official branch soon. 👍

zholub:
The PR still contains code which mentioned in #1584 as unreliable:

```
if ((e.getMessage() != null) && e.getMessage().contains("exists")) {

if ((e.getMessage() != null) && e.getMessage().contains("duplicate")) {
```

Not all RDBMS use these words in their error messages. Eg Oracle:
_"ORA-00955: name is already used by an existing object"
"ORA-00001: unique constraint (XXX) violated"_

In addition, in Oracle you can change the language of the server with NLS_LANG which affects error messages.
This message in german: _ORA-00955: Name wird bereits von einem vorhandenen Objekt verwendet_

For preparing oracle error message i would search for oracle error codes (Eg: ORA-00955) which is same for all languages. 

I suppose that other RBMDS also can have non matching error messages. 
A solution for this could be is simply ignoring the error messages and repeat the initialization for any errors ONCE.
In a second attempt the temporal errors (in this case the race condition) won't happen again.

schrieveslaach:
@zholub, I think the new version of this PR handles these race conditions better. Do you mind to have a look or even test the PR?

schrieveslaach:
Can anybody tell me why is the test failing to login to Artifactory?

nvoxland:
I pushed the current version of master to your fork to fix the build issue.

schrieveslaach:
@nvoxland, I rebased the PR so that includes the latest development version. Now, the CI checks work.

What do you think about the PR's content? Do you think it is a valuable contribution to Liquibase?

schrieveslaach:
Superseded by #2198

kataggart:
Thanks @schrieveslaach for updating!

PR #1910, opened 2021-06-16
DAT-6972 Added logic that sends operation to hub for dropAll

riakivets:


suryaaki2 (approved):

liquibase-core/src/main/java/liquibase/hub/HubService.java
```diff
@@ -23,6 +19,8 @@
 
     Project getProject(UUID projectId) throws LiquibaseHubException;
 
+    Project findProjectByConnectionIdOrJdbcUrl(UUID connectionId, String jdbcUrl) throws LiquibaseHubException;
```
nvoxland, line 22: It is never very safe to add methods to an interface, because that will be a breaking API change to anyone that is implementing this interface. It's a fairly internal/specialized interface, so probably OK for this case.

nvoxland (changes requested):
This was branched from master after 4.4.x was made, so it needs to ether be rebased or cherry picked back to 4.4.x post-merge-to-master.

liquibase-core/src/main/java/liquibase/hub/HubService.java
riakivets: Good point @nvoxland!  As  this method is being used for hub only. I'll leave it.

riakivets:
This branch was created from master but should be 4.4.x. 
@riakivets will recreate this PR with fix

PR #1917, opened 2021-06-18
DAT-6769 fix bug with loading property elements from liquibase.properties

riakivets:


nvoxland (approved):
My worry with the new @PropertyElement annotation is that it's an odd, non-maven thing and so it is easy for contributors to add fields that don't include it.

But, we area also planning on rewriting all these maven classes hopefully soon, so I'm fine with it as at least a temporary solution.

nvoxland (changes requested):
Because this was branched off master, it can't be brought into 4.4.x without being rebased or something so it doesn't have the commits already in master that can't go into 4.4.1

riakivets:
This branch was created from master but should be 4.4.x.
@riakivets will recreate this PR with fix

PR #1936, opened 2021-06-24
DAT-7230 Fixed sonar issues

riakivets:


liquibase-core/src/main/java/liquibase/parser/core/formattedsql/FormattedSqlChangeLogParser.java
```diff
@@ -138,7 +139,7 @@ public DatabaseChangeLog parse(String physicalChangeLogLocation, ChangeLogParame
                             }
                             continue;
                         } catch (NumberFormatException | NullPointerException nfe) {
-                            throw new ChangeLogParseException("Unknown ignoreLines syntax in changeset " + changeSet.toString(false));
+                            throw new ChangeLogParseException("Unknown ignoreLines syntax");
```
wwillard7800, line 142: What was the Sonar issue resolved here?  Did it involve the ChangeSet class?  Is that why it was removed?

riakivets: Sonar was complaing that changeSet.toString(false) can produce NPE because it is never initialized and it's true it's alway null

liquibase-core/src/main/java/liquibase/example/StartH2Main.java
```diff
@@ -48,7 +49,13 @@ public static void main(String[] args) throws Exception {
             long seconds = 60;
             long millis = 1000;
             while (true) {
-                Thread.sleep(seconds * millis);
+                try {
+                    Thread.sleep(seconds * millis);
+                } catch (InterruptedException interruptedException) {
+                        System.err.println("Error starting H2");
```
suryaaki2, line 55: @riakivets Can you remove System.errr?

suryaaki2 (changes requested):

nvoxland (changes requested):

liquibase-core/src/main/java/liquibase/example/StartH2Main.java
riakivets: @suryaaki2 I can remove it. What is your concern to keep it?

liquibase-maven-plugin/src/main/java/org/liquibase/maven/plugins/LiquibaseChangeLogSyncSQLMojo.java
```diff
@@ -64,30 +64,34 @@ protected boolean isPromptOnNonLocalDatabase() {
 	@Override
 	protected Liquibase createLiquibase(Database db)
 			throws MojoExecutionException {
-		Liquibase liquibase = super.createLiquibase(db);
+		try (Liquibase liquibase = super.createLiquibase(db)) {
```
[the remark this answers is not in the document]
riakivets: Ok I got it, I will revert changes and add just SuppressWarnings for Sonar

liquibase-core/src/main/java/liquibase/util/xml/DefaultXmlWriter.java
```diff
@@ -19,6 +20,8 @@
     public void write(Document doc, OutputStream outputStream) throws IOException {
         try {
             TransformerFactory factory = TransformerFactory.newInstance();
+            factory.setAttribute(XMLConstants.ACCESS_EXTERNAL_DTD, "");
```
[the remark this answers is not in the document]
riakivets: I've reverted it

riakivets:
Will be reopen later

PR #1955, opened 2021-06-30
DAT-7230 Sonar issues

riakivets:


liquibase-core/src/main/java/liquibase/util/xml/DefaultXmlWriter.java
```diff
@@ -13,12 +14,15 @@
...
 import java.io.OutputStreamWriter;
 
+@SuppressWarnings("java:S2755")
 public class DefaultXmlWriter implements XmlWriter {
 
     @Override
     public void write(Document doc, OutputStream outputStream) throws IOException {
         try {
             TransformerFactory factory = TransformerFactory.newInstance();
+            factory.setAttribute(XMLConstants.ACCESS_EXTERNAL_DTD, "");
```
nvoxland, line 24: In the past, setting these has wrecked our ability to validate against external XSD files. Both our shipped xsd files and also 3rd party ones that users create for their own custom changes.

Have we tested that well?

nvoxland (changes requested):

liquibase-core/src/main/java/liquibase/util/xml/DefaultXmlWriter.java
suryaaki2: @XDelphiGrl Can we test this with external XSD files?

nvoxland: Yes, that is what we need to make sure to do. 
- Make sure XSD validation is working in general
- Add an external xsd and make sure it doesn't complain
- Add something that references an element from the external xsd and see that it's good

nvoxland (approved):

PR #1957, opened 2021-06-30
DAT-6972 Logif for dropAll for autoregister flow

riakivets:


liquibase-core/src/main/java/liquibase/command/core/InternalDropAllCommandStep.java
```diff
@@ -78,6 +75,7 @@ public void run(CommandResultsBuilder resultsBuilder) throws Exception {
             lockService.waitForLock();
 
             DatabaseChangeLog changeLog;
+            HubRegisterResponse hubRegisterResponse = null;
             if (StringUtil.isNotEmpty(commandScope.getArgumentValue(CHANGELOG_FILE_ARG))) {
```
nvoxland, line 79: Changelog_file_argument should really be removed as an argument to dropAll in general. If that is never specified, what behavior happens?

suryaaki2 (approved):

PR #2103, opened 2021-09-20
DAT-8203: allow supplying a current value when prompting using UIService

StevenMassaro:
Allow specifying a `currentValue` when using the `prompt` method in the `UIService`. If a non-null `currentValue` is provided, it will be shown instead of the `defaultValue`. If the input handler allows empty inputs, and the user strikes enter at the prompt when a `currentValue` is shown, that value will be returned.

### DIFF
--- a/liquibase-core/src/main/java/liquibase/ui/ConsoleUIService.java
+++ b/liquibase-core/src/main/java/liquibase/ui/ConsoleUIService.java
@@ -68,6 +68,11 @@ public class ConsoleUIService extends AbstractExtensibleObject implements UIServ
 
     @Override
     public <T> T prompt(String prompt, T defaultValue, InputHandler<T> inputHandler, Class<T> type) {
+        return prompt(prompt, defaultValue, null, inputHandler, type);
+    }
+
+    @Override
+    public <T> T prompt(String prompt, T defaultValue, T currentValue, InputHandler<T> inputHandler, Class<T> type) {
         //
         // Check the allowPrompt flag
         //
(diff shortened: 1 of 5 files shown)


suryaaki2 (approved):

nvoxland:
In the end, the change is so that there can be a difference in prompts between a "default" and a "current" value?

I worry about the complexity of the implementation requirement (when to return defaultValue vs. currentValue) and the fact that it is a breaking API change. 

Is there a prompt we can use besides default/current that would work for either use case? Or, would it make sense to have prompt configuration on the InputHandler?

nvoxland (changes requested):
In the end, this is to support sometimes having a "default" prompt and sometimes a "current" one?

I worry about the complexity of implementation and use (correctly using/passing the currentValue vs. defaultValue) as well as the fact that it is a breaking API change.

Is there a different wording we can use in the prompt to avoid needing to deal with default vs. current prompts? (From an end-user standpoint, what is that distinction trying to get across to them?)

Or, would it make sense to have the prompt configuration logic part of the InputHandler?

StevenMassaro:
> In the end, this is to support sometimes having a "default" prompt and sometimes a "current" one?
> 
> I worry about the complexity of implementation and use (correctly using/passing the currentValue vs. defaultValue) as well as the fact that it is a breaking API change.
> 
> Is there a different wording we can use in the prompt to avoid needing to deal with default vs. current prompts? (From an end-user standpoint, what is that distinction trying to get across to them?)
> 
> Or, would it make sense to have the prompt configuration logic part of the InputHandler?

I considered putting the prompt configuration logic inside the InputHandler, but I felt that would make it more confusing.

The distinction that we are trying to get across to a user is that pressing the "enter" key selects the _current_ value of a parameter for a quality check, not the default value.

Ultimately, this code complexity is not necessary. The issue here is that product does not like the word "default" showing up in the prompt when the value being shown to the user is not a default value.

nvoxland (approved):

PR #2134, opened 2021-10-01
fix: Use default localhost finder for MacOs - fix #2098

### LINKED ISSUE #2098: NetUtils.getLocalHostName() takes 30s
## Environment

**Liquibase Version**: 4.3.5

**Liquibase Integration & Version**: maven

**Operating System Type & Version**: MacOS 11.5.2 Big Sur

## Description

During liquibase startup, the call `database.getDefaultSchemaName();` in `liquibase.changelog.ChangeLogParameters.java:61` leads to resolving the current hostname by `liquibase.util.NetUtil.getLocalHostName` in `NetUtil.java:71`. The subsequent line 75 takes 30s to resolve the hostname for the IP given by line 73. Interestingly, line 73 doesn't return 127.0.0.1 (from network interface eth0) but another IP taken from another network interface (eth1 / company internal network IP, here '/192.168.254.80' - yes starting with the slash. This IP can't resolved as fast as normally done by java vm.

`liquibase.util.NetUtil.getLocalHostName()`
```java
InetAddress localhost = getLocalHost(); // => '/192.168.254.80' ... yes with the leading slash
if (localhost != null) {
    return localhost.getHostName(); // needs 30s
}
```

## Expected/Desired Behavior
Same as normal JVM resolution, tested by the follwing:
```java
    @Test
    void testResolveLocalhost() throws UnknownHostException {
        InetAddress ia = InetAddress.getLocalHost();
        String hostname = ia.getHostName(); // 6ms
    }

    @Test
    void testResolveLiquibase() throws UnknownHostException, SocketException {
        String ia = NetUtil.getLocalHostAddress(); //30s
    }
```

## Additional Context
There are solution on Stackoverflow for the general host resolving problem in JVM. ( https://stackoverflow.com/questions/39636792/jvm-takes-a-long-time-to-resolve-ip-address-for-localhost/39698914#39698914 ) But the solutiuons given there didn't work.

grzi:
This will fix #2098

The NetUtil retrieves a 'wrong' localhost addr on local, that is impossible to map under 30 seconds, using the standard way makes it instant.

- - -
#### Links
* Fixed Issue: https://github.com/liquibase/liquibase/issues/2098
* Test Results: https://github.com/liquibase/liquibase/pull/2134/checks

#### Testing
* Steps to Reproduce: https://github.com/liquibase/liquibase/issues/2098
* Guidance:
  * Impact: Changes hostname lookup used in a variety of places. Easiest to see may be in the databasechangeloglock table
  * Change unifies logic across linux/mac/windows, but the underlying java calls may act differently in different OS's
  * To ensure that the hostname is working correctly:
    *  run `liquibase update`
    * ctrl-c to kill liquibase after it has locked the database
    *  `liquibase list-locks` to ensure that the hostname is not localhost
  * To test speed, run a `liquibase update` with an empty changelog so that there is little to do. On mac, it should be about as fast as windows/linux

#### Dev Verification
Ran updated NetUtilTest on linux and windows. Don't have a mac to run it on

#### Test Requirements (Liquibase Internal QA)
Performance testing is out of scope for this issue. QA will validate that the correct hostname is used and that update is successful, but not how long it takes an update to run.

 * **Setup** 
-  You will need a Mac and a Windows machine to test this fix.
- The database platform does not matter but example snippets are for MySQL.
- On the database, ensure AUTOCOMMIT is disabled.
- Create a table, "accounts":

```
CREATE TABLE accounts (
    name VARCHAR(100) NOT NULL,
    balance DEC(15,2) NOT NULL
);
```

- In a database IDE, start an insert transaction to the accounts table:
```
BEGIN;

INSERT INTO accounts(name,balance)
VALUES('Abernathy',50000);
```

- Do not commit the insert!!
- Create a formatted SQL file named addColumn-changelog.sql with the following content:
```
--liquibase formatted sql

--changeset liquibase:lb2180
ALTER TABLE accounts
ADD COLUMN phone VARCHAR(15);`
```

**Manual Tests**

**TEST 1: Verify the DATABASECHANGELOGLOCK.LOCKEDBY has the test machine's name and _not_ localhost.**
 Start a Liquibase update to add a column to table accounts: 
 * `liquibase update --changelog-file addColumn-changelog.sql`
 
 In another command prompt, run:
*  `liquibase list-locks`
The list-locks command outputs the lock information with the machine name, for example:
 ` - LENOVO (172.21.224.1) at Dec 2, 2021 3:54:40 PM` , where LENOVO is the machine name.

**TEST 2: Verify an update operation is successful.**
Execute Liquibase update:
*  `liquibase update --changelog-file addColumn-changelog.sql`

* Automated Tests
  * No new functional or harness tests required for this fix.

┆Issue is synchronized with this [Jira Bug](https://datical.atlassian.net/browse/LB-2180) by [Unito](https://www.unito.io)

### DIFF
--- a/changelog.txt
+++ b/changelog.txt
@@ -1,6 +1,68 @@
 Liquibase Core Changelog
 ===========================================
 
+Changes in version 4.17.0 (2022.10.05)
+
+## Notable Changes
+* [PRO] The `liquibase flow` command and flow files are available with a Pro license. It allows you to create portable, platform-independent  workflows that can run across different tools without modifying each workflow.
+  -  Learn more about flow here: [liquibase flow file and command](https://docs.liquibase.com/commands/flow/flow.html)
+  -  If you want to provide feedback on this capability, please email mario@liquibase.com.
+
+* [PRO] Pro license users can now integrate Amazon S3 with Liquibase. You can extend Liquibase to use remote file locations to enable secure, centralized file management.
+  -  Learn more here: [S3 Remote Files](https://docs.liquibase.com/tools-integrations/remote-files/s3.html)
+* [PRO] Pro license users can use searchpath to find files located on S3 that cannot be found by an absolute path.
+  -  Learn more about the [Searchpath command here](https://docs.liquibase.com/concepts/changelogs/how-liquibase-finds-files.html).
+  -  Learn more about [Searchpath with S3 here](https://docs.liquibase.com/tools-integrations/remote-files/searchpath.html).
+
+## API Breaking Changes
+* The liquibase.resource.ResourceAccessor APIs were updated to take advantage of the new liquibase.resource.Resource abstraction that was introduced. Code that USES the existing ResourceAccessor methods will continue to work as before with no changes, but if you have written custom implementations of ResourceAccessor there are new and changed methods to implement. For more information, see [javadocs.liquibase.com](https://javadocs.liquibase.com)
+
+## Enhancements
+* Created new liquibase.resource.Resource interface by @nvoxland in https://github.com/liquibase/liquibase/pull/3064
+* Created ThreadLocalScopeManager by @nvoxland in https://github.com/liquibase/liquibase/pull/3240
+* Create liquibase_autocomplete.zsh by @szandany in https://github.com/liquibase/liquibase/pull/3130
+
+## Fixes
+
+* AlterSequence: include NOORDER clause ordered="false" is specified by @LeBezout in https://github.com/liquibase/liquibase/pull/1044
+* Upper case result columns only for case-insensitive databases by @fbiville in https://github.com/liquibase/liquibase/pull/3102
+* Postgresql: Fix for case when current search_path is an empty string by @gadget in https://github.com/liquibase/liquibase/pull/3233
+* Fix missing catalog on add column rollback by @jandurovec in https://github.com/liquibase/liquibase/pull/921
+* Fix an error which can occur if getAutoCommit is called on a closed connection by @gpsfl in https://github.com/liquibase/liquibase/pull/3135
+* apply sqlVisitors to CompoundStatements on DB2z by @MichaelKern-IVV in https://github.com/liquibase/liquibase/pull/3220
+* Improved message when precondition onFail : MARK_RAN is set by @MultiM25 in https://github.com/liquibase/liquibase/pull/2238
+* Add default catalog condition to SequenceSnapshotGenerator for Oracle by @gabrielnardes in https://github.com/liquibase/liquibase/pull/3152
+* Fixed issue with h2 loadUpdateData not correctly handling values with the string " values " in the inserted data  by @tomyy in https://github.com/liquibase/liquibase/pull/1831
+* Fix the smallserial issue not being parsed as a SmallInt type for Postgres by @MalloD12 in https://github.com/liquibase/liquibase/pull/3234
+* Remove snowflake unsupported logging from snapshot command by @abrackx in https://github.com/liquibase/liquibase/pull/3277
+* Fix "double" handling with H2  by @MalloD12 in https://github.com/liquibase/liquibase/pull/3273
+* Firebird: fix createIndex missing space around "computed" by @Markus-Patt in https://github.com/liquibase/liquibase/pull/1500
+* Test Race Condititons During Liquibase Locking by @schrieveslaach in https://github.com/liquibase/liquibase/pull/2327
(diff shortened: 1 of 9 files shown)


nvoxland:
I merged the current master branch into the fork. 

I also changed the implementation a bit in the fork and added more tests.

Since it seems that the `InetAddress.getLocalHost()` is so much faster, we are now always calling that regardless of the OS and then falling back to the "loop through the interfaces" setup only if the initial fast call returned a localhost address.

Testing on windows, the code didn't fall back into the loop through.
Testing on linux, the code DID fall through to the interface looping.
Both those behaviors were as they were before.

nvoxland (approved):

suryaaki2 (approved):

XDelphiGrl:
This PR removes OS-specific code to get the hostaddress and uses `InetAddress foundHost = InetAddress.getLocalHost();` for Mac, Windows and Linux. If the `foundHost` is a [loopback address](https://en.wikipedia.org/wiki/Localhost) or a [link-local address](https://en.wikipedia.org/wiki/Link-local_address), Liquibase iterates through the network interfaces to locate a more "descriptive" hostaddress. 

---

_Validate update-sql INSERT for DATABASECHANGELOGLOCK.LOCKEDBY includes hostname and hostaddress._
* Mac **FAIL** (see issues logged during test)

```
UPDATE PUBLIC.DATABASECHANGELOGLOCK SET LOCKED = TRUE, LOCKEDBY = '10.0.0.8 (10.0.0.8)', LOCKGRANTED = '2022-02-14 14:47:04.238' WHERE ID = 1 AND LOCKED = FALSE;
```

* Windows **PASS**

```
UPDATE testdb1.DATABASECHANGELOGLOCK SET `LOCKED` = 1, LOCKEDBY = 'LENOVO (10.0.0.11)', LOCKGRANTED = NOW() WHERE ID = 1 AND `LOCKED` = 0;
```

_Verify update using an empty changelog on a Mac takes less than 30 seconds._ **PASS**

` $ time liquibase update --changelog-file blank-changelog.sql`

| Trial 1  | Trial 2 | Trial 3 |
| ------------- | ------------- |------------- |
| real 0m2.377s  | real 0m2.376s   | real 0m2.471s |
| user 0m5.976s | user 0m5.831s  | user 0m5.522s
| sys	0m0.373s  | sys 0m0.356s    |sys	0m0.343s |

* See Limitations in Test Methodology for more information.

---

### Limitations in Test Methodology 

* The reproduction steps to see `localhost` in the DATABASECHANGELOGLOCK.LOCKEDBY column did not work for me. On Windows, the Liquibase CLI correctly gets the hostname + hostaddress (IP) and uses those to populate the LOCKEDBY value. On Mac, however, even with the fix, Liquibase only ever gets hostaddress and the LOCKEDBY value ends up being `hostaddress(hostaddress)`. This is entered as Issue 2539.

* There are no performance benchmarks available for empirical analysis of the speed of NetUtils.getLocalHost(). Anecdotally, my Mac performs Liquibase operations slower than on my Windows machine, but this is likely a moot point, as the two machines do not have equivalent system resources.  

* The tests in `liquibase-core/src/test/groovy/liquibase/util/NetUtilTest.groovy` ran significantly faster on master than on the fix branch; ~80ms on master and ~428ms on the fix branch. I am sending this back to @nvoxland to take a look.

---
**Issues Logged During Test (loosely related to this PR)**
[Issue 2539](https://github.com/liquibase/liquibase/issues/2539)
[Issue 2537](https://github.com/liquibase/liquibase/issues/2537)

---
**Test Environment**
Liquibase Core fix/slow_get_localhost_on_mac/1081/aa2358, Pro: master/343/d9abdc
Mac Big Sur (11.4)
Windows 10
[Passing Functional Tests]( )

nvoxland:
I don't see anything new in master that would make this be so much faster in master vs. here besides changes to this logic.

If you're seeing this code faster on you rmac vs the old version, @grzi , we'll have to do more research into why and what the best logic to use is. I don't want to just accept this PR and end up making it slower for a majority of people.

nvoxland:
I guess I see that there is some additional caching in NetUtil to avoid so many hostname/IP lookups. That must be the main change.

@MalloD12 fixed up the hostname handling so it will check localhost for a hostname when the external interface's hostname is not set. That is the code used by the databasechangeloglock description that you were seeing off, @XDelphiGrl

FBurguer:
Tested this PR against MacOS and it looks like its showing the hostname instead of localhost now.

PR #2105, opened 2021-09-21
get lock only if needed

VlasyukA:
## Environment

**Liquibase Version**: 4.4.3

**Liquibase Integration & Version**: Spring

**Liquibase Extension(s) & Version**: liquibase-core

**Database Vendor & Version**: Any

**Operating System Type & Version** Any:

## Pull Request Type

- [ ] Bug fix (non-breaking change which fixes an issue.)
- [x] Enhancement/New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to change)

## Description
Springliquibase lock only if have new changesets.
https://liquibase.jira.com/browse/CORE-3044

## Steps To Reproduce
SpringLiquibase doesn't get db lock if now new changesets

## Actual Behavior
SpringLiquibase always get db lock 

## Expected/Desired Behavior
Springliquibase lock only if have new changesets.

## Fast Track PR Acceptance Checklist:

- [x] Build is successful and all new and existing tests pass
- [x] Added [Unit Test(s)](https://liquibase.jira.com/wiki/spaces/LB/pages/1274937609/How+to+Write+Liquibase+Core+Unit+Tests)
- [ ] Added [Integration Test(s)](https://liquibase.jira.com/wiki/spaces/LB/pages/1276608569/How+to+Write+Liquibase+Core+Integration+Tests)
- [ ] Added [Test Harness Test(s)](https://github.com/liquibase/liquibase-test-harness/pulls)
- [ ] Documentation Updated

## Need Help?
Come chat with us on our [discord channel](https://discord.com/channels/700506481111597066/700506481572839505)

Issue #1453, opened 2020-10-05
Improve LockService functionality and flexibility


nvoxland:
## Description

Because concurrent update/rollback/etc. operations to a database will interfere with each other, Liquibase has a [LockService](https://github.com/liquibase/liquibase/blob/master/liquibase-core/src/main/java/liquibase/lockservice/LockService.java) which we use to coordinate access so only one instance can run against a database at time.

Examples of when you can run into concurrency trouble:
- Liquibase is embedded on startup of an application server, and you have a cluster of machines that may be starting at once
- Liquibase runs as part of a build job, and concurrent builds can happen at once

The [StandardLockService](https://github.com/liquibase/liquibase/blob/master/liquibase-core/src/main/java/liquibase/lockservice/StandardLockService.java) is the normal implementation that uses a `databasechangeloglock` table. Liquibase inserts a record and commits it at startup time, and if other instances see that record they wait until it is cleared.

The big problem with this standard implementation is that IF the instance doing the update is killed before it commits the clearing of the record, that value is stuck until `liquibase clearLocks` is ran. 

This problem has been around forever in Liquibase, but has been exasperated in recent years with systems like AWS that auto-kill processes that seem slow to start up. We wrap everything in try/catch blocks, but process killing happens too out-of-band for us to deal with.

## Related issues/PRs

Descriptions and proposed fixes:
- #829 
- #1417 
- #1311 
- Others?

#### BYO Locking

One option is to leave the locking to the users. There are many situations where Liquibase is ran within a system where the user knows only one instance can be ran at a time. For example, cloud providers have an "init" phase where the platform ensures just one container runs and Liquibase can be moved to there. Or, updates are done via a managed deployment script or build process which is ensuring only one version of an app is being deployed at once. 

In these cases, the lock service is just getting in the way. Ideally, the default lock service should work just fine and whether it runs unnecessarily or not isn't worth the hassle of turning it off. But, it's a fallback option to consider. We do have [the nodatabaselock](https://github.com/liquibase/liquibase-nochangeloglock) extension which could be folded into the main liquibase code with an easy flag to switch to that implementation

#### Database Session Based Locking

Databases have ways to lock tables for the duration of a connection that automatically unlocks them when the connection closes.

By using that functionality, we can rely on the database to clean up the lock. We don't even necessarily need the databasechangeloglock table anymore, because we can lock the databasechangelog table itself. 

[...]

Issue #1312, opened 2020-08-10
Liquibase is very slow on startup with spring boot


ggdn:
## Environment
**Liquibase Version**: 3.8.5
**Liquibase Integration & Version**: spring boot
**Liquibase Extension(s) & Version**: none
**Database Vendor & Version**: oracle v12
**Operating System Type & Version**: unix rhel v7

## Description
Hello 
Sometime Liquibase can take a lot of time to startup ~3 minutes without any changelog to apply. I don't know why, and the problem didn't occurred with the last version i used 3.6.3. I run Liquibase in spring boot applications.

## Steps To Reproduce
I don't know how to reproduce it. It appends randomly 

## Screenshots (if appropriate)
```
2020-08-09 06:40:16.956 [main] INFO  o.h.annotations.common.Version.<clinit> - HCANN000001: Hibernate Commons Annotations {5.1.0.Final}
2020-08-09 06:40:16.763 [main] INFO  org.hibernate.Version.logVersion - HHH000412: Hibernate Core {5.4.10.Final}
2020-08-09 06:40:16.677 [main] INFO  o.h.jpa.internal.util.LogHelper.logPersistenceUnitInformation - HHH000204: Processing PersistenceUnitInfo [name: default]
2020-08-09 06:40:16.558 [main] INFO  l.lockservice.StandardLockService.info - Successfully released change log lock
2020-08-09 06:40:16.493 [main] INFO  liquibase.executor.jvm.JdbcExecutor.info - SELECT * FROM DATABASECHANGELOG ORDER BY DATEEXECUTED ASC, ORDEREXECUTED ASC
2020-08-09 06:40:16.492 [main] INFO  l.c.StandardChangeLogHistoryService.info - Reading from DATABASECHANGELOG
2020-08-09 06:40:16.491 [main] INFO  liquibase.executor.jvm.JdbcExecutor.info - SELECT COUNT(*) FROM DATABASECHANGELOG
2020-08-09 06:40:16.483 [main] INFO  liquibase.executor.jvm.JdbcExecutor.info - SELECT MD5SUM FROM DATABASECHANGELOG WHERE MD5SUM IS NOT NULL AND ROWNUM=1
2020-08-09 06:35:51.224 [main] INFO  l.lockservice.StandardLockService.info - Successfully acquired change log lock
[...]

PR #2190, opened 2021-11-05
CORE-3044 avoid locking if no updates pending

RichardBradley:
Based on changes proposed by Hannu Hartikainen <hannu@hrtk.in>

See #2105, #846, #1453

## Environment
**Liquibase Version**: current / all

**Liquibase Integration & Version**: all

**Liquibase Extension(s) & Version**:  ??

**Database Vendor & Version**: all

**Operating System Type & Version**: all

## Pull Request Type

* [ ] Bug fix (non-breaking change which fixes an issue.)
* [x] Enhancement/New feature (non-breaking change which adds functionality)

Fixes https://liquibase.jira.com/browse/CORE-3044

## Description
If no changes are necessary, Liquibase still takes out an exclusive write lock on the DB.

This adds significant overhead in environments where multiple instances of the same service are booting in parallel, such as Cloud services that use Liquibase. 

It is better to check whether any changes are necessary while not holding the write lock and skip updates if possible. It is important to note that a second check is necessary after taking the write lock, to avoid a race condition between the read and the write (hence the name "double checked locking").

## Steps To Reproduce
Launch multiple instances in parallel of a service that uses Liquibase to manage db schema. 

## Actual Behavior
The on-boot liquibase check is unnecessarily serialized by the db write lock.

## Expected/Desired Behavior
In the case where no changelogs are needed to run, the services ought to be able to boot in parallel.

In the case where changelogs are needed, the services should not run any changelogs in parallel, i.e. locking is still needed.

n/a

## Additional Context
I have not added any integration tests, but some are probably warranted. Do you have instructions on how to run the integration tests? I couldn't find any.

* [I think so?] Build is successful and all new and existing tests pass
* [No, need help ] Added [Integration Test(s)](https://liquibase.jira.com/wiki/spaces/LB/pages/1276608569/How+to+Write+Liquibase+Core+Integration+Tests)
* [??? ] Added [Test Harness Test(s)](https://github.com/liquibase/liquibase-test-harness/pulls)
* [Is the locking behaviour documented? ] Documentation Updated

## Need Help?

- - -
#### Links
* Fixed Issue: Stand alone PR
* Test Results: https://github.com/liquibase/liquibase/pull/2190/checks

#### Testing
* Steps to Reproduce: Run `liquibase update-sql` with no changesets to deploy and see that there is no lock statements made
* Guidance:
  * Impact: Impacts only `update` operation

#### Dev Verification
Revied code

┆Issue is synchronized with this [Jira Bug](https://datical.atlassian.net/browse/LB-2203) by [Unito](https://www.unito.io)

### DIFF
--- a/liquibase-core/src/main/java/liquibase/Liquibase.java
+++ b/liquibase-core/src/main/java/liquibase/Liquibase.java
@@ -81,6 +81,7 @@ public class Liquibase implements AutoCloseable {
     private ChangeLogSyncListener changeLogSyncListener;
 
     private UUID hubConnectionId;
+    private Map<String, Boolean> upToDateFastCheck = new HashMap<>();
 
     private enum RollbackMessageType {
         WILL_ROLLBACK, ROLLED_BACK, ROLLBACK_FAILED
@@ -233,6 +234,10 @@ public class Liquibase implements AutoCloseable {
      */
     public void update(Contexts contexts, LabelExpression labelExpression, boolean checkLiquibaseTables) throws LiquibaseException {
         runInScope(() -> {
+            if (isUpToDateFastCheck(contexts, labelExpression)) {
+                return;
+            }
+
             LockService lockService = LockServiceFactory.getInstance().getLockService(database);
             lockService.waitForLock();
 
@@ -241,7 +246,7 @@ public class Liquibase implements AutoCloseable {
 
             Operation updateOperation = null;
             BufferedLogService bufferLog = new BufferedLogService();
-            DatabaseChangeLog changeLog = null;
+            DatabaseChangeLog changeLog;
             HubUpdater hubUpdater = null;
             try {
                 changeLog = getDatabaseChangeLog();
@@ -309,6 +314,41 @@ public class Liquibase implements AutoCloseable {
                 setChangeExecListener(null);
             }
         });
+    }
+
+    /**
+     * Performs check of the historyService to determine if there is no unrun changesets without obtaining an exclusive write lock.
+     * This allows multiple peer services to boot in parallel in the common case where there are no changelogs to run.
+     * <p>
+     * If we see that there is nothing in the changelog to run and this returns <b>true</b>, then regardless of the lock status we already know we are "done" and can finish up without waiting for the lock.
+     * <p>
+     * But, if there are changelogs that might have to be ran and this returns <b>false</b>, you MUST get a lock and do a real check to know what changesets actually need to run.
+     * <p>
+     * NOTE: to reduce the number of queries to the databasehistory table, this method will cache the "fast check" results within this instance under the assumption that the total changesets will not change within this instance.
+     */
+    protected boolean isUpToDateFastCheck(Contexts contexts, LabelExpression labelExpression) throws LiquibaseException {
+        String cacheKey = contexts +"/"+ labelExpression;
+        if (!this.upToDateFastCheck.containsKey(cacheKey)) {
+            try {
+                if (listUnrunChangeSets(contexts, labelExpression, false).isEmpty()) {
+                    LOG.fine("Fast check found no un-run changesets");
+                    upToDateFastCheck.put(cacheKey, true);
+                } else {
+                    upToDateFastCheck.put(cacheKey, false);
+                }
+            } catch (DatabaseException e) {
+                LOG.info("Error querying Liquibase tables, disabling fast check for this execution. Reason: " + e.getMessage());
+                upToDateFastCheck.put(cacheKey, false);
+            } finally {
+                // Discard the cached fetched un-run changeset list, as if
+                // another peer is running the changesets in parallel, we may
+                // get a different answer after taking out the write lock
+
+                ChangeLogHistoryService changeLogService = ChangeLogHistoryServiceFactory.getInstance().getChangeLogService(database);
+                changeLogService.reset();
+            }
+        }
+        return upToDateFastCheck.get(cacheKey);
     }
 
     /**
@@ -477,10 +517,14 @@ public class Liquibase implements AutoCloseable {
 
                 outputHeader("Update Database Script");
 
-                LockService lockService = LockServiceFactory.getInstance().getLockService(database);
-                lockService.waitForLock();
+                if (isUpToDateFastCheck(contexts, labelExpression)) {
+                    Scope.getCurrentScope().getSingleton(ExecutorService.class).getExecutor("logging", database).comment("Database is up to date, no changesets to execute");
+                } else {
+                    LockService lockService = LockServiceFactory.getInstance().getLockService(database);
+                    lockService.waitForLock();
 
-                update(contexts, labelExpression, checkLiquibaseTables);
+                    update(contexts, labelExpression, checkLiquibaseTables);
+                }
 
                 flushOutputWriter(output);
 
--- a/liquibase-core/src/test/groovy/liquibase/LiquibaseTest.groovy
+++ b/liquibase-core/src/test/groovy/liquibase/LiquibaseTest.groovy
@@ -1,7 +1,5 @@
 package liquibase
 
-import liquibase.changelog.ChangeLogHistoryService
-import liquibase.changelog.ChangeLogHistoryServiceFactory
 import liquibase.changelog.ChangeLogIterator
 import liquibase.changelog.ChangeSet
 import liquibase.changelog.DatabaseChangeLog
--- a/liquibase-core/src/test/groovy/liquibase/LiquibaseTest.groovy
+++ b/liquibase-core/src/test/groovy/liquibase/LiquibaseTest.groovy
@@ -1,6 +1,7 @@
 package liquibase
 
 import liquibase.changelog.ChangeLogIterator
+import liquibase.changelog.ChangeSet
 import liquibase.changelog.DatabaseChangeLog
 import liquibase.database.Database
 import liquibase.database.core.MockDatabase
@@ -93,8 +94,19 @@ class LiquibaseTest extends Specification {
 
         mockHubService = (MockHubService) Scope.currentScope.getSingleton(HubServiceFactory).getService()
         mockHubService.reset()
+        def databaseChangeLog = new DatabaseChangeLog(changeLogId: MockHubService.randomUUID.toString())
+        databaseChangeLog.addChangeSet(new ChangeSet(
+                "test",
+                "test",
+                false,
+                false,
+                "test",
+                "",
+                "",
+                databaseChangeLog));
         ChangeLogParserFactory.getInstance().register(new MockChangeLogParser(changeLogs: [
-                "com/example/changelog.mock": new DatabaseChangeLog(changeLogId: MockHubService.randomUUID.toString())
+                "com/example/changelog.mock":
+                        databaseChangeLog
         ]))
     }
 
--- a/liquibase-core/src/test/groovy/liquibase/LiquibaseTest.groovy
+++ b/liquibase-core/src/test/groovy/liquibase/LiquibaseTest.groovy
@@ -1,5 +1,7 @@
 package liquibase
 
+import liquibase.changelog.ChangeLogHistoryService
+import liquibase.changelog.ChangeLogHistoryServiceFactory
 import liquibase.changelog.ChangeLogIterator
 import liquibase.changelog.ChangeSet
 import liquibase.changelog.DatabaseChangeLog
@@ -501,6 +503,32 @@ class LiquibaseTest extends Specification {
 
         then:
         assertSqlOutputAppliesTags(writer.toString(), "1.1");
+    }
+
+    def validateContextLabelEntryHasNotBeenAddedPreviously() {
+        when:
+        h2Connection = getInMemoryH2DatabaseConnection()
+        Liquibase liquibase = createDatabaseAtTag(h2Connection, "1.0")
+        Contexts context = new Contexts("testContext")
+        LabelExpression label = new LabelExpression("testLabel")
+
+        then:
+        assertFalse(liquibase.isUpToDateFastCheck(context, label))
+
+    }
+
+    def validateContextLabelEntryHasBeenAddedPreviously() {
+        when:
+        h2Connection = getInMemoryH2DatabaseConnection()
+        Liquibase liquibase = new Liquibase("liquibase/test-changelog-fast-check.xml", new ClassLoaderResourceAccessor(),
+                h2Connection);
+        Contexts context = new Contexts("testContext")
+        LabelExpression label = new LabelExpression("testLabel")
+        liquibase.update();
+
+        then:
+        assertTrue(liquibase.isUpToDateFastCheck(context, label))
+
     }
 
     private JdbcConnection getInMemoryH2DatabaseConnection() throws SQLException {
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -248,6 +248,7 @@ jobs:
           - oracle
           - postgresql
           - sqlite
+          - firebird
     needs: build
     timeout-minutes: 30
     steps:
--- a/liquibase-core/pom.xml
+++ b/liquibase-core/pom.xml
@@ -113,6 +113,13 @@
         </dependency>
 
         <dependency>
+            <groupId>org.junit.jupiter</groupId>
+            <artifactId>junit-jupiter-params</artifactId>
+            <version>5.8.1</version>
+            <scope>test</scope>
+        </dependency>
+
+        <dependency>
             <groupId>com.opencsv</groupId>
             <artifactId>opencsv</artifactId>
             <version>5.7.1</version>
--- a/liquibase-core/src/main/java/liquibase/GlobalConfiguration.java
+++ b/liquibase-core/src/main/java/liquibase/GlobalConfiguration.java
@@ -31,6 +31,7 @@ public class GlobalConfiguration implements AutoloadedConfigurations {
     public static final ConfigurationDefinition<Boolean> SHOULD_SNAPSHOT_DATA;
     public static final ConfigurationDefinition<Boolean> PRESERVE_SCHEMA_CASE;
     public static final ConfigurationDefinition<Boolean> SHOW_BANNER;
+    public static final ConfigurationDefinition<Boolean> ALWAYS_DROP_INSTEAD_OF_REPLACE;
 
     public static final ConfigurationDefinition<DuplicateFileMode> DUPLICATE_FILE_MODE;
 
@@ -209,6 +210,10 @@ public class GlobalConfiguration implements AutoloadedConfigurations {
 
         SEARCH_PATH = builder.define("searchPath", String.class)
                 .setDescription("Complete list of Location(s) to search for files such as changelog files in. Multiple paths can be specified by separating them with commas.")
+                .build();
+
+        ALWAYS_DROP_INSTEAD_OF_REPLACE = builder.define("alwaysDropInsteadOfReplace", Boolean.class)
+                .setDescription("If true, drop and recreate a view instead of replacing it.")
                 .build();
     }
 
--- a/liquibase-core/src/main/java/liquibase/change/AbstractChange.java
+++ b/liquibase-core/src/main/java/liquibase/change/AbstractChange.java
@@ -780,9 +780,10 @@ public abstract class AbstractChange extends AbstractPlugin implements Change {
         SortedSet<String> names = new TreeSet<>();
         for (Map.Entry<String, ChangeParameterMetaData> entry : metaData.getParameters().entrySet()) {
             String lowerCaseKey = entry.getKey().toLowerCase();
-            if (lowerCaseKey.endsWith("name")
-                && !lowerCaseKey.contains("schema")
-                && !lowerCaseKey.contains("catalog")) {
+            if ((lowerCaseKey.endsWith("name")
+                        && !lowerCaseKey.contains("schema")
+                        && !lowerCaseKey.contains("catalog"))
+                    || lowerCaseKey.equals("path")) {
                 Object currentValue = entry.getValue().getCurrentValue(this);
                 if (currentValue != null) {
                     names.add(entry.getKey()+"="+ currentValue);
--- a/liquibase-core/src/main/java/liquibase/change/AbstractChange.java
+++ b/liquibase-core/src/main/java/liquibase/change/AbstractChange.java
@@ -23,6 +23,8 @@ import java.lang.reflect.Modifier;
 import java.lang.reflect.Type;
 import java.util.*;
 
+import static liquibase.statement.SqlStatement.EMPTY_SQL_STATEMENT;
+
 /**
  * Standard superclass to simplify {@link Change } implementations. You can implement Change directly, this class is
  * purely for convenience but recommended.
@@ -496,7 +498,7 @@ public abstract class AbstractChange extends AbstractPlugin implements Change {
             throw new RollbackImpossibleException(e);
         }
 
-        return statements.toArray(new SqlStatement[statements.size()]);
+        return statements.toArray(EMPTY_SQL_STATEMENT);
     }
 
     /**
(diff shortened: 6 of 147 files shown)


dancek:
Thanks for mentioning me but that's not really needed. The other PR is based on copy-pasted code I've written and that I didn't get credited for. This is new code that does smarter things (I think, but I've forgotten how liquibase works).

molivasdat:
Hi @RichardBradley Thanks for the PR.  Did the instructions in the Fast Track section provide enough details around integration tests?
What should be changed in the documentation to make it clearer?
What help do you need?

- Additional tests ([unit ](https://liquibase.jira.com/wiki/spaces/LB/pages/1274937609/How+to+Write+Liquibase+Core+Unit+Tests)and/or [integration](https://liquibase.jira.com/wiki/spaces/LB/pages/1276608569/How+to+Write+Liquibase+Core+Integration+Tests)) and/or [Test Harness Tests](https://github.com/liquibase/liquibase-test-harness/pulls)

RichardBradley:
> Hi @RichardBradley Thanks for the PR. Did the instructions in the Fast Track section provide enough details around integration tests?

No, I was not able to find any docs on how to run the integration tests.

What do you mean by "the Fast Track section"? I can't find "fast track" anywhere near here. I looked in the ["How to contribute code to Liquibase" guide](https://www.liquibase.org/community/contribute/code), the main readme and I searched google for "liquibase fast track"

> What help do you need?

Please could I have instructions on how to run the existing integration tests, and where might be best to add an integration test to cover this functionality? I have changed the main `update` method in the entry `Liquibase` class: what integration tests cover that?

Thanks,


Rich

nvoxland (approved):

nvoxland:
The changes seem good. Thanks for collecting up all the work from the related PRs!

There is probably no better tests to add at this point. We're working on an easier/more powerful testing framework currently but it's not quite ready yet, unfortuanately.

suryaaki2 (approved):

RichardBradley:
Thanks all, looking forward to seeing this in live!

> dancek commented on Nov 7, 2021
> Thanks for mentioning me but that's not really needed.

No problem. I created this patch largely by moving your code up one level, so thanks for the initial work on this

> molivasdat commented on Nov 23, 2021
> Hi @RichardBradley Thanks for the PR. Did the instructions in the Fast Track section provide enough details around integration tests?
> 
> RichardBradley commented on Nov 23, 2021
> What do you mean by "the Fast Track section"? I can't find "fast track" anywhere near here.

I see what you mean now, @molivasdat: the "Fast Track PR Acceptance Checklist" section in the PR comment template. That's a very useful page, thanks. I think nvoxland is right that this doesn't fit nicely into any integration tests at present as it's a layer below most of the code exercised by those (the change types etc.).

Perhaps someone could add a link to the "How to Write Liquibase Integration Tests" page from the "How to contribute code to Liquibase" guide and the main readme?

nvoxland:
In testing this PR, we found that it's not handling a few use cases correctly. Especially around running update-sql.

I'm in the middle of writing a suite of tests around the lock logic and will look at capturing those failing test cases in the automated tests as well as the fixes for it as part o that effort over the next couple weeks.

consultantleon:
We just upgraded to liquibase 4.8.0 and our tests for Oracle 12 run into massive delays, startup of tomcat /UI now takes around 7 minutes (instead of ~ 1 minute). Several tests timed out and the root cause seems to be that liquibase has become a lot slower. 
This was a reason not to upgrade to 4.8.0 way back, but as whitesource is now indicating a critical vulnerability we are forced to upgrade to at least 4.8.0 -> is there a version available without such performance penalty; what is the status of this issue?

RichardBradley:
> what is the status of this issue?

I haven't heard anything since the comment in Feb. As far as I know, this is ready to merge and should improve performance quite a bit for many users.

It won't change performance when changelogs _are_ being run (only when they don't need to be run and primarily for multi-server clusters), so it probably won't help you with your slow tests. 

As I explained [in this comment](https://github.com/liquibase/liquibase/issues/1312#issuecomment-1006927091), there was a separate Oracle perf issue which is unrelated to this change. The #1312 issue is marked as closed+fixed, but I haven't tested it myself as I don't use Oracle. It may be that there are still open Oracle perf issues that could be improved.

> In testing this PR, we found that it's not handling a few use cases correctly. Especially around running update-sql.

Any update on this, @nvoxland ? Anything I can help with? What were the unhandled cases?

nvoxland:
Doing a bit of testing with it, the problem case I found was doing an update-sql when there are no changes generates a script of:

```
--  *********************************************************************
--  Update Database Script
--  *********************************************************************
--  Change Log: com/example/changelog.workspace.xml
--  Ran at: 9/6/22 3:27 PM
--  Against: lbuser@172.22.0.1@jdbc:mysql://localhost:33062/lbcat
--  Liquibase version: [local build]
--  *********************************************************************

--  Lock Database
UPDATE lbcat.DATABASECHANGELOGLOCK SET `LOCKED` = 1, LOCKEDBY = 'thinkpad (192.168.1.34)', LOCKGRANTED = NOW() WHERE ID = 1 AND `LOCKED` = 0;
```

which just locks the database with no unlock.

If there are changes to apply, it ends up working correctly, like this:

```
--  *********************************************************************
--  Update Database Script
--  *********************************************************************
--  Change Log: com/example/changelog.workspace.xml
--  Ran at: 9/6/22 3:28 PM
--  Against: lbuser@172.22.0.1@jdbc:mysql://localhost:33062/lbcat
--  Liquibase version: [local build]
--  *********************************************************************

--  Lock Database
UPDATE lbcat.DATABASECHANGELOGLOCK SET `LOCKED` = 1, LOCKEDBY = 'thinkpad (192.168.1.34)', LOCKGRANTED = NOW() WHERE ID = 1 AND `LOCKED` = 0;

--  Changeset com/example/changelog.workspace.xml::2::example
CREATE TABLE lbcat.test_table2 (id INT NULL);

INSERT INTO lbcat.DATABASECHANGELOG (ID, AUTHOR, FILENAME, DATEEXECUTED, ORDEREXECUTED, MD5SUM, `DESCRIPTION`, COMMENTS, EXECTYPE, CONTEXTS, LABELS, LIQUIBASE, DEPLOYMENT_ID) VALUES ('2', 'example', 'com/example/changelog.workspace.xml', NOW(), 2, '8:054a4683a59bacad0c3ba66a5a18eaaf', 'createTable tableName=test_table2', '', 'EXECUTED', NULL, NULL, 'DEV', '2496133708');

--  Release Database Lock
UPDATE lbcat.DATABASECHANGELOGLOCK SET `LOCKED` = 0, LOCKEDBY = NULL, LOCKGRANTED = NULL WHERE ID = 1;
```

I pushed a change that applies the logic to both update and updateSql using a new common method on `liquibase.Liquibase`.

One potential performance drawback to this is that we can end up selecting from databasechangelog twice. My additional commit added caching to make sure it's at most twice since the updateSql path goes through the new `isUpToDateFastCheck` a couple times. But there can still be the query in `isUpToDateFastCheck` and then the actual query once we have the lock. For people with a large databasechangelog table that may be an issue, but I'm good with seeing what real life impact this ends up having since it should be helping them as well and that query **_should_** generally be fast.

Let me know if anyone has comments on my changes.

Things to be aware of:
- Only impacts update/update-sql, not rollback or any other operations

Things to worry about:
- Nothing

liquibase-core/src/main/java/liquibase/Liquibase.java
```diff
@@ -285,6 +288,38 @@ public void update(Contexts contexts, LabelExpression labelExpression, boolean c
...
+     * <p>
+     * If we see that there is nothing in the changelog to run and this returns <b>true</b>, then regardless of the lock status we already know we are "done" and can finish up without waiting for the lock.
+     * <p>
+     * But, if there are changelogs that might have to be ran and this returns <b>false</b>, you MUST get a lock and do a real check to know what changesets actually need to run.
+     * <p>
+     * NOTE: to reduce the number of queries to the databasehistory table, this method will cache the "fast check" results within this instance under the assumption that the total changesets will not change within this instance.
+     */
+    protected boolean isUpToDateFastCheck(Contexts contexts, LabelExpression labelExpression) throws LiquibaseException {
+        String cacheKey = contexts +"/"+ labelExpression;
+        if (!this.upToDateFastCheck.containsKey(cacheKey)) {
```
RichardBradley, line 303: Not a big deal, but since Java 8, there's a "[computeIfAbsent](https://docs.oracle.com/javase/8/docs/api/java/util/Map.html#computeIfAbsent-K-java.util.function.Function-)" method on Map that lets you go a "get-or-put" op atomically without needing to do the map lookup twice:

```
return this.upToDateFastCheck.computeIfAbsent(cacheKey, k -> {
      return ... // compute uncached value here
})```

File: liquibase-core/src/main/java/liquibase/executor/jvm/ColumnMapRowMapper.java [hunk 2]
```diff
@@ -24,6 +24,12 @@
 @SuppressWarnings({"unchecked"})
 public class ColumnMapRowMapper implements RowMapper {
 
+    private final boolean caseSensitiveDatabase;
```
RichardBradley, line 27: This caseSensitive change looks unrelated to this PR #2190 ?
Probably no harm to include it if you were planning to add this anyway, ofc

RichardBradley:
Looks good to me, thanks for fixing this up.

> One potential performance drawback to this is that we can end up selecting from databasechangelog twice. ...  For people with a large databasechangelog table that may be an issue, but I'm good with seeing what real life impact this ends up having since it should be helping them as well and that query should generally be fast.

Yes, agreed. Hopefully the downsides will be outweighed by the improvements. In the 99% case where the app boots with no changes to apply, people with an excessively large databasechangelog table won't see any difference.

liquibase-core/src/main/java/liquibase/executor/jvm/ColumnMapRowMapper.java
nvoxland: Thanks for noticing that. Yes, I hadn't cleaned my master branch well enough before updating yours. 

That PR should go in first, so I won't bother fixing the git history and force pushing. So it should be fine as it is.

liquibase-core/src/main/java/liquibase/Liquibase.java
nvoxland: Good point. I always forget computeIfAbsent saves the value that gets computed...

I'm good just leaving it as it is for now and we can do a general fixup of that pattern across the codebase at some point.

RichardBradley:
> there can be a lot of different existing setups and I'm not sure this quick check before all the init logic is going to be robust enough.

The "fast check" calls directly through to the existing codebase at `Liquibase.listUnrunChangeSets`, so any unusual setups should be handled in there, if they are handled at all, I would think?

> we should probably be doing things like checking for md5sum versions and other "is the setup correct" which I don't think is being done.

Again, if this is currently part of `Liquibase.listUnrunChangeSets` then it will be included for free. If it's not then I don't see why we should start doing it now?

>  With the current code, it ends up breaking because the orderexecuted column isn't there yet

What happens when the current version of `listUnrunChangeSets` is run without "fast check" in such a scenario?

I suppose if that code transparently updates the table structure, then it perhaps ought to be done inside Liquibase's lock. I hadn't realised that `listUnrunChangeSets` might be a write op. Perhaps it needs a "read only" flag when it is being called from within the fast check?

> The last thing is that the check is in the liquibase.Liquibase interface which is one facade for working with liquibase, but not the only way anymore.

What other ways are there?

> I'm thinking this "fast check" needs to be moved more into the LockService interface itself so it can be more consistently used and database-specific cases (like the postgresql error handling) more extensible.

It really doesn't feel to me like it's part of "locking". It is part of the logic of Updating and it uses locks. If we move it inside LockService then you will be mixing the scopes of a) examining the changesets, b) deciding whether to run anything and c) locking against concurrent access.

The high level algo should look like:

```
update() {
  if (listUnrunChangeSets(readonly = true) is not empty) {
    lock {
      run (listUnrunChangeSets(readonly = false))
    }
  }
}
```

>the orderexecuted column isn't there yet ... some databases like postgresql don't handle exceptions in queries well without a connection reset

That doesn't sound to me like it should be a problem -- I wouldn't expect that selecting from a non-existent column would leave a psql connection in an unusable state, but you will certainly know more than me on that front. I will make sure the integration test covers this case.


I'm happy to take a look at fixing up the code for this edge case, if you think the general approach is still sound, given my comments above?

filipelautert:
Hi @RichardBradley - I changed liquibase check tables flag to true and with that all the tests are passing. Did you set it to false for any reason? This way we don't need to have any try /catch or extra handling.

@nvoxland By now I would be fine to keep this code here as early next year we may review it when we refactor it to the command framework... what do you think?

RichardBradley:
You didn't respond to my queries about these issues from my comment of Sep 27

> Hi @RichardBradley - I changed liquibase check tables flag to true and with that all the tests are passing. Did you set it to false for any reason? This way we don't need to have any try /catch or extra handling.

I think I took the impression that check tables=true would cause the check code to write to the tables if they needed fixing, meaning that it should only be called inside the lock, otherwise one would fail if two servers reached this code at the same time (in the case where the tables do need updating). Is that true? It has been a while since I had a close look.

If that is true then we have two choices:
 A: do the unlocked first check with check tables = false
 B: do the unlocked first check with check tables = true

In the case where the liquibase tables do need updating (which must be *very* rare, right? When is the last time that the liquibase core table structure changed?), then in the case where two servers boot at the same time:

 * With option A, both of those servers may fail to start if reading tables of the wrong structure does indeed leave the db connection in an invalid state. I'm not sure I believe this though. Do you have more details or a repro? I wouldn't expect a "select x from y" call where table y had no column x to completely invalidate a db connection.

 * With option B, there will be an unlocked write. One server will succeed and the other will fail. I suppose the cluster will continue to make forward progress though, so maybe this is better.

> @nvoxland By now I would be fine to keep this code here as early next year we may review it when we refactor it to the command framework... what do you think?

If you are about to rework all the startup code then it probably does make sense to park this. 

Please don't just park this indefinitely if that is not imminent though. I think it's worth reiterating that this change would be a small but significant improvement to many thousands of Liquibase users across the world. When Liquibase was first written, servers would live for a long time and clusters of multiple identical servers were rare. These days, I would expect that most projects using Liquibase in production will use clusters of VMs that get recreated regularly and in parallel. At present, each of those VM startups will do a write op to take a lock; then do a read; discover that there are no migrations needed (99% of the time); then do another write op to release the lock. With this change, we will skip those two write ops entirely. Across each of the thousands or millions of VM restarts using this code across the world that will save quite a bit of compute power.

filipelautert:
> > there can be a lot of different existing setups and I'm not sure this quick check before all the init logic is going to be robust enough.
> 
> The "fast check" calls directly through to the existing codebase at `Liquibase.listUnrunChangeSets`, so any unusual setups should be handled in there, if they are handled at all, I would think?
> 
Yes, but it just handle those changes when we ask it to check Liquibase tables that was set to false.

> > we should probably be doing things like checking for md5sum versions and other "is the setup correct" which I don't think is being done.
> 
> Again, if this is currently part of `Liquibase.listUnrunChangeSets` then it will be included for free. If it's not then I don't see why we should start doing it now?
In fact the update code does that at line 254 , but fast checks runs before it.

> 
> > With the current code, it ends up breaking because the orderexecuted column isn't there yet
> 
> What happens when the current version of `listUnrunChangeSets` is run without "fast check" in such a scenario?
> 
> I suppose if that code transparently updates the table structure, then it perhaps ought to be done inside Liquibase's lock. I hadn't realised that `listUnrunChangeSets` might be a write op. Perhaps it needs a "read only" flag when it is being called from within the fast check?

Yes, you are right. I missed this condition when I set the flag to true.

> 
> > The last thing is that the check is in the liquibase.Liquibase interface which is one facade for working with liquibase, but not the only way anymore.
> 
> What other ways are there?

We are moving code to the UpdateCommandStep, but so far I think all locations are calling Liquibase.update code.

> 
> > I'm thinking this "fast check" needs to be moved more into the LockService interface itself so it can be more consistently used and database-specific cases (like the postgresql error handling) more extensible.
> 
> It really doesn't feel to me like it's part of "locking". It is part of the logic of Updating and it uses locks. If we move it inside LockService then you will be mixing the scopes of a) examining the changesets, b) deciding whether to run anything and c) locking against concurrent access.
> 
> The high level algo should look like:
> 
> ```
> update() {
>   if (listUnrunChangeSets(readonly = true) is not empty) {
>     lock {
>       run (listUnrunChangeSets(readonly = false))
>     }
>   }
> }
> ```
> 

So let's keep it here, and if we notice it can be used somewhere else we can think about refactoring.

> > the orderexecuted column isn't there yet ... some databases like postgresql don't handle exceptions in queries well without a connection reset
> 
> That doesn't sound to me like it should be a problem -- I wouldn't expect that selecting from a non-existent column would leave a psql connection in an unusable state, but you will certainly know more than me on that front. I will make sure the integration test covers this case.

Postgresql usually does that if you have problems with dml requests, not queries. I tested the try/catch approach and it worked fine with Postgresql too.

filipelautert:
> You didn't respond to my queries about these issues from my comment of Sep 27
> 
Sorry, just did it.

> > Hi @RichardBradley - I changed liquibase check tables flag to true and with that all the tests are passing. Did you set it to false for any reason? This way we don't need to have any try /catch or extra handling.
> 
> I think I took the impression that check tables=true would cause the check code to write to the tables if they needed fixing, meaning that it should only be called inside the lock, otherwise one would fail if two servers reached this code at the same time (in the case where the tables do need updating). Is that true? It has been a while since I had a close look.
> 
> If that is true then we have two choices: A: do the unlocked first check with check tables = false B: do the unlocked first check with check tables = true
> 
> In the case where the liquibase tables do need updating (which must be _very_ rare, right? When is the last time that the liquibase core table structure changed?), then in the case where two servers boot at the same time:
> 
> * With option A, both of those servers may fail to start if reading tables of the wrong structure does indeed leave the db connection in an invalid state. I'm not sure I believe this though. Do you have more details or a repro? I wouldn't expect a "select x from y" call where table y had no column x to completely invalidate a db connection.
> * With option B, there will be an unlocked write. One server will succeed and the other will fail. I suppose the cluster will continue to make forward progress though, so maybe this is better.

Changes in Liquibase tables are very rare as they are well defined and tested by now, so as  this patch aims to make things faster AND I wasn't able to reproduce any connection problems with wrong sqls I think it is best going with option **A**.  B would be fine as the concurrent code is short and fast, but we would be playing with luck and hopping that all servers in the internet are well configured, until someday someone would open an issue about it and we would try to patch it without the context that we have here. 

I pushed the code using try/catch so we can see the tests passing in all databases.

> 
> > @nvoxland By now I would be fine to keep this code here as early next year we may review it when we refactor it to the command framework... what do you think?
> 
> If you are about to rework all the startup code then it probably does make sense to park this.
> 
> Please don't just park this indefinitely if that is not imminent though. I think it's worth reiterating that this change would be a small but significant improvement to many thousands of Liquibase users across the world. When Liquibase was first written, servers would live for a long time and clusters of multiple identical servers were rare. These days, I would expect that most projects using Liquibase in production will use clusters of VMs that get recreated regularly and in parallel. At present, each of those VM startups will do a write op to take a lock; then do a read; discover that there are no migrations needed (99% of the time); then do another write op to release the lock. With this change, we will skip those two write ops entirely. Across each of the thousands or millions of VM restarts using this code across the world that will save quite a bit of compute power.

The commands code will be something that will take time to rework, so let's get this Pr to master and hopefully in time for the next release👍

filipelautert (approved):
Speeds up update command execution when there are no changesets to be applied.

liquibase-core/src/test/groovy/liquibase/LiquibaseTest.groovy
```diff
@@ -93,8 +94,19 @@ class LiquibaseTest extends Specification {
 
         mockHubService = (MockHubService) Scope.currentScope.getSingleton(HubServiceFactory).getService()
         mockHubService.reset()
+        def databaseChangeLog = new DatabaseChangeLog(changeLogId: MockHubService.randomUUID.toString())
+        databaseChangeLog.addChangeSet(new ChangeSet(
```
MalloD12, line 98: Hi @RichardBradley, if I'm not misunderstanding this part of the code. Are we not missing the `isUpToDateFastCheck()` here? If not could you please clarify to me how this test works? 

Thanks,
Daniel.

RichardBradley:
Thanks, @filipelautert , that all sounds good to me

liquibase-core/src/test/groovy/liquibase/LiquibaseTest.groovy
RichardBradley: (This comment is based on my recollection only; I haven't re-run this code since I made this PR in Nov 2021.)

This change is not directly relevant to this PR.

When working on this PR, I found that this test was failing already. The failure was in unrelated code when constructing a changelog with no changesets in it. I needed to add an empty changeset to the changelog in order to fix the test.

> Are we not missing the isUpToDateFastCheck() here? 

No, client code should not directly invoke "isUpToDateFastCheck", but should always call the public methods on Liquibase such as update()

>  If not could you please clarify to me how this test works?

I'm not exactly sure what this test is attempting to verify. It fails before this change and passes after it, so I thought it was an improvement.

MalloD12: Ohh I see, I thought this test was intended to test this change (cc: @XDelphiGrl ). Thanks for clarifying and thank you for your contributions on this PR. 

Daniel.

XDelphiGrl: Thanks for the response, @RichardBradley - fixing broken tests is always a good thing!

@MalloD12, can we add a test for `isUpToDateFastCheck`?

MalloD12: Yeap,  we can do that!

MalloD12: I've added two unit test to validate `upToDateFastCheck()` method. One scenario to validate a context-label entry does not exists and one to validate it does.

Could you @XDelphiGrl  please review this when you have a chance?

Thanks,
Daniel.

liquibase-core/src/test/groovy/liquibase/LiquibaseTest.groovy
```diff
@@ -491,6 +503,32 @@ class LiquibaseTest extends Specification {
         assertSqlOutputAppliesTags(writer.toString(), "1.1");
     }
 
+    def validateContextLabelEntryHasNotBeenAddedPreviously() {
```
XDelphiGrl, line 506: Thanks for adding these tests, @MalloD12! Should we have a test for the case where there are no labels and no contexts? I'm unclear as to what should happen in that case.

MalloD12: Hi Erz, I think that scenario is not needed because if I'm not wrong if we provide a changeset with context/label as empty strings will work in the same way as with any string values because it will then create a key to validate whether changeset has been already processed or not without the need to acquire a lock. Here is part of the logic I'm talking about:
```
 String cacheKey = contexts +"/"+ labelExpression;
 if (!this.upToDateFastCheck.containsKey(cacheKey)) {
<some code>
}
```
Now, if you meant without providing context/label properties in the changeset then this method would not be called because this is going through the path of this update:
`public void update(Contexts contexts, LabelExpression labelExpression, boolean checkLiquibaseTables)` 

Thanks,
Daniel.

XDelphiGrl: @MalloD12, hi - It was the second case I was referring to. Thanks for explaining that case will not go through the same code.

XDelphiGrl (approved):
This clever PR prevents the serialization of the on-boot liquibase check by avoiding the db write lock in cases where no changes need to be deployed. This allows multiple instances of Liquibase writing to the same DATABASECHANGELOG to boot without blocking one another if no changes need to be deployed. When there are changes to deploy, locking & blocking enforces serialization.

Thank you, @RichardBradley, for this awesome enhancement!

APPROVED

troligsownder:
This improvements are not used anymore since 4.21.0. I created a bug report for this: https://github.com/liquibase/liquibase/issues/4312

Issue #2082, opened 2021-09-09
IDEA logs contains errors


baev:
## Environment

**Liquibase Version**: 4.4.3

**Liquibase Integration & Version**: <Pick one: CLI, maven, gradle, spring boot, servlet, etc.>

**Liquibase Extension(s) & Version**: 

**Database Vendor & Version**:

**Operating System Type & Version**:

## Description

Every millisecond (!!!) the following error is appeared in IDEA's logs:

```
2021-09-08 12:48:43,010 [418783120]  ERROR - red.platform.impl.ChunkManager - IntelliJ IDEA 2021.1.2  Build #IU-211.7442.40 
2021-09-08 12:48:43,010 [418783120]  ERROR - red.platform.impl.ChunkManager - JDK: 11.0.11; VM: Dynamic Code Evolution 64-Bit Server VM; Vendor: JetBrains s.r.o. 
2021-09-08 12:48:43,010 [418783120]  ERROR - red.platform.impl.ChunkManager - OS: Mac OS X 
2021-09-08 12:48:43,011 [418783121]  ERROR - red.platform.impl.ChunkManager - Last Action: GotoDeclaration 
2021-09-08 12:48:43,011 [418783121]  ERROR - red.platform.impl.ChunkManager - /Users/baev/.gradle/caches/modules-2/files-2.1/org.liquibase/liquibase-core/4.4.3/20bcb8d97a8619c7d184f03d7247dfb85f8a9f7a/liquibase-core-4.4.3-sources.jar!/var/jenkins/workspace/liquibasePro/4.4.x/53/liquibase/liquibase-dist/target/archive/GETTING_STARTED.txt 
java.io.FileNotFoundException: /Users/baev/.gradle/caches/modules-2/files-2.1/org.liquibase/liquibase-core/4.4.3/20bcb8d97a8619c7d184f03d7247dfb85f8a9f7a/liquibase-core-4.4.3-sources.jar!/var/jenkins/workspace/liquibasePro/4.4.x/53/liquibase/liquibase-dist/target/archive/GETTING_STARTED.txt
	at com.intellij.openapi.vfs.impl.ZipHandlerBase.getEntryCrc(ZipHandlerBase.java:83)
	at com.intellij.openapi.vfs.impl.jar.JarFileSystemImpl.getEntryCrc(JarFileSystemImpl.java:143)
[...]

PR #2224, opened 2021-11-17
Fix javadoc and source generation

nvoxland:
## Description

Improvements to the github actions build to correctly capture *-javadoc.jar and *-sources.jar files into the liquibase-artifacts.

Prior to this PR, the javadocs files were missing and the liquibase-0-SNAPSHOT-sources.jar file was empty. Also prior to this PR, the sources.jar incorrectly had complete paths in it (#2082)

## Other changes

**github-action-* pattern**
Because this change impacts build.yml but PRs are built via pull_request_target, I added a `github-action-*` pattern to `push` and `pull_request_target`, so if we make local branches called `github-action-...` like this one is, we'll still get a build from it. 

Is there a better way to be able to test workflow logic in a branch? We still get a double build now (push and pull_request_target trigger), but I think once this pull_request_target gets into master that will go away

**Create Release logic**
This PR also includes fixed up logic for doing the creating a release. That is currently failing when I tested it, possibly because of the double actions being ran still in this PR. That is all triggered from a separate workflow, so if nothing looks weird in the code review it can be safely merged and then I'll keep working on it in a separate branch after merge.

**Zip file includes branch name in filename**

Makes it easy to track what we've downloaded.
---

#### Links

- Fixed Issue: 
  - https://github.com/liquibase/liquibase/issues/2082
  - See above
- Test Results: https://github.com/liquibase/liquibase/pull/2224/checks

#### Testing

- Steps to Reproduce: See above
- Guidance:
  - Impact: only impacts creation of sources and javadoc jar files
  - Testing is around what is in the liquibase-artifacts.zip file: Are the javadoc files there? Are the source files there? If you unzip the javadocs.jars (aka zips) and open index.html does it look right? No pro code in sources and no pro code docs in javadocs.

#### Dev Verification

- Verified that javadocs.jars and sources.jars look right


jnewton03:
adding @mcred and his big github actions brain to check this out as well

jnewton03 (approved):

suryaaki2 (approved):

PR #2232, opened 2021-11-23
Include org name in version info

nvoxland:
## Description

Development builds include the branch in the version info, but not the owner. That is useful information when using a build from a fork.


nvoxland:
Going to do this as part of a different PR

PR #2258, opened 2021-12-08
Removed gpg signing from pom since it is now in the build logic

nvoxland:
## Description

The gpg config in the pom files should be unnecessary now.

Creating PR to test out that out.


nvoxland:
Asc files were still created fine in test PR. Merging unneeded code

PR #2265, opened 2021-12-10
Interactive prompting of init project command (DAT-8804)

StevenMassaro:
## Environment

**Liquibase Version**:

## Pull Request Type

## Description

## Steps To Reproduce

- Please be precise and ensure private information is redacted
- Include things like
  - Files used - sql scripts, changelog file(s), property file(s), config files, POM Files
  - Exact commands used - CLI, maven, gradle, spring boot, servlet, etc.

## Actual Behavior

## Expected/Desired Behavior

- [ ] Added [Unit Test(s)](https://liquibase.jira.com/wiki/spaces/LB/pages/1274937609/How+to+Write+Liquibase+Core+Unit+Tests)
- [ ] Documentation Updated

## Need Help?


liquibase-core/src/main/java/liquibase/command/CommandDefinition.java
```diff
@@ -17,7 +17,7 @@
      */
     private final SortedSet<CommandStep> pipeline;
 
-    private final SortedMap<String, CommandArgumentDefinition<?>> arguments = new TreeMap<>();
+    private final Map<String, CommandArgumentDefinition<?>> arguments = new LinkedHashMap<>();
```
nvoxland, line 20: I don't think changing this definition to be insertion order for the purposes of prompting is good. As we get to a better CommandStepPipline refactor, the insertion order will tend toward the more random feeling. Also, if/when other features are interested in the arguments in different orders than what the prompting requires we'll have conflicts.

I think it's best to keep this in the simple alphabetical order, and figure out a way to have the logic of the order in which to prompt be more orthogonal.

One idea: allow the interactivity logic to be able to say what configurations they depend on being set, then use DependencyUtil logic to dynamically figure out the order?

liquibase-cli/src/main/java/liquibase/integration/commandline/LiquibaseCommandLine.java
```diff
@@ -467,6 +468,7 @@ public int getPrecedence() {
         } else {
             Scope.getCurrentScope().getLog(getClass()).fine("Cannot find local defaultsFile " + defaultsFile.getAbsolutePath());
         }
+        liquibaseConfiguration.registerProvider(new InteractivePromptingValueProvider());
```
nvoxland, line 471: I like the idea of plugging interactions into the ConfigurationValueProvider system.

liquibase-core/src/main/java/liquibase/command/CommandArgumentDefinition.java
```diff
@@ -146,6 +150,28 @@ public String toString() {
...
     }
 
+    /**
+     * Interactively prompt in the CLI for a user to supply a value for this command argument.
+     * @param previouslyPromptedValues the values already provided by the user for other arguments in this interactive
+     *                                 prompting session
+     * @return the value provided by the user, or null if the shouldPrompt method of {@link InteractivePromptableCustomizationWrapper}
+     * signifies that this prompt should not occur
+     */
+    public DataType interactivelyPrompt(Map<String, Object> previouslyPromptedValues) {
```
nvoxland, line 160: Why are we having to pass the previouslyPromptedValues here?

liquibase-core/src/main/java/liquibase/command/CommandArgumentDefinition.java
```diff
@@ -146,6 +150,28 @@ public String toString() {
...
+        List<DynamicRuleParameter> collect = previouslyPromptedValues.entrySet().stream().map(v -> new DynamicRuleParameter(v.getKey(), v.getValue())).collect(Collectors.toList());
```
nvoxland, line 161: This seems to be hard-coding in the DynamicRuleParameter here in generic/API code? This layer of the code shouldn't have rule-specific references in it?

liquibase-core/src/main/java/liquibase/command/CommandArgumentDefinition.java
```diff
@@ -217,6 +243,11 @@ public String toString() {
             return this.defaultValue(defaultValue, description);
         }
 
+        public Building<DataType> interactivePrompt(InteractivePromptableCustomizationWrapper promptableCustomizationWrapper) {
```
nvoxland, line 246: I like having the promotability defined as part of the CommandArgumentDefinition. But, the InteractivePromptableCustomizationWrapper name seems a bit unweildy

liquibase-core/src/main/java/liquibase/command/CommandFactory.java
```diff
@@ -13,7 +13,7 @@
 public class CommandFactory implements SingletonObject {
     private Collection<CommandStep> allInstances;
 
-    private final Map<String, Set<CommandArgumentDefinition<?>>> commandArgumentDefinitions = new HashMap<>();
+    private final Map<String, Set<CommandArgumentDefinition<?>>> commandArgumentDefinitions = new LinkedHashMap<>();
```
nvoxland, line 16: Same issue with LinkedHashMap as above

liquibase-core/src/main/java/liquibase/command/CommandScope.java
```diff
@@ -134,6 +137,46 @@ public CommandResults execute() throws CommandExecutionException {
...
+            boolean commandHasAnyInteractivelyPromptableParameters = commandDefinition.getArguments().values().stream().anyMatch(CommandArgumentDefinition::isInteractivelyPromptable);
+            if (commandHasAnyInteractivelyPromptableParameters) {
+                // If the command does have any interactively promptable parameters, then generate a list of parameters which
+                // are interactively promptable and for which the user did not explicitly provide a value.
+//                SortedSet<ConfigurationDefinition<?>> registeredDefinitions = LiquibaseConfiguration.getInstance().getRegisteredDefinitions(true);
+                List<CommandArgumentDefinition<?>> interactivelyPromptableParametersWithDefaultValue =
+                        commandDefinition.getArgumentsInInsertionOrder().values().stream().filter(cad -> (getConfiguredValue(cad).wasDefaultValueUsed() || !getConfiguredValue(cad).found()) && cad.isInteractivelyPromptable())
+                                .collect(Collectors.toList());
+
+                if (!interactivelyPromptableParametersWithDefaultValue.isEmpty()) {
```
nvoxland, line 154: Asking the user in this generic code about setting up a new liquibase.properties file and sample changelog seems too "init" specific. I think there are good use cases for prompting the user for arguments where we are never going to persisting those anywhere like for the password etc.

This logic of asking about saving values should belong in the init logic, not here.

liquibase-core/src/main/java/liquibase/configuration/core/InteractivePromptingValueProvider.java
```diff
@@ -0,0 +1,30 @@
+package liquibase.configuration.core;
+
+import liquibase.configuration.AbstractMapConfigurationValueProvider;
+
+import java.util.Map;
+import java.util.SortedMap;
+import java.util.TreeMap;
+
+/**
+ * This value provider handles values obtained during interactive CLI prompting.
```
nvoxland, line 10: Eventually this should apply to more than just the CLI. Should this class be made CLI-specific and live in liquibase.integration.commandline with an eventual superclass? Or left where it is, but described in more generic terms?

liquibase-core/src/main/java/liquibase/ui/interactive/FileTypeEnum.java
```diff
@@ -0,0 +1,5 @@
+package liquibase.ui.interactive;
+
+public enum FileTypeEnum {
```
nvoxland, line 3: This doesn't seem like it's a globally applicable enum. There are tons of file types we can deal with, and the list of what there CAN be is extensible. This should maybe be a nested enum somewhere instead of something top-level?

liquibase-core/src/main/java/liquibase/ui/interactive/InteractivePromptableCustomizationWrapper.java
```diff
@@ -18,9 +18,7 @@
     private final Function<List<DynamicRuleParameter>, Boolean> shouldPrompt;
 
     public InteractivePromptableCustomizationWrapper(IInteractivelyPromptableEnum parameter) {
```
nvoxland, line 20: Why the "II" in the classname? We don't do hungarian notation

liquibase-core/src/main/java/liquibase/ui/interactive/InteractivelyPromptableParameters.java
```diff
@@ -0,0 +1,83 @@
...
+
+import java.util.Arrays;
+
+import static liquibase.integration.commandline.LiquibaseCommandLineConfiguration.DEFAULTS_FILE;
+
+/**
+ *
+ */
+public enum InteractivelyPromptableParameters implements IInteractivelyPromptableEnum {
+    PROJECT_DIR(null, "Enter a relative path to desired project directory, or (n) to cancel:", ".", (String)null, new DirectoryGetter()),
```
nvoxland, line 16: This all seems to be "init" specific logic and prompts, not something that belongs in the global liquibase.ui.interactive API.

liquibase-core/src/main/java/liquibase/ui/interactive/getter/DirectoryGetter.java
```diff
@@ -0,0 +1,29 @@
+package liquibase.ui.interactive.getter;
+
+import liquibase.ui.interactive.AbstractCommandLineValueGetter;
+
+import java.io.File;
+
+public class DirectoryGetter extends AbstractCommandLineValueGetter<String> {
```
nvoxland, line 7: Since we are plugging the interactivity into the ConfigurationValueProvider system, why isn't this using the ConfigurationValueConverter structures we already have?

nvoxland (changes requested):
Some overall good directions in here, but I think we need to put more thought into the generic interactivity API to keep it more open and future-proof. It's a bit tied to our first specific use case and also maybe not well set up to work with the future multi-step command pipelines.

I like the idea of being able to attach interactivity options to CommandArgumentDefinitions, but remember that a future setup will be something like the "update" operation having 3 steps: one that collects up hub settings, one that collects up connection information, and one that takes a database connection and a hub connection and does an update against those (regardless of how those objects got set up). And that extensions may introduce additional steps and arguments into that pipeline.

An alternative approach could be to pull all the interactivity out of the CommandArgumentDefintion and the generic CommandScope.execute() logic in general in favor of speicifc CommandSteps for commands that want to provide interactivity. That would allow us to still have an isolated "when possible, ask for this specific stuff" flow without building it into the more pure-business-logic CommandStep implementation. But, prompting does seem to be enough of a general-purpose thing that it would be nice to pull that logic into the main framework as we're trying here.

liquibase-core/src/main/java/liquibase/command/CommandDefinition.java
StevenMassaro: DependencyUtil is a neat idea, I didn't know it existed. I will change this so we don't depend on insertion order.

liquibase-cli/src/main/java/liquibase/integration/commandline/LiquibaseCommandLine.java
StevenMassaro: This solves a lot of problems, so I'm glad you're on board.

liquibase-core/src/main/java/liquibase/command/CommandArgumentDefinition.java
StevenMassaro: Good question.

This is a carry over from the prompting logic that was born out of the need in `checks customize`. We needed a way to control prompts based on the "answers" that the user provided in previous prompts. Hence, previously prompted values.

Example that applies to `init project`: if the user answers "s" (indicating "skip") to the question of "what filename should I use for the changelog", we need to know to skip a later prompt that asks what type of changelog they want. This implementation worked well in the `checks customize` flow, but feels a bit hamfisted here. I was intending to revisit this in a future story where I actually implement this "skip later prompts based on prior prompts logic".

liquibase-core/src/main/java/liquibase/command/CommandArgumentDefinition.java
StevenMassaro: Absolutely not. I intend/intended (perhaps forgot) to refactor this out. Again, the issue here is that this came over from the prompting from `checks customize` and I was trying to not make breaking changes.

liquibase-core/src/main/java/liquibase/command/CommandScope.java
StevenMassaro: I intentionally put this logic here to support the future vision. In the future, running commands like `liquibase update` and omitting the `changeLogFile` parameter will automatically route the user into an interactive "init project" prompting session.

liquibase-core/src/main/java/liquibase/command/CommandArgumentDefinition.java
StevenMassaro: Do you have an alternative suggestion? I agree this is unwieldy. How about `CustomizedInteractivePrompt`?

liquibase-core/src/main/java/liquibase/configuration/core/InteractivePromptingValueProvider.java
StevenMassaro: I'm not seeing the same future plans that you are. What would this apply to in the future besides interactive prompting in the CLI?

liquibase-core/src/main/java/liquibase/ui/interactive/InteractivelyPromptableParameters.java
StevenMassaro: Not sure I agree here given the future vision for this system. When this interactive flow is enabled for `liquibase update` (and other core commands), these parameters will be called from multiple locations.

liquibase-core/src/main/java/liquibase/command/CommandDefinition.java
StevenMassaro: DependencyUtil turned out to be a bit overkill for this situation, when all I really need is a list. I went with the simpler solution for now.

StevenMassaro:
Outcomes of meeting:
- Move prompting logic out of global spot to init project specific spot
- Need to do more analysis and planning when we handle interactive prompting application-wide
- Deferring the design discussion for an interactive prompting API to a later date

PR #2277, opened 2021-12-16
Fix for working with Firebird 2 and newer versions. Fixes nullables and boolean type.

mngsgoncalves:
Fix for working with Firebird 2 and newer versions. Fixes nullables and boolean type.

Current implementation does not handle correctly working on Firebird 2.5 and Firebird 3 databases. It always assumes a Firebird 3 database, which incurs in invalid SQL statements.

The proposed fix eliminates the Firebird3Database class and instead includes a "version2" attribute in the FirebirdDatabase class. This attribute is populated when the connection is set.

I've fixed the tests and impacted classes (BooleanType and SetNullableGenerator) for this implementation.

### DIFF
--- a/liquibase-core/src/main/java/liquibase/datatype/core/BooleanType.java
+++ b/liquibase-core/src/main/java/liquibase/datatype/core/BooleanType.java
@@ -6,6 +6,7 @@ import liquibase.database.core.*;
 import liquibase.datatype.DataTypeInfo;
 import liquibase.datatype.DatabaseDataType;
 import liquibase.datatype.LiquibaseDataType;
+import liquibase.exception.DatabaseException;
 import liquibase.exception.UnexpectedLiquibaseException;
 import liquibase.statement.DatabaseFunction;
 import liquibase.util.StringUtil;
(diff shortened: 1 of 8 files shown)


kataggart:
Thanks @mngsgoncalves for taking the time to look into this and propose the fix. We are reviewing.

nvoxland:
I like the removal of the Firebird3 class since that is not really our standard way of working. However, if anyone has extensions that depend on Firebird3 that will break them, so we may need to re-introduce it as a deprectated class to be safe.

I'm currently working on a big improvement to the integration test system, and am going to leave this open for a little bit since it will be a good test of it as I add firebird support in. Should be able to get to it soon, though.

nvoxland (approved):
I switched the custom isVersion2() logic for the standard getDatabaseMajorVersion check, with an assumption it's 3+ since 2.x versions are no longer supported by firebird.

Otherwise looks good, and the test-harness tests should catch any issues with firebird 3 or 4

nvoxland:
The failing functional test is unrelated to the changes

liquibase-core/src/main/java/liquibase/datatype/core/BooleanType.java
```diff
@@ -19,11 +20,18 @@ public class BooleanType extends LiquibaseDataType {
     @Override
     public DatabaseDataType toDatabaseDataType(Database database) {
         String originalDefinition = StringUtil.trimToEmpty(getRawDefinition());
-        if ((database instanceof Firebird3Database)) {
+        if ((database instanceof FirebirdDatabase)) {
+            try {
```
XDelphiGrl, line 24: As a non-developer, I find the use of a try/catch instead of if/else interesting. If anyone has time, can you please explain the advantages of this approach? I'm always happy to learn something new. :)

XDelphiGrl (approved):
This change replaces the Firebird3Database class with more streamlined logic to determine Firebird version. Depending on the Firebird version, different SQL is generated for both setting the boolean datatype and setting the nullable state of a column. 

* Fix specific to the Firebird database.
* Existing unit tests updated to reflect removal of the Firebird3Database class.

APPROVED

[Functional Tests](https://github.com/liquibase/liquibase-pro-tests/actions/runs/2399029188)
[Passing Test Harness Execution](https://github.com/liquibase/liquibase-test-harness/actions/runs/2399030506)
NOTE: Test failures in the functional test suite are due to mismatch in expected console output and actual console output for the checks feature. These are unrelated to the changes in this PR and can be disregarded.

liquibase-core/src/main/java/liquibase/datatype/core/BooleanType.java
suryaaki2: @nvoxland Do we want to log the exception??

suryaaki2 (changes requested):

liquibase-core/src/main/java/liquibase/datatype/core/BooleanType.java
nvoxland: I added a log of the exception

suryaaki2 (approved):

PR #2308, opened 2021-12-31
Make ClassLoaderResourceAccessor implement Closable

Delir4um:
## Description
The ClassLoaderResourceAccessor opens resources but has no method for closing them. Add a close() method to clean it up as part of the AutoClosabable interface

- - -
#### Links
* Fixed Issue: Stand alone PR
* Test Results: https://github.com/liquibase/liquibase/pull/2308/checks

#### Testing
* Steps to Reproduce: None
* Guidance:
  * Impact: This adds a close method which our code does not use. 3rd party integrations would be able to take advantage of it, however.

#### Dev Verification
Reviewed code

┆Issue is synchronized with this [Jira Bug](https://datical.atlassian.net/browse/LB-2205) by [Unito](https://www.unito.io)

### DIFF
--- a/liquibase-core/src/main/java/liquibase/resource/ClassLoaderResourceAccessor.java
+++ b/liquibase-core/src/main/java/liquibase/resource/ClassLoaderResourceAccessor.java
@@ -270,7 +270,7 @@ public class ClassLoaderResourceAccessor extends AbstractResourceAccessor implem
                     try (JarFile jar = new JarFile(URLDecoder.decode(jarPath, StandardCharsets.UTF_8.name()))) {
                         String comparePath = path;
                         if (comparePath.startsWith("/")) {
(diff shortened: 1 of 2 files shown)


nvoxland (approved):
I added a log message when close failed, but otherwise good. Thanks!

nvoxland:
Just making it auto-closable doesn't help the existing code which should be calling the method. 

But, in looking through the code we have there is often too much of a disconnect between when the ClassLoaderResourceAccessor is opened/created and when it would need to be closed. So I'm good with just adding the method for now. I think that over the next few months as we do more refactoring of the code we can take better advantage of this new close method. And possibly consider adding it further up the ResourceAccessor chain

suryaaki2 (approved):

kataggart:
@nvoxland please note this was approved as ready to merge

PR #2312, opened 2022-01-03
New Test System management

nvoxland:
## Description

The infrastructure for managing test systems for integration tests is old and needs to be improved.


liquibase-extension-testing/pom.xml
```diff
@@ -10,13 +10,61 @@
...
+        <dependency>
+            <groupId>org.testcontainers</groupId>
+            <artifactId>db2</artifactId>
+        </dependency>
+
+        <dependency>
+            <groupId>org.slf4j</groupId>
+            <artifactId>slf4j-jdk14</artifactId>
+            <version>1.7.33</version>
+        </dependency>
```
StevenMassaro, line 66: It might be a good idea to have these with `<scope>test</scope>`.

liquibase-core/src/main/java/liquibase/util/CollectionUtil.java
```diff
@@ -88,4 +88,34 @@
         }
     }
 
+    /**
+     * Converts a set of nested maps (like from yaml/json) into a flat map with dot-seprated properties
```
StevenMassaro, line 92:
```suggestion
     * Converts a set of nested maps (like from yaml/json) into a flat map with dot-separated properties
```

liquibase-core/src/test/java/liquibase/integration/commandline/MainTest.java
```diff
@@ -117,8 +117,8 @@ public void testLocalProperties() throws Exception {
         Main cli = new Main();
         cli.parseOptions(args);
 
-        assertTrue("Read context from liquibase.local.properties", ((cli.contexts != null) && cli.contexts.contains
-            ("local-context-for-liquibase-unit-tests")));
+//        assertTrue("Read context from liquibase.local.properties", ((cli.contexts != null) && cli.contexts.contains
+//            ("local-context-for-liquibase-unit-tests")));
```
StevenMassaro, line 121: Do you want this commented?

liquibase-extension-testing/src/main/java/liquibase/extension/testing/setup/TestSetupEnvironment.java
```diff
@@ -0,0 +1,35 @@
...
+    public TestSetupEnvironment(DatabaseTestSystem testSystem, DatabaseTestSystem altSystem) throws SQLException {
+        this.connection = testSystem.getConnection();
+        this.url = testSystem.getConnectionUrl();
+        this.username = testSystem.getUsername();
+        this.password = testSystem.getPassword();
+
+        this.altUrl = this.url.replace(testSystem.getCatalog(), testSystem.getAltCatalog());
+        this.altUsername = this.username;
+        this.altPassword = this.password;
+        this.altConnection = DriverManager.getConnection(this.altUrl, this.altUsername, this.altPassword);
```
StevenMassaro, line 32: Why are you getting the connection differently for the altSystem? If this is intentional, a comment here explaining why would be useful to future generations of Liquibase maintainers.

liquibase-extension-testing/src/main/java/liquibase/extension/testing/testsystem/command/TestSystemUpCommand.java
```diff
@@ -0,0 +1,70 @@
...
+
+        NAME = builder.argument("name", String.class).required()
+                .description("The name of the system to").build();
+        VERSION = builder.argument("version", String.class)
+                .description("Override version to use").build();
+        PROFILES = builder.argument("profiles", String.class)
+                .description("Set profile(s)").build();
+        ACCEPT_LICENSES = builder.argument("acceptLicense", Boolean.class)
+                .description("Accept licenses for any systems used/accessed")
+                .defaultValue(false)
```
StevenMassaro, line 34: Why is this a default of false? This seems like a nuisance.

liquibase-extension-testing/pom.xml
nvoxland: No, they'll have to be non-test.

The liquibase-extension-testing module is a separate jar which other modules and eventually 3rd party extensions use for testing. The "production" code in there is to be a test framework for other stuff, so everything needed to run that test framework (like the testcontainers and slf4j-jdk14) are part of its production code.

The other modules that use this will declare liquibase-extension-testing as scope=test, which will make these dependencies test dependencies at that point.

liquibase-core/src/test/java/liquibase/integration/commandline/MainTest.java
nvoxland: Yeah, there was a liquibase.local.properties file that had gotten committed to github for this test to run, but having it there could get in the way of trying to use it for testing. Taking that file out made this test fail. It's in legacy code so I'm OK with that not being tested right now.

liquibase-extension-testing/src/main/java/liquibase/extension/testing/setup/TestSetupEnvironment.java
nvoxland: This is for the existing CommandTest stuff. The way the tests are set up, they need to be able to have an "alt system" for things like diff commands which need to work against two separate connections.

What we have in this TestSetupEnvironment was made for the CommandTests, but with an eye toward being part of a generic way of managing the system. I'm planning on trying to refactor this code somehow, but I'm not quite sure how yet. I want to see how the tests we write over the next few weeks can leverage it and what they need.

So I figure this code here is pretty temporary, and just there as a shim for the existing CommandTests until we can do it better soon.

liquibase-extension-testing/src/main/java/liquibase/extension/testing/testsystem/command/TestSystemUpCommand.java
nvoxland: Yes, it is a nuisance, but I'm not a lawyer and in talking to @kevin-at-datical he figured that 3rd party license agreements should be made by the user, not automatically by us. So this is what I did to try to keep it as non-annoying as possible while still making the license decision be in the user's hands.

Besides this argument, there is also a liquibase.sdk.testSystem.acceptLicenses config which can be set in your liquibase.properties or liquibase.sdk.local.yaml file too.

liquibase-extension-testing/src/main/java/liquibase/extension/testing/testsystem/TestSystemFactory.java
```diff
@@ -0,0 +1,85 @@
...
+import java.util.function.Function;
+import java.util.stream.Collector;
+import java.util.stream.Collectors;
+
+/**
+ * Factory for getting {@link TestSystem} implementations.
+ */
+public class TestSystemFactory extends AbstractPluginFactory<TestSystem> {
+
+    private Map<TestSystem.Definition, TestSystem> systems = new HashMap<>();
```
StevenMassaro, line 22: Should this be static? Or are we guaranteed that there will only ever be one instance of this factory?

liquibase-extension-testing/src/main/java/liquibase/extension/testing/testsystem/spock/LiquibaseIntegrationMethodInterceptor.java
```diff
@@ -0,0 +1,70 @@
...
+            env.start();
+
+        }
+    }
+
+    private void stopContainers(List<FieldInfo> containers, IMethodInvocation invocation) throws Exception {
+        for (FieldInfo field : containers) {
+            TestSystem env = readContainerFromField(field, invocation);
+
+            env.stop();
```
StevenMassaro, line 62: Shouldn't you wrap line 62 in a try/catch? If there are multiple containers to stop, and the first one fails, the other ones will not stop.

The same problem will occur with the startContainers method, but it seems like a desirable side effect there.

liquibase-extension-testing/src/main/java/liquibase/extension/testing/util/DownloadUtil.java
```diff
@@ -0,0 +1,80 @@
...
+        final String url = "https://repo1.maven.org/maven2/" + groupId.replace(".", "/") + "/" + artifactId + "/" + version + "/" + artifactId + "-" + version + ".jar";
+
+        final File pathAsFile = path.toFile();
+        if (pathAsFile.exists()) {
+            log.fine("Artifact " + groupId + ":" + artifactId + ":" + version + " is available at " + path);
+        } else {
+            log.info("Downloading " + url + " to " + path);
+
+            try {
+                final HttpURLConnection connection = (HttpURLConnection) new URL(url).openConnection();
```
StevenMassaro, line 52: Might want to use try with resources here

liquibase-extension-testing/src/main/resources/META-INF/services/liquibase.command.CommandStep
```diff
@@ -0,0 +1,2 @@
+liquibase.extension.testing.testsystem.command.TestSystemDownCommand
```
StevenMassaro, line 1: If you are interested, we used a maven plugin in pro that generates these files dynamically at build time. I can point you to more info, let me know.

liquibase-integration-tests/src/test/java/liquibase/dbtest/h2/H2IntegrationTest.java
```diff
@@ -32,10 +33,9 @@ public H2IntegrationTest() throws Exception {
         this.dbmsExcludeChangelog = "changelogs/h2/complete/dbms.exclude.changelog.xml";
     }
 
-    @Override
-    protected boolean isDatabaseProvidedByTravisCI() {
-        // H2 is an in-process database
-        return true;
+    @Test
+    public void showVersion() throws DatabaseException {
+        System.out.println(getDatabase().getDatabaseProductVersion());
```
StevenMassaro, line 38: Does this test do anything?

liquibase-integration-tests/src/test/java/liquibase/statementexecute/AbstractExecuteTest.java
```diff
@@ -119,22 +109,22 @@ private void test(String[] expectedSql, Class<? extends Database>[] includeDatab
...
-                    }
-                    statement.execute(sqlToRun);
-                } catch (Exception e) {
-                    System.out.println("Failed to execute against " + availableDatabase.getShortName() + ": " + sqlToRun);
-                    throw e;
-
-                }
-            }
-        }
+//        for (Database availableDatabase : DatabaseTestContext.getInstance().getAvailableDatabases()) {
```
StevenMassaro, line 112: Confirming that you want all of the code in this file to be commented out, and not removed.

liquibase-extension-testing/src/main/java/liquibase/extension/testing/testsystem/command/TestSystemUpCommand.java
StevenMassaro: That makes sense. When documentation comes around, we should make sure that the config option is in the docs.

liquibase-extension-testing/src/main/java/liquibase/extension/testing/testsystem/TestSystemFactory.java
nvoxland: There should just be one instance of the factory. And if there are multiple we might want them handling their own managed systems.

liquibase-extension-testing/src/main/java/liquibase/extension/testing/testsystem/spock/LiquibaseIntegrationMethodInterceptor.java
nvoxland: Yeah, I added try catch to just 62 and not the start(). 

This whole spock interceptor is something I got going enough to see how it works but haven't gotten a chance to really use and vet yet. There are likely other improvements we'll need to use once we start wrting tests with it. The existing integration tests were junit

liquibase-extension-testing/src/main/java/liquibase/extension/testing/util/DownloadUtil.java
nvoxland: HttpURLConnection doesnt implement AutoCloseable so I can't. There is no close method on it even.

liquibase-extension-testing/src/main/resources/META-INF/services/liquibase.command.CommandStep
nvoxland: Yeah, it's worth thinking about at some point but not as part of this PR

liquibase-integration-tests/src/test/java/liquibase/dbtest/h2/H2IntegrationTest.java
nvoxland: No, it was something I added to see how I could change versions through settings. I'll take it out.

liquibase-integration-tests/src/test/java/liquibase/statementexecute/AbstractExecuteTest.java
nvoxland: Thanks, this was some code I needed to leave and had just meant to temporarily comment until I could refactor it correctly.

I'll get it working again

suryaaki2 (approved):

liquibase-extension-testing/src/main/java/liquibase/extension/testing/testsystem/TestSystem.java
```diff
@@ -0,0 +1,346 @@
...
+import org.junit.runners.model.MultipleFailureException;
+import org.junit.runners.model.Statement;
+
+import java.util.*;
+import java.util.regex.Matcher;
+import java.util.regex.Pattern;
+
+/**
+ * TestSystem implementations define and manage a connection to an external system to test.
+ * Ideally the implementation can start and top the test systems, but that is not necessary.
```
suryaaki2, line 23:
```suggestion
 * Ideally the implementation can start and stop the test systems, but that is not necessary.
```

suryaaki2 (changes requested):

PR #2316, opened 2022-01-04
Jdk 17 Support

nvoxland:
## Description

- Replace JDK 16 with 17 for builds


nvoxland:
Closing in favor of #2317 which builds without pull_request_target

PR #2318, opened 2022-01-04
Run automated tests in JDK17 rather than 16

nvoxland:
## Description

Replace java 16 in "build & test" with 17 since 16 is no longer a supported JVM

We still only ship the 1.8-compiled version, so the new 17 build is still thrown away like the 16-built one was.

## Changes
* (Test code issue) For the build, we were on an old enough version of mockito for testing, that it had to be upgraded to support the java 17 bytecode
* (Test code issue) The tests for the CDI needed new JVM flags to avoid complaints of being unable to access classes

┆Issue is synchronized with this [Jira Bug](https://datical.atlassian.net/browse/LB-2202) by [Unito](https://www.unito.io)


nvoxland:
I split out the installer-shipping of java 17 into #2602 

But, I did set the java version we ship in this PR to the latest java 11 rather than the version of 11 we had been shipping before so it is still a slight upgrade.

XDelphiGrl:
All automated tests passing.
* Except for one functional test that is failing due to a change in the quality checks feature output (not related to this PR).

No new tests required.
Moving to ready to merge.

[Functional Tests](https://github.com/liquibase/liquibase-pro-tests/actions/runs/1934766329)

PR #2330, opened 2022-01-05
Correctly catch potentially null value

inanemed:
check if referenceDatabase not null instead of referenceSnapshot.

---

#### Links

- Stand alone PR
- Test Results: https://github.com/liquibase/liquibase/pull/2330/checks

#### Testing

- Steps to Reproduce: Not sure. The fix obviously avoids a nulls a null pointer which could happen, but I'm not sure how to set up something naturally happening in liquibase to trigger it. Not sure what could lead to a null value, but if anything does it will handle it better now. Probably needs no additional manual testing?
- Guidance:
  - Impact: Stops potential null pointer
  
#### Dev Verification

Reviewed code.

### DIFF
--- a/liquibase-core/src/main/java/liquibase/diff/DiffGeneratorFactory.java
+++ b/liquibase-core/src/main/java/liquibase/diff/DiffGeneratorFactory.java
@@ -84,7 +84,7 @@ public class DiffGeneratorFactory {
 
     public DiffResult compare(DatabaseSnapshot referenceSnapshot, DatabaseSnapshot comparisonSnapshot, CompareControl compareControl) throws DatabaseException {
         Database referenceDatabase = referenceSnapshot.getDatabase();
-        if (comparisonSnapshot !=null && referenceSnapshot!=null) {
+        if (comparisonSnapshot !=null && referenceDatabase!=null) {
             if (referenceDatabase.getDefaultCatalogName() == null || referenceDatabase.getDefaultCatalogName().isEmpty()) {
                 referenceDatabase.setDefaultCatalogName(comparisonSnapshot.getDatabase().getDefaultCatalogName());
             }


nvoxland (approved):
Thanks. referenceDatabase gets called in the if statement, so checkign that it is null and not referenceSnapshot is what's neede.

liquibase-core/src/main/java/liquibase/diff/DiffGeneratorFactory.java
```diff
@@ -84,7 +84,7 @@ public DiffResult compare(Database referenceDatabase, Database comparisonDatabas
 
     public DiffResult compare(DatabaseSnapshot referenceSnapshot, DatabaseSnapshot comparisonSnapshot, CompareControl compareControl) throws DatabaseException {
         Database referenceDatabase = referenceSnapshot.getDatabase();
-        if (comparisonSnapshot !=null && referenceSnapshot!=null) {
+        if (comparisonSnapshot !=null && referenceDatabase!=null) {
```
suryaaki2, line 87: @nvoxland Do we need to do a null check for referenceSnapshot too?

suryaaki2 (changes requested):

liquibase-core/src/main/java/liquibase/diff/DiffGeneratorFactory.java
nvoxland: No, referenceSnapshot is never null so it doesn't have to be checked. We're already referencing it on line 86 so if it was a problem it would have been seen breaking there sometime which it isn't.

suryaaki2 (approved):

XDelphiGrl:
I reviewed the changes in the commit and agree, @nvoxland, that there is nothing additional to test from a QA perspective. Should there be a unit test for this change?

All automated tests passed.
No new tests required.
No manual testing required. 
[Passing Funcitonal Tests](https://github.com/liquibase/liquibase-pro-tests/actions/runs/1660190320)

nvoxland:
@XDelphiGrl I think probably not a useful enough unit test. The code is trying to do a lot of things there, and a lot of it is around how it's interacting with the outside systems and the coordination between components.

I'd like to eventually be able to refactor that code to have a better separation between what we'd test as a unit and what takes integration/coordination tests. But since we don't want to get into that right now for this change, I think the existing tests cover it well enough even if they are a bit slower to test it.

XDelphiGrl:
@nvoxland, thank you for taking a look at adding a test; makes sense to skip since it would add little value. Moving to ready to merge.

Issue #2352, opened 2022-01-10
includeAll cannot process directories in embedded jar files


hayeskl:
## Environment
Grails 5.0.3

**Liquibase Version**:
4.6.2

**Liquibase Integration & Version**: <Pick one: CLI, maven, gradle, spring boot, servlet, etc.>
Core

**Liquibase Extension(s) & Version**: 

**Database Vendor & Version**:

**Operating System Type & Version**:
Mac Monterey 12.1

## Description

A clear and concise description of the issue being addressed.
When running the application as a jar file the migrations are not found.  The following error is returned:

liquibase.exception.SetupException: Could not find directory or directory was empty for includeAll 'current/'
        at liquibase.changelog.DatabaseChangeLog.includeAll(DatabaseChangeLog.java:553)
        at liquibase.changelog.DatabaseChangeLog.handleChildNode(DatabaseChangeLog.java:409)
        at liquibase.changelog.DatabaseChangeLog.load(DatabaseChangeLog.java:318)
        at liquibase.parser.core.xml.AbstractChangeLogParser.parse(AbstractChangeLogParser.java:23)
        ... 96 more
Caused by: liquibase.exception.SetupException: Could not find directory or directory was empty for includeAll 'current/'
        at liquibase.changelog.DatabaseChangeLog.includeAll(DatabaseChangeLog.java:544)
        ... 99 more

## Steps To Reproduce

Package migrations into jar file relative to changeLogFile
Build a spring bootable jar for application and include the jar with migrations
run the spring bootable jar

## Actual Behavior
The above exception is thrown because ClassLoaderResourceAccessor cannot parse the resource path:
jar:file:/Users/khayes/griffin.git/trifleet/build/libs/trifleet-12.0.0.jar!/BOOT-INF/lib/common-12.0.0-plain.jar!/current/

## Expected/Desired Behavior
Migrations should be loaded from the embedded jar file

hayeskl:
This issue is fixed with https://github.com/liquibase/liquibase/pull/2353

hayeskl:
I've confirmed that this issue still exists with liquibase 4.24.0.

PR #2353, opened 2022-01-10
fixed includeAll when loaded from embedded jar files (fix for #2352)

hayeskl:
## Environment
Grails 5.0.3

**Liquibase Version**:
4.6.2

Core

Mac Monterey 12.1

## Description

A clear and concise description of the issue being addressed.
When running the application as a jar file the migrations are not found.  The following error is returned:

liquibase.exception.SetupException: Could not find directory or directory was empty for includeAll 'current/'
        at liquibase.changelog.DatabaseChangeLog.includeAll(DatabaseChangeLog.java:553)
        at liquibase.changelog.DatabaseChangeLog.handleChildNode(DatabaseChangeLog.java:409)
        at liquibase.changelog.DatabaseChangeLog.load(DatabaseChangeLog.java:318)
        at liquibase.parser.core.xml.AbstractChangeLogParser.parse(AbstractChangeLogParser.java:23)
        ... 96 more
Caused by: liquibase.exception.SetupException: Could not find directory or directory was empty for includeAll 'current/'
        at liquibase.changelog.DatabaseChangeLog.includeAll(DatabaseChangeLog.java:544)
        ... 99 more

## Steps To Reproduce

Package migrations into jar file relative to changeLogFile
Build a spring bootable jar for application and include the jar with migrations
run the spring bootable jar

## Actual Behavior
The above exception is thrown because ClassLoaderResourceAccessor cannot parse the resource path:
jar:file:/Users/khayes/griffin.git/trifleet/build/libs/trifleet-12.0.0.jar!/BOOT-INF/lib/common-12.0.0-plain.jar!/current/

## Expected/Desired Behavior
Migrations should be loaded from the embedded jar file

### DIFF
--- a/liquibase-core/src/main/java/liquibase/resource/ClassLoaderResourceAccessor.java
+++ b/liquibase-core/src/main/java/liquibase/resource/ClassLoaderResourceAccessor.java
@@ -5,6 +5,7 @@ import liquibase.changelog.DatabaseChangeLog;
 import liquibase.util.StreamUtil;
 
 import java.io.File;
+import java.io.FileInputStream;
 import java.io.IOException;
 import java.io.InputStream;
 import java.net.URISyntaxException;
@@ -17,6 +18,7 @@ import java.nio.file.attribute.BasicFileAttributes;
 import java.util.*;
 import java.util.jar.JarEntry;
 import java.util.jar.JarFile;
+import java.util.jar.JarInputStream;
 
 /**
  * An implementation of {@link FileSystemResourceAccessor} that builds up the file roots based on the passed {@link ClassLoader}.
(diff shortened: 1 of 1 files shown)


joannabartosz001:
Hi, any updates on this? We have met the same problem in our app.

nvoxland:
This has possibly been fixed by other changes, but I'm not able to reproduce the problem even on the listed 4.6.2 version.

Can you create a sample project I can run that replicates the issue, @hayeskl ?

hayeskl:
> This has possibly been fixed by other changes, but I'm not able to reproduce the problem even on the listed 4.6.2 version.
> 
> Can you create a sample project I can run that replicates the issue, @hayeskl or @joannabartosz001 ?

This is definitely still and issue.  The problem is that the ClassLoaderResourceAccessor is only loading the first jar file it sees on the url path instead of processing all of the jars listed on the path.  For example, my path jarPath is: jar:file:/Users/khayes/griffin.git/trifleet/build/libs/trifleet-12.0.0.jar!/BOOT-INF/lib/common-12.0.0-plain.jar!/current/

Since the ClassLoaderResourceAccessor is stopping after the first ! it does not see the directory in the second jar file.

nvoxland:
I see what you mean, @hayeskl : we're not correctly finding jars inside other jars in the ClassLoaderResourceAccessor.

We're currently doing a fairly large refactoring of the resource handling stuff in #3064 but that is also not going to handling that well.

I'll see how to incorporate this into the setup.

**_Question:_** You list the issue as "grails" but are you using grails to start a Spring application? Or are you running liquibase more directly from grails?

hayeskl:
Grails is more of a wrapper around a Spring Application.

nvoxland:
OK, thanks @hayeskl.

I'm going to close this as "duplicate" because I think the changes we are doing in #3064 should resolve this issue. That PR ended up resulting in a larger revisit of how the ResourceAccessors work in general including better handling of the SpringResourceAccessor to rely on spring for finding files. 

Since spring is correctly handling the nested jars and exposes a way for us to search in within it, we don't have to build our own logic for it.

If you are able to test the code in that branch to make sure it solves your problem prior to it being merged that would be great. Follow up on 3064 on what you find. Otherwise, if it's still an issue in whatever release has 3064 in it (probably 4.17) you can open a new issue.

Issue #2354, opened 2022-01-11
Print logs on ChangeSet exception


maxencelaurent:
## Environment

**Liquibase Version**: 4.7.0 with CDI Plugin
**Liquibase Integration & Version**: 
**Liquibase Extension(s) & Version**: any
**Database Vendor & Version**: any

**Operating System Type & Version**: any

## Description
Despite LiquiBase logs plenty of information, there is not a single statement about exceptions which may occur when applying changesets (FINE level set for all LiquiBase loggers). 

The current behaviour is to simply throw exceptions, without any logs.
 
When Liquibase is run in a CDI environment (weld, in my case), there is no way to understand what's wrong as the only relevant log statement is:
```
org.glassfish.deployment.common.DeploymentException: CDI deployment failure:WELD-000049: 
    Unable to invoke public void liquibase.integration.cdi.CDILiquibase.onStartup() on liquibase.integration.cdi.CDILiquibase@403e8b5
WELD-000049: Unable to invoke public void liquibase.integration.cdi.CDILiquibase.onStartup() on liquibase.integration.cdi.CDILiquibase@403e8b5
```

## Steps To Reproduce 
- configure liquibase to run in CDI environnement
- write some changeset that will fail, eg:
```xml
<update tableName="doesNotExist">
  <column name="doesNotExistEither" value="" />
</update>
```
- Run it.

## Actual Behavior
Liquibase does not log any relevant information about what failed.
Only way to understand underlying errors is to use a debugger... 

## Expected/Desired Behavior
Big fat error message which explain what's wrong

kataggart:
Thanks for taking the time to submit this detail @maxencelaurent . We are reviewing and will get back to you.

nvoxland:
@maxencelaurent: 
I opened an alternate implementation of #2397. That seems to me to be a more general case solution, but I don't personally use CDI enough to know for sure whether it's still going to solve your problem.

What do you think on #2397 vs. #2355?

XDelphiGrl:
The Liquibase CDI integration needs some love from QA. Instead of blocking this PR from merging until a full/official certification of Liquibase CDI completes,  @nvoxland added a new integration test, `onStartupExceptionsAreCorrectlyHandled()` to `liquibase-cdi/src/test/java/liquibase/integration/cdi/CDILiquibaseTest.java`.  This test ran and [passed](https://github.com/liquibase/liquibase/pull/2397/checks). 

[...]

PR #2355, opened 2022-01-11
Print logs on ChangeSet exception

maxencelaurent:
see #2354

### DIFF
--- a/liquibase-core/src/main/java/liquibase/changelog/ChangeLogIterator.java
+++ b/liquibase-core/src/main/java/liquibase/changelog/ChangeLogIterator.java
@@ -121,6 +121,7 @@ public class ChangeLogIterator {
                 }
             });
         } catch (Exception e) {
+            log.warning(e.toString());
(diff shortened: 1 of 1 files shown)


maxencelaurent:
Here is an example of the log message:

```
[2022-01-11T13:41:40.757+0100] [] [WARNING] [] [liquibase.changelog] [tid: _ThreadID=1 _ThreadName=main] [timeMillis: 1641904900757] [levelValue: 900] [[
  liquibase.exception.MigrationFailedException: Migration failed for change set META-INF/dbchangelogs/1641469325.xml::1641469325-1::maxence:
     Reason: liquibase.exception.DatabaseException: ERROR: column "georges" does not exist
  Position: 53 [Failed SQL: (0) UPDATE public.sometablename SET georges = '1234']]]
```

maxencelaurent:
Alternative #2397 is preferred

nvoxland:
Thanks for checking. We'll work on getting that in

PR #2356, opened 2022-01-11
point users to init project command if they forget a supported argument in any command (DAT-8641)

StevenMassaro:
This is a bit of a heavy-handed approach to this, and it does not have to be.

There are two commits:
First, the actual change. This is to adjust the message thrown by the exception during the validation stage of each argument.
Second, refactoring all existing commands which use any of the supported arguments to all use common code for the initialization of the argument. I did this primarily because I was concerned that the equality check would fail if one developer instantiated a command argument in a slightly different way (`changelogfile` vs `changelogFile`, for example)


liquibase-core/src/main/java/liquibase/command/CommandArgumentDefinition.java
```diff
@@ -112,8 +114,12 @@ public String getDefaultValueDescription() {
      */
     public void validate(CommandScope commandScope) throws CommandValidationException {
         final DataType currentValue = commandScope.getArgumentValue(this);
+        // This is a list of the arguments which the init project command supports. The thinking here is that if the user
+        // forgets to supply one of these arguments, we're going to remind them about the init project command, which
+        // can help them figure out what they should be providing here.
+        final List<String> initProjectArguments = Arrays.asList(CommandArgumentNames.CHANGELOG_FILE, CommandArgumentNames.URL, CommandArgumentNames.USERNAME, CommandArgumentNames.PASSWORD);
```
nvoxland, line 120: This seems like the wrong spot in the code to be putting this. The general CommandArgumentDefinition needs to be generic and not know anything about implementations. Adding logic for init project here is adding dependencies in the wrong direction.

If the init project doesn't have those as required, it should say that itself. Or if we are finding a use case that isn't currently handled by our CommandArgument metadata we should expand the metadata that can be attached to the arguments globally vs. hard-coding init into here.

liquibase-core/src/main/java/liquibase/command/CommandArgumentNames.java
```diff
@@ -0,0 +1,11 @@
+package liquibase.command;
+
+/**
+ * A common place to store commonly used command argument names.
+ */
+public class CommandArgumentNames {
```
nvoxland, line 6: This is maybe better as an enum?

Also maybe better named "CommonArgumentNames" vs. CommandArgumentNames"?

liquibase-core/src/main/java/liquibase/exception/CommandValidationException.java
```diff
@@ -6,7 +6,11 @@
 public class CommandValidationException extends CommandExecutionException {
 
     public CommandValidationException(String argument, String message) {
-        super("Invalid argument '" + argument + "': " + message);
+        this(argument, message, false);
+    }
+
+    public CommandValidationException(String argument, String message, boolean hasInitProjectParameters) {
```
nvoxland, line 12: hasInitProjectParameters as an argument here seems out of place here. As a generic extension, it shouldn't be containing init related messaging?

nvoxland (changes requested):

liquibase-core/src/main/java/liquibase/command/CommandArgumentDefinition.java
StevenMassaro: @nvoxland The acceptance criteria of the story which required this work says that we should be changing this message: "Error parsing command line: Invalid argument '--changelog-file': missing required argument", **for all commands** to "Error parsing command line: Invalid argument '--changelog-file': missing required argument. If you need to configure new liquibase project files and arguments, run the 'liquibase init project' command."

With that information, putting this logic elsewhere would require it be duplicated in many different places, wouldn't it? In other words, we want it here, because we want it to apply to every command which uses an argument which is also supported by `init project`.

liquibase-core/src/main/java/liquibase/command/CommandArgumentNames.java
StevenMassaro: Yes and yes. Resolved.

liquibase-core/src/main/java/liquibase/exception/CommandValidationException.java
StevenMassaro: Good point. I moved this logic up a layer.

liquibase-core/src/main/java/liquibase/command/CommandArgumentDefinition.java
nvoxland: I figured it was something around that, but I wasn't sure how many of the arguments it should show that message on.

Having gotten rid of the extra parameter on the exception has helped, since this code is at least encapsulated for now. Ideally it maybe should be in the CLI code since it just talks about CLI functionality (even though this code is integration-agnostic). Or could be a separate CommandStep along side the init code that catches the exception and outputs the message. 

But not sure if either are worth the effort or not. The encapsulation makes it fine if we want.

nvoxland (approved):

suryaaki2 (approved):

liquibase-core/src/main/java/liquibase/command/CommandArgumentDefinition.java
StevenMassaro: I moved this logic up a level, to the CLI code. Seems worth it to me.

liquibase-core/src/main/java/liquibase/command/core/CalculateChecksumCommandStep.java
```diff
@@ -20,9 +20,9 @@
 
     static {
         CommandBuilder builder = new CommandBuilder(COMMAND_NAME);
-        CHANGELOG_FILE_ARG = builder.argument("changelogFile", String.class).required()
+        CHANGELOG_FILE_ARG = builder.argument(CommonArgumentNames.CHANGELOG_FILE, String.class).required()
                 .description("The root changelog file").build();
-        URL_ARG = builder.argument("url", String.class).required()
+        URL_ARG = builder.argument(CommonArgumentNames.URL, String.class).required()
             .description("The JDBC database connection URL").build();
         DEFAULT_SCHEMA_NAME_ARG = builder.argument("defaultSchemaName", String.class)
```
wwillard7800, line 27: While they aren't as commonly used, defaultSchemaName and driverPropertiesFile are part of the set of common arguments across commands.

wwillard7800 (approved):

liquibase-core/src/main/java/liquibase/command/core/CalculateChecksumCommandStep.java
StevenMassaro: I was contemplating changing around the other arguments as well. I'll do that in another PR to keep things nice and separate.

nvoxland (approved):

PR #2365, opened 2022-01-13
Add Sonar scanning to builds

nvoxland:
## Description

Add sonar scanning to standard buildserver build logic. 

The results show as a PR status check, but also the build fails if the quality gate fails. 

It's currently adding a comment in PRs when it fails, but that is configurable within the sonar UI.

Builds running this check: https://github.com/liquibase/liquibase/actions?query=branch%3Agithub-action-sonar


.github/workflows/build.yml
```diff
@@ -150,6 +150,32 @@ jobs:
...
+        with:
+          fetch-depth: 0  # Shallow clones should be disabled for a better relevancy of analysis
+      - name: Set up JDK
+        uses: actions/setup-java@v2
+        with:
+          java-version: 11
+          distribution: 'adopt'
+          cache: 'maven'
+      - name: Cache SonarCloud packages
+        uses: actions/cache@v1
```
jnewton03, line 168: @nvoxland Any reason we can't use v2?
https://github.com/actions/cache

jnewton03 (commented):
@nvoxland can we also add the markdown badges?
https://sonarcloud.io/project/information?id=liquibase

.github/workflows/build.yml
nvoxland: Pushed v2 to see if it works.

nvoxland:
Added the badge

jnewton03 (approved):

PR #2453, opened 2022-02-01
DAT-8992: add start-H2 message to error message in JDBC connection creation

StevenMassaro:


liquibase-maven-plugin/src/main/java/org/liquibase/maven/plugins/MavenUtils.java
```diff
@@ -171,42 +171,6 @@ private static void addFile(Set<URI> urls, File f, Log log, boolean verbose)
         }
     }
 
-    public static Connection getDatabaseConnection(ClassLoader classLoader,
```
StevenMassaro, line 174: Idea reported that this code was not used anywhere, so I removed it.

suryaaki2 (approved):

PR #2504, opened 2022-02-07
DAT-9268 execute setup and build only if human has applied label 

ap-liquibase:
DAT-9268 execute setup and build only if human has applied label  SafeToBuild

## Environment

**Liquibase Version**:

## Pull Request Type

## Description

## Steps To Reproduce

- Please be precise and ensure private information is redacted
- Include things like
  - Files used - sql scripts, changelog file(s), property file(s), config files, POM Files
  - Exact commands used - CLI, maven, gradle, spring boot, servlet, etc.

## Actual Behavior

## Expected/Desired Behavior

- [ ] Added [Unit Test(s)](https://liquibase.jira.com/wiki/spaces/LB/pages/1274937609/How+to+Write+Liquibase+Core+Unit+Tests)
- [ ] Documentation Updated

## Need Help?

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -3,9 +3,7 @@ name: Build and Test
 on:
   pull_request_target:
     types:
-      - opened
-      - reopened
-      - synchronize
+      - labeled
     branches-ignore:
       - github-action-* #to allow testing of build logic in branches, don't do PR builds for github-action-* branches and instead use branch builds
 
@@ -20,6 +18,7 @@ on:
 jobs:
   setup:
     name: Setup
+    if: ${{ github.event.label.name == 'SafeToBuild' }}
     runs-on: ubuntu-latest
     outputs:
       proBranchName: ${{ steps.find-branches.outputs.proBranchName }}
@@ -109,6 +108,7 @@ jobs:
   build:
     name: Build & Test (Java ${{ matrix.java }})
     needs: setup
+    if: ${{ github.event.label.name == 'SafeToBuild' }}
     runs-on: ubuntu-latest
     strategy:
       fail-fast: false
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -19,9 +19,26 @@ on:
       - '**.md'
 
 jobs:
+  check_build_safety:
+    name: Check if Safe to Build
+    runs-on: ubuntu-latest
+    steps:
+      - name: Emit success code
+        if: ${{ github.repository == 'liquibase/liquibase' || github.event.label.name == 'SafeToBuild') }}
+        run: |
+          echo "PR is safe to build"
+          exit 0 // add success code
+
+      - name: Emit failure code
+        if: ${{ github.repository != 'liquibase/liquibase' && github.event.label.name == 'SafeToBuild') }}
+        run: |
+          echo "PR from Fork is NOT safe to build"
+          exit 1 // add failure code
+
+
   setup:
-    if: ${{ github.repository == 'liquibase/liquibase' || (github.repository != 'liquibase/liquibase' && github.event.label.name == 'SafeToBuild') }}
     name: Setup
+    needs: check_build_safety
     runs-on: ubuntu-latest
     outputs:
       proBranchName: ${{ steps.find-branches.outputs.proBranchName }}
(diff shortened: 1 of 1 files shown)


ap-liquibase:
Not sure if subsequent steps  (below `setup` and `build` should also get if-statement looking for SafeToBuild label.

jnewton03 (commented):
@ap-liquibase can we conditionally add an "if owner not liquibase, then require the label" ?

.github/workflows/build.yml
```diff
@@ -3,9 +3,7 @@ name: Build and Test
 on:
   pull_request_target:
     types:
-      - opened
```
nvoxland, line 6: I don't think we can change the types here to labeled. We still want it to run on opened, reopened, and synchronized and don't always want labeled to be required

.github/workflows/build.yml
```diff
@@ -20,6 +18,7 @@ on:
 jobs:
   setup:
     name: Setup
+    if: ${{ github.event.label.name == 'SafeToBuild' }}
```
nvoxland, line 21: Rather than putting this as if blocks within all the steps, maybe add a new top-level job of "Check Build Logic" or something that is just a "fail the build" if it's an external PR and it doesn't have the SafeToBuild label? Then the rest end up not needing to care? The failure would explain to the user and us about how it needs to be labeled by a maintainer

nvoxland (changes requested):

.github/workflows/build.yml
ap-liquibase: yeah. I thought that too. My plan was. to revert this change while adding additional conditionals below.

.github/workflows/build.yml
ap-liquibase: I think thats definitely more elegant but given the time-pressure I'm gonna keep it simpler (for me). The "if" block only needs to be in the setup and maybe the build steps since subsequent steps are dependent via the "needs:" statements

.github/workflows/build.yml
```diff
@@ -19,6 +20,7 @@ on:
 
 jobs:
   setup:
+    if: ${{ github.repository == 'liquibase/liquibase' || (github.repository != 'liquibase/liquibase' && github.event.label.name == 'SafeToBuild') }}
```
StevenMassaro, line 23: You could simplify this to: 
```suggestion
    if: ${{ github.repository == 'liquibase/liquibase' || github.event.label.name == 'SafeToBuild' }}
```

nvoxland: It's not `github.event.label.name` that matters, because we want it to build on all events as long as at some point it was labeled.

So it probably needs to be some sort of "does the PR have the SaveToBuild label on it" which is true when it's first labeled and also on every future build.

.github/workflows/build.yml
```diff
@@ -6,6 +6,7 @@ on:
       - opened
       - reopened
       - synchronize
+      - labeled
```
nvoxland, line 9: Can this be limited to just the SafeToBuild label being added? There is a ton of labels we add and we don't want a new build every time we add labels.

ap-liquibase: Maybe. I saw some examples of looking for specific labels, but I don't think it was up the type section for `pull_request_target`. Checking now to see if there exists a construction for that...

.github/workflows/build.yml
```diff
@@ -18,8 +19,26 @@ on:
...
+    runs-on: ubuntu-latest
+    steps:
+      - name: Emit success code
+        if: ${{ github.repository == 'liquibase/liquibase' || github.event.label.name == 'SafeToBuild') }}
+        run: |
+          echo "PR is safe to build"
+          exit 0 // add success code
+
+      - name: Emit failure code
+        if: ${{ github.repository != 'liquibase/liquibase' && github.event.label.name == 'SafeToBuild') }}
```
StevenMassaro, line 33: This statement looks incorrect to me. I think it should be something like:

```suggestion
        if: ${{ github.repository != 'liquibase/liquibase' || github.event.label.name != 'SafeToBuild') }}
```

Otherwise, no build will ever succeed, since adding the label to a non-liquibase PR will not get past this check.

nvoxland (approved):

jnewton03 (approved):

PR #2550, opened 2022-02-17
Add "KEY" as a keyword to HsqlDatabase adapter

felixgonschorek:


nvoxland:
@felixgonschorek, as far as I can tell, KEY isn't a keyword in hsqldb, just in mysql. Is that correct? Or are you seeing issues in hsql when creating the column named `key` too? 

When I tried it with:
```
    <changeSet author="example" id="1">
        <createTable tableName="foo">
            <column name="id" type="int"/>
            <column name="key" type="int"/>
        </createTable>
    </changeSet>
```
it ran fine for me against hsqldb.

felixgonschorek:
Hi @nvoxland,
thank you for taking a look at this.
I have this changeset:

```
<changeSet
    id="test"
    author="felix">
    <createTable tableName="kvs">
        <column
            name="key"
            type="VARCHAR(255)">
            <constraints nullable="false" />
        </column>
    </createTable>
</changeSet>
```
Which fails with the message:
`org.hsqldb.HsqlException: unexpected token: ( required: ON`

I have created a test project with this test:
https://github.com/felixgonschorek/liquibase-hsqldb-test/blob/main/src/test/java/hsqldb/KeyKeywordTest.java

If I monkeypatch for the "KEY" keyword (testWorksMeWithKeywordPatch()) the test runs fine.
Without patching for "KEY" keyword, the error message above is generated.

I use liquibase and hsqldb in the latest version (afaik)
```
    testImplementation 'org.liquibase:liquibase-core:4.8.0'
    testImplementation "org.hsqldb:hsqldb:2.6.1"
```

Maybe I oversee something but I hope the test helps to clarify things :-)
Thank you again
Felix

felixgonschorek:
Having a second thought, I think the error is the following:

We are using HSQLDB to write tests for an application which uses MySQL in production.
To make HSQLDB compatible with this setup, I am using the HSQLDB MySQL extension (sql.syntax_mys=true) 
We are using the  "ON DUPLICATE KEY UPDATE ..." syntax and the error message I receive seems to relate to that specific syntax. The MySQL extension is using "KEY" as a keyword and expected... could that be the case here?

nvoxland:
Thanks for the repo repository. That helped.

It is coming from the changes in hsql behavior depending on the compatibility mode. Your 

```    
<changeSet id="test" author="felix">
        <createTable tableName="kvs">
            <column
                    name="key"
                    type="VARCHAR(255)">
            </column>
        </createTable>
    </changeSet>
```
changeset runs just fine against `jdbc:hsqldb:mem:lbcat` but fails against `jdbc:hsqldb:mem:lbcat;sql.syntax_mys=true` with `unexpected token: ( required: ON`. There no `ON DUPLICATE KEY` stuff in the test so it's not from that. It is from hsql in mysql mode requiring an ON keyword after KEY.

I can't take in this PR because it changes the behavior for the standard hsql dialect and I'm sure doesn't suddenly make us work with mysql mode completely. Trying to update the HsqlDatabase dialect class to handle not just standard hsql but all the compatibility modes will be too much to be able to support. So we have to keep it as just hsql-standard dialect.

In your case, you have a few options to solve the problem:
1. Don't run liquibase against hsql with sql.syntax_mys=true. That would rely on liquibase's cross-database compatibility to make your changesets run against both mysql and hsql rather than rely on hsql's mysql compatibility. I assume you want the rest of your app to run against hsql in mysql mode, so I'm if this is an option for you or not. Depends on your connection management
2. Add `objectQuotingStrategy="QUOTE_ALL_OBJECTS"` to the changeSets which hit problems with quotes needed differently. 
3. Use `<modifySql>` or just plain `<sql>` and specify your mysql-compliant SQL for changesets that have problems with hsql's mysql mode. That lets you control the sql vs. quoting all objects.

felixgonschorek:
Nathan, thank you very much for having a detailed view at our problem and propose solutions and remedy.
We tried all three fixes, but we learned in the process, that hsqldb is not the perfect match for this specific test scenario.
We switched to using "testcontainers" to start up a docker container with an MySQL instance and are testing against a real MySQL server now. This takes considerably longer for the tests to run, but in the end this was the best approach here.
In other tests we still heavily rely on hsqldb for testing sql statements, so thanks again for taking your time to investigate and thank you and the rest of the hsqldb team for your great work!

PR #2571, opened 2022-02-23
Add logic to Derby.getDefaultDriver() to check for availability of different driver classes

### LINKED ISSUE #1835: Derby driver class name for 10.15.X.X issue
## Environment
**Liquibase Version**: 4.3.5

**Liquibase Integration & Version**: test-harness

**Liquibase Extension(s) & Version**: na

**Database Vendor & Version**: Apache derby 10.15.1.3

## Description
Starting from version 10.15.X.X Derby jdbc driver was changed. The class org.apache.derby.jdbc.ClientDriver.class is not present in the derbyclient.jar anymore. Instead of it they have org.apache.derby.client.ClientAutoloadedDriver.class. This causes throwing exception by Liquibase: 
_liquibase.exception.DatabaseException: java.lang.RuntimeException: Cannot find database driver: org.apache.derby.jdbc.ClientDriver_

I guess the problem is caused by **getDefaultDriver()** method at liquibase-core/src/main/java/liquibase/database/core/DerbyDatabase.java because the driver names are hardcoded.

AlexanderSashchenko:
## Description

Newer versions of derby use a new driver name, and the old driver is no longer available.

Fix for Derby getDefaultDriver() method to load appropriate driver classes based on db version. 

Fixes #1835

### DIFF
--- a/liquibase-core/src/main/java/liquibase/database/core/DerbyDatabase.java
+++ b/liquibase-core/src/main/java/liquibase/database/core/DerbyDatabase.java
@@ -43,27 +43,28 @@ public class DerbyDatabase extends AbstractJdbcDatabase {
         if (url == null) {
             return null;
         } else if (url.toLowerCase().startsWith("jdbc:derby://")) {
-                //Derby client driver class name for versions 10.15.X.X and above.
-                String derbyNewDriverClassName = "org.apache.derby.client.ClientAutoloadedDriver";
-                //Derby client driver class name for versions below 10.15.X.X.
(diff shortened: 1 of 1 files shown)


nvoxland (approved):
The change looks good. I did an auto-reformat and added in a comment that got lost.

nvoxland:
Code review and test results:

Things to be aware of
- Behavior difference is derby 10.15 vs. earlier
- Uses the same "look for the different possible classes in the classloader" trick that we do in MysqlDatabase for a similar problem
- Only impacts derby

Things to worry about
- Nothing

suryaaki2 (approved):

PR #2572, opened 2022-02-23
properly generate changelogs for postgres columns that use SET DEFAULT (DAT-8779)

StevenMassaro:
In Postgres, columns created like this:

```
ALTER TABLE ONLY schema1."AutoBids" ALTER COLUMN "orderIndex" SET DEFAULT nextval('schema1."AutoBids_orderIndex_seq"'::regclass);
```

are reported as both auto increment and default value. This PR updates that code to handle this case.


liquibase-core/src/main/java/liquibase/diff/output/changelog/core/MissingTableChangeGenerator.java
```diff
@@ -36,7 +36,7 @@ public static void setDefaultValue(ColumnConfig columnConfig, Column column, Dat
         Object defaultValue = column.getDefaultValue();
         if (defaultValue == null) {
             // do nothing
-        } else if (column.isAutoIncrement()) {
+        } else if (column.isAutoIncrement() && defaultValue == null) {
```
StevenMassaro, line 39: @nvoxland This feels like it might reach too far, do you think I should constrain this change to only apply to Postgres databases?

liquibase-core/src/main/java/liquibase/diff/output/changelog/core/MissingTableChangeGenerator.java
```diff
@@ -148,7 +148,7 @@ public int getPriority(Class<? extends DatabaseObject> objectType, Database data
             }
             columnConfig.setType(typeString);
 
-            if (column.isAutoIncrement()) {
+            if (column.isAutoIncrement() && column.getDefaultValue() == null) {
```
StevenMassaro, line 151: This could be another place where this change might be best constrained to only Postgres databases.

suryaaki2 (approved):

nvoxland:
> This is a good idea. Just to clarify though, we're making sure that columns which have a default value are not also marked as auto increment, not the inverse.

For postgresql, though, all auto-increment columns have a default database in the database. So what we are doing is trying to  figure out during snapshot is which of those function default values are saying the column is auto-increment, and which are just default values.

With the current changes, I worry it's losing the ability to detect auto-increment columns anymore and treating everything as having a default value. Did you test that @StevenMassaro ?

StevenMassaro:
> > This is a good idea. Just to clarify though, we're making sure that columns which have a default value are not also marked as auto increment, not the inverse.
> 
> For postgresql, though, all auto-increment columns have a default database in the database. So what we are doing is trying to figure out during snapshot is which of those function default values are saying the column is auto-increment, and which are just default values.
> 
> With the current changes, I worry it's losing the ability to detect auto-increment columns anymore and treating everything as having a default value. Did you test that @StevenMassaro ?

Looks like you're right, this is causing problems generating a changelog from a database containing a table like:

```
create table steve(
   stevecol serial
);
```

nvoxland (approved):

StevenMassaro:
@nvoxland and @suryaaki2 You may want to rereview this, I moved the logic up a level but it's largely the same.

nvoxland (approved):

PR #2576, opened 2022-02-23
Remove disruptive and superfluous Exception declaration from ClassLoaderResourceAccessor.close()

famod:
## Environment

**Liquibase Version**: 4.8.0

**Liquibase Integration & Version**: Maven, Quarkus

## Pull Request Type

## Description

The method handles IOException internally and there is no other checked exception, so `throws Exception` really shouldn't be there, as it forces callers to add bogus exception handling.

## Steps To Reproduce

not relevant

## Actual Behavior

n/a

## Expected/Desired Behavior

n/a

n/a

## Additional Context

I'm currently adding calls to this new method to the Quarkus extensions, which is when I found this issue.

### DIFF
--- a/liquibase-core/src/main/java/liquibase/resource/ClassLoaderResourceAccessor.java
+++ b/liquibase-core/src/main/java/liquibase/resource/ClassLoaderResourceAccessor.java
@@ -357,7 +357,7 @@ public class ClassLoaderResourceAccessor extends AbstractResourceAccessor implem
     }
 
     @Override
-    public void close() throws Exception {
+    public void close() {
         if (rootPaths != null) {
             for (final FileSystem rootPath : rootPaths) {
                 try {


nvoxland:
The reason we have it throw Exception is to match the interface method and to allow us to shift the implementation over time which may end up requiring us to throw exceptions.

So worry with taking it off is that while we don't throw an exception right now, if/when we find we have to it will be a breaking API change for everyone who is not attempting to handle anything coming from this method.

Or am I missing anything in that thought process, @famod?

famod:
@nvoxland well, I'm a big fan of the YAGNI and KISS principles. As things are, this declaration is just a burden for callers.

Sure, if you plan to actually throw an exception in the near future it might make sense to declare it.
But if it never comes to pass it's just noise and will probably stay there forever for no good reason.

As for API design (which I'm no expert for!), `Exception` is also _very_ broad which makes it hard to handle properly in case the same try-with-resources block can also throw other exception for more serious cases.

nvoxland:
Yes, Exception is broad but we're also trying to keep the API able to handle a large variety of future options while remaining future-compatible.

I'm going to close this ticket, but if anyone has other thoughts feel free to add to it.

PR #2585, opened 2022-03-01
Fix missing attributes in serialization

tms-91:
## Environment

**Liquibase Version**:  4.7.1+

**Liquibase Integration & Version**: CLI and Java API

**Liquibase Extension(s) & Version**: none

**Database Vendor & Version**: H2 2.1 and MSSQL

## Pull Request Type

## Description

The ChangeSet attributes  "runInTransaction", "runOrder" and "ignore" were never serialized since they were missing in the serializableFields list.

## Steps To Reproduce

1. Have a ChangeLog containing a ChangeSet with  "runInTransaction", "runOrder" and "ignore" attributes with non default values.
2. Let Liquibase read the Changelog.
3. Export the ChangeLog with the for example the XMLChangeLogSerializer

## Actual Behavior
After exporting the ChangeLog again the attributes "runInTransaction", "runOrder" and "ignore" are missing on the ChangeSet.

## Expected/Desired Behavior
The ChangeSet retaining the attributes "runInTransaction", "runOrder" and "ignore" after export.

- [x] Added [Unit Test(s)](https://liquibase.jira.com/wiki/spaces/LB/pages/1274937609/How+to+Write+Liquibase+Core+Unit+Tests)
extended already existing Tests
- [ ] Documentation Updated

## Need Help?


liquibase-core/src/main/java/liquibase/changelog/ChangeSet.java
```diff
@@ -955,14 +955,26 @@ public boolean isAlwaysRun() {
         return alwaysRun;
     }
 
+    public void setAlwaysRun(boolean alwaysRun) {
```
nvoxland, line 958: Why did you make these settable @tms-91 ? 

We've been trying to keep ChangeSet immutable vs. how it was originally constructed to be safe. But do you have a use case where you need to change them?

nvoxland (changes requested):

nvoxland (approved):

liquibase-core/src/main/java/liquibase/changelog/ChangeSet.java
nvoxland: Thanks! Looks good

nvoxland:
Code Review and test results:

Things to be aware of
- Changes make sense and have good tests added for them
- Liquibase code doesn't use the new behavior, only external callers

Things to worry about
- Nothing

XDelphiGrl:
Hello, @tms-91! Thanks for the PR! Is the reason we do not have a test for the formatted SQL changelog because there is no serializer for the "raw" SQL format? 

@nvoxland, hi - Why wouldn't this code be used by the CLI? I am able to add `runOrder`, `ignore` and `runInTransaction` on an XML changeset and successfully validate and update using the changelog. Is it because the CLI does not use XMLChangeLogSerializer to export the changelog?

wwillard7800 (approved):
Looks good.

nvoxland:
> @nvoxland, hi - Why wouldn't this code be used by the CLI? I am able to add `runOrder`, `ignore` and `runInTransaction` on an XML changeset and successfully validate and update using the changelog. Is it because the CLI does not use XMLChangeLogSerializer to export the changelog?

@XDelphiGrl Liquibase uses XMLChangeLogSerializer to generate the XML from constructed changelogs, but we don't set runOrder, ignore, or runInTransaction to the changesets we generate. So there isn't a way from our code (currently) for them to get set.

XDelphiGrl:
@nvoxland, Liquibase `update-sql` and 'update' do use `runOrder` and `ignore` in an XML changelog. If you use this changelog:

```
<databaseChangeLog 
xmlns="http://www.liquibase.org/xml/ns/dbchangelog" xmlns:ext="http://www.liquibase.org/xml/ns/dbchangelog-ext" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" 
xsi:schemaLocation="http://www.liquibase.org/xml/ns/dbchangelog-ext http://www.liquibase.org/xml/ns/dbchangelog/dbchangelog-ext.xsd 
http://www.liquibase.org/xml/ns/dbchangelog http://www.liquibase.org/xml/ns/dbchangelog/dbchangelog-latest.xsd">


	<changeSet author="liquibase-user" id="runLast" labels="runOrderWorksQuestionMark" runOrder="last">
        <createTable tableName="testtable1">
            <column name="supplier_id" type="numeric(4)"/>
            <column name="supplier_name" type="VARCHAR(50)"/>
        </createTable>
    </changeSet>

	<changeSet author="liquibase-user" id="runFirst" labels="runOrderWorksQuestionMark" runOrder="first">
        <createTable tableName="testtable2">
            <column name="supplier_id" type="numeric(4)"/>
            <column name="supplier_name" type="VARCHAR(50)"/>
        </createTable>
    </changeSet>
	
	<changeSet author="liquibase-user" id="ignoreMe" labels="runOrderWorksQuestionMark" ignore="true">
        <createTable tableName="testtable2">
            <column name="supplier_id" type="numeric(4)"/>
            <column name="supplier_name" type="VARCHAR(50)"/>
        </createTable>
    </changeSet>
</databaseChangeLog>
```
Then update-sql will show that the second changeset runs first (as set in runOrder) and the first changeset runs last (again, in runOrder). The third changeset is entirely ignored. 

When you update, the changes run in the order shown by update-sql:

```
Running Changeset: runOrder-ignore-changelog.xml::runFirst::liquibase-user
Running Changeset: runOrder-ignore-changelog.xml::runLast::liquibase-user
```
I don't know how to validate `runInTransaction` but it could also be the case that the CLI knows/uses that, too. 

I'm testing with a feature branch that is (relatively) up-to-date with liquibase/liquibase master. I believe this test should be executed against this PR's build to ensure that there are no regressions (at least for `runOrder` and `ignore`). I'm going to remove the autocandidate tag from this ticket. 

CC @FBurguer @ecarneiro01 @kataggart

tms-91:
> Hello, @tms-91! Thanks for the PR! Is the reason we do not have a test for the formatted SQL changelog because there is no serializer for the "raw" SQL format?

Hello @XDelphiGrl ! Thanks for reviewing this PR. 

From a quick glance at the code it doesn't seem to me that the FormattedSqlChangeLogSerializer uses the getSerializableFields() method of ChangeSet, so there should be no changes in behaviour here. I might have overlooked something though.

There should also be no changes to the parsing (reading of changelogs) since ChangeSet does not extend AbstractLiquibaseSerializable and implements its own load and serialize (stub in case of serialize) methods. The load method of ChangeSet doesnt use the getSerializableFields method either and is unchanged. 

An example use case that would be affected (and fixed) by this PR is the conversion between different Changelog file formats by using Liquibase as a library (not cli):

1. let Liquibase read the changelog file and create the Java objects.
2. use the XML/YAML/JSON-Serializers to export the Java objects to the output changelog file

In this case, without this PR, if a changeset had "runInTransaction", "runOrder", or "ignore" set to non-default values, they would be missing in the generated output changeset and treated as if they had default values, which would alter the semantics of the changeset.

XDelphiGrl:
Hello, @tms-91! Thank you for the excellent explanations in response to my questions. Based on your answers, I will move this ticket to Ready to Merge & you should see your fix in the next release. 

It makes my day to hear from someone in the Community - thank you for that!

XDelphiGrl (approved):
* Changes impact consumers of Liquibase as a library.
  * No impact to the CLI or other integrations.
* Tests added for each of the serializable changelog formats (XML, JSON, YAML).
  * Formatted SQL changelogs are not impacted by this fix.
* @tms-91 is an excellent collaborator. As a QA, I love Devs who take the time to explain the code to me.

APPROVED

[Passing Functional Tests](https://github.com/liquibase/liquibase-pro-tests/actions/runs/2517623764)
[Passing Test Harness Execution](https://github.com/liquibase/liquibase-test-harness/actions/runs/2517625475)

PR #2589, opened 2022-03-01
Reset the ranChangesets cache if we know there were changes to the da…

### LINKED ISSUE #2580: Issue updating older md5 checksums
## Environment

4.7.1, 4.8.0:

**Liquibase Integration & Version**: Calling directly from code

**Database Vendor & Version**: H2, Postgres, MySQL

**Operating System Type & Version**: Mac, linux

## Description

When the md5 checksum from an older version of Liquibase is in DATABASECHANGELOG and you call `Liquibase.listUnrunChangeSets` liquibase is supposed to update the md5sum. First by setting it to null, and then subsequently updating it. However, calling `listUnrunChangeSets` actually sets the md5sum to null, but never recalculates it because in the in-mem cached list of ChangeLogs, the version number is still the old version number. When calling `Liquibase.update` (which calls `listUnrunChangeSets`), the update will actually fail, because even though it nulls the actual entry, the cache still contains the old checksums, and those are not going to be equal to the newly calculated checksums 

## Steps To Reproduce

Build app with liquibase v3.5.3
Run app with change log and call `Liquibase.update`
Build app with liquibase v4.8.0
Run app with change log and call `Liquibase.update`
See it fail, complaining about incompatible md5 checksums
```
liquibase.exception.ValidationFailedException: Validation Failed:
     1 change sets check sum
          changeset1.sql::raw::includeAll was: 1:xxx but is now: 8:d0b70c31a2294f905d221f34acfa29eb


	at liquibase.changelog.DatabaseChangeLog.validate(DatabaseChangeLog.java:296)
	at liquibase.Liquibase$16.run(Liquibase.java:1988)
	at liquibase.Scope.lambda$child$0(Scope.java:180)
	at liquibase.Scope.child(Scope.java:189)
	at liquibase.Liquibase.runInScope(Liquibase.java:2405)
	at liquibase.Liquibase.listUnrunChangeSets(Liquibase.java:1978)
	at liquibase.Liquibase.listUnrunChangeSets(Liquibase.java:1969)
```

Integration test. Add this to AbstractIntegrationTest.java and run H2IntegrationTest.java:
```
    @Test
    public void testUpdateChangelogChecksum() throws Exception {
        assumeNotNull(this.getDatabase());

        String schemaName = getDatabase().getDefaultSchemaName();
        if (schemaName == null) {
            return;
        }

        getDatabase().setOutputDefaultSchema(false);
        getDatabase().setOutputDefaultCatalog(false);

        Liquibase liquibase = new Liquibase(includedChangeLog, new JUnitResourceAccessor(), database.getConnection());
        liquibase.setChangeLogParameter("loginuser", testSystem.getUsername());
        liquibase.update(contexts);

        Connection conn = ((JdbcConnection) database.getConnection()).getUnderlyingConnection();
        conn.createStatement().execute("update DATABASECHANGELOG set md5sum = '1:xxx'");
        conn.commit();

        liquibase.getDatabase().getRanChangeSetList();
        // uncommenting the next line will make the test pass again
        //         ChangeLogHistoryServiceFactory.getInstance().resetAll();
        liquibase.update(contexts);
    }
```
The sequence `getRanChangeSetList`, `update` is a minimal sequence that shows this behaviour that I narrowed it down to.

## Actual Behavior
Exception saying md5sum validation failed the first time
Update happens correctly the second time

## Expected/Desired Behavior
First time running update, the md5sum is updated to the current version

nvoxland:
## Description

Fixes #2580 by resetting the ranChangesets cache if we know there were changes to the databasechangelog table (including checksum clearing) in the init() logic.

### DIFF
--- a/liquibase-core/src/main/java/liquibase/changelog/StandardChangeLogHistoryService.java
+++ b/liquibase-core/src/main/java/liquibase/changelog/StandardChangeLogHistoryService.java
@@ -286,6 +286,11 @@ public class StandardChangeLogHistoryService extends AbstractChangeLogHistorySer
                     " " + getDatabase().getShortName() + " when checking databasechangelog table");
             }
         }
+
+        if (statementsToExecute.size() > 0) {
+            //reset the cache if there was a change to the table. Especially catches things like md5 changes which might have been updated but would still be wrong in the cache
+            this.ranChangeSetList = null;
+        }
         serviceInitialized = true;
     }
 
--- a/liquibase-integration-tests/src/test/java/liquibase/dbtest/AbstractIntegrationTest.java
+++ b/liquibase-integration-tests/src/test/java/liquibase/dbtest/AbstractIntegrationTest.java
@@ -1081,4 +1081,30 @@ public abstract class AbstractIntegrationTest {
     public void setDefaultSchemaName(String defaultSchemaName) {
         this.defaultSchemaName = defaultSchemaName;
     }
+
+    @Test
+    public void testUpdateChangelogChecksum() throws Exception {
+       //NOTE: This test does as much to test the handing of the StandardHistoryService.ranChangeSetList cache when the underlying checksums need to be updated as anything
+       //NOTE: The end-user behaviour it's replicating is: `liquibase update` with a version that uses an old checksum version then `liquibase update` with the current version
+        assumeNotNull(this.getDatabase());
+
+        String schemaName = getDatabase().getDefaultSchemaName();
+        if (schemaName == null) {
+            return;
+        }
+
+        getDatabase().setOutputDefaultSchema(false);
+        getDatabase().setOutputDefaultCatalog(false);
+
+        Liquibase liquibase = new Liquibase(includedChangeLog, new JUnitResourceAccessor(), database.getConnection());
+        liquibase.setChangeLogParameter("loginuser", testSystem.getUsername());
+        liquibase.update(contexts);
+
+        Connection conn = ((JdbcConnection) database.getConnection()).getUnderlyingConnection();
+        conn.createStatement().execute("update DATABASECHANGELOG set md5sum = '1:xxx'");
+        conn.commit();
+
+        liquibase.getDatabase().getRanChangeSetList();
+        liquibase.update(contexts);
+    }
 }


nvoxland:
Code review and test results:

Things to be aware of
- I tried to keep this PR as isolated as possible. There's a general cleanup planned for the changelog history service, so don't want to solve the problem too generally right now. But the "update should run on old checksum versions without a validation error" bug is a major issue
- The existing logic that determines if md5sums should be updated looks only at a the first row that comes back from the databasechangelog table. There are possibly times that can be wrong in real life, but changing that was more than I'd like to get into at this point. It might also never actually be a problem in real life either. But it bit me when trying to validate the fix because I only updated the checksum in the 2nd of my two ran changeSets.

Things to worry about
- Nothing

suryaaki2 (approved):

XDelphiGrl:
MD5 checksums generated by older versions of Liquibase are now correctly updated in the DATABASECHANGELOG table when a more recent version of Liquibase executes against the database. 

**Notable&Nice** 
Thank you, @leblonk for suggesting an integration test to demonstrate the problem in [GH Issue](https://github.com/liquibase/liquibase/issues/2580)
Thank you, @nvoxland for adding the test along with the fix in this PR.

---

### Test Methodology
I was unable to reproduce this bug as an end user at the CLI. I validated v7 of Liquibase MD5 checksums are correctly updated to to the latest (v7) checksum version when a newer Liquibase client is used during an update. Here is the workflow I followed:
* Install Liquibase 3.3.5.
* Deploy three tables.
* Validate the DBCL row for each createTable changetype has a v7 checksum, i.e., `7:22008842caa48f09ca3f57e2868a826e`
  * `SELECT * FROM DATABASECHANGELOG;`
* Install Liquibase 4.8.0
* Execute update-sql to validate the correct messaging and UPDATE statements are generated to update the MD5 checksums.
* Deploy a forth table.
* Validate all rows in the the DBCL have a v8 checksum.

---
* Verify update-sql generates UPDATE statements to re-version MD5 checksums generated by an older Liquibase client. **PASS**
```
UPDATE public.databasechangelog SET MD5SUM = NULL;
UPDATE public.databasechangelog SET MD5SUM = '8:b04ec4e3e870123b4e32ac75e05c0ae5' WHERE ID = '1' AND AUTHOR = 'r2' AND FILENAME = 'checksumsTest-changelog.xml';
UPDATE public.databasechangelog SET MD5SUM = '8:cabe9529c9277678de8b0834fb367648' WHERE ID = '2' AND AUTHOR = 'liqbibase' AND FILENAME = 'checksumsTest-changelog.xml';
UPDATE public.databasechangelog SET MD5SUM = '8:41d3498f8a9e799572c8243100542af8' WHERE ID = '3' AND AUTHOR = 'liquibase' AND FILENAME = 'checksumsTest-changelog.xml';
UPDATE public.databasechangelog SET MD5SUM = '8:6ae429778bd51ab85eb1ddd44a70c6ca' WHERE ID = '4' AND AUTHOR = 'liquibase' AND FILENAME = 'checksumsTest-changelog.xml';
```

* Verify update-sql output includes human readable/comment information explaining the MD5 changes. **PASS**
```
DatabaseChangeLog checksums are an incompatible version.  Setting them to null so they will be updated on next database update
```

* Verify update successfully re-versions the MD5 checksums. **PASS**

### 

---
**Test Environment**
Liquibase Core fix-historyservice-cache-on-md5-upgrade/1685/717dcd, Pro: master/658/91c564
Postgres 12.6
[Functional Tests](https://github.com/liquibase/liquibase-pro-tests/actions/runs/1918717095)
* There are MySQL Pro feature test failures related to a mismatch between test validation text and output to the console. Not a blocker for this PR (completely unrelated).

PR #2477, opened 2022-02-04
BUGFIX: Fix H2 auto increment syntax

drivera-armedia:
H2 has updated such that the old auto increment syntax no longer works

## Environment

**Liquibase Version**: 4.7.1

**Liquibase Integration & Version**: Java embedded

**Liquibase Extension(s) & Version**: no extensions

**Database Vendor & Version**: H2 2.1.210

**Operating System Type & Version**: Ubuntu Linux 21.04

## Pull Request Type

- [x] Bug fix (non-breaking change which fixes an issue.)
- [ ] Enhancement/New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to change)

## Description

H2 has modified its syntax to no longer support AUTO_INCREMENT unless being used in MySQL mode. In other modes, it's been advised that the standard GENERATED AS IDENTITY syntax. Furthermore, it doesn't support the _exact_ syntax so the comma used between the START WITH and INCREMENT BY clauses needed to be (configurably) removed, so I added a means to do just that which only takes effect for H2Database.

## Steps To Reproduce

List the steps to reproduce the behavior.
- Create a changeset which includes an autoIncrement column
- Apply said changeset to an H2 database in PostgreSQL mode

## Actual Behavior
- Changeset will fail with unrecognized syntax on AUTO_INCREMENT

## Expected/Desired Behavior
- Changeset will succeed and properly render tables with/modify existing auto-increment columns

## Fast Track PR Acceptance Checklist:

- [x] Build is successful and all new and existing tests pass
- [x] Added [Unit Test(s)](https://liquibase.jira.com/wiki/spaces/LB/pages/1274937609/How+to+Write+Liquibase+Core+Unit+Tests)
- [x] Added [Integration Test(s)](https://liquibase.jira.com/wiki/spaces/LB/pages/1276608569/How+to+Write+Liquibase+Core+Integration+Tests)
- [ ] Added [Test Harness Test(s)](https://github.com/liquibase/liquibase-test-harness/pulls)
- [ ] Documentation Updated

PR #2597, opened 2022-03-03
Refactor start-h2 logic

### LINKED ISSUE #2481: Simplify start-h2 usage and code
## Description

Now that we have introduced `liquibase init start-h2` as a command, there is no need to have the start-h2/start-h2.bat scripts which are sensitive to h2 driver upgrades. This will require updating documentation talking about start-h2 as a separate command in favor of `liquibase init start-h2`

As part of that change, do three minor bugfixes as well:
1. Refactor the StartH2Main code so it lives in StartH2CommandStep instead of being a separate class
2. Update the error message about `the h2-<version>.jar file in liquibase/lib` to be `the h2 jar file in liquibase/lib` to avoid the filename that looks like it has `<version>` in it and better describe what version will run (since users can change it).
3. Force the http url shown and opened to be "localhost" rather than whatever "what is the server's IP address" logic h2 uses. We've seen VPN setups that confuse h2's logic

## Acceptance Criteria

1. Retain start-h2 files in the examples directory, but they call `liquibase init start-h2` and not `StartH2Main`
2. No documentation references to start-h2, only `liquibase init start-h2`
3. No more StartH2Main class
4. If you run `liquibase init start-h2` with no h2 jar file in your liquibase/lib dir, the error message is valid
5. Browser opens to "localhost" not an IP address
6. URLs shown in the console output are "localhost" and not an IP address
7. Tests for the refactored StartH2CommandStep

nvoxland:
## Description

Moved StartH2Main from being a stand-alone command to being integrated into the StartH2CommandStep.

The old examples/start-h2 scripts are still there, but are now calls to `liquibase init start-h2` rather than a separately boostrapped java call to StartH2Main.

Fixes #2481


nvoxland:
Code review and test results:

Things to be aware of
- There were two copies of the start-h2 scripts, I deleted the uneeded copy and made sure the remaining copy was correctly copied (preserved unix permissions and dos/unix line endings ) into the final tar.gz file
  - Would like a double-check that these files still work from official build
- The code in StartH2CommandStep is copy-pasted from the old StartH2Main
  - But I did move some hard-coded settings like ports and username/password to be arguments now that that is easy to do
- I found the `h2.bindAddress` system property, which looks like the one h2 uses to decide the IP/host to connect to. Without that setting, it would guess at the best one and sometimes get it wrong. 
  - In our case, we want it to be localhost for security reasons, but there is a new arg for people who don't want that
- There is a corresponding pro PR and also Datical/liquibase-docs#137

Things to worry about
- Does the old start-h2 scripts still work on mac/linux and windows? Does liquibase init start-h2 still work?
  - The "does it start" is the main thing that changed. The logic within is mostly the same.

suryaaki2 (approved):

nvoxland:
> @nvoxland Are you going to remove the `com.datical.liquibase.ext.command.init.StartH2CommandStep` from the pro code?

Yes, there is a corresponding PR that removes that

StevenMassaro:
@nvoxland I've noticed that when I CTRL+C in the prompt where H2 is running, it takes a few seconds before the command is interrupted. Is that something that can be fixed, and is it fixed by your changes here?

nvoxland:
I added a commit to better handle that pause after "ctrl-c"

I also updated the readme and getting_started pages

FBurguer:
This fix adds a new command  `liquibase init start-h2` to start an example h2 DB, keeping the bat script (the bat command uses `liquibase init start-h2` now too). The only functional test that i could think so are using  `liquibase init start-h2`  command with this build and then use the bat file on an older build and compare their behavior. 
Both ways work the same way so i procided to check bugfixes and the things to be aware of by Nathan:

1. Retain start-h2 files in the examples directory, but they call liquibase init start-h2 and not StartH2Main. `PASS`
2. No documentation references to start-h2, only liquibase init start-h2 `PASS`
3. If you run liquibase init start-h2 with no h2 jar file in your liquibase/lib dir, the error message is valid `PASS`
4. Browser opens to "localhost" not an IP address `FAIL`
5. URLs shown in the console output are "localhost" and not an IP address `FAIL`
(4 and 5 failed in my environment but after trying on my coworker environement it looks like my case is uncommon, it is a thing to be aware of though)

Test environment
Liquibase Version: [Core: //start-h2-simplify/2477/50379f/2022-05-02 17:04+0000, Pro: start-h2-simplify/1078/c6d556/2022-05-02T16:59:19Z] #2477 built at 2022-05-02 17:04+0000
h2 2
jdk 1.8.0_291

PR #2599, opened 2022-03-04
Added new liquibase.monitorPerformance setting for CLI

nvoxland:
## Description

Adds a new `monitorPerformance` setting that is used only by the CLI to enable JDK Flight Recorder on startup. 

This makes it easier for us to get performance data from users seeing slowness problems, without having to explain how to set a `-XX:StartFlightRecording=disk=true,dumponexit=true,filename=liquibase.jfr,maxsize=1024m,settings=profile` JVM argument.

## Usage

- Running `liquibase --monitor-performance=false update` (default setting) will NOT do any additional profiling
- Running `liquibase --monitor-performance=true update` will  save a `liquibase-TIMESTAMP.jfr` file in your current working directory containing the performance data
- Running `liquibase --monitor-performance=/tmp/liquibase.jfr` will save a `/tmp/liquibase.jfr` file containing the performance data

NOTE: If running under a JVM version less than 11, the flag is ignored.

On startup, a message is output to stderr if monitoring is enabled or an exception happens trying to enable monitoring. If monitoring is disabled, a FINE-level log is output

Docs PR: Datical/liquibase-docs#145


nvoxland:
Code notes and test results:

Things to be aware of
- Startup code had to be written via reflection, since the classes do not exist in the java-8 version we build with
- Entire code is in a try/catch block so any problems with enabling should never block running
- Only applies to the CLI at this point. May want to expand it at some point, but not yet
- The saved file can be opened by installing https://jdk.java.net/jmc/8/

Things to worry about
- Nothing

liquibase-cli/src/main/java/liquibase/integration/commandline/LiquibaseCommandLine.java
```diff
@@ -360,6 +366,53 @@ public int execute(String[] args) {
...
+            String filename = monitorPerformanceValue;
+            if (filename.equalsIgnoreCase("true")) {
+                filename = "liquibase-" + new ISODateFormat().format(new Date()).replaceAll("\\W", "_") + ".jfr";
+            }
+            if (!filename.endsWith(".jfr")) {
+                filename = filename + ".jfr";
+            }
+
+            final Class<?> configurationClass = Class.forName("jdk.jfr.Configuration");
+            final Class<?> recordingClass = Class.forName("jdk.jfr.Recording");
```
neupsh, line 396: This could be a good place to catch class not found exception here and print out some helpful messages here. For example, letting the user know that they need to be running liquibase with jdk instead of jre.

liquibase-core/src/main/java/liquibase/integration/commandline/LiquibaseCommandLineConfiguration.java
```diff
@@ -63,12 +64,18 @@
         LOG_FILE = builder.define("logFile", File.class).build();
         OUTPUT_FILE = builder.define("outputFile", File.class).build();
 
+        MONITOR_PERFORMANCE = builder.define("monitorPerformance", String.class)
+                .setDescription("Enable performance tracking. Set to 'false' to disable. If set to 'true', data is stored to a `liquibase-TIMESTAMP.jfr` file in your working directory. Any other value will enable tracking and be used as the name of the file to write the data to.")
```
neupsh, line 68: we can include the jdk requirements here too.

kataggart: @nvoxland looks like @neupsh has some suggestions; could you review? Thanks!

liquibase-cli/src/main/java/liquibase/integration/commandline/LiquibaseCommandLine.java
nvoxland: We're catching whatever exceptions happen and then logging that out, including class not found exceptions. With Java 11+ there isn't really JREs anymore so assuming something like that in the output gets harder to do. I think the existing message about "can't enable logging because of e.getMessage()" is probably as a good as we can do

liquibase-core/src/main/java/liquibase/integration/commandline/LiquibaseCommandLineConfiguration.java
nvoxland: It should work with any Java 11+ version, and nothing under 11 is supported by oracle so I didn't think it was worth calling out in there. If someone tries to run with an old version it will tell them at the time which is probably good enough?

suryaaki2 (approved):

PR #2602, opened 2022-03-04
Updated Installer JDK to 17

nvoxland:
## Description

Updates the version of Java installed as part of Liquibase in the installers to the latest Java 17 rather than 11, since 17 is the newest LTS version.


nvoxland:
Notes and test results:

Things to be aware of
- Branched this off master rather than #2318 to make managing the two PRs independently easier. That does mean we're not testing anything with 17 in this branch right now, but after 2318 gets merged, we can sync master into this branch and then we will.
- This only impacts the version of java installed with the install4j-based installers
- File handling by installer (see below)

Things to worry about
- Nothing

_File Handling Notes_

My main worry was whether install4j would cleanly remove files from the 11 jvm that no longer exist in 17. If it didn't, that's also a potential problem even as we do patch releases. I know install4j doesn't remove the old driver jars on upgrade, but thought perhaps it handled it's bundled JVM differently.

To test the jdk file handling, I:

1. Installed an older version of Liquibase with the installer
2. Ran `tree` to list all the files in the installed dir
3. Upgraded that installation to a build from this branch
4. Ran `tree` on the upgraded dir
5. Compared the output from 2 and 4

From the comparison, I saw that new files like the h2 driver in the lib were added, but "removed" files were not actually removed in the lib dir as expected. 

BUT, the `jre` dir DID correctly had old files removed from it in addition to newly added files. So the java upgrade is handled well out of the box by install4j

I did test this only on windows though, not mac.

suryaaki2 (approved):

FBurguer:
For this Pr i compared the directories of an updated and a freshly installed builds from the windows installer, the only difference between them is autoUninstall file (there are 2 in the updated version). I also checked if liquibase is installing and running by default in java 17, and it looks like it. I also tried running a couple of commands on both builds and there is not any noticeable issue in any of them.

Test environment:
`Liquibase version: Liquibase Version: [Core: //jdk-17-installer/2546/9eb094/2022-05-09 15:52+0000, Pro: master/1193/d83fd3/2022-05-09T15:14:48Z]
Liquibase Community [Core: //jdk-17-installer/2546/9eb094/2022-05-09 15:52+0000, Pro: master/1193/d83fd3/2022-05-09T15:14:48Z] by Liquibase`
Database: h2 2
SO: Windows 10
Java: 17.0.3

PR #2614, opened 2022-03-08
move hub tests and infrastructure code to pro repo (DAT-9288)

StevenMassaro:


suryaaki2 (approved):

.github/workflows/build.yml
```diff
@@ -70,32 +70,32 @@ jobs:
           github-token: ${{ secrets.BOT_TOKEN }}
           script: |
             const helper = require('./.github/util/workflow-helper.js')({github, context});
-            
+
```
StevenMassaro, line 73: I can undo these formatting/whitespace changes.

StevenMassaro: Looks like Intellij did this automatically when resolving merge conflicts, so I'm going to leave it so we don't have to deal with it later.

.github/workflows/build.yml
```diff
@@ -270,7 +270,7 @@ jobs:
           password: ${{ secrets.ARTIFACTORY_TOKEN }}
 
       - name: Run Tests
-        run: mvn -B jar:jar jar:test-jar surefire:test -DtrimStackTrace=false -Dliquibase.sdk.testSystem.test=hub,${{ matrix.testSystem }} -Dliquibase.sdk.testSystem.acceptLicenses=${{ matrix.testSystem }} -Dtest=*IntegrationTest,*ExecutorTest -DfailIfNoTests=false
```
nvoxland, line 273: I think taking "hub" out of here will make the integration tests get skipped, since they are currently written to expect a hub system to be there. 

We'll have to either remove the hub usage (whatever that was) from the integration tests so they are not skipped or figure out a way to keep a hub available for testing enough in this codebase for these tests to still pass.

nvoxland (approved):
I was worried that the existing integration tests would get skipped without the HubTestSystem being there, but I think that since you took the hub reference out of AbstractIntegrationTest, I think they will continue to run correctly.

StevenMassaro:
> I was worried that the existing integration tests would get skipped without the HubTestSystem being there, but I think that since you took the hub reference out of AbstractIntegrationTest, I think they will continue to run correctly.

They didn't get skipped, they failed, which is why I removed the HubTestSystem. I did open up a second PR to test the GitHub actions changes and the tests passed there.

PR #2643, opened 2022-03-15
preserve default value of invalid autoincrement column types in Postgres (DAT-8670)

StevenMassaro:
Postgres reports that varchar columns are autoincrement and have a default value when created like this:

```
CREATE SEQUENCE mysequence INCREMENT 5 START 100;
CREATE TABLE products ( product_no varchar(20) DEFAULT nextval('mysequence'));
```

and this change allows Liquibase to snapshot these columns properly. The default value is preserved, and the autoincrement information is removed.


suryaaki2 (approved):

liquibase-core/src/main/java/liquibase/datatype/core/VarcharType.java
```diff
@@ -12,6 +12,20 @@
 @DataTypeInfo(name="varchar", aliases = {"java.sql.Types.VARCHAR", "java.lang.String", "varchar2", "character varying"}, minParameters = 0, maxParameters = 1, priority = LiquibaseDataType.PRIORITY_DEFAULT)
 public class VarcharType extends CharType {
 
+    /**
+     * While surprising, Postgres allows varchar columns to be autoincrement.
+     */
+    private boolean autoIncrement;
```
nvoxland, line 18: Rather than expanding our generic concept of auto-increment to varchars (and later to whatever other types postgresql allows to be "auto-increment", is it maybe better to not mark it as auto-increment but instead just preserve the default value?

In this case, it's probably less that the column is "auto-increment" and more that it has a default value that happens to pull from a sequence? 

So it would be to update the snapshot logic that determines if it's auto-increment on postgresql and therefore strips the defaultValue to only do that if the type is an int/bigint/smallint?

nvoxland (changes requested):

liquibase-core/src/main/java/liquibase/datatype/core/VarcharType.java
StevenMassaro: I'm open to that. I was concerned that we'd keep discovering that Postgres allows you to auto-increment whatever column type you like, and we'd end up with auto-increment properties on every type, even though it doesn't really make sense.

On the other hand, I wasn't sure if the conditional logic to strip the auto-increment property from the column was "okay". It felt a bit like we were thinking we knew better than the database. However, the actual outputs from the way you suggested are better, since the table DDL now includes the default value!

Let me know if what I came up with here is what you had in mind.

nvoxland (approved):

PR #2651, opened 2022-03-17
DAT-9524 enable apple notarization of macos installer 

ap-liquibase:
DAT-9524 enable apple notarization of macos installer


jnewton03 (commented):
@ap-liquibase looks like there are some comments that can be cleaned up (lines 25, 32-35).  Are there any alternatives to `sleep` commands?  They tend to be inconsistent :)

jnewton03 (approved):

Issue #2586, opened 2022-03-01
The <property /> definition for column type does not work in version 4.8.0, which works well in 3.x.x


JosephCen:
## Environment

**Liquibase Version**: 4.6.1

**Liquibase Integration & Version**: spring boot

**Liquibase Extension(s) & Version**: <n/a>

**Database Vendor & Version**: postgres-12 in container 'postgres:12-alpine'

**Operating System Type & Version**: Mac OS

## Description

An exception `java.lang.ArrayIndexOutOfBoundsException` happened when run with below change log.

``` xml
<databaseChangeLog ...>

    <property name="bytesarray_type" value="BYTEA" global="false" dbms="postgresql" />
    <property name="bytesarray_type" value="java.sql.Types.BLOB" global="false" dbms="hana" />

    <changeSet author="i521084" id="1583641912012-4">
        <createTable tableName="asyncevent">
            <column name="id" type="BIGINT">
                <constraints primaryKey="true" primaryKeyName="asyncevent_pkey"/>
            </column>
            <column name="messagebodybytes" type="${bytesarray_type}"/>
            <!-- ... -->
        </createTable>
    </changeSet>
</databaseChangeLog>
```

## Steps To Reproduce

1. Using postgres DB in docker image 'postgres:12-alpine'.
1. Using a changelog as above.
1. Run the changelog with 'liquibase.integration.spring.SpringLiquibase.performUpdate'.
1. The exception will happened with handle above 'create table' change set.

## Actual Behavior
An exception `ArrayIndexOutOfBoundsException` happened.

## Expected/Desired Behavior
Relevant table should be created with above change log. And the same change log works well under Liquibase version 3.x.x.

JosephCen:
This issue is same to https://github.com/liquibase/liquibase/issues/2231. But NOT fixed well in release 4.8.0.

JosephCen:
In release 4.8.0, there is a 'isDuplicate(...)' function added to check whether new added parameter is existing. But the logic of 'isDuplicate(...)' may not correct.

Relevant code:
https://github.com/liquibase/liquibase/blob/v4.8.0/liquibase-core/src/main/java/liquibase/changelog/ChangeLogParameters.java#L143

Please take a look at below screen-shot:
![image](https://user-images.githubusercontent.com/15362546/156156621-8c1ae440-2f0b-4e4a-b0f3-a81b8b9871e9.png)
[...]

PR #2656, opened 2022-03-19
Add configuration options for missing property handling

dwieland:
## Pull Request Type

- [x] Enhancement/New feature (non-breaking change which adds functionality)

## Description

Adds a new `liquibase.missingPropertyMode` setting with options of `PRESERVE`, `EMPTY`, and `ERROR`. The default value is PRESERVE which follows the old behavior of keeping the unexpanded property in the changelog file. EMPTY will replace it with an empty value, and ERROR will exit parsing with an error.

Example: if a changelog as `Using value ${myValue} here` and myValue is not defined, the result will be:
- PRESERVE: `Using value ${myValue} here`
- EMPTY: `Using value  here`
- ERROR: Exits with error

This PR also does a cleanup/refactoring of the changelog property handling and the classes/methods associated with it to make the mode more maintainable and usable, as well as better avoid bugs such as #2586 and the ArrayIndexOutOfBoundsException issue listed below

#### Overview 

The initial issue I encountered was a strange `ArrayIndexOutOfBoundsException` when running Liquibase. After some analysis I found out that it was related to #2586 - I tried to use specific varchar specifications for Oracle (`VARCHAR(100 CHAR)`) while keeping compatibility with MySQL. So I specified the type like this: `type: VARCHAR(100 ${sql.type.varchar.char})`, triggering the above issue as I forgot the set the property explicitly for MySQL.

After investigation I found that property placeholders are left untouched if the property could not be found in the current scope. After that the `DataTypeFactory` inspects the data type and assumes some embedded information based on the occurrence of a `{`. It tries to split the string in key and value and fails horribly doing so.

This PR tries to address multiple aspects of the issue. It allows to configure the behavior when trying to expand a placeholder of an unknown property - keeping backwards compatibility by default. The two new modes are: "replace with empty string" and "throw meaningful exception". It also tries to give a more meaningful exception when some placeholder reaches the `DataTypeFactory`.

#### Steps To Reproduce

Use a placeholder in `createTable.columns.column[].type` that is not backed by a property.

#### Actual Behavior

`java.lang.ArrayIndexOutOfBoundsException` is thrown, leaving the normal user clueless about what actually happened.

#### Expected/Desired Behavior

A meaningful error is presented to the user. The behavior in the described case is configurable.

Please tell me if additional unit or integration tests are needed.
I can update the documentation if I get some hints.

- [x] Build is successful and all new and existing tests pass
- [x] Added [Unit Test(s)](https://liquibase.jira.com/wiki/spaces/LB/pages/1274937609/How+to+Write+Liquibase+Core+Unit+Tests)
- [ ] Documentation Updated


JosephCen:
I guess that key should be compare in method 'isDuplicate' in class ChangeLogParameter [here](https://github.com/liquibase/liquibase/pull/2656/files#diff-3a3e312df551992c399205e47b782d18119d917f09552e9b95ca8f0a9399b32bL355). 

Refer below screen-shot:
image

nvoxland:
Thanks for the PR!

In reviewing it, I took it over a bit to do a more extensive cleanup of the property handling code. We had been trying to just do some workaround patches in the property handling vs. the needed refactoring to avoid taking time to retest existing functionality.

But, your switch from the regexp-based searching to scanning makes me want to make sure we do a pass through the overall functionality already so adding in the other changes I'd been holding of on.

**Changes I made to your original PR functionality:**
- I renamed the missingPropertyMode options from LEGACY, EMPTY, THROW to PRESERVE, EMPTY, ERROR to be more descriptive and less java-y
- The "error" setting throws a ChangeLogParseException vs. an UnexpectedLiquibaseException

That seem ok to you @dwieland ?

**Beyond that, it was:**
- Adding missing javadocs/improving error messages
- Re-implemented the global vs. local parameters implementation in ChangeLogParameters, clearing out all the "TODO: This is ugly" comments that it resolved
- Created new `set()` vs `setLocal()` functions that made it more obvious what the passed DatabaseChangeLog was for and not
- Adjusted some class/function visibility based on how internal they should be
- Fixed issue with parameters being local by default in formatted sql when they are global everywhere else by default
- Cleaned up existing unit tests and added more

**Notes and test results:**

Things to be aware of
- Made some breaking API changes to classes/methods that were either previously broken or were ones that there was really no good reason for people to have been using. 
- I preserved API compatibility in any methods that seemed at all likely people would be using

Things to worry about
- While I did a round of testing myself and added more automated tests, the changelog parameter logic can be complex and this rewrites a good chunk of it. So I'd like another set of eyes on it to see if I missed anything

dwieland:
> That seem ok to you @dwieland ?

That's totally fine. Thanks for looking into it!

nvoxland (approved):

nvoxland:
The failing functional test is unrelated to this change, so moving this along to test.

Things to be aware of
- The functional tests have existing tests of property support. My initial changes failed one of these tests (a value which has a property value as well)

Things to worry about
- This ended up a major refactoring of our property support, so we need good testing of the property behavior in general to make sure nothing was lost in the change.

FBurguer:
For this PR, i checked if the new option `--missing-property-mode` with each param on hanaDB and MySQL, also I used  the SQL script done by @JosephCen on issue #2586 with the modification they commented there. 
**PRESERVE**: Using value ${myValue} here 
- [ ] HanaDb
- [ ] Mysql
`UpdateSQL` Output: `CREATE TABLE asyncevent (id BIGINT NOT NULL, messagebodybytes ${BYTESARRAY_TYPE} NULL, CONSTRAINT asyncevent_pkey PRIMARY KEY (id));`
**EMPTY**: Using value here
- [ ] HanaDb
- [ ] Mysql
`UpdateSQL` Output: `CREATE TABLE asyncevent (id BIGINT NOT NULL, messagebodybytes  NULL, CONSTRAINT asyncevent_pkey PRIMARY KEY (id));`
**ERROR**: Exits with error
- [ ] HanaDb
- [ ] Mysql
`UpdateSQL` Output: `Unexpected error running Liquibase: Could not resolve expression `${bytesarray_type}` in file changelog.xml`
It looks like liquibase with the option `--missing-property-mode`  has the right behavior listed above.

Test Environment:
OS: Windows 10
Java 11

kataggart:
@nvoxland can you please also update the docs? Thanks!

minasgull:
Thanks for this enhancement. I run into its absence a few days ago and started looking for workarounds. But now i just wait till next release. Thank you a lot

PR #2667, opened 2022-03-22
Run build & test against JDK 18

nvoxland:
## Description

Java 18 is now available, so adding it as a build & test target to ensure everything works.

Keeping Java 17 because that is an LTS version


suryaaki2 (approved):

XDelphiGrl:
* Change extends Java versions against which liquibase/liquibase builds and tests.
* No tests required.
* All unit and integration tests passing (which, yay! Liquibase works with Java 18!)
* Changes meet requirements to bypass manual QA testing.

PR #2679, opened 2022-03-24
simplify license checks and move code into Pro repo (DAT-9602)

StevenMassaro:
- Simplify the license checks that occur in the core code
- Remove ability to store licenses in any other format (like a file) other than a base64 string
- Move some of the pro license specific code into the pro repo


liquibase-core/src/main/java/liquibase/license/LicenseServiceUtils.java
```diff
@@ -13,34 +13,17 @@
...
-                "go to https://www.liquibase.org/download";
-      validationErrors.addError(message);
-      return validationErrors;
-    }
-
   /**
    * Check for a Liquibase Pro License.
    * @return true if licensed, false if not
    */
   public static boolean isProLicenseValid() {
```
suryaaki2, line 20: Should this method be renamed to isLicenseValid()?

suryaaki2 (changes requested):

liquibase-core/src/main/java/liquibase/license/LicenseServiceUtils.java
StevenMassaro: No, it still checks to see if there is a valid pro license. If the installed license is an empire license, this will return true.

wwillard7800 (approved):

liquibase-core/src/main/java/liquibase/license/Location.java
```diff
@@ -14,82 +14,37 @@
 public class Location {
```
nvoxland, line 14: Are we able to get rid of this entire Location class?

nvoxland (approved):
It seems good beyond my question of "Do we still need Location?"

liquibase-core/src/main/java/liquibase/license/Location.java
StevenMassaro: I think we'll still need the location class, because the license service requires knowing both the value of the license, and metadata about where the license came from. I did simplify the license class a bit.

PR #2685, opened 2022-03-25
Release Logic Bugfixes

nvoxland:
## Description

Fixes a couple issues with the release logic:
- Cannot use macos for re-version.sh since it has a case-sensitive filesystem
- Do not include `,1)` in the OSGI version info since it is invalid. Just have `version="$version" `


ap-liquibase (approved):
Looks good. Thanks!

PR #2686, opened 2022-03-25
Introduce generateStatementsFromRows for extensions

fbiville:
This will allow extensions to override just the statement
generation while relying on the existing CSV parsing logic.

This is needed by the Neo4j extension.


liquibase-core/src/main/java/liquibase/change/core/LoadDataChange.java
```diff
@@ -884,6 +829,90 @@ public String getSerializedObjectNamespace() {
...
+                    if (insertStatement instanceof InsertOrUpdateStatement) {
+                        ((InsertOrUpdateStatement) insertStatement).setAllowColumnUpdate(columnName, column.getAllowUpdate() == null || column.getAllowUpdate());
+                    }
+                }
+
+                statements.add(insertStatement);
+            }
+        }
+        if (rows.stream().anyMatch(LoadDataRowConfig::needsPreparedStatement)) {
+            // If we have only prepared statements and the database supports batching, let's roll
```
fbiville, line 869: I think the refactoring exhibits a bit more clearly a bug that was already there.

"any prepared statements" gets conflated with "only prepared statements". As a result, regular statements will be skipped as soon as 1 prepared statement is created.

fbiville:
Note: all existing tests for this class pass locally and I added an extra one.

nvoxland (approved):
The changes look good. I moved the LoadDataRowConfig inner class definition to the bottom since that's where they tend to be in the liquibase code, and made `databaseSupportsBatchUpdates()` into just `supportsBatchUpdates()`

Otherwise, I see how you just moved around the existing code and appreciate the extra unit test.

suryaaki2 (approved):

kataggart:
@nvoxland do we need to update any docs here?

nvoxland:
No we don't need additional docs. This change will be captured in the generated javadocs well enough

XDelphiGrl (approved):
* Change impacts only those who override generateStatementsFromRows from a Liquibase extension.
* New test added validating the override.
* Code refactored to be more readable (thanks, @fbiville !!).
* No further testing required.

APPROVED

[Passing Harness Tests](https://github.com/liquibase/liquibase-test-harness/actions/runs/2322045949)
[Passing Functional Tests](https://github.com/liquibase/liquibase-pro-tests/actions/runs/2322118577)

PR #2715, opened 2022-04-04
Quality Checks for Databases (DAT-9298)

StevenMassaro:


liquibase-core/src/main/java/liquibase/command/AbstractCliWrapperCommandStep.java
```diff
@@ -35,6 +36,19 @@ public final void run(CommandResultsBuilder resultsBuilder) throws Exception {
         }
         resultsBuilder.addResult("statusCode", statusCode);
 
+/*
```
nvoxland, line 39: This is just turns-out-unneeded code that should be removed before merging, right?

liquibase-core/src/main/java/liquibase/command/AbstractDatabaseCommandStep.java
```diff
@@ -0,0 +1,92 @@
...
+import liquibase.resource.ResourceAccessor;
+
+import java.util.ResourceBundle;
+
+import static java.util.ResourceBundle.getBundle;
+
+/**
+ * Convenience base class for {@link CommandStep} implementations which need a Database object.
+ */
+public abstract class AbstractDatabaseCommandStep extends AbstractCommandStep {
```
nvoxland, line 18: Using inheritance like this to manage the database/connection in commands is not the architectural direction we are trying to move to. The plan is to leverage the "pipeline" structure we built into the command framework. 

So while this is helpful in the short term, I don't like adding it to our public API for everyone to use. 

At a minimum, it would probably be good to mark it as "deprecated, this class is going to be removed in a future version" so 3rd party people can know to avoid it. We may even want to make it package protected or even moved out of liquibase-core to avoid the temptation?

liquibase-core/src/main/java/liquibase/command/core/SnapshotCommandStep.java
```diff
@@ -41,21 +62,135 @@ public class SnapshotCommandStep extends AbstractCliWrapperCommandStep {
...
     }
 
     @Override
-    protected String[] collectArguments(CommandScope commandScope) throws CommandExecutionException {
-        return collectArguments(commandScope, Arrays.asList(SNAPSHOT_FORMAT_ARG.getName()), null);
+    public void adjustCommandDefinition(CommandDefinition commandDefinition) {
+        commandDefinition.setShortDescription("Capture the current state of the database");
+    }
+
+    public static CatalogAndSchema[] parseSchemas(Database database, String... schemas) {
```
nvoxland, line 82: Is this a copy of InternalSnapshotCommandStep.parseSchemas? I didn't see it removed from there so it doesn't seem to be a move?

A static utility function like this doesn't really belong on SnapshotCommandStep. It doesn't really belong on InternalSnapshotCommandStep, but since that is "internal" it is less of a stable API we are creating. 

I'd suggest either just leaving it on InternalSnapshotCommandStep until we have a cleaner spot for it, moving this to a utility class (maybe SqlUtil or JdbcUtil?), or making it a protected non-static method for just this class if it's all that's using it.

liquibase-core/src/main/java/liquibase/structure/DatabaseObject.java
```diff
@@ -92,5 +92,17 @@ public interface DatabaseObject extends Comparable, LiquibaseSerializable {
      */
     DatabaseObject setAttribute(String attribute, Object value);
 
+    /**
+     * Given a database object, build its fully qualified name by looking up the at its containing objects.
+     * @return the fully qualified name
+     */
+    default String buildFullyQualifiedName() {
```
nvoxland, line 99: This is probably better named `getFullyQualifiedName` ?

nvoxland: I'm also not sure this method should exist. The object model is database agnostic, but the qualified name is often times going to vary depending on the database, and having this simply return a dot-separated string will cause problems as people try to use it in SQL statements or something and it starts failing when the object names need to be quoted etc.

Maybe this should be moved to be a method on Database? And maybe the return type should be a String[] so needed logic like quoting etc. can be applied as needed? Or maybe even there should be a FullyQualifiedName return type so callers can understand what those types are since quoting etc. rules will depend on that? And the generic "just dot them together" could be the toString() method on FullyQualifiedName?

liquibase-core/src/main/java/liquibase/util/TableOutput.java
```diff
@@ -19,7 +20,7 @@ public class TableOutput {
      * @param writer                   Writer to use for output
      *
      */
-    public static void formatOutput(List<List<String>> table, int[] maxWidths, boolean leftJustifiedRows, Writer writer) throws LiquibaseException {
+    public static void formatOutput(List<List<String>> table, List<Integer> maxWidths, boolean leftJustifiedRows, Writer writer) throws LiquibaseException {
```
nvoxland, line 23: This (and the below version) is an API breaking change. Is there a compelling reason for that?

liquibase-maven-plugin/src/main/java/org/liquibase/maven/plugins/AbstractLiquibaseMojo.java
```diff
@@ -312,20 +314,36 @@ public abstract class AbstractLiquibaseMojo extends AbstractMojo {
...
      */
     @PropertyElement
-    protected String liquibaseProLicenseKey;
+    @Deprecated
+    private String liquibaseProLicenseKey;
+
+    /**
+     * Specifies your Liquibase license key.
+     *
+     * @parameter property="liquibase.liquibaseLicenseKey"
```
nvoxland, line 330: This property should be `liquibase.licenseKey`. The pro one was made wrong but no need to continue that problem.

nvoxland (changes requested):

liquibase-core/src/main/java/liquibase/command/core/SnapshotCommandStep.java
wwillard7800: This method can just be private to this class.  It is not used anywhere else.

liquibase-core/src/main/java/liquibase/command/AbstractCliWrapperCommandStep.java
wwillard7800: Removed

liquibase-core/src/main/java/liquibase/command/AbstractDatabaseCommandStep.java
wwillard7800: I will remove the class.

liquibase-core/src/main/java/liquibase/util/TableOutput.java
StevenMassaro: There is not. I will overload this method so it's not a breaking change.

liquibase-core/src/main/java/liquibase/structure/DatabaseObject.java
StevenMassaro: Returning a `FullyQualifiedName` object that can be toStringed like you're describing. Is that your preferred solution to this, rather than returning a String[]?

suryaaki2 (approved):

nvoxland (changes requested):

liquibase-core/src/main/java/liquibase/command/CommandScope.java
```diff
@@ -136,7 +136,19 @@ public CommandScope setOutput(OutputStream outputStream) {
      */
     public CommandResults execute() throws CommandExecutionException {
         CommandResultsBuilder resultsBuilder = new CommandResultsBuilder(this, outputStream);
+        return executeWithResultsBuilder(resultsBuilder);
```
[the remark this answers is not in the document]
StevenMassaro: It's only usage is here. I'll remove this change.

nvoxland (approved):

PR #2725, opened 2022-04-05
The type of attributes "ordered" and "cycle" should be boolean

### LINKED ISSUE #2724: The type of attributes "ordered" and "cycle" should be boolean
## Environment

**Liquibase Version**: 4.9.1

## Description

In sequenceAttributes group contains ordered and cycled attributes with xsd:string, but correct values for this attributes is true or false 
example from documentation: https://docs.liquibase.com/change-types/alter-sequence.html

## Actual Behavior
cycled - xsd:string
ordered - xsd:string

## Expected/Desired Behavior
cycled - booleanExp
ordered - booleanExp

DartVerder:
## Description

Fixes #2724


nvoxland (approved):
Thanks for the fix!

I updated the PR to not change the 4.9.xsd. Changes only need to go into the `-next.xsd` and they will get included in the next xsd version.

I also took out the `default="false"` change, we don't include defaults in the XSD, so that extensions can override defaults in the code as needed.

Code review and test results:

Things to be aware of
- Only changes the XSD better limit what can be set in ordered/cycled to just true/false or expressions 
- Only impacts people who use the next xsd file, not existing xsd files.

Things to worry about
- Nothing

XDelphiGrl:
Added ReleaseMajor label because XSD changes must go into a major release.

XDelphiGrl (approved):
* Changes to XSD (in -next).
  * Fix limits the ordered and cycled sequence attributes to boolean values.

APPROVED

[Passing Harness Tests](https://github.com/liquibase/liquibase-test-harness/actions/runs/2330056052)
[Passing Functional Tests](https://github.com/liquibase/liquibase-pro-tests/actions/runs/2330054869)

PR #2730, opened 2022-04-06
Fix code coverage metrics

nvoxland:
## Description

Currently our code coverage metrics are incomplete because they are only capturing what gets covered in the "Sonar Scan" job, not in the matrix of runs across the different database types or OS or JVM versions. 

As a result, there is a lot of code which does get exercised by our tests, but will always show as "not covered".

This PR is trying to solve that by running with coverage info across all our different test runs, then in the Sonar job it aggregates that data together and reports on that.

### DIFF
--- a/.github/util/jacoco-aggregate.pom.xml
+++ b/.github/util/jacoco-aggregate.pom.xml
@@ -1,0 +1,14 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<project xmlns="http://maven.apache.org/POM/4.0.0" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
+         xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 http://maven.apache.org/xsd/maven-4.0.0.xsd">
+    <modelVersion>4.0.0</modelVersion>
+
+    <parent>
+        <groupId>org.liquibase</groupId>
+        <artifactId>liquibase</artifactId>
+        <version>0-SNAPSHOT</version>
+    </parent>
+
+    <artifactId>liquibase-aggregate</artifactId>
+
+</project>
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -306,8 +306,8 @@ jobs:
           rm -rf artifacts/*results-*macos-*
           rm -rf artifacts/*results-*windows-*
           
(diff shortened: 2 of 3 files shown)


nvoxland:
**PROBLEMS WITH CURRENT IMPLEMENTATION**

This is not capturing any coverage info, because:
```
Warning:  Classes in bundle 'liquibase-aggregate' do not match with execution data. For report generation the same class files must be used as at runtime.
Warning:  Execution data for class liquibase/sqlgenerator/core/AddColumnGeneratorSQLite$1 does not match.
Warning:  Execution data for class liquibase/serializer/core/string/StringSnapshotSerializerReadable$4 does not match.
Warning:  Execution data for class liquibase/serializer/core/string/StringSnapshotSerializerReadable$1 does not match.
Warning:  Execution data for class liquibase/serializer/core/string/StringSnapshotSerializerReadable$2 does not match.
Warning:  Execution data for class liquibase/serializer/core/string/StringSnapshotSerializerReadable does not match.
Warning:  Execution data for class liquibase/serializer/core/string/StringSnapshotSerializerReadable$3 does not match.
....
```

in Sonar Scan -> Sonar Analyze -> mvn jacoco:report.

Each run of the tests stores the "jacoco.exec" binary formatted file because that has the level of detail to later merge together. Like some lines of InsertDataGenerator are ran in the "unit" test runs, other lines are ran when oracle or mysql or whatever is enabled when the tests are ran. The final coverage needs to be the union of all that info. Jacoco:merge should be doing that correctly.

**_BUT: sonar can't read the exec format, only the XML report format. So we run jacoco:report on the merged exec. But jacoco:report needs the class files that were tested and EXACTLY the class files that were tested. If they're different (like compiled from source with a different JVM version or a different OS) it doesn't report any coverage._**

I tried to make it grab the pre-built code which **_should_** have been used for the different tests runs, but it's thinking there's a difference still. I deleted out the exec files that I expect to be different (non-ubuntu, non-java-8) but it's still not working.

Maybe there's a simple problem in my logic? I don't see it so am leave it alone for now and posting this as a draft.

My best suggestion is to circle back to this as part of the larger build refactoring project we're looking at -- better bake in that we build liquibase once and then test that jar and report on that jar. Never re-compile the production code again.

Issue #2281, opened 2021-12-17
loadData and loadUpdateData + relativeToChangelogFile fails when using a logicalFilePath


nvoxland:
## Description

Like #1277 but with loadData and loadUpdateData

smainz:
Doesn't #2073 fix this?
Maybe loadUpdateData is not covered.

nvoxland:
I thought it should have, and it seemed the same as your #2060 PR so tried to combine the two.

I'll do a bit more testing to see what is missing/wrong.

nvoxland:
I tried a variety of setups with 4.6.3 but was not able to reproduce issues.  I tried with both the CLI and with spring.

Could someone who is having problems give a sample directory structure and details on how they are running liquibase?

lalo-mx:
@nvoxland Example with the sqlFile  file error [sample-liquidbase-4.6.2.zip](https://github.com/liquibase/liquibase/files/7764067/sample-liquidbase-4.6.2.zip). The error is not happening with 3.8.6.
 I just realized that you mentioned 4.6.3. Where could I get that version?

joaoqalves:
I can reproduce this problem. I have both Liquibase in Spring and `liquibase-maven-plugin`. In the Spring one, it works without any problem. But when running `liquibase:status` on Maven it shows me it would need to run all the database migrations.

# Database CSV export

```csv
id,author,filename,dateexecuted,orderexecuted,exectype,md5sum,description,comments,tag,liquibase,contexts,labels,deployment_id
01-create-first-tables,joaoqalves,classpath:db/changelog/01-create-first-tables.xml,2022-01-04 16:25:56.878316,1,EXECUTED,8:7119a6ddd5a46c6747e3f1b823b09035,"createTable tableName=configuration; createTable tableName=shedlock; createTable tableName=user; addUniqueConstraint constraintName=uk_user_slug, tableName=user; createTable tableName=team; createIndex indexName=idx_team_marca_id, tableName=team; ...",,,3.10.2,,,1309956214
02-populate-init-data,operez,classpath:db/changelog/02-populate-init-data.xml,2022-01-04 16:25:58.371892,2,EXECUTED,8:3cff518d7668d8a48734e036277d2540,loadData tableName=user; loadData tableName=statistic_type; loadData tableName=week,,,3.10.2,,,1309956214
[...]

PR #2758, opened 2022-04-14
Spring Boot: better support relativeToChangelogfile when ResourceLoaders return FilteredReactiveWebContextResources

erik-meuwese-topicus:
## Pull Request Type

- [X] Bug fix (non-breaking change which fixes an issue.)

## Description

Subclasses of [FilteredReactiveWebContextResource ](https://github.com/spring-projects/spring-boot/blob/main/spring-boot-project/spring-boot/src/main/java/org/springframework/boot/web/reactive/context/FilteredReactiveWebContextResource.java) and potentially others, make `exists() always returns false in order to avoid exposing the whole classpath in a non-servlet environment.` which impacts Liquibase's logic in figuring out relative paths.

The change to SpringResourceAccessor to rely on exists() was introduced in Liquibase 4.0, so this will fix some changelogs that worked in 3.x but not any versions.

Improves spring support as mentioned in #2281 for at least some configurations. 

## Example Fixed Scenario

file:/...../target/classes/liquibase/init/test.xml
```
                   xsi:schemaLocation="http://www.liquibase.org/xml/ns/dbchangelog http://www.liquibase.org/xml/ns/dbchangelog/dbchangelog-3.5.xsd">
    <changeSet>
        <sqlFile path="test.sql" relativeToChangelogFile="true" splitStatements="false"/>
    </changeSet>
</databaseChangeLog>
```
file:/...../target/classes/liquibase/init/test.sql

Liquibase 4.9.1 fails to find test.sql relative to test.xml which contains the changelog

## Actual Behavior
```
org.springframework.beans.factory.BeanCreationException: Error creating bean with name 'liquibase' defined in class path resource [org/springframework/boot/autoconfigure/liquibase/LiquibaseAutoConfiguration$LiquibaseConfiguration.class]: Invocation of init method failed; nested exception is liquibase.exception.LiquibaseException: liquibase.exception.UnexpectedLiquibaseException: java.io.IOException: The file test.sql was not found in
    - file:/...../target/classes/
    - file:/...../target/classes/
Specifying files by absolute path was removed in Liquibase 4.0. Please use a relative path or add '/' to the classpath parameter.
```

## Expected/Desired Behavior
```
2022-04-14 16:06:45.705  INFO 36588 --- [  restartedMain] liquibase.database                       : Set default schema name to public
2022-04-14 16:06:45.910  INFO 36588 --- [  restartedMain] liquibase.lockservice                    : Successfully acquired change log lock
2022-04-14 16:06:46.688  INFO 36588 --- [  restartedMain] liquibase.changelog                      : Reading from public.databasechangelog
Running Changeset: liquibase/init/test.xml::test-003::tester
2022-04-14 16:06:46.930  INFO 36588 --- [  restartedMain] liquibase.changelog                      : SQL in file test.sql executed
2022-04-14 16:06:46.938  INFO 36588 --- [  restartedMain] liquibase.changelog                      : ChangeSet liquibase/init/test.xml::test-003::tester ran successfully in 80ms
2022-04-14 16:06:46.966  INFO 36588 --- [  restartedMain] liquibase.lockservice                    : Successfully released change log lock
```

### DIFF
--- a/liquibase-core/src/main/java/liquibase/integration/spring/SpringResourceAccessor.java
+++ b/liquibase-core/src/main/java/liquibase/integration/spring/SpringResourceAccessor.java
@@ -20,9 +20,11 @@ import java.util.TreeSet;
 public class SpringResourceAccessor extends AbstractResourceAccessor {
 
     private final ResourceLoader resourceLoader;
+    private final DefaultResourceLoader fallbackResourceLoader;
 
     public SpringResourceAccessor(ResourceLoader resourceLoader) {
         this.resourceLoader = resourceLoader;
+        this.fallbackResourceLoader = new DefaultResourceLoader(resourceLoader.getClassLoader());
     }
 
     @Override
@@ -103,7 +105,7 @@ public class SpringResourceAccessor extends AbstractResourceAccessor {
             if (url.contains("!")) {
                 return url.replaceFirst(".*!", "");
             } else {
-                while (!resourceLoader.getResource("classpath:" + url).exists()) {
+                while (!getResource("classpath:" + url).exists()) {
                     String newUrl = url.replaceFirst("^/?.*?/", "");
                     if (newUrl.equals(url)) {
                         throw new UnexpectedLiquibaseException("Could determine path for " + resource.getURL().toExternalForm());
@@ -133,7 +135,7 @@ public class SpringResourceAccessor extends AbstractResourceAccessor {
             relativeTo = relativeTo.replace("\\", "/");
 
             boolean relativeIsFile;
-            Resource rootResource = new DefaultResourceLoader(resourceLoader.getClassLoader()).getResource(relativeTo);
+            Resource rootResource = getResource(relativeTo);
             relativeIsFile = resourceIsFile(rootResource);
 
             if (relativeIsFile) {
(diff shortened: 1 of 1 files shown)


nvoxland:
@erik-meuwese-topicus thanks for tracking down and describing the `exists()` problem in https://github.com/liquibase/liquibase/issues/2281#issuecomment-1021036440 and making the PR for it.

I am not able to reproduce the problem to see your change fix it due to not getting spring set up in whatever configuration exposes it, but I understand how this can solve it.

I did push a change where I extracted the DefaultResourceLoader creation to the constructor and made a new `getResource(String)` method to wrap the logic you added to make it more extendable by people with other odd resource loader configurations. I also didn't want to completely abandon the regular resourceLoader in case it's able to find things not in the classpath. **Can you check that my changes didn't break your fix, @erik-meuwese-topicus ?**

I also updated the description to have it a bit more focused on the exists() problem vs. a more generic "fixes spring boot". Let me know if there is anything I got wrong in that change

nvoxland (approved):
Code review and test results:

Things to be aware of
- I couldn't reproduce the problem to see the fix work, but I was able to tell the change didn't break the setup I had
- Fix improves the handling within certain spring boot setups without impacting others by falling back to the old logic if the new lookup doesn't find a resource

Things to worry about
- No tests added, but I need to spend more time that we have right now to figure out the right way to integrate spring testing into our suite. Unit tests didn't seem like they'd add enough value to come up with.

XDelphiGrl (approved):
* Code change introduces an appropriate redundancy check for ensuring Liquibase loads the expected classpath.
* Change is limited to Spring Boot.

**Please Note!**  @nvoxland (Liquibase Founder) and several other Liquibase Engineers reviewed the code change and deem it low risk. However, there are no automated functional tests in place for Spring Boot. Be advised there is a larger possibility of unexpected side-effect due to the change than for similar changes with robust testing in place.

XDelphiGrl (changes requested):
I submitted a PR approval but just noticed that the build did not complete. Requesting that the build run and pass all developer-level tests prior to merging this fix.

suryaaki2 (approved):

erik-meuwese-topicus:
[nvoxland] this fixed our issue, thanks

PR #2763, opened 2022-04-14
Correctly handle null class package names in the logging system

### LINKED ISSUE #2633: Default logger does not handle null package names
## Description

From https://clojurians.slack.com/archives/C03S1KBA2/p1646949527191689

Class.getPackage() does not always return a value and can return null too, and when it does, it throws NPE here: https://github.com/liquibase/liquibase/blob/master/liquibase-core/src/main/java/liquibase/logging/core/JavaLogService.java#L45

Newer jdk versions have a getPackageName() method, but jdk 8 only has .getPackage() which could return null. https://docs.oracle.com/javase/8/docs/api/java/lang/Class.html#getPackage--

This likely impacts many non-java languages on the JVM besides Clojure

nvoxland:
## Description

The Class.getPackage() method can return null, especially in non-java JVM languages like clojure.

Updates the JavaLogService to use the class name as the log name if the passed class object returns a null package.

Fixes #2633

### DIFF
--- a/liquibase-core/src/main/java/liquibase/logging/core/JavaLogService.java
+++ b/liquibase-core/src/main/java/liquibase/logging/core/JavaLogService.java
@@ -42,8 +42,12 @@ public class JavaLogService extends AbstractLogService {
      * For example, all {@link liquibase.change.Change} classes will return a log name of "liquibase.change" no matter what class name or package name they have.
      */
     protected String getLogName(Class clazz) {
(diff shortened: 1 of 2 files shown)


suryaaki2 (approved):

XDelphiGrl:
* Fix addresses a null pointer.
* Unit test added to validate behavior.
* Change meets requirements to bypass manual QA testing.

PR #2771, opened 2022-04-19
Updated pull request template

nvoxland:
## Description

Make pull request template more focused on the PR as a code change vs. trying to re-collect information that belongs in the issue

### DIFF
--- a/.github/pull_request_template.md
+++ b/.github/pull_request_template.md
@@ -1,53 +1,33 @@
-<!--- This environment context section helps us quickly review your PR. 
-      Please take a minute to fill-out this information. -->
-## Environment
-
-**Liquibase Version**:
-
-**Liquibase Integration & Version**: <Pick one: CLI, maven, gradle, spring boot, servlet, etc.>
-
-**Liquibase Extension(s) & Version**: 
-
-**Database Vendor & Version**:
-
-**Operating System Type & Version**:
-
-## Pull Request Type
-
-<!--- What types of changes does your code introduce?
-      Put an `x` in all the boxes that apply: 
-      If this PR fixes an existing GH issue, edit the next line to add "closes #XXXX" to auto-link.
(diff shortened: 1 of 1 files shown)


nvoxland:
We had some good internal discussion on this. Summarizing it a bit here for further discussion and feedback:

I have a couple improvements I'm looking to make with this template change:

1. Better guide people to descriptions/discussions **_PRs_** around the proposed code changes vs. the descriptions/discussions in **_issues_** being around proposed end-user functionality changes. 
  a. The existing PR template is primarily designed around collecting information that belongs in the issue
  b. I've noticed that having so much duplicated "issue" information keeps people from including enough information about the code changes
2. Incorporating best-practice wording and section styles we've been starting to use into the default we suggest to everyone

The worries from the team were:
- It is requiring people to create issues for everything. And/or when people don't have an issue we won't get the repo information we need
  - I think that creating a separate issue **_is_** the policy we should be suggesting more, and a focus of this PR. 
  - But we still don't require separate issues. 
    - I expanded the `If there is NOT an existing issue, consider creating one.` section to be a bit more explicit on that
    - If we notice people creating stand-alone PRs without good repo steps, we'll iterate on it

nvoxland:
I added back the "Pull Request Type" section but renamed to "Impact". I thought that would still be good to keep at the top

suryaaki2 (approved):

kevin-atx (approved):
I agree that having the "steps to reproduce" in the Bug (issues) template makes more sense. 

This looks like a better template for PRs. Let's merge this and evaluate the new PRs that come through to make sure we're getting good enough information to more easily accept new code changes.

kataggart (approved):

XDelphiGrl (approved):
Everything looks good here; this is much less busy than the previous template. APPROVED.

PR #2772, opened 2022-04-19
Do not throw a confusing error if a datatype specified as a changelog parameter is not expanded

nvoxland:
## Description

If a user specifies something that looks like a changelog parameter but doesn't match a specified parameter, liquibase simply users it as-is. 

However, the DataType.fromDescription function that is used to determine the actual datatype to return cannot correctly handle a datatype that is formatted like a changelog parameter.

This fixes the datatype handling so it handles a datatype that looks like `${ ... }` as just a datatype it doesn't understand.

## Example Problem Changeset

```
<createTable tableName="test_table">
    <column name="id" type="${invalid}"/>
</createTable>
```

## Current behavior

Throws an `ArrayIndexOutOfBoundsException` exception

## Fixed behavior

Fails with an error like `Syntax error in SQL statement "CREATE TABLE test_table (id ${INVALID }` which allows the user to recognize that they did not correctly set the changelog parameter

### DIFF
--- a/liquibase-core/src/main/java/liquibase/datatype/DataTypeFactory.java
+++ b/liquibase-core/src/main/java/liquibase/datatype/DataTypeFactory.java
@@ -112,6 +112,10 @@ public class DataTypeFactory {
         if (dataTypeDefinition == null) {
             return null;
         }
+        if (dataTypeDefinition.matches("^\\$\\{.*}$")) {
+            return new UnknownType(dataTypeDefinition);
+        }
+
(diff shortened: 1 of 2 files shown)


suryaaki2 (approved):

XDelphiGrl:
* Fix is straightforward and improves user experience.
* New unit test added.

These changes have no impact on harness or functional tests.

XDelphiGrl (commented):
Approved.

XDelphiGrl (approved):

PR #2777, opened 2022-04-20
Remove custom log message filtering

### LINKED ISSUE #2756: Performance regression in liquibase versions greater than 4.3.5
## Environment
JDK 17, Mac OS as well as Alpine Docker
**Liquibase Version**:4.9.1

**Liquibase Integration & Version**: spring-boot 2.6.6 version.

**Database Vendor & Version**: Postgres

**Operating System Type & Version**: Mac OS and Docker(Azul alpine jdk 11 and 17)

## Description

I start from a clean postgresql database and run my change sets with liquibase version specifically set to 4.3.5, it finishes in less than 40 seconds.
For versions above it including 4.9.1 it takes ~240 seconds to run the same changes against the same clean database.

## Steps To Reproduce

With both versions the spring boot jar is started with `java -jar jar-name.jar` command. The application has *no* other logic, it is only used for initializing the db.

I am attaching an html flame graph html file which profiles the version with 4.9.1. It looks to be taking a lot of time in the logger of all places with the later version.
[out7.html.zip](https://github.com/liquibase/liquibase/files/8485960/out7.html.zip)

nvoxland:
## Description

We had originally introduced code to do an additional check and filtering of log messages flowing through both our Loggers and the java.util.logger configured in the CLI. 

However, that introduces a heavy performance cost and reduces flexibility in what can be logged. It was never used for anything beyond hub API key filtering, and any obfuscation that needs to happen really belongs at the point where the message is created.

Fixes #2756

## Things to be aware of

- I left the existing methods and interfaces there but made them no-ops and marked as deprecated for anyone building extensions that rely on these functions/interfaces.
- The old global configuration setting still exists and so will not break people with it set, but is now marked as "not commonly used" and the description says it's deprecated.
- I went through all uses of the hub API key configuration value to ensure the value is obfuscated in any log messages. Only one message was impacted

## Things to worry about

- Is there any logging of the hub API key that got missed? We should do some final testing of log outputs to make sure none were.


wwillard7800 (approved):

suryaaki2 (approved):

FBurguer:
After checking the log file with Nathan we found that Liquibase does not obscure the API key for this specific operation:

`[2022-05-18 11:24:54] DETALLADO [sun.net.www.protocol.http.HttpURLConnection writeRequests] sun.net.www.MessageHeader@589b028e8 pairs: {POST /api/v1/organizations/099cdb7d-5f53-4695-bc33-22dece33fba0/projects/853197bf-81b6-4964-bee2-d16f54c35833/operations/d994f748-5593-4708-b8e3-35658baf1e97/change-events HTTP/1.1: null}{User-Agent: Liquibase DEV}{Authorization: Bearer *APIKEYHERE*}{Accept: application/json}{Content-Type: application/json}{Host: hub.liquibase.com}{Connection: keep-alive}{Content-Length: 3857}`

nvoxland:
I added a filter to the CLI launcher to not enable logging for `sun.net.www.protocol.http.HttpURLConnection` which will get rid of that

FBurguer:
There are no more `sun.net.www.protocol.http.HttpURLConnection` entries, so it looks there isnt any other line showing the API key in the log file.

PR #2778, opened 2022-04-20
Fixed SQLite not correctly snapshotting tables with underscores in the name

nvoxland:
## Description

The jdbc spec says calls to getTables() etc. take a "table pattern" meaning _'s and %'s count as wildcards. We introduced a change to ensure tables with _'s in them escape those wildcards out.

However, the sqlite jdbc driver (as of 3.34.0) has a hard coded `sql.append(" WHERE TABLE_NAME LIKE '").append(tblNamePattern).append("' ...`

and the sqlite like escape syntax is `LIKE 'pattern' ESCAPE '\'` which doesn't fit into that generated sql.

Therefore, while introducing this back in will make it so table names that only differ on a _ like `test_table` vs `testrtable` will not snapshot correctly, the snapshot is no longer completely broken

## Things to be aware of

- Testing is done via test-harness. Change types like addColumn needs to snapshot the database and this snapshot problem is what was breaking those

## Things to worry about

- Nothing beyond the lack of escaping of wildcards. Trying to change the snapshot algorithm to support client-side filtering is beyond the scope of this particular fix


suryaaki2 (approved):

XDelphiGrl:
@nvoxland, can you review the test failures in the harness execution and make sure none are related to the changes to snapshot? 

[Test Harness Results](https://github.com/liquibase/liquibase-test-harness/actions/runs/2270382484)

nvoxland:
Build is passing now

XDelphiGrl (approved):
* Fix addresses a bug in generating SQL for SQLLite tables with underscores in name.
* Fix tested by an existing harness test.
  * [Test](https://github.com/liquibase/liquibase-test-harness/blob/main/src/main/resources/liquibase/harness/change/changelogs/createTable.xml)
  * Test harness tests passing.

[Passing Test Harness Tests](https://github.com/liquibase/liquibase-test-harness/actions/runs/2296038692)
[Passing Functional Tests](https://github.com/liquibase/liquibase-pro-tests/actions/runs/2296036469)

APPROVED

PR #2785, opened 2022-04-22
Change value modifier to make StandardLockService extendable for future extension

Cliftonz:
This pull request changes the modifiers of standard lock service to protected for the case where only certain parts of the lock service need to be rewritten. 

An example is this pr in the Keyspace extension.
https://github.com/liquibase/liquibase-amazon-keyspaces/pull/8

### DIFF
--- a/liquibase-core/src/main/java/liquibase/lockservice/StandardLockService.java
+++ b/liquibase-core/src/main/java/liquibase/lockservice/StandardLockService.java
@@ -37,7 +37,7 @@ import java.util.*;
 import static java.util.ResourceBundle.getBundle;
 
 public class StandardLockService implements LockService {
-    private static final ResourceBundle coreBundle = getBundle("liquibase/i18n/liquibase-core");
+    protected static final ResourceBundle coreBundle = getBundle("liquibase/i18n/liquibase-core");
 
     protected Database database;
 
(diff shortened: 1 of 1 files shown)


nvoxland (approved):
The changes make sense. I held off a bit on it since we're planning some level of refactoring around StandardLockService at some point (including to make it more extensible) and making the internal fields protected will make us have to work harder to preserve them if/when we do the refactoring.

But, I'm not sure the timeline on that and so don't want to continue to block this on the off chance something changes then.

Cliftonz:
@nvoxland Thanks, I will work on getting this passed the CI and we can go from there.

Cliftonz:
@nvoxland  Also this PR needs the "SafeToBuild" tag for the CI.

Cliftonz:
@nvoxland It looks like the CI thinks that this pr is not safe to build...

Cliftonz:
@nvoxland  What do I need to do to make Liquibase pro stable?

XDelphiGrl:
> @nvoxland What do I need to do to make Liquibase pro stable? or is this even needed for this to merge in?

Hello, @Cliftonz! There is nothing you can do from your side to help with the stability of the Liquibase Pro build. I am kicking off the functional and test harness test suites manually; when those pass, I'll mark this as ready to merge. Thank you for the PR and for your patience!

PR #2790, opened 2022-04-22
Upgraded mssql driver to 10.2.0

nvoxland:
## Description

Upgraded SqlServer jdbc driver to the newest major version

## End User Impact

The driver changed the encryption default from "false" to "true" between the old 8.x and the new 10.x driver. That means that anyone who has a self-signed certificate in their database will need to add either `encrypt=false` OR `trustServerCertificate=true` OR add the server certificate to their java trusted certificate list.


suryaaki2 (approved):

kataggart:
https://techcommunity.microsoft.com/t5/sql-server-blog/jdbc-driver-10-2-for-sql-server-released/ba-p/3100754

PR #2795, opened 2022-04-25
Updated information in "licenses" directory

nvoxland:
## Description

Updated the information in the licenses directory to remove components we don't actually ship, and to add additional ones we now do.


kevin-at-datical (commented):
@nvoxland - thank you for including the new license files and updating the README.txt. @suryaaki / @suryaaki2 - could you also review this list and make sure it includes everything you need for Pro? 

I am working on tickets for the next sprint to split these lists apart into OSS and Pro to make it easier for people to understand which licenses go with which edition of Liquibase.

PR #2823, opened 2022-05-04
Fix #2817 Load XSD files under OSGI

### LINKED ISSUE #2817: Load XSD files under OSGI
A few months ago Liquibase changed default policy for loading XSD files (global option "secureParsing"). But there is a problem with  loading XSD files from jar archive under OSGI via existing solution. Probably a better way for this case would be trying to load resources via existing ResourceAccessor and in case that openStreams() returns empty InputStreamList we would try to load a resource directly via getEntry for liquibase-core bundle. 

Copy javadoc for Bundle.getEntry :

Returns a URL to the entry at the specified path in this bundle. This bundle's classloader is not used to search for the entry. Only the contents of this bundle are searched for the entry.

The specified path is always relative to the root of this bundle and may begin with "/". A path value of "/" indicates the root of this bundle. 

Environment : Liquibase 4.9.1, Java 17, Karaf 4.3.6

jherkel:
Environment: liquibase 4.9.1, Java 17, Karaf 4.3.6

### DIFF
--- a/liquibase-core/src/main/java/liquibase/parser/core/xml/LiquibaseEntityResolver.java
+++ b/liquibase-core/src/main/java/liquibase/parser/core/xml/LiquibaseEntityResolver.java
@@ -9,9 +9,13 @@ import liquibase.resource.ResourceAccessor;
 import org.xml.sax.InputSource;
 import org.xml.sax.SAXException;
 import org.xml.sax.ext.EntityResolver2;
-
 import java.io.IOException;
 import java.io.InputStream;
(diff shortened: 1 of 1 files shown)


nvoxland:
Hi @jherkel , Does the fix from #2830 fix this issue in OSGI as well? Or if there something extra about OSGi that keeps that from working still? If so, what?

jherkel:
Hi,

I will check the fix from #2817 and I will inform you.

Jakub

PR #2830, opened 2022-05-06
Fixed local XSD lookup

nvoxland:
## Pull Request Type

- [X] Bug fix (non-breaking change which fixes an issue.)

## Description

There is an issue in the XSD lookup logic, where it always pulls the XSD file over the network if the primary ResourceAccessor fails to find it and the "fallback" one does.

When we'd get to the fallback code, the `&& SECURE_PARSING)` would be false and make the `if` block fail regardless of whether it found resources or not.

The standard Liquibase integrations (CLI, Maven, etc.) set up the primary resource accessor so we never hit the problem code path, but custom setups could. The integration tests, for example, did hit the problem code.


nvoxland:
Notes and test results:

Things to be aware of
- With my internet broken, the integration tests failed without this fix and pass with it in
- Added "autocandidate" since it's not something reachable from the CLI, maven, etc.
- It's an isolated fix to fallback logic

Things to worry about
- Nothing

PR #2859, opened 2022-05-19
Added whitelist for CLI log channels to include

nvoxland:
## Impact

Marked as a breaking change because 3rd party (non-liquibase) logging will no longer be logged without also specifying the --log-channels flag.

## Description

While we can't control the log configuration in most environments liquibase runs in, we do control the log setup in the CLI. 

Liquibase does not log any sensitive information, but we have no control over what other libraries will log and want to make sure nothing unexpected gets logged in user environments. 

Therefore, when applying the "logLevel" setting in the CLI, we are switching to a whitelisting of channels to include. The whitelist defaults to only "liquibase", but it can be controlled via a new `liquibase.logChannel` setting. For example, `--log-channels=liquibase,com.microsoft.sqlserver.jdbc`. We also support `--log-channels=all`

This replaces the previous reflection-based blacklisting of `net.sun.www.protocol.http.HttpURLConnection` which caused a warning in Java 11 and an ignored error in 17

## Things to be aware of
- Impacts the logging to both stdout and the log file
- Only impacts in the CLI

## Things to worry about
- Nothing

## Additional Context

I was not able to control the log level of `net.sun.www.protocol.http.HttpURLConnection` with the setting. I thing their use of the internal PlatformLogger is getting in the way. If anyone ever needs that, we can look at options for that. I was able to see it work with `org.mariadb` after adding `?log=true` to the mariadb url

## Testing

**Setup:** Configure liquibase to point to a mssql database

- Running `liquibase update` with no logLevel should print no logs
- Running `liquibase --log-level=FINE update` should log only messages with `[liquibase ...]` as the channel. No `[com.microsoft...]` channels
- Running `liquibase --log-level=FINE --log-channels=liquibase,com.microsoft update` should log messages with both the liquibase and com.microsoft channels 
- Running the above update calls with and without --log-channels works as expected in Java 8 and 17
- Running the update with and without --log-channels but with --log-file should have the same messages present/missing in the file as was in the console.

### DIFF
--- a/liquibase-cli/src/main/java/liquibase/integration/commandline/LiquibaseCommandLine.java
+++ b/liquibase-cli/src/main/java/liquibase/integration/commandline/LiquibaseCommandLine.java
@@ -595,16 +595,13 @@ public class LiquibaseCommandLine {
     private void configureLogging(Level logLevel, File logFile) throws IOException {
         configuredLogLevel = logLevel;
 
-        System.setProperty("java.util.logging.SimpleFormatter.format", "[%1$tF %1$tT] %4$s [%2$s] %5$s%6$s%n");
-
-        java.util.logging.Logger liquibaseLogger = java.util.logging.Logger.getLogger("liquibase");
-
         final JavaLogService logService = (JavaLogService) Scope.getCurrentScope().get(Scope.Attr.logService, LogService.class);
+        java.util.logging.Logger liquibaseLogger = java.util.logging.Logger.getLogger("liquibase");
         logService.setParent(liquibaseLogger);
 
+        System.setProperty("java.util.logging.SimpleFormatter.format", "[%1$tF %1$tT] %4$s [%2$s] %5$s%6$s%n");
 
         java.util.logging.Logger rootLogger = java.util.logging.Logger.getLogger("");
-
         Level cliLogLevel = logLevel;
 
         if (logFile != null) {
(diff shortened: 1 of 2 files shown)


suryaaki2 (approved):

nvoxland:
In testing `--log-channels`, we had issues with the sqlserver driver not logging with `--log-channels=liquibase,com.microsoft` on java 8 on a system with a non-EN_US locale on windows.

It logged correctly on that system with `--log-channels=all` and it logged correctly on java 8 with an english locale.  
It logged correctly on the non-EN_US system with java 11 and 17

Our theory is that there is either a bug in java 8 or a bug in the mssql driver not handling the fact that java 8 (unlike 11 and 17) translates the log level. Either way, it's not something to change the implementation over given that 8 is old at this point and it does work with --log-channels=all.

If you run into this behavior and it causes problems for you, feel free to open a separate issue.

FBurguer:
For testing, i checked the points Nathan listed above:

- Running liquibase update with no logLevel should print no logs `PASS`
- Running liquibase --log-level=FINE update should log only messages with [liquibase ...] as the channel. No [com.microsoft...] channels `PASS`
- Running liquibase --log-level=FINE --log-channels=liquibase,com.microsoft update should log messages with both the liquibase and com.microsoft channels `PASS`
- Running the above update calls with and without --log-channels works as expected in Java 8 and 17 `PASS`
(Aparently, when your env its not in english, Java 8 level names are not in english and this cause some issues when choosing the channel, further investigation is needed)
- Running the update with and without --log-channels but with --log-file should have the same messages present/missing in the file as was in the console. `PASS`
-  Running the update with and with --log-channles=all, shows all the messages. `PASS`

Testing Environment: 
Windows 10
Java 8,11,17
mssql-2019

PR #2870, opened 2022-05-24
Include inherited labels in databasechangelog table

### LINKED ISSUE #2824: include labels attribute is not propagating labels
## Environment
XML

**Liquibase Version**:
4.9.1

CLI

Postgresql 

Windows 10

## Description

When using a master changeset and using `<include>` tag with `labels` attribute, the labels are not propagated to the included changesets' `databasechangelog` rows .

## steps to reproduce:
Using three changelog files, a master and two included files.
master.xml:
`<include labels="a" file="include_a.xml"/>`
`<include labels="b" file="include_b.xml"/>`
include_a.xml:
`<changeSet ... />`
include_b.xml:
`<changeSet labels="x" ... />`

sql: `delete * from databasechangelog;`
cli: `liquibase changelog-sync`
sql: `select * from databasechangelog order by orderexecuted;` 

## actual behaviour observed in  `databasechangelog`:
changeset rows from include_a have labels ""
changeset rows from include_b have labels "x"

## expected behaviour:
changeset rows from include_a should have labels "a"
changeset rows from include_b should have labels "b, x"

Related issue: https://github.com/liquibase/liquibase/issues/1813

### LINKED ISSUE #2883: Context from databaseChangeLog is not added to DATABASECHANGELOG
## Environment
any (onboarding stage)

**Liquibase Version**:
4.10 and 4.11

CLI

n/a

MS SQL 2019 1(5.0.4153.1)

Windows Server 2016/Windows 10

## Description

We want to have a single master change log xml that has <includeAll path="some_folder"/> (so it includes all the change log files from a particular folder, each file has databaseChangeLog element with a context="release-xyz"). Here's an example of a changelog file from that folder (note the use of context). When we run Liquibase with --contexts="Release-2022.3,patch", the file with the correct context is selected but only the change that had patch will have context="Release-2022.3 AND patch" while the rest of the change will have contexts=null. 

<databaseChangeLog context="Release-2022.3"
		http://www.liquibase.org/xml/ns/dbchangelog/dbchangelog-4.11.xsd
		http://www.liquibase.org/xml/ns/dbchangelog-ext http://www.liquibase.org/xml/ns/dbchangelog/dbchangelog-ext.xsd">

A clear and concise description of the issue being addressed.
- 
The default context from databaseChangeLog is not persisted to DATABASECHANGELOG unless changeSet specifies its own context (so then both the default and the particular context is saves to the table). 

## Steps To Reproduce
Using CLI:

liquibase update --contexts="XYZ-2022.3,patch"

Master changeset: 

<databaseChangeLog changeLogId="master-change-log"
		http://www.liquibase.org/xml/ns/dbchangelog/dbchangelog-4.11.xsd
		http://www.liquibase.org/xml/ns/dbchangelog-ext http://www.liquibase.org/xml/ns/dbchangelog/dbchangelog-ext.xsd">

				<includeAll path="changelogs"/>	
</databaseChangeLog>

Individual change log file:

<databaseChangeLog context="XYZ-2022.3"
		http://www.liquibase.org/xml/ns/dbchangelog/dbchangelog-4.11.xsd
		http://www.liquibase.org/xml/ns/dbchangelog-ext http://www.liquibase.org/xml/ns/dbchangelog/dbchangelog-ext.xsd">

	
		
	<changeSet id="XYZ-9866" author="xyzz.zz" runOnChange="false">
		<comment>initial table creation, the default context wont show up in the DATABASECHANGELOG table</comment>
		<sqlFile 
			encoding="utf8"
			path="sql/XYZ-9866_create db.sql"
			relativeToChangelogFile="false"
			splitStatements="true"
			stripComments="true"
		/>		
	</changeSet>

	<changeSet id="zzzz-9867-patch-example" author="zzz.zz" runAlways="false" context="patch" >
		<comment>MPL-2022.3.1 patch this will actually get the default context and the specified context</comment>			
		<sqlFile 
			encoding="utf8"
			path="sql/zzz-9867.sql"
			relativeToChangelogFile="false"
			splitStatements="true"
			stripComments="true"
		/>		
	</changeSet>
	
	
</databaseChangeLog>

## Actual Behavior

Only "zzzz-9867-patch-example" change will have contexts="XYZ-2022.3 AND patch", the other change will have contexts=null but contexts=XYZ-2022.3 should be written to DATABASECHANGELOG table.

nvoxland:
## Impact

## Description

Changes MarkChangeSetRanGenerator so that if a changeSet was included with an include/includeAll containing a `labels` attribute, those labels are included in the `labels` column of `databasechangelog`. Makes the behavior match what the `contexts` attribute was generally doing before.

Fixes #2824
Fixes #2883 (please test)

However, when looking at the current behavior, I found the following wrong logic and made the following changes:
- The contexts from the include are only logged **_if_** changeSet has contexts as well
  - Now the include's context is always tracked
  - The tracked context is always null vs. sometimes an empty string
  - New logic is only applied to new changests going forward, we don't update existing values in the databasechangelog table
- The include/includeAll functions on DatabaseChangeLog took a `LabelExpression` and not a `Labels` like it should. In the changelog file, we define just comma separated lists of labels, not a label expression.
  - The java objects now store Labels not LabelExpression objects
  - I added deprecated versions of the DatabaseChangeLog's `include()`, `includeAll()`, and `setIncludeLabels()` functions to match the old versions that took a LabelExpression and converted them into Labels.
  - I did NOT add deprecated version of `DatabaseChangeLog.getIncludedLabels()` or `ChangeSet.getInheriteableLabels()` since java cannot overload return values. I don't think those should be methods generally used by extensions and so it is not worth switching to less useful names to preserve the API compatibility.

## Additional Context

I'm adding the ReleaseMajor label due to the changes in logic and APIs listed in the description. They should not be serious breaks and will likely not impact anyone, but something for users to be aware of.


suryaaki2 (approved):

ecarneiro01:
I tested the issues linked to this PR and both of them are fixed in this build. I also checked that both of them could be applied simultaneously:

- Check label values are propagating correctly to changelog files included in a main changelog.
	- `include` tag. **PASS**
- Check context values are propagating correctly to changelog files included in a main changelog.
	- `includeAll` tag. **PASS**
	- `include` tag. **PASS**

[2870.zip](https://github.com/liquibase/liquibase/files/8856128/2870.zip)

PR #2873, opened 2022-05-24
Added connector-api.jar to internal/lib

nvoxland:
## Description

The jaybird.jar we ship requires access to `javax.resources` which comes with the connector-api.jar.

This adds that required dependency and updates the licenses file accordingly.

## Current behavior

Running `liquibase update` against firebird with an out-of-the-box liquibase configuration fails with a "ClassNotFound" exception

## Updated Behavior

Running `liquibase update` against firebird with an out-of-the-box liquibase configuration runs successfully

### DIFF
--- a/liquibase-dist/pom.xml
+++ b/liquibase-dist/pom.xml
@@ -114,6 +114,13 @@
             <optional>true</optional>
         </dependency>
 
+        <!-- required for firebird driver -->
+        <dependency>
+            <groupId>javax.resource</groupId>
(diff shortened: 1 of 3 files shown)


FBurguer:
`liquibase update` runs succesfully with an out-of-the-box liquibase configuration. I tried running liquibase without the `connector-api.jar` and it fails with a` javax/resource/ResourceException` so it is definitly needed for firebird databases.

Testing Environment:
OS: Windows 10
Java 11
Firebird 3.0.7

PR #2885, opened 2022-05-26
Quote liquibase.jar path in bash launcher

nvoxland:
## Description

The current `liquibase` bash script does not quote the liquibase.jar path, so it will not run if there is a space in the path to LIQUIBASE_HOME. It fails with a `Error: Unable to access jarfile` error.

This quotes the path so it works correctly

## Repo Steps

1. Untar liquibase.tar.gz to a path containing a space in the filename
2. Run `./liquibase --version`


suryaaki2 (approved):

ecarneiro01:
I confirmed that the command `--version` can be executed even if there are spaces in the LIQUIBASE_HOME path.

image

PR #2898, opened 2022-05-31
Make additional query for more column information for MariaDB DAT-8693

wwillard7800:
Fixed issue where the `DatabaseMetadata.getColumns()` query for MariaDB does not return enough information to allow us to correctly generate SQL for `DATETIME(<precision>)` columns.


suryaaki2 (approved):

wwillard7800:
The Xpand flavor of MariaDB did not work with `NOW()`.  It did work with `CURRENT_TIMESTAMP()`.  Mike and I arrived at that same conclusion independently.  Because `CURRENT_TIMESTAMP()` worked for regular MariaDB in my testing, I felt like it was a safe change to make for both kinds.

nvoxland (changes requested):
@wwillard7800 figured out this isn't solving the actual problem.

nvoxland (approved):
Looks good

PR #2905, opened 2022-06-02
Upgraded maven dependency to 3.8.5

nvoxland:
## Description

Updates the version of maven we compile against to 3.8.5 to avoid security scanners flagging the project.

Because we're not changing any code this does not impact versions of maven we support, but going forward we will have to be careful of new maven API method calls we make until we can update the test automation to test against older versions

### DIFF
--- a/liquibase-extension-examples/pom.xml
+++ b/liquibase-extension-examples/pom.xml
@@ -12,10 +12,6 @@
 	<packaging>jar</packaging>
 	<description>Liquibase Examples</description>
 
-	<properties>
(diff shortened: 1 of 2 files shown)


suryaaki2 (approved):

XDelphiGrl (approved):
* Maven target version upgrade impacts build and test, not the released code.
* Unit and integration tests passing with the newer Maven version (3.8.5).

[Maven 3.8.5 Release Notes](https://maven.apache.org/docs/3.8.5/release-notes.html)

XDelphiGrl (approved):

PR #2914, opened 2022-06-03
Improved postgresql prepared statement performance

nvoxland:
## Description

Addresses two performance issues:
1. There are commonly used `matches()` calls made which rebuild the regexp pattern each time. It's not an excessive amount of time, but I was seeing perhaps 5% of our run time in that. 
  a. Converted them to compile and store the pattern once, then re-use those
2. Running prepared statements on postgresql is surprisingly slow (details below)
  a. Updated the code to reuse prepared statements
  b. Found a flag that can be passed to a postgesql-specific method that helped.

## Prepared Statement Performance Detailed Findings

In general, Liquibase does not use prepared statements. We generate the complete sql so that update vs. update-sql are doing exactly the same thing and just use the regular Statement class. But in loadData and loadUpdateData we sometimes need prepared statements for blobs etc. and so will auto-use them as needed and also allow the user to specify using them with the usePreparedStatements attribute.

For the regular insert statements we do, prepared statements SHOULD be as fast or faster. But postgresql is **_significantly_** slower. You can see it just by changing the usePreparedStatements value on a simple CSV with just strings/numbers.

The current code wasn't re-using the prepared statement objects like one normally should because of how our flow is structured. I added a vaguely ugly caching of those objects in hopes that re-using them correctly would meet what postgresql was expecting and help the performance. It helped, but didn't make a big difference. That is a better object use in general, though, so I left it in. It may help performance in some use cases and other databases even if my microbenchmark didn't see a big change.

What made the biggest difference was doing the execute with the `QUERY_ONESHOT` flag. Postgresql must be doing some expensive optimization in preparation for things being called a bunch of times and that is the cause of the slowness? 

That flag can only be set with a pg-specific `executeWithFlags()` call, though, so I had to call that via reflection since the driver classes aren't in the compile classpath. That change only applies to postgresql-related databases and we fall back to the regular execute if the reflection fails.

## Things to be aware of
- Only impacts postgresql, cockroach, edb, etc. 

## Things to worry about
- Does other versions of postgresql have different performance characteristics where QUERY_ONESHOT hurts more than helps? 
  - Does databases based on postgresql like edb have different performance characteristics too? Based on my understanding of what ONE_SHOT is saying (this statement will not be reparsed a bunch of times down the road) it is a valid hint for all the databases.

### DIFF
--- a/liquibase-core/src/main/java/liquibase/database/AbstractJdbcDatabase.java
+++ b/liquibase-core/src/main/java/liquibase/database/AbstractJdbcDatabase.java
@@ -10,7 +10,6 @@ import liquibase.changelog.DatabaseChangeLog;
 import liquibase.changelog.RanChangeSet;
 import liquibase.changelog.StandardChangeLogHistoryService;
 import liquibase.GlobalConfiguration;
-import liquibase.configuration.ConfigurationDefinition;
 import liquibase.configuration.ConfiguredValue;
 import liquibase.database.core.OracleDatabase;
 import liquibase.database.core.PostgresDatabase;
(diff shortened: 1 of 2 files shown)


nvoxland:
Code Review and test results:

Things to be aware of
- Difficult to see differences beyond watching the code path taken
- Didn't find a good way to add unit tests given how the code is structured, but existing automated tests should catch any unexpected breaking behavior.

Things to worry about
- Nothing

XDelphiGrl:
@nvoxland and @kataggart , this PR includes a change to the XSD. Should there be a label to indicate this?

rimuln:
Hi I got an issued with this PR. I created comment about it on Liquibase Forum: https://forum.liquibase.org/t/classcastexception-when-usepreparestatement-true-for-loaddata/8890/2

PR #2917, opened 2022-06-05
Added support for new "resourceRoots" global configuration

nvoxland:
## Impact

## Description

Adds new resourceRoots configuration

### DIFF
--- a/liquibase-cli/src/main/java/liquibase/integration/commandline/LiquibaseCommandLine.java
+++ b/liquibase-cli/src/main/java/liquibase/integration/commandline/LiquibaseCommandLine.java
@@ -13,14 +13,12 @@ import liquibase.configuration.LiquibaseConfiguration;
 import liquibase.configuration.core.DefaultsFileValueProvider;
 import liquibase.exception.CommandLineParsingException;
 import liquibase.exception.CommandValidationException;
-import liquibase.exception.UnexpectedLiquibaseException;
 import liquibase.hub.HubConfiguration;
 import liquibase.license.LicenseService;
 import liquibase.license.LicenseServiceFactory;
 import liquibase.logging.LogService;
 import liquibase.logging.core.JavaLogService;
-import liquibase.resource.CompositeResourceAccessor;
-import liquibase.resource.FileSystemResourceAccessor;
+import liquibase.resource.*;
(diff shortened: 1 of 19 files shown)


wwillard7800 (approved):
Looks good.  I think maybe a short note somewhere here to point out the extension points that need to be implemented would be nice.

FBurguer:
For this PR, we tested on the 3 different integrations.
Apache Ant:
For Ant, every task that supports searchpath work as expected, if the task doesnt support searchpath and it was added as an attribute, you get a error msg saying that the task doesnt support searchpath. List of tasks that support search path:
-update
-futureRollbackSQL
-rollback
-markAsRun
-changeLogSync
-changeLogSyncToTag

Test Environment:
OS: Windwos 10
MySQL Docker Container v8.0.27
Ant version 1.10.12

FBurguer:
For CLI, all seems to work as expected. Tried changing the searchpath to new locations where I have my changelog, and there is no notable issue.

ecarneiro01:
I did a couple more checks in the CLI and Maven integration and they worked just fine:

Maven:
	- Adding the `searchPath` property in the POM file replaces the paths that were set in the `classpath` when looking for resources (I checked with changelogFiles and included files in a changelog). **PASS**
	- Adding the property in the CLI replaces the paths that were set in the `classpath` when looking for resources. **PASS**
	- Multiple paths can be added in the `searchPath` value. **PASS**
	- Not using the `searchPath` property sets it's value to be the same as the `classpath`. **PASS**
	- Files in the `searchPath` paths are found properly (and error messages are shown for duplicates). **PASS** 
 
CLI:
	- Adding the `searchPath` property in the properties file replaces the paths that were set in the `classpath` when looking for resources. **PASS**
	- Adding the property in the CLI replaces the paths that were set in the `classpath` when looking for resources. **PASS**
	- Multiple paths can be added in the `searchPath` value. **PASS**
	- Not using the `searchPath` property sets it's value to be the same as the `classpath`. **PASS**
	- When using both `searchPath` and `classpath` the paths in the later are ignored when looking for resources. **PASS**
	- Files in the `searchPath` paths are found properly (and error messages are shown for duplicates). **PASS**

PR #2932, opened 2022-06-09
Cleanup relative paths preventing sql files being found by the class loader

marcsowen:
Cleanup relative paths of type "classpath*:/org/camunda/bpm/engine/db/liquibase/../upgrade/mysql_engine_7.16_to_7.17.sql" which is otherwise not being found in JAR by class loader.

## Impact

## Description

While using Camunda BPM engine, the database could not be upgraded to the latest schema version because the upgrade SQL file could not be found in class path. The reason for this is that the class loader will compare the path "org/camunda/bpm/engine/db/liquibase/../upgrade/mysql_engine_7.16_to_7.17.sql" to the ZipEntry of the corresponding JAR file via a hash function. The hashes differ because of the "liquibase/.." in between the class path string.

This patch will cleanup such relative paths resulting in path searchPath like "classpath*:/org/camunda/bpm/engine/db/upgrade/mysql_engine_7.16_to_7.17.sql" which can be found by the class loader.

## Things to be aware of

- cleanPath might do some of the tasks done manually in finalizeSearchPath and could therefore be removed.

## Things to worry about

- cleanPath seems to work with the "classpath*:/" prefix but nonetheless I'm not sure what other types of search paths may occur and break things.

## Additional Context

The problem is also described [here](https://forum.camunda.io/t/camunda-update-with-camunda-provided-liquibase-scripts/37401) and [here](https://jira.camunda.com/browse/CAM-14629).


nvoxland (approved):
I think the changes make sense. It's always hard to know the impact of the path normalization logic spring might be doing, but between it and the path normalization we're doing in liquibase I think it will be good.

I added a unit test for your case

PR #2936, opened 2022-06-10
Exclude beanutils from dependency list

nvoxland:
## Impact

## Description

We do not actually use the beanutils code paths from opencsv, so we should exclude it

## Things to be aware of

- Matches what we ship with the CLI

## Things to worry about

- Nothing


suryaaki2 (approved):

FBurguer:
For this pr i compared the dependencies between this build and an older one. Looks like the only dependency that has been excluded is beanutils. 
This is `mvn dependency:tree` output.
Left is the older build and the right one is this PRs build
imagen_2022-06-10_142229862

PR #2938, opened 2022-06-13
hide CommandFailedException stacktrace when thrown to set exit code (DAT-9608)

StevenMassaro:
## Impact

## Description

Stacktraces will not be printed to the logs when a CommandFailedException is set to "hidden".

## Things to worry about

`isHidden` is perhaps not the best name we can come up with.

### DIFF
--- a/liquibase-core/src/main/java/liquibase/command/CommandFailedException.java
+++ b/liquibase-core/src/main/java/liquibase/command/CommandFailedException.java
@@ -3,11 +3,13 @@ package liquibase.command;
 public class CommandFailedException extends Exception {
     private final CommandResults results;
     private final int exitCode;
+    private final boolean hidden;
 
-    public CommandFailedException(CommandResults results, int exitCode, String message) {
+    public CommandFailedException(CommandResults results, int exitCode, String message, boolean hidden) {
         super(message);
         this.results = results;
         this.exitCode = exitCode;
+        this.hidden = hidden;
     }
 
     public CommandResults getResults() {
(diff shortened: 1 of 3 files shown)


suryaaki2 (approved):

liquibase-core/src/main/java/liquibase/command/CommandFailedException.java
```diff
@@ -3,11 +3,13 @@
 public class CommandFailedException extends Exception {
     private final CommandResults results;
     private final int exitCode;
+    private final boolean hidden;
```
nvoxland, line 6: Yeah, "hidden" isn't the best name since the error isn't hidden just how stacktrace gets logged.

Maybe:
- logStacktrace? (be explicit, but not 'future-proof' )
- unexpected? (more generic, default is true, only log stacktrace (or other future things) if unexpected). 
  - Would `expected` be nicer to avoid the double negative? Or `normal` etc.?
- ??

So far, I'd maybe lean to something more generic like `unexpected` and then have the javadoc something like: "CommandFailedException is thrown any time a command did not succeed. If it did not succeed due to normal and expected reasons, mark it as unexpected=false. If the CommandFailedException is marked as unexpected=true, the code calling the command may want to do additional logging or handling of the exception because it knows the command was surprised by the result."

nvoxland (changes requested):

liquibase-core/src/main/java/liquibase/command/CommandFailedException.java
StevenMassaro: I'm going to go with `expected`, since the double negative of `unexpected` is a bit confusing.

PR #2942, opened 2022-06-14
Add additional info in CLI's --version output

nvoxland:
## Impact

## Description

Currently `--version` output lists `internal\lib\connector-api.jar: connector-api UNKNOWN` because the connector-api.jar only defines a "specification version". This updates the message creation to use the specification version if nothing else is available as a version.

Also adds a "By ..." for the vendor information if it is present. It gets verbose for some like `internal\lib\hsqldb.jar: HSQLDB 2.5.2 By The HSQL Development Group` but others like `internal\lib\ojdbc8.jar: JDBC 21.5.0.0.0 By Oracle Corporation` were not descriptive enough without it.

## Things to be aware of

- Only impacts CLI `--version` call

## Things to worry about

- Nothing


suryaaki2 (approved):

FBurguer:
All the shipped drivers that have vendor information have it now.

PR #2943, opened 2022-06-14
Fixed before/after/at column arguments in addColumn

nvoxland:
## Impact

## Description

The beforeColumn/afterColumn/position attributes on `addColumn` were not being well checked for database support, and even for databases that used them they were not actually being applied.

What I found:
- BEFORE is supported by h2 and hsql SQL only
- AFTER is supported by mysql and h2 SQL only
- POSITION is supported by no databases (firebird only supports position on alter columns)
- I checked oracle and mssql syntax and neither seemed to support any of those arguments. I didn't check other databases, but the existing code only talked about h2, hsql, mysql, and firebird

I don't think we should attempt to force support at this point for databases that don't natively support that syntax.

## Things to be aware of

- **Breaking change:** Because I fixed the validation logic, anyone who has a `beforeColumn` or `afterColumn` or `position` attribute on addColumn for a database that doesn't support that will now get a validation error vs. it just being ignored.
- Should check the change works correctly for mysql, h2, hsqldb, and firebird

## Things to worry about
- Nothing


wwillard7800 (approved):

FBurguer:
For this PR I checked the next scenarios:

-I you add the `beforeColumn` attribute  on `addColumn` on a hsql db creates the column in the position before the column specified in `beforeColumn` : `PASS`
- If you dont specify the attribute `beforeColumn` on `addColumn` on a hsql db, the column is created in the first position available:  `PASS`

-I you add the `beforeColumn` attribute  on `addColumn` on a h2 db creates the column in the position before the column specified in `beforeColumn` : `PASS`
- I you add the `afterColumn` attribute  on `addColumn` on h2 creates the column in the position after the column specified in `beforeColumn` : `PASS`
- If you dont specify the attributes `beforeColumn` or `afterColumn` on `addColumn` on a h2 db, the column is created in the first position available:  `PASS`

- I you add the `afterColumn` attribute  on `addColumn` on MySQL creates the column in the position after the column specified in `beforeColumn` : `PASS`
- If you dont specify the attribute `afterColumn` on `addColumn` on MySQL db, the column is created in the first position available:  `PASS`

-If you use the atributte `position` on a Firebird db, liquibase throws the right error: `PASS`
Example of the error msg:
          addAtPosition is not allowed on firebird, *changelog file name*::*changeset id*::*authors name*

Test Environment:
OS: Windows 10
Java 8
Firebird version: 3.0.7
MySQL version: 8.0.27
hSQL version: 2.4.0
h2 version: 2.1.212

mjaggard:
@nvoxland I think your change has prevented me from adding a column in `FIRST` position on MySQL. Previously this appeared to be possible by using `beforeColumn` although I appreciate that this was likely not working. I think there should be a way to specify the first column as that is supported - maybe using `position=0`?

kataggart:
@mjaggard I created #3195 so your request doesn't get lost, as this PR is already merged. Please add any additional questions/feedback/details to #3195. Many thanks.

axiopisty:
I think this merge was a poor decision. It requires us to downgrade to an older version of liquibase because of the backwards incompatibility issue. It would have been nice to provide some sort of configuration option to disable treating this as an error and be able to maintain backwards compatibility. I think printing a warning would be fine.

MnM:
> I think this merge was a poor decision. It requires us to downgrade to an older version of liquibase because of the backwards incompatibility issue. It would have been nice to provide some sort of configuration option to disable treating this as an error and be able to maintain backwards compatibility. I think printing a warning would be fine.

Agreed.

So now I need to write separate migrations for databases that support addColumnBefore/After, and ones that don't? This breaks a lot of our existing migrations that worked fine on both MySQL (where the feature is supported) and MSSQL (where it isn't).

I can't simply remove the attribute because that changes the checksum. Breaking old migrations is definitely something to worry about!

axiopisty:
I still stand by my previous comment and still think this was a poor change. But as a follow up to my previous comment, I did find work around that allowed me to move forward without downgrading. I don't like this approach, but it did work for my situation.

The work around I used is as follows:

Mutated my failing changeSet by:
    1.  Added the `runOnChange="true"` attribute to the `<changeSet/>` tag.
    2. Removed the attribute from the column definition(s) causing the validation error. In my case this meant removing the `afterColumn="<column_name>"` attribute from the `<column/>` tag.
    3. Added the following `<preConditions/>` to the mutated `<changeSet/>`

```xml
<preCondition onFail="MARK_RAN">
    <not>
        <columnExists tableName="<table_name>" columnName="<column_name>" />
    </not>
</preCondition>
```

I don't like this solution simply because it violates the principle of never mutating change sets that have already been applied to a database. However, it did serve as a work around to allow me to upgrade the liquibase version. Here is how it works:

Adding the runOnChange attribute bypasses liquibases checksum validation because it enables the change set to be executed each time the changeset it modified (going against the best practice of not mutating change sets). Adding the precondition allowed the changeset to be applied conditionally. 

If the change set was to be run against a brand new database where the change set HAD NOT previously been applied, then change set will run as currently defined (because the precondition will be satisfied) which will add the new column without specifying that the column should exist in some particular order. 

If the change set was to be run against an existing database where the change set HAD previously been applied, then the change set will NOT run as currently defined (because the precondition will NOT be satisfied). This is exactly what we want in this situation because the column already exists and should not be modified. The `onFail="MARK_RAN"` will effectively just allow liquibase to continue it's flow.

I hope this helps anyone else stumbling on this ticket.

If this comment helps you in your situation, please thumbs up it. :)

MnM:
@axiopisty 

An easier solution might be to:
1. remove the attribute that fails validation
2. add this to the beginning of the changeset, to skip it if it has been execution already:

        <validCheckSum>ANY</validCheckSum>

axiopisty:
@MnM 

I considered using that approach, but doing so would allow future modifications to the same change set. I decided to use the other approach that I described above, which worked nicely.

PR #2947, opened 2022-06-15
Refactored code to move methods to ChangeSet class so that they are more utilitarian DAT-10283

wwillard7800:
## Impact

## Description

Moved private methods from MarkRanChangeSetGenerator to ChangeSet so that they can be used other places.  These methods are available to return labels and contexts for a ChangeSet which include included values and the values on the ChangeSet itself.


StevenMassaro (approved):

suryaaki2 (approved):

liquibase-core/src/main/java/liquibase/changelog/ChangeSet.java
```diff
@@ -45,6 +44,12 @@ public class ChangeSet implements Conditional, ChangeLogChild {
      */
     private CheckSum storedCheckSum;
 
+    public static final String AND = " AND ";
```
nvoxland, line 47: I think these should be private, since they're just used in here and any global use of those sort of strings should belong somethwere other than ChangeSet

nvoxland (changes requested):

liquibase-core/src/main/java/liquibase/changelog/ChangeSet.java
wwillard7800: Got it

nvoxland (approved):

PR #2971, opened 2022-06-17
Introduced "contextFilter" and "labelFilter"  replacement settings

nvoxland:
## Impact

## Description

Contexts vs. Labels is a common confusion. I think part of the problem is that the names of the attributes are just "labels" and "context(s)" on both the runtime arguments and the changelog parameters so it's non-obvious how one side of each is a "filter" expression and which side is the filter is the difference.

This PR:
- Switches all the changelog file use of "context" to "contextFilter"
  - changeSet attribute
  - include attribute
  - includeAll attribute
  - modifySql attribute
  - databaseChangeLog attribute
  - In XML, Yaml, JSON, and formatted SQL changelog formats
- Switches the use of "labels" to "labelFilter" as an argument to:
  - CLI
  - Maven
  - Ant
  - Servlet
  - Spring
- Preserves old context/labels functionality for backwards compatibility
  - Introduces "alias" support to CommandArgumentDefinition 
  - Keeps old get/set methods, but marked as deprecated and calling the new get/set methods
- Testing:
  - Updates unit and integration tests to use contextFilter
  - Adds additional unit and integration tests for context/labels compatibility

## Things to be aware of
- Lots of files got changed, but it was primarily through automated refactoring of the get/setLabels and get/setContexts functions and/or straightforward search/replace
  - Existing automated tests caught the couple places I missed. The functionality worked for me on the first try
- Did not update any test-harness or functional tests changelogs, so they are currently testing the backwards compatibility
- CLI --help just lists labelFilter, not labels anymore. 
  - No need to remind people of the old and confusing argument

## Things to worry about
- Are there usages of contexts or labels I missed?


wwillard7800 (approved):
Looks good

kataggart:
@nvoxland I finally read this and completely understand! Does your change make this backwards compatible? So if I have changesets that still have `context` instead of 'contextFilter' they will still work? Similarly, if I use `labels` instead of 'labelFilter' in my CLI argument it will still work?

kataggart:
@nvoxland IGNORE ME! I can see this here now:

Preserves old context/labels functionality for backwards compatibility

    Introduces "alias" support to CommandArgumentDefinition
    Keeps old get/set methods, but marked as deprecated and calling the new get/set methods

I'd like to make sure we test that both the legacy and new alias values work as part of the testing here. Thanks.

suryaaki2 (approved):

PR #2972, opened 2022-06-18
Include should fail if a parser cannot be found for the specified file

nvoxland:
## Impact

## Description

Currently, include fails for files that cannot be found, but files which cannot be parsed as a changelog are silently ignored.
This makes Liquibase fail saying the file cannot be parsed (regardless of whether the file exists or not)

## Things to be aware of

- Error was introduced from re-use of the logic in includeAll where we skip files in an includeAll'ed directory which can't be parsed
- IncludeAll continues to fail if the given directory does not exist, but only warns on files within that directory that cannot be parsed

## Things to worry about

- Nothing

## Additional Context

DAT-10591

### DIFF
--- a/liquibase-core/src/main/java/liquibase/changelog/DatabaseChangeLog.java
+++ b/liquibase-core/src/main/java/liquibase/changelog/DatabaseChangeLog.java
@@ -375,7 +375,7 @@ public class DatabaseChangeLog implements Comparable<DatabaseChangeLog>, Conditi
                             includeContexts,
                             labels,
                             ignore,
-                            true);
+                            OnUnknownFileFormat.FAIL);
                 } catch (LiquibaseException e) {
                     throw new SetupException(e);
                 }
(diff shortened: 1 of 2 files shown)


XDelphiGrl (approved):
* Fix warns when a file included in the root changelog does not have a format recognized by a Liquibase parser, continuing on with the Liquibase operation.
  * If an entire directory included in the root changelog does not exist, Liquibase throws an error, halting Liquibase operation.
* New tests added for both the WARN and the FAIL conditions.

APPROVED

[Passing Functional Tests](https://github.com/liquibase/liquibase-pro-tests/actions/runs/2519553646)
[Passing Test Harness Execution](https://github.com/liquibase/liquibase-test-harness/actions/runs/2519554768)

* Fix updates the deprecated method include.

liquibase-core/src/main/java/liquibase/changelog/DatabaseChangeLog.java
```diff
@@ -744,4 +761,10 @@ public void addNestedPrecondition(Precondition precondition) {
         }
     }
 
+    public enum OnUnknownFileFormat {
+        SKIP,
+        WARN,
+        FAIL
+    }
```
StevenMassaro, line 768: Some docs here would be nice

StevenMassaro (approved):

liquibase-core/src/main/java/liquibase/changelog/DatabaseChangeLog.java
nvoxland: Thanks, I added some docs

PR #2999, opened 2022-06-23
do not allow individual commands to close the top-level OutputStream (System.out)

StevenMassaro:
## Impact

## Description

This change prevents individual commands from closing the output stream they are provided, which is actually `System.out`. It is bad practice for code to close `System.out` particularly because other command steps will still expect it to be open.

## Things to worry about

In very unusual circumstances, we might get more output than we did initially, because commands called after the RegisterChangelogCommandStep previously had a closed output stream. This isn't really a worry, and more of a positive thing.

### DIFF
--- a/liquibase-core/src/main/java/liquibase/command/CommandScope.java
+++ b/liquibase-core/src/main/java/liquibase/command/CommandScope.java
@@ -4,6 +4,7 @@ import liquibase.Scope;
 import liquibase.configuration.*;
 import liquibase.exception.CommandExecutionException;
 import liquibase.exception.CommandValidationException;
+import liquibase.io.UnclosableOutputStream;
 import liquibase.util.StringUtil;
 
 import java.io.OutputStream;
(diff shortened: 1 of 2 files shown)


liquibase-core/src/main/java/liquibase/command/CommandScope.java
```diff
@@ -40,7 +41,12 @@ public class CommandScope {
      * Creates a new scope for the given command.
      */
     public CommandScope(String... commandName) throws CommandExecutionException {
-        setOutput(System.out);
+        /*
+        This is an UncloseableOutputStream because we do not want individual command steps to inadvertently (or
+        intentionally) close the System.out OutputStream. Closing System.out renders it unusable for other command
+        steps which expect it to still be open.
+         */
+        setOutput(new UnclosableOutputStream(System.out));
```
nvoxland, line 49: I think the `new UnclosableOutputStream` call should be in the `setOutput()` function instead of here. No matter what people pass in as an output, it should be unclosable.

nvoxland (changes requested):

liquibase-core/src/main/java/liquibase/command/CommandScope.java
StevenMassaro: I can do that. That felt a bit more assertive to me, but I can also see why that would be useful, since we wouldn't want our code closing a stream that was given to us.

wwillard7800 (approved):

PR #3006, opened 2022-06-25
Added new liquibase.duplicateFileMode setting

nvoxland:
## Impact

## Description

Currently, when the same path is found in multiple locations liquibase fails with an error.

This PR adds a new `liquibase.duplicateFileMode` global setting which defaults to "ERROR" but can also be set to "WARN".

## Things to be aware of

- Existing behavior of "ERROR" logic is still the default.
- Refactored some usages of openStreams that duplicated the logic in openStream to just use openStream in order to take advantage of this new behavior.
- Searched the code for usages of openStream vs. openStreams and addressed all I found that should be updated

## Things to worry about

- Nothing


wwillard7800 (approved):

FBurguer:
Now when you `duplicateFileMode=WARN` you get msg explainng that liquibase found 2 changelogs with the same name but continues working with the "first" changelog it founds.
Both values work as expected and the default value is ERROR like in the description of the PR.

Test Environment:
OS: Windows 10
MySQL Docker Container v8.0.27
Java 8

PR #3009, opened 2022-06-27
Ensure Postgresql search_path entries are quoted correctly

### LINKED ISSUE #2928: redshift does not work with 4.11 release
see https://stackoverflow.com/a/72447000/8874837 and https://github.com/liquibase/liquibase/pull/2812#issuecomment-1148469330

nvoxland:
## Impact

## Description

In postgresql and it's varients (redshift, cockroachdb, etc), the search_path that gets set based on the defaultSchemaName uses the value returned from `show search_path`. Sometimes, objects such as `$user` in that shown search path are note quoted like is needed to re-set them. 

This PR updates the logic from handling only values that start with a `$` and only on cockroachdb to handling all non-quoted values on all postgresql versions. 

The change handles cases where a value comes back already quoted from `show search_path`. 

For example, if defaultSchemaName is being set to `test`,  it will `set search_path=test, "$user", "public"` whether the original `show seach_path` returned `"$user", public` or `$user, public`.

By quoting `"public"` like that, it will enforce the case on public, but in my testing the value in search_path was already correctly cased. If I called `set search_path=$user, PUBLIC"` then show search_path still returned `$user, public`. And if I called `set search_path=$user, "TEST", public` it showed as `"$user", "TEST", public`. So the change seems correct still.

Fixes #2928

## Things to be aware of
- Nothing beyond what is in description

## Things to worry about

- Nothing

### DIFF
--- a/liquibase-core/src/main/java/liquibase/database/core/DatabaseUtils.java
+++ b/liquibase-core/src/main/java/liquibase/database/core/DatabaseUtils.java
@@ -1,5 +1,6 @@
 package liquibase.database.core;
 
+import liquibase.GlobalConfiguration;
 import liquibase.Scope;
 import liquibase.database.Database;
 import liquibase.database.OfflineConnection;
@@ -8,6 +9,7 @@ import liquibase.executor.Executor;
 import liquibase.executor.ExecutorService;
 import liquibase.statement.core.RawSqlStatement;
 import liquibase.structure.core.Schema;
+import liquibase.util.StringUtil;
 
 public class DatabaseUtils {
     /**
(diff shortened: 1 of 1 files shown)


wwillard7800 (approved):

ecarneiro01:
I checked that the generated SQL of `update-sql` could be applied on the databases and that the object `$user` was being quoted correctly:

Postgres:
 - Use default values for the search path and apply changes to the DB. **PASS**
 - Use defaultSchemaName property and apply changes to the DB. **PASS**

Redshift:
 - Use default values for the search path and apply changes to the DB. **PASS**
 - Use defaultSchemaName property and apply changes to the DB. **PASS**
  
CockroachDB:
 - Use default values for the search path and apply changes to the DB. **PASS**
 - Use defaultSchemaName property and apply changes to the DB. **PASS**

fatso83:
This was a breaking change (#3205) for a lot of people, but was not marked as such in the release notes 😞

PR #3016, opened 2022-06-28
warn if XSD version does not match build version in validate command (DAT-9874)

StevenMassaro:
## Impact

## Description

Warn if specified XSD version is different than the version of the build when running the `validate` command.

## Things to be aware of

I made this change in a way that this check is "best effort". It should not ever cause a user to experience an uncaught exception.

## Things to worry about

It is possible that the regex I'm using is not robust enough to properly get the version out of the XSD URL. It is also possible that there is a better way to do this than using a regex.

### DIFF
--- a/liquibase-core/src/main/java/liquibase/Liquibase.java
+++ b/liquibase-core/src/main/java/liquibase/Liquibase.java
@@ -33,6 +33,7 @@ import liquibase.logging.core.BufferedLogService;
 import liquibase.logging.core.CompositeLogService;
 import liquibase.parser.ChangeLogParser;
 import liquibase.parser.ChangeLogParserFactory;
+import liquibase.parser.core.xml.XMLChangeLogSAXParser;
 import liquibase.resource.InputStreamList;
 import liquibase.resource.ResourceAccessor;
 import liquibase.serializer.ChangeLogSerializer;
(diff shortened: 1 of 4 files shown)


abrackx (approved):

nvoxland (changes requested):
I’m not sure this is a good idea?

People usually have a large set of “previous” change logs that have built up over the years. Like a v1.0.xml and v1.1.xml … v8.3.xml

All those old versions were working fine, including the xsd they are referencing. Ideally as you move on to new changelog files you shouldn't be updating those old files, but this change will bother them until they do.

I get that people can have issues if they don't notice that they are on an old xsd, but we are also now suggesting they don't use a versioned xsd and instead use -latest unless they have a good reason. Under that best practice, this warning makes even less sense because they have picked a specific version for a reason and so even less of a “warning “ ?

StevenMassaro:
> I’m not sure this is a good idea?
> 
> People usually have a large set of “previous” change logs that have built up over the years. Like a v1.0.xml and v1.1.xml … v8.3.xml
> 
> All those old versions were working fine, including the xsd they are referencing. Ideally as you move on to new changelog files you shouldn't be updating those old files, but this change will bother them until they do.
> 
> I get that people can have issues if they don't notice that they are on an old xsd, but we are also now suggesting they don't use a versioned xsd and instead use -latest unless they have a good reason. Under that best practice, this warning makes even less sense because they have picked a specific version for a reason and so even less of a “warning “ ?

We're going to display this message only when running the `validate` command, and display it at an info level.

PR #3033, opened 2022-07-01
Improve connection exception by including underlying driver exception trace

hessjcg:
## Impact

## Description

When there is an unexpected exception while connecting to the database, liquibase should populate the "causedBy" 
exception so that it is possible to troubleshoot connection problems. Without this, users are shooting
in the dark if the connection fails.

## Things to worry about

- Underlying JDBC drivers may leak private data in its exception. This is probably beyond the scope of the liquibase code.

## Additional Context

I'm submitting this PR because of problems that my customers have using liquibase with 
google cloud sql. Often the problems are related to configuration issues that prevent them
from connecting and are easily resolved. The configuration problems are explained in exception
messages thrown by the custom google cloud sql socket factory. 

However, when my customers Liquibase, these messages are hidden. It would be helpful
to let them see the full stack trace.


nvoxland (approved):
Makes sense. I think the "caused by" must have gotten missed there, there should be no reason to suppress that.

hessjcg:
@StevenMassaro thanks for the quick review. Would you mind adding the tag `SafeToBuild` so that the builds can run?
The builds are currently blocked because... "A member of Liquibase team needs to apply label 'SafeToBuild'! "

FBurguer:
Reproduced this PR on MySQL and looks like it works as excpected and showing the right complete stack trace

Test Environment
OS: Windows 10
MySQL container 8.0.27
Java 8

PR #3034, opened 2022-07-01
Keep runInTransaction="false" from being included in generated changelogs

nvoxland:
## Impact

## Description

PR #2585 fixed our lack of serializing runInTransaction into changelogs, but the diff logic was incorrectly setting generated changesets as runInTransaction="false" which used to be ignored and therefore defaulted back to "true" but no longer is.

## Things to be aware of

- This is fixing an unreleased bug
- In the current build from master, all changesets in a generateChangeLog will have `runInTransaction="false"`. With this fix, there will be no runInTransaction flag included in the changesets, like it is in 4.12.0

## Things to worry about

- Nothing


StevenMassaro (approved):

FBurguer:
Using `generateChangeLog` does no longer set `runInTransaction="false"`. Tested it on a h2 db and a mssql db with a build that had this bug and then with this pr build. 

Test Environment
OS:Windows 10
Java 8
Docker container mssql 2017
h2 2.1.212

PR #3045, opened 2022-07-06
Output "may loose settings" warning on mysql/mariadb for more impacted change types

nvoxland:
## Impact

## Description

MySQL/MariaDB's `alter table ... modify column` syntax requires a complete redefinition of the column when modifying a column, such as setting a comment, changing a datatype, etc.

The the setColumnRemarks, addDefaultValue, etc change does not support all possible column features, only the data type so anything else (like nullability etc) will be lost. 

This PR expands the existing warning we had for modifyDataType, since that is all we can do for this.

## Things to be aware of

- Only impacts Mysql/MariaDB. 
- Adds warnings to addDefaultValue, setColumnRemarks, addNotNull, and dropNotNull changes. 

## Things to worry about

- Nothing


suryaaki2 (approved):

FBurguer:
For this PR tested if every change type listed before had a warning msg on MySQL and MariaDB:
addDefaultValue 
setColumnRemarks 
addNotNull, and dropNotNull

PR #3047, opened 2022-07-07
Add support for `AS <dataType>` clause for H2 since version 2.0

marcus-nl:
## Impact

## Description

Currently Liquibase explicitly excludes H2 when emitting the `AS <dataType>` clause. However, H2 has support for this feature since version 2.0. This effect of this change is that the CreateSequenceGenerator will only omit the clause for versions before that.

## Things to be aware of

I also extended the existing unit test to verify the new behavior. In the process I also attempted to improve the code 
for better readability and consistency. For example, I used the generatorUnderTest field instead of creating a new CreateSequenceGenerator instance.

This also fixes the issue of an unsupported "dataType" on old h2 versions being ignored instead of failing validation like it should have. This will impact on old versions of h2 using dataType as an attribute.

## Things to worry about

n/a

## Additional Context

n/a

### DIFF
--- a/liquibase-core/src/main/java/liquibase/sqlgenerator/core/CreateSequenceGenerator.java
+++ b/liquibase-core/src/main/java/liquibase/sqlgenerator/core/CreateSequenceGenerator.java
@@ -38,16 +38,21 @@ public class CreateSequenceGenerator extends AbstractSqlGenerator<CreateSequence
             validationErrors.checkDisallowedField("maxValue", statement.getMaxValue(), database, FirebirdDatabase.class, H2Database.class, HsqlDatabase.class);
         }
 
-        if (isPostgreWithoutAsDatatypeSupport(database)) {
-            validationErrors.checkDisallowedField("AS", statement.getDataType(), database, PostgresDatabase.class);
-        }
-
         validationErrors.checkDisallowedField("ordered", statement.getOrdered(), database, HsqlDatabase.class, PostgresDatabase.class, MSSQLDatabase.class);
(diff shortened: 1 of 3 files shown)


nvoxland (approved):
Code review and test results:

Things to be aware of:
- change is isolated to H2 and makes sense to me
- includes a mock connection test

Things to worry about:
-

nvoxland:
On further review, while you were preserving the old "just ignore the unsupported dataType" logic from before, we are working to make any unsupported attributes be validation errors so it isn't surprising when requested changelog settings aren't applied. I also moved the similar postgresql check to the same area of the code, and added an integration test entry that uses the attribute to validate that the generated SQL is correct.

I pushed the changes up to your branch and updated my review notes above.

nvoxland:
Code review and test results:

Things to be aware of:
- Change are isolated to h2
- Adds validation check for invalid use of the field in h2 < 2.0 and uses it in h2 >= 2.0

Things to worry about:
- Nothing

suryaaki2 (approved):

XDelphiGrl (approved):
* Fix adds support for sql generation of the "AS" clause for H2 versions 2.0 or later.
  * If H2 is less than 2.0, a validation error is returned by Liquibase.
  * If H2 is 2.0 or greater, generated SQL includes `AS` clause, i.e., `CREATE SEQUENCE SCHEMA_NAME.SEQUENCE_NAME AS BIGINT`.
* New tests added validating behavior for H2 versions older than 2.0 and equal to 2.0.
* Passing test harness execution; functional tests N/A.

APPROVED

PR #3063, opened 2022-07-14
Improved "include" fix performance: optimized normalizePath method

### LINKED ISSUE #1856: Optimize updates
Based on some profiling data we did for a client, it seemed that Liquibase was recompiling a lot of often used regexes, e.g.:
image

The latest code also seems to contain lots of them, e.g. https://github.com/liquibase/liquibase/blob/master/liquibase-core/src/main/java/liquibase/changelog/DatabaseChangeLog.java#L645-L650, resulting in a horrible performance.

Extracting every regex on the critical path shouldn't be difficult. Creating fewer Strings may be more difficult, but would be worth it.
I'd gladly help, but I don't have the reproducer code available, our client does.

### LINKED ISSUE #2221: normalizePath super slow?
## Environment

**Liquibase Version**: 4.3.5
**Liquibase Integration & Version**: spring boot
**Database Vendor & Version**: mysql
**Operating System Type & Version**: linux

## Description

Super slow run after liquibase upgrade. It takes 5~10 minutes to run. In previous versiones it took < 30 seconds.

## Steps To Reproduce

I have a change set xml file with more than 500 xml includes, like this:

`<include file="another.xml" relativeToChangelogFile="true" />`

## Actual Behavior
Takes a lot of time to run command

## Expected/Desired Behavior
Similar run time than previous versions

I've run with profiler, and got this:

screenshot2021-11-17T11:51:46-03:00

liquibase.changelog.DatabaseChangeLog#normalizePath
this method takes a lot of time.
i suspect that precompiling the regexp will help.

## Additional Context
if it looks good, i can provide a patch.

lzxgyh:
## Impact

## Description

Replaces on-the-fly regular expression usage with pre-compiled versions.

Fixes #2221 
Fixes #1856

## Things to be aware of

- Nothing

## Things to worry about

- Nothing

### DIFF
--- a/liquibase-core/src/main/java/liquibase/changelog/DatabaseChangeLog.java
+++ b/liquibase-core/src/main/java/liquibase/changelog/DatabaseChangeLog.java
@@ -25,6 +25,8 @@ import liquibase.util.FilenameUtil;
 import java.io.IOException;
 import java.io.InputStream;
 import java.util.*;
+import java.util.regex.Matcher;
+import java.util.regex.Pattern;
 
 /**
  * Encapsulates the information stored in the change log XML file.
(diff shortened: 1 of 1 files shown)


nvoxland:
Thanks @lzxgyh!
I updated the description to remove the boilerplate template with what was done.

I worry about the creation of the matcher static variables, though. If this gets used in a multithreaded way, I think those will cause problems. If you just create the matchers as local variables instead of resetting the static variables does it loose a lot of the performance gains? The original "compile" call should be what makes the most difference, right?

lzxgyh:
> Thanks @lzxgyh! I updated the description to remove the boilerplate template with what was done.
> 
> I worry about the creation of the matcher static variables, though. If this gets used in a multithreaded way, I think those will cause problems. If you just create the matchers as local variables instead of resetting the static variables does it loose a lot of the performance gains? The original "compile" call should be what makes the most difference, right?

lzxgyh:
> Thanks @lzxgyh! I updated the description to remove the boilerplate template with what was done.
> 
> I worry about the creation of the matcher static variables, though. If this gets used in a multithreaded way, I think those will cause problems. If you just create the matchers as local variables instead of resetting the static variables does it loose a lot of the performance gains? The original "compile" call should be what makes the most difference, right?

Hi  [nvoxland](https://github.com/nvoxland)

I think you are right, in the past, I always thought that liquibase was a single-threaded tool

lzxgyh:
> Thanks @lzxgyh! I updated the description to remove the boilerplate template with what was done.
> 
> I worry about the creation of the matcher static variables, though. If this gets used in a multithreaded way, I think those will cause problems. If you just create the matchers as local variables instead of resetting the static variables does it loose a lot of the performance gains? The original "compile" call should be what makes the most difference, right?

Hi  [nvoxland](https://github.com/nvoxland)

I think you are right, in the past, I always thought that liquibase was a single-threaded tool

nvoxland:
Are you able to fix that up quick, @lzxgyh?

lzxgyh:
Sure, I fpushed a new commit to remove the matcher static variables , Please review @nvoxland

nvoxland (approved):
Code review and test results:

Things to be aware of:
- Just moves implementation of where the regular expressions are managed from one spot to another, no functionality change
- I did not re-test the performance improvements, but I can see from the performance dumps that this what a large problem in the code and the PR will fix it as the contributor says it does

Things to worry about:
- Nothing

liquibase-core/src/main/java/liquibase/changelog/DatabaseChangeLog.java
```diff
@@ -725,14 +733,12 @@ public static String normalizePath(String filePath) {
...
         }
-        return filePath.replaceFirst("^classpath:", "")
-                .replaceAll("\\\\", "/")
-                .replaceAll("//+", "/")
-                .replaceAll("/\\./", "/")
-                .replaceFirst("^[a-zA-Z]:", "")
-                .replaceFirst("^\\.?/", "")
-                ;
-
+        String noClassPathReplaced = NON_CLASSPATH_PATTERN.matcher(filePath).replaceFirst("");
```
XDelphiGrl, line 736: Thank you, @lzxgyh - This is a nice improvement to code readability!

XDelphiGrl (approved):
* Fix improves performance of normalizePath.
* Fix improves readability of REGEX matching in DatabaseChangelog.java

APPROVED

[Passing Functional Tests](https://github.com/liquibase/liquibase-test-harness/actions/runs/2702642521)
[Passing Test Harness Execution](https://github.com/liquibase/liquibase-test-harness/actions/runs/2702642521)

suryaaki2 (approved):

PR #3073, opened 2022-07-18
generate-changelog does not capture same-name triggers defined on different tables (DAT-10112)

abrackx:
## Impact

- [?] Breaking change (fix or feature that would cause existing functionality to change)
 
## Description

Removes use of getDbObjectComparator when finding missing objects.

## Things to be aware of

Removes what I believe is an unnecessary comparator and instead uses scoped comparators. 

## Things to worry about

This may cause other items that were previously "identical" to generate.

## Additional Context
N/A


StevenMassaro (commented):
I'm not going to approve this because I don't have any idea why this code was originally written. I'd wait for one of the other three reviewers to share their opinion here.

nvoxland (changes requested):
I don't think this is the way we can fix it. Removing the DatabaseObjectComparator logic in favor of just equals/hashcode and also breaking the sorting seems like it will likely have too many side effects. 

I think the needed approach is be to get the correct comparator for the known type using DatabaseObjectComparatorFactory. It looks like we don't have a public method for just getting the best comparator for a given type? Maybe need to add that?

PR #3074, opened 2022-07-18
error when JAVA_HOME is set to invalid value (DAT-10545)

StevenMassaro:
## Impact

## Description

Throw an error and fail to start Liquibase if the JAVA_HOME environment variable is set and the path it points to does not exist.

## Things to worry about

If there is an edge case we didn't think about or test for, this will prevent Liquibase from starting in those scenarios.


nvoxland (approved):
Looks good. Both bash and batch only trigger the "does the file exist" check if JAVA_HOME is set so it won't break people not setting JAVA_HOME and relying on their PATH

PR #3078, opened 2022-07-19
Improve CLI error messages

nvoxland:
## Impact

## Description

Currently the CLI attempts to figure out the "best" message to show from all the different "caused by" reasons in an uncaught exception and sends that to the UI as the error message. Example:

```
Unexpected error running Liquibase: liquibase.parser.core.ParsedNodeException: Unknown precondition 'invalid'. Check for spelling or capitalization errors and missing extensions such as liquibase-commercial.
```

It picks the message to show by finding the "lowest" caused by. But, there can be valuable information in a variety of the caused by sections, and just because an exception is lowest doesn't mean it's the most useful.

This changes the CLI message to include all the messages in the stacktrace. Example:
```
Unexpected error running Liquibase: Connection could not be created to jdbc:mysql://localhost:33063/lbcat with driver 
  - Caused by: Connection refused: connect
```

## Things to be aware of

- Improved the logic to remove the exception class names from the UI error message
- Added code to removed duplicate messages that sometimes appear in the exceptions
- Only impacts the CLI

## Things to worry about

- Nothing


suryaaki2 (approved):

FBurguer:
Cases like conecttion error and invalid precondition are more clear now and error msgs that dont have Caused by doesnt seem to be impacted.

PR #3079, opened 2022-07-19
when generating SelectFromDatabaseChangeLogLock statement, cast LOCKED column to integer for MariaDB (DAT-10798)

StevenMassaro:
## Impact

## Description

The newer versions of the MariaDB driver, which have been shipped with recent versions of Liquibase, return the `LOCKED` column as a `BitSet` by default. This adds casting to the select statement, so that the column is returned as an integer.

## Things to be aware of

I tested this with the driver that is shipped with older versions of Liquibase (which did not exhibit the `BitSet` issue) and this change is backwards compatible with those drivers.

## Things to worry about

Perhaps there was a better way to solve this problem that I'm not aware of.

I was unable to write an integration test for this. For some reason, the integration tests returned the `LOCKED` column as an integer, even before I added the casting.

### DIFF
--- a/liquibase-core/src/main/java/liquibase/sqlgenerator/core/SelectFromDatabaseChangeLogLockGenerator.java
+++ b/liquibase-core/src/main/java/liquibase/sqlgenerator/core/SelectFromDatabaseChangeLogLockGenerator.java
@@ -3,6 +3,7 @@ package liquibase.sqlgenerator.core;
 import liquibase.change.ColumnConfig;
 import liquibase.database.Database;
 import liquibase.database.ObjectQuotingStrategy;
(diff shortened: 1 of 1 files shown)


StevenMassaro:
I actually might have found a better way to solve this.

StevenMassaro:
Turns out that this was a bug in the MariaDB driver, which was fixed here: https://github.com/mariadb-corporation/mariadb-connector-j/commit/17e60c886a7ebf2d234e3c772a8c392b3a9a54c7

And that was incorporated into Liquibase here: https://github.com/liquibase/liquibase/pull/3021

I confirmed that the latest builds on master do not exhibit this problem.

PR #3080, opened 2022-07-19
DAT-10473 make package signing a scheduled step

ap-liquibase:
## Impact

## Description

## Things to be aware of

## Things to worry about

- List unknowns you have


jnewton03 (approved):
looks okay to me, but would like to get @nvoxland @StevenMassaro takes first.

.github/workflows/installer-build-check.yml
```diff
@@ -0,0 +1,243 @@
+name: Build and Test
+
+# Job will run nightly at 03:05 EDT / 02:05 CDT
+# Time below is UTC
+on:
+  schedule:
+    - cron: "5 7 * * *"
+  workflow_dispatch:
+  push:
```
nvoxland, line 9: I'd suggest not making a `gha-scheduled-*` branch pattern for this to trigger by. Instead just have the cron run and the manual run option. The manual run option would let people run it with whatever branches they want on demand, without having to know about the magical `gha-scheduled-` pattern. Plus, as we have more scheduled jobs, we dont want all of them firing just because we made a branch with a pattern.

.github/workflows/build.yml
```diff
@@ -364,8 +364,8 @@ jobs:
           ##create zip file from tar content
           (cd liquibase-dist/target/liquibase-0-SNAPSHOT && zip -r ../liquibase-0-SNAPSHOT.zip *)
 
-          ##create installers
-          (cd liquibase-dist && ${{ github.workspace }}/.github/util/package-install4j.sh 0-SNAPSHOT)
+          ##create installer - disabled here but run as nightly job and as part of release workflow
+          # (cd liquibase-dist && ${{ github.workspace }}/.github/util/package-install4j.sh 0-SNAPSHOT)
```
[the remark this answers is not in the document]
nvoxland: Are we wanting to completely avoid making installers in branches? Or do we only want to not sign them?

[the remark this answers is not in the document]
StevenMassaro: I don't have a preference.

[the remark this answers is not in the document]
ap-liquibase: Hi @StevenMassaro The` gha-scheduled-*` trigger was for my development/testing purposes only (similar to github-action-*). I'm planning to removed that from the workflow (after PR merge)

ap-liquibase:
Hi @nvoxland I turned off all installer making completely for simplicity's sake. I did consider an alternative option where I would add another command-line argument and additional logic to `package-install4j.sh` which would for the script to execute `INSTALL4J_ARGS="$INSTALL4J_ARGS --disable-signing`. I decided against that since I could accomplish Steve's and Wes' ask by just turning off all installer building and moving that into the scheduled job.

ap-liquibase:
This comment just in from @nvoxland:
"If we're good with removing the installer creation from the build process (I'm fine with that) I think your scheduled job shouldn't be all that stuff copied from the build.yml we aren't ever going to do anymore.
Instead, it should be a copy of the 'build-installers' part of create-release.yml instead. Just:
Replace "Restore Completed Artifacts" with the action-download-artifact call to get liquibase-artifacts and get the .tar.gz we need for the installers out of there
Replace the "Attach Files to Draft Release" with the upload-action
The way you are replicating the now-non-existant build process isn't going to be catching the things that this scheduled action will need to catch (problems in the create-release process before release day). Ideally we'll replace the then-duplicate logic in create-release.yml with a call to this action down the road even."

ap-liquibase:
@nvoxland Please re-review. Successful action action execution here: https://github.com/liquibase/liquibase/actions/runs/2778007917

.github/workflows/installer-build-check.yml
```diff
@@ -0,0 +1,79 @@
...
+          workflow: build.yml
+          workflow_conclusion: success
+          branch: master
+          name: liquibase-artifacts
+          path: re-version/final
+
+      - name: Set up JDK for GPG
+        uses: actions/setup-java@v3
+        with:
+          java-version: '8'
```
jnewton03, line 32: @ap-liquibase @nvoxland should this be java 8 or java 11?

jnewton03: also should we add the `cache` options that the setup-java action has?

nvoxland (approved):
Thanks for making the changes

.github/workflows/installer-build-check.yml
ap-liquibase: I think for builds were using java 8

PR #3093, opened 2022-07-21
Improve failure message of RowCountPrecondition to preserve expected row count

martinspielmann:
## Impact

## Description

The existing failure message for `RowCountPrecondition` will always tell _Table [NAME] is not empty._, even if there is an expected row count greater than zero.

E.g. you set `expectedRows` to 5 and your table has 10 rows, you'll get the following failure message, which is somehow misleading, because you do not expect to have an empty table at all:
_Table [NAME] is not empty._

The change updates the error message to be more specific and contain the `expectedRows`:
_Table [NAME] does not have the expected row count of 5._

## Things to worry about

- I'm afraid this very small change changes the method signature of `getFailureMessage`, so classes extending from `RowCountPrecondition` may in fact break if they did override this method.


nvoxland (approved):
Code review and test results:

Things to be aware of:
- Change makes sense to me
- No test, but I can't think of a useful test to add at this point
- I'm OK with the minor API change

Things to worry about:
- Nothing

FBurguer:
Precondition work as expected, it shows how many rows you have vs expected rows. If it doesnt match, changset its not executed/warns you (depends on how did you set up the precondition).

PR #3098, opened 2022-07-21
[CORE-3202] Fix NUMBER not compatible with H2

nick318:
## Impact

## Description

Fixes NUMBER not compatible with H2 in PostreSql compatibility mode (or suggested for h2 in general).
Full description is here - https://liquibase.jira.com/browse/CORE-3202

As H2 supports numeric, we can use it as a default type as for others DBs


nvoxland:
Code review and test results:

Things to be aware of:
- Code makes sense and is isolated to h2
- Test harness "diff" test is failing for what seems to be other reasons

Things to worry about:
- Nothing

XDelphiGrl (commented):
* PR fixes a long-standing H2 bug where explicitly defined 'numeric' datatype in changesets was replaced with 'number', which is unsupported on H2.
  * An example of a changeset that was generating invalid SQL and now produces valid SQL containing `NUMERIC` is:

```
- changeSet:
      changes:
      - addColumn:
          tableName: talent
          columns:
          - column:
              name: rate
              type: numeric(20, 2)
```

* New unit tests added for H2.

KushnirykOleh (approved):
Investigated test-harness build failure. It is failing due to this branch being outdated, bringing latest master changes should fix the issue. Or this failure can be ignored as we already know the reason.
I'll take care of test-harness part after that

XDelphiGrl (approved):

PR #3100, opened 2022-07-22
fix pro commands in maven plugin

StevenMassaro:
## Impact

## Description

The jar split work changed the liquibase-core dependency to contain only open source code. This PR readds the pro code dependency for the maven plugin.

## Additional Context

https://github.com/liquibase/liquibase/pull/3050


StevenMassaro:
This needs to be fixed, but I don't think this is the right way to fix it.

PR #3102, opened 2022-07-22
Upper case result columns only for case-insensitive databases

### LINKED ISSUE #3071: `liquibase.executor.jvm.ColumnMapRowMapper` does not consider `liquibase.database.AbstractJdbcDatabase#isCaseSensitive`
## Environment

**Liquibase Version**: 4.13.0

**Liquibase Integration & Version**: core

**Database Vendor & Version**: Neo4j

## Description

Despite overriding `liquibase.database.AbstractJdbcDatabase#isCaseSensitive` in the Neo4j extension, the columns of results coming from `liquibase.executor.Executor#queryForList(liquibase.statement.SqlStatement)` are still upper-cased.

The row mapping happens via `liquibase.executor.jvm.ColumnMapRowMapper` and this class [always upper-cases](https://github.com/liquibase/liquibase/blob/83faa2125953a4761e6a711cbf5818b267a9b3b9/liquibase-core/src/main/java/liquibase/executor/jvm/ColumnMapRowMapper.java#L57-L59) result column names, which is a very surprising behavior when fetching results from case-sensitive databases such as Neo4j.

## Expected/Desired Behavior
`liquibase.executor.jvm.JdbcExecutor` should propagate the case-sensitivity information to the mapper, so that it only upper-cases columns when the database is NOT case-sensitive.
Note: this is a breaking change, there might be better, backward compatible solutions.

fbiville:
## Impact

## Description

Updates the "ResultSet into a Map" logic to preserve the column casing for case sensitive databases instead of always uppercasing them.

Marked as a breaking change, because if any code is using queries that return non-uppercased column names but then is reading from the map as upper case it will no longer get the column's value. That old behavior was wrong and this fixes it, but there may be 3rd party code that relies on that buggy behavior.

Fixes #3071


nvoxland (approved):
Code review and test results:

Things to be aware of:
- Only changes behavior if the database says it supports mixed case identifiers
- Current behavior is wrong in that case and likely not working, and this fixes that
- Relies on the driver to be correct about whether it supports mixed case identifiers

Things to worry about:
- Nothing

nvoxland:
Thanks for doing the update @fbiville . That may be the cause of the failing tests from before, I'll keep an eye on the new build

suryaaki2 (approved):

nvoxland:
Our build process isn't correctly connecting some of the needed changes in our pro code to get a fully clean build with this.

We have plans to improve that build process to be able to work with a setup like this where the liquibase-side PR is from a fork, but we don't want to hold this PR up for that work. 

So since the changes **_should_** be good due to me running most of the automated tests locally, I'll merge the PRs to master at a non-busy time in order to get a complete build that gets correctly automatically tested and if there is any surprising failures we can react then by quicky fixing whatever problems are found then or by reverting the changes until we can address them.

fbiville:
@nvoxland would it help if the branch is pushed to the main repository instead of a fork?

nvoxland:
@fbiville I don't think it's work making a separate branch/PR for this in the main repo. We're working on improving the build process currently so hopefully it shouldn't be an issue going forward.

XDelphiGrl (approved):
This fix prevents Liquibase from snapping column names TO UPPER when the database platform is case sensitive. 

APPROVED

Note: We had some build issues that prevented us from running all functional tests in a branch. The plan is to merge this to liquibase/liquibase master & re-run the tests. @nvoxland and I will be watching to see if there are adverse effects due to the merge. If so, we will fix or revert (depending on severity).

PR #3103, opened 2022-07-22
allow passing any arguments to flow commands from maven plugin (DAT-10944)

StevenMassaro:
## Impact

## Description

Add a `flowCommandArguments` `Map` to the mojo definition for `flow` and `flow.validate` maven goals. The arguments will be passed to every command that is run in flow.

## Things to be aware of

- Users are responsible for using the correct format of command arguments.
- Is it okay to create the getter for `argumentValues` in `CommandScope`?

## Things to worry about

Future maven plugin refactoring might render this method obsolete.


liquibase-core/src/main/java/liquibase/command/CommandScope.java
```diff
@@ -89,6 +89,10 @@ public <T> CommandScope addArgumentValue(CommandArgumentDefinition<T> argument,
         return this;
     }
 
+    public SortedMap<String, Object> getArgumentValues() {
```
nvoxland, line 92: I don't think it's a good idea to expose this private variable. It's not something that we want directly accessible by external clients as it would expose some internal implementation details and lets them go around the correct way of doing things and/or get unexpected results thinking that getArgumentValues() is a list of all the configured arguments and/or the correct values for them. 

I think you need to change your addAll() call below to loop through the flowCommandArguments and call `addArgumentValue()` for each.

nvoxland (changes requested):

liquibase-core/src/main/java/liquibase/command/CommandScope.java
StevenMassaro: Nathan and I talked about this offline. He gave me some good ideas, and we basically landed on creating a new value provider that provides the values from the mojo, which will allow downstream usage in the pro code very easily.

nvoxland (approved):

suryaaki2 (approved):

PR #3121, opened 2022-07-28
Changed what sequence attributes are included in diff/generate changelog for Snowflake

yodzhubeiskyi:
## Impact

## Description

Snowflake only supports operating 'startWith' and 'incrementBy' sequence attributes. The rests are set by default and are not allowed to be set or changed. In order for Liquibase to correctly work with those 'fixed' attributes we need to:

- Throw a validation error when the 'fixed' attributes are present in changesets that are being applied to Snowflake
- Include the values in JSON snapshots so the database is accurately and completely represented BUT
- Exclude the 'fixed' attributes from any changesets generated by diffChangelog or generateChangelog so that we don't generate invalid changesets

### DIFF
--- a/liquibase-snowflake/src/main/java/liquibase/diff/output/changelog/core/ChangedSequenceChangeGeneratorSnowflake.java
+++ b/liquibase-snowflake/src/main/java/liquibase/diff/output/changelog/core/ChangedSequenceChangeGeneratorSnowflake.java
@@ -1,0 +1,47 @@
+package liquibase.diff.output.changelog.core;
+
+import liquibase.change.Change;
+import liquibase.change.core.AlterSequenceChange;
+import liquibase.database.Database;
+import liquibase.database.core.SnowflakeDatabase;
+import liquibase.diff.ObjectDifferences;
+import liquibase.diff.output.DiffOutputControl;
+import liquibase.diff.output.changelog.ChangeGeneratorChain;
+import liquibase.structure.DatabaseObject;
+import liquibase.structure.core.Sequence;
+
+import java.util.ArrayList;
+import java.util.List;
+
+public class ChangedSequenceChangeGeneratorSnowflake extends ChangedSequenceChangeGenerator{
+
+    @Override
+    public int getPriority(Class<? extends DatabaseObject> objectType, Database database) {
+        int priority = super.getPriority(objectType, database);
+        if ((Sequence.class.isAssignableFrom(objectType)) && (database instanceof SnowflakeDatabase)) {
+            priority += PRIORITY_DATABASE;
+        }
+        return priority;
+    }
+
+    @Override
+    public Change[] fixChanged(DatabaseObject changedObject, ObjectDifferences differences, DiffOutputControl control, Database referenceDatabase, Database comparisonDatabase, ChangeGeneratorChain chain) {
+        Sequence sequence = (Sequence) changedObject;
+
+        List<Change> changes = new ArrayList<>();
+        AlterSequenceChange accumulatedChange = createAlterSequenceChange(sequence, control);
+
+        if (differences.isDifferent("incrementBy")) {
+            AlterSequenceChange change = createAlterSequenceChange(sequence, control);
+            change.setIncrementBy(sequence.getIncrementBy());
+            accumulatedChange.setIncrementBy(sequence.getIncrementBy());
+            changes.add(change);
+        }
+
+        if (changes.isEmpty()) {
+            return null;
+        } else {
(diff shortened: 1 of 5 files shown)


nvoxland (approved):

liquibase-snowflake/src/main/java/liquibase/diff/output/changelog/core/MissingSequenceChangeGeneratorSnowflake.java
```diff
@@ -0,0 +1,38 @@
...
+        return priority;
+    }
+
+    @Override
+    public Change[] fixMissing(DatabaseObject missingObject, DiffOutputControl control, Database referenceDatabase, Database comparisonDatabase, ChangeGeneratorChain chain) {
+        Sequence sequence = (Sequence) missingObject;
+
+        CreateSequenceChange change = new CreateSequenceChange();
+        change.setSequenceName(sequence.getName());
+        if (control.getIncludeSchema()) {
```
KushnirykOleh, line 29: i would rather treat catalogs in DiffOutputControl like in parent class

       ``` if (control.getIncludeCatalog()) {
            change.setCatalogName(sequence.getSchema().getCatalogName());
        }```

KushnirykOleh (changes requested):
one ask for a change, other than that looks good

KushnirykOleh (approved):

liquibase-snowflake/src/main/java/liquibase/diff/output/changelog/core/ChangedSequenceChangeGeneratorSnowflake.java
```diff
@@ -0,0 +1,47 @@
...
+import liquibase.diff.ObjectDifferences;
+import liquibase.diff.output.DiffOutputControl;
+import liquibase.diff.output.changelog.ChangeGeneratorChain;
+import liquibase.structure.DatabaseObject;
+import liquibase.structure.core.Sequence;
+
+import java.util.ArrayList;
+import java.util.List;
+
+public class ChangedSequenceChangeGeneratorSnowflake extends ChangedSequenceChangeGenerator{
```
suryaaki2, line 16: Just wondering if the name of the class should be be SnowflakeChangedSequenceChangeGenerator? Can you please check if there is a pattern, if not am fine..

liquibase-snowflake/src/main/java/liquibase/diff/output/changelog/core/MissingSequenceChangeGeneratorSnowflake.java
```diff
@@ -0,0 +1,41 @@
...
+import liquibase.change.Change;
+import liquibase.change.core.CreateSequenceChange;
+import liquibase.database.Database;
+import liquibase.database.core.SnowflakeDatabase;
+import liquibase.diff.output.DiffOutputControl;
+import liquibase.diff.output.changelog.ChangeGeneratorChain;
+import liquibase.structure.DatabaseObject;
+import liquibase.structure.core.Sequence;
+
+public class MissingSequenceChangeGeneratorSnowflake extends MissingSequenceChangeGenerator{
```
suryaaki2, line 12: Same comment as above

liquibase-snowflake/src/main/java/liquibase/sqlgenerator/core/CreateSequenceGeneratorSnowflake.java
```diff
@@ -0,0 +1,54 @@
...
+
+import liquibase.database.Database;
+import liquibase.database.core.*;
+import liquibase.exception.ValidationErrors;
+import liquibase.sql.Sql;
+import liquibase.sql.UnparsedSql;
+import liquibase.sqlgenerator.SqlGeneratorChain;
+import liquibase.statement.core.CreateSequenceStatement;
+
+public class CreateSequenceGeneratorSnowflake extends CreateSequenceGenerator{
```
suryaaki2, line 11: same comment

suryaaki2 (changes requested):

suryaaki2 (approved):

liquibase-snowflake/src/main/java/liquibase/diff/output/changelog/core/ChangedSequenceChangeGeneratorSnowflake.java
yodzhubeiskyi: The naming here goes according to the rest of the classes in Snowflake extension, so I'd consider it to be correct

PR #3127, opened 2022-07-28
Handle open of defaults file with new Resource API DAT-10444

wwillard7800:
DAT-10444

In order to support the ability to open defaults files that are not on the file system, we use the `PathHandlerFactory` to open the file, rather than directly opening it with the Java API


nvoxland (approved):

liquibase-cli/src/main/java/liquibase/integration/commandline/LiquibaseCommandLine.java
```diff
@@ -526,23 +526,25 @@ private List<ConfigurationValueProvider> registerValueProviders(String[] args) t
         }
 
         final File defaultsFile = new File(defaultsFileConfig.getValue());
```
nvoxland, line 528: Can this line be removed? Seems like it shouldn't be used anymore?

wwillard7800: Thanks, I removed it.

PR #3134, opened 2022-08-01
Modify command tests to prevent search path from using old value DAT-10772 

wwillard7800:


wwillard7800:
Code has already been merged.

PR #3140, opened 2022-08-03
Use PathHandlerFactory when building driver properties from file (DAT-11067)

abrackx:
## Impact

## Description

Updates DatabaseFactory#buildDriverProperties method to and to use PathHandlerFactory to load the driver properties file.

## Things to be aware of

Removed existing implementation in favor of using a resource accessor to grab the input stream.

## Things to worry about
N/A

## Additional Context
N/A


liquibase-core/src/test/java/liquibase/database/DatabaseFactoryTest.java
```diff
@@ -93,7 +96,11 @@ public void openConnectionLoadsDriverPropertiesFromGivenFile() throws Exception
         }
         String propsFilePath = propsFile.getAbsolutePath();
 
-        DatabaseConnection dbConnection = databaseFactory.openConnection("jdbc:h2:mem:DatabaseFactoryTest", "sa", "", null, null, propsFilePath, null, resourceAccessor);
+        Map<String, String> resources = new HashMap<>();
+        resources.put(propsFilePath, propsFile.toString());
+        ResourceAccessor mockAccessor = new MockResourceAccessor(resources);
```
abrackx, line 101: Unsure if this is the best approach but `mock(ResouceAccessor.class)` doesn't have anything populated so this test fails since I changed the buildDriverProperties method to actually use a ResourceAccessor.

liquibase-core/src/main/java/liquibase/database/DatabaseFactory.java
```diff
@@ -291,23 +294,26 @@ private Properties buildDriverProperties(String username, String password, Strin
...
-                    LOG.fine(
-                            "Loading properties from the file:'" + driverPropertiesFile + "'"
-                    );
-                    FileInputStream inputStream = new FileInputStream(propertiesFile);
-                    try {
-                        driverProperties.load(inputStream);
-                    } finally {
-                        inputStream.close();
+                try {
+                    String driverPropertiesFilePath = resourceAccessor.resolve(null, driverPropertiesFile);
```
nvoxland, line 298: You'll want to use the PathHandlerFactory in here as well too. In the soon-incoming `add-resource` branch there is a single call that will do this logic of looking for a file first via the way it's specified (including s3 etc.) and then falling back to the resource accessor as needed.

It's maybe best to wait for that PR to get through instead of doing it this way and then changing it for that?

liquibase-core/src/test/java/liquibase/database/DatabaseFactoryTest.java
abrackx: Irrelevant with different approach

liquibase-core/src/main/java/liquibase/database/DatabaseFactory.java
```diff
@@ -291,23 +292,26 @@ private Properties buildDriverProperties(String username, String password, Strin
...
+                try {
+                    PathHandlerFactory pathHandlerFactory = Scope.getCurrentScope().getSingleton(PathHandlerFactory.class);
+                    Resource driverProperty = pathHandlerFactory.getResource(driverPropertiesFile, true);
+                    if (driverProperty != null) {
+                        InputStream stream = driverProperty.openInputStream();
+                        LOG.fine(
+                                "Loading properties from the file:'" + driverPropertiesFile + "'"
+                        );
+                        driverProperties.load(stream);
+                        stream.close();
```
StevenMassaro, line 304: Can you do a try with resources here so the stream is always closed?

StevenMassaro (approved):

liquibase-core/src/main/java/liquibase/database/DatabaseFactory.java
abrackx: Updated!

suryaaki2 (approved):

nvoxland (approved):

PR #3152, opened 2022-08-05
Add default catalog condition to SequenceSnapshotGenerator for Oracle

### LINKED ISSUE #3138: Sequence precondition and sequence creation behave differently in Oracle when the catalog is not defined in the changeset
## Environment

**Liquibase Version**: 4.9.1

**Liquibase Integration & Version**: Spring Boot 2.7.2, OpenJDK 11.0.12, Maven 3.8.6

**Database Vendor & Version**: Oracle Express Edition 18.4.0

**Operating System Type & Version**: Ubuntu 20.04

**Infrastructure Type/Provider**: -

## Description

When trying to execute two equivalent changesets to create the same sequence, liquibase throws an error, even with precondition configured. That is a common scenario in a project where other libraries create the same sequence.

## Steps To Reproduce

Execute the following changelog connected to an Oracle database:
```
<databaseChangeLog xmlns="http://www.liquibase.org/xml/ns/dbchangelog/1.9" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
                   xsi:schemaLocation="http://www.liquibase.org/xml/ns/dbchangelog/1.9
                      http://www.liquibase.org/xml/ns/dbchangelog/dbchangelog-1.9.xsd">

    <changeSet id="10" author="gabrielgiampietro">
            <not><sequenceExists sequenceName="HIBERNATE_SEQUENCE" /></not>
        </preConditions>
        <createSequence sequenceName="HIBERNATE_SEQUENCE"/>
    </changeSet>

    <changeSet id="20" author="gabrielgiampietro">
            <not><sequenceExists sequenceName="HIBERNATE_SEQUENCE" /></not>
        </preConditions>
        <createSequence sequenceName="HIBERNATE_SEQUENCE"/>
    </changeSet>

</databaseChangeLog>
```
## Actual Behavior
When creating the sequence, Liquibase takes the default catalog name, in `AbstractJdbcDatabase`:
```
public String escapeObjectName(String catalogName, String schemaName, final String objectName,
                                   final Class<? extends DatabaseObject> objectType) {
...
    } else {
        catalogName = this.getDefaultCatalogName();

        if (catalogName == null) {
            return escapeObjectName(objectName, objectType);
        } else {
            if (defaultCatalogSet || (isDefaultCatalog(catalogName) && getOutputDefaultCatalog())) {
                return escapeObjectName(catalogName, Catalog.class) + "." + escapeObjectName(objectName, objectType);
            } else {
                return escapeObjectName(objectName, objectType);
            }
        }
    }
...
}
```
The SQL generated to create the sequence is the following: `CREATE SEQUENCE SYSTEM.HIBERNATE_SEQUENCE`

When executing the precondition in the changeset with id 20, in `SequenceSnapshotGenerator`, method `protected String getSelectSequenceSql(Schema schema, Database database)`, the generated query is the following:
```
SELECT sequence_name, 
CASE WHEN increment_by > 0 
     THEN CASE WHEN min_value=1 THEN NULL ELSE min_value END
     ELSE CASE WHEN min_value=(-999999999999999999999999999) THEN NULL else min_value END
END AS min_value, 
CASE WHEN increment_by > 0 
     THEN CASE WHEN max_value=999999999999999999999999999 THEN NULL ELSE max_value END
     ELSE CASE WHEN max_value=last_number THEN NULL else max_value END 
END  AS max_value, 
CASE WHEN increment_by = 1 THEN NULL ELSE increment_by END AS increment_by, 
CASE WHEN cycle_flag = 'N' THEN NULL ELSE cycle_flag END AS will_cycle, 
CASE WHEN order_flag = 'N' THEN NULL ELSE order_flag END AS is_ordered, 
LAST_NUMBER as START_VALUE, 
CASE WHEN cache_size = 20 THEN NULL ELSE cache_size END AS cache_size 
FROM ALL_SEQUENCES WHERE SEQUENCE_OWNER = 'null'
```
Oracle searches for a sequence where `SEQUENCE_OWNER = 'null'`, but it doesn't find any, so it tries to create the sequence. Then, the error `ORA-00955: name is already used by an existing object` shows up. This is the full log output: [complete.log](https://github.com/liquibase/liquibase/files/9243465/complete.log)

## Expected/Desired Behavior
The precondition on the changeset with id 20 should find the sequence and not try to create the same sequence.

## Additional Context
A possible solution would be to change `SequenceSnapshotGenerator#getSelectSequenceSql()` in case the catalog is not set in the changeset (for Oracle only), and one of these could happen: to search for the default catalog name of the database, or to not add the `WHERE` clause.
The same bug happens with an older setup: Liquibase 3.x, Oracle Express Edition 12.2.0.2, JDK Oracle 7u80, and Maven 3.0.4.

gabrielnardes:
## Impact

## Description

When verifying if a sequence exists in a Oracle, Liquibase sets the column SEQUENCE_OWNER according to the catalog name provided in the changeset. If the catalog is not set, Liquibase will try to search for a SEQUENCE_OWNER='null'. We can set this value to the default catalog name of the database, because, according to the docs, this is a **not null** column: https://docs.oracle.com/en/database/oracle/oracle-database/21/refrn/ALL_SEQUENCES.html

Fixes #3138


nvoxland (approved):
Code review and test results:

Things to be aware of:
- The changes make sense to me. It's ensuring the catalog name is correctly set in the existing query

Things to worry about:
- Nothing

suryaaki2 (approved):

XDelphiGrl (approved):
PR prevents Oracle sequence owner being incorrectly set to "null" when catalogName is not specified on a changeset. Oracle defines the ALL_SEQUENCES.SEQUENCE_OWNER as not-null. This means that when a changeset does not include catalogName, it is safe to use the default catalog as the SEQUENCE_OWNER when querying for an Oracle sequence.
* Very nice [new integration test added](liquibase-core/src/test/groovy/liquibase/snapshot/jvm/SequenceSnapshotGeneratorTest.groovy) which makes the behavior very clear. 

APPROVED

tabbyf00:

PR #3158, opened 2022-08-08
use PathHandlerFactory for outputFile parameter (DAT-11034)

StevenMassaro:
## Impact

## Description

Use the PathHandlerFactory to write to files specified in the `outputFile` parameter.

## Things to worry about

This should not be a breaking change.

## Additional Context

This is being done to support the S3 work.


abrackx (approved):

nvoxland (approved):
That is the right new pattern to be using for writing to files, I'd assume there will be more to fix as we find them.

PR #3164, opened 2022-08-10
Write changelogs created by generate-changelog and diff-changelog in S3 (DAT-11044)

StevenMassaro:
## Impact

## Description

Use PathHandlers for output files for generate-changelog and diff-changelog commands, allowing output of those files to S3.

## Things to worry about

- I had to pretty significantly change the way that output files are written. Existing tests continue to pass, but it is possible that the changes I made exclude an edge case.


liquibase-core/src/main/java/liquibase/diff/output/changelog/DiffToChangeLog.java
```diff
@@ -91,23 +92,38 @@ public void print(PrintStream out) throws ParserConfigurationException, IOExcept
...
+        final PathHandlerFactory pathHandlerFactory = Scope.getCurrentScope().getSingleton(PathHandlerFactory.class);
+        Resource resource = pathHandlerFactory.getResource(changeLogFile, true);
+        boolean fileExists;
+        // todo, an exists method would be useful here, instead of this hack
+        try (InputStream is = resource.openInputStream()) {
+            fileExists = is != null;
+        } catch (IOException | NullPointerException e) {
+            fileExists = false;
+            Scope.getCurrentScope().getLog(getClass()).warning("Unable to open input stream to file " + changeLogFile, e);
+        }
```
StevenMassaro, line 104: This is not ideal. How about adding an `exists` method to PathHandler?

liquibase-core/src/main/java/liquibase/diff/output/changelog/DiffToChangeLog.java
```diff
@@ -91,23 +92,38 @@ public void print(PrintStream out) throws ParserConfigurationException, IOExcept
...
+            objectsDir = CommonsFilenameUtils.concat(parent, "objects-" + new Date().getTime());
         }
 
         if (objectsDir != null) {
-            if (objectsDir.exists()) {
-                throw new UnexpectedLiquibaseException("The generatechangelog command would overwrite your existing stored logic files. To run this command please remove or rename the '"+objectsDir.getCanonicalPath()+"' dir in your local project directory");
-            }
+//            if (objectsDir.exists()) {
+//                throw new UnexpectedLiquibaseException("The generatechangelog command would overwrite your existing stored logic files. To run this command please remove or rename the '"+objectsDir.getCanonicalPath()+"' dir in your local project directory");
+//            }
```
StevenMassaro, line 126: I can't think of a good way to do this check.

liquibase-core/src/main/java/liquibase/diff/output/changelog/DiffToChangeLog.java
```diff
@@ -91,23 +92,38 @@ public void print(PrintStream out) throws ParserConfigurationException, IOExcept
...
+            fileExists = is != null;
+        } catch (IOException | NullPointerException e) {
+            fileExists = false;
+            Scope.getCurrentScope().getLog(getClass()).warning("Unable to open input stream to file " + changeLogFile, e);
+        }
 
         final Map<String, Object> newScopeObjects = new HashMap<>();
 
-        File objectsDir = null;
+        String objectsDir = null;
```
StevenMassaro, line 108: Test that the objects directory logic still works.

liquibase-core/src/main/java/liquibase/diff/output/changelog/DiffToChangeLog.java
StevenMassaro, line 104: Group agrees that this should be replaced with an `exists(String path)` method in the PathHandler interface.

liquibase-core/src/main/java/liquibase/diff/output/changelog/DiffToChangeLog.java
```diff
@@ -91,23 +92,38 @@ public void print(PrintStream out) throws ParserConfigurationException, IOExcept
...
+        if (resource != null) {
+            parent = CommonsFilenameUtils.getPath(resource.getPath());
+        } else {
+            parent = CommonsFilenameUtils.getPath(changeLogFile);
+        }
         if (changeLogFile.toLowerCase().endsWith("sql")) {
             DeprecatedConfigurationValueProvider.setData("liquibase.pro.sql.inline", "true");
         } else if (this.diffResult.getComparisonSnapshot() instanceof EmptyDatabaseSnapshot) {
-            objectsDir = new File(file.getParentFile(), "objects");
+            objectsDir = CommonsFilenameUtils.concat(parent, "objects");
```
StevenMassaro, line 118: Does this work in S3?

liquibase-core/src/main/java/liquibase/util/CommonsFilenameUtils.java
```diff
@@ -0,0 +1,1684 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one or more
+ * contributor license agreements.  See the NOTICE file distributed with
+ * this work for additional information regarding copyright ownership.
+ * The ASF licenses this file to You under the Apache License, Version 2.0
+ * (the "License"); you may not use this file except in compliance with
+ * the License.  You may obtain a copy of the License at
+ *
```
StevenMassaro, line 8: Bring in only what is needed here. Delete unused methods.

Also, make sure to bring in unit tests.

liquibase-core/src/main/java/liquibase/diff/output/changelog/DiffToChangeLog.java
StevenMassaro: Wes thinks that this is possible with the aws cli

s3 ls bucket:foldername (without slash) - won't be found
s3 ls bucket:foldername (with slash) - will respond yes, that a directory does exist here

liquibase-core/src/main/java/liquibase/diff/output/changelog/DiffToChangeLog.java
```diff
@@ -124,15 +140,17 @@ public void print(String changeLogFile, ChangeLogSerializer changeLogSerializer)
...
                     try {
-                        if (!file.exists()) {
+                        if (!finalFileExists) {
                             //print changeLog only if there are available changeSets to print instead of printing it always
-                            printNew(changeLogSerializer, file);
+                            printNew(changeLogSerializer, changeLogFile);
                         } else {
-                            Scope.getCurrentScope().getLog(getClass()).info(file + " exists, appending");
+                            Scope.getCurrentScope().getLog(getClass()).info(changeLogFile + " exists, appending");
+                            StringBuilder fileContents = new StringBuilder(StreamUtil.readStreamAsString(resource.openInputStream()));
```
StevenMassaro, line 153: @nvoxland - is it okay that we're loading the changelog file into memory here? Would it be better to instead download the file from S3 to temp directory and continue to use the same `RandomAccessFile` logic that used to be here?

liquibase-core/src/main/java/liquibase/resource/PathHandler.java
```diff
@@ -37,4 +37,12 @@ public interface PathHandler extends Plugin {
      * @throws IOException if the path cannot be written to
      */
     OutputStream createResource(String path) throws IOException;
+
+    /**
+     * Check if a resource exists at the specified path.
+     * @return true if the file exists, false otherwise
+     */
+    boolean exists(String path);
+
```
StevenMassaro, line 46: Javadoc

liquibase-core/src/main/java/liquibase/resource/Resource.java
```diff
@@ -42,4 +42,6 @@ public interface Resource {
      * For example, a file resource may return a path of <code>my/file.txt</code> and a uri of <code>file:/tmp/project/liquibase/my/file.txt</code> for a resource accessor using <code>file:/tmp/project/liquibase</code> as a root
      */
     URI getUri();
+
+    String getAbsolutePath();
```
StevenMassaro, line 46: Javadoc

liquibase-core/src/main/java/liquibase/util/CommonsFilenameUtils.java
```diff
@@ -0,0 +1,902 @@
...
+ *      http://www.apache.org/licenses/LICENSE-2.0
```
wwillard7800, line 9: We talked about only copying the code for the methods that we needed from FilenameUtils.  Is this all of them or the trimmed down version?

StevenMassaro: Believe it or not this is the trimmed down version.

wwillard7800 (approved):

StevenMassaro:
Note to self - make sure to add tests for FilenameUtils and other classes.

liquibase-core/src/main/java/liquibase/resource/PathResource.java
```diff
@@ -35,13 +35,36 @@ public InputStream openInputStream() throws IOException {
...
-    public OutputStream openOutputStream() throws IOException {
+    public OutputStream openOutputStream(boolean createIfNeeded) throws IOException {
+        if (!exists()) {
+            if (createIfNeeded) {
+                path.getParent().toFile().mkdirs();
+            } else {
+                throw new IOException("File " + this.getUri() + " does not exist");
+            }
+        }
+
```
StevenMassaro, line 67: These changes are fine. I just want to point out that `Files.newOutputStream(this.path)` will truncate the existing file before writing to it. I found this behavior a bit surprising originally but figured it was intentional. Now, with the new `createIfNeeded` parameter, this behavior is much more surprising. If it's intentional and desired, let's call it out in the documentation.

liquibase-core/src/main/java/liquibase/resource/AbstractResource.java
```diff
@@ -60,4 +60,18 @@ public boolean equals(Object obj) {
...
+     */
+    protected String resolvePath(String other) {
+        return getPath() + "/" + other;
+    }
+
+    /**
+     * Convenience method for computing the relative path in {@link #resolveSibling(String)} implementations.
+     */
+    protected String resolveSiblingPath(String other) {
+        return getPath().replaceFirst("/[^/]*$", "") + "/" + other;
```
StevenMassaro, line 75: This looks brittle. Is that a problem?

liquibase-core/src/main/java/liquibase/diff/output/changelog/DiffToChangeLog.java
nvoxland: With the new interfaces only exposing input/output streams, I think copying like this is all we can do. The files are probably generally small enough that streaming it to a temp file and then using RandomAccessFile is probably unneeded?

But I wouldn't be against doing it too. I suppose best now when we remember the code is in git history?

liquibase-core/src/main/java/liquibase/resource/PathResource.java
nvoxland: It seems like the truncating is the behavior we want? I was a bit worried that we'd have to add logic to do that, but sounds like we're not?

I think the expectation is "openOutputStream lets you write what you want for content to the file" and not have to worry about "if you write less bytes than the current file has, the final result is a combination of what you wrote and what was there before". 

But why are you thinking that especially with the createIfNeeded it's extra surprising? I'm maybe not following?

liquibase-core/src/main/java/liquibase/resource/AbstractResource.java
nvoxland: I thought maybe it was OK? And since its a protected method, it can be overriden for subclasses where it doesn't quite work.

But what types of paths did you think it would be  a problem for?

liquibase-core/src/main/java/liquibase/resource/PathResource.java
StevenMassaro: Truncating probably is the behavior we want. I just think it needs to be called out expressly in the documentation, because it might not be obvious to a user. It would be confusing to a user who thinks to themselves "I told you not to create the file, why are you overwriting the entire thing".

> I think the expectation is "openOutputStream lets you write what you want for content to the file" and not have to worry about "if you write less bytes than the current file has, the final result is a combination of what you wrote and what was there before".

I can see where you're going here. I can also see a not insignificant number of users who are surprised by the fact that "opening" an output stream to a file also erases the entire file contents.

liquibase-core/src/main/java/liquibase/diff/output/changelog/DiffToChangeLog.java
StevenMassaro: One vote here for keeping the logic the way its written - loading the file into memory. The change will still be easily found in the git history when this is merged. I am not worried about that.

liquibase-core/src/main/java/liquibase/resource/AbstractResource.java
StevenMassaro: What about a path that uses backslashes instead of forward slashes?

StevenMassaro: Coming up with examples of how to break this is not easy. That's why I used the Commons IO code when I did the first round on this, because I was relatively confident that they have come up with good solutions to these situations.

liquibase-core/src/main/java/liquibase/integration/spring/SpringResource.java
```diff
@@ -31,7 +59,10 @@ public InputStream openInputStream() throws IOException {
     }
 
     @Override
-    public OutputStream openOutputStream() throws IOException {
+    public OutputStream openOutputStream(boolean createIfNeeded) throws IOException {
+        if (!resource.exists() && !createIfNeeded) {
+            throw new IOException("Resource "+getUri()+" does not exist");
+        }
```
StevenMassaro, line 65: This logic should be abstracted up a layer, since this check will have to occur in every implementation of Resource.

liquibase-core/src/test/groovy/liquibase/resource/PathHandlerFactoryTest.groovy
```diff
@@ -25,6 +26,45 @@ class PathHandlerFactoryTest extends Specification {
...
+        (pathHandlerFactory.getResource(path).exists()) == existsWithoutResourceAccessor
+        Scope.child(Scope.Attr.resourceAccessor, new JUnitResourceAccessor(), { ->
+            assert (pathHandlerFactory.getResource(path, true).exists()) == existsWithResourceAccessor
+        })
+
+
+        where:
+        path                                                               | existsWithoutResourceAccessor | existsWithResourceAccessor
+//        "src/test/groovy/liquibase/resource/PathHandlerFactoryTest.groovy" | true                          | true
+//        "invalid/path.txt"                                                 | false                         | false
```
StevenMassaro, line 44: Do you want these commented out?

StevenMassaro:
FYI there are test failures in the pro-side of this PR

liquibase-core/src/main/java/liquibase/resource/AbstractResource.java
nvoxland: That path should be the standardized "path in the resourceAccessor" so it should already just use `/` and not have any odd `..` or whatever in them. So shouldn't need to much fancy logic?

liquibase-core/src/test/groovy/liquibase/resource/PathHandlerFactoryTest.groovy
nvoxland: No, those should get put back. Had been skipping them to debug later ones that were failing. Thanks for catching that

liquibase-core/src/main/java/liquibase/integration/spring/SpringResource.java
nvoxland: I'm not sure how it can, though? The `resource` variable there is a springframework.Resource class, so that is implementation-specific code.

nvoxland (approved):

liquibase-core/src/main/java/liquibase/resource/PathHandlerFactory.java
```diff
@@ -95,32 +98,21 @@ public Resource getResource(String resourcePath, boolean includeResourceAccessor
...
      */
-    public boolean isAbsolute(String path) throws IOException {
-        PathHandler plugin = getPlugin(path);
-        if (plugin == null) {
-            throw new IOException("Cannot parse resource location: '" + path + "'");
-        }
-        return plugin.isAbsolute(path);
-    }
-
     private static class FoundResourceAccessor implements ResourceAccessor {
```
wwillard7800, line 116: It would be good to have a description of this inner class.

wwillard7800 (approved):
Looks good. Minor comment.

liquibase-core/src/main/java/liquibase/resource/PathHandlerFactory.java
StevenMassaro: There is a description for this class, must have been moved or something due to the way the diff is rendered, but it's definitely there.

PR #3187, opened 2022-08-22
DAT-10810 QC: checks show changes (--show-cols flag, drop some, etc)

dyadyushko:
## Impact

## Description

Minor changes within TableOutput class.
Adding toLowerWithoutWhitespaces method to StringUtil class.
example: 
    "  InPu t " -> "input"

## Additional Context
Changes were made to support https://github.com/liquibase/liquibase-pro/pull/628.


wwillard7800 (approved):

liquibase-core/src/main/java/liquibase/util/StringUtil.java
```diff
@@ -1031,6 +1031,14 @@ public static byte[] getBytesWithEncoding(String string) {
         return string.getBytes(StandardCharsets.UTF_8);
     }
 
+    /**
+     * @param value string to process
+     * @return string without any whitespaces formatted to lowercase.
+     */
+    public static String toLowerWithoutWhitespaces(String value) {
+        return value.toLowerCase().replaceAll("\\s+", "");
+    }
```
StevenMassaro, line 1040: Should probably handle the null case here. Maybe something like:

```suggestion
    public static String toLowerWithoutWhitespaces(String value) {
        if (value == null) {
            return null;
        }
        return value.toLowerCase().replaceAll("\\s+", "");
    }
```

liquibase-core/src/test/groovy/liquibase/util/StringUtilTest.groovy
```diff
@@ -118,6 +119,17 @@ class StringUtilTest extends Specification {
...
+    def "to lower without whitespaces"() {
+        expect:
+        StringUtil.toLowerWithoutWhitespaces(string) == expected
+
+        where:
+        string                | expected
+        "First Value"         | "firstvalue"
+        "secondValue"         | "secondvalue"
+        "   Third Value   \n" | "thirdvalue"
+    }
```
StevenMassaro, line 131: Add a test for the null case here. Maybe also add a test for empty string.

liquibase-core/src/main/java/liquibase/util/StringUtil.java
```diff
@@ -1031,6 +1031,14 @@ public static byte[] getBytesWithEncoding(String string) {
         return string.getBytes(StandardCharsets.UTF_8);
     }
 
+    /**
+     * @param value string to process
+     * @return string without any whitespaces formatted to lowercase.
+     */
+    public static String toLowerWithoutWhitespaces(String value) {
```
nvoxland, line 1038: This should be updated to return null if value is null instead of throwing a nullpointerexception

nvoxland (approved):
Added one comment on something good to fix, but still marking it as approved so I don't have to be a bottleneck in re-approving it

liquibase-core/src/main/java/liquibase/util/StringUtil.java
dyadyushko: Nice catch, going to fix!

liquibase-core/src/main/java/liquibase/util/StringUtil.java
dyadyushko: Definitely, gonna do that

PR #3189, opened 2022-08-22
Removed invalid warning in Maven

nvoxland:
## Impact

## Description

When using maven, if you did not specify an outputFileEncoding setting, it gave a warning about the encoding being system dependent. That is not the case, it defaults to UTF-8 so I removed the warning. We used to fall back to the system encoding instead of UTF-8, but changed that quite a while ago (maybe 4.6?)

I also cleaned up the code slightly so that there is consistent creation of the stream regardless of how the encoding was determined.

## Things to be aware of

- Only impacts maven

## Things to worry about

- Nothing


XDelphiGrl (approved):
* Fix removes a misleading warning output from Liquibase Maven output.
  * If an encoding is set, Liquibase uses that.
  * If an encoding is not set, Liquibase defaults to UTF8.

APPROVED

nvoxland (commented):
Code review and test results:

Things to be aware of:
- Changes make sense
- No automated tests, but we don't have a great maven automated test setup yet so we'll manually test it

Things to worry about:
- Nothing

PR #3194, opened 2022-08-23
Adds flow file examples (DAT-11287)

abrackx:
## Impact

## Description

Adds example flow files.


StevenMassaro:
Why is the flowvariables file only in the xml folder? Shouldn't it be in all of the folders?

suryaaki2:
> Why is the flowvariables file only in the xml folder? Shouldn't it be in all of the folders?

If it is the same file then why it should be in all the folders?

abrackx:
Good question! I was just following the [ticket](https://datical.atlassian.net/browse/DAT-11287), I'll confirm with mario this was intentional.

abrackx:
This was _not_ intentional, all directories should be the same. Good catch yall!

- [ ] Add flowvariables to all directories

StevenMassaro (approved):

PR #3197, opened 2022-08-24
DAT-11427 add manual build trigger

ap-liquibase:
## Impact

## Description

DAT-11427 add manual build trigger for QA team

## Things to be aware of

- Nothing

## Things to worry about

- Nothing

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -2,10 +2,6 @@ name: Build and Test
 
 on:
   workflow_dispatch:
-    inputs:
-      branch:
-        description: 'Branch to build manually (Defaults to master)'
-        required: false
 
   pull_request_target:
     types:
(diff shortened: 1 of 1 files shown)


jnewton03:
@ap-liquibase wouldn't the user just select the branch to run instead of typing it into a text box?

ap-liquibase:
@jnewton03 I think you're right, and it would simplify things too. I assume the branch selection pulldown appears automatically if I get rid of the following, right.
 ```    
inputs:
      branch:
        description: 'Branch to build manually (Defaults to master)'
        required: false
```

nvoxland:
I updated the PR description to remove the template text that should have been replaced vs. left.

nvoxland (approved):
I feel like this is probably not going to work like hoped, but we'll have to actually get it merged to master to test it to know for sure. 

Reasons I worry it might not work:
- I don't know what the context.payload is for manual runs, and may not have the correct info required for the "get current branch" and other logic which maybe expects it to be a pull request or a push. 
- The logic on the pro-tests side that looks for the corresponding branch on liquibase doesn't actually look for a corresponding "branch" is maybe less about finding a branch and more about assuming it's a pull request it's finding. So the pro-tests may not actually find this still
- When this build runs, it attaches artifacts to the commit hash, which is not going to be unique just because you made a branch. This may cause use to get builds attached to "master" for example, which have liquibase-commercial jars in them from non-master liquibase-pro branches which will wrong.

I'm not sure if either of those 3 will be problems, or if there will be other problems I haven't thought of. But it's possible thatthis is not an "easy" work-around to the problem that is hoped.

PR #3232, opened 2022-09-01
simplify file loading logic in remaining classes (DAT-11569)

StevenMassaro:
## Impact

## Description

File loading logic needed to be corrected and simplified in a few places.


suryaaki2 (approved):

liquibase-core/src/main/java/liquibase/change/core/CreateViewChange.java
```diff
@@ -166,11 +166,11 @@ protected InputStream openSqlStream() throws IOException {
         try {
             ResourceAccessor resourceAccessor = Scope.getCurrentScope().getResourceAccessor();
 
-            String path = getPath();
             if (ObjectUtil.defaultIfNull(getRelativeToChangelogFile(), false)) {
-                path = resourceAccessor.get(getChangeSet().getChangeLog().getPhysicalFilePath()).resolveSibling(path).getPath();
+                return resourceAccessor.get(getChangeSet().getChangeLog().getPhysicalFilePath()).resolveSibling(getPath()).openInputStream();
```
nvoxland, line 170: Does this (and the similar change inSQLFileChange) throw a good error when referencing a file that does not exist in this branch?

StevenMassaro: It depends on the implementation (currently). `PathResource` throws a `FileNotFoundException` but the `S3Resource` logs a message and returns null.

PR #3233, opened 2022-09-01
Postgresql: Fix for case when current search_path is an empty string

gadget:
## Impact

## Description

In certain cases the search path in a postgres database can be empty (not null, but an empty string). For us it happens when running postgres with testcontainers for IT tests, probably on rollbacks (as you call initializeDatabase for every rollback to reset default schema). In case searchPath is empty we should not try and process it, just re-set the default schema.

### DIFF
--- a/liquibase-core/src/main/java/liquibase/database/core/DatabaseUtils.java
+++ b/liquibase-core/src/main/java/liquibase/database/core/DatabaseUtils.java
@@ -10,6 +10,7 @@ import liquibase.executor.ExecutorService;
 import liquibase.statement.core.RawSqlStatement;
 import liquibase.structure.core.Schema;
 import liquibase.util.StringUtil;
+import org.apache.commons.lang3.StringUtils;
 
 public class DatabaseUtils {
     /**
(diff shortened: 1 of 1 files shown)


nvoxland (approved):
Code review and test results:

Things to be aware of:
- Changes make sense to me, it's keeping an invalid search_path with a trailing , from being made
- Probably not a code path that gets hit in our code's setup, and not something that is easy to write a test for at this point. But existing tests should catch anything surprising it breaks.

Things to worry about:
- Nothing

suryaaki2 (approved):

XDelphiGrl (approved):
PR fixes a bug to prevent empty Postgres SEARCH_PATH values from being set. In the case where the SEARCH_PATH is empty, Liquibase will use the default schema as the SEARCH_PATH.

APPROVED

Note: Functional test failures on DB2 LUW and MariaDB are due to test validations out of sync with current Liquibase master; this is not related to this PR (which only impacts Postgres, which is passing).

tabbyf00:

PR #3246, opened 2022-09-07
Fix generateChangeLog and diffChangeLog logic to avoid including default schema name when it should not

### LINKED ISSUE #2776: Postgres generateChangeLog includeSchema
## Environment

**Liquibase Version**: 4.8.0

**Liquibase Integration & Version**: <Pick one: CLI, maven, gradle, spring boot, servlet, etc.> CLI

**Database Vendor & Version**: Postgresql 12.4

**Operating System Type & Version**: Amazon Aurora

## Description

I am trying to generate DDL from the postgres database using Liquibase.  I would like the schema to be suppressed in the output.  However, when I run:
      liquibase --changeLogFile=./ddl-build/changelog.etl1.new.sql --includeSchema=false --schemas=wetl generateChangeLog
I get oracle as schema, such as:
     CREATE TABLE "oracle"."activity_defn"
I don't know where this "oracle" is coming from, I am pointing to a schema called "wetl".  

## Steps To Reproduce

This is the liquibase command I have run:
liquibase --changeLogFile=./ddl-build/changelog.etl1.new.sql --includeSchema=false --schemas=wetl generateChangeLog

## Actual Behavior
Object owner called "oracle" gets appended to the objects when I generate the changelog and set includeSchema=false
Example: CREATE TABLE "oracle"."activity_defn"

## Expected/Desired Behavior
There should be no object owner in the changelog output when includeSchema is set to false.  
Example: CREATE TABLE "activity_defn"

n/a

## Additional Context
n/a

MalloD12:
## Impact

## Description

Fix generateChangeLog and DiffChangeLog logic to avoid including default schema name when `includeSchema=false`.

FIXES #2776 

## Things to be aware of

- Update logic by setting up outputDefaultSchema flag depending on value set of `includeSchema` flag parameter.
- Only impacts generateChangeLog and diffChangelog CLI.

## Things to worry about

- Nothing

## Additional Context

- Nothing


nvoxland (approved):
Code review and test results:

Things to be aware of:
- The changes make sense to me
- New tests (and changed behavior of existing setup) is shown in liquibase/liquibase-pro-tests#447 and those look correct
- Only impacts generateChangeLog with formatted sql
The generated sql will be different for people that are used to having the schema included. But they can use the existing flags to (now correctly) configure that behavior. It's good the formatted sql output now matches the default values.

Things to worry about:
- Nothing

XDelphiGrl:
@MalloD12 and @nvoxland, hi -

In reading the linked issue, it looks like the problematic command included both `include-schema` and `schemas`.  
* Does the bug occur only when the `schemas` attribute was used in conjunction with `include-schema false`? 
* Does the bug reproduce if `include-schema` is omitted from the problematic command? I ask because the default for `include-schema` is `false`.  
* If the answer to either of those questions is "yes", the functional tests will need to be adjusted. 

Thanks!

suryaaki2 (approved):

MalloD12:
> @MalloD12 and @nvoxland, hi -
> 
> In reading the linked issue, it looks like the problematic command included both `include-schema` and `schemas`.
> 
> * Does the bug occur only when the `schemas` attribute was used in conjunction with `include-schema false`?
> * Does the bug reproduce if `include-schema` is omitted from the problematic command? I ask because the default for `include-schema` is `false`.
> * If the answer to either of those questions is "yes", the functional tests will need to be adjusted.
> 
> Thanks!

Hi @XDelphiGrl, sorry for the late reply but I just saw these questions. Below my answers:

1.  The problem here was basically how include-schemas, because even without specifying any schema but saying to liquibase to do not include schema (`include-schema=false`) when generating an sql format changelog default schema name was present. So that's what we have fixed.
2. Yeap, behaviour is the same. If we don't specify `include-schema=false` default schema name is also present when generating changelog.

Thanks for your valuable questions, and again my apologises for my late response. 

Daniel.

XDelphiGrl:
> > @MalloD12 and @nvoxland, hi -
> > In reading the linked issue, it looks like the problematic command included both `include-schema` and `schemas`.
> > 
> > * Does the bug occur only when the `schemas` attribute was used in conjunction with `include-schema false`?
> > * Does the bug reproduce if `include-schema` is omitted from the problematic command? I ask because the default for `include-schema` is `false`.
> > * If the answer to either of those questions is "yes", the functional tests will need to be adjusted.
> > 
> > Thanks!
> 
> Hi @XDelphiGrl, sorry for the late reply but I just saw these questions. Below my answers:
> 
>     1. The problem here was basically how include-schemas, because even without specifying any schema but saying to liquibase to do not include schema (`include-schema=false`) when generating an sql format changelog default schema name was present. So that's what we have fixed.
> 
>     2. Yeap, behaviour is the same. If we don't specify `include-schema=false` default schema name is also present when generating changelog.
> 
> 
> Thanks for your valuable questions, and again my apologises for my late response.
> 
> Daniel.

@MalloD12, no need to apologize! We are all very busy. Thank you for answering my questions, I appreciate the extra clarity.

XDelphiGrl (approved):
Fix prevents schema names from being included in generated changelogs when include-schema is false. 
* Functional tests were updated to match the output from the fixed Core code.

APPROVED

PR #3272, opened 2022-09-13
snowflake no longer requires extension

kataggart:
readme update - moved snowflake out of list requiring extension to list of included databases


nvoxland:
This list has been removed with #3280

PR #3274, opened 2022-09-13
Improve int/tinyint/smallint/bigint handling in H2

### LINKED ISSUE #3250: Table creation fails with H2 v2 when a change set specifies an integer size, e.g. INT(10)
## Environment

**Liquibase Version**: 4.9.1

**Liquibase Integration & Version**: SpringBoot v2.7.3

**Database Vendor & Version**: H2 v2.1.214

**Operating System Type & Version**: MacOS v12.5

## Description

Since upgrading to H2 v2 my change set no longer applies.

## Steps To Reproduce

Create a change set such as this:

```xml
 <changeSet author="mark" id="createPersonsTable">
     <createTable tableName="persons">
         <column name="personId" type="INT(10)"/>
         <column name="lastName" type="VARCHAR(255)"/>
         <column name="firstName" type="VARCHAR(255)"/>
         <column name="address" type="VARCHAR(255)"/>
         <column name="city" type="VARCHAR(255)"/>
     </createTable>
 </changeSet>
```

The magic part is the `INT(10)`.

Recreation is [here](https://github.com/CynanX/LiquibaseH2Issue).

## Actual Behavior

An exception is throw: -

```
2022-09-07 10:11:41.855  WARN 23829 --- [           main] o.s.w.c.s.GenericWebApplicationContext   : Exception encountered during context initialization - cancelling refresh attempt: org.springframework.beans.factory.BeanCreationException: Error creating bean with name 'liquibase' defined in class path resource [org/springframework/boot/autoconfigure/liquibase/LiquibaseAutoConfiguration$LiquibaseConfiguration.class]: Invocation of init method failed; nested exception is liquibase.exception.LiquibaseException: liquibase.exception.MigrationFailedException: Migration failed for change set changelog/1.0.0/tables/create_persons_table.xml::createPersonsTable::mark:
     Reason: liquibase.exception.DatabaseException: Syntax error in SQL statement "CREATE TABLE PUBLIC.persons (personId INT[*](10), lastName VARCHAR(255), firstName VARCHAR(255), address VARCHAR(255), city VARCHAR(255))"; expected "ARRAY, INVISIBLE, VISIBLE, NOT NULL, NULL, AS, DEFAULT, GENERATED, ON UPDATE, NOT NULL, NULL, AUTO_INCREMENT, DEFAULT ON NULL, NULL_TO_DEFAULT, SEQUENCE, SELECTIVITY, COMMENT, CONSTRAINT, COMMENT, PRIMARY KEY, UNIQUE, NOT NULL, NULL, CHECK, REFERENCES, AUTO_INCREMENT, ,, )"; SQL statement:
CREATE TABLE PUBLIC.persons (personId INT(10), lastName VARCHAR(255), firstName VARCHAR(255), address VARCHAR(255), city VARCHAR(255)) [42001-214] [Failed SQL: (42001) CREATE TABLE PUBLIC.persons (personId INT(10), lastName VARCHAR(255), firstName VARCHAR(255), address VARCHAR(255), city VARCHAR(255))]
```

## Expected/Desired Behavior

The table should be created.

## Additional Context

The above change set works fine when using H2 v1.x.

### LINKED ISSUE #3300: BIGINT with precision doesn't work anymore with H2 database
## Environment

**Liquibase Version**: `4.16.1`

**Liquibase Integration & Version**: `CLI`

**Database Vendor & Version**: `H2 2.1.214`

**Infrastructure Type/Provider**: Mac OS Monterey 12.1

## Description

Liquibase doesn't translate `BIGINT` (`BOOLEAN`?) field to correct value in H2 database. 

## Steps To Reproduce

<Changelog/Changeset Example>

## Actual Behavior
A clear and concise description of what happens in the software with the **version used**.

## Expected/Desired Behavior
Column with `BIGINT(xy)` (maybe `BOOLEAN` too) is correctly translated.

MalloD12:
## Impact

## Description

Implement support to consider integer precision by parsing to the specific integer type (tiny, integer or bigint) depending on the specified size.

Fixes #3250 
Fixes #3300 

## Things to be aware of

- This change only impacts integer data type for H2.

## Things to worry about

- Nothing.

## Additional Context

- Nothing.

### DIFF
--- a/liquibase-core/src/main/java/liquibase/datatype/core/BigIntType.java
+++ b/liquibase-core/src/main/java/liquibase/datatype/core/BigIntType.java
@@ -73,6 +73,9 @@ public class BigIntType extends LiquibaseDataType {
         if (database instanceof SybaseASADatabase) {
             return new DatabaseDataType("BIGINT");
         }
+        if (database instanceof H2Database) {
+            return new DatabaseDataType("BIGINT");
+        }
         return super.toDatabaseDataType(database);
     }
 
--- a/liquibase-core/src/main/java/liquibase/datatype/core/IntType.java
+++ b/liquibase-core/src/main/java/liquibase/datatype/core/IntType.java
@@ -77,7 +77,9 @@ public class IntType extends LiquibaseDataType {
             int intParameter = Integer.valueOf(getParameters()[0].toString());
             if(intParameter >= 1 && intParameter <= 3) {
                 return new DatabaseDataType("TINYINT");
-            } else if (intParameter > 3 && intParameter <= 10) {
+            } else if (intParameter > 3 && intParameter <= 5) {
+                return new DatabaseDataType("SMALLINT");
+            } else if (intParameter > 5 && intParameter <= 10) {
                 return new DatabaseDataType("INTEGER");
             } else if (intParameter > 10) {
                 return new DatabaseDataType("BIGINT");
--- a/liquibase-core/src/main/java/liquibase/datatype/core/IntType.java
+++ b/liquibase-core/src/main/java/liquibase/datatype/core/IntType.java
@@ -72,6 +72,17 @@ public class IntType extends LiquibaseDataType {
         if (database instanceof SybaseASADatabase) {
             return new DatabaseDataType("INTEGER");
         }
+        String rawDefinitionDataType = getRawDefinition().toLowerCase();
+        if (database instanceof H2Database && rawDefinitionDataType.matches("int\\([1-9]?[0-9]\\)")) {
+            int intParameter = Integer.valueOf(getParameters()[0].toString());
+            if(intParameter >= 1 && intParameter <= 3) {
+                return new DatabaseDataType("TINYINT");
+            } else if (intParameter > 3 && intParameter <= 10) {
+                return new DatabaseDataType("INTEGER");
+            } else if (intParameter > 10) {
+                return new DatabaseDataType("BIGINT");
+            }
+        }
         return super.toDatabaseDataType(database);
     }
 
--- a/liquibase-core/src/main/java/liquibase/datatype/core/SmallIntType.java
+++ b/liquibase-core/src/main/java/liquibase/datatype/core/SmallIntType.java
@@ -62,6 +62,10 @@ public class SmallIntType extends LiquibaseDataType {
             return new DatabaseDataType("SMALLINT"); //always smallint regardless of parameters passed
         }
 
+        if (database instanceof H2Database) {
+            return new DatabaseDataType("SMALLINT");
+        }
+
         return super.toDatabaseDataType(database);
     }
 
--- a/liquibase-core/src/main/java/liquibase/datatype/core/TinyIntType.java
+++ b/liquibase-core/src/main/java/liquibase/datatype/core/TinyIntType.java
@@ -42,6 +42,9 @@ public class TinyIntType  extends LiquibaseDataType {
         if (database instanceof OracleDatabase) {
             return new DatabaseDataType("NUMBER",3);
         }
+        if (database instanceof H2Database) {
+            return new DatabaseDataType("TINYINT");
+        }
         return super.toDatabaseDataType(database);
     }
 
--- a/liquibase-core/src/test/groovy/liquibase/datatype/core/BigIntTypeTest.groovy
+++ b/liquibase-core/src/test/groovy/liquibase/datatype/core/BigIntTypeTest.groovy
@@ -1,0 +1,36 @@
+package liquibase.datatype.core;
+
+import liquibase.database.core.*;
+import liquibase.datatype.DataTypeFactory;
+import spock.lang.Specification;
+import spock.lang.Unroll;
+
+class BigIntTypeTest extends Specification {
+
+    @Unroll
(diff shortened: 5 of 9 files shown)


nvoxland (changes requested):
I think it's fine to be checking the size here for int and converting to the correct actual types, since new h2 does not support precision like that.

But, the issue still exists where someone specifying `smallint(3)` or `bigint(15)` or whatever will still fail. So I think we need fixes to TinyIntType, SmallIntType, etc. that simply ignore the parameters passed since h2 doesn't support that. I don't think we need to do logic like "smallint(10) means bigint". 

Also, could you add a changeset to the h2 integration test changelog with the different int, smallint, etc. types for columns to ensure the final SQL is valid?

MalloD12:
> I think it's fine to be checking the size here for int and converting to the correct actual types, since new h2 does not support precision like that.
> 
> But, the issue still exists where someone specifying `smallint(3)` or `bigint(15)` or whatever will still fail. So I think we need fixes to TinyIntType, SmallIntType, etc. that simply ignore the parameters passed since h2 doesn't support that. I don't think we need to do logic like "smallint(10) means bigint".
> 
> Also, could you add a changeset to the h2 integration test changelog with the different int, smallint, etc. types for columns to ensure the final SQL is valid?

@nvoxland If I didn't get it wrong, what you are asking for here is being addressed as part of this [PR](https://github.com/liquibase/liquibase/pull/3303/files). Can you please have a look and confirm?

Thanks,
Daniel.

nvoxland:
@MalloD12 yes, that other PR picks up the other problems, I hadn't seen that one yet. Since it's touching similar code we could have done it as a single PR addressing both issues, but fine to keep them separate since you've made both already.

Both could use integration test items for them still, though.

MalloD12:
> @MalloD12 yes, that other PR picks up the other problems, I hadn't seen that one yet. Since it's touching similar code we could have done it as a single PR addressing both issues, but fine to keep them separate since you've made both already.
> 
> Both could use integration test items for them still, though.

if you are ok with that, I can still unify these two PRs in just a single one and then close one of them? I think some of them already have integration test created, but yeah some others don't.

nvoxland (approved):
Code review and test results:

Things to be aware of:
- H2 doesn't support arguments on any of the int datatypes, so this correctly removes them if they are set
- Includes tests to ensure they are as expected
- Only impacts h2

Things to worry about:
- Nothing

VeluchamySC:
Could you please include the fix for BOOLEAN Types as well?

Currently with H2 2.1.214, Liquibase is not able to translate 1 to TRUE (or) 0 to FALSE. With previous version of H2 1.4.200, Liquibase was able to translate it.

Steps to reproduce:
databaseChangeLog:
- changeSet:
    id: '1'
    author: author
    comment: 'initial schema'
    changes:
    - addColumn:
        tableName: order
        columns:
        - column:
            name: order_status
            type: VARCHAR(255)
            constraints:
                nullable: false
            value: NONE
    - update:
        columns:
        - column:
            name: order_status
            value: EXECUTED
        tableName: order
        where: executed = 1

Reason: liquibase.exception.DatabaseException: Values of types "BOOLEAN" and "INTEGER" are not comparable; SQL statement:
UPDATE PUBLIC.order SET order_status = 'EXECUTED' WHERE executed = 1 [90110-214] [Failed SQL: (90110) UPDATE PUBLIC.order SET order_status = 'EXECUTED' WHERE executed = 1]

nvoxland:
@VeluchamySC that won't be something we can fix. The `where` clause can be arbitrarily complex and so it's something we can try to parse and understand. All we do is pass it along as the update statement's where clause exactly as you have it. 

It looks like a change in what h2 will do for data type conversions between 1.x and 2.x. If you have just switched to h2 2.x you can update the statement to use `executed = true`. (You will need to add a validCheckSum tag to it knows you know what you are doing). If you still need to support both versions, you can use a changelog parameter like ${trueValue} which you set to 1 or true depending on the version of h2 you are using.

XDelphiGrl (approved):
H2 does not support precision in the SQL syntax, but it is valid to include precision in changeset datatype column definitions. This fix correctly "translates" integer datatypes specified with precision when SQL is generated for an H2 database, generating SQL with TINYINT, SMALLINT, INTEGER or BIGINT. Liquibase determines the correct type of integer following the H2 documentation here:

[H2 Data Types](http://www.h2database.com/html/datatypes.html)

* Comprehensive set of integration tests added. 

APPROVED

MalloD12:
Hi @TomBenjamins We are looking for getting another review on this PR. Once that happens we will be ready to merge this. We are expecting to have this close soon to include it in the next release,  which hopefully should be in about two weeks.

Thanks,
Daniel.-

suryaaki2 (approved):

kamil-j:
Hi,
Has this issue been fixed for raw sql changes as well? After migration from 4.4.0 to 4.18.0 scripts are failing on H2 with exactly the same message.

Example changeset:
`<sql>
            CREATE TABLE SampleTable
            (
                id          bigint(20) NOT NULL PRIMARY KEY AUTO_INCREMENT
            )
</sql>`

foxpluto:
Bump for the fix...

PR #3277, opened 2022-09-16
Remove snowflake unsupported logging from snapshot command (DAT-11291)

abrackx:
## Impact

## Description
Removes message `INFO This command might not yet capture Liquibase Pro additional object types on snowflake` when running snapshot against snowflake.

## Things to be aware of
Removed duplicate code from SnapshotCommandStep.

## Things to worry about
N/A

## Additional Context
N/A


liquibase-core/src/main/java/liquibase/command/core/InternalSnapshotCommandStep.java
```diff
@@ -112,19 +112,15 @@ public static String printSnapshot(CommandScope commandScope, CommandResults sna
...
 
     public static void logUnsupportedDatabase(Database database, Class callingClass) {
         if (LicenseServiceUtils.isProLicenseValid()) {
             if (!(database instanceof MSSQLDatabase
                     || database instanceof OracleDatabase
                     || database instanceof MySQLDatabase
                     || database instanceof DB2Database
-                    || database instanceof PostgresDatabase)) {
+                    || database instanceof PostgresDatabase
+                    || database.getShortName().equalsIgnoreCase("snowflake"))) {
```
StevenMassaro, line 123: Do we have to do this hacky because adding a dependency on the snowflake module would cause a circular dependency?

abrackx: Yep. Until that module is fully integrated into core we've got to use this hack. I talked it through with @nvoxland but I should probably also add a TODO. I'll do that now!

StevenMassaro (approved):

liquibase-core/src/main/java/liquibase/command/core/InternalSnapshotCommandStep.java
StevenMassaro: 🤮

nvoxland (approved):
It is the easiest way to make the change. We'd like something different, but it will take some thought on how to organize the code

PR #3279, opened 2022-09-16
DAT-11591: use complete path of resource when generating load data statements

StevenMassaro:
## Impact

## Description

Use the complete path of the newly created resources which represent the data to be loaded.

## Things to worry about

This could break some other integration I didn't think of.

### DIFF
--- a/liquibase-core/src/main/java/liquibase/diff/output/changelog/core/MissingDataExternalFileChangeGenerator.java
+++ b/liquibase-core/src/main/java/liquibase/diff/output/changelog/core/MissingDataExternalFileChangeGenerator.java
@@ -139,7 +139,7 @@ public class MissingDataExternalFileChangeGenerator extends MissingDataChangeGen
                 }
 
                 LoadDataChange change = new LoadDataChange();
-                change.setFile(fileName);
+                change.setFile(externalFileResource.getPath());
                 change.setEncoding(GlobalConfiguration.OUTPUT_FILE_ENCODING.getCurrentValue());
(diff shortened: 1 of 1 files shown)


liquibase-core/src/main/java/liquibase/diff/output/changelog/core/MissingDataExternalFileChangeGenerator.java
```diff
@@ -139,7 +139,7 @@ public Change[] fixMissing(DatabaseObject missingObject, DiffOutputControl outpu
                 }
 
                 LoadDataChange change = new LoadDataChange();
-                change.setFile(fileName);
+                change.setFile(externalFileResource.getPath());
```
nvoxland, line 142: What gets stored in here? Is it a searchpath relative path like `mydata/populated.csv` ? 

Or does it have a "direct" path like `s3:/bucket/mydata/populated.csv` or `/my-project/src/mydata/populated.csv`  ?

StevenMassaro: Relative, like your first example: `mydata/populated.csv` if your `--data-output-directory` were `mydata`

nvoxland: OK, sounds good. We don't want it to be absolute.

nvoxland (approved):

PR #3285, opened 2022-09-19
Rewrite Maven Integration to use Command Framework

nvoxland:
## Impact

## Description

We introduced the command framework to be a consistent business logic layer across all the integrations. Same command names, same arguments for commands, same extensibility pipeline, etc. With the command framework, each integration should just be a thin layer translating the user request that is specific to that integration to the generic command calls.

We rewrote the CLI to use the command framework a while ago, this applies the same pattern to Maven.

## Things to be aware of

- See discussion

## Things to worry about

- Need to preserve existing maven goals/arguments for backwards compatibility


filipelautert:
@abrackx I"m closing this PR, but just tagging you here as you're working with maven.

PR #3286, opened 2022-09-19
Create generic "Integration" layer

nvoxland:
## Impact

## Description

The command framework is our primary interface for running commands, but it is also fairly low level and requires coordination with a variety of other code when used as part of an "integration" like the CLI, maven, grade, spring, etc.

As we wrote the CLI to use the command framework we build that integration logic into the CLI code, but now that we are expanding the command framework to Maven and other integrations it would be nice to pull that code out of the CLI into something that can be re-used and be consistent across all our integrations. Make them easier to write and less error-prone.

## Things to be aware of

- Keeping this a separate PR from #3285 so far so we can more easily review the generic "maven command framework wrapper" changes from "refactor the CLI to extract common code" work. This will likely get merged into 3285 before 3285 goes to master.

## Things to worry about

- Too early to tell


nvoxland:
This is likely not even compiling yet and very much a work in progress.

The first big question I've hit in the refactoring is output handling. Specifically if it's time/right to replace the separate UIService.sendMessage / sendErrorMessage with Log.info / warning / error calls. In particular, there is code in the CLI that I want to pull to the Integration layer, but it's `System.err.println(...)` calls and we need to figure out how to refactor that.

We originally created the separate UIService from the log under the theory that the CLI and likely other interfaces will want to treat "messages that go to the user" separately from "messages that go in the log". Like maybe imagine a GUI where we have things show up in a dialog vs. going to a log.

But, what I've found is:
- The CLI is really the only integration so far with a concept of a separate "UI". Maven, gradle, spring, etc. all really just have the log
- We end up with a separate call to log.info() every time we call UIService.sendMessage because we want everything that shows on the screen to go to the log
- We have defined the info level as "messages that a user would normally expect to see in the CLI output"
- We would like "quiet" mode in the CLI which shows less info and the UIService doesn't have methods to differentiate that.

So my suggestion is:
- Make the log the only thing (with UIService still there as a wrapper to the log for compatibility reasons)
- The CLI simply configures a custom logger that sends info+ messages to stderr without the normal log formatting
- The CLI can have a "quiet" mode that makes it only send warning+ messages to stderr without the normal log formatting
- The CLI's log-level flag sets the log level and changes the formatting to include the normal log format
- In the future, UIService could be used for "progress indicator" type output which doesn't go to the log. So the class may never completely go away, but will be deprecated for now.

Then, all the code in the integration is just sending messages etc. to the correct log levels and it's up to the integration to set up the logger as it needs as it's interface.

Thoughts?

PR #3296, opened 2022-09-21
Updates generate-changelog to validate file exists via PathHandlerFactory (DAT-11559)

abrackx:
## Impact

## Description
When using the s3 extension liquibase will overwrite an existing changelog file in s3 instead of erroring due to the file existing. This is because it is looking locally using `File` not trying to pull the file as a `Resource`. This change updates the generate-changelog command to use PathHandlerFactory to find the resource and determine whether it exists. 

## Things to be aware of
I removed the logic where once we've determined the file doesn't exist, we then attempt to delete it. Resource does not include a delete method.

## Things to worry about
See above 

## Additional Context
N/A


liquibase-core/src/main/java/liquibase/integration/commandline/Main.java
```diff
@@ -1577,23 +1574,11 @@ protected void doMigration() throws Exception {
                     //will output to stdout:
                     currentChangeLogFile = "";
                 }
-
-                File file = new File(currentChangeLogFile);
+                final PathHandlerFactory pathHandlerFactory = Scope.getCurrentScope().getSingleton(PathHandlerFactory.class);
+                Resource file = pathHandlerFactory.getResource(currentChangeLogFile);
```
StevenMassaro, line 1578: Should we also be checking the resource accessor here?

abrackx: I don't _think_ so because I could see the scenario where we try to pull a file in s3, it doesn't exist so it falls back to another resourceaccessor, then on windows it sees a file with multiple colons in the uri and throws that unexpected ":" error we've seen. Does that make sense? 🤔

StevenMassaro: Definitely. But aren't changelogs the one place that we want to be checking the resource accessor? I admit that I am not familiar with the code you're changing, so I don't know if its relevant or not.

nvoxland: No, we should not be checking the resource accessor here. It's only reading changelog files that should be using the resource accessor. Because this is mainly an output operation, this only takes a direct file path.

Thinking through the generateChangelog/diffChangeLog impact when originally wanting to expanding resourceAccessor usage was a big part of what made me say "no, let's not change our existing logic to start taking the searchPath into account without a lot more thought because it will likely be problematic"

nvoxland (approved):

PR #3301, opened 2022-09-23
Add 'ifExists' parameter to the dropView statement

gzsombor:
## Impact

## Description

Add "ifExists" flag to the "dropView" statement, if the attribute is set to true, then the migration can progress further even when the view is not there.   

## Things to be aware of

The change is backward compatible, so if that attribute is not specified, the same 'drop view X' statement is generated as before.

## Things to worry about

None

## Additional Context

-

### DIFF
--- a/liquibase-core/src/main/java/liquibase/change/core/CreateViewChange.java
+++ b/liquibase-core/src/main/java/liquibase/change/core/CreateViewChange.java
@@ -260,7 +260,7 @@ public class CreateViewChange extends AbstractChange {
 		}
 
         if (!supportsReplaceIfExistsOption(database) && replaceIfExists) {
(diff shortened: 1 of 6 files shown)


suryaaki2 (approved):

XDelphiGrl (approved):
This PR adds support for the `ifExists` attribute to the `dropView` changetype. This allows users to tell Liquibase to check for a view's existence prior to blindly attempting to drop the view. Note that Oracle, and therefore Liquibase, does not support `IF EXISTS` syntax for views.
* New integration test added; run against all DBMS platforms _except_ Oracle.
* XSD change --> requires a major version bump at time of release.

APPROVED

awhitford-cip:
Yet another excellent feature that is sadly not documented: https://docs.liquibase.com/change-types/drop-view.html

PR #3302, opened 2022-09-23
Check for existence before adding resources directory in registerChangeLog Mojo

wwillard7800:


liquibase-core/src/main/java/liquibase/diff/output/changelog/core/MissingDataExternalFileChangeGenerator.java
```diff
@@ -139,7 +139,7 @@ public Change[] fixMissing(DatabaseObject missingObject, DiffOutputControl outpu
                 }
 
                 LoadDataChange change = new LoadDataChange();
-                change.setFile(fileName);
+                change.setFile(externalFileResource.getPath());
```
nvoxland, line 142: Does this have to be changed?
It's just change.setFile(filename) in master, and the getPath() behavior is currently a bit undefined so I prefer to not rely on it if we don't have to

StevenMassaro: Yes this does have to be changed. The dataOutputDirectory is part of the externalFileResource.getPath, but not part of fileName.

suryaaki2 (approved):

PR #3308, opened 2022-09-26
Refactored release logic to use liquibase-sdk

ap-liquibase:
## Impact

## Description

Updates the logic used for creating release artifacts to:
- Pull current versions of external components at "create artifact time" vs. expecting the .tar.gz to be correct
- Use logic that was moved to liquibase-sdk for building final files

Fixes DAT-11427

## Things to be aware of

- Only needs to impact release logic
- Prepares us to change the upstream build logic in a separate PR

## Things to worry about

- We can see the final artifacts in a test draft release and manually check them, but the real publish of them isn't tested until release day


nvoxland:
For reviewing this PR, I ended up writing proposed changes vs. describing them. You can see them at https://github.com/liquibase/liquibase/pull/3308/commits/5810f8b42bdb4569a235d69d2dcce3c7f0e7261a and I see it as a discussion point more than "how it has to be".

Changes:
- Use liquibase-sdk:download-snapshot-artifacts vs. action-download-artifact
- Removed some lines that seemed unneeded now
- Separated the "save the results" from "build intallers" partly because of the attempted re-use of create-release-artifact which can't run on mac and partly to prepare for other steps later like "make RPMs" etc.
- Some wording/naming changes
- Added liquibase-sdk to the base pom.xml file

I ran the action to see what works and doesn't. It failed now because the create-release-artifact call I do in the "finish" line to sign the new filese and  build the liquibase-additional.zip fails due to not finding any snapshot files to reversion. 
That is easily fixed, but I didn't want to bother with that if we ended up wanting to do things completely differently than I have here.

ap-liquibase:
This PR was left open originally to facilitate discussion with Nathan about the use of sdk. However it is deprecated now since   we will use Github Packages as a way to share build artifacts with downstream builds.

nvoxland:
I think all the logic we have in here for improving how the release process works is just as valid even if we're getting snapshot builds from github packages vs. off the builds?

So not sure it should just be closed for that reason?

PR #3318, opened 2022-09-28
Fixes problem of view snapshot evaluation as metadata is not being located when it is created with objectQuotingStrategy="QUOTE_ALL_OBJECTS" and evaluated without it

### LINKED ISSUE #2659: Precondition tableExists crashes on H2 with objectQuotingStrategy=LEGACY, if a camelCaseView already exists.
## Environment

**Liquibase Version**: 4.9.0

**Liquibase Extension(s) & Version**:  -

**Database Vendor & Version**: H2 Database, 2.1.210

**Operating System Type & Version**: Microsoft Windows [Version 10.0.19044.1586]

## Description

Preconditions (e.g. `tableExists`) crash on H2 database, if the db schema already contains a view, whose name uses camelCase (i.e. it was created with `objectQuotingStrategy="QUOTE_ALL_OBJECTS"` before), but the changeSet that contains the precondition (or the complete changelog file) uses `objectQuotingStrategy="LEGACY"`.

Here comes the example `changelog.xml`:
```
                                       http://www.liquibase.org/xml/ns/dbchangelog/dbchangelog-4.9.xsd"
                   objectQuotingStrategy="LEGACY">
    
    <changeSet id="1 Create MyView" author="mamue" objectQuotingStrategy="QUOTE_ALL_OBJECTS">
        <createView  viewName="MyView" replaceIfExists="true">select 1</createView>
    </changeSet>

    <changeSet id="2 Drop NOT_EXISTENT_TABLE" author="mamue">
        <preConditions onFail="MARK_RAN"><tableExists tableName="NOT_EXISTENT_TABLE"/></preConditions>
        <dropTable tableName="NOT_EXISTENT_TABLE"/>
    </changeSet>

</databaseChangeLog>
```

The first changeSet is executed without problems.
The evaluation of the precondition of the second changeSet crashes with this error message: `Unexpected error running Liquibase: Expected single row from liquibase.statement.core.GetViewDefinitionStatement@9255c05 but got 0`

I used the following `liquibase.properties` file:
```
changeLogFile: changelog.xml
driver: org.h2.Driver
url: jdbc:h2:mem:
username: sa
password: 
liquibase.secureParsing: false
```

## Steps To Reproduce

* Download `liquibase-4.9.0.zip` and extract it to a new folder.
* Place the `changelog.xml` and `liquibase.properties` files mentioned above in the root folder of the extracted archive.
* Start the `liquibase.bat` with the argument `update`.
* Watch the output. It should be something like this:

> Running Changeset: changelog.xml::1 Create MyView::mamue
> Running Changeset: changelog.xml::2 Drop NOT_EXISTENT_TABLE::mamue
> 
> Unexpected error running Liquibase: Expected single row from liquibase.statement.core.GetViewDefinitionStatement@9255c05 but got 0
> 
> For more information, please use the --log-level flag

## Actual Behavior
The evaluation of the precondition of the second changeSet of the example `changelog.xml` runs into the already mentioned error. The execution of the `changelog.xml` is stopped prematurely.

## Expected/Desired Behavior
The execution of the precondition of the second changeSet detects that the table  `NOT_EXISTENT_TABLE`  does not exist and therefore skips the execution of the action of the second changeSet and marks that changeSet RUN in the table `DATABASECHANGELOG`. The execution of the `changelog.xml` finishes without error.

## Additional Context
I found that also other preconditions fail, that check for the existence of some database object. The error message was the same. The error also happens with file based H2 databases.

filipelautert:
## Impact

## Description
When a view is created using objectQuotingStrategy="QUOTE_ALL_OBJECTS" , the name is saved as defined (ie: "MyView" ) - otherwise it is all upcase (MYVIEW) . As the default strategy for h2 is LEGACY, it fails on subsequent changesets unless QUOTE_ALL_OBJECTS is specified (tested on H2 and postgres: H2 failed, postgresql not).
It fails using the "update" command - if we execute the command "snapshot"  , the strategy is already set by default to QUOTE_ALL_OBJECTS so it works.

This PR changes the strategy to QUOTE_ALL_OBJECTS, snapshots the views and then sets it back to what it was before (as it is done in some others parts of the code).
 
## Things to be aware of
- fixes #2659 
-

### DIFF
--- a/liquibase-core/src/main/java/liquibase/snapshot/jvm/ViewSnapshotGenerator.java
+++ b/liquibase-core/src/main/java/liquibase/snapshot/jvm/ViewSnapshotGenerator.java
@@ -4,6 +4,7 @@ import liquibase.CatalogAndSchema;
 import liquibase.Scope;
 import liquibase.database.AbstractJdbcDatabase;
 import liquibase.database.Database;
+import liquibase.database.ObjectQuotingStrategy;
 import liquibase.database.core.InformixDatabase;
 import liquibase.database.core.MariaDBDatabase;
 import liquibase.database.core.MySQLDatabase;
@@ -77,7 +78,9 @@ public class ViewSnapshotGenerator extends JdbcSnapshotGenerator {
                 CatalogAndSchema schemaFromJdbcInfo = ((AbstractJdbcDatabase) database).getSchemaFromJdbcInfo(rawCatalogName, rawSchemaName);
                 view.setSchema(new Schema(schemaFromJdbcInfo.getCatalogName(), schemaFromJdbcInfo.getSchemaName()));
 
+                ObjectQuotingStrategy originalQuotingStrategy = database.getObjectQuotingStrategy();
                 try {
+                    database.setObjectQuotingStrategy(ObjectQuotingStrategy.QUOTE_ALL_OBJECTS);
                     String definition = database.getViewDefinition(schemaFromJdbcInfo, view.getName());
 
                     if (definition.startsWith("FULL_DEFINITION: ")) {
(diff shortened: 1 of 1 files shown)


filipelautert:
Moving to pr 3319

PR #3323, opened 2022-09-29
Dat 9372

dyadyushko:
## Impact

## Description
isn't ready

## Things to be aware of

## Things to worry about

- List unknowns you have

### DIFF
--- a/liquibase-core/src/main/java/liquibase/exception/MigrationFailedException.java
+++ b/liquibase-core/src/main/java/liquibase/exception/MigrationFailedException.java
@@ -2,10 +2,15 @@ package liquibase.exception;
 
 import liquibase.changelog.ChangeSet;
 
+import java.util.Collections;
+import java.util.LinkedList;
+import java.util.List;
+
 public class MigrationFailedException extends LiquibaseException {
 
     private static final long serialVersionUID = 1L;
     private final String failedChangeSetName;
+    private List<ChangeSet> failedChangeSets = null;
     
     public MigrationFailedException() {
         failedChangeSetName = "(unknown)";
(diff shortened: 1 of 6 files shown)


liquibase-core/src/main/java/liquibase/integration/commandline/Main.java
```diff
@@ -1842,7 +1845,17 @@ protected void doMigration() throws Exception {
...
             try {
                 if (COMMANDS.UPDATE.equalsIgnoreCase(command)) {
-                    liquibase.update(new Contexts(contexts), new LabelExpression(getLabelFilter()));
+                    try {
+                        liquibase.update(new Contexts(contexts), new LabelExpression(getLabelFilter()));
+                    } catch (MigrationFailedException e) {
+                        if (!e.containsFailedChangeset() || !rollbackOnError) {
+                            throw e;
+                        }
+                        LicenseServiceUtils.checkProLicenseAndThrowException(command, ROLLBACK_ON_ERROR.getName());
```
StevenMassaro, line 1854: You probably don't need this license check here, since there is a license check inside the rollback one changeset command in the pro repo.

liquibase-core/src/main/java/liquibase/exception/MigrationFailedException.java
```diff
@@ -14,17 +19,40 @@ public MigrationFailedException() {
...
         this.failedChangeSetName = failedChangeSet.toString(false);
+        this.failedChangeSets = filterFailedChangeSets(failedChangeSet);
+    }
+
+    public boolean containsFailedChangeset() {
+        return !(failedChangeSets == null || failedChangeSets.isEmpty());
+    }
+
+    public List<ChangeSet> getFailedChangeSets() {
+        return new LinkedList<>(failedChangeSets);
```
StevenMassaro, line 43: Any reason you're creating a new `LinkedList` object here? I would think you could return `failedChangeSets` directly.

liquibase-core/src/main/java/liquibase/integration/commandline/Main.java
dyadyushko: True!

liquibase-core/src/main/java/liquibase/exception/MigrationFailedException.java
dyadyushko: I was thinking of preventing devs from mutating the object outside of the class. I would've used `Collections.unmodifiableList` but it prevents us from reversing the list in the function `rollbackChangeSets` down below.

PR #3339, opened 2022-10-04
Use PathHandler for writing log files, which allows using S3 and other extensions (DAT-11515)

StevenMassaro:
## Impact

## Description

Create log files using the `PathHandler` interfaces. Add support for different `OpenOptions` for resources, like appending and truncating existing files.

## Things to be aware of

- Liquibase erroneously closed the `Handler` in `LiquibaseCommandLine`. This is actually unnecessary as the Java log framework will handle closing it at the appropriate time. Additionally, Liquibase was closing the handler too early, which caused an exception to be thrown by the logging framework itself: `java.lang.IllegalStateException: Current state = FLUSHED, new state = CODING`
- Not all implementations of resource will support appending to existing files. Implementations which do not support appending will throw an exception if the file already exists and append is set to true.


liquibase-cli/src/main/java/liquibase/integration/commandline/LiquibaseCommandLine.java
```diff
@@ -644,8 +644,9 @@ private void configureLogging(Level logLevel, File logFile) throws IOException {
 
         if (logFile != null) {
             if (fileHandler == null) {
-                fileHandler = new FileHandler(logFile.getAbsolutePath(), true);
-                fileHandler.setFormatter(new SimpleFormatter());
+                final LogFileFactory logFileFactory = Scope.getCurrentScope().getSingleton(LogFileFactory.class);
```
nvoxland, line 647: Why do we need to be defining a new log-specific resource type? I get the limitation we are working around with s3, but it seems wrong to be singling out log files when this will be a general problem for other files and also how even **having** a log file is a not something that is assumed. There are lot of cases where there is no log file available.

Instead of introducing these new interfaces, could we update the ResourceAccessor/PathHandler/Resource APIs' openStream type methods to specify whether it should be an append or a replace operation?

nvoxland (changes requested):

liquibase-cli/src/main/java/liquibase/integration/commandline/LiquibaseCommandLine.java
StevenMassaro: I added this log-specific resource type, because we want to append to log files by default. That is how Liquibase currently works. However, for S3, we're going to throw an exception if the log file the user is pointing to already exists in S3. So we need the ability to behave slightly differently for S3 than in all other circumstances.

> Instead of introducing these new interfaces, could we update the ResourceAccessor/PathHandler/Resource APIs' openStream type methods to specify whether it should be an append or a replace operation?

This would not allow us to have an S3 specific implementation for log file writing.

StevenMassaro: An alternative option for resolving this would be to remove the log file plugin, and instead change the S3 code to throw an exception if the user tries to use the append open option and the file already exists. That would be a relatively transparent change as far as the open source code is concerned. What do you think of that?

liquibase-core/src/main/java/liquibase/integration/spring/SpringResource.java
```diff
@@ -59,10 +60,13 @@ public InputStream openInputStream() throws IOException {
     }
 
     @Override
-    public OutputStream openOutputStream(boolean createIfNeeded) throws IOException {
-        if (!resource.exists() && !createIfNeeded) {
+    public OutputStream openOutputStream(OpenOptions openOptions) throws IOException {
+        if (!resource.exists() && !openOptions.isCreateIfNeeded()) {
```
StevenMassaro, line 64: This needs a null check on openOptions

liquibase-core/src/main/java/liquibase/resource/PathHandlerFactory.java
```diff
@@ -75,17 +74,30 @@ public Resource getResource(String resourcePath) throws IOException {
...
+     * Returns the outputStream from {@link #getResource(String)}, using settings from the passed {@link OpenOptions}.
+     *
+     * @return null if resourcePath does not exist and {@link OpenOptions#isCreateIfNeeded()} is false
+     * @throws IOException if there is an error opening the stream
+     */
+    public OutputStream openResourceOutputStream(String resourcePath, OpenOptions openOptions) throws IOException {
         Resource resource = getResource(resourcePath);
         if (!resource.exists()) {
-            if (createIfNotExists) {
+            if (openOptions.isCreateIfNeeded()) {
```
StevenMassaro, line 94: This needs a null check on openOptions

liquibase-core/src/main/java/liquibase/integration/spring/SpringResource.java
StevenMassaro: This is identified as not null in the javadocs, and adding null checks everywhere it is used makes the code pretty complicated, so I'm not going to do this.

liquibase-core/src/main/java/liquibase/resource/PathHandlerFactory.java
StevenMassaro: This is identified as not null in the javadocs, and adding null checks everywhere it is used makes the code pretty complicated, so I'm not going to do this.

nvoxland (approved):

Issue #1259, opened 2020-07-10
Logging start of change set file


Roland112263:
## Environment 

**Liquibase Version**: latest

**Liquibase Integration & Version**: maven liquibase-core

**Liquibase Extension(s) & Version**: latest

**Database Vendor & Version**: org.testcontainers PostgreSql 1.10.6

**Operating System Type & Version**: Windows 10

## Description

INFO level does not have information about start of execution of sql script.
This information contains DEBUG level, but besides there is a lot information that does not need

## Steps To Reproduce

1) Run any sql scripts in liqubase with log level INFO

## Actual Behavior
INFO level contains the information below:
2020-07-06 12:51:54.811 INFO 14057 — [ restartedMain] liquibase: classpath:/db/changelog/db.changelog-master.yaml: db/changelog/country/0000-init.sql::raw::includeAll: Custom SQL executed
2020-07-06 12:51:54.831 INFO 14057 — [ restartedMain] liquibase: classpath:/db/changelog/db.changelog-master.yaml: db/changelog/country/0000-init.sql::raw::includeAll: ChangeSet

## Expected/Desired Behavior
INFO level should contains the information below:
**2020-07-06 12:51:54.800 INFO 14057 — [ restartedMain] liquibase: classpath:/db/changelog/db.changelog-master.yaml: db/changelog/country/0000-init.sql::raw::includeAll: Executing Custom SQL**
2020-07-06 12:51:54.811 INFO 14057 — [ restartedMain] liquibase: classpath:/db/changelog/db.changelog-master.yaml: db/changelog/country/0000-init.sql::raw::includeAll: Custom SQL executed
2020-07-06 12:51:54.831 INFO 14057 — [ restartedMain] liquibase: classpath:/db/changelog/db.changelog-master.yaml: db/changelog/country/0000-init.sql::raw::includeAll: ChangeSet

molivasdat:
Thanks @Roland112263 for writing this issue.  We will add it to the issues list to process.

filipelautert:
We're currently triaging our backlog to ensure we can focus on the issues that matter most. With a large number of open issues, it becomes difficult to give each one the attention it deserves.

This issue has been closed due to inactivity — there have been no updates for over 30 days. If this is still relevant to you, please feel free to reopen it or create a new issue with updated context. We appreciate your contribution!

PR #3350, opened 2022-10-08
#1259: Logging start of change set file

skrivenko:
Changed log message from fine to info for the following text: "[INFO] Executing Statement: insert into person (name) values ('John')". Then it will show the start date of executing statement for custom sql.

Tags: hacktoberfest, firstPR

Examples:

Running Changeset: example-changelog.xml::8::your.name
[INFO] Executing Statement: insert into person (name) values ('John')
[INFO] Custom SQL executed
[INFO] ChangeSet example-changelog.xml::8::your.name ran successfully in 17ms
....
Running Changeset: example-changelog.xml::9::your.name
[INFO] Executing Statement: liquibase.statement.core.AddColumnStatement@52a7928a
[INFO] Columns middleName(varchar(20)) added to person
[INFO] ChangeSet example-changelog.xml::9::your.name ran successfully in 17ms

### DIFF
--- a/liquibase-core/src/main/java/liquibase/database/AbstractJdbcDatabase.java
+++ b/liquibase-core/src/main/java/liquibase/database/AbstractJdbcDatabase.java
@@ -1265,7 +1265,8 @@ public abstract class AbstractJdbcDatabase implements Database {
             if (statement.skipOnUnsupported() && !SqlGeneratorFactory.getInstance().supports(statement, this)) {
                 continue;
             }
-            Scope.getCurrentScope().getLog(getClass()).info("Executing Statement: " + statement);
+            Scope.getCurrentScope().getLog(getClass()).info("Executing Statement: " + statement.getClass());
+            Scope.getCurrentScope().getLog(getClass()).fine(statement.toString());
             try {
                 Scope.getCurrentScope().getSingleton(ExecutorService.class).getExecutor("jdbc", this).execute(statement, sqlVisitors);
             } catch (DatabaseException e) {
--- a/liquibase-core/src/main/java/liquibase/database/AbstractJdbcDatabase.java
+++ b/liquibase-core/src/main/java/liquibase/database/AbstractJdbcDatabase.java
@@ -1265,7 +1265,7 @@ public abstract class AbstractJdbcDatabase implements Database {
             if (statement.skipOnUnsupported() && !SqlGeneratorFactory.getInstance().supports(statement, this)) {
                 continue;
             }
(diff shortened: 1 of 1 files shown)


nvoxland:
Thanks for the PR, @skrivenko !

Your overall goal is to make the reporting of SQL shown at a higher log level, correct? Or is there more you are looking to address?

There is currently logging at the JdbcExecutor level based on the `liquibase.sql.logLevel` flag that should be allowing you to configure logging to be at info if you'd like. Is that not catching some of the statements you are looking for?

We don't log SQL at info level by default because it gets chatty and there are people who worry about the security implications of having their SQL sent to potentially insecure logs. So that flag is there to let people choose the level to see it at.

skrivenko:
Hello, thank you for you review! I thought that it was the issue https://github.com/liquibase/liquibase/issues/1259
But I understand your comment about sql and its values.
Would it be ok if I return back the debug level but I will add another info level like this:

[INFO] Executing: liquibase.statement.core.AddColumnStatement.class 

with no sql exposed? 
Thanks!

skrivenko:
For issue #3350

nvoxland (changes requested):
Sorry for the slow reply. 

I think the overall issue with the PR is that you're looking to log the pre-execution of a Statement vs the Change, and there isn't a lot of good descriptive methods on the Statement.

The breakdown between Change and Statement is that Change objects are the "higher level" command which may call one or more Statements, where statements more closely map to a single SQL statement. Sometimes it's a 1:1 mapping between change and statement, but not always.

For the original issue, they're wanting a "before start" and "after completion' for the Change object. The exiting INFO call is at ChangeSet line 684 and they're looking for a similar method about the Change on like line 682 or so.

We currently have a "we are going to start to execute a Statement FINE level log message which is what you are changing in the PR, but that is different. 

We don't want to change the Statement to fine level, since that is more of a detail of the already-INFO level logged Change execution. The current statement.toString() that's being logged isn't as good as your changed `.getName()` call since toString isn't well implemented on the statement so the message is a bit odd. But we do need to keep that one at FINE level and not (like this PR does) have it as info now.

If you'd like, you could update this PR to have the getName() fixed "executing statement" message still at FINE and a corresponding FINE message after the statement finishes executing of "executed statement + statement.getClass().getName()".

And then also add the log message to the end of ChangeSet.execute() asked for in the original issue.

MalloD12:
Hi @skrivenko,

I just came across this PR and I would like to know whether you are keen to keep working on it addressing Nathan's suggested changes or not. 

Today,  we have created an internal ticket to address #1259 which we plan to be working on at some point in the near future. 

Thanks,
Daniel.

PR #3355, opened 2022-10-10
Fix include/includeAll to correct track relativeToChangelog=true paths using ./ or ../

nvoxland:
## Impact

## Description

Changes in 4.17.0 made a changelog like `<include file="./changelog.workspace.sql" relativeToChangelogFile="true"/>` incorrectly include the `./` in the databasechangelog table and `status --verbose` among other places.

The correct fix is in the Resource's resolveSibling() function, but this is a temporary work around until we have a better solution in there.

## Things to be aware of

- Only impacts included changesets

## Things to worry about

- Nothing


borjapenman:
Using ../ when relativeToChangelog=true seems quite common use case. Is thsis going to be merged for the next release?

filipelautert (approved):
Fixed tests, so there is test coverage.

suryaaki2 (approved):

XDelphiGrl (approved):
PR addresses a regression in the path shown some output messages and the value stored in DATABASECHANGELOG tracking table when `include` or `includeAll` changesets specify `relativeToChangelog=true`. 
* New unit test added.

APPROVED

filipelautert:
> This impacts existing migrations so this is a breaking change wouldn't it be ?
> 
> On migration , for a sql file, the filename will include the path separator in the changelog table With this update that name no longer matches on a deployment so the existing migration attempts to run again checksums are equal , names are different

Well, it went back to as it was before 4.17.0 . So if you already faced this problem with 4.17 you need to run clearChecksums before upgrading.

PR #3363, opened 2022-10-12
Add skipOnFileExists maven property for skipping plugin execution when a file exists

mensinda:
## Impact

## Description

This PR adds a new property for the liquibase-maven-plugin to skip the plugin execution if the specified file exists. This feature is useful for certain edge cases, where the already existing `skip` property is not enough.

## Things to be aware of

- Impacts on the codebase: Minimal (just one more property and minimal logic)

## Things to worry about

None


nvoxland:
Thanks for the feature PR.

Could you explain your use case a little bit? I'm curious how you are using this. Largely because we are soon going to better unify the maven, the CLI, and other integrations to have as much of a common set of functionality as possible, and so wondering if it's something that really just need so to remain in maven or would apply well everywhere.

mensinda:
We basically need it to work around maven limitations.

Specifically, we have a company wide parent POM where we use the liquibase plugin for automatically generating test H2 DBs in the `target` dir. However, we don't want to do that every `mvn verify` run, so we also generate a `test.NO_REGEN.db` file. As long as this file exists, we don't want to regenerate the test DB. We already use the existing `skip` property for something else, so we also can't use hacks like this:

```xml
        <profile>
            <id>create-local-it-test-lq_cl-db-DISABLER</id>
            <activation>
                <file>
                    <exists>${basedir}/target/test.NO_REGEN.db</exists>
                </file>
            </activation>
            <properties>
                <skip-liquibase-test-db>true</skip-liquibase-test-db>
            </properties>
        </profile>

        <profile>
            <id>create-local-it-test-lq_cl-db-ENABLER</id>
            <activation>
                <file>
                    <missing>${basedir}/target/test.NO_REGEN.db</missing>
                </file>
            </activation>
            <properties>
                <skip-liquibase-test-db>false</skip-liquibase-test-db>
            </properties>
        </profile>
```

For the CLI this can be solved with a trivial `if [ ! -f "<file path here>" ]` in bash, but good luck trying to do complex conditional plugin activation in maven. Thus, this would be something maven specific IMHO.

mensinda:
Hi, is there a chance for this change to be merged? Would a more general solution be preferred? Is something missing?

nvoxland:
Sorry for the delay, @mensinda. I was holding the PR for a bit as some possibly related refactoring work was being considered. Moving it along in the review process now. 

Mostly what I was wondering was whether it should be a more generic fix across both the CLI and maven (and other integrations) or good to keep just in maven. We're working on bringing functionality to be identical between maven and the CLI, but there will always be at least some difference. As part of the full review process, we can think more on that and how it fits into the overall maven refactoring

mensinda:
No problem. Reading CLI, I (personally) don't really see the point for adding skip functionality, since thats what `if` is for in bash, etc. (see my comment from above). As for making it more generic:
- Add support for a `!` prefix for inverting the condition
- Add support for multiple files with `&&` and `||` evaluation + brackets

However, with each added feature, the skip expression will more and more resemble a scripting language, which probably wouldn't be desirable :)

My questions would thus be:
- Should I make it more generic?
  - If yes, what features do you want (or rather how complex)?
- Do you want it in the CLI regardles?

mensinda:
Is there a way to see why the "Run Functional Tests" CI is failing (so that I can try to fix it)? I get a 404 error when trying to view the details.

XDelphiGrl:
> Is there a way to see why the "Run Functional Tests" CI is failing (so that I can try to fix it)? I get a 404 error when trying to view the details.

@mensinda, hello - No, the functional tests are internal Liquibase-access only. The tests are passing last time they ran. My guess is there were environmental or "expected" test validation failures that needed to be updated to match the latest core code.

XDelphiGrl (approved):
This PR introduces a Liquibase Maven Plugin property to skip executing Liquibase if a particular file is present. The file name is defined in the property `skipOnFileExists`. When the file exists, Liquibase logs a warning that Liquibase did not run due to the presence of the file. When the file does not exist, Liquibase logs a warning that Liquibase execution continued because the file does not exist. Leaving `skipOnFileExists` undefined results in Liquibase executing for every applicable Maven execution.
* This is Liquibase Maven Plugin specific and is not available in the CLI.

APPROVED

suryaaki2 (approved):

PR #3366, opened 2022-10-12
Fix create index on Postgresql and MSSQL using wrong column names when  generating a changelog

### LINKED ISSUE #2953: Conditional index wrongly generated with CLI generate-changelog command
## Environment

**Liquibase Version**:
version 4.10.0 #2501 built at 2022-05-04 14:27+0000

CLI

PostgreSQL 11.15 on x86_64-apple-darwin21.3.0, compiled by Apple clang version 13.0.0 (clang-1300.0.29.30), 64-bit 

Mac

**Infrastructure Type/Provider**: <AWC, GCS, Azure, VM, etc>
Local

## Description

A condition index is wrongly converted when using:
```
liquibase --changelog-file=dbchangelog.xml generate-changelog
```

We created an index with following SQL:
```
CREATE UNIQUE INDEX db_tbl_uc01 ON s.abc (x, y, z) WHERE (b = TRUE);
```

This results in the following xml export:
```
 <createIndex indexName="db_tbl_uc01" schemaName="s" tableName="abc" unique="true">
            <column computed="true" name="(b = true)"/>
            <column computed="true" name="(b = true)"/>
            <column computed="true" name="(b = true)"/>
   </createIndex>
```

So the columns x, y, z are missing as columns

filipelautert:
## Impact

## Description

Postgresql and MSSQL server support conditional indexes (a WHERE when creating the index).  When a conditional index is exported by generate-changelog, liquibase is replacing the columns by the where condition. This fix makes sure that the columns are corrected exported. 
As discussed with Nathan, implementing the "where" condition would be too specific for those databases and is currently not supported by Liquibase, thus it's being ignored.

This fix makes sure that Postgresql and MSSQL use the column name instead of FILTER_CONDITION field when snapshoting indexes (as both support partial indexes, FILTER_CONDITION brings the filter and not the function value)

- Fixes #2953 
- Functional test fix: https://github.com/liquibase/liquibase-pro-tests/pull/502

## Things to be aware of

- where condition is being ignored.

## Things to worry about

- None

## Additional Context

- None


XDelphiGrl (approved):
This PR fixes a bug that caused indexes created with `WHERE` clauses to use the `WHERE` clause instead of the column name in a generated `createIndex` changeset. The bug, and fix, are specific to Postgres and SQL Server. 
* New integration test added.
* Internal functional tests to change test to match new/correct behavior.

APPROVED

Note: The wording in the integration test assertion could be updated, but I will not block the PR from approval based on that. If you can make the change before merging, @filipelautert, that would be very helpful.

PR #3367, opened 2022-10-12
Added new modifyChangeSets tag to support specification of runWith attribute for all change sets in include or includeAll DAT-7766

wwillard7800:
Added new modifyChangeSets tag to support specification of runWith attribute for all change sets in include or includeAll


liquibase-core/src/main/resources/www.liquibase.org/xml/ns/dbchangelog/dbchangelog-4.17.xsd
```diff
@@ -81,6 +81,36 @@
             <xsd:enumeration value="QUOTE_ONLY_RESERVED_WORDS"/>
         </xsd:restriction>
     </xsd:simpleType>
+    <!-- include -->
```
StevenMassaro, line 84: I am surprised that you modified the 4.17 xsd. Isn't this one already released, and so shouldn't be modified?

suryaaki2: @wwillard7800 Should this xsd change be only in latest.xsd?

wwillard7800: Yes, that is correct.  I will revert that change.

suryaaki2 (approved):

liquibase-core/src/main/java/liquibase/changelog/DatabaseChangeLog.java
```diff
@@ -48,6 +46,7 @@ public class DatabaseChangeLog implements Comparable<DatabaseChangeLog>, Conditi
     private String physicalFilePath;
     private String logicalFilePath;
     private String changeLogId;
+    private String includeRunWith;
```
nvoxland, line 49: I don't think we can be adding an object-level field for includeRunWith. It happens to work with how we tend to call methods on this object, but the scope seems wrong on it. It's not that there's a DatabseChangeLog-wide "includeRunWith", it's really just a scratch storage variable that should be scoped to where it actually applies. Otherwise it's going to be too easy to have bugs when people are calling include() in their own unique ways without knowing to somehow clean this up. Or as we evolve our usage of this class over time it may burn us.

Since we have a new concept of modifyChangeSets we probably need a class to contain that. It will just have runWith as a field on it now, but will allow us to keep adding more flags without breaking API compatibility. Updating the include() function to take the applicable ModifyChangeSets object (but keeping a deprecated vesrsion that doesn't have the parameter for API compatibility) makes it so that function doesn't have to rely on this object-level field.

But there is the question of how to store it on line 399 where we'd be creating it when working on one ParsedNode to where we need to get it on line 404 when being called for a different ParsedNode....

We could add a scratch Map to handleChildNode's signature and on line 399 do a nodeScratch.put("modifyChangeSets", new ModifyChangeSets(...)) and then on line 402 after all the "handleChildNode" call nodeScratch.remove("modifyChangeSets"). That's a bit ugly, but is at least keeping it scoped a bit cleaner? 

Ideally we'd be able to walk up the parent node tree to look for attached/parsed versions of the objects. But that is probably a bigger refactoring than we want right now.

Any other ideas?

liquibase-core/src/main/java/liquibase/changelog/ChangeLogInclude.java
```diff
@@ -9,12 +9,13 @@
 
 public class ChangeLogInclude extends AbstractLiquibaseSerializable implements ChangeLogChild {
     private String file;
+    private String runWith;
```
nvoxland, line 12: This doesn't feel like a change we should make, because the "include" doesn't have a runWith. Same with ChangeLogIncludeAll below. 

But looking at how they is used, I'm not finding any usages of them at all? Is this dead code that you thought might be needed to change and didn't undo after you found it wasn't needed? Or how is this being used and why does this have to be here?

nvoxland (changes requested):

liquibase-core/src/main/java/liquibase/changelog/DatabaseChangeLog.java
wwillard7800: I have re-worked the implementation.

nvoxland (approved):
I think the changes look good. You agree is seems cleaner, @wwillard7800 ?

wwillard7800:
> I think the changes look good. You agree is seems cleaner, @wwillard7800 ?

Yes, much cleaner

PR #3369, opened 2022-10-13
Handle absolute path in getAll method DAT-11890

wwillard7800:
Do not allow absolute paths in this method


StevenMassaro (commented):
Are you able to write a test around this change?

wwillard7800:
> Are you able to write a test around this change?

Yes.  I had written a unit test originally, and then it got lost.  I added it again.

suryaaki2 (approved):

nvoxland (approved):

PR #2608, opened 2022-03-06
enable 'includeAll' directive to work with liquibase resources provided in OSGi bundle

ktor:
## Environment

**Liquibase Version**: 4.4.0

**Liquibase Integration & Version**: liquibase-core 4.4.0 as a bundle in Apache Felix OSGi container, Liferay portal 7.4

**Liquibase Extension(s) & Version**: 

**Database Vendor & Version**: Postgresql 11.4

**Operating System Type & Version**: Linux

## Pull Request Type

- [ ] Bug fix (non-breaking change which fixes an issue.)
- [x] Enhancement/New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to change)

## Description

I've succesfully used `liquibase-core` deployed as a bundle to Liferay Portal. Another sibling bundle provided `changelog.xml` together with BundleActivator code which executed the `changelog.xml` just fine.

I wanted to organize migration steps in separate files under a directory and used `includeAll` `changelog.xml` directive. The executor code however haven't found any of the migration files except for the entrypoint `changelog.xml` in which `includeAll` directive was declared.

`BundleActivator.java`:
```java
        ...
        ResourceAccessor resourceAccessor = new OSGIResourceAccessor(BundleActivator.class);
        try (Liquibase liquibase = new Liquibase("changelog.xml", resourceAccessor, database)) {
            LockServiceFactory.reset();
            liquibase.update(new Contexts());
        }
        ...
```

The pull requests contains modification of existing `OSGiResourceAccessor` class so that it supports traversing of an OSGi bundle directory structure which provides entrypoint `changelog.xml` and migration resources/paths declared in it.

## Steps To Reproduce

1. Prepare OSGi bundle with following `changelog.xml` resource:
```xml
<?xml version="1.0" encoding="UTF-8"?>

<databaseChangeLog
    xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
    xmlns="http://www.liquibase.org/xml/ns/dbchangelog"
    xsi:schemaLocation="http://www.liquibase.org/xml/ns/dbchangelog http://www.liquibase.org/xml/ns/dbchangelog/dbchangelog-4.3.xsd">

    <includeAll path="steps/" relativeToChangelogFile="true"/>

</databaseChangeLog>
```
2. Add migration step `steps/0001-init.xml`:

```xml
<?xml version="1.0" encoding="UTF-8"?>

<databaseChangeLog
[...]

PR #927, opened 2019-10-17
fix: CompositeClassLoader does not implement getResource(String)

mselerin:
Fix for https://github.com/liquibase/liquibase-hibernate/issues/170

`CompositeClassLoader` does not implement the getResource() and getResources() methods : they always return null / empty collection.

This raise an exception inside `DefaultPersistenceUnitManager.determineDefaultPersistenceUnitRootUrl()` with the liquibase-hibernate plugin.

PR #3378, opened 2022-10-15
Load XSD files under OSGI

ponziani:
PR for https://github.com/liquibase/liquibase/issues/2817


nvoxland (changes requested):
Thanks for the PR! Sorry for the delay looking at it. 

The main thing you're doing is searching in the getClass.getClassLoader() vs. Thread.currentContextLoader right?

It's probably better to have that expanded classpath in what's returned by getSearchClassloader() so it can be better controlled by subclasses, since that method is the "first check the classloader" part of the logic.

Could you update that method to return a ClassLoader that combines the two?

ponziani:
Any suggestions on how to do that ? 

I was looking for a CompositeClassloader, but although it's mentioned in your changelog.txt https://github.com/liquibase/liquibase/pull/927, it's not there in the codebase. Any history on that ?

nvoxland:
I was hoping there was something in the general java APIs to have a composite classloader, but didn't see one. I pushed a commit to your fork with what I was generally thinking. Does that make sense? Does that still work with OSGi better?

ponziani:
Hi Nathan, I haven't tried it, but I assume it will work with OSGi.

Fyi, Drools has an implementation of this idea as well : https://github.com/kiegroup/drools/blob/c07f0d800ae7441fcb5b6fe7055088315569e2ae/kie-internal/src/main/java/org/kie/internal/utils/CompositeClassLoader.java

ponziani:
Hi @nvoxland, @filipelautert,

I believe an even better solution might be to make OSGiResourceAccessor a CompositeResourceAccessor

as suggested in https://github.com/liquibase/liquibase/pull/3597#issuecomment-1364647460

Please have a look at that other ticket, which is about the same issue.

filipelautert:
> Hi @nvoxland, @filipelautert,
> 
> I believe an even better solution might be to make OSGiResourceAccessor a CompositeResourceAccessor
> 
> as suggested in [#3597 (comment)](https://github.com/liquibase/liquibase/pull/3597#issuecomment-1364647460)
> 
> Please have a look at that other ticket, which is about the same issue.

Hello @ponziani  - I took a look at #3597 and it seems easy to implement. Can we close this PR and use 3597  as the fix?

ponziani:
@filipelautert , I think you can close this one indeed. But https://github.com/liquibase/liquibase/pull/3597 does not implement the suggested solution. 

https://github.com/liquibase/liquibase/pull/3597 makes the XSD's available for other bundles, which is good, but I feel liquibase-core should be able to access them itself, which can be achieved by making OSGiResourceAccessor a CompositeResourceAccessor imho.

filipelautert:
@ponziani oh , I see. So this one may be more related to PR #2608 ?

ponziani:
Yep. To be able to load the liquibase-core XSD-files, the OSGiResourceAccessor should be a combination of the ClassLoaderResourceAccessor for the calling bundle it is now with a ClassLoaderResourceAccessor for the liquibase-core bundle.

For https://github.com/liquibase/liquibase/pull/2608 the first ClassLoaderResourceAccessor could be extended as suggested in that PR.

filipelautert (approved):
Changes are fine, it just adds a seond classloader to make sure that resources can be loaded in multiclassloader environments.

suryaaki2 (approved):

XDelphiGrl (approved):
By default, Liquibase enables SECURE_PARSING to prevent loading potentially malicious remote XSD files. This logic breaks XSD loading in OSGI environments because Liquibase is not able to load its own XSDs from a jar file. This PR addresses that issue by implementing a composite classloader to get resources from multiple loaders.
* Test harness and functional test executions passing.

APPROVED

XDelphiGrl:
@nvoxland, have your concerns been addressed? If so, we can move this to ready to merge.

PR #3387, opened 2022-10-19
Use PathHandler for writing log files, which allows using S3 and other extensions (DAT-11515)

StevenMassaro:
Reverts liquibase/liquibase#3368


abrackx (approved):

nvoxland (approved):
Assuming it's the same as I approved before

PR #3389, opened 2022-10-20
check all relevant snapshot generators when determining types for a snapshot (DAT-11512)

StevenMassaro:
## Impact

## Description

Check all relevant snapshot generators when determining relevant types for a snapshot. Previously, only the first generator in the list was checked.

## Things to be aware of

I tried to write an integration test for this for quite some time and was unsuccessful. If I am successful, I will push it up.

### DIFF
--- a/liquibase-core/src/main/java/liquibase/snapshot/SnapshotGeneratorFactory.java
+++ b/liquibase-core/src/main/java/liquibase/snapshot/SnapshotGeneratorFactory.java
@@ -306,12 +306,13 @@ public class SnapshotGeneratorFactory {
         SortedSet<SnapshotGenerator> generators = getGenerators(type, database);
 
         if ((generators != null) && !generators.isEmpty()) {
-            SnapshotGenerator generator = generators.iterator().next();
-            addsTo = generator.addsTo();
-            if (addsTo != null) {
-                for (Class<? extends DatabaseObject> newType : addsTo) {
-                    returnSet.add(newType);
-                    getContainerTypes(newType, database, returnSet);
+            for (SnapshotGenerator generator : generators) {
+                addsTo = generator.addsTo();
+                if (addsTo != null) {
+                    for (Class<? extends DatabaseObject> newType : addsTo) {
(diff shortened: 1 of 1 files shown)


suryaaki2 (approved):

liquibase-core/src/main/java/liquibase/snapshot/SnapshotGeneratorFactory.java
```diff
@@ -306,12 +306,13 @@ private void getContainerTypes(Class<? extends DatabaseObject> type, Database da
         SortedSet<SnapshotGenerator> generators = getGenerators(type, database);
 
         if ((generators != null) && !generators.isEmpty()) {
-            SnapshotGenerator generator = generators.iterator().next();
```
nvoxland, line 309: I don't think this is a change we can make, because it breaks how the extension system is supposed to work.

The purpose of the getPriority() call and this spot that sorts the valid generators by that priority and then picks the single one with the highest priority is to allow extensions to replace the existing functionality and have the best implementation for the interface be used. 

For example, the standard ColumnSnapshotGenerator assumes a relational databases, and the MongodbColumnSnapshotGenerator knows how to work with mongo. Therefore this code needs to make sure that only the MongodbColumnSnapshot generator is used and NOT ColumnSnapshotGenerator.

Or even if ColumnSnapshotGenerator **_could_** work without error but OracleColumnSnapshotGenerator says  it has a higher priority, we need to just run the one implementation.

If you're thinking you need to execute all the implementations, it's probably pointing to a bug in one of the implementations which is either claiming it should run when it shouldn't, or not doing everything it should.

nvoxland (changes requested):

liquibase-core/src/main/java/liquibase/snapshot/SnapshotGeneratorFactory.java
StevenMassaro: This is the case. The fix is in the pro repo. Closing this PR.

PR #3396, opened 2022-10-21
Resets changelog cache upon acquiring lock

### LINKED ISSUE #2816: Upgrade lock not working correctly due to cache of `ranChangeSetList`
## Environment

**Liquibase Version**: 4.9.0

**Liquibase Integration & Version**: Quarkus 2.8.1

**Database Vendor & Version**: PSQL

## Description

When creating a liquibase instance and use them for two commands, the lock is not working correctly.

We have following code:

```java
  try (Liquibase liquibase = liquibaseFactory.createLiquibase()) {
          liquibase.validate();
          liquibase.update(liquibaseFactory.createContexts(), liquibaseFactory.createLabels());
  }
```

Following happens:
* Inside the `validate()` method, the `ranChangeSetList` (from `StandardChangeLogHistoryService`) is created and filled with the values from the change log table
* in the `update()` a lock on the table is created
  * the content on the table is not read again, the cached `ranChangeSetList` is used. -> so both services that loaded the `ranChangeSetList` before any change, will try to run the new changesets.
  * So the changesets are run twice

## Steps To Reproduce

Create following code: 

```java
  try (Liquibase liquibase = liquibaseFactory.createLiquibase()) {
          liquibase.validate();
          Thread.sleep(1000);
          liquibase.update(liquibaseFactory.createContexts(), liquibaseFactory.createLabels());
  }
```

Run the code twice in parallel. 

## Actual Behavior

The new changesets will be applied twice.

## Expected/Desired Behavior

The `ranChangeSetList` is revalidated during the duration of the lock, and so the changes are applied only once

filipelautert:
## Impact

## Description

- Resets changelog cache before running update making sure that it's using the latest information from database. In concurrent  process scenarios 
- Fix #2816 

## Things to be aware of
- None

## Things to worry about
- - how does it affects performance?


nvoxland (changes requested):
This is not the right spot for this resetAll.

The different history services should be managing their state correctly for normal usage, and the reset call should be more for one-off "we know we are doing something strange, and so need to tell them to reset" things. 

Adding a rest to the update operation is working around whatever the underlying bug is. Perhaps too aggressive caching in the implementation? Perhaps a specific case where we should know that the implementations can't know what's changing under them and do the reset there?

filipelautert (commented):
Tks, changed as discussed.

XDelphiGrl (approved):
PR handles cases where multiple threads are attempting to run the same changesets due to a stale cache of `ranChangeSetList`. This situation was encountered by users running multiple simultaneous Liquibase Docker instances. Fix resets the `ranChangeSetList` to prevent the contents from becoming stale over the lifetime of the lock.

APPROVED

XDelphiGrl:
@filipelautert and @FBurguer, hello!

We should attempt to recreate the bug and then validate the fix is working. The linked issue gives a code snippet that can be used to trigger the situation.  The user initially encountered the problem running with multiple Liquibase Docker containers. I believe that either the code snippet or using multiple Docker containers are equally good choices for testing. 

I suggest timeboxing the reproduction of the bug. Threading issues are difficult to recreate and can cost more time than is reasonable to merge the fix to master to see if the PR addresses the issue for the broader community.

-erz

filipelautert:
@XDelphiGrl @FBurguer I'm attaching here a java class that can be used to test it. 

I run 2 instances with the following parameters (it need ot be 2 processes, can't be threads that's why 2 executions):

* t1 15000
* t2 1

image

I hope it helps!

[RunnableProblem.txt](https://github.com/liquibase/liquibase/files/9930971/RunnableProblem.txt)

suryaaki2 (approved):

nvoxland (approved):

PR #3398, opened 2022-10-23
Do not write output files for computed columns when executing dbDoc.  Fixes #1088

### LINKED ISSUE #1088: Liquibase-core - 3.5.5.x -  dbDoc fails generating of documents
Prerequisites:
- Oracle DB(with at least one table with minimum necessary column(id, first_name));
- Windows 10;
- Liquibase-core 3.5.5.x;
- liquibase.properties
`driver=oracle.jdbc.driver.OracleDriver`
`changeLogFile=test_changelog.xml`
`username=liquibase_user`
`password=root`
`url=jdbc:oracle:thin:@localhost:1521:ORCL`
- test_changelog.xml
`<?xml version="1.0" encoding="UTF-8"?>`
     `<databaseChangeLog xmlns="http://www.liquibase.org/xml/ns/dbchangelog" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"` `xsi:schemaLocation="http://www.liquibase.org/xml/ns/dbchangelog http://www.liquibase.org/xml/ns/dbchangelog/dbchangelog-3.0.xsd">`
    `<changeSet runAlways="true" author="testuser" id="3">`
        `  <sql>CREATE INDEX TEST_INDEX ON PERSON (CASE WHEN (FIRST_NAME IS NOT NULL
            AND (FIRST_NAME='1' OR FIRST_NAME='2' OR FIRST_NAME='3' OR FIRST_NAME='4' OR FIRST_NAME='5'`
            `OR FIRST_NAME='6' OR FIRST_NAME='7' OR FIRST_NAME='8' OR FIRST_NAME='9' OR FIRST_NAME='10' OR FIRST_NAME='11'`
            `OR FIRST_NAME='12' OR FIRST_NAME='13' OR FIRST_NAME='14' OR FIRST_NAME='15' OR FIRST_NAME='16' OR FIRST_NAME='17' OR FIRST_NAME='18'))
            THEN 1 ELSE 0 END)</sql>`
    `</changeSet>`

`</databaseChangeLog>`

Steps to reproduce:
1. Execute the next command in cmd:
`java -jar liquibase-core-3.5.5.jar --driver=oracle.jdbc.driver.OracleDriver --classpath=<path_to_jdbc_driver> --changeLogFile=<path_to_test_changelog.xml> --url="jdbc:oracle:thin:@localhost:1521:ORCL" --username=<user_name> --password=<password> update`
2. `java -jar liquibase-core-3.5.5.jar --driver=oracle.jdbc.driver.OracleDriver --classpath=<path_to_jdbc_driver> --changeLogFile=<path_to_test_changelog.xml> --url="jdbc:oracle:thin:@localhost:1521:ORCL" --username=<user_name> --password=<password> dbDoc folder_to_out_docs`

**Expected result**: dbDoc generates html files for each table column as well as other attributes.
**Actual result:** `Unexpected error running Liquibase: dbDoc_out\columns\liquibase_user.person.case__when___first_name__is_not_null_and___first_name___1__or__first_name___2__or__first_name___3__or__first_name___4__or__first_name___5__or__first_name___6__or__first_name___7__or__first_name___8__or__first_name___9__or__first_name___10__or__first_name___11__or__first_name___12__or__first_name___13__or__first_name___14__or__first_name___15__or__first_name___16__or__first_name___17__or__first_name___18____then_1_else_0_end.html (The filename, directory name, or volume label syntax is incorrect).`

RCA:
The reason of this error is dbDoc tries to create html doc with index body in filename for table column. As we can see the index body is too long and file system is not allow to create files with such too long file name(255 characters constraint). So, from my perspective, dbDoc identifies index body as column and generates doc for it.

Please fix it.

rozenshteyn:
## Impact

## Description

Fix dbDoc fails to generate documents for an Oracle database when an index is created using a filter clause.  Fixes #1088.
Skip writing an html file for Oracle database columns which have the computed flag set (i.e. the column was created based on the content of the filter clause of an index).

## Things to be aware of

- I constrained the code change to only apply to the Oracle database.  I'm not sure if other database types could have the same issue whereupon the column name is set to the content of the index filter clause in the https://github.com/liquibase/liquibase/blob/9d6d14455e209fc60d1a7c9ce3477a56b42bf11e/liquibase-core/src/main/java/liquibase/snapshot/jvm/IndexSnapshotGenerator.java#L296-L297
- Since I was not sure how to properly mock a static method in groovy/spock, I created DBDocVisitorTest.java.

## Things to worry about

- None at this time.

## Additional Context

None at this time.


filipelautert:
Hi @rozenshteyn ! Thanks for the PR. 
I ran the same test case from the issue on Postgres and had the same problem, so you were right that it's not just an Oracle problem.  But I tried a different approach: I used the original class but changed the SnapshotControl to just bring Table and Columns as below (DBDocVisitor line 129), and it worked on Postgres (index wasn't listed as column):

```
        DatabaseSnapshot snapshot = SnapshotGeneratorFactory.getInstance().createSnapshot(
                database.getDefaultSchema(), database, new SnapshotControl(database, Table.class, Column.class));
```

Could you help test it on Oracle? So maybe this would be a simpler approach to fix this?

rozenshteyn:
@filipelautert I tested your proposed fixed with the Oracle db.   While it does resolve the issue of failing to write out db documentation, it creates a new issue whereupon the index info is missing in the produced db documentation.  Since you've mentioned that original issue affects not just the Oracle db, I went ahead and modified my fix by removing the restriction that it applies to the Oracle db only.

filipelautert:
@rozenshteyn thanks for testing out my approach! So all seems good, moving along. Thanks for the fix.

filipelautert (approved):
Working fine, also adds tests.

liquibase-core/src/test/java/liquibase/changelog/visitor/DBDocVisitorTest.java
```diff
@@ -0,0 +1,90 @@
...
+import org.junit.rules.TemporaryFolder;
+import org.junit.runner.RunWith;
+import org.junit.runners.Parameterized;
+import org.junit.runners.Parameterized.Parameters;
+import org.mockito.MockedStatic;
+
+import liquibase.CatalogAndSchema;
+import liquibase.database.Database;
+import liquibase.database.core.OracleDatabase;
+import liquibase.database.core.PostgresDatabase;
```
XDelphiGrl, line 26: I do not believe the import for liquibase.database.core.PostgresDatabase is needed for this test.

suryaaki2 (approved):

PR #3406, opened 2022-10-25
Split statements going through SqlChangeLogParser

nvoxland:
## Impact

## Description

Previously, when parsing "raw" sql files, we would not attempt to split the content. We did that back in the day when we didn't have as smart of an SQL parser, since it would just split on `;` even if it was in a stored procedure or something.

In effect, since we couldn't trust our ability to split the statements, we rely on you to split them yourself if using includeAll.

Now that we are able to better differentiate between delimiters in quotes and in begin/end blocks, and handle ; and GO and / it's perhaps time to change that default?

## Things to be aware of

- For people that have no end delimiters in their sql file, there is no change
- For people that have a single statement with complex SQL, there SHOULD be no change, since we SHOULD parse that OK now. But there may be cases where we don't and previously successful changelog will not fail.
- For people that have multiple statements in their SQL files, their changelog now executes instead of failing.

## Things to worry about

- Is there any statements people used to use in raw SQL includes that cannot be split with our current logic?


filipelautert (dismissed):
Besides tests were added, I believe manual tests would be interesting to confirm case 3: "For people that have multiple statements in their SQL files, their changelog now executes instead of failing."

mariochampion:
hey all, this PR should **NOT be merged** yet, as we need to test this change against the Native Executors and Quality Checks, which will occur in DAT ticket (to be written and appended here soon). After that we'll get it into a a final code review and merge plan. thanks!

filipelautert:
@XDelphiGrl  - After Nathan's comments, should we go ahead with this one?

MalloD12:
Hi @nvoxland,

Hope you are doing very well. I would like to check with you if based on your knowledge you think this would be still needed or not. Recently, there were a few changes related to split statements and more directly related to `endDelimiter`.  When I saw this:

```java
change.setSplitStatements(false);
```

I thought it could break things, but to be honest I'm not 100% sure that could be that risky.

Thanks,
Daniel.

PR #3413, opened 2022-10-26
Fixed resolveSibling method by updating URL resource type

### LINKED ISSUE #3393: `relativeToChangeLogFile` not working with `SpringResource#resolveSibling`
## Environment

**Liquibase Version**: 4.17

**Liquibase Integration & Version**: Spring Boot 3.0.0-RC1

## Description

Folder structure:
```
src/main/resources/db/changelog/
├── changelog.xml
└── data.csv
```

changelog.xml:
```xml
<databaseChangeLog xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
                   xsi:schemaLocation="http://www.liquibase.org/xml/ns/dbchangelog http://www.liquibase.org/xml/ns/dbchangelog/dbchangelog-4.5.xsd">

    <changeSet id="relative import" author="kzander91">
        <createTable tableName="my_table">
            <column name="test_col" type="bigint"/>
        </createTable>
        <loadData tableName="my_table"
                  file="data.csv"
                  relativeToChangelogFile="true">
            <column name="test_col" type="bigint"/>
        </loadData>
    </changeSet>

</databaseChangeLog>
```

The `loadData` change resolves `file` to `/data.csv` instead of `/db/data.csv`. This worked previously with 4.9.1 and broke with 4.17.0.
The bug is located in `SpringResource`:
https://github.com/liquibase/liquibase/blob/c9ceb05141eee992c7aaa3002451bb5f531b49d5/liquibase-core/src/main/java/liquibase/integration/spring/SpringResource.java#L40-L49

If `resource` is a `org.springframework.core.io.UrlResource`, `createRelative()` uses `new URL(URL, String)` to resolve the relative URL:
https://github.com/spring-projects/spring-framework/blob/v6.0.0-RC2/spring-core/src/main/java/org/springframework/core/io/UrlResource.java#L310-L317

This effectively calls
```java
new URL(new URL("file://C:/Temp/demo/target/classes/db/changelog/changelog.xml"), "../data.csv")
```
Which resolves to `file://C:/Temp/demo/target/classes/db/data.csv` instead of `file://C:/Temp/demo/target/classes/db/changelog/data.csv`.

## Steps To Reproduce

- Use `<loadData relativeToChangelogFile="true"/>`.
- Start migration with Spring Boot.

## Actual Behavior
Relative file is resolved incorrectly, migration aborts due to "unknown file".

## Expected/Desired Behavior
Relative paths are resolved correctly as before.

## Reproducer
[demo.zip](https://github.com/liquibase/liquibase/files/9838779/demo.zip)

* Unzip and run `mvnw test`, test fails.
* Uncomment `liquibase.version` property in pom and rerun: test succeeds.

## Possibly related
#3354 
#3365 
#3064

MalloD12:
## Impact

## Description

This change has changed the resource type and the way it uses to create a relative path to a changelog file.

Fixes #3393

## Things to be aware of

- ResolveSibling logic updated for a Spring resource.

## Things to worry about

- This change has an impact only with SpringBoot.

## Additional Context

- Nothing


OleksandrShkurat:
Py project is also affected by this issue, so I really expect this fix to be released as soon as possible.

MalloD12:
> My project is also affected by this issue, so I really expect this fix to be released as soon as possible. Currently, I forced to rollback liquibase to 4.16.1

Hi @OleksandrShkurat thanks for sharing. I'll do my best to have this change ready and available to merge as soon as possible. I would need first to get some review/feedback from the team/community.

Thanks,
Daniel.

OleksandrShkurat:
I've tried your change locally in my project and it helps

liquibase-core/src/main/java/liquibase/integration/spring/SpringResource.java
```diff
@@ -40,7 +41,7 @@ public liquibase.resource.Resource resolve(String other) {
     @Override
     public liquibase.resource.Resource resolveSibling(String other) {
         try {
-            Resource otherResource = this.resource.createRelative("../"+other);
+            UrlResource otherResource = new UrlResource(this.resource.createRelative(other).getURL());
```
nvoxland, line 44: I don't think this is the correct fix. Switching an existing Resource implementation to UrlResouce class is just working around the problem that there are differences between the Resource implementations that shouldn't be there.

I'm thinking maybe #3355 is the more wide-ranging fix for this?

Viserius:
Any updates yet? @nvoxland

MalloD12:
> Any updates yet? @nvoxland

Hi @Viserius,

We have a fix already implemented for this, but we are not able yet to move forward with this due to a test dependency issue we are having with one of our projects. We have part of the team already working on it and I hope we can solve that issue asap so we can move forward with this. I'll keep you guys posted here. 

Thanks for your patience,
Daniel.

filipelautert (approved):
Fixes the path handling.

XDelphiGrl (commented):
PR fixes Liquibase Springboot startup migrations for changelogs with `loadData` changesets. The bug manifested when the data file (csv) was relative to the changelog because an incorrect path to the csv was constructed. This PR removes an unecessary `..\` from the path to the csv.
* Passing PR to manual QA to verify the bug is fixed.

XDelphiGrl (approved):

liquibase-core/src/main/java/liquibase/integration/spring/SpringResource.java
```diff
@@ -3,6 +3,7 @@
 import liquibase.exception.UnexpectedLiquibaseException;
 import liquibase.resource.OpenOptions;
 import org.springframework.core.io.Resource;
+import org.springframework.core.io.UrlResource;
```
konikvranik, line 6: this looks like unused import

MalloD12: Hi @konikvranik,

Thanks for pointing this out. I'm going to remove this unused import in a different PR.

PR #3436, opened 2022-11-03
The onSqlOutput attribute does not work for preconditions in formatted SQL changelogs.

dyadyushko:
## Impact

## Description
The attribute to control the behavior of update-sql when using preconditions is onUpdateSQL for formatted SQL changelogs.  For all other changelog formats, the attribute is onSqlOutput.  We should be consistent so that the same attribute works for all changelog formats. That is, users of formatted sql changelogs should be able to use onSqlOutput.

## Things to be aware of
 - None
## Things to worry about
- None


liquibase-core/src/main/java/liquibase/parser/core/formattedsql/FormattedSqlChangeLogParser.java
```diff
@@ -422,11 +423,16 @@ public DatabaseChangeLog parse(String physicalChangeLogLocation, ChangeLogParame
                                 Matcher onFailMatcher = onFailPattern.matcher(body);
                                 Matcher onErrorMatcher = onErrorPattern.matcher(body);
                                 Matcher onUpdateSqlMatcher = onUpdateSqlPattern.matcher(body);
+                                Matcher onSqlOutputMatcher = onSqlOutputPattern.matcher(body);
 
                                 PreconditionContainer pc = new PreconditionContainer();
                                 pc.setOnFail(StringUtil.trimToNull(parseString(onFailMatcher)));
                                 pc.setOnError(StringUtil.trimToNull(parseString(onErrorMatcher)));
-                                pc.setOnSqlOutput(StringUtil.trimToNull(parseString(onUpdateSqlMatcher)));
+                                if (onSqlOutputMatcher.matches()) {
```
suryaaki2, line 431: @dyadyushko Can you handle the condition that if changelog contains both onUpdateSql and onSqloutput in which case, we should throw an error, "Please modify the changelog to have preconditions set with either "onUpdateSql" or "onSqlOutput"?

suryaaki2 (changes requested):

nvoxland (approved):

suryaaki2 (approved):

filipelautert (approved):

PR #3482, opened 2022-11-17
DAT-11899 Create Snapshot builds of liquibase for downstream consumption

jnewton03:
## Impact

## Description

## Things to be aware of

## Things to worry about

- List unknowns you have


.github/workflows/build-main.yml
```diff
@@ -0,0 +1,66 @@
...
+        uses: actions/setup-java@v3
+        with:
+          java-version: '17'
+          distribution: 'temurin'
+          cache: 'maven'
+          server-id: github
+
+      # Version artifact based off of branch and commit SHA.
+      - name: Version Artifact
+        run: mvn versions:set -DnewVersion="$(git rev-parse --short HEAD)-SNAPSHOT"
```
StevenMassaro, line 60: You might be able to use `${{github.sha}}` here.

.github/workflows/build-main.yml
```diff
@@ -0,0 +1,66 @@
+name: Build Main/Master SNAPSHOT
+
+on:
+  workflow_dispatch:
+  schedule:
+    - cron: "5 6 * * *"
```
StevenMassaro, line 6: Personally, I think a nightly build is unnecessary if we're building on every push. Seems like a waste of resources to me.

StevenMassaro (approved):

mcred (approved):

abrackx (approved):

.github/workflows/build-main.yml
jnewton03: @StevenMassaro doesn't that give the full commit has and not the short version?

.github/workflows/build-main.yml
jnewton03: Easy to remove if you'd like

.github/workflows/build-main.yml
StevenMassaro: Good point, it does give the full hash.

PR #3483, opened 2022-11-17
Refactoring of Maven plugin code to better support secret store extensions DAT-12158

wwillard7800:
This PR is based on work done by @nvoxland to make better use of the Command Framework for execution of Maven goals.  This refactoring will also allow secret store extensions to work in the Maven plugin.  Previously, the Mojo base class was responsible for opening a database connection.  This causes an issue when the URL and credentials are secrets, as their values have not yet been retrieved.  Now, the commands associated with the goals will open the connection.

For now, extension configuration arguments/parameters will need to be specified as either Maven properties on the command line, or in environment variables.


liquibase-maven-plugin/pom.xml
```diff
@@ -75,23 +75,42 @@
...
+            <!--                            <goal>helpmojo</goal>-->
+            <!--                        </goals>-->
+            <!--                    </execution>-->
+            <!--                </executions>-->
+            <!--                <configuration>-->
+            <!--                    <mojoDependencies>-->
+            <!--                        <param>org.liquibase:liquibase-maven-plugin</param>-->
+            <!--                    </mojoDependencies>-->
+            <!--                </configuration>-->
+            <!--            </plugin>-->
```
StevenMassaro, line 113: Do you want this to be commented out?

liquibase-maven-plugin/pom.xml
```diff
@@ -172,14 +193,14 @@
...
-    </reporting>
+    <!--    <reporting>-->
+    <!--        <plugins>-->
+    <!--            <plugin>-->
+    <!--                <groupId>org.apache.maven.plugins</groupId>-->
+    <!--                <artifactId>maven-plugin-plugin</artifactId>-->
+    <!--                <version>3.6.4</version>-->
+    <!--            </plugin>-->
+    <!--        </plugins>-->
+    <!--    </reporting>-->
```
StevenMassaro, line 204: Do you want this to be commented out?

liquibase-maven-plugin/pom.xml
```diff
@@ -10,7 +10,7 @@
     </parent>
 
     <artifactId>liquibase-maven-plugin</artifactId>
-    <packaging>maven-plugin</packaging>
+    <packaging>jar</packaging>
```
StevenMassaro, line 13: Do we actually want to change the packaging?

liquibase-maven-plugin/src/main/java/org/liquibase/maven/plugins/AbstractLiquibaseMojo.java
```diff
@@ -916,16 +917,17 @@ private static InputStream handlePropertyFileInputStream(String propertyFile) th
...
+//        try {
+//            return MavenUtils.getArtifactClassloader(project,
+//                    includeArtifact,
+//                    includeTestOutputDirectory,
+//                    getClass(),
+//                    getLog(),
+//                    verbose);
+//        } catch (MalformedURLException e) {
+//            throw new MojoExecutionException("Failed to create artifact classloader", e);
+//        }
```
StevenMassaro, line 930: Should this be commented out? If so can the method be deleted?

liquibase-maven-plugin/src/main/java/org/liquibase/maven/plugins/LiquibaseMojo.java
```diff
@@ -0,0 +1,163 @@
...
+    private Map<String, Object> configuration;
+    private MavenProject project;
+    private MojoExecution mojoExecution;
+
+    public LiquibaseMojo() {
+    }
+
+    @Override
+    public void addComponentRequirement(ComponentRequirement componentRequirement, Object o) throws ComponentConfigurationException {
+        //TODO: How does this get used?
```
StevenMassaro, line 33: Is this important?

liquibase-maven-plugin/src/main/java/org/liquibase/maven/plugins/LiquibaseMojo.java
```diff
@@ -0,0 +1,163 @@
...
+        @Override
+        protected boolean keyMatches(String wantedKey, String storedKey) {
+            if (super.keyMatches(wantedKey, storedKey)) {
+                return true;
+            }
+            if (wantedKey.startsWith("liquibase.command.")) {
+                return super.keyMatches(wantedKey.replaceFirst("^liquibase\\.command\\.", ""), storedKey);
+            }
+
+            return super.keyMatches(wantedKey.replaceFirst("^liquibase\\.", ""), storedKey);
```
StevenMassaro, line 150: I know I've personally duplicated this logic in another place already. Do you think it would be possible to move this to an abstract class and remove the duplication here and elsewhere?

liquibase-maven-plugin/src/main/java/org/liquibase/maven/plugins/MavenResourceAccessor.java
```diff
@@ -1,21 +1,53 @@
...
  * Extension of {@link liquibase.resource.ClassLoaderResourceAccessor} for Maven which will use a default or user specified {@link ClassLoader} to load files/resources.
  */
-public class MavenResourceAccessor extends ClassLoaderResourceAccessor {
+public class MavenResourceAccessor extends CompositeResourceAccessor {
 
-    public MavenResourceAccessor(ClassLoader classLoader) {
-        super(classLoader);
+    public MavenResourceAccessor(MavenProject project) throws DependencyResolutionRequiredException, IOException {
+        for (String element : project.getCompileClasspathElements()) {
+            //this.addResourceAccessor(new DirectoryResourceAccessor(new File(element)));
```
StevenMassaro, line 21: Do you intend for this to be commented out?

liquibase-maven-plugin/src/main/java/org/liquibase/maven/plugins/MavenResourceAccessor.java
```diff
@@ -1,21 +1,53 @@
...
+        }
+
+        Set<Artifact> dependencies = project.getArtifacts();
+        if (dependencies != null) {
+            for (Artifact artifact : dependencies) {
+                //this.addResourceAccessor(new DirectoryResourceAccessor(artifact.getFile()));
```
StevenMassaro, line 27: Do you intend for this to be commented out?

liquibase-maven-plugin/src/main/java/org/liquibase/maven/plugins/MavenResourceAccessor.java
```diff
@@ -1,21 +1,53 @@
...
+        File projectArtifactFile = project.getArtifact() != null ? project.getArtifact().getFile() : null;
+        if (projectArtifactFile == null) {
+            if (project.getBuild() != null) {
+                this.addResourceAccessor(new DirectoryResourceAccessor(new File(project.getBuild().getOutputDirectory())));
+            }
+        } else {
+            this.addResourceAccessor(new DirectoryResourceAccessor(projectArtifactFile));
+        }
+
+//TODO        if (includeTestOutputDirectory) {
```
StevenMassaro, line 47: Do you intend for this to be commented out?

liquibase-maven-plugin/pom.xml
```diff
@@ -75,23 +75,42 @@
...
+                        <id>default-liquibase-build</id>
+                        <phase>generate-resources</phase>
                         <goals>
-                            <goal>helpmojo</goal>
+                            <goal>build-plugin-xml</goal>
                         </goals>
+                        <configuration>
+                            <liquibaseClassesDir>../liquibase-core/target/classes</liquibaseClassesDir>
+                            <liquibaseProClassesDir>../../liquibase-pro/pro/target/classes</liquibaseProClassesDir>
+                            <outputFile>src/main/resources/META-INF/maven/plugin.xml</outputFile>
```
StevenMassaro, line 91: I see you've also committed to the plugin.xml file. Should that be committed or gitignored? Regardless, what's the process for automatically updating that file on a regular basis?

liquibase-maven-plugin/src/main/java/org/liquibase/maven/plugins/LiquibaseMojo.java
nvoxland: This was probably a comment I put in as I was trying to figure out how the plugin system works

liquibase-maven-plugin/src/main/java/org/liquibase/maven/plugins/LiquibaseMojo.java
nvoxland: I had started on #3286 which was part of my same "how can maven work with the command framework" spike as this, but I kept separate because it was "how do we add an 'integration api' layer" initial experimentation. A big goal of that is to reduce duplicated logic like this. 

It's probably worth keeping the duplication for now and getting a command-API version of maven out there. Then use the 3rd integration we convert (ant?) as the time we find and extract the shared logic from all three integrations.

liquibase-maven-plugin/pom.xml
nvoxland: In automation this is going to be a problem relying on the pro classes because there are no pro classes when building this code. We maybe even have the liquibase-commercial.jar available currently, since that's just brought in as a special case in building liquibase-dist possibly at the end.

nvoxland (changes requested):
I talked to wes about how we need to put in the "support backwards compatibility" work prior to being able to merge this in.

PR #3486, opened 2022-11-18
DAT-11579: when generating changelogs for MySQL, ignore table column order for PKs

StevenMassaro:
## Impact

## Description

MySQL requires that a table's primary key is created when the table is created. Liquibase was generally respecting this requirement, but if the order of the columns in the PK was different than the order of the columns in the table, Liquibase would create the primary key in a separate change set, which MySQL did not like:

```
Unexpected error running Liquibase: Migration failed for changeset generated.xml::1668700616367-2::Steve (generated):
     Reason: liquibase.exception.DatabaseException: Incorrect table definition; there can be only one auto column and it must be defined as a key [Failed SQL: (1075) CREATE TABLE steve2.str4 (col1 INT NOT NULL, col2 INT AUTO_INCREMENT NOT NULL, col3 INT NULL)]
```

This change allows primary keys in generated changelogs from MySQL to have the primary key in a different order than the order of the columns in the database.

## Things to be aware of

This should only affect MySQL.


abrackx (approved):

filipelautert (approved):

suryaaki2 (approved):

wwillard7800 (approved):

filipelautert (approved):
For Mysql fixes column order when required to make sure thar primary key is specified at creation for auntoincrement column.
Tests added.

PR #3489, opened 2022-11-19
Feature to allow checksum upgrade to support either new normalized filepaths or legacy stored file paths

vuparkar:
## Impact

- [x ] Enhancement/New feature (adds functionality without impacting existing logic)
 
## Description
When attempting to upgrade checksums in the databasechangelog table, Liquibase generates an `UPDATE` statement whose `WHERE` clause uses the normalized file path. This will cause `RanChangeSets` that used the older non-normalized paths, for eg with a `classpath:<some/file/path>`, in them to ignore the update and the checksum is never saved.

- Change being submitted
  - Adding a new `GlobalConfiguration` definition flag `liquibase.storeNormalizedFileNameInDatabaseChangelogTable`
  - Default value is `True`
  - Configuring this to `False` will allow fallback to the older Update Checksum behaviour but keep the new normalized file check everywhere else
  - Added tests to verify this behaviour
  - Added a fix to recent upstream code changes that cause the generated changesets to forget the stored filepaths

- Steps to Reproduce
  - Create a project that uses Liquibase 3.x (for eg. 3.5.3)
  - Configure the changelog file location using a classpath: specification
  - Verify the checksum generated in the databasechangelog table contains the file name with classpath: in it
  - Verify that the MD5Sum column begins with 7 and not 8 as in the latest versions
  - Now update the project to use latest liquibase versions
  - Rerun the project and the Checksum will not be updated
  - Debug logging will indicate 0 records updated
  - This happens since the where clause uses the normalized file specification which doesnt have classpath: in it.

## Things to be aware of

- The changes in the upstream classes `ChangeLogIterator` and `AbstractChangeLogHistoryService`  can be avoided with a simple change in `DatabaseChangeLog.getChangeSets(RanChangeSets ...)` but the functionality is not clear to me so left unchanged there. 
- Since this is an extra field that is being updated in the `ChangeSet` other functionality that is not using it should be unaffected 
  
## Things to worry about

- No tests for the `ChangeLogIterator` and `AbstractChangeLogHistoryService`

### DIFF
--- a/liquibase-core/src/main/java/liquibase/sqlgenerator/core/UpdateChangeSetChecksumGenerator.java
+++ b/liquibase-core/src/main/java/liquibase/sqlgenerator/core/UpdateChangeSetChecksumGenerator.java
@@ -13,6 +13,7 @@ import liquibase.statement.SqlStatement;
 import liquibase.statement.core.UpdateChangeSetChecksumStatement;
 import liquibase.statement.core.UpdateStatement;
 import liquibase.structure.core.Column;
+import liquibase.util.StringUtil;
 
 public class UpdateChangeSetChecksumGenerator extends AbstractSqlGenerator<UpdateChangeSetChecksumStatement> {
     @Override
@@ -45,7 +46,8 @@ public class UpdateChangeSetChecksumGenerator extends AbstractSqlGenerator<Updat
 
     private String getFilePathToUseInWhereClause(ChangeSet changeSet) {
         String changeSetFilePath = changeSet.getFilePath();
-        if (!GlobalConfiguration.STORE_NORMALIZED_FILE_NAME_IN_DATABASECHANGELOG_TABLE.getCurrentValue()) {
+        if (!GlobalConfiguration.STORE_NORMALIZED_FILE_NAME_IN_DATABASECHANGELOG_TABLE.getCurrentValue()
+                && StringUtil.isNotEmpty(changeSet.getStoredFilePath())) {
             changeSetFilePath = changeSet.getStoredFilePath();
         }
         return changeSetFilePath;
--- a/liquibase-core/src/main/java/liquibase/sqlgenerator/core/UpdateChangeSetChecksumGenerator.java
+++ b/liquibase-core/src/main/java/liquibase/sqlgenerator/core/UpdateChangeSetChecksumGenerator.java
@@ -1,5 +1,6 @@
 package liquibase.sqlgenerator.core;
 
+import liquibase.GlobalConfiguration;
 import liquibase.changelog.ChangeSet;
 import liquibase.changelog.column.LiquibaseColumn;
 import liquibase.database.Database;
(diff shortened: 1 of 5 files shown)


nvoxland:
It's a good catch on the

> When attempting to upgrade checksums in the databasechangelog table, Liquibase generates an UPDATE statement whose WHERE clause uses the normalized file path. This will cause RanChangeSets that used the older non-normalized paths, for eg with a classpath:<some/file/path>, in them to ignore the update and the checksum is never saved.

problem. That's something that needs to be addressed regardless.

However, I don't think making a change to how we store and manage the path is the approach we can take. 

Our overall strategy is to have the stored path a normalized path and consistent path regardless of the OS and version of liquibase, but there have been improvements (and sometimes bugs) in how that normalization happens over the years. So while  we try to keep that stored value correctly normalized (because people do reporting and comparisons outside of liquibase), when actually determining equality when finding which changesets have ran or not we do a more fuzzy matching. 

So I see the cause for the bug being that the "upgrade the checksums" logic doesn't have that fuzzy matching logic in there like it should, not that we need to be changing how we store the value. Partly because we want to stay consistently normalized in what we store for user expectation, but also because it should "just work" for users instead of making them have to think about another global setting.

To fix that, we may have to make the "update checksum" logic not a simple "update where" sql but instead selecting all rows and then checking them using the java-side code. Or, we could just select the old-versioned checksums and if/when there are any of those re-normalize those potentially "legacy" formatted values and update them before doing the where check?

vuparkar:
Thank you for the review comments

> So I see the cause for the bug being that the "upgrade the checksums" logic doesn't have that fuzzy matching logic in there like it should, not that we need to be changing how we store the value. Partly because we want to stay consistently normalized in what we store for user expectation, but also because it should "just work" for users instead of making them have to think about another global setting.
> 
Agreed the root issue here is that liquibase has changed how the file name is now stored in the Changelog.

Unfortunately since this has been released for a while, we will now have both types of users, ones with legacy non normalized filepaths and others with the new normalized filepaths.

Adding this flag was an attempt to ensure that the new users still remain unaware of normalization (until there is a future breaking change) and the legacy users can still upgrade liquibase library with no disruption to the changelog.


> To fix that, we may have to make the "update checksum" logic not a simple "update where" sql but instead selecting all rows and then checking them using the java-side code. Or, we could just select the old-versioned checksums and if/when there are any of those re-normalize those potentially "legacy" formatted values and update them before doing the where check?

Wouldnt the first option slow down the entire process esp for big changesets when checking in java side and especially as it would be run for everyone whether they need it or not.

Do you think it a good idea to provide the second option as a "command" to liquibase to allow users control over when to upgrade their changelog path as a distinct operation in their workflow?

nvoxland:
It would be nice to avoid a separate command if at all possible, since that would require users understanding what Liquibase is doing behind the scenes. It's nicest to not have to have them care and/or have to coordinate that. There is also a lot of use cases around "embedded liquibase" like in spring boot etc. where people can't call commands as easily.

I'm wondering if the trick is to better track the stored path way back when we originally read it from the databasechangelog info?

The UpdateChangeSetChecksumStatement is created from AbstractChangeLogHistoryySerivce.replaceChecksum() which is called from getRunStatus() and upgradeChecksums() and in both cases we had the value stored in the table but end up not preserving all the way to the spot where UpdateChangeSetChecksumGenerator needs it. Maybe optionally pass along RanChangeSet's `storedChangeLog` to UpdateChangeSetChecksumStatement and use that in the where clause if it's set?

nvoxland:

I haven't heard back from @vuparkar for a while, so I'm moving this PR along as half PR, half bug report. The base problem that is being reported should be addressed one way or the other. But I think we need to find a way that is transparent and handled automatically by liquibase and not something that requires flags or understanding by the user.

Questions I have:
- None

Potential risks:
- None

- Nothing

liquibase-core/src/main/java/liquibase/sqlgenerator/core/UpdateChangeSetChecksumGenerator.java
```diff
@@ -34,11 +36,20 @@ public Sql[] generateSql(UpdateChangeSetChecksumStatement statement, Database da
                     .setWhereClause(database.escapeObjectName("ID", LiquibaseColumn.class) + " = ? " +
                             "AND " + database.escapeObjectName("AUTHOR", LiquibaseColumn.class) + " = ? " +
                             "AND " + database.escapeObjectName("FILENAME", LiquibaseColumn.class) + " = ?")
-                    .addWhereParameters(changeSet.getId(), changeSet.getAuthor(), changeSet.getFilePath());
+                    .addWhereParameters(changeSet.getId(), changeSet.getAuthor(), getFilePathToUseInWhereClause(changeSet));
```
filipelautert, line 39: If we always use the now filled changeSet.getStoredFilePath() instead of changeSet.getFilePath() would this problem be solved?

vuparkar:
Yes. We just need to ensure the stored file path attribute is always set to the configured path and never modified. Here that's done in 2 places maybe there is a better place to update this attribute

Get Outlook for Android<https://aka.ms/AAb9ysg>
________________________________

filipelautert:
Hello @vuparkar ! Thanks for the PR. 
We are working on checksums upgrades, and your PR helped us to fix it. We created a similar PR based on yours -> https://github.com/liquibase/liquibase/pull/4210 .

PR #3496, opened 2022-11-22
match databases with numbers (like H2) in test filter

StevenMassaro:
## Impact

## Description

The test filtering logic needed to be expanded to match against databases that have numerics in their name, like H2.

## Things to worry about

None - this is a test only change.


suryaaki2 (approved):

StevenMassaro:
> Change is fine but seems that test-harness needs to be fixed to match it.

It's unlikely that the test-harness is using this code. I added this code months ago to allow developers to run a single CommandTest, rather than running the entire CommandTest suite. Take a look at the "filtering" section of the liquibase-pro repo readme for more information.

PR #3500, opened 2022-11-23
Define and reuse constants. Use an empty array styles to convert a collection to an array.

arturobernalg:
## Impact

## Description

Define and reuse constants. Use an empty array styles to convert a collection to an array.
There are two styles to convert a collection to an array:
 
A pre-sized array, for example, c.toArray(new String[c.size()])
An empty array, for example, c.toArray(new String[0])

In older Java versions, using a pre-sized array was recommended, as the reflection call necessary to create an array of proper size was quite slow.
However, from  OpenJDK 6, this call was intrinsified, making the performance of the empty array version the same, and sometimes even better, compared to the pre-sized version. 

## Things to be aware of

N/A

## Things to worry about

N/A

## Additional Context

N/A


nvoxland:

Yes, changes make sense. Getting the size for an array is unnecessary nowadays and a shared constant is good.

Questions I have:

- Are the constants defined in the right spots? Maybe could put them on the interfaces they are an array of? Like Statement.EMPTY_STATEMENT_ARRAY and Change.EMPTY_CHANGE_ARRAY etc?

Potential risks:

- None, if it's just changing new Object[size] instantiations to constants.

- Nothing

MalloD12:

This PR looks good to me. I have only applied a few minor changes of moving some constants from an abstract class to the related interfaces. From the Liquibase project, this looks ok (all tests passed), but before approving it I would wait until all checks are green.

Thanks again @arturobernalg for your contribution, I really like this kind of change to reuse lines of code and improve readability.

### Things to be aware of:
- None

### Things to worry about:
- None

MalloD12 (approved):
Approved!

arturobernalg:
thank you all.

PR #3506, opened 2022-11-24
Fixes behaviour of includeAll by keeping the trailing slash on the path

filipelautert:
## Impact

## Description

PR #3355 aimed to fix issues related to "./" and "../" , but by using Paths.get it is also removing trailing slashs. The changed method has logic to append the trailing slash, but it is located above the new code - so it baceme useless. This PR moves the code that fixes the path to be executed after Paths.get.

## Things to be aware of

- Fixes includeAll on S3 extension.
- Tests restored as they were before #3355 

## Things to worry about

- None

## Additional Context

- Issue raised by S3 extension tests.

### DIFF
--- a/liquibase-core/src/main/java/liquibase/changelog/DatabaseChangeLog.java
+++ b/liquibase-core/src/main/java/liquibase/changelog/DatabaseChangeLog.java
@@ -652,10 +652,10 @@ public class DatabaseChangeLog implements Comparable<DatabaseChangeLog>, Conditi
                             .replace("\\", "/");
                 }
 
+                path = path.replace("\\", "/");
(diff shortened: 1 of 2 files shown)


suryaaki2 (approved):

abrackx (approved):

XDelphiGrl (approved):
PR addresses a regression introduced while fixing the path shown in some output messages and the value stored in DATABASECHANGELOG tracking table when `include` or `includeAll` changesets specify `relativeToChangelog=true`. The original fix was too broad, and stripped trailing as well as leading slashes; only leading should be removed. This fix restores the correct behavior of keeping trailing slashes.
* S3 extension test that exposed the bug is fixed by this PR. 
* New unit test added.

APPROVED

PR #3518, opened 2022-11-30
Avoid ClassCastException when loading LogService from Scope

mattbertolini:
## Impact

## Description

This PR fixes a bug where a `ClassCastException` occurs when alternative logger implementations are on the classpath when the CLI is used. This was [previously fixed](https://github.com/liquibase/liquibase/blob/master/liquibase-core/src/main/java/liquibase/integration/commandline/Main.java#L290-L293) on the older legacy `Main` class but the newer `LiquibaseCommandLine` still has the issue.

Motivation is to avoid issues like [this](https://github.com/mattbertolini/liquibase-slf4j/issues/16) from being created.

## Things to worry about

None that I know of. This change is backwards compatible.

### DIFF
--- a/liquibase-cli/src/main/java/liquibase/integration/commandline/LiquibaseCommandLine.java
+++ b/liquibase-cli/src/main/java/liquibase/integration/commandline/LiquibaseCommandLine.java
@@ -632,9 +632,11 @@ public class LiquibaseCommandLine {
     private void configureLogging(Level logLevel, String logFile) throws IOException {
         configuredLogLevel = logLevel;
 
-        final JavaLogService logService = (JavaLogService) Scope.getCurrentScope().get(Scope.Attr.logService, LogService.class);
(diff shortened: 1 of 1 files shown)


nvoxland:

Change makes sense to me, we don't want to be assuming a specific logger when the CLI code is running.

Questions I have:
- None

Potential risks:
- None

- Nothing

StevenMassaro (approved):

filipelautert (approved):

XDelphiGrl (approved):
This PR addresses a bug impacting extensions such as [https://github.com/mattbertolini/liquibase-slf4j/](https://github.com/mattbertolini/liquibase-slf4j/) that rely on the Liquibase logging framework. 
 * The failing test harness cases are due to an environmental issue in hsql and firebird setup.

APPROVED

Issue #3269, opened 2022-09-12
 MD5SUM updating at every execution


doakd:
## Environment

**Liquibase Version**: Community 4.14

**Liquibase Integration & Version**: CLI

**Liquibase Extension(s) & Version**: Oracle 4.14, Teradata 4.14, neo4j 4.14, netezza 4.14 (dbeaver extension)

**Database Vendor & Version**: Oracle 12.1.0.2

**Operating System Type & Version**: Linux 5.4.0-1084-aws #91~18.04.1-Ubuntu SMP x86_64 GNU/Linux

**Infrastructure Type/Provider**: AWS

## Description

I have a customer who has a large number of changesets that get the MD5SUM updated in databasechangelog at the start of every deployment, even though the changesets have not been modified and the changesets are not actually executed.  The Liquibase output looks like this:

```
[2022-09-02 15:46:24] INFO [liquibase.executor] UPDATE TST4CIS.DATABASECHANGELOG SET MD5SUM = '8:cf74d98bf9923254d88d77efba90e1ec' WHERE ID = 'ads_cisd015.pks' AND AUTHOR = 'xxx' AND FILENAME = 'Packages/db-changelog-1.xml'
[2022-09-02 15:46:24] INFO [liquibase.executor] 1 row(s) affected
[2022-09-02 15:46:24] INFO [liquibase.executor] UPDATE TST4CIS.DATABASECHANGELOG SET MD5SUM = '8:a1d70adee4b5b830d4b5ab103c371548' WHERE ID = 'ads_cisd109.pks' AND AUTHOR = 'xxx' AND FILENAME = 'Packages/db-changelog-1.xml'
[2022-09-02 15:46:24] INFO [liquibase.executor] 1 row(s) affected
[2022-09-02 15:46:24] INFO [liquibase.executor] UPDATE TST4CIS.DATABASECHANGELOG SET MD5SUM = '8:859cb9b9b4174f5322329ee434232c55' WHERE ID = 'bobfixtca.pkb' AND AUTHOR = 'xxx' AND FILENAME = 'Packages/db-changelog-1.xml'
[2022-09-02 15:46:24] INFO [liquibase.executor] 1 row(s) affected
```

It's strange because it's not all of his changesets, only the changesets that contain PL/SQL (packages, procedures, trigggers, etc) , executed via sqlFile, and the files have .pkb, .pks, .prc, and .sql extensions.  

This seems to have started after two changes we made:
1. Upgraded to 4.12
[...]

PR #3530, opened 2022-12-02
Fixes md5 generation of SQLVisitors that was changed on 4.16.0

filipelautert:
## Impact

## Description

On Liquibase 4.16, PR #2971 introduced "contextFilter" replacement field. As it is used to generate the md5sum, the md5sums changed. 
This PR 'fixes' those fields names when calculating the md5 to keep compatibility with Liquibase < 4.16 . 

## Things to be aware of

- MD5Sums generated using Liquibase 4.16.0 -> 4.18.0 with changeset containing SqlVisitors using  "context", "contexts" 
or "contextFilter" are different from the other versions.

## Things to worry about

- None.

## Additional Context

- Reported on issue #3269, but it is not same problem.

### DIFF
--- a/liquibase-core/src/main/java/liquibase/sql/visitor/AbstractSqlVisitor.java
+++ b/liquibase-core/src/main/java/liquibase/sql/visitor/AbstractSqlVisitor.java
@@ -76,7 +76,12 @@ public abstract class AbstractSqlVisitor implements SqlVisitor {
 
     @Override
     public CheckSum generateCheckSum() {
-        return CheckSum.compute(new StringChangeLogSerializer().serialize(this, false));
+        String stringToMd5 = new StringChangeLogSerializer().serialize(this, false);
+        if (this.getContextFilter() != null) {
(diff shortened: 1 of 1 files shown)


suryaaki2 (approved):

liquibase-core/src/main/java/liquibase/sql/visitor/AbstractSqlVisitor.java
```diff
@@ -76,7 +76,12 @@ public void setLabels(Labels labels) {
 
     @Override
     public CheckSum generateCheckSum() {
-        return CheckSum.compute(new StringChangeLogSerializer().serialize(this, false));
+        String stringToMd5 = new StringChangeLogSerializer().serialize(this, false);
```
XDelphiGrl, line 79: Should `labels` and `labelsFilter` also be included in this change? Or are labels not included in the checksum calculation?

CC  @filipelautert

filipelautert:
Will be handled at https://github.com/liquibase/liquibase/issues/3549 .

PR #3535, opened 2022-12-05
Correctly handle indexes with descending columns in snapshot DAT-11447

wwillard7800:
In the snapshot command, indexes with columns marked as descending were not getting serialized in a way that would allow a changelog to be successfully generated.  With this fix, columns are now explicitly serialized as objects, rather than as references to column objects.


nvoxland (approved):
It overall looks good to me. 

My only question is if `belongsToIndex` is the right name for it. Partly it's a bit wordy and partly just looking at a Column object it maybe sounds like it's saying in general if the column belongs to an index vs more that "this specific object is describing how the column is used in an index". 

I don't have a better name just off hand, though. `forIndex`? `indexedVersion`? Something better?

filipelautert (approved):
I read @nvoxland comments about the field name but I don't have a better name too - everything I can think of would be longer. I'm fine with belongsToIndex or changing it to forIndex.

suryaaki2 (approved):

PR #3543, opened 2022-12-07
fix overwriteOutputFile parameter for GenerateChangelog (DAT-12036)

StevenMassaro:
## Impact

## Description

In #3064 , the logic to delete an existing file if `overwriteOutputFile` is set to true was accidentally removed. This effectively eliminated the `overwriteOutputFile` functionality of the generate changelog command. This PR readds that logic, but in a different location.


abrackx (approved):

filipelautert (approved):
I was going to suggest that flag overwriteOutputFile could be changed to overwriteChagelogOutputFile or something like that, but as we are restoring the behaviour that`s fine.

PR #3550, opened 2022-12-08
Delegate change set equality tests

droy-sandbox:
## Impact

## Description

Moving the responsibility of determining ChangeSet/RanChangeSet equality from the change log filters to the to the RanChangeSet.isSameAs() method.  The change is logically equivalent to what existed before but centralizes the equality logic to the RanChangeSet class.

This was done so that extension developers can override the ChangeSet and RanChangeSet classes to provide a different behavior by overriding the isSameAs() and toString() methods.  My particular use case is identifying change sets only by id to support refactoring of database change logs.  This allows me to move change sets around without having them mis-identified as "new" change sets.

## Things to be aware of

N/A

## Things to worry about

N/A

## Additional Context

N/A


nvoxland:

Code cleanup is always appreciated, thanks!

Questions I have:
- None

Potential risks:
- None

- Nothing

MalloD12 (approved):
**Approved**.

This is looking good to me and all checks are green. Thanks @droy-sandbox for this clean up, very appreciate it.

filipelautert (approved):
Seems fine, removes redundant methods.

suryaaki2 (approved):

XDelphiGrl (approved):
PR removes redundant logic from several changeset classes, relying instead upon the logic in the `isSameAs` method in [https://github.com/liquibase/liquibase/blob/master/liquibase-core/src/main/java/liquibase/changelog/RanChangeSet.java](https://github.com/liquibase/liquibase/blob/master/liquibase-core/src/main/java/liquibase/changelog/RanChangeSet.java).

APPROVED

PR #3552, opened 2022-12-08
Bump up the max number of code points for JSON/YAML parser DAT-12657

wwillard7800:
Refactored the creation of the Yaml object in order to pass a LoaderOptions instance which sets the maximum number of code points higher so that larger JSON files can be parsed.


filipelautert (approved):

liquibase-core/src/main/java/liquibase/parser/core/yaml/YamlSnapshotParser.java
```diff
@@ -11,19 +11,25 @@
...
+import org.yaml.snakeyaml.representer.Representer;
+import org.yaml.snakeyaml.resolver.Resolver;
 
 import java.io.InputStream;
 import java.io.InputStreamReader;
 import java.util.Map;
 
 public class YamlSnapshotParser extends YamlParser implements SnapshotParser {
 
+    public static final int CODE_POINT_LIMIT = 9 * 1024 * 1024;
```
StevenMassaro, line 27: Why not just make this MAX_VALUE? I see no reason to limit the size on our end.

liquibase-core/src/main/java/liquibase/parser/core/yaml/YamlSnapshotParser.java
```diff
@@ -11,19 +11,25 @@
...
+
     @SuppressWarnings("java:S2095")
     @Override
     public DatabaseSnapshot parse(String path, ResourceAccessor resourceAccessor) throws LiquibaseParseException {
-        Yaml yaml = new Yaml(new SafeConstructor());
+        Yaml yaml = createYaml();
```
StevenMassaro, line 32: We create a `Yaml` object in the pro code too, I wonder if we should increase the code point limit there as well.

liquibase-core/src/main/java/liquibase/parser/core/yaml/YamlSnapshotParser.java
wwillard7800: Agreed.

liquibase-core/src/main/java/liquibase/parser/core/yaml/YamlSnapshotParser.java
wwillard7800: I'll take a look.

StevenMassaro: This may be best done with another story, up to you.

suryaaki2 (approved):

PR #3553, opened 2022-12-08
No need redundant conditions.

arturobernalg:
## Impact

## Description

No need redundant conditions. 
Condition like obj != null && obj instanceof String, the null-check is redundant as instanceof operator implies non-nullity.
Same case with obj != null && "anyString".equals(obj).

https://docs.oracle.com/javase/specs/jls/se11/html/jls-15.html#jls-15.20.2


nvoxland:

Thanks for the code cleanup

Questions I have:
- None

Potential risks:
- None

- Nothing

filipelautert (approved):

XDelphiGrl (approved):
PR removes extraneous `x != null` checks from `if` statements where there is also a check for `x instanceof aClass` or `"x".equals(y)`. 

APPROVED

suryaaki2 (approved):

Issue #2018, opened 2021-08-03
Create ThreadLocal ScopeManager


nvoxland:
## Overview

The current imlementation of ScopeManager is based around a singleton instance. Sometimes that works well, sometimes not. Alternative ScopeManager implementations are left as an exercise for the user (such as in#1768).

A ThreadLocalScopeManager would allow liquibase to run concurrently in separate threads.

## Requirements

Liquibase should continue to use the SingletonScopeManager as the default, but provide a way to optionally activate the separate ThreadLocalScopeManager. Not sure the best way off hand, the LiquibaseConfiguration system is too late in the process. May need to depend on a system property or even a registered subclass of ThreadLocalScopeManager? _Maybe other ideas?_

Will need to add tests to ensure that the threading is working correctly, and that there is nothing else in liquibase that conflicts in a multi-threaded environment. The work will be as much about ensuring Liquibase overall works in concurrent threads as that the Scope object supports it.

## Implementation

The example in #1768 is a good start.

jamey-clari:
I'd actually argue that the thread-local implementation should likely be the preferred default as most users of Liquibase are in an enterprise type environment and it's not unlikely that multiple threads might be using Liquibase tooling against multiple databases concurrently.  Aside from this a pluggable thread-safe alternative to the default is a big step in the right direction.  Thanks to all for getting this going as we will definitely be able to make use of it.

PR #3561, opened 2022-12-10
Make updates of many un-Scoped factories state safer

ettavolt:
Also less locking when using Scope.

## Impact

## Description

Follow up #2018.

Migrating 11 databases with four threads rarely led to an NPE due to concurrency in ReflectionSerializer.
I've decided to not stop with just that but also look at all static modifiable fields and make access to them safer.

## Things to be aware of

Mostly it's CopyOnWriteArrayList and ConcurrentHashMap with compute/IfAbsent.
I've tried hard to preserve even "unregister" order (sometimes it deletes the last instance, rather than the first one).
I've run this version several times with different number of threads (incl.11) for those 11 databases (tmpfs FTW).

## Things to worry about

There is a couple if (get() == null) put() usages of ConcurrentHashMap I've fixed, but there could be more.


nvoxland:
Thanks for working through that!

**Notes for reviewer**

Quick glance looked like thread-safe cleanup of code with no functionality or API changes. Removed some usage of SmartMap, did it remove all of it and we could remove that class? Or still some left and/or do we want to keep SmartMap around regardless (or at least deprecate)?

Questions I have:
- None

Potential risks:
- None

- Nothing

ettavolt:
> Removed some usage of SmartMap, did it remove all of it and we could remove that class? Or still some left and/or do we want to keep SmartMap around regardless (or at least deprecate)?

SmartMap is still used by AbstractExtensibleObject. In a one private field, accessible through three methods. Relevant part may be inlined, I guess.

If you'd like, I can do this too. I enjoy erasing deprecated code. :smile:

nvoxland:
If you're good with inlining the usage on this PR @ettavolt I'd appreciate it :) 

Probably can't completely remove SmartMap yet, since it might be used by extensions. But you could mark it as deprecated for now at least.

MalloD12 (commented):
@nvoxland , I have reviewed this PR and changes made here looks consistent to me across the different updated files. But from my current java/product knowledge I cannot guarantee there is not a potential risk. Build and tests have been executed successfully.

Could you please review this PR and approve it if looks good to you too?

Thanks,
Daniel.

MalloD12:
Hi @ettavolt, after triggering our GH jobs I saw sonar is failing due to a few [bugs/code smells](https://sonarcloud.io/project/issues?resolved=false&types=CODE_SMELL&sinceLeakPeriod=true&pullRequest=3561&id=liquibase). Would you mind taking a look to it?

Thanks,
Daniel.

ettavolt:
Sure!

ettavolt:
Sorry, I've already messed history with my shallow-copy approach, before I realized that it would have been better to append rather than squash.

MalloD12:
Hi @ettavolt, 

I have resolved some merge conflicts, and tried to build again but am still seeing some sonar issues/code smells. Do you think you can fix them, so we can move forward with this PR?

Thanks,
Daniel.

ettavolt:
Yep!
(This time I'll try to make your reviewing job easier 😀)

liquibase-core/src/main/java/liquibase/diff/output/StandardObjectChangeFilter.java
```diff
@@ -36,8 +36,12 @@ public class StandardObjectChangeFilter implements ObjectChangeFilter {
     public StandardObjectChangeFilter(FilterType type, String filter) {
         this.filterType = type;
         if (databaseObjects.isEmpty()) {
```
MalloD12, line 38: @ettavolt is the first `IF` required because we have the same comparison inside of the `sinchronized` block?

liquibase-core/src/main/java/liquibase/statement/ExecutablePreparedStatementBase.java
```diff
@@ -122,50 +102,21 @@ public void execute(PreparedStatementFactory factory) throws DatabaseException {
...
-    }
-
     protected void executePreparedStatement(PreparedStatement stmt) throws SQLException {
         if (database instanceof PostgresDatabase) {
             //postgresql's default prepared statement setup is slow for normal liquibase usage. Calling with QUERY_ONESHOT seems faster, even when we keep re-calling the same prepared statement for many rows in loadData
             try {
-                if (executeWithFlagsMethod == null) {
-                    executeWithFlagsMethod = stmt.getClass().getMethod("executeWithFlags", int.class);
-                    executeWithFlagsMethod.setAccessible(true);
+                if (postgresExecuteWithFlagsMethod == null) {
```
MalloD12, line 109: Same thing here, we have the same comparison outside and inside of the synchronized block.

ettavolt:
Yup, it's  https://en.m.wikipedia.org/wiki/Double-checked_locking

MalloD12:
**Approved**.

Besides, I am not a Thread safe expert, but I have seen code changes made and compared with what we had before and the changes look good to me. Some of the changes made have been to start using CopyOnWriteArrayList and ConcurrentHashMap data structures which are thread-safe and which allow us to use methods like `compute` and `IfAbsent`. As Nathan mentioned there are no code functionality or API changes made on this PR so I think we are ok to move forward. 

Build and tests passed successfully.

This is a good clean-up @ettavolt, thanks for taking care of this. And, to have fixed sonar issues/code smells.   

### Things to be aware of:
- None

### Things to worry about:
- Besides the testing, you have performed @ettavolt . I'm wondering whether we can add some unit tests around these changes. For example for BaseSerializerFactory where we have a couple of methods that are used in several places.

MalloD12 (approved):

liquibase-core/src/main/java/liquibase/diff/output/StandardObjectChangeFilter.java
ettavolt: Rereading Wiki, I thought there could be a problem with nothing being `volatile`, potentially giving different results for the second conditional (which is inside `synchronized`). How, java.util.SynchornizedCollections don't use any `volatile` to provide thread-safety guarantee.

That said, current code relies on the fact that there is just one instance of writing, which might intendedly change in the future. To be on a safe side, let me initialize the field to null, and then set it to an unmodifiable list.

ettavolt:
As for testing BaseSerializerFactory: I think it's pretty nice covered by ChangeLogSerializerFactoryTest, which seem to call all four public methods. Just in case, I'm making BaseSerializerFactory package-local, so that it is not considered an API.

liquibase-core/src/main/java/liquibase/statement/ExecutablePreparedStatementBase.java
MalloD12: @ettavolt don't we need to also make the same change you made on StandardObjectChangeFilter here too by making `executeWithFlagsMethod` a volatile field as well?

ettavolt: `postgresExecuteWithFlagsMethod` is already defined as `volatile`. (It was earlier version that used `executeWithFlagsMethod` name).

MalloD12: You're totally right, I think I saw it wrong. Thank you!

nvoxland (approved):
Nothing looks surprising or unexpected to me. Thanks!

ettavolt:
I guess that `com.datical.liquibase.ext.diff.FormattedDiffObject.getSerializableFields` should filter retrieved fields collection via stream(), or first create a new one off it.
It might even fix some bug!

ettavolt:
@XDelphiGrl, can you tell me from where is `com.datical.liquibase.ext.diff.FormattedDiffObject.getSerializableFields` getting the collection it tries to remove elements from? I can make that method to always return new modifiable collections.

filipelautert:
> @XDelphiGrl, can you tell me from where is `com.datical.liquibase.ext.diff.FormattedDiffObject.getSerializableFields` getting the collection it tries to remove elements from? I can make that method to always return new modifiable collections.

Hello @ettavolt - sure,  it gets from super:

```
    @Override
    public Set<String> getSerializableFields() {
        Set<String> fields = super.getSerializableFields();
        fields.remove("objectTypeHeader");
        return fields;
    }
```

and super is `AbstractLiquibaseSerializable `

liquibase-standard/src/main/java/liquibase/plugin/AbstractPluginFactory.java
```diff
@@ -22,21 +24,18 @@ protected AbstractPluginFactory() {
...
      * <p>
-     * However, if there is a {@link Scope} key of "liquibase.plugin.${plugin.interface.class.Name}", an instance of that class will always be ran first.
+     * However, if there is a {@link Scope} key of "liquibase.plugin.${plugin.interface.class.Name}",
+     * an instance of that class is returned or {@link RuntimeException} is thrown.
      *
+     * @throws RuntimeException when cannot find registred plugin for specified name.
      * @return null if no plugins are found or have a priority greater than zero.
      */
-    protected T getPlugin(final Object... args) {
+    protected T getPlugin(final PriorityArgsClosure<? super T> chooser) {
```
ettavolt, line 38: That's probably my third time using `super` type bound. ☺

MalloD12:
Hi @ettavolt, 

Could you please look at the Postgres integration tests failing (https://github.com/liquibase/liquibase/actions/runs/5394370602/jobs/9795625506?pr=3561), and see whether the failures are related to your changes or not?

Thanks,
Daniel.

ettavolt:
Likely they are, because there are successful builds [before](https://github.com/liquibase/liquibase/actions/runs/5393962323) and [after](https://github.com/liquibase/liquibase/actions/runs/5401309105) that.

I'll try to run this integration test locally and see what it does wrong.

kevin-atx:
@ettavolt - any luck getting the integration tests to pass?

ettavolt:
I didn't see new tests results. I'll merge the mainline and fix the conflicts.

ettavolt:
If you are referring to PostgreSQL integration tests: those were fixed like two months ago.

There were merge conflicts with master, I've resolved these now.

PR #3563, opened 2022-12-12
Fixes generate-changelog table/view comment and table/view column comment generation for MSSQL (DAT-12045)

abrackx:
## Impact

## Description
This PR does a few things:

- Reworks how we find views and view comments when taking a snapshot for MSSQL.
  - This is because previously we were relying on the jdbc connection metadata to find views and view comments. The code backing that reads from `sp_tables` which [does not return any remarks](https://learn.microsoft.com/en-us/sql/relational-databases/system-stored-procedures/sp-tables-transact-sql?view=sql-server-ver16#result-sets) despite it being included as a column in the result set.
- Adds support for view comments and view column comments for MSSQL.
- Adds field ~`isView`~ `ownerType` which is a new enum to `SetColumnRemarksChange` and `SetColumnRemarksStatement` to differentiate between view column comments and table column comments. 
  - The syntax is different between the two for MSSQL and we needed some method to determine which we were using.
- Reworks table comment and table column comment generated sql for MSSQL. 
  - See `Things to be aware of`

## Things to be aware of
Changes how existing SQL Server table comments as well as column comments are generated. I switched the syntax to something you are able to use multiple times in a single batch by avoiding declaring any variables. This is because I was encountering a batch variable scoping errors such as `The variable name '@TableName' has already been declared. Variable names must be unique within a query batch or stored procedure.` when using the generated table and view comments together in the same changeset.

## Things to worry about
See above

## Additional Context
You can reproduce these issues with the changeset used in the new test case.


liquibase-core/src/main/java/liquibase/sqlgenerator/core/SetTableRemarksGenerator.java
```diff
@@ -37,44 +37,35 @@ public Sql[] generateSql(SetTableRemarksStatement statement, Database database,
...
+					" END " +
+					" ELSE " +
+					" BEGIN " +
+					" EXEC sys.sp_addextendedproperty @name = N'MS_Description'" +
+					" , @value = N'" + remarksEscaped + "'" +
+					" , @level0type = N'SCHEMA'" +
+					" , @level0name = N'" + schemaName + "'" +
+					" , @level1type = N'TABLE'" +
+					" , @level1name = N'" + tableName + "'" +
+					" END";
```
StevenMassaro, line 68: Can you summarize the changes you made to this block of SQL?

liquibase-core/src/test/java/liquibase/verify/saved_state/compareGeneratedSqlWithExpectedSqlForMinimalChangesets/setColumnRemarks/mssql.sql
```diff
@@ -2,4 +2,4 @@
...
+IF EXISTS(  SELECT extended_properties.value FROM SYS.EXTENDED_PROPERTIES WHERE major_id = OBJECT_ID('dbo.person') AND name = N'MS_DESCRIPTION' AND minor_id = ( SELECT column_id FROM SYS.COLUMNS WHERE name = 'id' AND object_id = OBJECT_ID('dbo.person')) ) BEGIN  EXEC sys.sp_updateextendedproperty @name = N'MS_Description' , @value = N'A String' , @level0type = N'SCHEMA' , @level0name = N'dbo' , @level1type = N'TABLE' , @level1name = N'person' , @level2type = N'COLUMN' , @level2name = N'id' END  ELSE  BEGIN  EXEC sys.sp_addextendedproperty @name = N'MS_Description' , @value = N'A String' , @level0type = N'SCHEMA' , @level0name = N'dbo' , @level1type = N'TABLE' , @level1name = N'person' , @level2type = N'COLUMN' , @level2name = N'id' END;
```
StevenMassaro, line 5: @wwillard7800 Are we allowed to make changes to these baselines?

StevenMassaro (approved):

liquibase-core/src/main/java/liquibase/sqlgenerator/core/SetTableRemarksGenerator.java
abrackx: Sure! I removed the query variables from the query. Originally would set up a table name variable `DECLARE @TableName SYSNAME; SET @TableName = N'BLAH';` I have removed this in favor of directly using the table name etc in the query. 
So before we had: 
```
DECLARE @TableName SYSNAME
set @TableName = N'person';
DECLARE @FullTableName SYSNAME;
SET @FullTableName = N'dbo.person';
DECLARE @MS_DescriptionValue NVARCHAR(3749);
SET @MS_DescriptionValue = N'A String';
DECLARE @MS_Description NVARCHAR(3749)
set @MS_Description = NULL;
SET @MS_Description = (SELECT CAST(Value AS NVARCHAR(3749)) AS [MS_Description]
                       FROM sys.extended_properties AS ep
                       WHERE ep.major_id = OBJECT_ID(@FullTableName)
                         AND ep.name = N'MS_Description'
                         AND ep.minor_id = 0);
IF @MS_Description IS NULL
    BEGIN
        EXEC sys.sp_addextendedproperty @name = N'MS_Description', @value = @MS_DescriptionValue,
             @level0type = N'SCHEMA', @level0name = N'dbo', @level1type = N'TABLE', @level1name = @TableName;
    END
ELSE
    BEGIN
        EXEC sys.sp_updateextendedproperty @name = N'MS_Description', @value = @MS_DescriptionValue,
             @level0type = N'SCHEMA', @level0name = N'dbo', @level1type = N'TABLE', @level1name = @TableName;
    END;
```
Now it looks like: 
```
IF EXISTS(SELECT extended_properties.value
          FROM SYS.EXTENDED_PROPERTIES
          WHERE major_id = OBJECT_ID('dbo.person')
            AND name = N'MS_DESCRIPTION'
            AND minor_id = 0)
    BEGIN
        EXEC sys.sp_updateextendedproperty @name = N'MS_Description', @value = N'A String', @level0type = N'SCHEMA',
             @level0name = N'dbo', @level1type = N'TABLE', @level1name = N'person'
    END
ELSE
    BEGIN
        EXEC sys.sp_addextendedproperty @name = N'MS_Description', @value = N'A String', @level0type = N'SCHEMA',
             @level0name = N'dbo', @level1type = N'TABLE', @level1name = N'person'
    END;
```
The reason I had to change this is we generate table comments and table column comments together. The repeated usage of the same `@TableName` even when redeclared throws errors.

liquibase-core/src/main/java/liquibase/change/core/SetColumnRemarksChange.java
```diff
@@ -17,6 +17,7 @@ public class SetColumnRemarksChange extends AbstractChange {
     private String columnName;
     private String remarks;
     private String columnDataType;
+    private boolean isView = false;
```
nvoxland, line 20: Adding a not null/defaulted field to the change type like this is going to modify the definition used to generate checksums. So I think this will break everyone using setColumnRemark currently.

I think you have to leave it as Boolean and leave it null by default.

I also wonder if instead of a more specific isView with a boolean it should be an enumeration like `ownerType` or `relationType` that currently can have values of TABLE or VIEW but later on could be COLLECTION or whatever else different databases may want to add columns to?

liquibase-core/src/test/java/liquibase/verify/saved_state/compareGeneratedSqlWithExpectedSqlForMinimalChangesets/setColumnRemarks/mssql.sql
nvoxland: As long as we know that the new SQL is valid and works it's fine to change. Having it in a test like this lets us see that theColumnRemarks SQL changed and how, but always make sure it's working as you expect.

nvoxland (changes requested):

liquibase-core/src/main/java/liquibase/change/core/SetColumnRemarksChange.java
StevenMassaro: This makes me think that we should have tests around the checksum generation code so that we can be sure we don't accidentally break it going forward with changes like this.

StevenMassaro (approved):

liquibase-core/src/test/groovy/liquibase/serializer/core/string/StringChangeLogSerializerTest.groovy
```diff
@@ -309,6 +310,8 @@ public class StringChangeLogSerializerTest extends Specification {
                 //nothing
             } else if (field.getType().equals(LoadDataChange.LOAD_DATA_TYPE.class)) {
                 //nothing
+            } else if (field.getType().equals(ColumnOwnerType.class)) {
+                //nothing
```
abrackx, line 314: Is this the correct way to handle this new enum field (that is ignored in serialization by overriding `getSerializableFields()`) @nvoxland @filipelautert? The test `general field serialization check` fails without it.

filipelautert: I think it's fine, otherwise you would need to add it to most of the tests above it.

liquibase-core/src/main/java/liquibase/change/core/SetColumnRemarksChange.java
nvoxland: Yes, very much we do, in an automatically complete way. What we have now is too manual and scattered and one-off

liquibase-core/src/main/java/liquibase/util/ColumnOwnerType.java
```diff
@@ -0,0 +1,10 @@
+package liquibase.util;
+
+/**
+ * Used to indicate what type of relation owns a column.
+ */
+public enum ColumnOwnerType {
+    TABLE,
+    VIEW,
+    UNKNOWN
```
nvoxland, line 9: I don't think we should have UNKNOWN as a type. These definitions exist for us to know what to do, and a type of UNKNOWN is not something we can know what to do with, so it just adds code and user complexity.

The variables of this enum type can be null, so better to rely on that single way to define "don't know the column type" than introduce this.

liquibase-core/src/test/groovy/liquibase/serializer/core/string/StringChangeLogSerializerTest.groovy
nvoxland: It maybe seems best to actually set a value there to be able to test against instead of just ignoring it. Otherwise we're not really testing how it serializes a field of that type. You can maybe just randomly pick one like "table" for the test? 

But I'm also not sure what the test failure was. Was it failing because it was catching a real problem in the code and ignoring it just avoids the real bug? Or was it just "don't know what to use for a value?" and so giving value instead of ignoring it is fine? Or something else?

nvoxland (changes requested):

liquibase-core/src/main/java/liquibase/change/core/SetColumnRemarksChange.java
StevenMassaro: I logged a tech debt story: https://datical.atlassian.net/browse/DAT-12752

liquibase-core/src/main/java/liquibase/util/ColumnOwnerType.java
abrackx: Good point. I removed the "UNKNOWN" value.

liquibase-core/src/test/groovy/liquibase/serializer/core/string/StringChangeLogSerializerTest.groovy
abrackx: Updated to set the value in the test case. It was failing because it did not know what to do with the value.

suryaaki2 (approved):

nvoxland (approved):

PR #3564, opened 2022-12-12
Fully qualified path to the changelog or snapshot in S3 is not allowed, so needs better msg DAT-11478

diadiushko:
## Impact

## Description
Now if user inputs an s3 path as changelogFile, he is going to receive according error message.
## Things to be aware of
Now any `InvalidPathException` will be checked for an s3 path.


liquibase-core/src/main/java/liquibase/parser/core/xml/XMLChangeLogSAXParser.java
```diff
@@ -150,6 +151,15 @@ public void fatalError(SAXParseException exception) throws SAXException {
             }
 
             throw new ChangeLogParseException("Invalid Migration File: " + reason, e);
+        } catch (InvalidPathException e) {
+            if (e.getInput().startsWith("s3://")) {
```
StevenMassaro, line 155: Do you think we need this conditional? Could we throw this exception in all circumstances?

liquibase-core/src/main/java/liquibase/parser/core/xml/XMLChangeLogSAXParser.java
```diff
@@ -150,6 +151,15 @@ public void fatalError(SAXParseException exception) throws SAXException {
...
+                final String message = "ERROR: The file '" + e.getInput() + "' was not found. " +
+                        "The <changelog|snapshot> file property cannot be configured with " +
+                        "a fully qualified path, but must be a relative path on the property, " +
+                        "and any local or remote base of the path set on the '--search-path' property.";
+                throw new ChangeLogParseException(message);
```
StevenMassaro, line 160: Make sure you either log or include the original exception `e` with the new exception you are throwing. Log would probably be best.

dyadyushko: done!

liquibase-core/src/main/java/liquibase/parser/core/xml/XMLChangeLogSAXParser.java
```diff
@@ -150,6 +151,13 @@ public void fatalError(SAXParseException exception) throws SAXException {
             }
 
             throw new ChangeLogParseException("Invalid Migration File: " + reason, e);
+        } catch (InvalidPathException e) {
+            final String message = "ERROR: The file '" + e.getInput() + "' was not found. " +
+                    "The <changelog|snapshot> file property cannot be configured with " +
+                    "a fully qualified path, but must be a relative path on the property, " +
+                    "and any local or remote base of the path set on the '--search-path' property.";
+            Scope.getCurrentScope().getUI().sendErrorMessage("ERROR: " + e.getMessage(), e);
```
StevenMassaro, line 159: This is probably more appropriate in the logs than in the console, since the exception you're throwing on line 160 will show up in the console.

```suggestion
            Scope.getCurrentScope().getLog(getClass()).error("ERROR: " + e.getMessage(), e);
```

StevenMassaro (approved):

liquibase-core/src/main/java/liquibase/parser/core/xml/XMLChangeLogSAXParser.java
dyadyushko: good point!

dyadyushko: I think there is no `error` method, I am going to use `severe` then.

StevenMassaro (approved):
Perfect!

filipelautert (approved):

PR #3572, opened 2022-12-14
fix: snowflake has wrong priority

### LINKED ISSUE #3538: Snowflake Implements have error priority
## Environment

**Liquibase Version**: 4.17.2

**Liquibase Integration & Version**: gradle

**Liquibase Extension(s) & Version**:  hanadb-4.17.2

## Description

In version 4.12.0 support for Snowflake database has been moved from the external extension into the main Liquibase artifact.
but some implements like SequenceSnapshotGeneratorSnowflake , SchemaSnapshotGeneratorSnowflake  have a wrong priority
https://github.com/liquibase/liquibase/blob/256675bca43a089e32c940e10ed8ed7a5f9d66f6/liquibase-core/src/main/java/liquibase/snapshot/SnapshotGeneratorFactory.java#L76-L89

https://github.com/liquibase/liquibase/blob/93f3a7a828f0a683cdf9ead9feb7062358dbf43b/liquibase-snowflake/src/main/java/liquibase/snapshot/jvm/SequenceSnapshotGeneratorSnowflake.java#L13-L19

As you can see, **SequenceSnapshotGeneratorSnowflake** will add to **validGenerators** when we use hanadb (other db is ok)

Then when **getNextValidGenerator** we will get **SequenceSnapshotGeneratorSnowflake**

https://github.com/liquibase/liquibase/blob/9ee66e827221fb2a1138d70fe9ee7ceeef02bbac/liquibase-core/src/main/java/liquibase/snapshot/SnapshotGeneratorChain.java#L43

Then when call **getSelectSequenceSql**
https://github.com/liquibase/liquibase/blob/9ee66e827221fb2a1138d70fe9ee7ceeef02bbac/liquibase-core/src/main/java/liquibase/snapshot/jvm/SequenceSnapshotGenerator.java#L43

Then will call **super.getSelectSequenceSql**

https://github.com/liquibase/liquibase/blob/93f3a7a828f0a683cdf9ead9feb7062358dbf43b/liquibase-snowflake/src/main/java/liquibase/snapshot/jvm/SequenceSnapshotGeneratorSnowflake.java#L27-L36

Then will throw this exception

https://github.com/liquibase/liquibase/blob/93f3a7a828f0a683cdf9ead9feb7062358dbf43b/liquibase-core/src/main/java/liquibase/snapshot/jvm/SequenceSnapshotGenerator.java#L285

## Steps To Reproduce

Use liquibase version 4.17.2 and hana-ext 4.17.2, and use a liquibase changelog like
``` xml
<changeSet id="create-pkey-sequence" author="test">
        <createSequence sequenceName="tenant_lcm_event_seq" />
    </changeSet>
```

## Actual Behavior

Throw UnexpectedLiquibaseException("Don't know how to query for sequences on hana"); 

## Expected/Desired Behavior

Will create sequence successful

Yogurt-lei:
## Impact

## Description

FIxes #3538
Snowflake implementions merge to liquibase-core.jar with wrong priority


nvoxland:

If I remember right, the "priority" in snapshot is a bit odd in that it's used more for "adds to" than "replaces" logic?

The changes in this need to be made, since otherwise the snowflake code gets used for all databses. But they maybe also have to override replaces() and `return new Class[] { SchemaSnapshotGenerator.class };` ?

Questions I have:
- None

Potential risks:
- None

- Nothing

LonwoLonwo:
Hello @Yogurt-lei 

Would you like some help? Because I want to see these changes in the code too :)

filipelautert (approved):
Now that's funny. I was testing this PR and got an error saying that "liquibase don't know how to generate sequence for snowflake... etc" . And guess what: I had the bigquery extension on my lib directory, and it has exactly the same problem that this PR is fixing  🤣 :  https://github.com/liquibase/liquibase-bigquery/blob/main/src/main/java/liquibase/ext/bigquery/snapshot/jvm/BigQuerySequenceSnapshotGenerator.java

Removing bigquery extension fixed it and snowflake worked fine. So yes, this PR really fixes the problem AND now we can fix big query too based on it. Thanks!

XDelphiGrl (approved):
PR fixes a bug introduced by the migration of Snowflake from an extension into Core. The bug set an incorrect priority for Snowflake in some snapshot generator classes, resulting in exceptions running Liquibase with HanaDB. 

APPROVED

PR #3578, opened 2022-12-14
Issue 3562 - Required expression operator for context and label filters

### LINKED ISSUE #3562: Feature Request: contextFilter indicator that context is explicitly specified
## Description

Currently, where when no contexts are provided for a Liquibase run then all changesets in the provided changelog will be run, regardless of if they have contextFilter attributes or not. Essentially, contextFilter attributes are completely ignored if no contexts are provided for the run.

While this has merit in certain cases, it is not always ideal. Such a situation is pointed out in you own documentation [here](https://docs.liquibase.com/concepts/changelogs/attributes/contexts.html). In the test data situation noted in that documentation, if someone were to run the changelog without any contexts, test data would get installed. Instead, it would be nice to be able to require that the "test" context be provided in order for that changeset to be run. the contextFilter attribute already accepts "!", "AND", "OR", "(", and ")", so logical evaluation is already being done on the contextFilter. A new operator "@" (only used as an example... could be any character you wanted to use), could be used to designate that this context must be explicitly provided when Liquibase is run for the changeset to be run. For example, the attribute `contextFilter: @test` would make it so that a changeset would not be run if Liquibase were run without any contexts being provided.

## Steps To Reproduce

Create a Liquibase changelog and change set(s) with 1 or more contextFilter attributes specified.

## Actual Behavior

All of the change sets will be run.

## Expected/Desired Behavior

A new operator could be applied to contexts in the contextFilter attribute that would allow them to only be run if the context is explicitly provided when Liquibase is run.

jasonlyle88:
## Impact

## Description

Implements feature request from [issue 3562](https://github.com/liquibase/liquibase/issues/3562).

Updated the `ExpressionMatcher` class to add a new logical operator, `@`. This operator makes an expression required - as in it will not match an empty item set. Further updates were needed in the `ContextChangeSetFilter`, `LabelChangeSetFilter`, `ContextExpression`, and `LabelExpression` classes to allow for evaluation of context/label expressions against empty item sets.

Unit tests were also implemented for the new operator in `ExpressionMatcherTest.groovy` and `ContextChangeSetFilterTest.java`. Additionally, I discovered that the `LabelChangeSetFilter` class did not have any unit tests against it, so I created a `LabelChangeSetFilterTest.java` file for unit tests against that class. The unit tests started as a copy of the unit tests for the `ContextChangeSetFilter` class and were then properly adjusted to apply to the `LabelChangeSetFilter` class.

Fixes #3562 

## Things to be aware of

My original issue I raised was in regards to contexts, however since contexts and labels are implemented almost identically, and labels can benefit from the same additional functionality, I applied the required expression operator to both contexts and labels.

## Things to worry about

Running the Maven test lifecycle generated 78 new artifacts that I have also included in this pull request. Not sure if that is how you guys want things done or not, happy to change as necessary.


nvoxland:
This is an interesting approach to solving the "didn't specify a contexts/labelFilter" problem I hadn't thought of. Thanks, @jasonlyle88 !

My initial thought, though, is whether this is going to be surprising/unexpected for users, and in the end just drive people to always have to add `@` to all their labels/contexts?

Having a character to mean "this shouldn't match the 'empty' set" is great when there is just a selection of contexts/labels where you want to make sure it's applied, but is that a case people will hit often? For example, if I know that some environments don't have contexts set and that there are a couple changesets that REALLY need the "qa" context explicitly set, being able to say `contextFilter="@qa"` solves the problem. But it seems like you'd probably want to do `@qa` for **_all_** your contexts not just some? And same with `prod` or `staging` or whatever other contexts you have? 

It makes me feel like it needs to be a global setting to change the behavior so just using `contextFilter="qa"` is still enough? Or are there times I'm not thinking of where "some contexts/labels are required and some are not" will be a useful thing to handle?

jasonlyle88:
Hey @nvoxland, happy to contribute! Thanks for all the work you do here (I saw your name all over the git commit logs when I was working through the code!).

I had similar thoughts when I was proposing this change and took a while to consider them before creating the issue/coding the pull request. Below are my thoughts on the topic and why I ultimately still went with this approach.

For my first consideration, I took in the theory behind contexts/labels. The ideology between the two is similar, the difference between the two being who has control (aka, who writes the filter expressions).

With labels, the end user writes the filter expression. In this case, being able to specify a global switch determining if filters match an empty expression makes sense because the end user writes the filter expression for the labels and gets to decide if those filter expressions match an empty set of labels or not.

With contexts, the developers write the filter expressions. So, in the case of contexts, developers are supposed to be able to completely control what happens when an end user provides a list of contexts. The exact opposite control structure as labels. As such, we have what I see as an issue: it does not make sense for end users to have control of the filter expression matching an empty set or not because the end user is not the one writing the expression filter, the developer is. By providing the end user this global flag, we are providing a way for end users to run contexts in a potentially unsupported manner.

This differentiation in who writes the filter expression led me to believe that rather than having a global setting it is best left to the individual filter level.

My second consideration is roll out: it would be very easy to break an old change log if accidentally applying the new global setting to contexts in a project not intended to use the global setting. It would have to be more deliberate for an end user to go in and change the existing filter expression to add the `@` operator. Now, I realize that this isn't really a huge issue... end users should run things correctly. However... I've seen end users do lots of... silly... things over the years. I like keeping power in the developers hands where possible. So... yes, this is a slightly biased point, but one I wanted to at least list so it could be considered.

My third consideration was actual practical situations where this could be useful and developers would not just end up using the `@` operator on every label. I had a recent project with a setup that leant itself to this (and the inspiration behind all this): a modular project. Lets say we have a project structure that follows: a core set of functionality that must always be installed, and a set of  modules that can optionally be installed. For this we have no context for the core since it must always be present, and module 1 represented by the context `m1`, module 2 represented by the context `m2` and module 3 being represent by the context `m3`. We can also have a special context `mAll` to install all modules. An additional context can also be present for testing purposes, `qa`.  Depending on the project, I can easily see two main setups: 1) if no runtime contexts are provided, then 1) install the core and all modules or 2) install the core and no modules.

In scenario 1, we would want the empty set to match all the modules even though they have a filter provided. Each module would then have the following filters:

- Module 1: `m1 or @mAll`
- Module 2: `m2 or @mAll`
- Module 3: `m3 or @mAll`

But then we also have to consider the items that are installed for the `qa` context. Anything that should only be installed for the `qa` context would have filters like the following:

- Core: `@qa`
- Module 1: `(@m1 or @mAll) and @qa`
- Module 1: `(@m2 or @mAll) and @qa`
- Module 1: `(@m3 or @mAll) and @qa`

In scenario 2, we would NOT want the empty set to match all the modules. Each module would then have the following filters

-  Module 1: `@m1 or @mAll`
-  Module 2: `@m2 or @mAll`
-  Module 3: `@m3 or @mAll`

And again we also have to consider the items that are installed for the `qa` context. Anything that should only be installed for the `qa` context would have filters like the following:

- Core: `@qa`
- Module 1: `(@m1 or @mAll) and @qa`
- Module 1: `(@m2 or @mAll) and @qa`
- Module 1: `(@m3 or @mAll) and @qa`

So, as you can see, your concern is correct that the label could end up getting applied almost everywhere. But, if we created a global setting, we would be getting rid of the ability to implement situation 1.

I tend to give flexibility to developers where possible. While flexibility is great... it can also lead to a very complicated product. So I understand that my solution may not be the best answer, but I believe it is the most flexible for this problem space.

So thats it, sorry for the long response, just a lot of my thoughts on the topic. Happy to discuss other ideas on how this could be improved or implemented in a different way though!

nvoxland:
Sorry, @jasonlyle88 , I've been letting this sit in my pre-review pile too long. Moving it along in the review process to the "has the allotted time to make sure we work through it all" phase

MalloD12 (approved):
**Approved**.

Code changes look good to me. It also contains a good set of unit tests which also looks good to me. From my side I have review the code changes applied here, manually tested the changes and checked with product manager whether we are happy to move forward with this from a product point of view, and we all are ok with it. 

Build and all tests checks have been executed successfully.

### Things to be aware of:
- These changes doesn't affect previous context/label behaviour. It's just extending that behaviour with the addition of a new operator ('@').

### Things to worry about:
- None

jasonlyle88:
@MalloD12 @filipelautert Hey all! Looks like this has been reviewed, approved, and ready to go for a bit. Is there anything preventing this from getting merged in?

MalloD12:
> @MalloD12 @filipelautert Hey all! Looks like this has been reviewed, approved, and ready to go for a bit. Is there anything preventing this from getting merged in?

Hi @jasonlyle88, it's sitting in the queue so other members of the team can review it as well before it gets merged. My apologies for the time it's taking. 

Thanks,
Daniel.

jasonlyle88:
@filipelautert This was my first merge request, and on subsequent merge requests I've been told my maven runs are generating files that you generally don't commit in a PR like this. In this case, I believe only 8 of those files are actually files I changed, and the other 76 are files generated by the maven test lifecycle. SO hopefully those are a quick review... or we just remove the files from the PR.

filipelautert:
> @filipelautert This was my first merge request, and on subsequent merge requests I've been told my maven runs are generating files that you generally don't commit in a PR like this. In this case, I believe only 8 of those files are actually files I changed, and the other 76 are files generated by the maven test lifecycle. SO hopefully those are a quick review... or we just remove the files from the PR.

Yes, you are right. All the files under "verify/saved_state" can be removed. Good!

filipelautert (approved):

suryaaki2 (approved):

rberezen:
Hello @jasonlyle88 ! I attempted to execute Liquibase using the attached artifacts in this pull request and I discovered that the "@"  works correctly for the context, but it does not function as expected for labels. Unfortunately, the "@" symbol is treated as a regular character for labels. Could you kindly review this issue?
Screenshot from 2023-06-19 19-52-53

cc @filipelautert @MalloD12

jasonlyle88:
Hey @rberezen! Your example is showing expected behavior. Labels work with a control structure that is flipped around from contexts. This means that instead of the operators being present in the changelogs, label operators are made to be used on the command line with the labelFilter argument. You can test this with released version of liquibase and any of the existing operators (`AND`, `OR`, `!`, and `()`).

Using the `@` operator (or any other existing operator) in a changelog is only valid for `contextFilter` attributes.

cc @filipelautert @MalloD12

rberezen:
@jasonlyle88 I apologize for my mistake.

This PR introduces a new logical operator, "@" for labels and contexts. This operator enforces that an expression is required, meaning it will not match an empty item set. Consequently, if you have a changeset marked with contextFilter that starts with "@," and you run Liquibase without specifying the "--contexts" attribute, such changeset will not be executed.

And if you have a changeset labeled as "test" and other changesets without any labels, and you only want to run the changeset with the specific label, you need to execute Liquibase by specifying the "--labels=@test" attribute.

- New unit tests added.
- No additional testing required.

APPROVED

rberezen (approved):

jasonlyle88:
@rberezen No worries! I don't even know how many times I screwed that up during the coding and testing of this PR 😳 Thats the main reason I was able to recognize the mix up right away!

PR #3588, opened 2022-12-16
Performance improvements

filipelautert:
## Impact

## Description

1.  Fix strange break logic (LIQUIBASE_COMPLETE):
- After reading the entire table with all it columns, the current logic makes the column to always be reloaded if it is nullable. In cloud databases like BigQuery, it means an overhead (sometimes of around ~1s) per not nullable column when generating the database snapshot.
- It was introduced by a commit from 2018 with comment [Add additional logic to snapshot nullable column](https://github.com/liquibase/liquibase/commit/f98cd9f4edcc477a84ea48a578c06d17f1237b0f)  . 

2. ~~Do not add checksums to identifiers of changelogs, that is not required.~~ Forget about it , it is required to find duplicated changesets.

3. When running an update for a changelog with multiple included changeset files, disabling xml validation may grant a huge performance improvement - sometimes up to 50% speed increase. The more files you have more ou felll this improvement. This change does: 
* Creates a flag to toggle xsd validation (defaults to validate, keeping the current behaviour) . 
* Users can disable validation by using the new parameter "liquibase.validateXmlChangelogFiles"

## Things to be aware of
* Introduces a new liquibase parameter  "liquibase.validateXmlChangelogFiles"
* I'm not able to reproduce the issues reported by the users on the issues below, but from the performance dumps this is as close as I can get now. 

## Things to worry about
~~* Should we change default value for XML validations to false? Or keep it true and users would turn it off as required?~~


liquibase-core/src/test/groovy/liquibase/parser/core/xml/XMLChangeLogSAXParser_RealFile_Test.groovy
```diff
@@ -281,7 +282,16 @@ public class XMLChangeLogSAXParser_RealFile_Test extends Specification {
         e.message.startsWith("The file ${path} was not found")
     }
 
-    def "ChangeLogParseException thrown if changelog has invalid tags"() throws Exception {
+    def "ChangeLogParseException is thrown if the changeset has invalid tags"() throws Exception {
+        when:
+        def d = new XMLChangeLogSAXParser().parse("liquibase/parser/core/xml/malformedChangeLog.xml", new ChangeLogParameters(), new JUnitResourceAccessor())
```
XDelphiGrl, line 287: Is the variable `d` necessary? It looks like maybe it was used for debugging?

liquibase-core/src/test/groovy/liquibase/parser/core/xml/XMLChangeLogSAXParserTest.groovy
```diff
@@ -32,6 +32,21 @@ class XMLChangeLogSAXParserTest extends Specification {
...
+<!DOCTYPE databaseChangeLog [
+        <!ENTITY insecure SYSTEM "file:///invalid.txt">
+        ]>
+
+<databaseChangeLog xmlns="http://www.liquibase.org/xml/ns/dbchangelog"
+                   xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
+                   xsi:schemaLocation="http://www.liquibase.org/xml/ns/dbchangelog
+                      http://www.liquibase.org/xml/ns/dbchangelog/dbchangelog-4.19.xsd">
+
+    <iDontKnowWhatImDoing />
```
XDelphiGrl, line 48: Two things: 

1. The `<iDontKnowWhatImDoing/> made me laugh!
2. We should probably change the dbchangelog XSD version to be *-latest.xsd unless there is a reason to "pin" the version to 4.19.

liquibase-core/src/main/java/liquibase/snapshot/jvm/ColumnSnapshotGenerator.java
```diff
@@ -200,7 +200,7 @@ protected void addTo(DatabaseObject foundObject, DatabaseSnapshot snapshot) thro
                     Column column = readColumn(row, relation, database);
                     setAutoIncrementDetails(column, database, snapshot);
                     populateValidateNullableIfNeeded(column, metaDataNotNullConst, database);
-                    column.setAttribute(LIQUIBASE_COMPLETE, !column.isNullable());
+                    column.setAttribute(LIQUIBASE_COMPLETE, true);
```
XDelphiGrl, line 203: I agree that this was very strange logic. I went back to the original (internal to Liquibase) bug ticket that was linked to the PR where this `!column.isNullable()` was introduced. I tried to replicate the reported bug both with and without this change. Both code paths work correctly. My best guess is that at some point, the `validate` attribute was not in the XSD for `addNotNullConstraint` and that somehow made that other code useful. 

In any case, I am no longer worried about this change breaking a bug fixed a very, very long time ago.

liquibase-core/src/main/java/liquibase/changelog/ChangeLogIterator.java
```diff
@@ -179,7 +179,7 @@ protected String createKey(ChangeSet changeSet) {
         Labels labels = changeSet.getLabels();
         ContextExpression contexts = changeSet.getContextFilter();
 
-        return changeSet.toString(true)
+        return changeSet.toString(false)
```
XDelphiGrl, line 182: This class is missing JavaDoc annotations. Should I raise a separate GH Issue to address this?

XDelphiGrl (changes requested):
Please see my comments for more details ...

CHANGES REQUESTED

liquibase-core/src/test/groovy/liquibase/parser/core/xml/XMLChangeLogSAXParser_RealFile_Test.groovy
filipelautert: I forgot it there, thanks. Fixed!

liquibase-core/src/test/groovy/liquibase/parser/core/xml/XMLChangeLogSAXParserTest.groovy
filipelautert: Agreed, changed

liquibase-core/src/main/java/liquibase/GlobalConfiguration.java
```diff
@@ -215,6 +216,11 @@ public class GlobalConfiguration implements AutoloadedConfigurations {
         ALWAYS_DROP_INSTEAD_OF_REPLACE = builder.define("alwaysDropInsteadOfReplace", Boolean.class)
                 .setDescription("If true, drop and recreate a view instead of replacing it.")
                 .build();
+
+        VALIDATE_XML_CHANGELOG_FILES = builder.define("validateXmlChangelogFiles", Boolean.class)
```
suryaaki2, line 220: @filipelautert Do we need to mention in the description about the performance overhead with enabling XSD validation?

suryaaki2 (changes requested):

liquibase-core/src/main/java/liquibase/GlobalConfiguration.java
filipelautert: I"m addind some short comments about it.

liquibase-core/src/test/groovy/liquibase/parser/core/xml/XMLChangeLogSAXParserTest.groovy
XDelphiGrl: Github is still showing me `http://www.liquibase.org/xml/ns/dbchangelog/dbchangelog-4.19.xsd">` ... Is it lying, lol?

filipelautert: I guess it is 🤣

XDelphiGrl: So it was! I didn't see any of your changes, lol. We're all good now, though.

XDelphiGrl (commented):
This PR addresses performance degradation during XSD validation for very large XML changelogs. A new global property, `VALIDATE_XML_CHANGELOG_FILES` allows users to disable XSD validation. Please note that if `VALIDATE_XML_CHANGELOG_FILES` is disabled, the `LIQUIBASE_SECURE_PARSING` property is no longer considered as the XSD will not be used or downloaded. 
* The default for `VALIDATE_XML_CHANGELOG_FILES` is `true` to maintain backwards compatibility. 
* The valid values for `VALIDATE_XML_CHANGELOG_FILES` are `true` and `false`. 
* Disabling XSD validation improves performance.
* Disabling XSD validation circumvents logic to find errors in changeset definitions prior to running a Liquibase `update`. 
* Test harness executions are passing.
* There is one functional test failure that is under investigation; if that is a test framework issue and is resolved, I will approve this PR. If the test failure is valid (i.e., caused by this PR), this will go back to Development.

APPROVED (pending passing tests)

**UPDATE**
The functional tests are passing. A change to the ChangeLogIterator class was reverted from the original PR to fix the failure. Here is the commit that reverted the problematic line of code: [https://github.com/liquibase/liquibase/pull/3588/commits/9ef2aace1aaf9ae7afd276db15762d215da4261f](https://github.com/liquibase/liquibase/pull/3588/commits/9ef2aace1aaf9ae7afd276db15762d215da4261f)

XDelphiGrl (approved):

suryaaki2 (approved):

filipelautert:
Opened https://datical.atlassian.net/browse/PD-3156 to handle documentation update.

PR #3590, opened 2022-12-16
Remove redundant conditions tha covered by a subsequent condition.

arturobernalg:
## Impact

## Description

The Java 11 Language Specification expresses this concisely in section 15.20.2, "Type comparison operator instanceof". (Java 17 expresses this less concisely, after the introduction of instanceof patternmatching.)

"At run time, the result of the instanceof operator is true if the value of the RelationalExpression is not null and the reference could be cast to the ReferenceType without raising a ClassCastException. Otherwise the result is false."

So if the operand is null, the result is false.

https://docs.oracle.com/javase/specs/jls/se11/html/jls-15.html#jls-15.20.2

## Things to be aware of

## Things to worry about

- List unknowns you have


MalloD12 (approved):
**Approved**

This PR LGTM. I have reviewed updated conditions, also updated few others I found and make sure tests are all passing. Thank you, @arturobernalg!

### Things to be aware of:
- None

### Things to worry about:
- None

liquibase-core/src/main/java/liquibase/change/ChangeMetaData.java
```diff
@@ -137,6 +137,6 @@ public Set<String> getAppliesTo() {
     }
 
     public boolean appliesTo(DatabaseObject databaseObject) {
-        return (appliesTo != null) && appliesTo.contains(databaseObject.getObjectTypeName());
+        return appliesTo.contains(databaseObject.getObjectTypeName());
```
filipelautert, line 140: Here appliesTo can be null as the cosntructor sets it to null in some circunstances. Can you review it?

liquibase-core/src/main/java/liquibase/diff/compare/core/SchemaComparator.java
```diff
@@ -105,15 +105,15 @@ private String getSchemaAfterComparison(Database accordingTo, String schemaName1
 
     private String getFinalSchemaAfterComparison(Database accordingTo, String schemaName1, String comparisonSchema1, String comparisonSchema2, String finalSchema1) {
         if (CatalogAndSchema.CatalogAndSchemaCase.ORIGINAL_CASE.equals(accordingTo.getSchemaAndCatalogCase())){
-            if ((comparisonSchema1 != null) && comparisonSchema1.equals(schemaName1)) {
+            if (comparisonSchema1.equals(schemaName1)) {
```
filipelautert, line 108: comparisioSchema1 and 2 can be null, as they are provided by method getComparisonSchemaOrCatalog that can return null in some specific situations.

filipelautert (changes requested):

liquibase-core/src/main/java/liquibase/diff/compare/core/SchemaComparator.java
filipelautert: Equals logic is right, but if comparisonSchema1 is null then calling comparisonSchema1.equals will throw the nullpointer (null.equals) .

arturobernalg:
Hi  guys,
I'll check your comment this afternoon.
TY

liquibase-core/src/main/java/liquibase/change/ChangeMetaData.java
MalloD12: Added back this null check.

liquibase-core/src/main/java/liquibase/diff/compare/core/SchemaComparator.java
MalloD12: Applied back these checks too.

liquibase-core/src/main/java/liquibase/util/StringClauses.java
```diff
@@ -501,7 +501,7 @@ public boolean isEmpty() {
             return true;
         }
         for (Object clause : clauses.values()) {
-            if ((clause != null) && !"".equals(clause.toString().trim())) {
+            if (!"".equals(clause.toString().trim())) {
```
XDelphiGrl, line 504: This style of preventing null pointers is less readable than using `isEmpty`. Can you please see if this code can be cleaned up to use `isEmpty`?

XDelphiGrl (changes requested):
Please see [https://github.com/liquibase/liquibase/pull/3590/files#r1081576866](https://github.com/liquibase/liquibase/pull/3590/files#r1081576866) for requested changes.

CHANGES REQUESTED

arturobernalg:
Sorry I'm lost here. Do I have to check/change something here?

liquibase-core/src/main/java/liquibase/util/StringClauses.java
MalloD12: Hey @XDelphiGrl, yes we could make this change, but I would personally recommend doing it in a separate PR as it would be a separate readability improvement not related to changes introduced here. I just look for a comparison like this one and there are a couple of places where we need to make this update.

filipelautert:
> Sorry I'm lost here. Do I have to check/change something here?

No, it's fine. Daniel is helping from here, thanks!

liquibase-core/src/main/java/liquibase/util/StringClauses.java
XDelphiGrl: Makse sense to me. This code serves the same purpose, it is just a bit of an outdated approach.

MalloD12: @XDelphiGrl (cc: @filipelautert ) I totally agree it would make it a bit more readable what you are suggesting. I can quickly make a separate PR and make this update in a couple of files I saw yesterday we might need to update.

MalloD12: @XDelphiGrl created this [issue](https://github.com/liquibase/liquibase/issues/3718) to address empty string check.

filipelautert (approved):

PR #3594, opened 2022-12-17
Remove redundant modifiers.

arturobernalg:
## Impact

## Description

Remove redundant modifiers. The resulting code will be shorter, but the behaviour and meaning will remain the same.

## Things to be aware of

## Things to worry about

- List unknowns you have

### DIFF
--- a/liquibase-core/src/main/java/liquibase/diff/compare/DatabaseObjectComparator.java
+++ b/liquibase-core/src/main/java/liquibase/diff/compare/DatabaseObjectComparator.java
@@ -8,10 +8,10 @@ import java.util.Set;
 
 public interface DatabaseObjectComparator {
 
-    final int PRIORITY_NONE = -1;
-    final int PRIORITY_DEFAULT = 1;
-    final int PRIORITY_TYPE = 5;
-    final int PRIORITY_DATABASE = 10;
+    int PRIORITY_NONE = -1;
+    int PRIORITY_DEFAULT = 1;
+    int PRIORITY_TYPE = 5;
+    int PRIORITY_DATABASE = 10;
 
     int getPriority(Class<? extends DatabaseObject> objectType, Database database);
 
--- a/liquibase-core/src/main/java/liquibase/parser/NamespaceDetails.java
+++ b/liquibase-core/src/main/java/liquibase/parser/NamespaceDetails.java
@@ -5,7 +5,7 @@ import liquibase.servicelocator.PrioritizedService;
 
 public interface NamespaceDetails extends PrioritizedService{
 
-    public static int PRIORITY_EXTENSION = PRIORITY_DATABASE;
+    int PRIORITY_EXTENSION = PRIORITY_DATABASE;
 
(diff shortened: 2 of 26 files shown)


MalloD12 (approved):
*Approved*.


Another clean-up PR which also LGTM. Thank you @arturobernalg for making look better this code.

### Things to be aware of:
- None

### Things to worry about:
- None

liquibase-core/src/main/java/liquibase/parser/NamespaceDetails.java
```diff
@@ -5,7 +5,7 @@
 
 public interface NamespaceDetails extends PrioritizedService{
 
-    public static int PRIORITY_EXTENSION = PRIORITY_DATABASE;
+    int PRIORITY_EXTENSION = PRIORITY_DATABASE;
```
XDelphiGrl, line 8: Is there any functional reason it makes sense to allow multiple copies of `PRIORITY_EXTENSION` in the JVM?

arturobernalg: no. its just redundant.
https://docs.oracle.com/javase/specs/jls/se8/html/jls-9.html

liquibase-core/src/main/java/liquibase/diff/compare/DatabaseObjectComparator.java
```diff
@@ -8,10 +8,10 @@
 
 public interface DatabaseObjectComparator {
 
-    final int PRIORITY_NONE = -1;
-    final int PRIORITY_DEFAULT = 1;
-    final int PRIORITY_TYPE = 5;
-    final int PRIORITY_DATABASE = 10;
+    int PRIORITY_NONE = -1;
```
XDelphiGrl, line 11: Is there any functional reason to allow callers to modify these values at runtime?

arturobernalg: idem above

liquibase-core/src/main/java/liquibase/parser/NamespaceDetails.java
XDelphiGrl: Hi, @arturobernalg!  Thanks for doing so much code cleanup for the project - it is awesome! Is this the bit of the documentation that applies? 
```
Every field declaration in the body of an interface is implicitly public, static, and final.
```

I'm not a Java developer so thank you for being kind about answering my questions. :)

XDelphiGrl (approved):
This PR continues with @arturobernalg's cleanup and enhanced readability of the Liquibase code! Redundant modifiers are removed from interface members where modifiers are already implicitly defined by Java itself.
* No impact to Liquibase functionality.

APPROVED

liquibase-core/src/main/java/liquibase/parser/NamespaceDetails.java
arturobernalg: Hi, exactly. All variables declare in an interface are by default public static final.

Thank you.

filipelautert (approved):

PR #3620, opened 2022-12-26
Issue 3619 - Allow control of recursion for includeAll via minDepth and maxDepth attributes

jasonlyle88:
## Impact

## Description

Adds two new attributes to the changelog `includeAll` attribute: `minDepth` (inclusive, default `1`) and `maxDepth` (inclusive, default `Integer.MAX_VALUE`). These defaults allow `includeAll` to behave as it did before these provided changes.

For `minDepth`, `0` and `1` are treated the same way and denote the directory specified in the `includeAll`'s `path` attribute. Having a `minDepth` of 2 would skip any files specified in the `includeAll`'s `path` attribute and only start searching for files to include in immediate subdirectories of the `includeAll`'s `path` attribute.

For `maxDepth`, `1` is the minimum value. A value less than `minDepth` will result in no files being included. The `maxDepth` attribute is computed off of the `includeAll`'s `path` attribute, not based on the `minDepth` attribute.

Examples:

`minDepth`=`1`,`maxDepth`=`1`: Include all the files in the specified `path` attribute directory and do not look in any subdirectories (identical to recursive: false).

`minDepth`=`1`,`maxDepth`=`2`: Include all the files in the specified `path` attribute directory and look in immediate subdirectories and those subdirectories' immediate subdirectories.

... So on and so forth for maxDepth

`minDepth`=`2`,`maxDepth`=`null`: Do not include any files in the specified `path` attribute directory and look in all subdirectories for files.

## Things to be aware of

Not only did this require the change of the `includeAll` attribute, but also the `ResourceAccessor` interface.

As this change occurred on the base interface, all `ResourceAccessors` changed to support this model.

## Things to worry about

1. I've never worked with the Spring framework before. Because of this, I as unsure how/if I should handle the depth bounded search function for the `SpringResourceAccessor` class. For now, I created the method and just immediately throw an error saying it is not implemented. Not sure if this is acceptable or if this should be fully implemented.
2. I wasn't sure how/if I should implemented the depth bounded search function for the `MockResourceAccessor` class. For now, I created the method and just immediately throw an error saying it is not implemented. Not sure if this is acceptable or if this should be fully implemented.
3. I am unsure how to properly write tests for this when it is based on a filesystem configuration.

### DIFF
--- a/liquibase-core/src/main/java/liquibase/changelog/DatabaseChangeLog.java
+++ b/liquibase-core/src/main/java/liquibase/changelog/DatabaseChangeLog.java
@@ -494,6 +494,8 @@ public class DatabaseChangeLog implements Comparable<DatabaseChangeLog>, Conditi
                         includeContextFilter,
                         labels,
                         ignore,
+                        node.getChildValue(null, "minDepth", Integer.valueOf(1)),
+                        node.getChildValue(null, "maxDepth", Integer.MAX_VALUE),
                         (ModifyChangeSets)nodeScratch.get("modifyChangeSets"));
                 break;
             }
@@ -604,7 +606,9 @@ public class DatabaseChangeLog implements Comparable<DatabaseChangeLog>, Conditi
                 resourceAccessor,
                 includeContextFilter,
                 labels,
-                ignore);
+                ignore,
+                1,
+                Integer.MAX_VALUE);
     }
     public void includeAll(String pathName,
                            boolean isRelativeToChangelogFile,
@@ -614,10 +618,12 @@ public class DatabaseChangeLog implements Comparable<DatabaseChangeLog>, Conditi
                            ResourceAccessor resourceAccessor,
                            ContextExpression includeContextFilter,
                            Labels labels,
-                           boolean ignore)
+                           boolean ignore,
+                           Integer minDepth,
+                           Integer maxDepth)
             throws SetupException {
         includeAll(pathName, isRelativeToChangelogFile, resourceFilter, errorIfMissingOrEmpty, resourceComparator,
-                   resourceAccessor, includeContextFilter, labels, ignore, new ModifyChangeSets(null));
+                   resourceAccessor, includeContextFilter, labels, ignore, minDepth, maxDepth, new ModifyChangeSets(null));
     }
 
     public void includeAll(String pathName,
@@ -629,6 +635,8 @@ public class DatabaseChangeLog implements Comparable<DatabaseChangeLog>, Conditi
                            ContextExpression includeContextFilter,
                            Labels labels,
                            boolean ignore,
+                           Integer minDepth,
+                           Integer maxDepth,
                            ModifyChangeSets modifyChangeSets)
             throws SetupException {
         try {
@@ -659,8 +667,7 @@ public class DatabaseChangeLog implements Comparable<DatabaseChangeLog>, Conditi
                 LOG.fine("includeAll for " + pathName);
                 LOG.fine("Using file opener for includeAll: " + resourceAccessor.toString());
 
-                //TODO: JML update to have minDepth and maxDepth attributes
-                unsortedResources = resourceAccessor.search(path, 1, Integer.MAX_VALUE);
+                unsortedResources = resourceAccessor.search(path, minDepth, maxDepth);
             } catch (IOException e) {
                 if (errorIfMissingOrEmpty) {
                     throw e;
--- a/liquibase-core/src/main/java/liquibase/changelog/DatabaseChangeLog.java
+++ b/liquibase-core/src/main/java/liquibase/changelog/DatabaseChangeLog.java
@@ -659,7 +659,8 @@ public class DatabaseChangeLog implements Comparable<DatabaseChangeLog>, Conditi
                 LOG.fine("includeAll for " + pathName);
                 LOG.fine("Using file opener for includeAll: " + resourceAccessor.toString());
 
-                unsortedResources = resourceAccessor.search(path, true);
+                //TODO: JML update to have minDepth and maxDepth attributes
+                unsortedResources = resourceAccessor.search(path, 1, Integer.MAX_VALUE);
             } catch (IOException e) {
                 if (errorIfMissingOrEmpty) {
                     throw e;
--- a/liquibase-core/src/main/java/liquibase/resource/ResourceAccessor.java
+++ b/liquibase-core/src/main/java/liquibase/resource/ResourceAccessor.java
@@ -129,6 +129,30 @@ public interface ResourceAccessor extends AutoCloseable {
      * </ul>
      *
      * @param path      The path to lookup resources in.
+     * @param minDepth  Minimum folder depth from path to start searching (default: 0)
+     * @param maxDepth  Maximum folder depth from path to stop searching (default: null)
+     * @return empty set if nothing was found
+     * @throws IOException if there is an error searching the system.
+     */
+    List<Resource> search(String path, Integer minDepth, Integer maxDepth) throws IOException;
+
+    /**
+     * Returns the path to all resources contained in the given path.
+     * Multiple resources may be returned with the same path, but only if they are actually unique files.
+     * Order is important to pay attention to, they should be returned in a user-expected manner based on this resource accessor.
+     * <br><br>
+     * Should return an empty list if:
+     * <ul>
+     *     <li>Path does not exist</li>
+     * </ul>
+     * Should throw an exception if:
+     * <ul>
+     *     <li>Path is null</li>
+     *     <li>Path is not a "directory"</li>
+     *     <li>Path exists but cannot be read from</li>
+     * </ul>
+     *
+     * @param path      The path to lookup resources in.
      * @param recursive Set to true and will return paths to contents in subdirectories as well.
      * @return empty set if nothing was found
      * @throws IOException if there is an error searching the system.
--- a/liquibase-core/src/main/java/liquibase/changelog/ChangeLogIncludeAll.java
+++ b/liquibase-core/src/main/java/liquibase/changelog/ChangeLogIncludeAll.java
@@ -14,9 +14,13 @@ public class ChangeLogIncludeAll extends AbstractLiquibaseSerializable implement
     private String resourceFilter;
     private ContextExpression context;
 
+    private Integer minDepth;
+
+    private Integer maxDepth;
+
     @Override
     public Set<String> getSerializableFields() {
-        return new LinkedHashSet<>(Arrays.asList("path", "errorIfMissingOrEmpty", "relativeToChangelogFile", "resourceFilter", "context"));
+        return new LinkedHashSet<>(Arrays.asList("path", "errorIfMissingOrEmpty", "relativeToChangelogFile", "resourceFilter", "context", "minDepth", "maxDepth"));
     }
 
     @Override
@@ -68,4 +72,12 @@ public class ChangeLogIncludeAll extends AbstractLiquibaseSerializable implement
     public void setContext(ContextExpression context) {
         this.context = context;
     }
+
+    public Integer getMinDepth() { return minDepth; }
+
+    public void setMinDepth (Integer minDepth) { this.minDepth = minDepth; }
+
+    public Integer getMaxDepth() { return maxDepth; }
+
+    public void setMaxDepth (Integer maxDepth) { this.maxDepth = maxDepth; }
 }
--- a/liquibase-core/src/main/java/liquibase/integration/spring/SpringResourceAccessor.java
+++ b/liquibase-core/src/main/java/liquibase/integration/spring/SpringResourceAccessor.java
@@ -48,6 +48,11 @@ public class SpringResourceAccessor extends AbstractResourceAccessor {
     }
 
     @Override
+    public List<liquibase.resource.Resource> search(String searchPath, Integer minDepth, Integer maxDepth) throws IOException {
+        throw new UnexpectedLiquibaseException("Method not implemented");
+    }
+
+    @Override
     public List<liquibase.resource.Resource> search(String searchPath, boolean recursive) throws IOException {
         if (recursive) {
             searchPath += "/**";
--- a/liquibase-core/src/main/java/liquibase/resource/AbstractPathResourceAccessor.java
+++ b/liquibase-core/src/main/java/liquibase/resource/AbstractPathResourceAccessor.java
@@ -69,7 +69,7 @@ public abstract class AbstractPathResourceAccessor extends AbstractResourceAcces
     }
 
     @Override
-    public List<Resource> search(String startPath, boolean recursive) throws IOException {
+    public List<Resource> search(String startPath, Integer minDepth, Integer maxDepth) throws IOException {
         if (startPath == null) {
             throw new IllegalArgumentException("Path must not be null");
         }
@@ -92,10 +92,20 @@ public abstract class AbstractPathResourceAccessor extends AbstractResourceAcces
             throw new IOException("'" + startPath + "' is a file, not a directory");
         }
 
+        if (minDepth == null) {
+            minDepth = 1;
+        }
+
+        final int finalMinDepth = minDepth;
+
+        if (maxDepth == null) {
+            maxDepth = Integer.MAX_VALUE;
+        }
+
         SimpleFileVisitor<Path> fileVisitor = new SimpleFileVisitor<Path>() {
             @Override
             public FileVisitResult visitFile(Path file, BasicFileAttributes attrs) throws IOException {
-                if (attrs.isRegularFile()) {
+                if (attrs.isRegularFile() && residesInDepthBoundaries(file)) {
                     addToReturnList(file);
                 }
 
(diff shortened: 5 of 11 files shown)


liquibase-core/src/main/java/liquibase/resource/ResourceAccessor.java
```diff
@@ -112,6 +112,30 @@ default SortedSet<String> list(String relativeTo, String path, boolean recursive
...
+     *     <li>Path exists but cannot be read from</li>
+     * </ul>
+     *
+     * @param path      The path to lookup resources in.
+     * @param minDepth  Minimum folder depth from path to start searching (default: 0)
+     * @param maxDepth  Maximum folder depth from path to stop searching (default: null)
+     * @return empty set if nothing was found
+     * @throws IOException if there is an error searching the system.
+     */
+    List<Resource> search(String path, Integer minDepth, Integer maxDepth) throws IOException;
```
nvoxland, line 137: Having a new search() version that is smarter allows us to add nice functionality like you have here.

However, just adding this new search function to the interface will break any extensions which simply implement this directly. I also see an unending set of other options people will want to add to the search logic over time which will lead to a large number of methods that ResourceAccessor implementations will have to think about and callers have to decide between.

I'm thinking it would be better to have ResourceAccessor.SearchOptions inner class with fields of minDepth, maxDepth, and recursive and change this new method to be `search(String path, SearchOptions searchOptions)` instead. That way future new like "file name pattern" or whatever can be included without having to make new method permutations.

Then, I'm wondering if we could default that new search method to call the existing `search(path, boolean)` call and do in-memory filtering of the results? That would allow ResourceAccessor implementors to generally just have the one method to implement and keep getting the additional functionality as it's added? But still leaving them to implement the new `search(String, SearchOptions)` if they feel they can do a better job?

nvoxland (changes requested):

liquibase-core/src/main/java/liquibase/resource/ResourceAccessor.java
jasonlyle88: I really like that idea, the more extensible the better! Especially in this situation where things are community driven and there is this extension capability. I do like the default implementation in the interface to allow it to move forward without breaking current implementations and allow future implementors to override only if they want.

I can definitely take a stab at that... however I am not a Java programmer. I picked this up as we are using Liquibase for our project and this is the first time I've touched Java in ~15 years (and I've only been using Liquibase for ~ 1 month). As such, please feel free to let me know where improvements can be made as my Java knowledge is very outdated and I'm still learning Liquibase!

jasonlyle88:
@nvoxland I believe I have implemented the `SearchOptions` class how you were describing it. I also implemented the method in the `ResourceAccessor` interface as a default method so that any existing/future extensions don't have any issues after this is merged. 

Here are some updates to my original post for this merge request:

# Things to be aware of

I believe this default `search(String, SearchOptions)` method in the interface strategy takes care of the items I mentioned above in the "Things to be aware of" section.

# Things to worry about

1. `SpringResourceAccessor` still does not have this implemented. I don't think thats an issue... but wanted to make sure I called it out at least.
2. `MockResourceAccessor` has the method implemented, but doesn't take any of the SearchOption attributes into account. It previously did not look at the `recursive` value either, so I figure this is the intended functionality, but again just want to call it out to be thorough.
3. Not knowing how to implement tests based on filesystem hierarchy is still a concern.
4. (NEW) I opted not to create a `recursive` attribute of this class because that is really just a specific expression of `minDepth` and `maxDepth`. I did create `getRecursive` and `setRecursive` methods, but those just set `minDepth` and `MaxDepth` to appropriate values. Let me know if this is an issue, I just saw it as a better way of doing this since they really tell you the same information and could just cause problems trying to keep `recursive` and `minDepth`/`maxDepth` in sync.

nvoxland:
Good point that `recursive` is just a limited version of min/max depth. No need to have that too. 

The changes is looking generally good, so sending it along for the deeper review. I like that for our ResourceAccessor implementations you have fixed them up to implement the new SearchOptions search. But I wonder if it's possible to have the default ResourceAccessor SearchOptions method you made not just call the old recursive search and ignore the min/max settings, but still respect them? Maybe take the list of Resources returned from the recursive search and loop through them and pull out resources too high or low in the tree? Like what you did in `residesInDepthBoundaries` but with the more limited API of Resource. You might only have splitting on "/" or "\" but that would still better respect those settings for ResourceAccessors that only implement the recursive search version.

jasonlyle88:
Awesome, I'm hoping the review goes well. I was able to get a basic depth bounding into the default `search(String, SearchOptions)` implementation based off `Resource.getPath()` and counting the `/` characters since `Resource.getPath()` says it should always be relative and use `/` as the separator. Let me know if you see any problems with this implementation.

MalloD12 (approved):
**Approved**.

Changes looks good to me. As part of the review we have performed the following actions:

- Fixed failing unit tests
- Remove attribute default values in XSD
- Remove attribute types in XSD
- Added validation errors when setting up minDepth and maxDepth
- Change types from integer to primitive int type
- Update minDepth initialisation to be set with 0, instead of 1.
- Update getRecursive and setRecursive methods.
- Add validation error to return exception if maxDepth is less than minDepth
- ResourceAccessor search logic updated to make proper use of `searchRecursive` flag.
- MockResourceAccessor updated to only make use of recursive search method. Also, search method has been updated to make a proper use of **recursive** boolean flag.
- Manual testing performed with below combinations:
> 0-0
> 0-1
> 1-0
> 1-1
> 0-2
> 1-2
> 0-3
> 3-123123

- Unit test added to test `includeAll` with new min/max boundaries.
- Integration tests updated to also make use of these new `includeAll` properties.

### Things to be aware of:
- None: this is only having an impact on includeAll tag, but this is still supporting previous includeAll definition (without needing to specify recursive boundaries).

### Things to worry about:
- None

jasonlyle88:
@MalloD12 No comments here! I've been following along and everything has made sense. Thank you very much for your work on getting this across the finish line!

Only question I have is if this will make it into the next release. I see a merge freeze check in the git checks and wasn't sure.

filipelautert:
Hello @jasonlyle88  - we are going to release 4.19.1 this week but this PR won't be able to make it. As we try to release monthly the next release that it may make into is around beggining of April.

liquibase-core/src/main/java/liquibase/resource/ResourceAccessor.java
```diff
@@ -257,4 +308,51 @@ public OutputStream openOutputStream(OpenOptions openOptions) throws IOException
...
         }
     }
+
+    class SearchOptions {
+        private int minDepth;
+        private int maxDepth;
+
+        public SearchOptions() {
+            minDepth = 0;
+            maxDepth = 1;
```
filipelautert, line 318: Should not the default value be  Integer.MAX_VALUE ? This way it will behave the same as before if new people just do `new SearchOptions() `

filipelautert (commented):
thanks for the code! I just added one comment about the default values.

liquibase-core/src/main/java/liquibase/resource/ResourceAccessor.java
MalloD12: In my opinion, it wouldn't change that much when SearchOptions is initialized it will be followed up with min/maxDepth seters. It we do initialize it with `Integer.MAX_VALUE` it will still define whether it's a recursive search depending on the values set. To keep recursive compatibility as it was before this change, there is a `search()` which expect a `recursive` boolean flag instead of a SearchOptions (as well as an IncludeAll method which does not expect min/max values which will end up performing a recursive search `min=0` - `max=Integer.MAX_VALUE`)

filipelautert (approved):

suryaaki2 (approved):

jasonlyle88:
@nvoxland not sure if you have to approve the changes for this to move forward once the merge freeze is done, but figured I'd ask now and get the ball rolling in case.

I think everything you asked for is covered here though!

liquibase-core/src/main/java/liquibase/changelog/DatabaseChangeLog.java
```diff
@@ -995,4 +1005,27 @@ public enum OnUnknownFileFormat {
...
+     * @param minDepth
+     * @param maxDepth
+     * @return ResourceAccessor.SearchOptions
+     * @throws SetupException in case maxDepth is less than minDepth
+     */
+    private ResourceAccessor.SearchOptions initializeAndSetMinAndMaxDepth(int minDepth, int maxDepth) throws SetupException {
+        ResourceAccessor.SearchOptions searchOptions = new ResourceAccessor.SearchOptions();
+        try {
+            if (maxDepth < minDepth) {
+                throw new IllegalArgumentException("maxDepth argument must be greater than minDepth");
```
XDelphiGrl, line 1019: Nice error handling!

XDelphiGrl:
This ticket needs to be tested on Windows. Internal Liquibase QA will take this action item.

XDelphiGrl:
**Test Methodology: includeAll**
To validate various combinations of `includeAll` `minDepth` and `maxDepth` properties, create a set of nested directories. Each directory will have one XML changelog with one changeset, except for one nested directory which is empty. The top-level directory contains the "parent" YAML chaneglogs; each of these changelogs has one `includeAll` changetype with varying combinations of `minDepth` and `maxDepth`. Parent changelogs are named according to the settings for `minDepth` and `maxDepth`. 

This set of tests is executed on Windows 11.

---

**Directory Structure**

~/ISSUE3619

~/ISSUE6319/depth1
issue3619-1-changelog.xml

~/ISSUE6319/depth1/depth2
issue3619-2-changelog.xml

~/ISSUE3619/depth1/depth2/depth3
issue3619-3-changelog.xml

~/ISSUE3619/depth1/depth2/depth3/depth4-no-changelog

~/ISSUE3619/depth1/depth2/depth3/depth4-no-changelog/depth5
issue3619-5-changelog.xml

---
**Parent Changelogs**
* ISSUE3619/issue3619-none-changelog.yaml (no min or max configured)
* ISSUE3619/issue3619-min0max1-changelog.yaml
* ISSUE3619/issue3169-min1-changelog.yaml
* ISSUE3619/issue3619-min1max2-changelog.yaml
* ISSUE3619/issue3619-min1max2-depth3-changelog.yaml (relative nested path using /depth1/depth2/depth3)
* ISSUE3619/depth1/depth2/depth3/depth4-no-changelog/depth5/issue3619-min1-max2-relativedepth3-changelog.yaml (relative path using `..` to ISSUE3619/depth1/depth2/depth3)
* ISSUE3619/issue3619-min1max0-changelog.yaml

---
**Tests: includeAll**
Changelog for Cases 1, 2, 3, 4, 5 and 7 are executed from directory `ISSUE3619`.
Changelogs for Case 6 is executed from directory `ISSUE3619\depth1\depth2\depth3\depth4-no-changelog\depth5`
* Relative path is `/../../../depth3` 

| Case # | Parent  | Min | Max | Deployed Directories | Expected Count Changesets Deployed | P/F |
| ------------- | ------------- | ------------- | ------------- | ------------- | ------------- | ------------- |
| 1 | issue3619-none | NOT SET  |  NOT SET | 1,2,3,4,5 | 4 | PASS |
| 2 | issue3619-min0max1 | 0  |  1 | 1 | 1 |  PASS |
| 3 |issue3619-min1 | 1  |  1 | 1,2,3,4,5 | 5 | PASS |
| 4 |issue3619-min1max2 | 1  | 2 | 1,2 | 2 | PASS |
| 5 |issue3619-min1max2-depth3 | 1 | 2  | 3 | 1 | PASS |
| 6 | issue3619-min1max2-relativedepth | 1 | 2 | 3 |  1 | PASS |
| 7 | issue3619-min1max0 | N/A  |  0 | ERROR | 0 | PASS |

---
**Tests: includeAll wrapped in modifyChangesets**

This test uses a parent changelogs with an `includeAll` wrapped in `modifyChangesets`.  Formatted SQL changelogs, each with one changeset are located in `ISSUE3619/depth1`, `ISSUE3619/depth1/depth2` and `ISSUE3619/depth1/depth2/depth3`. This test validates the following conditions:
* JSON `includeAll` supports `minDepth` and `maxDepth`
* XML `includeAll` supports `minDepth` and `maxDepth`
* An `includeAll` wrapped with `modifyChangesets` correctly uses the PSQL native exexutor (PRO feature).

**Example of includeAll Wrapped in modifyChangesets (JSON)**
```
   "modifyChangeSets": 
       {
           "runWith": "psql",
           "includeAll": {"path": "depth1", "relativeToChangelogFile": "true", "minDepth": "0", "maxDepth": "3"}
       }
```

Changelogs for Case 8 and Case 9 are executed from directory `ISSUE3619`.

| Case # | Parent  | Min | Max | Deployed Directories | Expected Count Changesets Deployed | P/F |
| ------------- | ------------- | ------------- | ------------- | ------------- | ------------- | ------------- |
| 8 | issue3619-modifychangeset.json | 0 |  1 | 1,2,3 | 3 | PASS |
| 9 | issue3619-modifychangeset.xml | 0 |  1 | 1,2,3 | 3 | PASS |

liquibase-core/src/test/groovy/liquibase/changelog/DatabaseChangeLogTest.groovy
```diff
@@ -419,7 +443,7 @@ create view sql_view as select * from sql_table;'''
                 "com/example/not/fileX.sql"     : "file X",
         ])
         def changeLogFile = new DatabaseChangeLog("com/example/root.xml")
-        changeLogFile.includeAll("com/example/missing", false, null, true, changeLogFile.getStandardChangeLogComparator(), resourceAccessor, new ContextExpression(), new Labels(), false)
+        changeLogFile.includeAll("com/example/missing", false, null, true, changeLogFile.getStandardChangeLogComparator(), resourceAccessor, new ContextExpression(), new LabelExpression(), false)
```
XDelphiGrl, line 446: Shouldn't `LabelExpression` be `Labels`? As I understand it, the changelog cannot have a label expression (or filter). In looking at  [the includeAll case statement near this line in DatabaseChangelog.java](https://github.com/liquibase/liquibase/blob/63705dc90d775723d48859e33bdcbbb67c237572/liquibase-core/src/main/java/liquibase/changelog/DatabaseChangeLog.java#L519), it looks like Labels is the correct object type.  However ... this is working so maybe I just need some general education on when to use `Labels` versus `LabelExpression`.

liquibase-core/src/test/groovy/liquibase/changelog/DatabaseChangeLogTest.groovy
```diff
@@ -590,4 +614,32 @@ http://www.liquibase.org/xml/ns/dbchangelog http://www.liquibase.org/xml/ns/dbch
...
+
+        def rootChangeLog = new DatabaseChangeLog(rootChangeLogPath)
+        rootChangeLog.load(new ParsedNode(null, "databaseChangeLog")
+                .addChildren([includeAll: [path: includedAllChangeLogPath, minDepth:minDepth, maxDepth:maxDepth, errorIfMissingOrEmpty:false]]), resourceAccessor)
+
+        then:
+        rootChangeLog.getChangeSets().size() == expectedIncludeAllChangesetsToDeploy
+
+        where:
+        minDepth | maxDepth | expectedIncludeAllChangesetsToDeploy
```
XDelphiGrl, line 636: This is a very nice set of test cases!

XDelphiGrl (approved):
This new feature allows you to control the depth of recursion applied by `includeAll`. To do so, set the `minDepth` and `maxDepth` properties on an `includeAll` changeset. Not setting either `minDepth` or `maxDepth` is the equivalent of pre-existing behavior and all directories are recursed to find changelogs to execute.
* This feature does not break backwards compatibility.
* This feature DOES have an XSD change and can only be released with a major version.
* This feature works with `includeAll` changesets wrapped in a `modifyChangesets` changetype; this is used for the PRO-only native executors, like SQLPlus, SQLCmd and PSQL.
* The Firebird test harness failures are due to an environmental issue with Titan and are not related to this PR. 
* The functional test failures are related to an issue fixed on the functional test master branch; again, these are not related to this PR and can be disregarded.
* Manual testing executed on Windows and Mac to validate there are no pathing issues related to this change.

APPROVED

liquibase-core/src/test/groovy/liquibase/changelog/DatabaseChangeLogTest.groovy
MalloD12: These tests are using the deprecated version of `includeAll` :

```
public void includeAll(String pathName,
                           boolean isRelativeToChangelogFile,
                           IncludeAllFilter resourceFilter,
                           boolean errorIfMissingOrEmpty,
                           Comparator<String> resourceComparator,
                           ResourceAccessor resourceAccessor,
                           ContextExpression includeContextFilter,
                           LabelExpression labelExpression,
                           boolean ignore)
```
Which we still use for backward compatibility. These tests were failing because passing labels (instead of label expression) would also require including these new parameters (minDepth/maxDepth), so at that moment I thought might be simpler to keep it that way, but I can update these tests and make use of the a non-deprecated includeAll method.

jasonlyle88:
Thanks everyone for your work here and helping me get this included! Very much appreciated!

PR #3629, opened 2022-12-29
DAT-12576 update release workflow to attach artifact by run_id

ap-liquibase:
## Impact

## Description

## Things to be aware of

## Things to worry about

- List unknowns you have

### DIFF
--- a/.github/workflows/create-release.yml
+++ b/.github/workflows/create-release.yml
@@ -8,7 +8,9 @@ on:
       branch:
         description: 'Branch to release (Defaults to master)'
         required: false
-
+      runId:
+        description: 'RunId of liquibase/liquibase artifacts to attach'
+        required: true
 jobs:
   setup:
     name: Setup
@@ -16,6 +18,7 @@ jobs:
     outputs:
       version: ${{ steps.collect-data.outputs.version }}
       branch: ${{ steps.collect-data.outputs.branch }}
+      runId: ${{ steps.collect-data.outputs.runId }}
     steps:
       - name: Collect Data
         id: collect-data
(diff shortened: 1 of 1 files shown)


jnewton03 (commented):
@ap-liquibase do we need to set a default value for `runId` ?

ap-liquibase:
setting a default value would be desirable. I will check with @XDelphiGrl to see what might be a good value for this.

XDelphiGrl:
> setting a default value would be desirable. I will check with @XDelphiGrl to see what might be a good value for this.

@jnewton03 and @ap-liquibase, hi - 

I do not think there is a way to programmatically chose a sensible default `runId`. The `attach-artifacts` job is manually executed and is not unreasonable to expect the person running the workflow come prepared with the `runId` of the artifacts they want to attach. 

Andre and I discussed the possibility of logic to have an either/or for `branch` and `runId`, as `branch` can always default to `master`. I feel that complexity is unnecessary at this time. Given that using the default branch sometimes chose the incorrect artifacts was the main reason QA submitted this ask, I think just specifying a `runId` is sufficient.

Andre is going to check to see if we can safely do a dry run of the functionality in this PR. I've asked him to find out where the repo tag happens, as people in the Community do watch for that; if we put a release tag there now, and don't release for a couple of weeks, people will notice. Andre suggested using a non-standard tag so people would know we are testing and then we could just delete the (test) draft release when we've tested the new workflow with `runId`.

jnewton03:
@ap-liquibase @XDelphiGrl that sounds like a good plan

jnewton03 (approved):

PR #3636, opened 2023-01-02
Reworks DirectoryPathHandler to allow colons in file names. (DAT-12363)

abrackx:
## Impact

## Description
Moves ":" into regex validating DirectoryPathHandler's priority. With this change, you are now able to generate changelogs with filenames such as `example:something-else.yml`. 

## Things to be aware of
Changes priorty check of DirectoryPathHandler.

## Things to worry about
Explicitly disallow files starting with `http://`, `https://` and `proto:` rather than just falling back on excluding `:`. Unsure if there is additional impact of this change. 

## Additional Context
N/A

### DIFF
--- a/liquibase-core/src/main/java/liquibase/resource/DirectoryPathHandler.java
+++ b/liquibase-core/src/main/java/liquibase/resource/DirectoryPathHandler.java
@@ -20,7 +20,7 @@ public class DirectoryPathHandler extends AbstractPathHandler {
             return PRIORITY_DEFAULT;
         }
 
-        if (root.startsWith("file:") || root.matches("^[A-Za-z]:.*")) {
+        if (root.startsWith("file:") || root.matches("^[A-Za-z,:].*")) {
             return PRIORITY_DEFAULT;
         } else {
             return PRIORITY_NOT_APPLICABLE;
--- a/liquibase-core/src/main/java/liquibase/resource/DirectoryPathHandler.java
+++ b/liquibase-core/src/main/java/liquibase/resource/DirectoryPathHandler.java
@@ -20,6 +20,10 @@ public class DirectoryPathHandler extends AbstractPathHandler {
             return PRIORITY_DEFAULT;
         }
 
+        if (root.matches("^https?://.*")) {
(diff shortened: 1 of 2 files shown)


suryaaki2 (approved):

liquibase-core/src/main/java/liquibase/resource/DirectoryPathHandler.java
```diff
@@ -20,7 +20,15 @@ public int getPriority(String root) {
...
-        if (root.startsWith("file:") || root.matches("^[A-Za-z]:.*")) {
+        if (root.matches("^https?://.*")) { // Do not support http files
+            return PRIORITY_NOT_APPLICABLE;
+        }
+
+        if (root.matches("^proto:.*")) { // Do not support protobuf files
+            return PRIORITY_NOT_APPLICABLE;
+        }
+
+        if (root.startsWith("file:") || root.matches("^[A-Za-z,:].*")) {
```
nvoxland, line 31: By making DirectoryPathHandler support arbitrary names with colons as "default priority", it will be conflicting with any more specific handlers that also use default priority.

I'm not sure if there are any currently we write. ZipPathHandler currently uses PRIORITY_SPECIALIZED. I can't see the s3 code anymore to know what that looks like. But I could also imagine external people choosing PRIORITY_DEFAULT for theirs and then getting random times their path handler is not used because of this.

I get that this is required for people to have colons in directories, but is that something we want to support? 

Besides the potential "bugs" in extensions it introduces, it will also be confusing for people who put a path root with a protocol like `ftp://...` or `s3:` and get a directory created instead of a "ftp not supported" (or whatever the protocol is) error like they may expect.

Maybe it would help to be not explicitly checking for proto/http/etc. protocols but instead looking for a `root.matches("^\\w+://")` like generic protocol-looking string and returning NOT_APPLICABLE for that?

StevenMassaro: > I get that this is required for people to have colons in directories, but is that something we want to support?

Seems like a pretty common use case for Windows users, no?

PR #3640, opened 2023-01-03
Remove potentially sensitive information from --monitor-performance 

nvoxland:
## Impact

## Description

By default, the file generated by `--monitor-performance` included system information such as environment variables and the CLI arguments which can potentially include sensitive information.

This disables capturing that information so performance results can be shared safely.

## Things to be aware of

- The contents of the profiling file can be read with `jfr metadata filename.jfr` and/or `jfr print --events "*" filename.jfr`
- Only impacts the --monitor-performance setup

## Things to worry about

- Nothing


MalloD12:
I have tested this setting `liquibase.monitorPerformance=true` and Java (Temuriun) 17 against master and this branch. In none of the cases I've seen any of these sensitive info:

```
- "jdk.InitialEnvironmentVariable"
- "jdk.InitialSystemProperty"
- "jdk.SystemProcess"
- "jdk.JVMInformation"
```

Below few questions for you @nvoxland: 

- Is there any specific setup I need to perform? Or any specific environment variable should be included as part of the results if this fix is not applied?
- This is a minor thing and not related to changes made in this PR, but it's with the configuration used for this. **MONITOR_PERFORMANCE** configuration is defined as below:
```
builder.define("monitorPerformance", String.class)
     .setDescription(<DESCRIPTION>)
      .setDefaultValue("false")
      .build();
```
Do you mind if change this to a Boolean instead of a String?

MalloD12 (approved):
**Approved**.

Code changes look good. Re-tested and compare master agains fix branch and verified fix is working working as expected. No sensitive information can be included now.

suryaaki2 (approved):

filipelautert (approved):

PR #3647, opened 2023-01-05
Add mdc to update command (DAT-12602)

abrackx:
## Impact

## Description
Adds mdc values for the update command.

## Things to be aware of
Opted to create a `MdcKey` class to use to keep track of all relevant keys, as well as an `MdcValue` class for any custom values that currently didn't map to something. 

## Things to worry about
This technically reaches beyond just the update command, but there's not much I can do about that without refactoring a good amount of code. Not necessary for a POC imo.

## Additional Context
N/A


liquibase-core/src/main/java/liquibase/integration/commandline/Main.java
```diff
@@ -1832,6 +1844,8 @@ protected void doMigration() throws Exception {
 
             try {
                 if (COMMANDS.UPDATE.equalsIgnoreCase(command)) {
+                    Scope.getCurrentScope().addMdcValue(MdcKey.LIQUIBASE_COMMAND_NAME, COMMANDS.UPDATE);
```
StevenMassaro, line 1847: Can you put the line for LIQUIBASE_COMMAND_NAME a level higher, and use the `command` variable instead of `COMMANDS.UPDATE`? Then we'd get this for free for most commands.

StevenMassaro (approved):

StevenMassaro:
Nice work here, this looks great!

liquibase-core/src/main/java/liquibase/integration/commandline/Main.java
abrackx: Great idea, updated!

filipelautert (approved):
All good!

suryaaki2:
@abrackx Can we add some tests?

wwillard7800 (approved):

PR #3042, opened 2022-07-05
Update Liquibase.java

kataggart:
Changeset Consistency Naming conventions #3041

## Impact

- [ ] Bug fix (non-breaking change which fixes expected existing functionality)
- [x ] Enhancement/New feature (adds functionality without impacting existing logic)
- [ ] Breaking change (fix or feature that would cause existing functionality to change)
 
## Description

A clear and concise description of the change being made.  

Steps To Reproduce

Run
liquibase history
against a database that has been updated by at least one changeset

Run
liquibase status
against a database that has at least one changeset needed to be updated.
Actual Behavior

The history command uses the name changesets
Database updated at 1/28/22, 4:53 PM. Applied 97 changeset(s) in 82.304s, DeploymentId: 3410389992

The status command calls uses the name change sets
2 change sets have not been applied to

Expected/Desired Behavior

Both commands use the same terminology
updated to changesets.

NOTE: I found one other case of Change sets that I also updated to Changesets

PR #3656, opened 2023-01-09
Make Pattern instance variables and avoid re calculating each time.

arturobernalg:
## Impact

## Description

Java [Pattern](http://docs.oracle.com/javase/7/docs/api/java/util/regex/Pattern.html) objects are thread safe and immutable (its the matchers that are not thread safe).

As such, there is no reason not to make them static if they are going to be used by each instance of the class (or again in another method in the class).

Making them instance variables, no matter how short (or long) their life time means that you are recompiling the regular expression each time you create an instance of a class.

One of the key reasons for this structure (Pattern being a factory for Matcher objects) is that compiling the regular expression into its finite automata is a moderately expensive action. However, one finds that often the same regular expression is used again and again in a given class (either through multiple invocations of the same method or different spots in the class).

## Things to be aware of

## Things to worry about

- List unknowns you have


nvoxland:

It is a good cleanup, thanks.

Questions I have:
- None

Potential risks:
- None

- Nothing

liquibase-core/src/main/java/liquibase/changelog/ChangelogRewriter.java
```diff
@@ -15,6 +15,12 @@
 import java.util.regex.Pattern;
 
 public class ChangelogRewriter {
+
+    public static final String XSD_PATTERN_STRING = "([dbchangelog|liquibase-pro])-3.[0-9]?[0-9]?.xsd";
```
MalloD12, line 19: I was thinking maybe can append a REGX/REGEX at the end of these regex constants. Also, I would lightly change the name of some of the patterns changing `XSD_PATTERN` for `XSD_NAME_PATTERN`. What do you think about it?

liquibase-core/src/main/java/liquibase/changelog/ChangelogRewriter.java
```diff
@@ -15,6 +15,12 @@
...
+    public static final Pattern XSD_PATTERN = Pattern.compile(XSD_PATTERN_STRING);
+    private static final String PATTERN_STRING = "(?ms).*<databaseChangeLog[^>]*>";
+    private static final Pattern PATTERN = Pattern.compile(PATTERN_STRING);
```
MalloD12, line 22: Also, maybe changing this to something a bit more representative. Do you agree?

liquibase-core/src/main/java/liquibase/database/OfflineConnection.java
```diff
@@ -47,11 +47,13 @@ public class OfflineConnection implements DatabaseConnection {
     private boolean sendsStringParametersAsUnicode = true;
     private String connectionUserName;
 
+    private static final Pattern PATTERN = Pattern.compile("offline:(\\w+)\\??(.*)");
```
MalloD12, line 50: Same here, regarding naming. Also, I would follow the same convention of extracting the regex as a separate constant.

liquibase-core/src/main/java/liquibase/datatype/core/TimeType.java
```diff
@@ -20,6 +20,7 @@
 public class TimeType extends LiquibaseDataType {
 
     protected static final int MSSQL_TYPE_TIME_DEFAULT_PRECISION = 7;
+    public static final Pattern PATTERN = Pattern.compile("(\\(\\d+\\))");
```
MalloD12, line 23: Same here, regarding naming and regex as a separate constant.

MalloD12:
Hi, @arturobernalg! 

Thanks again for submitting another PR to enhance the readability of our code. I have left you a few comments all regarding the same (pattern naming and extracting regex as a separate constant). Let me know what do you think about it. 

Daniel.

MalloD12:
Hi @arturobernalg, thank you for taking the time to address my review comments. If it's not a time-consuming thing for you, would you mind updating other regex updated in this PR as a separate constant and renaming some of the remaining generic `PATTERN` names? Otherwise, if you don't mind I can do it later?

Thanks,
Daniel.

arturobernalg:
> Hi, @arturobernalg!
> 
> Thanks again for submitting another PR to enhance the readability of our code. I have left you a few comments all regarding the same (pattern naming and extracting regex as a separate constant). Let me know what do you think about it.
> 
> Daniel.

Hi @MalloD12 
You're right. I was probably lazy that day. I have changed the last ones that I have modified. But there are still others missing. If you can wait I'll look at them today.
I just changed. let me know what do you think
TY

MalloD12 (approved):

Code looks good to me now. Thanks @arturobernalg for another PR and for applying my suggestions.

### Things to be aware of:
- None

### Things to worry about:
- None

filipelautert (approved):

suryaaki2 (approved):

liquibase-core/src/main/java/liquibase/changelog/ChangelogRewriter.java
```diff
@@ -15,6 +15,12 @@
 import java.util.regex.Pattern;
 
 public class ChangelogRewriter {
+
+    public static final String XSD_FILE_REGEX = "([dbchangelog|liquibase-pro])-3.[0-9]?[0-9]?.xsd";
```
XDelphiGrl, line 19: Do we need to take `*-latest` and version `4-` into account for any of these objects?

liquibase-core/src/main/java/liquibase/database/AbstractJdbcDatabase.java
```diff
@@ -88,15 +88,32 @@ public abstract class AbstractJdbcDatabase implements Database {
...
+    private static final String DATE_ONLY_REGEX = "^\\d{4}\\-\\d{2}\\-\\d{2}$";
+    private static final Pattern DATE_ONLY_PATTERN = Pattern.compile(DATE_ONLY_REGEX);
+
+    private static final String DATE_TIME_REGEX = "^\\d{4}\\-\\d{2}\\-\\d{2}[T ]\\d{2}:\\d{2}:\\d{2}(?:\\.\\d+)?$";
+    private static final Pattern DATE_TIME_PATTERN = Pattern.compile(DATE_TIME_REGEX);
+
+    private static final String TIMESTAMP_REGEX = "^\\d{4}\\-\\d{2}\\-\\d{2}T\\d{2}:\\d{2}:\\d{2}\\.\\d+$";
+    private static final Pattern TIMESTAMP_PATTERN = Pattern.compile(TIMESTAMP_REGEX);
+
+    private static final String TIMES_REGEX = "^\\d{2}:\\d{2}:\\d{2}$";
```
XDelphiGrl, line 109: Should this be `TIME_REGEX` instead of `TIMES_REGEX`?

liquibase-core/src/main/java/liquibase/changelog/ChangelogRewriter.java
MalloD12: This hasn't changed in the last two years. @arturobernalg just made the change of extracting this regular expression into a constant. I'm trying to understand whether this regular expression should be updated to include these versions (`4-` and `-latest`) or not.

XDelphiGrl: We seem to define a similar Pattern in multiple classes. In other places the regex is `dbchangelog-[\\w\\.]+.xsd` or `(?:-pro-|-)(?<version>[\\d.]*)\\.xsd`. Any chance we can use identical regex in all the different places? What about having the XSD pattern defined in one place?

liquibase-core/src/main/java/liquibase/parser/core/xml/StandardNamespaceDetails.java
```diff
@@ -11,12 +11,15 @@
 public class StandardNamespaceDetails implements NamespaceDetails {
 
     public static final String GENERIC_EXTENSION_XSD = "http://www.liquibase.org/xml/ns/dbchangelog/dbchangelog-ext.xsd";
-    private final Pattern standardUrlPattern;
-    private final Pattern oldStandardUrlPattern;
+    private static final String STANDARD_XSD_URL_REGEX = "http://www.liquibase.org/xml/ns/dbchangelog/(dbchangelog-[\\w\\.]+.xsd)";
+    private static final Pattern STANDARD_XSD_URL_PATTERN = Pattern.compile(STANDARD_XSD_URL_REGEX);
+
+    private static final String OLD_STANDARD_XSD_URL_REGEX = "http://www.liquibase.org/xml/ns/migrator/(dbchangelog-[\\w\\.]+.xsd)";
```
XDelphiGrl, line 17: Wow, this must be a _really_ old URL, for sure pre-3.0? Interesting new fact I learned from reading this!

liquibase-core/src/main/java/liquibase/changelog/ChangelogRewriter.java
MalloD12: I asked @nvoxland about it, and he was pointing me out that changelogID is a changeset attribute that has been introduced in version `4-` so we should be checking to add this field (changelogID) for version 3-0 (not worth doing it also for earlier versions).

liquibase-core/src/main/java/liquibase/database/AbstractJdbcDatabase.java
MalloD12: Hi @arturobernalg, 

Would you mind making this minor update?

Thanks,
Daniel.

liquibase-core/src/main/java/liquibase/parser/core/formattedsql/FormattedSqlChangeLogParser.java
```diff
@@ -26,6 +26,155 @@
 public class FormattedSqlChangeLogParser implements ChangeLogParser {
 
 
+    private static final String FIRST_LINE_REGEX = "\\-\\-\\s*liquibase formatted.*";
```
XDelphiGrl, line 29: The standard for using the word changeset is one word, not two (change set). The Strings and Pattern names that have `*CHANGE_SET*` should be `*CHANGESET*` for consistency.

liquibase-core/src/main/java/liquibase/parser/core/formattedsql/FormattedSqlChangeLogParser.java
```diff
@@ -26,6 +26,155 @@
...
+
+    private static final String ROLLBACK_MULTI_LINE_START_REGEX = "\\s*\\/\\*\\s*liquibase\\s*rollback\\s*$";
+    private static final Pattern ROLLBACK_MULTI_LINE_START_PATTERN = Pattern.compile(ROLLBACK_MULTI_LINE_START_REGEX, Pattern.CASE_INSENSITIVE);
+
+
+    private static final String WORD_RESULT_REGEX = "^(?:expectedResult:)?(\\w+) (.*)";
+    private static final String SINGLE_QUOTE_RESULT_REGEX = "^(?:expectedResult:)?'([^']+)' (.*)";
+    private static final String DOUBLE_QUOTE_RESULT_REGEX = "^(?:expectedResult:)?\"([^\"]+)\" (.*)";
+
+    private static final Pattern[] PATTERNS = new Pattern[]{
```
XDelphiGrl, line 172: I see the Pattern[] was named `patterns` in the previous code. That wasn't very confusing because `patterns` was declared where it was being used. Now that it is moved to a different location, a more descriptive name would be useful. Perhaps something like WORD_AND_QUOTING_PATTERNS?

XDelphiGrl (changes requested):
@arturobernalg, hi! I went regex by regex and copied and pasted to a text file to make sure nothing changed. I have to say, WOW! I did not find a single unintentional change to regex; the best I could come up with is what looks like a typo in one of the regex/pattern names (see comments). 

I do have a couple of additional comments for you or @MalloD12 to address. The primary ask is to please be consistent in naming for "changeset" (one word, not two). The secondary ask is to double-check the regex used for XSD pattern matching. Seems like we have a few ways of defining the regex so I'm curious if we can settle on one regex (or even one place where this is defined). 

As always, thank you so much for the code improvements, @arturobernalg !!!

liquibase-core/src/main/java/liquibase/database/AbstractJdbcDatabase.java
arturobernalg: Done

liquibase-core/src/main/java/liquibase/parser/core/formattedsql/FormattedSqlChangeLogParser.java
arturobernalg: Done

filipelautert:
@XDelphiGrl all that you request is done right?

liquibase-core/src/main/java/liquibase/parser/core/formattedsql/FormattedSqlChangeLogParser.java
XDelphiGrl: "Changeset" is one word so it shouldn't be split with the `_` in the variable names. Does that make more sense, @arturobernalg? Maybe? :)

MalloD12: Sorry @XDelphiGrl, is this not the regex to match the first line of a changelog in sql format?

`-- liquibase formatted sql`

XDelphiGrl: That is correct, @MalloD12. That is the correct regex for the starting line of a formatted SQL changelog.

I think part of the confusion is I added my comment on a line where it doesn't make sense. I put it on the first line in the file because it applies to multiple lines BUT I can see where this is confusing. The line for an example is Line 39 (CHANGE_SET_REGEX ... I suggest we change it to CHANGESET_REGEX). 

My issue is just with the names of the variables being "CHANGE_SET" instead of "CHANGESET". We've done some cleanup lately to be consistent:
[https://github.com/liquibase/liquibase/pull/3042](https://github.com/liquibase/liquibase/pull/3042).

XDelphiGrl: After discussion with @nvoxland and @filipelautert, we're going to stay with the `CHANGE_SET` and not make a change to `CHANGESET`. Resolving this conversation.

XDelphiGrl (approved):
Liquibase uses regular expressions in a multitude of places. As @arturobernalg mentions in the PR description, instance variables of regular expressions introduce a computational load compared to constant definitions for instance variables. This PR "converts" instance variables of regular expressions to constants.
* Functional and test harness executions did pass prior to this branch becoming out of date with master.
  * Functional test failures are related to the new `history` command `format` property. 
  * Tests will pass when master merges to this branch.

APPROVED

PR #3657, opened 2023-01-09
Remove Unnecessary operation/modifier:

arturobernalg:
## Impact

## Description

Remove Unnecessary operation/modifier:

* Unnecessary call to 'toString()'
* Unnecessary conversion to 'String'
* Unnecessary modifier
* Unnecessary semicolon

## Things to be aware of

- None

## Things to worry about

- None 

## Additional Context

- Nothing


nvoxland:

When you're making the PR, @arturobernalg, if the default template gets in the way it's better to clear out the unused stuff than leave the placeholder text there. 

You're description at the top is good, so you could move that to the normal "description" section and then just delete the "things to be aware of" and the rest. If you had anything to add in the other sections you could have, but in cases like this where it's a straightforward change it can get to be just noise if you leave them in with simple "nothing" entries.


Questions I have:
- None

Potential risks:
- None

- Nothing

liquibase-core/src/main/java/liquibase/database/AbstractJdbcDatabase.java
```diff
@@ -1270,7 +1270,7 @@ public void execute(final SqlStatement[] statements, final List<SqlVisitor> sqlV
                 Scope.getCurrentScope().getSingleton(ExecutorService.class).getExecutor("jdbc", this).execute(statement, sqlVisitors);
             } catch (DatabaseException e) {
                 if (statement.continueOnError()) {
-                    Scope.getCurrentScope().getLog(getClass()).severe("Error executing statement '" + statement.toString() + "', but continuing", e);
+                    Scope.getCurrentScope().getLog(getClass()).severe("Error executing statement '" + statement + "', but continuing", e);
```
MalloD12, line 1273: Where's the statement taking the toString() from?

liquibase-core/src/main/java/liquibase/snapshot/SnapshotIdService.java
```diff
@@ -18,6 +18,6 @@ private SnapshotIdService() {
     }
 
     public String generateId() {
-        return base+Integer.toString(nextId++);
+        return base+ nextId++;
```
MalloD12, line 21: Please correct me if I'm wrong, but I think it's a bit more efficient doing something like: `String.format("%s%d", base, nextId);`, instead of this explicit conversion.

MalloD12:
@arturobernalg This PR is overall LGTM. I just left two questions for which you probably have a better fundament than me. I like this improvement in removing the unnecessary `toString()` calls. 

Thanks,
Daniel.

liquibase-core/src/main/java/liquibase/snapshot/SnapshotIdService.java
MalloD12: @arturobernalg please, ignore my previous comment. After reviewing your `StringBuilder.append()` [PR](https://github.com/liquibase/liquibase/pull/3668). I've been reading a bit more and found StringBuilder is more efficient than `String.format()`, and also that even though String concatenation is less efficient than StringBuilder,  it's better than format. I think for this particular case we can either leave it as it is considering it's a small concatenation or in alignment with your other PR update this method making use of StringBuilder as well.

MalloD12 (approved):
**Approved**

PR changes LGTM. Validate `toString()` method calls were correctly removed because they were being use in most of the cases a string concatenation, and also in most of these cases there was already a implemented `toString()` method for the classes being called in these mentioned string concatenations. Found a case for which there was not a definition for that I've created this enhancement [Issue](https://github.com/liquibase/liquibase/issues/3746). Build and tests have been successfully executed.

### Things to be aware of:
- None

### Things to worry about:
- None

filipelautert (approved):

XDelphiGrl (approved):
PR cleans up instances where `toString()` was applied to objects that have an implicit "toString()." PR also removes a couple of extra `;` in `if` blocks. 

APPROVED

Issue #3587, opened 2022-12-16
'liquibase status' is broken for Oracle


SIEDA-EliasHeydrich:
## Environment
**Liquibase Version**: 4.18.0 (but at least 4.16.1 was already affected too)
**Liquibase Integration & Version**: None, we just use a batch-script
**Database Vendor & Version**: Oracle 19
**Operating System Type & Version**: We use Microsoft Infrastructure, Win10 equivalents on DB-Server and Client
**Infrastructure Type/Provider**: DB-Server runs on a VMWare-VM

## Description
The `status` command reports 100% of changelogs inside a file as not executed, although they have been executed prior.
The documented behavior that it generates _a list of pending changesets_ is thus not fulfilled.

## Steps To Reproduce
* provide a DB and a list of changesets
* employ _Liquibase_'s `update` command to update the DB
* try `status`

For details, see below:

## Actual Behavior
Here is a sequence of console commands with all private and personal information redacted/replaced.
It shows that the behavior of the `status` command is incorrect, as it continues to report 34 changesets as not executed.
```
liquibase.bat --url="jdbc:oracle:thin:@ORA19SAS:1521/ORCL19SASPDB" --liquibase-schema-name="our-thingy" --default-schema-name="our-thingy" --username="our-thingy" --password="***" --changelog-file="changesets.xml" --headless=true --file-encoding="UTF-8" history
Starte Liquibase am 11:20:18 (Version 4.18.0 #5864, kompiliert am 2022-12-02 18:02+0000)
Liquibase Version: 4.18.0
Liquibase Community 4.18.0 by Liquibase
Liquibase History for jdbc:oracle:thin:@ORA19SAS:1521/ORCL19SASPDB

- Database updated at 14.12.22, 16:12. Applied 33 changeset(s) in 17.406s, DeploymentId: 1030919883
  file001::0::personA
  file001::1::personB
  file001::2::personB
  file001::3::personB
  file001::4::personB
  file001::5::personB
  file001::6::personA
  file001::7::personA
  file001::8::personA
  file001::pre-9::personA
  file001::9::personB
  file001::10::personA
  file001::11::personA
  file001::12::personB
  file001::13::personA
  file001::14::personA
  file001::15::personA
  file001::16::personA
  file001::17::personA
  file001::18::personA
  file001::19::personA
  file001::20::personA
  file001::21::personA
  file001::22::personA
  file001::23::personA
  file001::post-23::personA
  file001::24::personA
  file001::25a::personB
[...]

PR #3662, opened 2023-01-10
Consistent ARG-suffix for 'defaultSchemaName'-Parameter

SIEDA-EliasHeydrich:
## Impact
Minimal, ensures consistent code in all CLI-Command classes.
 
## Description

All **but two** _CommandStep_-classes employ the following naming scheme for the  'defaultSchemaName'-Parameter:

`public static final CommandArgumentDefinition<String> DEFAULT_SCHEMA_NAME_ARG;`

Except two classes, which used:

`public static final CommandArgumentDefinition<String> DEFAULT_SCHEMA_NAME;`

This PR fixes this.

## Things to worry about

None 😉 

## Additional Context

I started to investigate what the root cause of #3587 could be and noticed this while code-reading 😉

### DIFF
--- a/liquibase-core/src/main/java/liquibase/command/core/UpdateCommandStep.java
+++ b/liquibase-core/src/main/java/liquibase/command/core/UpdateCommandStep.java
@@ -11,7 +11,7 @@ public class UpdateCommandStep extends AbstractCliWrapperCommandStep {
 
     public static final CommandArgumentDefinition<String> CHANGELOG_FILE_ARG;
     public static final CommandArgumentDefinition<String> URL_ARG;
-    public static final CommandArgumentDefinition<String> DEFAULT_SCHEMA_NAME;
+    public static final CommandArgumentDefinition<String> DEFAULT_SCHEMA_NAME_ARG;
     public static final CommandArgumentDefinition<String> DEFAULT_CATALOG_NAME_ARG;
     public static final CommandArgumentDefinition<String> USERNAME_ARG;
     public static final CommandArgumentDefinition<String> PASSWORD_ARG;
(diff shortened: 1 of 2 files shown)


nvoxland:

Thanks for the fix, and keeping it separate.

In theory, other code may have referenced the old public static object. So it would be best to keep the old variable name but mark it as deprecated and just be a reference to the _ARG version. Eventually we can remove the old deprecated version. 

Questions I have:
- None

Potential risks:
- Just the other code issue.

- Nothing

SIEDA-EliasHeydrich:
> In theory, other code may have referenced the old public static object. So it would be best to keep the old variable name but mark it as deprecated and just be a reference to the _ARG version. Eventually we can remove the old deprecated version.

Building the code should reveal such issues but the CI is blocked because you have not marked it as safe (this whole _PR from Fork is NOT safe to build_-thingy).

filipelautert (approved):
Both params are working now.

suryaaki2 (approved):

XDelphiGrl (approved):
PR deprecates `DEFAULT_SCHEMA_NAME` in UpdateCommandStep and UpdateTestingRollbackCommand in favor of using `DEFAULT_SCHEMA_NAME_ARG`; the `*_ARG` naming convention was used everywhere except in these two classes. External callers using `DEFAULT_SCHEMA_NAME` should update code to use `DEFAULT_SCHEMA_NAME_ARG` in preparation for future removal of the deprecated variable name.
* Functional tests passing.
* There are test harness failures related to SQL Server table and column remarks; these are unrelated to this change and are not a blocker to merging this PR to master.

APPROVED

PR #3666, opened 2023-01-10
Delete unused comment code.

arturobernalg:
## Impact

## Description

Code that is commented out gets outdated very quickly and becomes misleading. It is better to delete commented out code completely and use the VCS history instead.

## Things to be aware of

## Things to worry about

- List unknowns you have


nvoxland:
This one will take a little more to look through. I'll sometimes leave code commented that will be a useful reminder in an upcoming expected change. While deleting it does leave it in history, people don't think to look there when they end up needing it.

Sometimes that expected change never comes and it's good to clean out, sometimes that time came and went and it needs to be cleaned out, sometimes it should still stay. I'll have to see if any of these seem worth keeping for a while.

MalloD12:
> This one will take a little more to look through. I'll sometimes leave code commented that will be a useful reminder in an upcoming expected change. While deleting it does leave it in history, people don't think to look there when they end up needing it.
> 
> Sometimes that expected change never comes and it's good to clean out, sometimes that time came and went and it needs to be cleaned out, sometimes it should still stay. I'll have to see if any of these seem worth keeping for a while.

I like this PR and I think I have been also tempted to delete some commented code as well in the past months. @nvoxland I think it would be nice to start removing cases like this, and only in exceptional cases we can leave some commented-out code but leaving a comment/reason why it has not been/cannot be deleted. If we want to keep a record of some deleted code I'm thinking I can create a confluence page and track the deleted code in there with a GH reference to the commit where these changes have been applied. What would you think about it? (cc: @filipelautert )

filipelautert:
@MalloD12 I verified some random samples,  and there is code that have been commented out since 2009. So I think it's fine to remove it, specially because there should be better ways to handle it nowadays :)

image

MalloD12:
> @MalloD12 I verified some random samples, and there is code that have been commented out since 2009. So I think it's fine to remove it, specially because there should be better ways to handle it nowadays :)
> 
> image

That's exactly what I was pointing out, but I think Nathan's point is valid. Although, I think we can implement a different idea/strategy (like the one I was talking about in my previous comment) if for some reason we would like to keep track of some specific code that has been commented out.

MalloD12 (approved):
**Approved**.

Thank you @arturobernalg. Glad you removed this unused code which from what I saw in most of the cases it was a very old code (around 10 years old). From the updated classes I have also removed some existing commented code.

### Things to be aware of:
- None

### Things to worry about:
- None

filipelautert (approved):

PR #3667, opened 2023-01-10
DAT-6636: Refactor of TagExistsCommandStep and extraction of code to new LockServiceCommandStep 

filipelautert:
## Impact

## Description

* Migrates code from Liquibase to TagExists Command step
* Create LockServiceCommandStep
* Refactor CommandStep ordering login: instead of using hardcoded order values to order the steps, introduces a dependency system where a step requires or provides resources and those dependencis are used to sort out the steps.
* Removes outdated methods and classes.

## Things to be aware of
* Simpler ordering way
* Requires integration test fixes from https://github.com/liquibase/liquibase-pro/pull/795

## Things to worry about
- The removed methods should be deprecated instead of removed?

## Additional Context
- Finishes the Command framework refactoring started with TagCommand.


liquibase-core/src/main/java/liquibase/command/AbstractCommandStep.java
```diff
@@ -1,29 +1,23 @@
...
-            }
-        }
-        return ORDER_NOT_APPLICABLE;
+    public List<Class<?>> requiredDependencies() {
+        return Collections.emptyList();
+    }
+
+    @Override
+    public List<Class<?>> providedDependencies() {
+        return Collections.emptyList();
```
StevenMassaro, line 20: Do we have a plan in mind for how to gracefully migrate any extensions or users who have classes which extend this class and make use of the `getOrder` method?

liquibase-core/src/main/java/liquibase/command/CommandFactory.java
```diff
@@ -63,13 +62,64 @@ public CommandDefinition getCommandDefinition(String... commandName) throws Ille
...
+                pipelineGraph.add(provider, step);
+                findDependenciesForCommand(pipelineGraph, allCommandStepInstances, provider);
+            }
+        }
+    }
 
-        return commandDefinition;
+    private CommandStep whoProvidesClass(Class<?> d, Collection<CommandStep> allCommandStepInstances) {
+        return allCommandStepInstances.stream().filter(cs -> cs.providedDependencies().contains(d))
+                .findFirst().orElseThrow(() -> new RuntimeException("Unable to find CommandStep provider for class " +  d.getName()));
```
StevenMassaro, line 122: Please add javadocs and inline documentation to these methods, which appear to be cognitively complex.

liquibase-core/src/main/java/liquibase/command/CommandScope.java
```diff
@@ -122,6 +124,16 @@ public <T> T getArgumentValue(CommandArgumentDefinition<T> argument) {
...
 
+    public  <T> CommandScope provideDependency(Class<T> clazz, T value) {
+        this.dependencies.put(clazz, value);
+
+        return this;
+    }
+
+    public <T> Object getDependency(Class<T> clazz) {
+        return this.dependencies.get(clazz);
+    }
```
StevenMassaro, line 135: Please add javadoc here, explaining what a dependency is.

liquibase-core/src/main/java/liquibase/command/core/TagExistsCommandStep.java
```diff
@@ -1,52 +1,47 @@
...
-                .description("Username to use to connect to the database").build();
-        PASSWORD_ARG = builder.argument(CommonArgumentNames.PASSWORD, String.class)
-                .description("Password to use to connect to the database")
-                .setValueObfuscator(ConfigurationValueObfuscator.STANDARD)
-                .build();
-        TAG_ARG = builder.argument("tag", String.class).required()
-                .description("Tag to check").build();
+        TAG_ARG = builder.argument("tag", String.class).required().description("Tag to check").build();
+
+        TAG_EXISTS_RESULT = builder.result("tagExistsResult", Boolean.class).build();
```
StevenMassaro, line 24: What is this new property? Can you add a description to it or make it hidden?

liquibase-integration-tests/src/test/resources/liquibase/extension/testing/command/tagExists.test.groovy
```diff
@@ -34,10 +35,6 @@ Optional Args:
                 password:   { it.password },
                 tag: "version_2.0",
         ]
-
-        expectedResults = [
-                statusCode   : 0
-        ]
```
StevenMassaro, line 40: Is the removal of this intentional?

liquibase-core/src/main/java/liquibase/command/AbstractCommandStep.java
StevenMassaro: Does the pro code compile successfully with the removal of the getOrder method?

liquibase-core/src/main/java/liquibase/command/CommandFactory.java
abrackx: 👍 Some comments here would be very helpful.

liquibase-core/src/main/java/liquibase/command/AbstractCommandStep.java
abrackx: I think we'll need to deprecate `getOrder`for a while and not remove it yet. I wasn't able to build core and pro with this change, and I'd expect some other extensions may throw a fit too.

liquibase-core/src/main/java/liquibase/command/core/LockServiceCommandStep.java
```diff
@@ -0,0 +1,61 @@
...
+
+    @Override
+    public List<Class<?>> requiredDependencies() {
+        return Collections.singletonList(Database.class);
+    }
+
+    @Override
+    public List<Class<?>> providedDependencies() {
+        return Collections.singletonList(LockService.class);
+    }
```
abrackx, line 31: Smells like DI in here 😎

liquibase-core/src/main/java/liquibase/command/core/LockServiceCommandStep.java
```diff
@@ -0,0 +1,61 @@
...
+
+    @Override
+    public void run(CommandResultsBuilder resultsBuilder) throws Exception {
+        CommandScope commandScope = resultsBuilder.getCommandScope();
+        Database database = (Database) commandScope.getDependency(Database.class);
```
abrackx, line 36: Can you explain how this Database dependency is getting set up?

liquibase-core/src/main/java/liquibase/command/AbstractCommandStep.java
filipelautert: I only verified core code and it does not use order, but I forgot that pro does. I rolledback and deprecated getOrder, also changing the logic to work with order and dependencies. I created a branch for this issue on pro and it is building now.
I think that in the next 2 or 3 commands this solution will become stable and then we can go ahead and refactor pro/other extensions code to get rid of getOrder.

liquibase-integration-tests/src/test/resources/liquibase/extension/testing/command/tagExists.test.groovy
filipelautert: Yes, the status code is not returned when we change the implementation to AbstractCommandStep . I discussed this with Nathan and the idea is that if we have a failure an Exception will be thrown, so we don't need to expect anything here.

liquibase-core/src/main/java/liquibase/command/core/LockServiceCommandStep.java
filipelautert: That's the plan! Maybe in the future we can have an @Autowired annotation 😃

liquibase-core/src/main/java/liquibase/command/core/LockServiceCommandStep.java
filipelautert: LockServiceCS  requires Database.class , and DbUrlCS provides Database.class . During command evaluation at CommandFactory, in the same place that we had the getOrder before now we validate those dependencies and the command steps are added to the CommandDefinition pipeline in the correct order. 

So: 
* CommandFactory finds a command like "TagExists".
*  TagExists requires _LockService_, 
* Factory finds who provides it (LockServiceCS) and adds it to the pipeline before TagExistsCs. 
* Now Factory verifies that LockServiceCS requires _Database_
* Factory finds who provides it (DbUrlCS) and adds it to the pipeline before LockService. 
* DBUrlCS does not requires anything else, so it end.

At the end we have a CommandDefinition with a pipeline of:
* **DbUrlCS** (create database) -> **LockServiceCS** (create lockservice and locks) -> **TagExistsCS** (does the tag magic) . 

Now, LockServiceCS and DbUrlCS also implement **CleanUpCommandStep**. After the commands executions, the pipeline is executed in the reverse order and if the step implements CleanUpCS the cleanUp method is executed. So:
* **LockServiceCS** (releases the lock) -> **DbUrlCS** (closes db connection)

StevenMassaro: What happens if multiple classes register themselves as a provider of `LockService`?

filipelautert: Right now it is a "findFirst" filter on the array of CommandSteps. And if nothing provides it an exception is thrown.

 But in next commands we'll need to come up with a way to define the priority as it may happen, specially when working with comands like DiffChangelog. Maybe something like the `@order(1) `used by spring?

liquibase-core/src/main/java/liquibase/command/core/TagExistsCommandStep.java
filipelautert: This new property is the output from execution that can obtained by calling `commandResults.getResult(TagExistsCommandStep.TAG_EXISTS_RESULT); `
I'm adding the description .

liquibase-core/src/main/java/liquibase/command/core/LockServiceCommandStep.java
StevenMassaro: I think you should throw an exception if there are multiple providers found. Until we add an ordering mechanism, randomly selecting one of the providers could cause inconsistent behavior. The exception would also make it very clear that whatever you just developed is not working properly.

liquibase-core/src/main/java/liquibase/command/core/TagExistsCommandStep.java
StevenMassaro: I am a bit confused though. Why does the tag exists command define its own result as a command argument?

liquibase-core/src/main/java/liquibase/command/core/LockServiceCommandStep.java
filipelautert: Makes sense! will do it. I'll also add the @Beta annotations to the methods to advise anyone trying to use it.

liquibase-core/src/main/java/liquibase/command/core/TagExistsCommandStep.java
filipelautert: It's not an argument.. It's a result :)
*Argument: commandScope.getArgumentValue(TagCommandStep.TAG_ARG)));
* Result:  resultsBuilder.addResult(TAG_EXISTS_RESULT, true);

And at Liquibase class you can see both of them in use:
```
        CommandResults commandResults = new CommandScope("tagExists")
                .addArgumentValue(DbUrlConnectionCommandStep.DATABASE_ARG, database)
                .addArgumentValue(**TagExistsCommandStep.TAG_ARG**, tagString)
                .execute();
        return commandResults.getResult(**TagExistsCommandStep.TAG_EXISTS_RESULT**);
```

StevenMassaro: Got it, that makes more sense. Thanks for adding the description!

liquibase-core/src/main/java/liquibase/command/CommandFactory.java
filipelautert: Done!

StevenMassaro (approved):

liquibase-core/src/main/java/liquibase/command/CommandDefinition.java
```diff
@@ -36,12 +36,16 @@ protected CommandDefinition(String[] name) {
...
             final int order = Integer.compare(o1.getOrder(this), o2.getOrder(this));
             if (order == 0) {
-                return o1.getClass().getName().compareTo(o2.getClass().getName());
+                // if o1 requires o2 dependencies it must come after o2
+                for (Class<?> required : o1.requiredDependencies()) {
+                    if (o2.providedDependencies().stream().anyMatch(required::equals)) {
+                        return 1;
+                    }
+                }
+                return -1;
```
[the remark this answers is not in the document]
filipelautert: Makes sense, thanks!

suryaaki2 (approved):

PR #3695, opened 2023-01-16
DAT-12039: detect circular references in includeAll statement

StevenMassaro:
## Impact

## Description

Throw an exception if a circular reference is detected in an includeAll statement.

## Things to be aware of

- This can be enabled/disabled via a new global configuration option.
- I added the global configuration option on my own, I'm not sure what product wants to call the property, so it may change.

## Things to worry about

False positive detection.

### DIFF
--- a/liquibase-core/src/main/java/liquibase/changelog/DatabaseChangeLog.java
+++ b/liquibase-core/src/main/java/liquibase/changelog/DatabaseChangeLog.java
@@ -11,6 +11,7 @@ import liquibase.database.ObjectQuotingStrategy;
 import liquibase.exception.*;
 import liquibase.logging.Logger;
 import liquibase.parser.ChangeLogParser;
+import liquibase.parser.ChangeLogParserConfiguration;
 import liquibase.parser.ChangeLogParserFactory;
 import liquibase.parser.core.ParsedNode;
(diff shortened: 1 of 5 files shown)


abrackx (approved):
Nice!

liquibase-core/src/main/java/liquibase/changelog/DatabaseChangeLog.java
```diff
@@ -656,6 +659,13 @@ public void includeAll(String pathName,
                 if (StringUtil.isNotEmpty(path) && !(path.endsWith("/"))) {
                     path = path + '/';
                 }
+
+                if (ChangeLogParserConfiguration.ERROR_ON_CIRCULAR_INCLUDE_ALL.getCurrentValue()) {
+                    if (seenChangelogPaths.contains(path)) {
+                        throw new SetupException("Circular reference detected in '" + path + "'. Set " + ChangeLogParserConfiguration.ERROR_ON_CIRCULAR_INCLUDE_ALL.getKey() + " if you'd like to ignore this error.");
```
filipelautert, line 665: So.. if we are adding a feature to disable this validation, should we call this an error? Maybe s/error/setting/g ?

filipelautert (approved):

PR #3696, opened 2023-01-16
Refactors CalculateChecksumCommandStep 

filipelautert:
## Impact

## Description

* Migrates code from Liquibase to CalculateChecksumCommandStep 

## Things to be aware of
I had to change the wait that  2  objects are obtained. Seems it is ok and bey the code I read I believe we are not changing the behaviour. But anyway the changes are:

1. ResourceAcessor: 
* was `this.getResourceAccessor() ; `
* now is  `Scope.getCurrentScope().getResourceAccessor(); `
2. ChangeLogParameter:
* was `this.getChangeLogParameter()`
* now is `new ChangeLogParameters(database)`

## Additional Context


liquibase-core/src/main/java/liquibase/Liquibase.java
```diff
@@ -2202,40 +2195,30 @@ public void run() throws Exception {
...
-
-        final ChangeSet changeSet = changeLog.getChangeSet(filename, author, id);
-        if (changeSet == null) {
-            throw new LiquibaseException(
-                    new IllegalArgumentException("No such changeSet: " + filename + "::" + id + "::" + author)
-            );
-        }
-
-        return changeSet.generateCheckSum();
+        return this.calculateCheckSum(String.format("%s::%s::%s", filename, id, author));
```
StevenMassaro, line 2221: Can you avoid the duplication of the string formatting that occurs in the line above?

StevenMassaro (approved):

liquibase-core/src/main/java/liquibase/Liquibase.java
filipelautert: Yes, thanks for spotting this. In fact I removed the log as it is already executed inside the command step.

suryaaki2 (approved):

suryaaki2 (approved):

liquibase-integration-tests/src/test/resources/liquibase/extension/testing/command/calculateChecksum.test.groovy
```diff
@@ -40,7 +40,7 @@ Optional Args:
         ]
 
         expectedResults = [
-                statusCode   : 0
+                checksumResult   : "8:b6084e5d5f46b534bbbe18a0d35d34e0"
```
XDelphiGrl, line 43: This test will fail when Liquibase upgrades the checksum version. That failure indicates the test is doing its job and letting us know that a change occurred, which is good. Whatever statusCode is (exit code??) probably would not fail with a checksum version upgrade, so I like this enhancement!

XDelphiGrl:
@filipelautert , hello! 

I see that we're checking the exception class type in the checksum integration tests. Could we also validate that the correct error message is returned for each of the negative test cases?

suryaaki2 (approved):

liquibase-integration-tests/src/test/resources/liquibase/extension/testing/command/calculateChecksum.test.groovy
filipelautert: Validation of error messages added!

PR #3702, opened 2023-01-17
Rework dropProcedure for Snowflake Log message when dropAll fails DAT-12916

wwillard7800:
This change only impacts dropProcedure for Snowflake

### DIFF
--- a/liquibase-core/src/main/java/liquibase/command/core/InternalDropAllCommandStep.java
+++ b/liquibase-core/src/main/java/liquibase/command/core/InternalDropAllCommandStep.java
@@ -100,6 +100,7 @@ public class InternalDropAllCommandStep extends AbstractCommandStep {
                     commandScope.getArgumentValue(DATABASE_ARG).dropDatabaseObjects(schema);
                 }
             } catch (LiquibaseException liquibaseException) {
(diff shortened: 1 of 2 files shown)


liquibase-core/src/main/java/liquibase/command/core/InternalDropAllCommandStep.java
```diff
@@ -100,6 +100,7 @@ public void run(CommandResultsBuilder resultsBuilder) throws Exception {
                     commandScope.getArgumentValue(DATABASE_ARG).dropDatabaseObjects(schema);
                 }
             } catch (LiquibaseException liquibaseException) {
+                log.severe("Error occurred during dropAll: " + liquibaseException.getMessage());
```
StevenMassaro, line 103: Good idea. Can you also include the raw exception in the log message so we get a stack trace in the logs?

```suggestion
                log.severe("Error occurred during dropAll: " + liquibaseException.getMessage(), liquibaseException);
```

wwillard7800: Yes, that makes sense.

filipelautert (approved):

StevenMassaro (approved):

suryaaki2 (approved):

Issue #3035, opened 2021-12-22
For snowflake tables having float datatype diff-changelog is generating the datatype as DOUBLE(0)


dsprof:
For a float column in snowflake liquibase diff-changelog and generate change log commands are generating the datatype as DOUBLE(0),
snowflake does not expect a size for the double datatype. A Json changelog column example below:

 "column": {
                  "name": "ABCD",
                  "type": "DOUBLE(0)"
                }

nvoxland:
This appears to be an issue still. Moving it to liquibase/liquibase since the snowflake code has moved there

filipelautert:
Fixed in PR #3703 (merged Jan 2023) which added a proper `Double` data type class to avoid the incorrect `DOUBLE(0)` output for Snowflake.

Closing this out.

PR #3703, opened 2023-01-17
#3035 add Double data type class to avoid incorrect double(0) databas…

LonwoLonwo:
add Double data type class to avoid incorrect double(0) database data type result

## Impact

## Description

As I can see, I can add a new condition to the DoubleType class, so I made these changes in the snowflake package.
Basically, this is a copy of HanaDoubleType and such kind of extension data type classes.
We fixed this problem that way in our snowflake extension fork and have used it since December. Works fine for us.

More about numeric types: https://docs.snowflake.com/en/sql-reference/data-types-numeric.html#double-double-precision-real

I can do "DOUBLE" instead of "FLOAT", but, I think, it will be not correct.

### DIFF
--- a/liquibase-snowflake/src/main/java/liquibase/datatype/core/DoubleDataTypeSnowflake.java
+++ b/liquibase-snowflake/src/main/java/liquibase/datatype/core/DoubleDataTypeSnowflake.java
@@ -1,0 +1,36 @@
+package liquibase.datatype.core;
+
+import liquibase.database.Database;
+import liquibase.database.core.SnowflakeDatabase;
+import liquibase.datatype.DataTypeInfo;
+import liquibase.datatype.DatabaseDataType;
+import liquibase.datatype.LiquibaseDataType;
+
+@DataTypeInfo(
+    name = "double",
+    aliases = {"java.sql.Types.DOUBLE", "java.lang.Double"},
+    minParameters = 0,
+    maxParameters = 2,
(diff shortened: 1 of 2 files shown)


LonwoLonwo:
Hi Liquibase Team. We want to update our liquibase jat version to the latest, but I can't it properly during your last changes. Snowflake extension is now inside your main library. We use Snowflake extension fork in our project with our changes. But with the latest changes this extension do not work anymore. So, I moved our changes in your main library and created PRs:

nvoxland:

Thanks for the fix, @LonwoLonwo . The changes make sense to me, I'll send it through the final check and review process to get it into a release.

Questions I have:
- None

Potential risks:
- None

- Nothing

filipelautert:
Added a harness test at https://github.com/liquibase/liquibase-test-harness/pull/481

filipelautert (approved):
Change specific to ❄️ , all fine according to the documentation .

XDelphiGrl (approved):
This Snowflake-specific fix allows Liquibase to correctly identify the `FLOAT` datatype. This prevents generation of invalid SQL for columns defined as `DOUBLE` during `generate-changelog` and `diff-changelog`.
* Functional tests passing.
* Failing SQLServer test-harness tests unrelated to this change; not a blocker to PR merge.
  * New test harness test added for the `DOUBLE` datatype.

APPROVED

PR #3704, opened 2023-01-17
Do not add default schema name to getSchemas method to avoid one sche…

LonwoLonwo:
Do not add default schema name to getSchemas method to avoid one schema result name

## Impact

## Description

Hello!

We are thinking about migrating our product to the latest Liquibase library version. 
Now we are using the Snowflake extension fork library, but with the latest changes here, this extension moved into liqubase main library. So we can't use our fork workarounds and hacks no more. But we can try to fix it here.

Unfortunately, I can't come up with a ticket for this issue. I can just see the consequences of this issue while I'm using liqubase as a library in our product. (diffChangeLog command)

I can't see the result of comparing two schemas in Snowflake if I did "USE SCHEMA <schemaName>" for any schema in my catalog earlier. If 

`"SELECT CURRENT_SCHEMA()"`

return not null for the database.

Expected:

2023-01-18 01_22_05-Schema Compare

Actual:

2023-01-18 01_23_03-Schema Compare

So, the problem is in the method `getSchemas`.
In original SchemaSnapshotGenerator#getDatabaseSchemaNames simple `database.getConnection()).getMetaData().getSchemas()` is used. 
But for Snowflake, you have `ResultSet getSchemas(String catalog, String schemaPattern) throws SQLException;` implementation.
This method returns you only one schema in case                 
`getSchemas(database.getDefaultCatalogName(), database.getDefaultSchemaName())`

2023-01-18 01_29_22-DBeaver Ultimate 22 3 2 - _TEST_DB 2_ Script-101

https://github.com/snowflakedb/snowflake-jdbc/blob/8c0a9ce074b6fea8befbe05727387924c53f9122/src/main/java/net/snowflake/client/jdbc/SnowflakeDatabaseMetaData.java#L3001-L3033

So. If you want to specify the catalog name - it is ok. But the schema pattern name is not appropriate here.

### DIFF
--- a/liquibase-snowflake/src/main/java/liquibase/snapshot/jvm/SchemaSnapshotGeneratorSnowflake.java
+++ b/liquibase-snowflake/src/main/java/liquibase/snapshot/jvm/SchemaSnapshotGeneratorSnowflake.java
@@ -27,7 +27,7 @@ public class SchemaSnapshotGeneratorSnowflake extends SchemaSnapshotGenerator {
         List<String> returnList = new ArrayList<>();
 
         try (ResultSet schemas = ((JdbcConnection) database.getConnection()).getMetaData().getSchemas(database
-                .getDefaultCatalogName(), database.getDefaultSchemaName())) {
+                .getDefaultCatalogName(), null)) {
             while (schemas.next()) {
                 returnList.add(JdbcUtil.getValueForColumn(schemas, "TABLE_SCHEM", database));
             }


nvoxland:

Yes, it doesn't seem like we should be passing along the default schema since it doesn't let you get other schemas out. I'll pass it along for the full review. I don't think there should be any downstream impacts from getting all the schemas back from this method. That is how it works on all the other databases.

Questions I have:
- None

Potential risks:
- None

- Nothing

filipelautert (approved):
This one is a tough one, you really need to execute command  "USE SCHEMA" to reproduce it .
Thansk for the fix!

XDelphiGrl:
@LonwoLonwo, hello! Thank you for the detailed bug description! As a QA Engineer, it makes me smile! I hope you have a wonderful day & thanks for the PR!

XDelphiGrl (approved):
This PR fixes an unpleasant bug in the `diff` logic for comparing two Snowflake schemas. 

APPROVED

LonwoLonwo:
@XDelphiGrl 😊

PR #3707, opened 2023-01-18
Map 'double' to SQL type 'DOUBLE PRECISION' for an Oracle database (CORE-3165)

maartenc:
## Impact

## Description

With this change, a 'double' type will map to the SQL type 'DOUBLE PRECISION' for an Oracle database. 
The current mapping to a FLOAT(24) is not good enough, as described in [CORE-3165](https://liquibase.jira.com/browse/CORE-3165).


nvoxland:

Thanks for the PR, @maartenc Glad for the help!

Double precision does seem like the correct choice for "double". It will be a change in behavior for oracle users so we'll need to make sure to document it in the release.

Questions I have:
- None

Potential risks:
- None

- Nothing

filipelautert:
Hello @maartenc ! Thanks for the code.
I executed the project build and it failed at class DoubleTypeTest - do you mind updating this test file to be according to your changes?

maartenc:
Thanks @filipelautert !
I've updated that DoubleTypeTest as well now :-)

filipelautert (approved):
Thanks!

liquibase-core/src/main/java/liquibase/datatype/core/DoubleType.java
```diff
@@ -44,7 +44,7 @@ public DatabaseDataType toDatabaseDataType(Database database) {
         }
 
         if (database instanceof OracleDatabase) {
-            return new DatabaseDataType("FLOAT", 24);
+            return new DatabaseDataType("DOUBLE PRECISION");
```
suryaaki2, line 47: Can we combine all if conditions for various databases to one single if with || ?

suryaaki2 (changes requested):

liquibase-core/src/main/java/liquibase/datatype/core/DoubleType.java
maartenc: Good suggestion, I pushed an extra change for this

XDelphiGrl (approved):
PR address an Oracle bug where `FLOAT` was being incorrectly translated to to `DOUBLE` instead of `DOUBLE PRECISION`. 

APPROVED

PR #3713, opened 2023-01-19
Make fields can be safely made final. All final fields have a value a…

arturobernalg:
## Impact

## Description

Make fields can be safely made final. All final fields have a value and this value does not change.


nvoxland:

The main thing to look out for in making things final is that even if we aren't changing values in our code, we sometimes leave fields changeable for extensions to modify.

However, in my quick glance I only saw private fields so that's not a problem for those. No other code can change them anyway, so the compiler will catch anything in our code trying to.

Questions I have:
- None

Potential risks:
- None

- Nothing

MalloD12 (approved):
**Approved**.

Changes looks ok to me. I had only to fix some sonar issues/code smells. Build and tests have been successfully executed. 

### Things to be aware of:
- None

### Things to worry about:
- None

liquibase-core/src/main/java/liquibase/snapshot/jvm/JdbcSnapshotGenerator.java
```diff
@@ -103,9 +103,6 @@ public void addStatusListener(DiffStatusListener listener) {
     }
 
     protected void updateListeners(String message) {
-        if (this.statusListeners == null) {
```
XDelphiGrl, line 106: @arturobernalg or @MalloD12, hi! Why is this check for null statusListeners no longer needed? In other places where `this.something == null` the code was updated to `this.something.isEmpty()`. I'm just curious why we didn't need to do that here.

filipelautert: Hi - it's because the variable has been initialized at line 27 and now it is final, so nothing can change it's value to null.

XDelphiGrl: Ah! Makes perfect sense. Thank you!

XDelphiGrl (approved):
PR updates many private fields to be `final`. In cases where the previous code had a check for `this.something == null` and `something` was marked `final`, the `== null` check has been removed because now there is no way for `something` to be inadvertently set to `null` (or any other value). 

APPROVED

filipelautert (approved):
That's interesting, there are so many final fields. How many of them could be changed to constants? :)

arturobernalg:
> That's interesting, there are so many final fields. How many of them could be changed to constants? :)

I'll check. i curious as well

suryaaki2 (approved):

PR #3716, opened 2023-01-19
Use constants instead of arrays with known lengths of zero

arturobernalg:
## Impact

## Description

We can save memory reusing the same array instance. Allocations of arrays with known lengths of zero when there is a constant for that in the class of the array's element type. Also, as zero-length arrays are immutable.

Please take a look to this PR 
https://github.com/liquibase/liquibase/pull/3500

### DIFF
--- a/liquibase-core/src/main/java/liquibase/change/core/DropAllForeignKeyConstraintsChange.java
+++ b/liquibase-core/src/main/java/liquibase/change/core/DropAllForeignKeyConstraintsChange.java
@@ -62,7 +62,7 @@ public class DropAllForeignKeyConstraintsChange extends AbstractChange {
             sqlStatements.addAll(Arrays.asList(change.generateStatements(database)));
         }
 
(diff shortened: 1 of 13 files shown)


MalloD12 (approved):
**Approved**.

Return of empty arrays has been improved by using a defined constant in the same way as it was done in PR #3500. 

### Things to be aware of:
- None

### Things to worry about:
- None

XDelphiGrl (approved):
PR replaces logic to define pre-sized arrays with constant zero-size arrays. The constants are defined in the abstract classes (see PR referenced in description). An example of the changes implemented in this PR is:

**Old Style**
```
return sqlStatements.toArray(new SqlStatement[0]);
```

**New Style Using Constant**
```
sqlStatements.toArray(SqlStatement.EMPTY_SQL_STATEMENT);
```

APPROVED

filipelautert (approved):

suryaaki2 (approved):

filipelautert (approved):

PR #3726, opened 2023-01-24
Stop publishing liquibase-extension-testing to Github on branch builds (DAT-13038)

abrackx:
## Impact

## Description
Keeps track of liquibase-extension-testing in cleanup jobs, publishes package on release instead.  

## Things to be aware of
This will currently publish liquibase-core and liquibase-extension-testing on release to github packages.

### DIFF
--- a/.github/workflows/cleanup-branch-builds.yml
+++ b/.github/workflows/cleanup-branch-builds.yml
@@ -9,6 +9,10 @@ jobs:
   delete-package:
     name: Delete Github Package for Branch
     runs-on: ubuntu-20.04
+    matrix:
+      package:
+        - org.liquibase.liquibase-core
+        - org.liquibase.liquibase-extension-testing
     steps:
       # Get version id(s) based on version name
     - uses: castlabs/get-package-version-id-action@v2.1
@@ -22,7 +26,7 @@ jobs:
         # Name of the package.
         # Defaults to an empty string.
         # Required if `package-version-ids` input is not given.
-        package-name: org.liquibase.liquibase-core
+        package-name: ${{ matrix.package }}
 
         # The number of latest versions to keep.
         # This cannot be specified with `num-old-versions-to-delete`. By default, `num-old-versions-to-delete` takes precedence over `min-versions-to-keep`.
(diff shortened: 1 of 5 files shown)


.github/workflows/release-published.yml
```diff
@@ -195,3 +195,31 @@ jobs:
...
+          distribution: 'temurin'
+          cache: 'maven'
+          server-id: github
+
+      - name: Version artifact
+        run: mvn versions:set -DnewVersion="${{ needs.setup.outputs.version }}"
+
+      # Publish to GitHub Packages
+      - name: Publish package to Github
+        run: mvn -B clean deploy -DskipTests=true
```
StevenMassaro, line 222: Is Maven configured to only deploy the modules we care about to GitHub packages? I am probably missing that somewhere.

StevenMassaro (approved):

.github/workflows/cleanup-branch-builds.yml
```diff
@@ -9,6 +9,10 @@ jobs:
   delete-package:
     name: Delete Github Package for Branch
     runs-on: ubuntu-20.04
+    matrix:
+      package:
+        - org.liquibase.liquibase-core
+        - org.liquibase.liquibase-extension-testing
```
jnewton03, line 15: Is deleting branch builds needed if we aren't publishing branch builds?

abrackx: Good point, removed!

.github/workflows/release-published.yml
abrackx: Yeah we have `<maven.deploy.skip>false</maven.deploy.skip>` set in the pom files of all submodules (except core), but in liquibase-extension-testing it is set to `true` in the profile `github-package` when maven detects the env var `TARGET_ARTIFACT_REPOSITORY` is set to `github`. This deploy step is separate from the one that actually pushes to sonatype (which does not set the env var), so in this new step we'll be pushing the core artifact and extension-testing artifact on release only to github packages.

jnewton03 (approved):
LGTM 🚀

filipelautert (approved):

PR #3737, opened 2023-01-27
DAT-12984 Change exception level for changeset error to severe

wwillard7800:


wwillard7800:
Base branch was incorrect.

PR #3747, opened 2023-01-30
Include deploymentId and liquibaseTargetUrl mdc values during exception log (DAT-13012)

abrackx:
## Impact

## Description
Include deploymentId and liquibaseTargetUrl mdc values during exception log.

## Things to be aware of
This change will also include any other mdc values that have not been manually removed from scope if an exception is encountered.

## Things to worry about
See above.


liquibase-core/src/main/java/liquibase/integration/commandline/Main.java
```diff
@@ -409,6 +409,7 @@ public Integer run() throws Exception {
                                 }
                             }
                         });
+                        Scope.getCurrentScope().getMdcManager().clear();
```
abrackx, line 412: @StevenMassaro do you think I need to do this?

StevenMassaro: Is this to remove the keys you're adding in Liquibase.java?

Why did you move this from line 461?

abrackx: Resolved via chat this line is necessary - best practice is to clean up threadlocal variables so in the event that someone is reusing threads with Liquibase they don't encounter any unexpected behavior. The finally block was removed because it cleans up all custom mdc key value pairs prior to logging the exception.

suryaaki2 (approved):

filipelautert (approved):

suryaaki2 (approved):

PR #3752, opened 2023-01-31
Add Snowflake structured data types to unmodifiable Data Types

LonwoLonwo:
Add Snowflake structured data types in the unmodifiable Data Types list to avoid the incorrect results of the diffChangeLog command

## Impact

## Description

Structured data types are compared incorrectly.

Structured data types in the Snowflake can not be created with parameters: https://docs.snowflake.com/en/sql-reference/data-types-semistructured.html

2023-01-31 12_26_53-DBeaver Ultimate 22 3 3 - _TEST_DB_ Script-85

But the result of comparing tables with these types of columns with the diffChangeLog command will be

```sql
CREATE OR REPLACE TABLE TEST_SEMI_STRUCTURED(
	var VARIANT(0),
	arr ARRAY(0),
	obj OBJECT(0));
```

because by default Liquibase returns data type name + parameters

And the simplest way to avoid it is adding data types in the unmodifiable types list:

https://github.com/liquibase/liquibase/blob/master/liquibase-core/src/main/java/liquibase/snapshot/jvm/ColumnSnapshotGenerator.java#L435

Steps:

1. Create a table in schema1:

```sql
CREATE TABLE TEST_SEMI_STRUCTURED(
	var VARIANT,
	arr ARRAY,
	obj OBJECT);
```
2. Compare schema1 with an empty schema.

If I need to add tests - please, give me advice on how to do it better. Maybe like MSSQLDatabaseTest#isUnmodifiable? 
If I need to create a ticket for the issue first - tell me also.

### DIFF
--- a/liquibase-snowflake/src/main/java/liquibase/database/core/SnowflakeDatabase.java
+++ b/liquibase-snowflake/src/main/java/liquibase/database/core/SnowflakeDatabase.java
@@ -24,7 +24,8 @@ public class SnowflakeDatabase extends AbstractJdbcDatabase {
 
     public SnowflakeDatabase() {
         super.setCurrentDateTimeFunction("current_timestamp::timestamp_ntz");
-        super.unmodifiableDataTypes.addAll(Arrays.asList("integer", "bool", "boolean", "int4", "int8", "float4", "float8", "numeric", "bigserial", "serial", "bytea", "timestamptz"));
+        super.unmodifiableDataTypes.addAll(Arrays.asList("integer", "bool", "boolean", "int4", "int8", "float4", "float8", "numeric",
+            "bigserial", "serial", "bytea", "timestamptz", "array", "object", "variant"));
         super.unquotedObjectsAreUppercased = true;
         super.addReservedWords(getDefaultReservedWords());
         super.defaultAutoIncrementStartWith = BigInteger.ONE;


nvoxland:

The change makes sense to me. I appreciate the detail in your PR. 

There is no need for an issue for specific bugfixes like this. It can be nice to track and discuss the feature request (issue) independently from implementation options for that feature (PR) for more complex things, but for "X doesn't work unless we do Y" type bugs it's generally not worth the hasstle. 

I'd tend to think this fix doesn't need a test. Tests should be reserved for "non-trivial code" especially "logic that can have bugs in it", but this basically just a declarative statement with no logic itself. The test would end up really just being "Is java able to correctly handle an array literal" which the answer is yes. We have tests elsewhere which check that "unmodifiable" values are being used correctly. 

What could be useful as a test is more of an overall test that those newly-added datatypes are handled correctly, either through test-harness based tests in the liquibase-snowflake module or in the integration-tests module. But, neither are well set up for that right now on snowflake and I don't think it's worth blocking this PR for getting that going.

Questions I have:
- None

Potential risks:
- None

- Nothing

LonwoLonwo:
Thanks for the detailed response @nvoxland

filipelautert (approved):
Lot's of Snowflake fixes for this release, thanks! How was it working before?

suryaaki2 (approved):

LonwoLonwo:
@filipelautert If we talk about us, we had the Snowflake extension fork with all these fixes :)

XDelphiGrl (approved):
This PR prevents Liquibase from adding invalid datatype parameters to generated changesets for objects containing `array`, `object` and `variant` datatypes.
* Fix impacts Snowflake only.

APPROVED

PR #3782, opened 2023-02-06
add MDC for rollback-one-changeset (DAT-13024)

StevenMassaro:
## Impact


liquibase-core/src/main/java/liquibase/integration/commandline/Main.java
```diff
@@ -394,7 +394,10 @@ public Integer run() throws Exception {
                         Map<String, Object> innerScopeObjects = new HashMap<>();
                         innerScopeObjects.put("defaultsFile", LiquibaseCommandLineConfiguration.DEFAULTS_FILE.getCurrentValue());
                         if (!Main.runningFromNewCli) {
-                            innerScopeObjects.put(Scope.Attr.resourceAccessor.name(), new ClassLoaderResourceAccessor(main.configureClassLoader()));
+                            new CompositeResourceAccessor(
+                                    new DirectoryResourceAccessor(Paths.get(".").toAbsolutePath().toFile()),
+                                    new ClassLoaderResourceAccessor(main.configureClassLoader())
+                            );
```
StevenMassaro, line 400: I am unsure whether this is a safe change to make. It was necessary in order to write an integration test for this change.

filipelautert: I think it's fine, but should not  it be set to "innerScopeObjects.put(Scope.Attr.resourceAccessor.name(),<here>)"  instead of just instantiated?

StevenMassaro: Uh yeah, it should. My test was passing so I didn't even notice this... Fixing now. Thanks @filipelautert!

filipelautert (approved):

PR #3783, opened 2023-02-06
Fix failing EnvironmentValueProviderTest

nvoxland:
## Impact

## Description

The `getProvidedValue doesn't require license` test is looking for a `java_home` environment variable, and so it fails on any development environments where that variable is not set.

I could have tried to find a different environment variable that's consistently available across all OS's and/or used a custom map like the other tests, but because the class under test has no references to LicenseService anymore it seemed like an unnecessary test to fix.

## Things to be aware of

Only removes a sometimes-failing test.

## Things to worry about

Nothing


filipelautert (approved):
Thanks, this error was boring.

XDelphiGrl (approved):
PR removes a test that is unnecessary due to the removal of the licensing requirement to use certain environment variables. 
* Functional and test harness executions passing, although this is irrelevant to the removal of a unit test.

APPROVED

PR #3784, opened 2023-02-06
Created ChangeDefinitionTest

nvoxland:
## Impact

- [X] New Test
 
## Description

Created a new "golden master" change definition test to catch unexpected changes in Change definitions and provide a summary for docs team to look at

## Things to be aware of

Nothing

## Things to worry about

Nothing


MalloD12:
Hi @nvoxland, 

I think these tests look ok to me. I was wondering whether we should keep them alongside the command definition tests we have in a single place (maybe a definition tests module). Do you think that is something feasible or should they be separate as they are now?

Thanks,
Daniel.

nvoxland:
Either is fine. This test doesn't actually need a database to connect to and so didn't want to tie it up into the "integration tests" which is where the command definition tests are (because they do interact with databases). 

But it's a good point that it could be nice to have one parent to look for for both types of "specifications". 

But, even if I moved it to the integration-tests module, they'd be in different sub-directories still since it's commands vs. changes and probably not something people would jump between anyway? So I'd tend to say  just leave them where they are, but if you think it would be easier to maintain I can move them quick

MalloD12 (approved):
**Approved**.

Change definition tests look good to me. 

@filipelautert @suryaaki2 what do you guys think about unifying definition tests into a single module for them. I don't think this change would provides a huge benefit, but I think it would be nice to have each different type of tests centralized in a single place (extension test for a side, integration tests for another one and these ones for another).  Happy to leave them as they are if you think it won't be worth to spend time on something like that.

suryaaki2 (approved):

PR #3789, opened 2023-02-07
Adds contexts, labels, contextFilters and labelFilters to mdc (DAT-12958)

abrackx:
## Impact
‼️ BUILT OFF DAT-12919 AND DAT-13024 ‼️ 

## Description
Also updates updateCount, updateToTag to include expected mdc values.


StevenMassaro (approved):

wwillard7800 (approved):

filipelautert (approved):

liquibase-core/src/main/java/liquibase/Liquibase.java
```diff
@@ -571,13 +574,16 @@ public void update(int changesToApply, Contexts contexts, LabelExpression labelE
             throws LiquibaseException {
         changeLogParameters.setContexts(contexts);
         changeLogParameters.setLabels(labelExpression);
+        Scope.getCurrentScope().addMdcValue(MdcKey.CHANGESET_LABEL_FILTER, labelExpression != null ? labelExpression.toString(): "");
```
suryaaki2, line 577: I see a duplicate of code -Can we move this two lines to a method and call that method instead?

suryaaki2 (changes requested):

liquibase-core/src/main/java/liquibase/Liquibase.java
abrackx: Updated!

PR #3797, opened 2023-02-08
Improve use of generics in code

arturobernalg:
## Impact

## Description

<!--
Avoid generic classes with omitted type parameters.  Such raw use of generic types is valid in Java, but it defeats the purpose of type parameters and may mask bugs.


nvoxland:

Thanks! More generics are always good. We will need to make sure to watch the API compatibility in the full review.

I don't remember the details on what impact this can have on compatibity, though. To some degree, java does generic type erasure so maybe it doesn't survive compilation to cause a problem? But it also seems like it maybe does? 

If it does get saved enough to impact people using liquibase as a library, we'll need to make sure the types are correct and not too narrow or to wide. It looked like it was mainly on variables and not methods which are also probably less API-impacting and so less of a worry?

MalloD12 (approved):
**Approved**.

Based on my knowledge and understanding about generics this looks ok to me, but I would appreciate if you @filipelautert can also look at this quickly to this. We have already discuss about some points (same with @nvoxland). I haven't notice any API compatibility issue/risk, but I think a second quick look would be nice. 

Again, thank you @arturobernalg for another contribution.

### Things to be aware of:
- None

### Things to worry about:
- None

arturobernalg:
> **Approved**.
> 
> ## Review and testing results:
> Based on my knowledge and understanding about generics this looks ok to me, but I would appreciate if you @filipelautert can also look at this quickly to this. We have already discuss about some points (same with @nvoxland). I haven't notice any API compatibility issue/risk, but I think a second quick look would be nice.
> 
> Again, thank you @arturobernalg for another contribution.
> 
> ### Things to be aware of:
> * None
> 
> ### Things to worry about:
> * None

In Java, both new ArrayList() and new ArrayList<>() create a new instance of the ArrayList class. However, there is a subtle difference between the two.

new ArrayList() is known as the raw type, while new ArrayList<>() is known as the diamond operator. The diamond operator was introduced in Java 7 to make it easier to declare generic types.

When using the diamond operator, the compiler can infer the type of the ArrayList based on the type of the variable it is assigned to. This can make code easier to read and less error-prone.

Additionally, using the diamond operator can help prevent type mismatch errors at compile time, as the compiler can ensure that the type of the ArrayList matches the type of the variable it is being assigned to.

Therefore, the benefit of changing new ArrayList() to new ArrayList<>() is that it makes the code more concise, easier to read, and less prone to type mismatch errors.

filipelautert (approved):

suryaaki2 (approved):

XDelphiGrl (approved):
This code safety PR minimizes compile-time type mismatch errors.
* Code has been reviewed by several Java developers to avoid API breaking changes.
  * There should be no functional changes for Liquibase consumers.
* Test harness failures for hsqldb and firebird are environmental (Titan failed to start). These failures are not caused by or related to the changes in this PR and can be disregarded.
* Functional tests have failures due to an issue with how the Commercial jars are loaded; there is a test PR to address this. No need to block merge of this PR due to these failures.

APPROVED

filipelautert:
@MalloD12 can you fix the conflicts for this pr so we can merge it? thanks!

PR #3804, opened 2023-02-10
Enhance empty string check.

Jeevananthammoorthy:
## Impact


MalloD12:
Hi @Jeevananthammoorthy thank you for creating this PR for us. The updated class is not the only place where we have this type of occurrence of this type of validation. Looking quickly through the code I have seen there several other places where we would need to update this as well. Would you mind looking for these classes and performing the same change over there?

Thanks,
Daniel.

Jeevananthammoorthy:
> Hi @Jeevananthammoorthy thank you for creating this PR for us. The updated class is not the only place where we have this type of occurrence of this type of validation. Looking quickly through the code I have seen there several other places where we would need to update this as well. Would you mind looking for these classes and performing the same change over there?
> 
> Thanks, Daniel.

Got it, will push the updated patch

Jeevananthammoorthy:
@MalloD12  QQ: Is it okay to create new PR with the requested review comments? Am using githubDesktop, and I couldn't find away to update this commit.
Lmk if any issue, in raising new PR. the changes are ready in my local.

Jeevananthammoorthy:
@MalloD12  Sure, will push update patch in sometime including null check before isEmpty().

Thanks,
Jeevanantham

MalloD12:
> @MalloD12 Sure, will push update patch in sometime including null check before isEmpty().
> 
> Thanks, Jeevanantham

I think it would depend on a case basis, maybe not all the cases need this null check. 

Thanks,
Daniel.

Jeevananthammoorthy:
@MalloD12  Updated the patch, kindly review.

Thanks,
Jeevanantham

MalloD12:
> @MalloD12 Updated the patch, kindly review.
> 
> Thanks, Jeevanantham

I have some unit tests failing, and I still need to confirm whether there is any relationship with the changes implemented in this PR or not. I might need to spend a bit more time on this. Also, I noticed today this PR has been created directly from your master branch, we should be creating a dedicated branch with a representative name of the changes we are trying to implement here. I'll keep you posted once I can determine the cause of these failing unit tests.

Thanks,
Daniel.

MalloD12:
This PR is a minor extension from https://github.com/liquibase/liquibase/pull/3590 where we discussed with the team adding String `isEmpty()` check, instead of doing something like `"".equals(<stringToCompareAgainst>)`. After we added some changes here the build of this PR started having some issues which might require a little bit more time to investigate the root cause of it, whether it's directly related to changes made here or if it's something else.

Move it back to "Foundation" queue for further prioritization, or we can also close it if we consider it is not worth expending too much time on it for the extra value it will provide us.

cc: @mariochampion

liquibase-standard/src/main/java/liquibase/parser/core/xml/XMLChangeLogSAXHandler.java
```diff
@@ -100,7 +100,8 @@ public void endElement(String uri, String localName, String qName) throws SAXExc
         ParsedNode node = nodeStack.pop();
         try {
             String seenText = this.textStack.pop().toString();
-            if (!"".equals(StringUtil.trimToEmpty(seenText))) {
+            String trimmedString = StringUtil.trimToEmpty(seenText);
+            if (!StringUtil.isEmpty(trimmedString)) {
```
filipelautert, line 104: Hello @Jeevananthammoorthy ! Can't we use the new "StringUtil.isBlank" here and in some other places instead of trimToEmpty then isEmpty?

filipelautert (approved):

filipelautert (changes requested):
Please see my comments in the code.

filipelautert:
@kevin-atx - it wasn't merged. The contributor removed their repository and with that the PR was closed.

kevin-atx:
that makes sense - thanks!

PR #3805, opened 2023-02-10
DAT-13145: set log level to severe if using a log file with no log level set

StevenMassaro:
## Impact

### DIFF
--- a/liquibase-cli/src/main/java/liquibase/integration/commandline/LiquibaseCommandLine.java
+++ b/liquibase-cli/src/main/java/liquibase/integration/commandline/LiquibaseCommandLine.java
@@ -630,7 +630,7 @@ public class LiquibaseCommandLine {
     }
 
     protected Map<String, Object> configureLogging() throws IOException {
-        Map<String, Object> returnMap = new HashMap<>();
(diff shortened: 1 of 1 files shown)


liquibase-cli/src/main/java/liquibase/integration/commandline/LiquibaseCommandLine.java
```diff
@@ -664,6 +664,8 @@ private void configureLogging(Level logLevel, String logFile) throws IOException
 
         java.util.logging.Logger rootLogger = java.util.logging.Logger.getLogger("");
         Level cliLogLevel = logLevel;
+        // If a log file is set, but no log level is set, this variable represents the log level that should be used.
```
wwillard7800, line 667: Do we also need to check `logFormat` here?

StevenMassaro: Nope, this applies to all log formats. This was existing functionality that has gradually broken over the years.

wwillard7800 (approved):

filipelautert (approved):

PR #3807, opened 2023-02-10
safely set code point limit for SnakeYaml (DAT-13189)

StevenMassaro:
## Impact

## Description

Previously, Liquibase increased the code point limit for SnakeYaml so that particularly large snapshots could be parsed. This made use of a recently added method in the SnakeYaml codebase, so users of Liquibase who are using an old version of SnakeYaml with a newer version of Liquibase would have their execution interrupted by an exception about a missing method. This fix should catch that exception so that execution is not interrupted.


filipelautert (approved):
Nice

wwillard7800 (approved):

PR #3810, opened 2023-02-13
Adds databaseChangelogSQL mdc key (DAT-13092)

abrackx:
## Impact

## Description
Adds databaseChangelogSQL mdc key, populates via StandardChangelogHistoryService and StandardLockService. Moves changesetSql mdc population into Changeset class. 

## Things to be aware of
‼️ This makes it so that end delimiters (ex: `;`) are included in the changesetSql mdc value. ‼️ I can control this, but because I'm now handling the sql at a level above statement execution it made the most sense to take this route to me. We're not always, but sometimes working with more than one statement per change.

### DIFF
--- a/liquibase-core/src/main/java/liquibase/changelog/ChangeSet.java
+++ b/liquibase-core/src/main/java/liquibase/changelog/ChangeSet.java
@@ -1493,10 +1493,12 @@ public class ChangeSet implements Conditional, ChangeLogChild {
     }
 
     private void addSqlMdc(Change change, Database database, boolean generateRollbackStatements) throws RollbackImpossibleException {
-        SqlStatement[] statements = generateRollbackStatements ? change.generateRollbackStatements(database) : change.generateStatements(database);
-        String sqlStatementsMdc = Arrays.stream(statements)
-                .map(statement -> SqlUtil.getSqlString(statement, SqlGeneratorFactory.getInstance(), database))
-                .collect(Collectors.joining("\n"));
-        Scope.getCurrentScope().addMdcValue(MdcKey.CHANGESET_SQL, sqlStatementsMdc);
+        if (change.supports(database)) {
+            SqlStatement[] statements = generateRollbackStatements ? change.generateRollbackStatements(database) : change.generateStatements(database);
+            String sqlStatementsMdc = Arrays.stream(statements)
+                    .map(statement -> SqlUtil.getSqlString(statement, SqlGeneratorFactory.getInstance(), database))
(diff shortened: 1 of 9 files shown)


liquibase-core/src/main/java/liquibase/changelog/ChangeSet.java
```diff
@@ -1484,4 +1491,14 @@ private void addChangeSetMdcProperties() {
         Scope.getCurrentScope().addMdcValue(MdcKey.CHANGESET_LABEL, labelMdc);
         Scope.getCurrentScope().addMdcValue(MdcKey.CHANGESET_CONTEXT, contextsMdc);
     }
+
+    private void addSqlMdc(Change change, Database database, boolean generateRollbackStatements) throws RollbackImpossibleException {
+        if (change.supports(database)) {
```
suryaaki2, line 1496: What would happen if change doesn't support MDC? Can you also add a javadoc comment to the method?

suryaaki2 (changes requested):

liquibase-core/src/main/java/liquibase/changelog/ChangeSet.java
abrackx: If the change doesn't support the database then we do not add it to the mdc, as we're going to skip it in execution.

suryaaki2 (approved):

filipelautert (approved):
Nice, getJdbcExecutor helps to make sure that we are always logging to MDC.

liquibase-core/src/main/java/liquibase/util/SqlUtil.java
```diff
@@ -341,4 +346,22 @@ public static String replacePredicatePlaceholders(Database database, String pred
...
+     * @param sqlGeneratorFactory the SqlGeneratorFactory instance to use to generate the sql
+     * @param database the database to generate sql against
+     * @return the sql string or an empty string if there are no statements to generate
+     */
+    public static String getSqlString(SqlStatement statement, SqlGeneratorFactory sqlGeneratorFactory, Database database) {
+        Sql[] sqlStatements = sqlGeneratorFactory.generateSql(statement, database);
+        if (sqlStatements != null) {
+            return Arrays.stream(sqlStatements)
+                    .map(sql -> sql.toSql() + sql.getEndDelimiter())
+                    .collect(Collectors.joining("\n"));
```
StevenMassaro, line 362: Should this use System.lineseparator? Maybe it is best to be consistent with the line ending.

StevenMassaro (approved):

PR #3811, opened 2023-02-13
PD-3151: Updating change type documentation

adrian-velonis1:
## Impact

## Description

Updated descriptions of core change types to match online documentation.

https://datical.atlassian.net/browse/PD-3151


liquibase-core/src/test/groovy/liquibase/configuration/core/EnvironmentValueProviderTest.groovy
```diff
@@ -1,87 +1,14 @@
...
-import liquibase.license.LicenseInstallResult
-import liquibase.license.LicenseService
-import liquibase.license.LicenseServiceFactory
-import liquibase.license.Location
 import spock.lang.Specification
 import spock.lang.Unroll
 
 class EnvironmentValueProviderTest extends Specification {
-    static boolean licenseValid = false;
-    static final mockLicenseService = new LicenseService() {
```
nvoxland, line 15: This is removing unnecessary tests which fail on environments without a "java.home" environment variable (most people).

nvoxland (approved):

liquibase-core/src/main/java/liquibase/change/core/AbstractModifyDataChange.java
```diff
@@ -59,7 +60,7 @@ public void setWhere(String where) {
     /**
      * @deprecated use getWhere().
      */
-    @DatabaseChangeProperty(isChangeProperty = false)
+    @DatabaseChangeProperty(isChangeProperty = false, description = "Deprecated. Use 'where'")
```
MalloD12, line 63: For this kind of deprecated method for which we already have a newer version. Don't we need to use `@Deprecated` annotation? @nvoxland (cc: @filipelautert )

liquibase-core/src/main/java/liquibase/change/core/AddAutoIncrementChange.java
```diff
@@ -51,7 +56,7 @@ public void setCatalogName(String catalogName) {
         this.catalogName = catalogName;
     }
 
-    @DatabaseChangeProperty(mustEqualExisting = "column.relation.schema")
+    @DatabaseChangeProperty(mustEqualExisting = "column.relation.schema", description = "Name of the database schema")
```
MalloD12, line 59: Hey Adrian, I have another question for you. For example for this method, we are adding a description, but checking the published [documentation](https://docs.liquibase.com/change-types/add-auto-increment.html?Highlight=AddAuto) we already have the same description for it. What’s the reason for this?

```
@DatabaseChangeProperty(since = "3.0", mustEqualExisting = "column.relation.catalog", description = "Name of the database catalog")
public String getCatalogName() {
    return catalogName;
}
```

liquibase-core/src/main/java/liquibase/change/core/AddAutoIncrementChange.java
```diff
@@ -60,7 +65,7 @@ public void setSchemaName(String schemaName) {
         this.schemaName = schemaName;
     }
 
-    @DatabaseChangeProperty(mustEqualExisting = "column.relation")
```
MalloD12, line 63: @nvoxland (cc: @adrian-velonis1 - @filipelautert  ) Another question I have is about "mustEqualExisting" property. It says it adds the specified column into changeParameterMetaData, what is it used for? Is it also reflected in the change documentation page as well?

liquibase-core/src/main/java/liquibase/change/core/AddAutoIncrementChange.java
XDelphiGrl: I am not entirely sure, but my guess is that @adrian-velonis1 did this so that what is already in Flare (documentation) matches what is in the code. This way when they autogenerate the documentation, we aren't missing something that was manually added to Flare. 

Adrian, am I understanding this correctly?

liquibase-core/src/main/java/liquibase/change/core/AddAutoIncrementChange.java
```diff
@@ -186,11 +195,13 @@ public ChangeStatus checkStatus(Database database) {
...
             if (getGenerationType() != null && column.getAutoIncrementInformation().getGenerationType() != null) {
-                result.assertCorrect(getGenerationType().equals(column.getAutoIncrementInformation().getGenerationType()), "Generation type is incorrect");
+                result.assertCorrect(getGenerationType().equals(column.getAutoIncrementInformation().getGenerationType()),
+                     "Generation type is incorrect");
             }
 
             if (getDefaultOnNull() != null && column.getAutoIncrementInformation().getDefaultOnNull() != null) {
-                result.assertCorrect(getDefaultOnNull().equals(column.getAutoIncrementInformation().getDefaultOnNull()), "Default on null is incorrect");
+                result.assertCorrect(getDefaultOnNull().equals(column.getAutoIncrementInformation().getDefaultOnNull()),
+                     "Default on null is incorrect");
```
XDelphiGrl, line 204: @adrian-velonis1, hi! Is the extra spacing here required for autogenerating the documentation? I'm curious because it is an unusual way to see the code formatted.

liquibase-core/src/main/java/liquibase/change/core/AddForeignKeyConstraintChange.java
```diff
@@ -369,6 +374,7 @@ public String getConfirmationMessage() {
      * @deprecated No longer supported in 3.0
      */
     @Deprecated
+    @DatabaseChangeProperty(description = "Deprecated")
```
XDelphiGrl, line 377: This is not related to the documentation, but we should probably remove code that's been deprecated for this long.

XDelphiGrl: Actually, and this is related to the documentation, @adrian-velonis1 .  If we do not remove this code, what should the description say to use instead? For example, just above we have a deprecation with this information: 

`@deprecated Use {@link #getOnDelete()}`

liquibase-core/src/main/java/liquibase/change/core/AddLookupTableChange.java
```diff
@@ -68,7 +68,7 @@ public void setExistingTableCatalogName(String existingTableCatalogName) {
         this.existingTableCatalogName = existingTableCatalogName;
     }
 
-    @DatabaseChangeProperty(mustEqualExisting ="column.relation.schema")
+    @DatabaseChangeProperty(mustEqualExisting ="column.relation.schema", description = "Name of the database schema of the existing table")
```
XDelphiGrl, line 71: A more complete description is, "Name of the database schema where the table containing data to extract resides."

liquibase-core/src/main/java/liquibase/change/core/AddLookupTableChange.java
```diff
@@ -95,8 +97,7 @@ public void setExistingColumnName(String existingColumnName) {
         this.existingColumnName = existingColumnName;
     }
 
-
-    @DatabaseChangeProperty(since = "3.0")
+    @DatabaseChangeProperty(since = "3.0", description = "Name of the database catalog of the table to create")
```
XDelphiGrl, line 100: Suggest that we include the word "lookup" in this description and others like it, i.e., "Name of the database catalog to create the lookup table."

liquibase-core/src/main/java/liquibase/change/core/AddLookupTableChange.java
```diff
@@ -105,6 +106,7 @@ public void setNewTableCatalogName(String newTableCatalogName) {
         this.newTableCatalogName = newTableCatalogName;
     }
 
+    @DatabaseChangeProperty(description = "Name of the database schema of the table to create")
```
XDelphiGrl, line 109: And similarly , "Name of the database schema for the lookup table to create."

liquibase-core/src/test/java/liquibase/verify/saved_state/compareGeneratedSqlWithExpectedSqlForMinimalChangesets/createView/asany.sql
```diff
@@ -1,4 +1,4 @@
 -- Database: asany
```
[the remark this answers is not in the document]
XDelphiGrl: Makes sense - thanks, @nvoxland !

liquibase-core/src/test/java/liquibase/verify/saved_state/compareGeneratedSqlWithExpectedSqlForMinimalChangesets/createView/db2i.sql
```diff
@@ -1,4 +0,0 @@
--- Database: db2i
```
XDelphiGrl, line 1: Why was this test deleted? Is it invalid, somehow? Same question for the MaxDB test.

liquibase-core/src/main/java/liquibase/change/core/CreateProcedureChange.java
```diff
@@ -132,7 +121,10 @@ public void setProcedureBody(String procedureText) {
         this.procedureText = procedureText;
     }
 
-    @DatabaseChangeProperty(isChangeProperty = false)
+    @DatabaseChangeProperty(
+        description = "The SQL creating the procedure. You need to define either this attribute or 'path'. " +
```
XDelphiGrl, line 125: It is true that the XML changelog does not have an attribute named `procedureText`. The `procedureText` is implicitly defined as "whatever appears in between `<createProcedure procedureName="name">` and `</createProcedure>`".  For example: 

```
   <changeSet author="lbuser" id="3-proc" labels="community" runOnChange="true" contextFilter="ctx">
        <createProcedure procedureName="simple_proc3">
			ANYTHING HERE IS THE procedureText
		</createProcedure>
    </changeSet>
```

I'm guessing that this implicit definition is what is meant by "nested" but that phrasing confused me. Perhaps, "inline" is better than "nested?"

liquibase-core/src/main/java/liquibase/change/core/CreateProcedureChange.java
```diff
@@ -90,8 +84,7 @@ public void setEncoding(String encoding) {
     }
 
     @DatabaseChangeProperty(
-        description = "File containing the procedure text. Either this attribute or a nested procedure text is " +
-            "required.",
+        description = "File containing the procedure text. Either this attribute or a nested procedure text is required.",
```
XDelphiGrl, line 87: I'm reviewing this file from bottom to top so here's that word "nested" again. I'd say "inline" here, too. Or maybe, "Either this attribute or inline SQL within the `createProcedure` definition is required."

liquibase-core/src/main/java/liquibase/change/core/CreateProcedureChange.java
```diff
@@ -22,16 +22,7 @@
 import java.io.UnsupportedEncodingException;
 import java.util.Map;
 
-@DatabaseChange(
-    name = "createProcedure",
```
XDelphiGrl, line 26: Just verifying ... we do **not** want to keep this in the documentation? I think removing it makes sense but I want to double-check it was intentional.

liquibase-core/src/main/java/liquibase/change/core/AddLookupTableChange.java
MalloD12: Yeah, I like more your proposed description. I can make this update if @adrian-velonis1 is happy with it.

liquibase-core/src/main/java/liquibase/change/core/AddForeignKeyConstraintChange.java
MalloD12: It was my mistake to not checked that the deprecated method was not being used. I can remove it if you both are ok with it.

liquibase-core/src/main/java/liquibase/change/core/CreateProcedureChange.java
TymofiiKritsak: INFO: [DAT-12813 PRO :: "procedureText" attribute doesn't work in JSON and YML changelogs](https://datical.atlassian.net/browse/DAT-12813)

liquibase-core/src/main/java/liquibase/change/core/CreateSequenceChange.java
```diff
@@ -89,7 +90,7 @@ public void setMinValue(BigInteger minValue) {
         this.minValue = minValue;
     }
 
-    @DatabaseChangeProperty(description = "Is the sequence generated in the requested order?")
```
XDelphiGrl, line 92: I appreciate that you removed the `?` from these. I understand the intent of the original description but it also kind of reads like, "hmm, will it or won't it? We'll let you try and find out!" Removing the `?` provides a more definitive statement.

liquibase-core/src/main/java/liquibase/change/core/CreateTableChange.java
```diff
@@ -1,6 +1,19 @@
 package liquibase.change.core;
 
-import liquibase.change.*;
+import static liquibase.statement.SqlStatement.EMPTY_SQL_STATEMENT;
+
+import java.util.ArrayList;
+import java.util.List;
+
+import liquibase.change.AbstractChange;
```
XDelphiGrl, line 8: I've wondered this in many of the classes I've reviewed, but does the documentation require expansion of imports down to the exact class name? The `package.name.*` is a more common annotation when most or all of a package is being imported.

liquibase-core/src/main/java/liquibase/change/core/CreateViewChange.java
```diff
@@ -127,6 +133,7 @@ public void setEncoding(String encoding) {
         this.encoding = encoding;
     }
 
+    @DatabaseChangeProperty(description = "A brief descriptive comment")
```
XDelphiGrl, line 136: "A brief descriptive comment stored in the view metadata" (see similar comment on CreateTableChange).

liquibase-core/src/main/java/liquibase/change/core/DropAllForeignKeyConstraintsChange.java
```diff
@@ -16,16 +22,14 @@
 import liquibase.structure.core.ForeignKey;
 import liquibase.structure.core.Table;
 
-import java.util.*;
-
-@DatabaseChange(name="dropAllForeignKeyConstraints", description = "Drops all foreign key constraints for a table", priority = ChangeMetaData.PRIORITY_DEFAULT, appliesTo = "table")
+@DatabaseChange(name = "dropAllForeignKeyConstraints", description = "Drops all foreign key constraints for a table", priority = ChangeMetaData.PRIORITY_DEFAULT, appliesTo = "table")
```
XDelphiGrl, line 25: Foreign keys are defined _on_ a table so this would be better as "Drops all foreign key constraints on a table".

liquibase-core/src/main/java/liquibase/change/core/DropColumnChange.java
```diff
@@ -258,7 +251,8 @@ public void addColumn(ColumnConfig column) {
     }
     
     @Override
-    @DatabaseChangeProperty(description = "Columns to be dropped if dropping multiple columns. If this is populated, the columnName attribute is ignored.", requiredForDatabase = "none")
+    @DatabaseChangeProperty(
+        description = "Columns to be dropped if dropping multiple columns. If this is populated, the columnName attribute is ignored.", requiredForDatabase = "none")
```
XDelphiGrl, line 255: I'm not sure how you want to document this, but `columns` is not a valid attribute for XML changelogs. When I run this changeset:

```
	<changeSet author="liquibase-user" id="1::dropColumns" labels="community">
        <dropColumn tableName="testtable1">
			<columns>
				<column name="supplier_id"/>
			</columns> 
        </dropColumn>
    </changeSet>
```

I get this validation error:

```
Unexpected error running Liquibase: Error parsing line 21 column 49 of communityOnly.xml: cvc-complex-type.3.2.2: Attribute 'columns' is not allowed to appear in element 'column'.
```

I'm also not sure what `columnName` attribute refers to; there's `name` but no `columnName` Additionally, if we ignore the `name` then ... we would not know which column to drop. I would remove this statement from the documentation.

liquibase-core/src/main/java/liquibase/change/core/DropForeignKeyConstraintChange.java
```diff
@@ -10,14 +10,15 @@
 /**
  * Drops an existing foreign key constraint.
  */
-@DatabaseChange(name="dropForeignKeyConstraint", description = "Drops an existing foreign key", priority = ChangeMetaData.PRIORITY_DEFAULT, appliesTo = "foreignKey")
+@DatabaseChange(name = "dropForeignKeyConstraint", description = "Drops an existing foreign key", priority = ChangeMetaData.PRIORITY_DEFAULT, appliesTo = "foreignKey")
```
XDelphiGrl, line 13: The word "existing" is not required; it's implied that the FK exists if you are dropping it.

liquibase-core/src/main/java/liquibase/change/core/DropPrimaryKeyChange.java
```diff
@@ -16,7 +16,8 @@
 /**
  * Removes an existing primary key.
  */
-@DatabaseChange(name="dropPrimaryKey", description = "Drops an existing primary key", priority = ChangeMetaData.PRIORITY_DEFAULT, appliesTo = "primaryKey")
+@DatabaseChange(name = "dropPrimaryKey", description = "Drops an existing primary key", priority = ChangeMetaData.PRIORITY_DEFAULT,
```
XDelphiGrl, line 19: Same here; the word "existing" is not necessary.

liquibase-core/src/main/java/liquibase/change/core/DropViewChange.java
```diff
@@ -46,8 +51,8 @@ public void setViewName(String viewName) {
         this.viewName = viewName;
     }
 
-    @DatabaseChangeProperty(supportsDatabase = ALL,
-            description = "Option to only drop the view, if it exists, and not failing, if it's not exists")
+    @DatabaseChangeProperty(since = "4.19.0", supportsDatabase = ALL,
+        description = "Appends IF EXISTS to the DROP VIEW statement. If ifExists=true, the view is only dropped if it already exists, but the migration continues even if the view does not exist. If ifExists=false and the view does not exist, the database returns an error. Default: false.")
```
XDelphiGrl, line 55: Nicely worded and way more descriptive than the previous description!

liquibase-core/src/main/java/liquibase/change/core/LoadDataChange.java
```diff
@@ -46,28 +46,7 @@
 import static liquibase.change.ChangeParameterMetaData.ALL;
 
 @DatabaseChange(name = "loadData",
-        description = "Loads data from a CSV file into an existing table. A value of NULL in a cell will be " +
```
XDelphiGrl, line 49: This information should be available somewhere. I agree it doesn't need to be in the JavaDocs but this is useful and valuable additional documentation for LoadDataChange.

liquibase-core/src/main/java/liquibase/change/core/RawSQLChange.java
```diff
@@ -10,29 +10,11 @@
 import liquibase.util.StringUtil;
 
 /**
- * Allows execution of arbitrary SQL.  This change can be used when existing changes are either don't exist,
+ * Allows execution of arbitrary SQL. This change can be used when existing changes are either don't exist,
  * are not flexible enough, or buggy. 
  */
-@DatabaseChange(name="sql",
```
XDelphiGrl, line 16: This is another chunk of information that should be available, although I agree it should not be in the JavaDocs directly. Should I enter PD tickets for these sorts of things, @adrian-velonis1 ?

liquibase-core/src/main/java/liquibase/change/core/RawSQLChange.java
```diff
@@ -10,29 +10,11 @@
 import liquibase.util.StringUtil;
 
 /**
- * Allows execution of arbitrary SQL.  This change can be used when existing changes are either don't exist,
+ * Allows execution of arbitrary SQL. This change can be used when existing changes are either don't exist,
```
XDelphiGrl, line 13: "This change can be used when existing changetypes don't exist or are not flexible enough." I'd drop the "or buggy," too.

liquibase-core/src/main/java/liquibase/change/core/RawSQLChange.java
```diff
@@ -10,29 +10,11 @@
...
-            "\n" +
-            "The sql change can also contain comments of either of the following formats:\n" +
-            "\n" +
-            "A multiline comment that starts with /* and ends with */.\n" +
-            "A single line comment starting with <space>--<space> and finishing at the end of the line.\n" +
-            "Note: By default it will attempt to split statements on a ';' or 'go' at the end of lines. Because of " +
-            "this, if you have a comment or some other non-statement ending ';' or 'go', don't have it at the end of a " +
-            "line or you will get invalid SQL.",
+@DatabaseChange(name = "sql",
+        description = "Allows you to directly specify whatever SQL you want.",
```
XDelphiGrl, line 17: I'd change this to, "Allows you to directly specify raw SQL to execute against the database."

liquibase-core/src/main/java/liquibase/change/core/RawSQLChange.java
```diff
@@ -53,7 +35,8 @@ public String getSql() {
         return super.getSql();
     }
 
-    @DatabaseChangeProperty(serializationType = SerializationType.NESTED_OBJECT, exampleValue = "What about Bob?")
+    @DatabaseChangeProperty(serializationType = SerializationType.NESTED_OBJECT, description = "A brief descriptive comment",
```
XDelphiGrl, line 38: @nvoxland, is this comment written to the database? If so, Adrian, we should update this description to say so.

liquibase-core/src/main/java/liquibase/change/core/CreateProcedureChange.java
MalloD12: I think we should have a similar PD ticket for this because only `procedureText` should be a valid attribute for `CreateProcedureChange`. Both get/set of `procedureBody` are deprecated fields (we should probably remove them from the code as well), so YAML documentation should be updated to reflect this, as well as the code if somehow it is still accepting it. Thank you for pointing this out, @TymofiiKritsak!

liquibase-core/src/main/java/liquibase/change/core/RenameColumnChange.java
```diff
@@ -88,7 +88,7 @@ public void setColumnDataType(String columnDataType) {
         this.columnDataType = columnDataType;
     }
 
-    @DatabaseChangeProperty(description = "Remarks of the column")
+    @DatabaseChangeProperty(description = "A brief descriptive comment")
```
XDelphiGrl, line 91: This should be "A brief descriptive comment written to the column metadata."

liquibase-core/src/main/java/liquibase/change/core/RenameSequenceChange.java
```diff
@@ -41,7 +42,7 @@ public void setSchemaName(String schemaName) {
         this.schemaName = schemaName;
     }
 
-    @DatabaseChangeProperty(mustEqualExisting = "sequence", description = "Name of the sequence to rename")
+    @DatabaseChangeProperty(mustEqualExisting = "sequence", description = "Name of the existing sequence to rename")
```
XDelphiGrl, line 45: :) We may have a different philosophical opinion on if "existing" is necessary or not.

liquibase-core/src/main/java/liquibase/change/core/SQLFileChange.java
```diff
@@ -20,29 +20,9 @@
 
 /**
  * Represents a Change for custom SQL stored in a File.
- * <p/>
  */
 @DatabaseChange(name = "sqlFile",
-        description = "The 'sqlFile' tag allows you to specify any sql statements and have it stored external in a " +
```
XDelphiGrl, line 26: @XDelphiGrl - follow up on this.

liquibase-core/src/main/java/liquibase/change/core/SetColumnRemarksChange.java
```diff
@@ -67,6 +71,7 @@ public void setColumnName(String columnName) {
         this.columnName = columnName;
     }
 
+    @DatabaseChangeProperty(description = "A brief descriptive comment")
```
XDelphiGrl, line 74: A brief descriptive comment written to the column metadata.

liquibase-core/src/main/java/liquibase/change/core/SetTableRemarksChange.java
```diff
@@ -55,6 +56,7 @@ public void setTableName(String tableName) {
         this.tableName = tableName;
     }
 
+    @DatabaseChangeProperty(description = "A brief descriptive comment")
```
XDelphiGrl, line 59: A brief descriptive comment stored in the table metadata.

liquibase-core/src/main/java/liquibase/change/core/RawSQLChange.java
MalloD12: I'm 92.5%  😁  sure most of the inline comments in a raw SQL are not stored in the DB.

liquibase-core/src/main/java/liquibase/change/core/TagDatabaseChange.java
```diff
@@ -9,7 +9,8 @@
 import liquibase.statement.SqlStatement;
 import liquibase.statement.core.MarkChangeSetRanStatement;
 
-@DatabaseChange(name="tagDatabase", description = "Applies a tag to the database for future rollback", priority = ChangeMetaData.PRIORITY_DEFAULT, since = "1.6")
+@DatabaseChange(name = "tagDatabase", description = "Applies a tag to the database for future rollback",
```
XDelphiGrl, line 12: Applies a tag to the database to specify where to stop a rollback.

XDelphiGrl (changes requested):
This PR ensures that the change type JavaDocs are accurate. This allows the Liquibase Documentation team to rely on autogenerated documentation instead of a combination of autogenerated documentation and manual edits on the website. 

@adrian-velonis1, this is an astounding amount of work! Thank you! I do have some comments with suggested changes. If you have any questions, please let me know.

CHANGES REQUESTED

liquibase-core/src/main/java/liquibase/change/core/AddAutoIncrementChange.java
adrian-velonis1: This one has always been mysterious to me, but my understanding is that `mustEqualExisting` tells the application how the value of the attribute _ought_ to be applied when applying the change to an existing high-level object in the database—there's potentially room for error. So in `AddAutoIncrementChange`, the value returned by `getTableName()` corresponds to `column.relation` since you're adding the increment on a particular column in a particular table. I believe this is supposed to be Liquibase's way of providing a generic "path" for modifying the object in whichever database you're using.

Nathan might be able to explain in more detail and/or correct me if I've misunderstood.

As far as I know, `mustEqualExisting` is only needed on attributes which:

1. Interact with objects, so we also don't use it for a boolean/whatever that defines some kind of general behavior
2. Interact with objects that, a high level, realistically exist on most/all databases (like tables)
3. Interact with _existing_ objects, so we do use it for `getTableName()` on `AddColumnChange`, but we don't use it for `getTableName()` on `CreateTableChange`

We probably wouldn't need to put this metadata on docs.liquibase.com, but it would be appropriate on Javadocs or contribute.liquibase.com. I don't have a strong opinion either way.

liquibase-core/src/main/java/liquibase/change/core/AddAutoIncrementChange.java
adrian-velonis1: @XDelphiGrl The indenation isn't technically required, I just wanted to avoid having to do a lot of left/right scrolling to edit the change type descriptions. Normally I would put `description = "blah"` on a new line only if it was getting close to 120ish characters.

This particular one for `getDefaultOnNull()` does look a little strange and was possibly an accident. At least I don't think this particular string would be something we'd grab for the docs.

liquibase-core/src/main/java/liquibase/change/core/AddForeignKeyConstraintChange.java
adrian-velonis1: According to this [old post](https://forum.liquibase.org/t/addforeignkeyconstraintchange-referencesuniquecolumn-support-in-3-0-1/2298), `referencesUniqueColumn` was removed because it "wasn’t really doing anything," but was kept in the code so people's changelogs didn't break. I wouldn't mind if it were removed from the code now, as that was 10 years ago and I doubt we have many people on 2.X. Nathan might have an opinion, I'm not sure what the standard operating procedure is.

liquibase-core/src/main/java/liquibase/change/core/AddLookupTableChange.java
adrian-velonis1: Go for it @MalloD12!

liquibase-core/src/test/java/liquibase/verify/saved_state/compareGeneratedSqlWithExpectedSqlForMinimalChangesets/createView/db2i.sql
adrian-velonis1: @XDelphiGrl Nathan and I were troubleshooting an issue I was getting in 4.18 (4.19?) that he wasn't getting in whichever slightly earlier version he was on. Clearing out some of the tests seemed to be part of the solution. Honestly, I don't really understand what the issue was, but Nathan would know. Might be fixed by now anyway?

liquibase-core/src/main/java/liquibase/change/core/CreateProcedureChange.java
adrian-velonis1: Intentional! Too long for this file. We have pretty much this exact text on the docs.

liquibase-core/src/main/java/liquibase/change/core/CreateProcedureChange.java
```diff
@@ -155,6 +151,7 @@ public void setDbms(final String dbms) {
         this.dbms = dbms;
     }
 
+    @DatabaseChangeProperty(description = "Comments")
```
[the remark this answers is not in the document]
adrian-velonis1: Good sleuthing. So a description like would be "Inline comments generated by `update-sql`. Not applied to the database" would be fine with me.

liquibase-core/src/main/java/liquibase/change/core/CreateTableChange.java
adrian-velonis1: `liquibase.change.*` should work! I think this may have been my IDE throwing a fit...

liquibase-core/src/main/java/liquibase/change/core/DropColumnChange.java
adrian-velonis1: On docs.liquibase.com we currently have a note "YAML and JSON changelogs using the 'column' tag must nest it within a 'columns' tag." or equivalent wherever relevant. I think this would be appropriate in the class defining the `column` tag for sure, as that would theoretically get pulled onto the changetype pages where it's valid.

`columnName` and `columnNames` are only used for change types that don't use the actual `<columns>` tag, like [modifyDataType](https://docs.liquibase.com/change-types/modify-data-type.html) and [addPrimaryKey](https://docs.liquibase.com/change-types/add-primary-key.html) respectively.

I'm pretty sure you wouldn't use `columnName` or `columnNames` with `dropColumn`. Not sure if this was copy+pasted from another class when it was originally written or if it has some significance I don't understand.

liquibase-core/src/main/java/liquibase/change/core/DropForeignKeyConstraintChange.java
adrian-velonis1: Agreed, though if we make this change we should be consistent with descriptions on `dropTableChange` etc. ("Drops an existing table" -> "Drops a table.") Sometimes it's useful to use emphasis words like "new" or "existing" but I try to only do this as a point of comparison (new/old, existent/not-yet-existent, etc) or if a description could otherwise be confusing without the terms.

liquibase-core/src/main/java/liquibase/change/core/LoadDataChange.java
adrian-velonis1: Thankfully the info is on the [docs page](https://docs.liquibase.com/change-types/load-data.html)!

liquibase-core/src/main/java/liquibase/change/core/RawSQLChange.java
adrian-velonis1: I think this info should be (mostly? Completely?) covered by the docs page on the [<sql>](https://docs.liquibase.com/change-types/sql.html) changetype, where it is also formatted better. If there is anything that needs clarification on that page, we can definitely make adjustments.

If you find anything that shouldn't be in Javadocs but should be in the user-facing docs somewhere, definitely make a PD ticket and tag one of us!

liquibase-core/src/main/java/liquibase/change/core/RenameSequenceChange.java
adrian-velonis1: I think this change was partially just to maintain consistency ("existing" is used in a few places throughout the docs), though I have been trying to think about ways the descriptions can be misinterpreted by users whose first language isn't English. In this case, I was trying to contrast the `oldSequenceName` and `newSequenceName` descriptions. "Old" has a different meaning than "current"/existing, which is something native/proficient speakers would immediately understand, but it's possible that novice English speakers may not immediately make the connection. i.e. conceivably they might assume we're asking for a third (old) sequence name rather than a current one.

Hard to say.

suryaaki2 (approved):

liquibase-core/src/main/java/liquibase/change/core/AddLookupTableChange.java
MalloD12: Hey @adrian-velonis1, if you are planning to address some other description updates, would you mind updating this proposed description too? Otherwise, I don't mind making the required changes here.

Thanks,
Daniel.

liquibase-core/src/main/java/liquibase/change/core/CreateProcedureChange.java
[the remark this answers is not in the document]
XDelphiGrl: That phrasing makes sense to me as well, @adrian-velonis1 .

liquibase-core/src/main/java/liquibase/change/core/EmptyChange.java
```diff
@@ -6,7 +6,7 @@
 import liquibase.database.Database;
 import liquibase.statement.SqlStatement;
 
-@DatabaseChange(name="empty", description = "empty", priority = ChangeMetaData.PRIORITY_DEFAULT)
+@DatabaseChange(name = "empty", description = "empty", priority = ChangeMetaData.PRIORITY_DEFAULT)
```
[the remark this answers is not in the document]
XDelphiGrl: lol! But ... why would someone want an empty array of non-existent SQL statements?!?

adrian-velonis1:
@XDelphiGrl Went through and updated the descriptions based on your comments. I will probably have more tweaks next sprint-ish, but this is the "big PR" out of the way. Thanks again for your detailed review!

@MalloD12 If Erz is all set too then this should be good to go.

liquibase-core/src/main/java/liquibase/change/core/AddForeignKeyConstraintChange.java
nvoxland: We've tended to not remove deprecated code until we're forced to in order to keep backwards compatibility as long as possible. Even if people are on a current version of liquibase now, they may still have changesets written from back then which have those attributes in their "already applied" changesets and removing it will cause problems.

liquibase-core/src/test/java/liquibase/verify/saved_state/compareGeneratedSqlWithExpectedSqlForMinimalChangesets/createView/db2i.sql
nvoxland: There were likely changes to these files based on some of the changed docs (probably the 'example' fixed up) which caused the tests to fail.

We deleted the files to get past them failing, but they should be re-generated and the updated versions committed by running the tests on a developer's machine. Then we can see what it's thinking is different.

XDelphiGrl (approved):
Thank you for the improvements and edits, @adrian-velonis1!
* For some reason, the build did not run for this PR. I've triggered it here: [https://github.com/liquibase/liquibase/actions/runs/4514955239/jobs/7951669951?pr=3811](https://github.com/liquibase/liquibase/actions/runs/4514955239/jobs/7951669951?pr=3811)

When that passes integration, unit and code scanning, this is ready to merge to master.

CC @MalloD12 ^^^

filipelautert (approved):

PR #3820, opened 2023-02-15
fix Liquibase classloader loading in multiple classloaders scenario

### LINKED ISSUE #3819: Liquibase has problem with multiple classloader Cannot find default log service error
## Environment

**Liquibase Version**:
4.19.0

4.19.0

Ubuntu 20.04

**Infrastructure Type/Provider**: 
VM

## Description
```java
        try (Connection conn = DriverManager.getConnection(
                "jdbc:postgresql://ha-ds.zextras.test:5432/ha", "ha", "ha")) {

            if (conn != null) {
                System.out.println("Connected to the database!");
            } else {
                System.out.println("Failed to make connection!");
            }
            Map<String, Object> config = new HashMap<>();
            Scope.child(config, () -> {
                Database database = DatabaseFactory
                        .getInstance()
                        .findCorrectDatabaseImplementation(new JdbcConnection(conn));
                try (Liquibase liquibase = new Liquibase("changelog.xml", new ClassLoaderResourceAccessor(), database)) {
                    liquibase.update(new Contexts(), new LabelExpression());
                }
            });
        } catch (SQLException e) {
            System.err.format("SQL State: %s\n%s", e.getSQLState(), e.getMessage());
        } catch (Exception e) {
            e.printStackTrace();
        }
    }
```
Code causes Cannot find default log service error when run jdk17
```log
liquibase.exception.UnexpectedLiquibaseException: Cannot find default log service
	at liquibase.Scope.getCurrentScope(Scope.java:87)
	at liquibase.Scope.enter(Scope.java:213)
	at liquibase.Scope.child(Scope.java:186)
	at org.example.Main.main(Main.java:31)
```
## Steps To Reproduce
Code causes Cannot find default log service error when run jdk17

sonersivri:
## Impact

## Description

Liquibase has problem with multiple classloader Cannot find default log service error

## Additional Context

Fixes #3819


liquibase-core/src/main/java/liquibase/servicelocator/StandardServiceLocator.java
```diff
@@ -16,21 +16,40 @@ public int getPriority() {
     @Override
     public <T> List<T> findInstances(Class<T> interfaceType) throws ServiceNotFoundException {
         List<T> allInstances = new ArrayList<>();
-
```
lorenzopiovesanzextras, line 19: @sonersivri , I think both the above List and below Set can be both final

lorenzopiovesanzextras (approved):

nvoxland:

We don't have example/test code using multiple classloaders to be able to test it out well. But if it impacts the "normal" single-classloader setup that would get caught by the existing tests. And I'm not sure it's worth blocking this fix to build more robust tests against it.

The original issue has a good example of the project setup that was hitting this problem.

Questions I have:
- None

Potential risks:
- None

- Nothing

sonersivri:
> **Initial/Pre-Review Thoughts**
> 
> We don't have example/test code using multiple classloaders to be able to test it out well. But if it impacts the "normal" single-classloader setup that would get caught by the existing tests. And I'm not sure it's worth blocking this fix to build more robust tests against it.
> 
> The original issue has a good example of the project setup that was hitting this problem.
> 
> Questions I have:
> 
> * None
> 
> Potential risks:
> 
> * None
> 
> What could make the full review difficult: (List what could trip us up and burn time during the review process)
> 
> * Nothing

This method aim is find implemented class I only added look at also given Class classLoader if classLoader is not same as method parameters classLoader.
This should not be problematic.
In our cases we create big jar with dependencies and using it in plugin system of carbonio.

filipelautert:
Hello 
I tried to reproduce it with a simple JakartaEE project that I created at https://github.com/filipelautert/liquibase-jakarta-demo . The "main" class is https://github.com/filipelautert/liquibase-jakarta-demo/blob/main/src/main/java/com/example/liqiubasejakartademo/liquibase/LiquibaseStartupUpdate.java .

I tried in 2 ways: 
* Starting a database test container and creating the jdbc connection
* Getting a connection from a server datasource
 *Tried with Liquibase 4.19 and 4.22 

It worked fine in the 2 cases. Do you have any idea of what I could change to reproduce it? Thanks!

kevin-atx:
@sonersivri - it looks like @filipelautert was unable to reproduce your issue. Would you be able to provide more detailed steps to reproduce?

kevin-atx:
@tati-qalified - could you try to reproduce this issue?

sonersivri:
> @sonersivri - it looks like @filipelautert was unable to reproduce your issue. Would you be able to provide more detailed steps to reproduce?

@kevin-atx @filipelautert you can reproduce using https://github.com/sonersivri/liquibase-bug-3820

filipelautert (approved):
hey @sonersivri , thanks for the example project. I can't see how we could reproduce it using junit as you're doing some serious manual classloading there. 
Anyway, I updated you PR to our current base (just I touched this class so it was simple to fix the merge conflicts) and added some comments trying to explain why you added the second call to findInstances. Could you review them?

And thanks again for the PR!

sonersivri:
> hey @sonersivri , thanks for the example project. I can't see how we could reproduce it using junit as you're doing some serious manual classloading there. Anyway, I updated you PR to our current base (just I touched this class so it was simple to fix the merge conflicts) and added some comments trying to explain why you added the second call to findInstances. Could you review them?
> 
> And thanks again for the PR!

@filipelautert Thanks for review and checks. I checked the PR it looks ok. 
I also test new code in my project it works fine.

rberezen (approved):

sonersivri:
Guys is there any progress about our company planned to change the using liquibase

sonersivri:
My solution is not solve our problem I am closing the PR

PR #3822, opened 2023-02-15
hide log format parameter (DAT-13501)

StevenMassaro:
## Impact

### DIFF
--- a/liquibase-core/src/main/java/liquibase/configuration/ConfigurationDefinition.java
+++ b/liquibase-core/src/main/java/liquibase/configuration/ConfigurationDefinition.java
@@ -37,6 +37,7 @@ public class ConfigurationDefinition<DataType> implements Comparable<Configurati
     private static final Pattern ALLOWED_KEY_PATTERN = Pattern.compile("[a-zA-Z0-9._]+");
 
     private boolean loggedUsingDefault = false;
+    private boolean hidden = false;
 
     /**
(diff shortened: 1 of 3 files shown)


liquibase-core/src/main/java/liquibase/configuration/ConfigurationDefinition.java
```diff
@@ -195,6 +196,13 @@ public boolean isInternal() {
         return internal;
     }
 
+    /**
+     * Return true if this configuration should not be printed to the console for any help command.
+     */
+    public boolean isHidden() {
```
filipelautert, line 202: Will it be a temporary fix just for 4.19.1 ? If so maybe it could be tagged as deprecated or Beta so no one else uses it.

filipelautert (approved):

liquibase-core/src/main/java/liquibase/configuration/ConfigurationDefinition.java
StevenMassaro: I envisioned this a long term addition to the code. I could see it being useful elsewhere. There is also a `isCommonlyUsed` function, with corresponding setting, which ironically isn't used anywhere in the code. Presumably that method (at one time) had the same effect of hiding the property. Since I wasn't sure, I didn't do anything with that method.

suryaaki2 (approved):

PR #3828, opened 2023-02-16
Fixes modules declaration order in parent pom to match build order for Maven 3.9.0

filipelautert:
## Impact

## Description

Maven 3.9.0 is building the modules in the order that they are specified in the liquibase parent pom, instead of sorting them out according to the depedency between them. This fix order the modules in the parent pom in a way that they provide the required dependencies to the next module.

This is not fully fixing Maven 3.9.0 build as there are maven defects that needs to be fixed so it won't break when pulling commercial modules, but it al least allow OSS members to compile full open source code .
Similar maven issues pending fix so we can use 3.9.+: https://issues.apache.org/jira/browse/MNG-7679 (https://github.com/apache/maven/pull/993)

```
[ERROR] NullPointerException
java.lang.NullPointerException
    at org.apache.maven.lifecycle.internal.MojoExecutor.executeForkedExecutions (MojoExecutor.java:439)
    at org.apache.maven.lifecycle.internal.MojoExecutor.doExecute (MojoExecutor.java:325)
    at org.apache.maven.lifecycle.internal.MojoExecutor.execute (MojoExecutor.java:213)

```


StevenMassaro:
@filipelautert Have you found or reported an issue to the maven team about the loss of dependency ordering? That is such a fundamental part of maven that it is hard to imagine them releasing a version where that is broken. I'm not entirely convinced that's what we're seeing here, though I will admit I have not looked into it at the same level of detail as you and @nvoxland

filipelautert:
@StevenMassaro not yet. I see that 3.9.0 fixed some issues related to reactor build order, but need to take more time to understand how it is affecting our build. I'll come back to it after we release 4.19.1 . I created issue https://github.com/liquibase/liquibase/issues/3852 to track it.

filipelautert:
Fixxing it won't solve the problem.

PR #3830, opened 2023-02-16
Adds new outputStream configuration. (DAT-13144)

abrackx:
## Impact

## Description
Adds new outputStream command line configuration. Used to redirect all logging to STDOUT/STDERR.

## Things to be aware of
~This likely has some impact on maven usage, but I have not tested it.~ Turns out it doesn't because it's gated behind params that do not exist in maven.


liquibase-cli/src/main/java/liquibase/integration/commandline/LiquibaseCommandLine.java
```diff
@@ -726,6 +728,27 @@ private void configureLogging(Level logLevel, String logFile, boolean wasDefault
...
+            StreamHandler streamHandler = new StreamHandler(logOutputStream.getOutputStream(), new SimpleFormatter());
+            streamHandler.setLevel(level);
+            setFormatterOnHandler(logService, streamHandler);
+            logger.addHandler(streamHandler);
+            switch (logOutputStream) {
+                case STDOUT:
+                    System.setErr(logOutputStream.getOutputStream());
+                case STDERR:
+                    System.setOut(logOutputStream.getOutputStream());
+            }
```
StevenMassaro, line 748: This should be a method on the enum class.

liquibase-cli/src/main/java/liquibase/integration/commandline/LiquibaseCommandLine.java
```diff
@@ -726,6 +728,27 @@ private void configureLogging(Level logLevel, String logFile, boolean wasDefault
...
 
+    /**
+     * Configures the system output stream and error stream based on supplied global parameter.
+     */
+    private void setOutputStream(Logger logger, LogService logService) {
+        final LogOutputStream logOutputStream = LiquibaseCommandLineConfiguration.LOG_STREAM.getCurrentValue();
+        final Level level = LiquibaseCommandLineConfiguration.LOG_LEVEL.getCurrentValue();
+        if (logOutputStream != null) {
+            Arrays.stream(logger.getHandlers()).forEach(logger::removeHandler); // Unsure of impact of this statement.
+            StreamHandler streamHandler = new StreamHandler(logOutputStream.getOutputStream(), new SimpleFormatter());
```
StevenMassaro, line 739: What if one of the handlers you've removed on line 738 isn't a stream handler? You never put it back. Is that okay?

abrackx: Good point, looks like it works when I only remove handlers that are streamhandlers so I updated it to work that way. I'm unsure of the impact of removing all handlers.

liquibase-cli/src/main/java/liquibase/integration/commandline/LiquibaseCommandLine.java
abrackx: Done!

StevenMassaro (approved):

filipelautert (approved):

liquibase-core/src/main/java/liquibase/logging/SystemOutputStream.java
```diff
@@ -0,0 +1,40 @@
+package liquibase.logging;
+
+import liquibase.util.StringUtil;
+
+import java.io.PrintStream;
+
+public enum SystemOutputStream {
```
suryaaki2, line 7: @abrackx Can you add javadoc comments to the class?

suryaaki2 (changes requested):

abrackx:
Closing in favor of new approach with new ticket.

PR #3835, opened 2023-02-17
set up maven using simpler GitHub action

StevenMassaro:
## Impact

## Description

The GitHub action s4u/setup-maven-action is a "catch-all" and contains a few different actions, many of which are unnecessary. This change simplifies the logic by calling the stCarolas/setup-maven action directly, rather than through another layer.

## Things to be aware of

I don't know for certain that this will fix the issue with the branch builds, but it is my best educated guess at the moment.


jnewton03 (approved):
LGTM 👍

filipelautert (approved):
Seems a very educated guess !

PR #3836, opened 2023-02-17
Remove not using imported classes and packages in the source file.

arturobernalg:
## Impact

## Description
Remove not using imported classes and packages in the source file.


nvoxland:

Does not actually impact functionality, just compilation so as long as it's only imports then "still compiles" is all the test that's really needed.

Questions I have:
- None

Potential risks:
- None

- Nothing

MalloD12 (approved):
Approved.

Changes look good to me. Integration tests were failing, but it was due some `SHOW_SUMMARY` missing changes I added to fix these tests. Functional tests are failing because of the same reason (not related with this PR). I'll investigate that separately with the team. 

### Things to be aware of:
- None

### Things to worry about:
- None

filipelautert (approved):

PR #3837, opened 2023-02-17
Replace with Collections.singletonList(), Collections.emptyList(), or…

arturobernalg:
… List.of() on JDK 9 and later, which will save some memory.

## Impact

## Description

<!--
In particular, Collections.emptyList() and List.of() with no arguments always return a shared instance, while Arrays.asList() with no arguments creates a new object every time it's called.


nvoxland:

Change makes sense, thanks

Questions I have:
- None

Potential risks:
- None

- Nothing

MalloD12 (approved):
**Approved**.

Thanks @arturobernalg!  Changes looks good to me. Making use of singleton for list makes a better use of memory. Build and tests successfully executed, except for functional tests which at the moment are failing because of [DAT-14259](https://datical.atlassian.net/browse/DAT-14259).

### Things to be aware of:
- None

### Things to worry about:
- None

filipelautert (approved):

PR #3839, opened 2023-02-18
Avoid expressions and conditions that always produce the same result.

arturobernalg:
## Impact

## Description
This pull request adds a feature to the project that reports expressions and conditions that always produce the same result. These expressions could be replaced with the corresponding constant value, which could improve code efficiency and maintainability. However, the presence of such expressions could also indicate a bug in the code.

To identify these expressions, the pull request includes new functionality that analyzes the code for expressions that always produce the same result, such as true, false, null, or zero. When such expressions are found, a report is generated that lists the affected expressions and provides recommendations for how to address them.

Overall, this pull request adds valuable functionality to the project that can help developers to improve the quality and reliability of their code. Please let me know if you have any questions or concerns about these changes.

### DIFF
--- a/liquibase-standard/src/main/java/liquibase/database/AbstractJdbcDatabase.java
+++ b/liquibase-standard/src/main/java/liquibase/database/AbstractJdbcDatabase.java
@@ -1202,7 +1202,7 @@ public abstract class AbstractJdbcDatabase implements Database {
 
         if (connection == null) {
             if (that.connection == null) {
-                return this == that;
+                return false;
             } else {
                 return false;
             }
--- a/liquibase-extension-testing/src/main/java/liquibase/extension/testing/testsystem/wrapper/DockerDatabaseWrapper.java
+++ b/liquibase-extension-testing/src/main/java/liquibase/extension/testing/testsystem/wrapper/DockerDatabaseWrapper.java
@@ -124,16 +124,14 @@ public class DockerDatabaseWrapper extends DatabaseWrapper {
             }
         }
 
-        if (ports != null) {
-            List<PortBinding> portBindings = new ArrayList<>();
-            for (int port : ports) {
-                portBindings.add(new PortBinding(Ports.Binding.bindPort(port), new ExposedPort(port)));
-            }
-
-            container.withCreateContainerCmdModifier((Consumer<CreateContainerCmd>) cmd ->
-                    cmd.withHostConfig(new HostConfig().withPortBindings(portBindings))
-            );
+        List<PortBinding> portBindings = new ArrayList<>();
+        for (int port : ports) {
+            portBindings.add(new PortBinding(Ports.Binding.bindPort(port), new ExposedPort(port)));
         }
+
+        container.withCreateContainerCmdModifier((Consumer<CreateContainerCmd>) cmd ->
+                cmd.withHostConfig(new HostConfig().withPortBindings(portBindings))
+        );
     }
 
     @Override
--- a/liquibase-maven-plugin/src/main/java/org/liquibase/maven/plugins/AbstractLiquibaseMojo.java
+++ b/liquibase-maven-plugin/src/main/java/org/liquibase/maven/plugins/AbstractLiquibaseMojo.java
@@ -1049,7 +1049,7 @@ public abstract class AbstractLiquibaseMojo extends AbstractMojo {
 
         Object defaultValue = getDefaultValue(f);
         if (defaultValue == null) {
-            return currentValue != null;
+            return true;
         } else {
             // There is a default value, check to see if the user has selected something other
             // than the default
--- a/liquibase-maven-plugin/src/main/java/org/liquibase/maven/plugins/LiquibaseRollback.java
+++ b/liquibase-maven-plugin/src/main/java/org/liquibase/maven/plugins/LiquibaseRollback.java
@@ -81,14 +81,11 @@ public class LiquibaseRollback extends AbstractLiquibaseChangeLogMojo {
             }
             type = RollbackType.COUNT;
         } else if (rollbackDate != null) {
-            if ((rollbackTag != null) || (rollbackCount > 0)) {
+            if (rollbackTag != null) {
                 throw new MojoFailureException(message);
             }
             type = RollbackType.DATE;
-        } else if (rollbackTag != null) {
-            if ((rollbackCount > 0) || (rollbackDate != null)) {
-                throw new MojoFailureException(message);
-            }
+        } else {
             type = RollbackType.TAG;
         }
     }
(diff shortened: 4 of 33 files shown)


nvoxland:

For the full code review, make sure to look at check that the removed code is actually pointless and/or if the real problem is that we're missing an extra "else throw an unexpected state" thing. 

Questions I have:
- None

Potential risks:
- None

- Nothing

liquibase-core/src/main/java/liquibase/change/core/DropColumnChange.java
```diff
@@ -215,7 +215,7 @@ public boolean createThisIndex(Index index) {
             sqlStatements = new SqlStatement[statements.size()];
             return statements.toArray(sqlStatements);
         }
-        return sqlStatements;
+        return null;
```
MalloD12, line 218: updating this to return an empty array as this is going to be use like this: `Arrays.asList(generateStatementsForSQLiteDatabase(database)`

liquibase-core/src/main/java/liquibase/change/core/DropPrimaryKeyChange.java
```diff
@@ -144,7 +144,7 @@ public boolean createThisIndex(Index index) {
             sqlStatements = new SqlStatement[statements.size()];
             return statements.toArray(sqlStatements);
         }
-        return sqlStatements;
+        return null;
```
MalloD12, line 147: same here

liquibase-core/src/main/java/liquibase/change/core/MergeColumnChange.java
```diff
@@ -128,7 +128,7 @@ public SqlStatement[] generateStatements(final Database database) {
         statements.addAll(Arrays.asList(addNewColumnChange.generateStatements(database)));
 
         String updateStatement = "";
-        if (database instanceof MySQLDatabase || database instanceof MariaDBDatabase) {
```
MalloD12, line 131: Hey @arturobernalg, why are we removing MariaDB check? As far as I know, we can use `CONCAT` function to do so.

liquibase-core/src/main/java/liquibase/diff/compare/DatabaseObjectComparatorChain.java
```diff
@@ -82,11 +82,11 @@ public ObjectDifferences findDifferences(DatabaseObject object1, DatabaseObject
         if ((object1 == null) && (object2 == null)) {
             return new ObjectDifferences(compareControl);
         }
-        if ((object1 == null) && (object2 != null)) {
+        if (object1 == null) {
             return new ObjectDifferences(compareControl).addDifference("Reference value was null", "this", null, null);
```
MalloD12, line 86: It's not directly related with the changes made here, but I'm seeing the only difference between the two `ObjectDifferences` object is the message set. I think this should be something like:
```
return new ObjectDifferences(compareControl).addDifference("Reference value was null", "this", null, object2);```
and:
```
return new ObjectDifferences(compareControl).addDifference("Reference value was null", "this", object1, null);```
```

File: liquibase-core/src/main/java/liquibase/diff/output/changelog/DiffToChangeLog.java [hunk 5]
```diff
@@ -608,11 +608,9 @@ protected void addDependencies(DependencyUtil.DependencyGraph<String> graph, Lis
                 String tabName = StringUtil.trimToEmpty((String)row.get("REFERENCED_SCHEMA_NAME")) +
                         "." + StringUtil.trimToEmpty((String)row.get("REFERENCED_NAME"));
 
-                if (!(tabName.isEmpty() || bName.isEmpty())) {
```
MalloD12, line 611: Why this if block is removed? Is there not chance both `tabName` and `bName` are empty?

liquibase-core/src/main/java/liquibase/snapshot/DatabaseSnapshot.java
```diff
@@ -447,9 +447,9 @@ private Object replaceObject(Object fieldValue) throws DatabaseException, Invali
             List<Object> newValues = new ArrayList<>();
             while (fieldValueIterator.hasNext()) {
                 Object obj = fieldValueIterator.next();
-                if ((fieldValue instanceof DatabaseObject) && !snapshotControl.shouldInclude(((DatabaseObject)
-                        fieldValue).getClass())) {
-                    return fieldValue;
+                if ((fieldValue instanceof DatabaseObject)) {
```
MalloD12, line 450: We have changed the comparison here and not returning anything here now. Is that intended?

liquibase-core/src/main/java/liquibase/snapshot/JdbcDatabaseSnapshot.java
```diff
@@ -1635,8 +1635,7 @@ private String createSql(String catalogName, String schemaName, String tableName
                         if (tableName != null) {
                             sql += " and table_name='" + tableName + "'";
                         }
-                    } else if ((database instanceof MySQLDatabase) || (database instanceof HsqlDatabase) || (database
```
MalloD12, line 1638: Why are we not including MariaDB now?

MalloD12:
Hi @arturobernalg,

Have you had a chance to look at my comments here?

Thanks,
Daniel.

arturobernalg:
> Hi @arturobernalg,
> 
> Have you had a chance to look at my comments here?
> 
> Thanks, Daniel.

Hi @MalloD12 
Sorry for the late response. I'll take a look to this PR and the other one this weekend
TY

MalloD12:
> > Hi @arturobernalg,
> > Have you had a chance to look at my comments here?
> > Thanks, Daniel.
> 
> Hi @MalloD12 Sorry for the late response. I'll take a look to this PR and the other one this weekend TY

Sure, I appreciate it @arturobernalg!

liquibase-core/src/main/java/liquibase/change/core/DropPrimaryKeyChange.java
arturobernalg: this is an inheritance changes

liquibase-core/src/main/java/liquibase/diff/compare/DatabaseObjectComparatorChain.java
arturobernalg: 'object2 != null))' this is true always when reach the conditions.

liquibase-core/src/main/java/liquibase/diff/output/changelog/DiffToChangeLog.java
arturobernalg: Because Condition '(fieldValue instanceof DatabaseObject) && !snapshotControl.shouldInclude(((DatabaseObject) ...' is always 'false'

arturobernalg:
I haven't review  all the case but  almost sure that the change its redudant

liquibase-standard/src/main/java/liquibase/database/AbstractJdbcDatabase.java
```diff
@@ -1202,7 +1202,7 @@ public boolean equals(final Object o) {
 
         if (connection == null) {
             if (that.connection == null) {
```
MalloD12, line 1204: @arturobernalg excuse me, but I don't see the point of having this inner IF block if in any case it will return `false`.

MalloD12:
Hi @arturobernalg, the other day when looking into this PR again I realised my changes were reverted (maybe because of the forced push) and can't see any new change here as well as no answers to my questions. Was this unintended, or do you think those changes make no sense and you would like to keep the original approach of this PR?   

Thanks,
Daniel.

arturobernalg:
Hey Everyone,

I have decided to close this pull request and begin anew. I apologize for any inconvenience caused by this decision. Thank you for your understanding.

Best regards,

filipelautert:
No problems!

Issue #3816, opened 2023-02-14
Execution/processing time for initial databasechangelog read is excessive when switching from v3.5.3 to v4.17.2


doakd:
## Environment

**Liquibase Version**: 
old: v3.5.3
new: v4.17.2

**Liquibase Integration & Version**: <Pick one: CLI, maven, gradle, spring boot, servlet, etc.>
CLI

**Liquibase Extension(s) & Version**: 

**Database Vendor & Version**:
Oracle 19c

**Operating System Type & Version**:
Linux

**Infrastructure Type/Provider**: <AWC, GCS, Azure, VM, etc>
On-Prem

## Description

A customer is updating from a v3.5.3 "self-supported" Liquibase solution to our v4.17.2 "officially-supported" Liquibase solution.

Using the same changelogs and database/schema:
In v3.5.3, the initial "Reading from databasechangelog" step is taking 1 second.
In v4.17.2 the initial "Reading from databasechangelog" step is taking 10 minutes, and consistently takes this long.  I monitored in the database and the actual query is taking < 1 second, so I believe the additional time is spent in Liquibase code.

I also had then purge the databasechangelog table of v.3.5.3 records, and run v4.17.2 "changelog-sync" to repopulate, didn't help.

## Expected/Desired Behavior
The "Reading from databasechangelog" should be approximately the same between v3.5.3 and v4.17.2

## Screenshots (if appropriate)
v3.5.3 output:
INFO 2/9/23 10:39 AM: liquibase: Reading from TPPONPREMDEV.DATABASECHANGELOG
DEBUG 2/9/23 10:39 AM: liquibase: Executing QUERY database command: SELECT * FROM TPPONPREMDEV.DATABASECHANGELOG ORDER BY DATEEXECUTED ASC, ORDEREXECUTED ASC
DEBUG 2/9/23 10:39 AM: liquibase: master.xml: session-start/metadata.xml::alter_session_start::Ebix: Computed checksum for inputStream as d2a9c9002a66f9fe4e8c4aa41810b63c

v4.17.2 output:
[2023-02-08 22:18:30] INFO [liquibase.changelog] Reading from TPPONPREMDEV.DATABASECHANGELOG
Running Changeset: session-start/metadata.xml::alter_session_start::Ebix
[2023-02-08 22:28:35] INFO [liquibase.executor] ALTER SESSION SET NLS_TIMESTAMP_FORMAT = 'DD-MON-YYYY HH24:MI:SS.FF6'
[...]

PR #3853, opened 2023-02-22
Performance Improvement: optimized DatabaseChangeLog.normalizePath()

nvoxland:
## Impact

## Description

PR #3063 improved performance of `DatabaseChangeLog.normalizePath()` by no longer re-compiling regular expressions, but it is a method that is called often and with a large changelog it still appears to be a bottleneck. 

This PR replaces the regular expressions with more efficient "replace" calls and/or adds a "if it might be a problem" gate around the regexp call.

From testing, it removed about half the time spent based on profiling data from #3816

Fixes partially #3816

## Things to be aware of

It may be possible to reduce the number of calls to this function in the first place as well. As part of the review process, it is probably worth looking to see if we keep re-calling normalizePath for the same paths over and over or not.

In testing this, the customer reported that it **caused some validation errors** (see #3816). I'm not sure how it could off hand, since it's just in the normalizePath, it's possibly causing some changesets to be identical now? So needs a bit more unit testing 

## Things to worry about

Not much, there are existing tests around the normalizePath() which would fail as I pulled out each regexp before replacing it with a comparable replace() call


nvoxland:
The validation errors appear to be caused by a separate change that was already in master. #3881

filipelautert (approved):
Let's get the performance back!

suryaaki2 (approved):

PR #3867, opened 2023-02-27
feat : https://datical.atlassian.net/browse/DAT-12842

sayaliM0412:
## Impact

`release-published.yml` will now include deploying `liquibase` and `liquibase-pro` `xsd` to cloud front distribution. 

## Description

`https://datical.atlassian.net/browse/DAT-12842` :  Automating the deployment of XSDs to new beta landing site.

## Things to be aware of

Need to make sure that the desired/ current `xsd`'s are indeed getting uploaded to s3 bucket at `release-published` workflow.


.github/workflows/release-published.yml
```diff
@@ -222,4 +222,32 @@ jobs:
...
+          path: liquibase-pro
+          repository: "liquibase/liquibase-pro"
+      
+      - name: Upload to s3
+        env:
+          AWS_ACCESS_KEY_ID: ${{ secrets.LIQUIBASEORIGIN_ACCESS_KEY_ID }}
+          AWS_SECRET_ACCESS_KEY: ${{ secrets.LIQUIBASEORIGIN_SECRET_ACCESS_KEY }}
+          AWS_DEFAULT_REGION: us-east-1
+        run: |
+          aws s3 sync $GITHUB_WORKSPACE/liquibase-core/src/main/resources/www.liquibase.org/xml/ns/dbchangelog/dbchangelog-${{ needs.setup.outputs.version }}.xsd s3://liquibaseorg-origin --only-show-errors
```
jandroav, line 252: I took a look at https://github.com/liquibase/liquibase/tree/2de02c6cb0cea9fa425be05755a0643d451aa671/liquibase-core/src/main/resources/www.liquibase.org/xml/ns/dbchangelog and I could not see `xsd` files for patch versions like `4.17.x`
I am not sure how the release workflow handles patch versions, it seems `xsd` files are handled by `major.minor` so not sure what is going to happened when somebody launches this workflow for `4.19.1`

jnewton03:
aside from Jandro's comment I think this looks good.  You might be able to use `working_directory` instead of the Github Path environment variables for cleaner looking code, but that's splitting hairs :)

I think once this is ready to test we'll want to setup a manual trigger and comment out everything else so that we can confirm the process works without having to wait on a release.

.github/workflows/release-published.yml
sayaliM0412: "There won't be any `xsd`'s for patch release, if there is an `xsd` update then it automatically becomes a major release". So for this it would be still : [dbchangelog-4.19.xsd](https://github.com/liquibase/liquibase/blob/2de02c6cb0cea9fa425be05755a0643d451aa671/liquibase-core/src/main/resources/www.liquibase.org/xml/ns/dbchangelog/dbchangelog-4.19.xsd)

sayaliM0412: 👍 added the logic to handle patch releases

filipelautert (approved):
No more XSD manual work! wheee

changelog.txt
```diff
@@ -197,11 +197,11 @@ Changes in version Liquibase v4.18.0 (2022.11.30)
 - Use together for increased security and convenience of centralized properties
 
 * [PRO]Security
-- Vaults offer much greater security than defaults files or environment variables 
+- Vaults offer much greater security than defaults files orenvironment variables
```
jnewton03, line 200: typo :)

jnewton03 (approved):
looks great @sayaliM0412!  Can be moved to `RFT`

changelog.txt
sayaliM0412: I saw that, but I din't change it so wasn't sure what it should be.

sayaliM0412: I am assuming it is "for"

sayaliM0412:
> 

the pr is has an unsuccessful check : 
@merge-freeze
mergefreeze — Frozen by Surya Aki

.github/workflows/release-published.yml
```diff
@@ -222,4 +222,46 @@ jobs:
...
+  deploy_xsd:
+      name: Upload xsds
+      needs: [ setup ]
+      runs-on: ubuntu-20.04
+      outputs:
+          tag: ${{ steps.collect-data.outputs.tag }}
+          version: ${{ needs.setup.outputs.version }}
+      steps:
+      - name: Download liquibase xsd
+        uses: actions/checkout@v3
```
nvoxland, line 236: Deploying from master rather than what is released risks drift between what's in master and what has been released. 

In theory they should be the same, and I know there was discussion on the implementation options, so I'll just note that we are OK with that risk.

.github/workflows/release-published.yml
```diff
@@ -222,4 +222,46 @@ jobs:
...
+          # Relative path under $GITHUB_WORKSPACE to place the repository
+          path: liquibase-core  
+          repository: "liquibase/liquibase"
+    
+      - name: Download liquibase-pro xsd
+        uses: actions/checkout@v3
+        with:
+        # Relative path under $GITHUB_WORKSPACE to place the repository
+          path: liquibase-pro
+          repository: "liquibase/liquibase-pro"
```
nvoxland, line 247: I don't think the default token has read access to liquibase-pro. You probably need to add `token: ${{ secrets.BOT_TOKEN }} to `with`

.github/workflows/release-published.yml
```diff
@@ -222,4 +222,46 @@ jobs:
...
+
+      - name: Upload to s3
+        # if patch_version is 0, only then upload to s3 as it's a new minor release 
+        if: ${{ env.PATCH_VERSION }} == 0
+        env:
+          AWS_ACCESS_KEY_ID: ${{ secrets.LIQUIBASEORIGIN_ACCESS_KEY_ID }}
+          AWS_SECRET_ACCESS_KEY: ${{ secrets.LIQUIBASEORIGIN_SECRET_ACCESS_KEY }}
+          AWS_DEFAULT_REGION: us-east-1
+        run: |
+          aws s3 sync $GITHUB_WORKSPACE/liquibase-core/src/main/resources/www.liquibase.org/xml/ns/dbchangelog/dbchangelog-${{ env.XSD_VERSION }}.xsd s3://liquibaseorg-origin --only-show-errors
```
nvoxland, line 266: I don't know how the s3 sync command works, but nether seem to have a target directory listed? They go in different spots in the s3 hierarchy, is that being managed somewhere?

.github/workflows/release-published.yml
suryaaki2: @nvoxland What's released will be pushed to master and step to upload XSD's comes right after that

sayaliM0412: we had some discussion around this and as those are xsd's which are not being committed with every push as such. So I still think that would be fine. But if not, I can do something like downloading the entire `.zip` and extracting the `xsd` from there.

.github/workflows/release-published.yml
sayaliM0412: I just got access to the SFTP server and it will include the full path in my next commit.

.github/workflows/release-published.yml
```diff
@@ -222,4 +222,47 @@ jobs:
...
+          repository: "liquibase/liquibase-pro"
+
+      - name: Get major.minor xsd
+        run: |
+          version=${{ needs.setup.outputs.version }}
+          arr=(${version//./ })
+          xsd_version=${arr[0]}"."${arr[1]}
+          patch_version=${arr[2]}
+          echo ::set-env name=XSD_VERSION::${xsd_version}
+          echo ::set-env name=PATCH_VERSION::${patch_version}
```
jandroav, line 257: `::set-env` was deprecated: https://github.blog/changelog/2020-10-01-github-actions-deprecating-set-env-and-add-path-commands/
You can use `echo "YOUR_VAR=value" >> $GITHUB_ENV` and then access it like `${{ YOUR_VAR }}`

.github/workflows/release-published.yml
jnewton03: @nvoxland is this comment resolved?

sayaliM0412: I have added a full path for the `xsd`'s which will grab the `xsd` from the source in `liquibase` GitHub repo to the target dir in `s3` which are different for `liquibase-pro` and `liquibase`.

jnewton03:
@sayaliM0412 this looks good to me.

PR #3868, opened 2023-02-27
DAT-13659 Fix nullpointers on table output

filipelautert:
Fix logic that clear out deploymentId when calling hub.register

## Impact

## Description

Since Liquibase 4.18, `StandardLockService.reset` calls out `changelogService.reset()` that clears de deployment id. HubService.regsiter is releasing the lock to wait for user input, so it's also clearing the deployment id.

### DIFF
--- a/liquibase-core/src/main/java/liquibase/Liquibase.java
+++ b/liquibase-core/src/main/java/liquibase/Liquibase.java
@@ -260,15 +260,13 @@ public class Liquibase implements AutoCloseable {
                     checkLiquibaseTables(true, changeLog, contexts, labelExpression);
                 }
 
-                //
+                changeLog.validate(database, contexts, labelExpression);
+
                 // Let the user know that they can register for Hub
-                //
                 hubUpdater = new HubUpdater(new Date(), changeLog, database);
                 hubUpdater.register(changeLogFile);
 
                 generateDeploymentId();
-
-                changeLog.validate(database, contexts, labelExpression);
 
                 //
                 // Create or retrieve the Connection if this is not SQL generation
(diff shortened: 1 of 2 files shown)


liquibase-core/src/main/java/liquibase/Liquibase.java
```diff
@@ -259,16 +259,17 @@ public void update(Contexts contexts, LabelExpression labelExpression, boolean c
...
-
-                changeLog.validate(database, contexts, labelExpression);
 
                 //
                 // Let the user know that they can register for Hub
                 //
                 hubUpdater = new HubUpdater(new Date(), changeLog, database);
                 hubUpdater.register(changeLogFile);
 
+                generateDeploymentId();
```
suryaaki2, line 269: @filipelautert deploymentId should be generated with or without hub, then why is generateDeploymentId method gets called after hub?

suryaaki2 (changes requested):

liquibase-core/src/main/java/liquibase/Liquibase.java
filipelautert: it happens that `hubUpdater.register(changeLogFile) `is calling `lockService.releaseLock()` before prompting the user if he wants to subscribe or not to Hub, and the releaseLock is clearing the deploymentId. By moving `generateDeploymentId()` to be called after Hub we don't have this problem.
I also verified and `hubUpdater.register(changeLogFile);` does not need the deploymentId .

I'm just reviewing the code as `changeLog.validate(database, contexts, labelExpression)` should still being called before hubUpdater.

abrackx (approved):

suryaaki2 (approved):

StevenMassaro (approved):

PR #3869, opened 2023-02-27
DAT-13664 - skip MDC logging for MarkChangeSetRanStatement SQLs 

filipelautert:
## Impact

## Description

Skip MDC logging for MarkChangeSetRanStatement SQLs preventing skipping order executed ids.

## Things to be aware of

I won't call this a fix - it's more of a temporary workaround.


suryaaki2 (approved):

wwillard7800:
I had two thoughts on this issue:

1.  To make this solution cleaner, we could add a `isLoggable` method to the `SqlStatement` interface, with a default implementation that returns true.  `GetNextChangeSetSequenceValueStatement` would return false.  
2.  We could implement logic to tell the `StandardChangeLogHistoryService` to return the next sequence value, without incrementing the private member variable `lastChangeSetSequenceValue`.  That sounds messy, though.

PR #3879, opened 2023-03-01
add helper command step to set output writer to console (DAT-13194)

StevenMassaro:
## Impact

## Description

This helper class can be run prior to any command (but usually the *-sql commands, like update-sql) to redirect the SQL to the console, rather than running it against an actual database.


filipelautert (approved):
Nice, I'll use this on DAT-6603 too instead of duplicating this code.

PR #3882, opened 2023-03-02
Run expected integration tests, fix tests that were not run (DAT-13751)

abrackx:
## Impact

## Description
Integration tests not marked with the suffix `IntegrationTest` are being ignored during our integration test workflow. This pr updates these tests to have the suffix and fixes `DiffChangelogIntegrationTest` which was failing locally but not in the pipeline, because it was not being executed.


wwillard7800 (approved):
Thanks for the fix!

StevenMassaro (approved):

filipelautert (approved):
After all tests passes it's ready to go!

filipelautert (approved):

StevenMassaro (approved):
Nice!

Do you think we should update the surefire plugin properties to also recognize files that end only in `Test`?

StevenMassaro:
> > Nice!
> > Do you think we should update the surefire plugin properties to also recognize files that end only in `Test`?
> 
> I think we should examine other options (like using `@Category` or `@Tag` in junit5) to run the specific suites. That may simplify things going forward and we won't need any specific suffix to match.

I was thinking along the lines of ensuring that no developer needs to remember to do "something" in order for the tests to run. They can just write a test class, call it `WhateverTest.groovy` or `WhateverIntegrationTest.groovy` or `WhateverIT.groovy` and it'll get picked up automatically. Is that not within the realm of possibility? Maybe it has other undesirable side effects I'm not thinking of.

PR #3595, opened 2022-12-17
Issue 3584: Add support for relativeToChangelogFile for ChangeLogProperty class/property attribute

jasonlyle88:
## Impact

- [ ] Bug fix (non-breaking change which fixes expected existing functionality)
- [X] Enhancement/New feature (adds functionality without impacting existing logic)
- [ ] Breaking change (fix or feature that would cause existing functionality to change)
 
## Description

A clear and concise description of the change being made.  

Add a new attribute to the `property` changeLog attribute for `relativeToChangelogFile`. Modified the `ChangeLogProperty` class to add the new property along with setter/getter function. Modified the `dbchangelog-latest.xsd` to know about the new attribute. Then modified the `DatabaseChangeLog` and `YamlChangeLogParser` classes to handle the `relativeToChangelogFile` property correctly in regards to the `property` `file` attribute.

## Things to be aware of

Used the code from the `include` attribute for the `relativeToChangelogFile`, so it should be solid.

I also defaulted the `relativeToChangelogFile` attribute to false so that it does not break current behavior.

PR #3901, opened 2023-03-04
Add errorIfMissingOrEmpty attribute to ChangeLogProperty and ChangeLogInclude classes

jasonlyle88:
## Impact

## Description

Add the `errorIfMissingOrEmpty` functionality that is available in `includeAll` to `ChangeLogInclude` and `ChangeLogProperty` classes (when `property` is using a file).

## Things to be aware of

I have set the default value for `errorIfMissingOrEmpty` to be `true` for both `property` and `include`. This is the functionality for `include` and was the functionality for `property` until PR https://github.com/liquibase/liquibase/pull/3595. So this PR adds the `errorIfMissingOrEmpty` functionality as well as fixes the change from original behavior introduced in PR https://github.com/liquibase/liquibase/pull/3595. 

Additionally I have created tests for all 18 possible use cases that could arise with a property file based on 3 possible values for `errorIfMissingOrEmpty` (unset, `true`, `false`) + the 3 possible values for `relativeToChangelogFile` (unset, `true`, `false`) + 2 categorical values for `file` (A value that can be resolved to a file and an value that cannot be resolved to a file). I believe that should cover everything, but let me know if not.

I have also added two additional tests for `include` and the `errorIfMissingOrEmpty` attribute.

## Things to worry about

Nothing of which I am aware.

## Additional Context

@filipelautert This is the change you asked me to tag you in from PR https://github.com/liquibase/liquibase/pull/3595


jasonlyle88:
Since the original PR (`errorIfMissingOrEmpty` for only `ChangeLogProperty`), I realized I forgot that I wanted to implement this functionality for both `ChangeLogProperty` and ``ChangeLogInclude`. So thats what the update of the title and original description are for (as well as the new commit)

nvoxland (commented):
The `var/folders` files should not have been committed

The new files in `test/java/liquibase/verify` are ones that could be committed since they get generated, but maybe best to leave out of this PR since I'd expect nothing in those directories to be impacted and the newly committed files makes it harder to see that.

Would you be able to clear those out, @jasonlyle88 ?

jasonlyle88:
@nvoxland Absolutely, easy enough to do. Do you want me to add the following entries to the .gitignore file as well:

1. liquibase-core/src/test/java/liquibase/verify/**
2. liquibase-core/var/folders/**

Or maybe:

1. */src/test/java/liquibase/verify/**
2. */var/folders/**

?

Just let me know

nvoxland:
Thanks for fixing it up, @jasonlyle88

> Do you want me to add the following entries to the .gitignore file as well?

We don't want the "verify" one in there. We want to be tracking files in there, there's just some that haven't been added like they should be, so the fix is to bring those in as an independent PR.

The "var" ones shouldn't be added ever, but I'm not sure where those came from. Maybe something unique to your setup? If that's the case, we shouldn't add it. If it was created by normal test running or something then we should consider it.

jasonlyle88:
@nvoxland Makes sense. I'm using IntelliJ Idea. The "var" files showed up when I ran the TEST maven lifecycle, nothing special as far as I'm concerned. Happy to add the `*/var/folders/**` to the gitignore if you want, or I can just make sure not to add them for now... either way works for me!

MalloD12 (approved):
**Approved**.

Thank you @jasonlyle88 for another PR. Overall this one looks good to me. I have made few minor changes which I describe below:
- Update some `DatabaseChanageLog` tests which were failing after resolving some merge conflicts.
- Rename attribute name from `errorIfMissingOrEmpty` to `errorIfMissing`. I'll leave this open for discussion, but I basically consider this change needed as we are not adding any extra logic to check and do some action if change log is empty.
- Add an error exception in case `includeAll()` fails when trying to read change logs from specified directory. 
- Tests successfully executed.

### Things to be aware of:
- None

### Things to worry about:
- None

jasonlyle88:
Hey @MalloD12 , no problem, thanks for taking care of everything for me again!

Only issue I have is that I never touched the `includeAll` stuff. That is all stuff that has been around for a bit, so I'm not sure if you want to change the `includeAll` from `errorIfMissingOrEmpty` to `errorIfMissing` for this specific case. I think `includeAll` does specific checks for empty directories. My stuff should absolutely be changed to `errorIfMissing` since it doesn't do anything special if the specified file is empty.

jasonlyle88:
I guess my concern is it is existing functionality for `includeAll` and this is a breaking change.  So anybody that has changelogs already utilizing the `errorIfMissingOrEmpty` attribute of an `includeAll` element will all of the sudden have their changelog fail

MalloD12:
> I guess my concern is it is existing functionality for `includeAll` and this is a breaking change. So anybody that has changelogs already utilizing the `errorIfMissingOrEmpty` attribute of an `includeAll` element will all of the sudden have their changelog fail

Yeap, this is changing a bit the logic I shared above in the case `errorIfMissingOrEmpty = false`. I think it shouldn't be a `&&` and we should change it to  an `||`, but doing that I'm not really sure whether it makes too much sense about having this one attribute for `includeAll`. What do you guys think about it @filipelautert @XDelphiGrl?

jasonlyle88:
My wife is going to be having our first baby any day now, so I'll probably be absent for a while! I'm good with everything here, just wanted to raise the points above to make sure they were recognized. Look forward to seeing this merged in when I eventually emerge from a daze of no sleep 🤣

XDelphiGrl:
Congratulations, @jasonlyle88!

MalloD12:
> My wife is going to be having our first baby any day now, so I'll probably be absent for a while! I'm good with everything here, just wanted to raise the points above to make sure they were recognized. Look forward to seeing this merged in when I eventually emerge from a daze of no sleep 🤣

Congratulations @jasonlyle88, wish you all the best!

filipelautert (approved):

XDelphiGrl:
Functional Tests Rerunning Here: [https://github.com/liquibase/liquibase-pro-tests/actions/runs/5003360452](https://github.com/liquibase/liquibase-pro-tests/actions/runs/5003360452)

Test Harness Execution Running Here: [https://github.com/liquibase/liquibase-test-harness/actions/runs/5003369502](https://github.com/liquibase/liquibase-test-harness/actions/runs/5003369502)

liquibase-standard/src/main/resources/www.liquibase.org/xml/ns/dbchangelog/dbchangelog-4.21.xsd
XDelphiGrl, line 1: @filipelautert , hi - I believe we need to make the change in `dbchangelog-latest.xsd`, not in the 4.21 version of the XSD.

XDelphiGrl:
Added `releaseMajor` label as there is an XSD change in this PR.

liquibase-standard/src/main/resources/www.liquibase.org/xml/ns/dbchangelog/dbchangelog-4.21.xsd
MalloD12: My apologies, I haven't seen your comment before. Yeah, I agree with you. I just moved these attribute changes to the latest version. Thank you!

MalloD12:
> Thanks for the well wishes @XDelphiGrl and @MalloD12! Everything went well and my wife and baby girl are doing great!
> 
> Was hoping to see this merged in by now though! I'm sure you guys are busy with the 4.22 issue with the new checksum algorithm, so I get it though! This does have a XSD change and so I see the releaseMajor tag was applied. Do we think this can be released with 4.23 this month? Or is there anything else that needs to be done here?

Glad to read everything went well @jasonlyle88, for some reason, I might miss Erzebet's comment. I just applied her suggestion so it's ready to be reviewed it again. 

Thanks,
Daniel.

suryaaki2 (approved):

liquibase-standard/src/main/resources/www.liquibase.org/xml/ns/dbchangelog/dbchangelog-4.21.xsd
XDelphiGrl: Thanks for making this change, @MalloD12.

liquibase-standard/src/main/java/liquibase/changelog/DatabaseChangeLog.java
```diff
@@ -622,7 +629,7 @@ public boolean isDbmsMatch(Set<String> dbmsSet) {
     public void includeAll(String pathName,
```
XDelphiGrl, line 629: Most `@deprecated` annotations include the method to use instead of the deprecated method. Can we add that information to this `@deprecated` as well? It might also be useful to note what should be used instead of `LabelExpression`. 

CC @MalloD12 @filipelautert

XDelphiGrl:
@MalloD12, hi -

Thanks for including this information in the PR! It was helpful during my review.
>     * Rename attribute name from `errorIfMissingOrEmpty` to `errorIfMissing`. I'll leave this open for discussion, but I basically consider this change needed as we are not adding any extra logic to check and do some action if the change log is empty.

CC @rberezen

liquibase-standard/src/main/java/liquibase/changelog/DatabaseChangeLog.java
```diff
@@ -750,6 +757,7 @@ public void includeAll(String pathName,
     @Deprecated
```
XDelphiGrl, line 757: This `@Deprecated` should also include what method to use instead of this one.

liquibase-standard/src/main/java/liquibase/parser/core/yaml/YamlChangeLogParser.java
```diff
@@ -101,12 +102,17 @@ private Map parseYamlStream(String physicalChangeLogLocation, Yaml yaml, InputSt
     private void loadChangeLogParametersFromFile(ChangeLogParameters changeLogParameters, ResourceAccessor resourceAccessor, DatabaseChangeLog changeLog, Map property, ContextExpression context, Labels labels, Boolean global) throws IOException, LiquibaseException {
         Properties props = new Properties();
         Boolean relativeToChangelogFile = (Boolean) property.get("relativeToChangelogFile");
+        Boolean errorIfMissing = (Boolean) property.get("errorIfMissing");
```
XDelphiGrl, line 105: Are there other changelog parser classes (like JSON??) that would also need this update?

CC @rberezen @MalloD12

liquibase-standard/src/test/groovy/liquibase/changelog/DatabaseChangeLogTest.groovy
```diff
@@ -473,14 +472,14 @@ http://www.liquibase.org/xml/ns/dbchangelog http://www.liquibase.org/xml/ns/dbch
                 "include-all-dir/include-all.xml": changelogText,
         ])
         def changeLogFile = new DatabaseChangeLog("com/example/root.xml")
-        changeLogFile.includeAll("include-all-dir", false, null, true, changeLogFile.getStandardChangeLogComparator(), resourceAccessor, new ContextExpression(), new LabelExpression(), false)
+        changeLogFile.includeAll("include-all-dir", false, null, true, changeLogFile.getStandardChangeLogComparator(), resourceAccessor, new ContextExpression(), new Labels(), false, 0, Integer.MAX_VALUE)
 
         then:
         SetupException e = thrown()
         assert e.getMessage().startsWith("liquibase.exception.SetupException: Circular reference detected in 'include-all-dir/'. Set liquibase.errorOnCircularIncludeAll if you'd like to ignore this error.")
```
XDelphiGrl, line 479: I did not even realize there is a `errorOnCircularIncludeAll` parameter! This probably should also be added to the documentation when `errorIfMissing` is added.

CC @MalloD12 @rberezen

XDelphiGrl (changes requested):
I've left a few questions and suggestions in comments.

CHANGES REQUESTED

liquibase-standard/src/main/java/liquibase/changelog/DatabaseChangeLog.java
MalloD12: Sure, I'll make this update.

liquibase-standard/src/main/java/liquibase/parser/core/yaml/YamlChangeLogParser.java
MalloD12: My understanding is that there is not, because for example, JSON changelog parser extends from this one (YamlChangeLogParser), so I think with this one we are ok.

liquibase-standard/src/test/groovy/liquibase/changelog/DatabaseChangeLogTest.groovy
MalloD12: Ups I didn't see this comment before of creating a PD ticket. I'll add a comment there so the guys can also include this when working on the created ticket.

MalloD12: I think it's not needed. I just found it's documented here https://docs.liquibase.com/parameters/error-on-circular-include-all.html?Highlight=errorOnCircularIncludeAll

liquibase-standard/src/test/groovy/liquibase/changelog/DatabaseChangeLogTest.groovy
```diff
@@ -558,13 +557,70 @@ http://www.liquibase.org/xml/ns/dbchangelog http://www.liquibase.org/xml/ns/dbch
...
+        null                        | null                          | "file.properties"
+        null                        | false                         | "file.properties"
+        null                        | true                          | "com/example/file.properties"
+        true                        | null                          | "file.properties"
+        true                        | false                         | "file.properties"
+        true                        | true                          | "com/example/file.properties"
+    }
+
+    @Unroll
+    def "no error is thrown when properties file is not found and is set to not error and property is not set"() {
```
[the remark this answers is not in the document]
MalloD12: yeah, it might sound a bit confusing and we can update it. I understood exactly what you described in your comment. I can rename it to "no error is thrown when properties file is not found and errorIfMissing" is either set as false or null. @jasonlyle88 do you agree? I'll update as I mentioned, but please feel free to make any other update on top of mine if you have a better name.

Thanks,
Daniel.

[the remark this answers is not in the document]
jasonlyle88: That is much clearer, thanks for the update. Amazing what makes sense when you are in the middle of something... but when you go back and read it you just question how you ever thought that was a good explanation 🤣 Thanks for the adjustment there!

liquibase-standard/src/main/java/liquibase/changelog/DatabaseChangeLog.java
XDelphiGrl: Thanks!

liquibase-standard/src/test/groovy/liquibase/changelog/DatabaseChangeLogTest.groovy
XDelphiGrl: I see; it's a global parameter and not a changeset property. My mistake. Thanks for the link to the documentation!

XDelphiGrl (approved):
This PR allows users to set the `errorIfMissing` attribute on the `includeAll` and `include` change types. The implementation aligns with the earlier PR to add `errorIfMissing` for the `property` changelog type (when `property` references a file, not inline key/value definitions). Adding `errorIfMissing="true"` to the `includeAll` or `include` change types prevents Liquibase from  halting execution if a referenced file is missing. This is useful in cases where it is expected for a particular file to be unavailable in a particular environment; for example, when there are `include`ed files that are required only in TEST but not in STAGING. 
* The default for `errorIfMissing` is `true`.
* Test harness execution passing.
* Functional test failures are due to a mismatch in Core/Pro code and tests that are ahead of that code. Failures are unrelated to this PR and can be disregarded.

In a final word, please join me on congratulating @jasonlyle88 and his wife on the birth of their new baby! @jasonlyle88 is a friendly and committed member of our Liquibase Community. That he's a new dad AND still working on OSS PRs is astounding and deserving of all the thanks in the world!

CC @rberezen

jasonlyle88:
Thank you for your kind words @XDelphiGrl! And thank you and all the other developers here for all the work you put in on this product!

PR #3907, opened 2023-03-06
Change Index.setTable method to take a Relation parameter DAT-13676

wwillard7800:
DAT-13676

## Impact

## Description

The method `Index.setTable(Table)` will not accept a `View` argument, and causes an exception during snapshot parsing.  To resolve this, we can internally change the attribute key in the in-memory model from `table` to `relation`, which causes the loading code to correctly call `setRelation(Relation)` instead.


liquibase-core/src/main/java/liquibase/structure/core/Index.java
```diff
@@ -82,8 +82,8 @@ public Table getTable() {
      * @deprecated Use {@link #setRelation(Relation)}
      */
     @Deprecated
-	public Index setTable(Table table) {
-		return setRelation(table);
+	public Index setTable(Relation table) {
```
nvoxland, line 85: I don't think this can be changed becauase it's a breaking API change?

The 3rd party source code doesn't haven't to change, but it maybe got compiled as "call for the setTable method with a Table argument" which no longer exists and the 3rd party apps have to be recompiled?

That's why we created a separate method of "setRelation" and kept this one around as deprecated.

That said, it has been deprecated for 5 years so if there is no way to fix the underlying problem differently, we could consider the breaking API change.

You could also do some testing to know for sure if it's a breaking change or not.

filipelautert (approved):

suryaaki2 (approved):

PR #3914, opened 2023-03-07
DAT-13285 - Fix CheckSum generation issues + improvement

### LINKED ISSUE #3549: SQL checksum not resilient to all formatting changes
## Description

When checksuming SQL based changesets like <sql> and <sqlFile> and <createProcedure> we run it through `liquibase.change.AbstractSQLChange.NormalizatingStream` which smooths out non-significant differences in the SQL.

For example, `create table test (id int)` and `create   table   test (id   int)` should generate the same checksum. 

That way, when editors or users reformat their SQL to make it more readable it's not going to keep them from deploying the changelog due to checksum errors. Same reason the checksum logic for other change types are not dependent on the XML formatting.

HOWEVER there are some semantically equivalent SQL which still generate different checksums.

**UPDATE**:  As part of this issue we are also going to update checksum generation logic to avoid including `contexts-filter`, `labels-filter`, and `dbms`.

## Steps To Reproduce

Run a  changelog with a <sql> tag in it, and keep running it against the same database but change the formatting of the SQL from one run to the next.

I saw that adding whitespace gets to be a problem. For example, `create table test (id int)` should be the same as `create table test ( id int )` but it was not. It should also be the same as `create table test (\nid int\n)` but was not.

Are there other ways can formatting change and still generate different checksums? Or is it just a "adding whitespace" problem?

## Additional Context

The impact to existing users need to be taken into account. I'm suspicious this will require a new checksum version, but we'll have to see what the actual required changes are.

### LINKED ISSUE #4192: Changing a property value doesn’t rerun a changeSet
### Search first

### Description

Some people have been facing the issue of a changeSet not being re-run when a property value that has a reference embedded into a SQL file has changed.  For example let's say we have below changelog file:

```
<databaseChangeLog xmlns="http://www.liquibase.org/xml/ns/dbchangelog" xmlns:ext="http://www.liquibase.org/xml/ns/dbchangelog-ext" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" xsi:schemaLocation="http://www.liquibase.org/xml/ns/dbchangelog-ext http://www.liquibase.org/xml/ns/dbchangelog/dbchangelog-ext.xsd http://www.liquibase.org/xml/ns/dbchangelog http://www.liquibase.org/xml/ns/dbchangelog/dbchangelog-latest.xsd">
    <property name="property1" value="1" global="true"/>

    <changeSet author="x" id="1" runOnChange="true">
        <sql>
            drop procedure if exists testHello();
      CREATE OR REPLACE PROCEDURE testHello()
          LANGUAGE plpgsql
          AS $$
            BEGIN
          raise notice '[${property1}]';
            END $$
        </sql>
        <rollback>
        </rollback>
    </changeSet>

    <changeSet author="x" id="2" runOnChange="true">
        <sql>
            drop procedure if exists testHello();
        </sql>
        <sqlFile path="changelog/procedure2.sql" />
    </changeSet>
</databaseChangeLog>

```
Where `procedure2.sql` content is:
```
CREATE OR REPLACE PROCEDURE testHello()
          LANGUAGE plpgsql
          AS $$
BEGIN
          raise notice '[${property1}]';
END $$
```
When the property value has been changed and the above changelog gets executed by a second time, the second changeSet (changelog.xml::2::x) does not re-run again. 

### Steps To Reproduce

1. Execute `update` command for the given changelog.
2. Update `property1` value.
3. Execute `update` command for the given changelog again.

### Expected/Desired Behavior

Both changeSets should be successfully re-run

### Liquibase Version

4.20.0

### Database Vendor & Version

PostgresSQL 14

### Liquibase Integration

CLI

### Liquibase Extensions

_No response_

_No response_

### Additional Context

None

filipelautert:
## Impact

## Description

Fixes #4192
Fixes #3549


suryaaki2 (approved):

PaulJulius:
Here's an example use case from the community forum where people are running into issues: https://forum.liquibase.org/t/changing-a-property-value-doesnt-rerun-a-changeset/8003/7

MalloD12:
@filipelautert - @suryaaki2 - @XDelphiGrl: guys, I think I'm done with the latest changes we have decided to incorporate as part of this PR (SQLFile and Change objects which can specify body as part of a SQL file). Please feel free to review this again and let me know if you have any questions/requested changes. 

Thanks,
Daniel.

liquibase-standard/src/main/java/liquibase/change/core/SQLFileChange.java
```diff
@@ -161,4 +161,25 @@ public String describe() {
                 ", relativeToChangelogFile=" + relativeToChangelogFile +
                 '}';
     }
+
+    @Override
+    public CheckSum generateCheckSum() {
```
nvoxland, line 166: The parent AbstractSQLChange class is already handing the checksum of the sql so a checksum of super.generateCheckSum plus doing this is double-computing?

Plus, I think we need excludeFromChecksum() to exclude the `path` an `relativeToChangeLogFile` properties in this file or the checksum will change if either of those change when it shouldn't be.

liquibase-standard/src/main/java/liquibase/change/AbstractChange.java
```diff
@@ -818,4 +837,29 @@ public String getDescription() {
...
         return description;
     }
+
+    /**
+     * This method will read the content of the given stream and make any parameter update on the content. For example,
+     * making a property replacement into procedure text read from the sql file.
+     * @param stream
+     * @return an updated {@link InputStream} if any replacement has been performed in the content of the original stream.
+     */
+    protected InputStream updateStreamContentIfApplicable(InputStream stream) {
```
nvoxland, line 847: I like the idea of this being a stream wrapper, but the function name is not very obvious what its for.  In thinking about a better name, I'm thinking it's maybe best to have it be a stand-alone InputStream class, like normally found? That makes it more usable for other places that may want to expand expressions on a stream beyond just a Change.

So maybe make a new `liquibase.changelog.PropertyExpandingStream` class as a wrapper around the updatedContentStream ByteArrayInputStream created in the function? Instead, make that updatedContentStream in the constructor and then have the various stream read methods call down to the updatedContentStream?

In theory it would be nice to not read the whole file into memory to do that, but that can be a later performance improvement if/when it's a problem.

nvoxland (changes requested):

liquibase-standard/src/main/java/liquibase/change/core/SQLFileChange.java
MalloD12: We have decided to update this `generateChecksum()` implementation by not having to include any checksum part of the parent because for this particular change object, we only care about the checksum generated from the SQL file content. No need to exclude any field as we are already ignoring any of the SQL file change attributes.

liquibase-standard/src/main/java/liquibase/change/AbstractChange.java
MalloD12: Done

MalloD12:
@nvoxland would you mind having a look at this when you have a chance? But it's basically what we've seen/discussed earlier.

cc: @filipelautert - @suryaaki2

MalloD12:
Adding a TEST note about the latest changes made here was an issue users reported [here](https://forum.liquibase.org/t/changing-a-property-value-doesnt-rerun-a-changeset/8003/7). The fix for this has ended up with changes on `CreateProcedure`, `CreateView` and `SQLFile`. This has been manually and unit tested by me, but if there is any other scenario anyone can think of I'm happy to add more.

Basically, the issue was property value was not being replaced in an indicated SQLFile. Here’s a changelog example of the discussed issue:

```
<databaseChangeLog
                      http://www.liquibase.org/xml/ns/dbchangelog/dbchangelog-3.8.xsd">

  <property name="property1" value="value1" global="true"/>

  <changeSet author="x" id="1" runOnChange="true">
    <sql>
      drop synonym if exists synonym1;
      CREATE SYNONYM synonym1 FOR [${property1}].[dbo].[TEST1]
    </sql>
    <rollback>
    </rollback>
  </changeSet>

  <changeSet author="x" id="2" runOnChange="true">
    <sql>
      drop synonym if exists synonym2;
    </sql>
    <sqlFile
      path="synonym2.sql"/>
    <rollback>
    </rollback>
  </changeSet>

</databaseChangeLog> 
```

Where `synonym2.sql` content file was:

`CREATE SYNONYM synonym1 FOR [${property1}].[dbo].[TEST1]`

nvoxland (approved):

Issue #3915, opened 2023-03-07
Threading and memory issues caused by static 'addedMdcEntries' map in Scope


bvremmeinfor:
### Search first

- [X] I searched and no similar issues were found

### Description

Using Liquibase 4.19.1 in a multi-threaded environment throws NullPointerExecptions like this:
```Caused by: java.lang.NullPointerException
        at liquibase.Scope.removeMdcObjectWhenScopeExits(Scope.java:427)
        at liquibase.Scope.addMdcValue(Scope.java:417)
        at liquibase.Scope.addMdcValue(Scope.java:406)
        at liquibase.executor.jvm.ChangelogJdbcMdcListener.addSqlMdc(ChangelogJdbcMdcListener.java:63)
        at liquibase.executor.jvm.ChangelogJdbcMdcListener.execute(ChangelogJdbcMdcListener.java:30)
        at liquibase.changelog.StandardChangeLogHistoryService.init(StandardChangeLogHistoryService.java:285)
        at liquibase.Liquibase.checkLiquibaseTables(Liquibase.java:2237)
        at liquibase.Liquibase$16.run(Liquibase.java:2309)
        at liquibase.Scope.lambda$child$0(Scope.java:187)
        at liquibase.Scope.child(Scope.java:196)
        at liquibase.Scope.child(Scope.java:186)
        at liquibase.Scope.child(Scope.java:165)
        at liquibase.Liquibase.runInScope(Liquibase.java:2714)
        ... 9 more
```

### Steps To Reproduce

Run a stress-test with multiple threads.
This can be used: https://github.com/bvremmeinfor/liquibase-threading-demo

### Actual Behavior

Liquibase fails with NullPointerExceptions, usually related to Scope.removeMdcObjectWhenScopeExits().

### Expected/Desired Behavior

Stable run without NullPointerExceptions

### Liquibase Version

4.19.1

### Database Vendor & Version

Any

### Additional Context

Please review changes in Scope [in this commit](https://github.com/liquibase/liquibase/pull/3574/files#diff-02cf9dc5731d4b4cab085adaefa3a0c592e2af76b14c0e0f781f4544c7153007).

* notice the static unsynchronized Map, when modified by multiple threads weird bugs will occur
* elements in map are never removed, looks like a memory leak (when used in long running JVMs)

Suggestions:
```java
  // Use ConcurrentHashMap for thread safety
  // Note: Lists in Map does not need synchronization, they belong to single Scope/thread
[...]

PR #3927, opened 2023-03-09
Clear entries from MDC map on scope exit

amrasarfeiniel:
Also make map a ConcurrentHashMap to handle multithreaded access, otherwise there may be livelock / NPE issues

## Impact

- [x ] Bug fix (non-breaking change which fixes expected existing functionality)
 
## Description

Fixes https://github.com/liquibase/liquibase/issues/3915.

Makes Scope's addedMdcEntries a ConcurrentHashMap to handle multithreaded access.
Removes the entries added to the map on scope exit to avoid leaking memory.


filipelautert:
Hello @amrasarfeiniel - I enable build and tests started to fail, so I pushed a fix to the branch (in fact it was an error with the tests, not with the fix ).

filipelautert (approved):
Fixes concurrency & cleanup objects used by MDC. 
It also fixes a case where a Scope could be created without an scopeId that was causing nullpointers on tests.
Tested using project from https://github.com/bvremmeinfor/liquibase-threading-demo and seems it solved all the issues.

StevenMassaro (approved):

XDelphiGrl (approved):
In addition to converting the MDC `HashMap`  to a `ConcurrentHashMap` to accommodate multithreading, this PR also cleans up MDC key/value pairs on scope exit.
* Fix prevents memory leaks caused by "lingering" MDC entries.
* Fix avoids thread collisions.

APPROVED

PR #3943, opened 2023-03-13
Upgrades snakeyaml for installer

filipelautert:
## Impact

## Description

Upgrade was missing for installer causing integration tests to fail.

### DIFF
--- a/liquibase-dist/src/main/maven/release.pom.xml
+++ b/liquibase-dist/src/main/maven/release.pom.xml
@@ -39,7 +39,7 @@
         <dependency>
             <groupId>org.yaml</groupId>
             <artifactId>snakeyaml</artifactId>
-            <version>1.33</version>
+            <version>2.0</version>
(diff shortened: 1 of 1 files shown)


liquibase-dist/src/main/maven/release.pom.xml
```diff
@@ -39,7 +39,7 @@
         <dependency>
             <groupId>org.yaml</groupId>
             <artifactId>snakeyaml</artifactId>
-            <version>1.33</version>
+            <version>2.0</version>
```
StevenMassaro, line 42: It would be better if we had a property in the topmost parent pom for the snakeyaml version, so we only had to change this in one place in the future.

filipelautert: That's true, this way I would not have missed this one. Maybe this should be adressed when we restructure/rename the projects?

Issue #3906, opened 2023-03-06
calling "liquibase --version" requires resolveable hostname after updating from 4.19.0 to 4.19.1


ColinMcMicken:
### Search first

- [X] I searched and no similar issues were found

### Description

Context: I maintain a minimal liquibase container image based on ubi8-micro. It is built using buildah and podman. during the container build process I verify that I can run `liquibase --version`.

Using 4.19.0 this works as it did before. When updating to 4.19.1 the `liquibase --version` command tries to resolve a temporary, random hostname, example below: `12d3de1bfed7`, in the container which is not resolvable and fails. Error and debug log provided below.

I have a work around by setting the hostname to something resolvable.

This is an unexpected change in behavior for a patch update.

### Steps To Reproduce

Run `liquibase --version` with liquibase version 4.19.1 on a system where the hostname is not resolvable.

### Actual Behavior

`+ buildah run -- <redacted container name> bash -c 'docker-entrypoint.sh --logLevel=debug --version'
[2023-03-06 12:15:31] FINE [liquibase.resource] Creating resourceAccessor for directory /liquibase
[2023-03-06 13:15:31] FINE [liquibase.configuration] No configuration value for liquibase.searchPath found
[2023-03-06 12:15:31] FINE [liquibase.resource] Creating resourceAccessor for file /liquibase/internal/lib/liquibase-core.jar
[2023-03-06 12:15:31] FINE [liquibase.resource] Creating resourceAccessor for directory /liquibase/lib
[2023-03-06 12:15:31] FINE [liquibase.resource] Creating resourceAccessor for file /liquibase/internal/lib/commons-text.jar
[2023-03-06 12:15:31] FINE [liquibase.resource] Creating resourceAccessor for file /liquibase/internal/lib/jaybird.jar
[2023-03-06 12:15:31] FINE [liquibase.resource] Creating resourceAccessor for file /liquibase/internal/lib/ojdbc8.jar
[2023-03-06 12:15:31] FINE [liquibase.resource] Creating resourceAccessor for file /liquibase/internal/lib/h2.jar
[2023-03-06 12:15:31] FINE [liquibase.resource] Creating resourceAccessor for file /liquibase/internal/lib/jaxb-api.jar
[2023-03-06 12:15:31] FINE [liquibase.resource] Creating resourceAccessor for file /liquibase/internal/lib/snowflake-jdbc.jar
[...]

PR #3960, opened 2023-03-15
[3906]  Don't break Liquibase if a resolveable hostname is not found

filipelautert:
## Impact

## Description

Fixes https://github.com/liquibase/liquibase/issues/3906

## Things to be aware of

Changed to use an Exception safe method that already exists in Liquibase Instead of using java.net directly.

## Things to worry about

Now we will always have an host name.


liquibase-core/src/main/java/liquibase/util/NetUtil.java
```diff
@@ -13,6 +13,10 @@ public class NetUtil {
     private static InetAddress localHost;
     private static String hostName;
 
+    private static final String UNKNOW_HOST_NAME = "unknown";
```
StevenMassaro, line 16: UNKNOW is spelled wrong, missing an N

StevenMassaro (approved):

liquibase-core/src/main/java/liquibase/util/NetUtil.java
filipelautert: Thanks!

suryaaki2 (approved):

XDelphiGrl (approved):
The new structured logging code (MDC) gathers information to build up industry-standard JSON objects to be used by monitoring systems. Unfortunately, the original implementation threw an error trying to resolve a hostname while gathering this information for the MDC. This PR fixes that issue.

APPROVED

PR #3963, opened 2023-03-16
DAT-12842

sayaliM0412:
## Impact

## Description

Implement - Add an entry for the liquibase-core-{version-number} in index.htm file which is under `/xml/ns/dbchangelog` for s3 CloudFront Distribution


.github/workflows/release-published.yml
```diff
@@ -261,3 +261,21 @@ jobs:
...
+          AWS_SECRET_ACCESS_KEY: ${{ secrets.LIQUIBASEORIGIN_SECRET_ACCESS_KEY }}
+          AWS_DEFAULT_REGION: us-east-1
+        # get the major.minor xsd version. grab the index.htm from s3, add the new verison of xsd and sync with the s3 again
+        run: |
+          version=${{ needs.setup.outputs.version }}
+          arr=(${version//./ })
+          xsd_version=${arr[0]}"."${arr[1]}
+          mkdir index-file
+          aws s3 cp s3://liquibaseorg-origin/xml/ns/dbchangelog/index.htm index-file
+          sed -ie "s/<\/ul>/  <li><a href=\"\/xml\/ns\/dbchangelog\/dbchangelog-${xsd_version}.xsd\">dbchangelog-${xsd_version}.xsd<\/a><\/li>\n<\/ul>/" index-file/index.htm
```
filipelautert, line 278: Besides this is the last part of the script, by just appeding at the end we are not able to rerun this job if it fails. It would keep on appending the link at each rerun.
I'm not sure if it's worth it, but maybe greping before for the version on the file and then adding it if it's not found would allow us to rerun it safely, what dfo you think?

sayaliM0412: good catch! Merci.

filipelautert (approved):

Issue #3945, opened 2023-03-13
executeCommand runs twice per changeSet


joserebelo:
### Search first

- [X] I searched and no similar issues were found

### Description

The `executeCommand` is being executed twice per changeSet, since Liquibase 4.19.1. This was working as expected on Liquibase 4.19.0.

We can see that the [ExecuteShellCommandChange](https://github.com/liquibase/liquibase/blob/63f445f85beda72e7aafc8bb18340713049ecc5c/liquibase-core/src/main/java/liquibase/change/core/ExecuteShellCommandChange.java#L122) uses the generateStatements function to execute the shell command conditionally. This seems to be now called [while reading the changesets](https://github.com/liquibase/liquibase/blob/63f445f85beda72e7aafc8bb18340713049ecc5c/liquibase-core/src/main/java/liquibase/changelog/ChangeSet.java#L714), which will effectively execute the command prematurely.

A git bisect tracked the commit that introduced this erratic behavior to 005857b1aa387896acf9a6bb12a813fbb34281a0

### Steps To Reproduce

Run the following changeSet:

```xml
<changeSet id="shell command" author="joserebelo">
    <executeCommand executable="/bin/bash">
        <arg value="-c"/>
        <arg value="echo hi >> /tmp/liquibase_shell_out"/>
    </executeCommand>
</changeSet>
```

Notice that `/tmp/liquibase_shell_out` will actually contains 2 lines, since the executeCommand change ran twice.

### Expected/Desired Behavior

The executeCommand change should only be executed once per changeSet.

### Liquibase Version

4.19.1, also confirmed on master (fa6ecb783a6931cddce5f07773f40217b6c9d7c0)

### Database Vendor & Version

PostgreSQL 9.6

### Liquibase Integration

cli

### OS and/or Infrastructure Type/Provider

Ubuntu

### Are you willing to submit a PR?

- [ ] I'm willing to submit a PR (Thank you!)

joschi:
Same issue here with Liquibase 4.19.1 and 4.20.0.

Example: https://github.com/dropwizard/dropwizard/pull/6776

filipelautert:
Hello - @wwillard7800 is working on a fix for that. Thanks for the reports!

filipelautert:
Thanks for the comments @kristofdepypere . Seems the same issue, but the PR won't fix it. We will take a look at a more generic fix in this case.

filipelautert:
[...]

PR #3967, opened 2023-03-16
Allow change to add a non-execute mode to prevent execution at certain points in the update or rollback.

wwillard7800:
## Impact

## Description

Fix an issue with `ExecuteShellCommandChange` where the command would be executed twice.  This happened 
because the code to add SQL to MDC calls the `generateStatements` method, which generates the statements and executes the command.  This will prevent the command execution.

Fixes https://github.com/liquibase/liquibase/issues/3945

### DIFF
--- a/liquibase-core/src/test/groovy/liquibase/changelog/ChangeSetTest.groovy
+++ b/liquibase-core/src/test/groovy/liquibase/changelog/ChangeSetTest.groovy
@@ -647,4 +647,22 @@ public class ChangeSetTest extends Specification {
         changeSet.execute(databaseChangeLog, new MockDatabase()) == ChangeSet.ExecType.EXECUTED
     }
 
+    def "execute does not execute shell command against when adding MDC"() {
+        when:
+        DatabaseChangeLog databaseChangeLog = new DatabaseChangeLog("com/example/test.xml")
+        def changeSet = new ChangeSet(databaseChangeLog)
+        ExecuteShellCommandChange executeShellCommandChange = new ExecuteShellCommandChange()
+        executeShellCommandChange.setExecutable("/bin/bash")
+        executeShellCommandChange.addArg("-c")
+        executeShellCommandChange.addArg("echo hi >> liquibase_shell_out")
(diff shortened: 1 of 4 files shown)


filipelautert (approved):
Thanks for fixing it @wwillard7800 !

liquibase-core/src/test/groovy/liquibase/changelog/ChangeSetTest.groovy
```diff
@@ -647,4 +647,22 @@ public class ChangeSetTest extends Specification {
         changeSet.execute(databaseChangeLog, new MockDatabase()) == ChangeSet.ExecType.EXECUTED
     }
 
+    def "execute does not execute shell command against when adding MDC"() {
```
XDelphiGrl, line 650: Let me make sure I understand this test. Here is how I'm reading it: 

* Create a changelog with a `executeShellCommandChange` that creates a file when executed.
* Call ChangeSet.execute for this changelog.
* ChangeSet.execute will _not_ run the `executeShellCommandChange` and _not_ create the file because the `allowNonExecuteMode=true` for the `ExecuteShellCommandChange`.

The test passes if the file is not created.
The test fails if the file is created.

Then, in cleanup, the test attempts to delete the file. Since the file will not exist (in the passing scenario), the File.delete will return false, but we don't care because that result is not used anywhere. 

Am I reading this correctly, @wwillard7800 ?

suryaaki2 (approved):

XDelphiGrl:
> Yes, that is it. The cleanup will be a no-op when the test passes.

Thank you!

XDelphiGrl (approved):
This PR fixes a bug where the structured logging (MDC) code improperly executed instead of building up the MDC object while running an ExecuteShellCommandChange. This is prevented by overriding the new `allowNonExecuteMode()` method to set execution mode to true; by default, this method returns false. 
* New unit test added to validate that nothing executes when ExecuteShellCommandChange runs with non-execute set to true.

APPROVED

filipelautert:
Replaced by https://github.com/liquibase/liquibase/pull/4054

PR #3978, opened 2023-03-19
Introducing removeChangeSetProperty feature 

### LINKED ISSUE #3091: restore best-effort behavior of afterColumn
## Environment

**Liquibase Version**: 4.13.0

**Liquibase Integration & Version**: gradle, spring boot

**Database Vendor & Version**: mysql, postgresql

**Infrastructure Type/Provider**:

## Description

https://github.com/spinnaker/orca (migrations are [here](https://github.com/spinnaker/orca/tree/master/orca-sql/src/main/resources/db/changelog)) uses `afterColumn` in some of its migrations, and tests those migrations using both mysql and postgresql.  Before version 4.13.0, and specifically https://github.com/liquibase/liquibase/pull/2943, afterColumn worked with mysql, and did nothing with postgresql.  With 4.13.0, afterColumn still works with mysql, but throws a validation error with postgresql.  For example:
```
PgTopApplicationExecutionCleanupPollingNotificationAgentSpec > initializationError FAILED
    com.netflix.spinnaker.kork.sql.test.SqlTestUtil$DatabaseInitializationFailed: liquibase.exception.ValidationFailedException: Validation Failed:
         4 changes have validation failures
              addAfterColumn is not allowed on postgresql, db/changelog/20180510-add-legacy-id-fields.yml::add-legacy-id-fields::cthielen
              addAfterColumn is not allowed on postgresql, db/changelog/20180510-add-legacy-id-fields.yml::add-legacy-id-fields::cthielen
              addAfterColumn is not allowed on postgresql, db/changelog/20180510-add-legacy-id-fields.yml::add-legacy-id-fields::cthielen
              addAfterColumn is not allowed on postgresql, db/changelog/20180510-add-legacy-id-fields.yml::add-legacy-id-fields::cthielen
        at com.netflix.spinnaker.kork.sql.test.SqlTestUtil.initDatabase(SqlTestUtil.java:248)
        at com.netflix.spinnaker.kork.sql.test.SqlTestUtil.initTcPostgresDatabase(SqlTestUtil.java:97)
        at com.netflix.spinnaker.orca.sql.cleanup.PgTopApplicationExecutionCleanupPollingNotificationAgentSpec.getDatabase(TopApplicationExecutionCleanupPollingNotificationAgentSpec.groovy:161)

        Caused by:
        liquibase.exception.ValidationFailedException: Validation Failed:
             4 changes have validation failures
                  addAfterColumn is not allowed on postgresql, db/changelog/20180510-add-legacy-id-fields.yml::add-legacy-id-fields::cthielen
                  addAfterColumn is not allowed on postgresql, db/changelog/20180510-add-legacy-id-fields.yml::add-legacy-id-fields::cthielen
                  addAfterColumn is not allowed on postgresql, db/changelog/20180510-add-legacy-id-fields.yml::add-legacy-id-fields::cthielen
                  addAfterColumn is not allowed on postgresql, db/changelog/20180510-add-legacy-id-fields.yml::add-legacy-id-fields::cthielen
            at liquibase.changelog.DatabaseChangeLog.validate(DatabaseChangeLog.java:307)
            at liquibase.Liquibase.lambda$update$1(Liquibase.java:228)
            at liquibase.Scope.lambda$child$0(Scope.java:180)
            at liquibase.Scope.child(Scope.java:189)
            at liquibase.Liquibase.runInScope(Liquibase.java:2400)
            at liquibase.Liquibase.update(Liquibase.java:208)
            at liquibase.Liquibase.update(Liquibase.java:194)
            at liquibase.Liquibase.update(Liquibase.java:190)
            at liquibase.Liquibase.update(Liquibase.java:182)
            at com.netflix.spinnaker.kork.sql.test.SqlTestUtil.initDatabase(SqlTestUtil.java:246)
            ... 2 more
```
One use of `afterColumn` is:
```
    - addColumn:
        tableName: orchestration_stages
        columns:
        - column:
            name: legacy_id
            type: varchar(500)
            afterColumn: id
```

## Expected/Desired Behavior

https://github.com/liquibase/liquibase/pull/2943 notes that this is the new behavior.  I'd love a way to go back to the old behavior, where addColumn works where supported, and is ignored where it's not.  There are some migrations where adding e.g.:
```
modifySql:
  dbms: "mysql"
  append:
    value: " after id"
```
works, but many where it doesn't.

### LINKED ISSUE #3195: Before/After/At change prevents adding a column in `FIRST` position on MySQL.
@nvoxland I think your change has prevented me from adding a column in `FIRST` position on MySQL. Previously this appeared to be possible by using `beforeColumn` although I appreciate that this was likely not working. I think there should be a way to specify the first column as that is supported - maybe using `position=0`?

_Originally posted by @mjaggard in https://github.com/liquibase/liquibase/issues/2943#issuecomment-1224219622_

kirangodishala:
## Impact

## Description
Introducing `removeChangeSetProperty` feature and AddColumnChangeVisitor implementation to remove afterColumn, beforeColumn or position if the db doesn't support. 

An example changelog in yaml format:
```yaml
databaseChangeLog:
- removeChangeSetProperty:
    change: addColumn
    dbms: postgresql
    remove: afterColumn    
- changeSet:
    id: add-column-example
    author: Kiran
    changes:
    - addColumn:
        tableName: example
        columns:
        - column:
            name: new_col
            type: varchar(50)
            afterColumn: id
    rollback:
    - dropColumn:
        tableName: example
        columnName: new_col
``` 
While loading the changelog, a `ChangeVisitor` (`AddColumnChangeVisitor`) object is created and loaded  but it gets added to changelog object's changeVisitors list only if the running database matches the dbms value. Then while loading the a change, modify(changeVisitor) method is called for all the applicable changeVisitors which applies the given modication. 

Fixes #3091 
Fixes #3195
  
## Things to be aware of
Introduction of modifyChange should take place before changeSets as the ChangeVisitor needs to be available while ChangeSets are being loaded.


filipelautert:
Hello @kirangodishala, thanks for the feature!
I'm adding the safe to build tag so we can see it build and test. Nathan may take a better look at you PR later this/next week and move it ahead for further reviews.

kirangodishala:
@nvoxland - any comments on the PR so far?

nvoxland:
Sorry for the delay, @kirangodishala. I've started to look at it a couple times over the last couple weeks but didn't get my thoughts fully collected.

The main thing is that I think we should switch around where the logic lives. You have a `modify` function that got added to the Change interface and which you implemented in AddColumnChange and will have to be similarly implemented in every change. You have the visitor class with "AddColumnChangeVistor" as an example, but I think that can also be more generic so we don't need a new class for every type of attribute that people might want to modify.

Instead, are you able to have a more generic `AddAttributeChangeVisitor` and `RemoveAttributeChangeVisitor` that take the attribute name and will add/remove whatever that given attribute name is? The Change objects have `set(String, Object)` method which should be able to modify any attribute name on any change type. 

Then, the logic for how to do that add/remove should live only in the ChangeVisitor implementations without adding the new `modify()` method to Change. Instead, the changeVisitor.modify(change) function will modify the change object. I don't think you need the clone() logic either, then, since the point where we'd call that modify() function will be when we've built up a specific change instance and I think we'd be OK directly changing it than dealing with a dupliate.

Think that would work?

kirangodishala:
@nvoxland - thanks for the feedback. I will get back to you.

nvoxland:
Have you gotten a chance to look at this more, @kirangodishala ?

kirangodishala:
@nvoxland - I will give an update by Monday.

kirangodishala:
@nvoxland - sorry for the delay. I have gone through your suggestions and looked into the code for the feasibility but I don't think the removal(setting the value to null) of `afterColumn/beforeColumn/position` can be done using `set(String, Object)`, simply because they are not `attributes` of the change object but the fields under AddColumnConfig. 
image (1)
image

`AddAttributeChangeVisitor` and `RemoveAttributeChangeVisitor` can be added to deal with attributes but they don't make intended changes on the `AddColumnConfig`

But the logic can be moved from `change.modify()` to `changeVisitor.modify()` (in this case `AddColumnChangeVisitor.modify()`) so that `Change` need not be modified

nvoxland:
Thanks for looking into it a bit more, @kirangodishala

I'll move it along the process to where we do a bit deeper of a review and can help think/work through implementation options better. I think we'll want to make at least some of the logic moving changes, but @filipelautert or whoever looks into it a bit more deeply can know better what to suggest.

kirangodishala:
@nvoxland - I would like to check with you if there is any further update on this.

MalloD12:
Hi Kiran (@kirangodishala), I have been looking into this PR but I think I still don't have a complete understanding of this change.  So I would like to ask you a few questions which might help me to move forward with the review:

- What is the idea behind this change or what do we expect with it? Using your given example we know Postgres does not support these attributes (before, after, or position) so if we have a changelog like this what will happen is this attribute will be set as null, but what do we expect to happen in the DB? Should the given column be still deployed and added to the table? I can see a column is added to the table always at the last position, but just want to make sure that's the intended behaviour.
- Do we expect this change only be available for changelos in a YAML format? If not I think we should probably add the `modifyChange` tag into the `dbchangelog-latest.xsd` file?

MalloD12:
> @MalloD12 - Thank you for looking into this PR.
> 
> > * What is the idea behind this change or what do we expect with it? Using your given example we know Postgres does not support these attributes (before, after, or position) so if we have a changelog like this what will happen is this attribute will be set as null, but what do we expect to happen in the DB? Should the given column be still deployed and added to the table? I can see a column is added to the table always at the last position, but just want to make sure that's the intended behaviour.
> 
> Yes. Since the feature is not available in Postgres we expect those attributes to be ignored and column be added at the last position without throwing any error.
> 
> > * Do we expect this change only be available for changelos in a YAML format? If not I think we should probably add the `modifyChange` tag into the `dbchangelog-latest.xsd` file?
> 
> I believe any change has to be compliant with any format. But I didn't look into the internals of xml processing logic as we are primarily interested in yaml. To keep it consistent with all formats, if it requires a change in xsd, yes it has to be added.

Hi @kirangodishala,

I appreciate your answers. I have a few questions and comments.  

- As of now with this change, we are only allowed to apply `modifyChange` function to only one DB per time, I was thinking it would be nice to apply the same **dbms** behaviour we do for changesets and some change objects which allows specifying a list of them.

- I think it would be nice to add the necessary changes to have this also available in XML format for which you might need to enter this new function in `dbchangelog-latest.xsd`.

-  Lastly, I think there is one more scenario we can add to AddColumnChangeTest:
- -  Validate that a new column is added to the last position of the table.
.
Thanks,
Daniel.

MalloD12:
Hi @kirangodishala, have you had a chance to follow up on my comments?
 
Thanks,
Daniel.

kirangodishala:
@MalloD12 - yeah, I saw your comments. I will get back to you in a couple of days.

kirangodishala:
> * As of now with this change, we are only allowed to apply `modifyChange` function to only one DB per time, I was thinking it would be nice to apply the same **dbms** behaviour we do for changesets and some change objects which allows specifying a list of them.

@MalloD12 - there was a long discussion on the changes and finally this design was agreed. You can find the details here: https://github.com/liquibase/liquibase/issues/3091

If you still suggest a change in design, I would need more details on how to handle all the scenarios because right now I tried to include the code which has minimal or no impact on the existing code.

kirangodishala:
> * I think it would be nice to add the necessary changes to have this also available in XML format for which you might need to enter this new function in `dbchangelog-latest.xsd`.

Sure. I guess any change should support all the formats. So if the changes are accepted, then I can go for xsd changes but I would need some guidance.

kirangodishala:
> * Validate that a new column is added to the last position of the table.

Sure, please confirm if the changes are fine otherwise(except xsd change)

MalloD12:
> > * As of now with this change, we are only allowed to apply `modifyChange` function to only one DB per time, I was thinking it would be nice to apply the same **dbms** behaviour we do for changesets and some change objects which allows specifying a list of them.
> 
> @MalloD12 - there was a long discussion on the changes and finally this design was agreed. You can find the details here: #3091
> 
> If you still suggest a change in design, I would need more details on how to handle all the scenarios because right now I tried to include the code which has minimal or no impact on the existing code.

Yeah, I have been following up on that discussion. I think it's a good idea to keep the code with minimal or no impact on the existing code, but I think it would be good if we can make this consistent with how we are using the `dmbs` property in other parts (changeset, change objects, etc). I think the change would be pretty simple by accepting/returning a `String` array instead of a single value, and also it would require changing the below if condition from `DatabaseChangeLog` class:
```
if(changeVisitor.getDbms().equalsIgnoreCase(changeLogParameters.getDatabase())) {
...
}
```
For something like:
```
if(DatabaseList.definitionMatches(changeVisitor.getDbms(), changeLogParameters.getDatabase(), false)) {
...
}

I think this way, the same changelog could be used in multiple platforms and we would only need to add the DB short names into the `dbms` property of the `modifyChange` section, instead of adding multiple `modifyChange` requests with only a different dbms. We have been discussing this with @filipelautert and we think it would be nice to make this change unless there is a reason we might not be aware of.

**Comment by MalloD12** [comment]
> > * I think it would be nice to add the necessary changes to have this also available in XML format for which you might need to enter this new function in `dbchangelog-latest.xsd`.
> 
> Sure. I guess any change should support all the formats. So if the changes are accepted, then I can go for xsd changes but I would need some guidance.

For including this into XML format you might need to add something like this inside of the `databaseChangeLog` element:
```
<xsd:element name="databaseChangeLog">
      <xsd:complexType>
          <xsd:sequence>
                
      <xsd:element name="modifyChange">
              <xsd:complexType>
                  <xsd:attribute name="change" type="xsd:string"/>
                  <xsd:attribute name="dbms" type="xsd:string"/> 
                  <xsd:attribute name="remove" type="xsd:string"/>           
              </xsd:complexType>
          </xsd:element>
```

@filipelautert please correct me if I'm wrong.

**Comment by MalloD12** [comment]
@kirangodishala I think if you can apply mentioned comments we would be ready to approve this PR and move forward with it. If you have any questions or concerns please share it with us, and again we appreciate you taking care of this change. 

Cheers,
Daniel.

**Comment by kirangodishala** [comment]
Thank you @MalloD12. I will work on the changes and get back to you if I have any questions.

**Comment by MalloD12** [comment]
> Thank you @MalloD12. I will work on the changes and get back to you if I have any questions.

Thank you, Kiran. I appreciate it!

**Comment by MalloD12** [comment]
Hi @kirangodishala, 

I hope you are doing very well,  I appreciate you pushing the suggested changes. The look good to me. I would like to ask you for a few more things I have been thinking about after reviewing this PR again:

- Rename this new feature from `modifyChange` to something like `removeChangeSetProperty` (or maybe any other name you can think of). I thought `modifyChange` could be confusing for some people when trying to use `modifyChangeSets` or vice versa. 

- Adding one more scenario we can add to AddColumnChangeTest:

- - Validate that a new column is added to the last position of the table. 

- Adding an integration test with this new functionality:
-  - This would be pretty much adding a `modifyChange` tag (or whatever it's finally called) in the `common.tests.changelog.xml`:
```
 <changeSet id="addColumn" author="mallod">
        <addColumn tableName="myTable">
            <column name="new_column" type="varchar(50)" afterColumn="test_col">
            </column>
        </addColumn>
    </changeSet>
``` 
and 
```
<modifyChange dbms="mysql, postgresql" change="addColumn" remove="afterColumn" />
```
at the beginning of the changelog. This way we add an extra check to make sure this new feature is run successfully (or not) in any DB regardless it supports or not this type of attribute.

I can implement all or any of these changes if you are ok with that, or you can do it if it's your preference. 

Thanks,
Daniel.

kirangodishala:
Hi @MalloD12 - I made all the changes you suggested -
- renamed `modifyChange` node to `removeChangeSetProperty`
- added integration test to verify removeChangeSetProperty
- added additional unit test to check if the column is added at the last position when removeChangeSetProperty takes effect

If you think of any more changes, I am okay for you taking over from here as all we(Spinnaker community) want is to push these changes at the earliest. If someone from liquibase is offering to expedite we are more than happy to handover the code :-)

MalloD12:
> Hi @MalloD12 - I made all the changes you suggested -
> 
> * renamed `modifyChange` node to `removeChangeSetProperty`
> * added integration test to verify removeChangeSetProperty
> * added additional unit test to check if the column is added at the last position when removeChangeSetProperty takes effect
> 
> If you think of any more changes, I am okay for you taking over from here as all we(Spinnaker community) want is to push these changes at the earliest. If someone from liquibase is offering to expedite we are more than happy to handover the code :-)

Hi @kirangodishala,

I appreciate you adding these new suggested changes to the PR. I'll review them in a bit and I'll let you know, but yeah I totally agree and fine with us taking over any new additional change I or someone from the team can suggest. 

Thanks,
Daniel.

MalloD12 (approved):
**Approved**.

Code changes looks good to me . This new functionality provides support to avoid getting a validation error when a given DB does not support adding a new column at a requested position (`beforeColum`/ `afterColumn` / `positionAt`). Execution of this new change type will make set the given attribute as `NULL` and add the new column at the end of the given table.

I manually test this change and it work as expected for both a DB which supports these type of properties and for DB that it doesn't. This PR contains a good amount of test scenarios (unit and integration tests).

Build and test checks have been successfully executed (except for Functional test for which failing tests seems to not be related with these changes. Same failures have been seen in a different PR).

### Things to be aware of:
- As mentioned this will work and behave in the same way either for a DB which supports properties like `beforeColum`/ `afterColumn` / `positionAt`.

### Things to worry about:
- None

MalloD12:
> **Approved**.
> 
> ## Review and testing results:
> Code changes looks good to me . This new functionality provides support to avoid getting a validation error when a given DB does not support adding a new column at a requested position (`beforeColum`/ `afterColumn` / `positionAt`). Execution of this new change type will make set the given attribute as `NULL` and add the new column at the end of the given table.
> 
> I manually test this change and it works as expected for both a DB that supports these types of properties and for DB that it doesn't. This PR contains a good amount of test scenarios (unit and integration tests).
> 
> Build and test checks have been successfully executed (except for the functional tests for which failing tests seem to not be related to these changes. Same failures have been seen in a different PR).
> 
> ### Things to be aware of:
> * As mentioned this will work and behave in the same way either for a DB that supports properties like `beforeColum`/ `afterColumn` / `positionAt`.
> 
> ### Things to worry about:
> * None


Again, thank you @kirangodishala for taking care of this fix!

filipelautert (approved):
Thanks for the new feature @kirangodishala !

This PR needs to be merged in a major (4.24.0) release as it changes CSD and introduces new public methods.

rberezen:
This pull request introduces the removeChangeSetProperty property, which allows the removal of attributes such as afterColumn, beforeColumn, or position for the addColumn Change Type.

- New unit tests added.
- Manual testing passed.
- Ticket to cover documentation created.

APPROVED

rberezen (approved):

PR #4005, opened 2023-03-22
DAT-13447

sayaliM0412:
## Impact

## Description

https://datical.atlassian.net/browse/DAT-13447

`liquibase-dist` module in `.github/workflow/build.yml` generates 2 things:

- a `.jar` file and a `.tar.gz` file.
- This PR helps build only `.tar.gz` file.

@StevenMassaro @suryaaki :  I think these are the changes to be made. But please let me know if these changes look good. I checked in the GHA [here](https://github.com/liquibase/liquibase/actions/runs/4493599239/jobs/7905201415#step:9:570) to check for the jars being built.


.github/util/re-version.sh
```diff
@@ -133,13 +133,13 @@ done
 
 ##### update zip/tar files
 mkdir -p $workdir/internal/lib
-cp $outdir/liquibase-core-$version.jar $workdir/internal/lib/liquibase-core.jar ##save versioned jar as unversioned to include in zip/tar
+# cp $outdir/liquibase-core-$version.jar $workdir/internal/lib/liquibase-core.jar ##save versioned jar as unversioned to include in zip/tar
```
filipelautert, line 136: ahhhh git is your friend, remove it 🤣

filipelautert (approved):

StevenMassaro (commented):
I agree with Filipe, remove these lines, not comment them. Also, there should be modifications to the pom of liquibase-dist so that it is not generating these unneeded artifacts.

PR #4021, opened 2023-03-25
Fix search path handle for PostgreSQL with JDBC connection with specified schema

### LINKED ISSUE #4014: Unable to migrate database on Postgres with a schema containing dashes
### Search first

### Description

In our environment, we use postgres databases containing dashes (name originates from git branches).
We found out that when running Liquibase on a Spring application hosted in JBoss that our deployment got stuck because it was failing to acquire a lock, after further debugging we found out that Liquibase is unable to handle schema names with dashes in them.

I have setup an example project which demonstrates the issue here: https://github.com/Thodor12/LiquibaseTest

Schema names with special characters aren't anything out of the ordinary, but they must be wrapped in double quotes in order for the query to be valid.

### Steps To Reproduce

- Checkout the application provided in the description
- Ensure you have Docker installed so you can create the database using compose (alternatively set it up with the identical configuration as in the project)
- Start the application
- Error appears on boot:

```
Caused by: liquibase.exception.LockException: liquibase.exception.DatabaseException: ERROR: syntax error at or near "-"
```

We solved the issue in our project by stripping every non-alphanumeric character from the schema in order to make it functional.
If this is the intended solution, is this mentioned anywhere in the documentation?

### Expected/Desired Behavior

The application can start fine and no errors appear

### Liquibase Version

4.19.0 (in project), 4.17.2 (in example)

### Database Vendor & Version

PostgreSQL 14.4

### Liquibase Integration

gradle

### Liquibase Extensions

N/A

Windows 10 / CentOS 7.9

### Additional Context

N/A

quonas:
## Impact

## Description
Fixed #4014
When using Postgres JDBC connect with  specificed schema, `SHOW SEARCH_PATH` will return the specific schema , not like the default case  `"$user",public`.  Therefore, this PR is intended to handle this particular scenario.


nvoxland:
Initial review:

Code makes sense off hand, thanks for the fix @quonas !

Things to be aware of:
- Only impacts postgresql (and related databases)

Things to worry about:
- Nothing

filipelautert (approved):
Fix schema handling when an url defining the default schema is used, such as url=jdbc:postgresql://localhost:5434/postgres?currentSchema=bt-1-aa .
This does not impact usage of "defaultSchemaName" property used by liquibase.
Thanks for the fix @quonas !

suryaaki2 (approved):

XDelphiGrl:
@filipelautert, hi! There are a couple of integration tests failing for this PR. Can you please take a look?

[https://github.com/liquibase/liquibase/actions/runs/4853381427](https://github.com/liquibase/liquibase/actions/runs/4853381427)

XDelphiGrl (approved):
Postgres supports schema names with dashes but Liquibase was throwing errors when the JDBC URL specified `currentSchema=schema-dash`. This PR fixes that bug. Please note that in the case of this PR, `search path` is the _Postgres_ search path, not the Liquibase `--search-path` property.
* Manual test of the `history` command validates the fix.

APPROVED

Issue #3289, opened 2022-09-19
Liquibase 4.8 breaks H2 1.x compatibility


consultantleon:
Since upgrading to liquibase 4.8.0 our application fails to create the schema and load data successfully regarding a table (named 'SETTINGS') which has a column named "Value".

Before with liquibase 4.5 "Value" was not marked as a reserved word and during creation of the table the column was not quoted so it'd be created by default in Uppercase.

Since liquibase 4.8.0 the column name is seen as a reserved word and the column name is quoted during creation. So not this column gets created in actual mixed-case "Value" (instead of VALUE).

Besides ending up with a table with one column 'uppercased' and the other 'case preserved' (NAME,Value) , data loading and our application code fails:
-  loadData fails from csv file with column names in uppercase:
```
"NAME","VALUE"
"passwords.enableComplexity","Medium"
"passwords.enableAge","false"
"passwords.enableHistory","true"
```

[ so column names in loadData are not checked for reserved words and quote automatically as for other statements...? ]

Our application code fails doing a select Value from SETTINGS where Name='xxx' as the application code is not designed to quote column names as liquibase 4.8 now started to do for the "Value" column ...

Why not keep a set of V1 keywords (keywordsV1 ) and keywords (V2) and check for the major version number as is done elsewhere in the H2Database class?

e.g. here:

```
@Override
    protected String getAutoIncrementClause() {
        try {
            if (getDatabaseMajorVersion() == 1) {
```

consultantleon:
Created a workaround, H2V1Database class which due to a higher priority supersedes H2Database once installed

Issue #2385, opened 2022-01-19
Liquibase 4.7.0 upgrade causing issue with table name


marceloverdijk:
I have a hobby project running on Spring Boot 2.7 snapshot which today upgraded from Liquibase 4.6.2 to 4.7.0.

With Liquibase 4.6.2 and H2 datasource the following changelog

```
databaseChangeLog:
  - changeSet:
      id: 1
      author: me
      changes:
        - createTable:
            tableName: user
            columns:
              - column:
                  name: id
                  type: int
                  autoIncrement: true
                  constraints:
                    nullable: false
                    primaryKey: true
                    primaryKeyName: user_pk
              - ..
```

will create UPPERCASE tables in H2:

![image](https://user-images.githubusercontent.com/197004/150146687-13c81f5a-ef47-4191-bc26-c7ce1a8325fc.png)

In fact I never really paid attention to the uppercase/lowercase table/column name in the H2 console despite having them lowercase in the changelog.

Also a changelog like this (lowercase names) works:

```
        - insert:
            tableName: user_role
            columns:
              - column:
                  name: user_id
                  valueComputed: select id from user where username = 'foo_bar'
              - column:
                  name: role_name
                  value: ROLE_USER
```

Doing some manual queries in the H2 console work for these:

- `SELECT * FROM USER`
- `SELECT * FROM user`
- `SELECT * FROM "USER"`

but not for `SELECT * FROM "user"`

Now with upgrade to Liquibase 4.7.0 I see other behaviour with above setup.

The application won't start as the changelog with the `insert` fails now:

```
Reason: liquibase.exception.DatabaseException: Table "USER" not found; SQL statement:
INSERT INTO PUBLIC.user_role (user_id, role_name) VALUES (select id from user where username = 'foo_bar', 'ROLE_USER') [42102-200] [Failed SQL: (42102) INSERT INTO PUBLIC.user_role (user_id, role_name) VALUES (select id from user where username = 'foo_bar, 'ROLE_USER')]
```

Although "USER" (uppercase) is not used like that in the insert it still fails.

Now I probably should have not used the table name `user` as `USER` is a reserved word, so I can only blame myself ;-) but this might impact others as well...

In the Liquibase 4.7.0 [release notes](https://docs.liquibase.com/release-notes/home.html) I don't see anything noteworthy that changed that could explain this behaviour.

Is this unexpected and a bug or expected behaviour?

kataggart:
[...]

PR #4024, opened 2023-03-27
Gently introduce JUnit 5

mches:
## Impact

## Description

Introduced JUnit 5 `test` dependency migrated JUnit 4 tests in the `liquibase.database`, `liquibase.database.core` packages to JUnit 5.

## Things to be aware of

Replaced JUnit 4 assertions with JUnit 5 and Hamcrest assertions rather than going all out with AssertJ assertions. Converted some tests to parameterized tests where it made sense.

## Things to worry about

N/A

## Additional Context

I have some ideas for fixing #2385, #3289. I thought it would be nice if the baseline for the test cases involved was JUnit 5.

### DIFF
--- a/liquibase-core/src/test/java/liquibase/database/AbstractJdbcDatabaseTest.java
+++ b/liquibase-core/src/test/java/liquibase/database/AbstractJdbcDatabaseTest.java
@@ -1,11 +1,11 @@
 package liquibase.database;
 
-import static org.junit.Assert.assertEquals;
-import static org.junit.Assert.assertFalse;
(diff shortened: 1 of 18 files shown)


mches:
On a second look, it seems JUnit 5 is already a dependency, so my branch name doesn't make sense. I'll have to submit a new PR to change the branch name. Sorry for the clutter. 🙇

filipelautert:
Hey @mches , no problems. We don't follow strict patterns for branch names (but trying to avoid names like "patch1" is good) .  Most important is the PR title, but that can be changed later.

PR #4032, opened 2023-03-27
Add mirror-console-messages-to-log parameter (DAT-13802)

abrackx:
## Impact

## Description
Adds new `mirror-console-messages-to-log` global parameter, **defaults to true**. Used to mirror console messages to logs. Handled using new `CompositeUIService` class which delegates to all provided output services and requires one input service to construct. Adds test.

## Things to be aware of
Mirrors all console messages to logs. This means if you set `--log-level` in your console you will see duplicated console messages (printed once to the console, once to the logs) unless you are also utilizing `--log-file`. You can disable this behavior by setting `--mirror-console-messages-to-log false`.


StevenMassaro (approved):
Not a huge fan of the name of the new argument, but implementation looks solid.

suryaaki2 (approved):

filipelautert (approved):

nvoxland:
Yeah, I'm not a fan of the new argument either. It seems to me to be yet another pointer that the whole UI vs. log separation is wrong and needs to be cleaned up instead of just continued to work around like we've been talking about for years. While that's a non-trivial refactoring, it's also not terribly difficult, so I'd suggest considering that larger change vs. adding a flag we'll have to continue to support if/when we make that larger change. Better to just spend the little more time to fix it right vs. continuing to patch the symptoms?

Implementation-wise, rather than adding more logic into ConsoleUIService and making it do multiple things, maybe it's better to instead add a CompositeUIService class, and then if the flag is set, set the UIService to be that CompositeUIService contining the ConsoleUIService plus the existing LoggerUIService? The more we can keep classes more single-purpose and instead rely on composition, the more it tends to be a more maintainable API for people I think.

StevenMassaro (approved):

suryaaki2 (approved):

PR #4043, opened 2023-03-28
Fix updateTesting rollback replacing rollbackToDate by rollbackCount

filipelautert:
## Impact

## Description

Fix updateTesting rollback: instead of using the date to mark where to rollback, uses the count of the number of changes executed to rollback them.

## Things to be aware of

For some databases like Snowflake, when the database is in a different timezone than the host running liquibase the time returned is different from the local time, thus causing updateTestingRollback to fail.

## Things to worry about

All updateTestingRollback  should keep on working with this new approach.

### DIFF
--- a/liquibase-core/src/main/java/liquibase/Liquibase.java
+++ b/liquibase-core/src/main/java/liquibase/Liquibase.java
@@ -1523,6 +1523,7 @@ public class Liquibase implements AutoCloseable {
         update(tag, contexts, labelExpression);
         changeLogService.reset();
         int changesetsToRollback = changeLogService.getRanChangeSets().size() - originalSize;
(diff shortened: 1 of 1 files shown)


suryaaki2:
@filipelautert Can we write tests?

suryaaki2 (approved):

awhitford-cip:
Thanks for addressing this!

Does this approach behave as expected when changesets have _preconditions_ that evaluate to `false`?

filipelautert:
Hello @awhitford-cip - we just changed the rollbackToDate by a rollbackCount in order to fix it - everything else should behave as before. Internally, "updateTestingRolback" is just chaining 3 commands as if they were run manually:
- update
- rollbackCount
- update

PR #4045, opened 2023-03-28
Handle MDC log level correctly DAT-13892

wwillard7800:
## Impact

## Description
Handle MDC log level correctly


liquibase-cli/src/main/java/liquibase/integration/commandline/LiquibaseCommandLine.java
```diff
@@ -296,6 +298,22 @@ private int handleException(Throwable exception) {
...
+    private Level determineLogLevel(Throwable throwable) {
+        if (throwable == null) {
+            return Level.SEVERE;
+        }
+        Level returnLevel = Level.SEVERE;
+        Throwable t = throwable;
+        while (t != null) {
+            if (t instanceof LiquibaseException) {
+                LiquibaseException le = (LiquibaseException) t;
+                returnLevel = le.getLogLevel();
```
filipelautert, line 310: I believe it should be set to returnLevel only if le.getLogLevel() is not null . Otherwise you may return null at the end and then set the log level as null at LiquibaseCommandLine

wwillard7800: You are exactly right.  I wrote this code the right way before and removed it, and then when I brought it back I fouled this up.  Thanks!

filipelautert (approved):

PR #4054, opened 2023-03-30
Do not allow custom change types to execute twice DAT-14051

### LINKED ISSUE #3945: executeCommand runs twice per changeSet

wwillard7800:
## Impact

## Description
The addition of context data to logging was causing custom changes to be executed twice.  These changes prevent that.

Fixes #3945


liquibase-core/src/main/java/liquibase/change/Change.java
```diff
@@ -140,4 +142,16 @@ public interface Change extends LiquibaseSerializable, Plugin, ExtensibleObject
...
+
+    /**
+     *
+     * Override this to allow this instance be in non-execute mode
+     * Default return is false
+     *
+     * @return   boolean       True if non-execute mode
+     *
+     */
+    default boolean allowNonExecuteMode() {
```
filipelautert, line 154: This method is not used anymore right?

liquibase-core/src/main/java/liquibase/change/core/ExecuteShellCommandChange.java
```diff
@@ -378,6 +370,11 @@ protected void customLoadLogic(ParsedNode parsedNode, ResourceAccessor resourceA
         }
     }
 
+    @Override
```
filipelautert, line 373: Here too?

liquibase-core/src/main/java/liquibase/change/core/ExecuteShellCommandChange.java
```diff
@@ -154,12 +146,12 @@ public Sql[] generate(Database database) {
                         throw new UnexpectedLiquibaseException("Error executing command: " + e.getLocalizedMessage(), e);
                     }
 
-                    return null;
+                    return new Sql[0];
```
filipelautert, line 149: Is this change is expected?

liquibase-core/src/main/java/liquibase/change/Change.java
wwillard7800: Correct.  I need to remove it.

liquibase-core/src/main/java/liquibase/change/core/ExecuteShellCommandChange.java
wwillard7800: Same

liquibase-core/src/main/java/liquibase/change/core/ExecuteShellCommandChange.java
wwillard7800: I followed a suggestion made by Sonar, but I probably shouldn't have made that change.  I will revert it, just to be safe.

XDelphiGrl:
Functional Tests Started Here: [https://github.com/liquibase/liquibase-pro-tests/actions/runs/4611625991](https://github.com/liquibase/liquibase-pro-tests/actions/runs/4611625991)

Test Harness Execution Started Here: [https://github.com/liquibase/liquibase-test-harness/actions/runs/4611629843](https://github.com/liquibase/liquibase-test-harness/actions/runs/4611629843)

tarioch:
Is it possible that this broke the SQL Generation mode for custom sql changes? I'm still trying to track this down, but it looks like since this change we're no longer getting the SQL generated for the custom changes.

PR #4057, opened 2023-03-30
Fix release lock exception

wwillard7800:
## Impact

## Description
Running `update*Sql` commands against an empty database caused an exception to be thrown because there was no DBCL lock table.  Overriding the new `cleanUp` method in these CommandStep classes is the fix.


wwillard7800:
This branch is no longer needed.

PR #4065, opened 2023-04-03
Reworks liquibaseOperation MdcKey to liquibaseInternalCommand (DAT-13996]

abrackx:
## Impact

## Description
Sets up keys so that `liquibaseCommand` is the command entered to the console and `liquibaseInternalCommand` is the current command being ran by liquibase (which can be internal commands).

## Things to be aware of
Adds new `isMdcKeyPresent` beta method to check for existing mdc keys to avoid any unintentional overriding due to multiple CommandScopes.


liquibase-core/src/main/java/liquibase/Liquibase.java
StevenMassaro, line 1: I think there are constant fields with the command names in them somewhere, can you use those instead of hardcoding the string?

liquibase-core/src/main/java/liquibase/integration/commandline/Main.java
StevenMassaro, line 1: Can you replace usages of `internalCommand` with their `COMMAND_NAME` from the respective class, rather than hardcoding?

StevenMassaro (changes requested):

liquibase-core/src/main/java/liquibase/integration/commandline/Main.java
abrackx: These are internal commands from pro, so I can't take that route

liquibase-core/src/main/java/liquibase/Liquibase.java
StevenMassaro: You made the right call, don't add them.

StevenMassaro (approved):

filipelautert (approved):

filipelautert (approved):

PR #4075, opened 2023-04-03
[DAT-6629] Refactoring of rollbackToDate[SQL] command.

filipelautert:
## Impact

### DIFF
--- a/liquibase-core/src/main/java/liquibase/command/core/AbstractRollbackCommandStep.java
+++ b/liquibase-core/src/main/java/liquibase/command/core/AbstractRollbackCommandStep.java
@@ -1,0 +1,213 @@
+package liquibase.command.core;
+
+import liquibase.RuntimeEnvironment;
+import liquibase.Scope;
+import liquibase.change.core.RawSQLChange;
+import liquibase.changelog.*;
+import liquibase.changelog.filter.*;
+import liquibase.changelog.visitor.ChangeExecListener;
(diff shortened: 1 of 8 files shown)


suryaaki2 (approved):

liquibase-core/src/main/java/liquibase/command/core/AbstractRollbackCommandStep.java
```diff
@@ -0,0 +1,213 @@
...
+        CommandBuilder builder = new CommandBuilder();
+        ROLLBACK_SCRIPT_ARG = builder.argument("rollbackScript", String.class)
+                .description("Rollback script to execute").build();
+    }
+
+    private enum RollbackMessageType {
+        WILL_ROLLBACK, ROLLED_BACK, ROLLBACK_FAILED
+    }
+
+    protected abstract String getOperationCommand();
```
StevenMassaro, line 49: It is not clear what this method is supposed to do. Also is it necessary or could you just call the existing method which returns the command name?

StevenMassaro (approved):

liquibase-core/src/main/java/liquibase/command/core/AbstractRollbackCommandStep.java
filipelautert: You are right, I"m duplicating stuff. Thanks

PR #4078, opened 2023-04-04
Added changesetOperationStop time to exception handling also made all start/stop times be UTC DAT-13963

wwillard7800:
## Impact


liquibase-core/src/main/java/liquibase/changelog/ChangeSet.java
```diff
@@ -596,7 +597,8 @@ public ExecType execute(DatabaseChangeLog databaseChangeLog, ChangeExecListener
         }
 
         long startTime = new Date().getTime();
-        Scope.getCurrentScope().addMdcValue(MdcKey.CHANGESET_OPERATION_START_TIME, new ISODateFormat().format(new Date()));
+        String startTimeString = Instant.ofEpochMilli(startTime).toString();
```
suryaaki2, line 600: Should we add a method getStartTime() instead of duplicating the code at 3 places?

suryaaki2 (changes requested):

filipelautert (approved):
Agree with @suryaaki2 comments.. or maybe remove the  "[start|stop]TimeString" and move the conversion straight to where it is used?

filipelautert (approved):

suryaaki2 (approved):

Issue #3923, opened 2023-03-08
generateChangeLog for MySQL produces BIT(3) for boolean columns defined as TINYINT(1).


victorrodrigueznadq:
### Search first

- [X] I searched and no similar issues were found

### Description

generateChangeLog for MySQL produces BIT(3) for boolean columns defined as TINYINT(1).  This produces unexpected results when testing migrations during our build.  As a work around, I have to do the following, but I shouldn't have to do that.  If it's a TINYINT(1) in the database, it should be a TINYINT(1) in the databaseChangeLog.xml file produced.

```
find . -type f -name 'databaseChangeLog.xml' -exec sed -i 's:BIT(3):TINYINT(1):g' {} \;
```

Example....

```
create table rms.ACCOUNT_DESIGNATIONS
(
    ID                  varchar(36)            not null
        primary key,
    PARTICIPANT_ID      int                    not null,
    MP_ID               varchar(32)            null,
    USER_ID             int                    null,
    USER_ACCOUNT        varchar(32)            null,
    BILLING_ENTITY_ID   int                    null,
    BILLING_ENTITY_SIGN varchar(32)            null,
    IS_AUTD             tinyint(1) default 0   null,
    XSTO_EQ_PL          varchar(4)             null,
    XHEL_EQ_PL          varchar(4)             null,
    XCSE_EQ_PL          varchar(4)             null,
    XSTO_EQ_LPS         tinyint(1) default 0   null,
    XHEL_EQ_LPS         tinyint(1) default 0   null,
    XCSE_EQ_LPS         tinyint(1) default 0   null,
    NOK_EQ_LPS          tinyint(1) default 0   null,
    XSTO_EQ_TOP         tinyint(1) default 0   null,
    XHEL_EQ_TOP         tinyint(1) default 0   null,
    XCSE_EQ_TOP         tinyint(1) default 0   null,
    XSTO_WAR_PL         varchar(4)             null,
    XHEL_WAR_PL         varchar(4)             null,
    XCSE_WAR_PL         varchar(4)             null,
    XSTO_ETF_LPS        tinyint(1) default 0   null,
    XHEL_ETF_LPS        tinyint(1) default 0   null,
    NOK_ETF_LPS         tinyint(1) default 0   null,
    VALID_FROM_DATE     date                   null,
    YEAR_AND_MONTH      date                   null,
    NOTES               varchar(256)           null,
    TERMINATE_DATE      date                   null,
[...]

Issue #2037, opened 2021-08-16
MySQL Datatype incorrectly applied TINYINT(1) -> BIT(1)


winklevos:
## Environment

**Liquibase Version**: Liquibase Community 4.3.5

**Liquibase Integration & Version**: CLI

**Liquibase Extension(s) & Version**: 

**Database Vendor & Version**: MySQL 8.0.21

**Operating System Type & Version**:  windows 10

## Description

- Any TINYINT(1) columns are converted to BIT(1) when generating a changelog. This is incorrectly applied boolean type for MySQL https://dev.mysql.com/doc/refman/8.0/en/other-vendor-data-types.html
- Some other external programs reading the database will not recognise BIT(1) as boolean for MySQL as it would with TINYINT(1)

## Steps To Reproduce
1. create table with boolean TINYINT(1) field
```CREATE TABLE `xxx` (
	`boolean` TINYINT(1) NOT NULL DEFAULT '0'
)
COLLATE='utf8mb4_0900_ai_ci'
ENGINE=InnoDB
;```

2. Generate changlog containing a TINYINT(1) field

## Actual Behavior

```
<changeSet author="xxx (generated)" id="xxx-x">
        <addColumn schemaName="xxx" tableName="xxxx">
            <column defaultValueBoolean="false" name="boolean" type="BIT(1)">
                <constraints nullable="false"/>
            </column>
        </addColumn>
</changeSet>
```

## Expected/Desired Behavior

```
<changeSet author="xxx (generated)" id="xxx-x">
        <addColumn schemaName="xxx" tableName="xxxx">
            <column defaultValueBoolean="false" name="boolean" type="TINYINT(1)">
                <constraints nullable="false"/>
            </column>
        </addColumn>
</changeSet>
```

molivasdat:
Hi @winklevos Thanks for creating this issue.  We will add this to the list of issues to process.

nvoxland:
This ends up being a "feature" of the mysql driver. When we ask it for the type of the `boolean` column in the example above, the driver tells us that it is of type bit and NOT tinyint like it actually is. 

There is a mysql URL parameter of `tinyInt1isBit=false` which changes the driver's behavior. See https://stackoverflow.com/questions/51258496/tinyint-returning-as-bit-in-databasemetadata-java

[...]

PR #4090, opened 2023-04-06
Fix mysql boolean behaviours

nwcm:
## Impact

## Description

MySQL boolean is defined as `TINYINT(1)` currently liquibase maps it to `BIT` which can cause problems reading from the tables.


nvoxland:
Initial review:

We should be following the vendor definitions of what the mapping should be. 

This will be a very impacting changes to users, because anyone with a changeset with `type="boolean"` on mysql will now get differently created tables depending on the version of liquibase they use. If that is the correct change to make, it's what we should do. But it will need to be well explained as part of the release notes.

Things to worry about:
- Need to test the various driver settings

kevin-atx:
@nvoxland  - could you please update this PR with the definition of the spike needed to research all of the current MySQL datatypes vs the suggested datatypes as specified by MySQL [here](https://dev.mysql.com/doc/connector-j/8.0/en/connector-j-reference-configuration-properties.html)? We will review Daniel/Filipe's output of this spike in the next sprint review meeting so we can better understand how to move forward safely (and completely).

nwcm:
Also for reference https://dev.mysql.com/doc/connector-j/8.0/en/connector-j-reference-type-conversions.html

MalloD12:
Hi @nwcm, 

Thank you for creating this PR to solve this boolean issue. Code changes look good to me, I have only changed one thing which is removing `tinyint(1)` alias from the `BooleanType` class because we don't want to limit it uses just a boolean. Also, I'm not approving it yet based on Nathan's comment it would be good to see if there is a way we can reduce the impact on users using previous versions. I'll discuss it next week with other members of the team to see if we can move forward with this PR.

Testing Note:  I manually tested this change on MySQL 5.7 and 8.0, and it behaves as expected in both cases boolean is deployed in the DB as a `tinyInt(1)` type.

MalloD12 (approved):
**Approved**. 

Code changes look ok to me and are now MySQL docs compliance. Although during testing we identify this requires to update some unit and test harness tests, and also we see behaviour is in a way for `update` command (a boolean type in a changelog will be deployed in the DB as a `TinyInt(1)`), but it does behave in a different way for `generate-changelog` and `snapshot` commands for which based on MySQL connection default setup when fetching a `TinyInt(1)` column it will return a `BIT` type. We have tracked this details in a PD [ticket](https://datical.atlassian.net/browse/PD-3514?atlOrigin=eyJpIjoiYTU3MTE1Mzg4NzUyNDBjYmFmY2MyOTZkYThhYzRiNjgiLCJwIjoiamlyYS1zbGFjay1pbnQifQ), so we get our documentation correctly updated with all this info. 

Also, as mentioned above it required us to make some updates on Test Harness suite for which @PavloTytarchuk has created this [PR](https://github.com/liquibase/liquibase-test-harness/pull/584). I very appreciate all your contribution with this PR, Pavlo!!

Build and rest of the tests (unit, integration and functional test) have been successfully executed.

MalloD12:
#3923 seems to be related to #2037, but I could not reproduce it. Having a `tinyInt(1)` column in the DB when generating a changelog I got a `BIT(1)` column.

rberezen:
MySQL boolean is defined as TINYINT(1) now.

- Test harness updated. 
- Unit tests updated.
- Manual testing with with MariaDB and MariaDB Connector/J of different versions passed.

APPROVED

rberezen (approved):

filipelautert:
@MalloD12 please see the comment above - seems we should have used only tinyint instead of tinyint(1) ...

MalloD12:
I have created #5069 to make the necessary updates. If anyone is keen to work on it, please feel free to create a PR for it.

Thanks,
Daniel.

PR #4110, opened 2023-04-10
Add support for Unicode roman numerals.

### LINKED ISSUE #4107: TokenMgrError: Lexical error
### Search first

### Description

When the sql statement field value contains Roman numerals II, an exception is thrown 【2023-04-10 11:17:30.424 [main] WARN  org.springframework.boot.web.servlet.context.AnnotationConfigServletWebServerApplicationContext | 559 | Exception encountered during context initialization - cancelling refresh attempt: org.springframework.beans.factory.BeanCreationException: Error creating bean with name 'liquibase' defined in class path resource [org/springframework/boot/autoconfigure/liquibase/LiquibaseAutoConfiguration$LiquibaseConfiguration.class]: Invocation of init method failed; nested exception is liquibase.util.grammar.TokenMgrError: Lexical error at line 8648, column 123.  Encountered: "\u2161" (8545), after : ""】

### Steps To Reproduce

INSERT INTO `std_standard_organization` VALUES (1, '001', 'Ⅱ', 'Ⅱ', 1, 9, b'1', '002', 99, 1640966400000, NULL, '/1/308/315/', b'0', 1679995673513, 1679995673513, 0);

### Expected/Desired Behavior

SQL statement field value can contain Roman numerals II

### Liquibase Version

_No response_

### Database Vendor & Version

_No response_

### Liquibase Integration

_No response_

### Liquibase Extensions

_No response_

_No response_

### Additional Context

_No response_

jccampanero:
## Impact

## Description

Modify the `javacc` `SimpleSqlGrammar.jj` file adding support for roman numerals as defined in the ["Number Forms" Unicode block](https://www.compart.com/en/unicode/block/U+2150).

Fixes #4107.

## Things to be aware of

Nothing.

## Things to worry about

Nothing.

## Additional Context

Is it necessary to define a new token block?

Tests are pending.


nvoxland (approved):

liquibase-core/src/main/javacc/liquibase/util/grammar/SimpleSqlGrammar.jj
```diff
@@ -370,6 +370,8 @@ Built list from http://stackoverflow.com/a/37668315/45756
     "\u213C"-"\u213F",
     "\u2145"-"\u2149",
     "\u214E",
+    "\u2160"-"\u216F", //Roman numerals
```
nvoxland, line 373: Should we just add the whole "Number Forms" block https://www.compart.com/en/unicode/block/U+2150 instead of these specifically?

jccampanero: I though about it when I performed the change @nvoxland. In the end I didn't include the whole block because, on one hand, `SimpleSqlGrammar.jj` already included the values `\u2183` and `\u2184`; on the other,  I think the range from `\u2180` to `\u218B` defines symbols that are used less frequently.

In any case, I will be glad to include the whole "Number Forms" block if you prefer. Should I change it?

MalloD12:
Hi @jccampanero, could you please share some more details on how to reproduce #4107, I was trying to add some tests but I realised I was needing some more context. I tried by first creating a table with a single year column (int type), and then tried performing some inserts into it by adding values like "IIIII" and "MMXXIII" expecting they were going to be parsed and inserted as integers, but got a DatabaseException:

```
Data conversion error converting "'IIIIIIIII' (ROMANNUMBERSTEST: ""year"" INTEGER)"; SQL statement:
INSERT INTO PUBLIC.romanNumbersTest ("year") VALUES ('IIIIIIIII')```

Were you able to reproduce this issue? Any specific DB or it's applicable to all and it's more a Spring issue? 

Thanks,
Daniel.

**Comment by jccampanero** [comment]
Hi @MalloD12.

Well, honestly, when I created the PR I just changed the JavaCC grammar being confident enough that it will solve the problem.

After your comment, I have tried reproducing the issue and I think this is the actual use case (spoiler, I am afraid it isn't what we are expecting for). Please, try the following simple code fragment:

```java
SqlParser.parse("INSERT INTO a VALUES (´Ⅱ´);");
```
A very important thing to note here is that the "single quotes" are not actually "single quotes", the ones associated with the question mark keyboard key, but those associated with the curly brackets key.

Everything works smoothly if you use usual single quotes.

But when you use this kind of "single quotes" the strange thing is that, if we define the symbol in the JavaCC grammar as explained in the PR, everything works properly and the error disappear.

MalloD12:
> Hi @MalloD12.
> 
> Well, honestly, when I created the PR I just changed the JavaCC grammar being confident enough that it will solve the problem.
> 
> After your comment, I have tried reproducing the issue and I think this is the actual use case (spoiler, I am afraid it isn't what we are expecting for). Please, try the following simple code fragment:
> 
> ```java
> SqlParser.parse("INSERT INTO a VALUES (´Ⅱ´);");
> ```
> 
> A very important thing to note here is that the "single quotes" are not actually "single quotes", the ones associated with the question mark keyboard key, but those associated with the curly brackets key.
> 
> Everything works smoothly if you use usual single quotes.
> 
> But when you use this kind of "single quotes" the strange thing is that, if we define the symbol in the JavaCC grammar as explained in the PR, everything works properly and the error disappear.
> 
> How should we deal with this?

Thank you, @jccampanero! I'll try this out and post it here my results later.

MalloD12:
@jccampanero I tried to execute `SqlParser.parse("INSERT INTO a VALUES ('Ⅱ');");` with and without your fixed and it worked the same way for me, I couldn't get the reported error in #4107.

jccampanero:
> @jccampanero I tried to execute `SqlParser.parse("INSERT INTO a VALUES ('Ⅱ');");` with and without your fixed and it worked the same way for me, I couldn't get the reported error in #4107.

Hi @MalloD12 Daniel.

In my tests the problem occurs only when using ´Ⅱ´, not 'Ⅱ': please, observe the difference in the quotes.

In a certain way, those kinds of quotes should be supported?

MalloD12:
> > @jccampanero I tried to execute `SqlParser.parse("INSERT INTO a VALUES ('Ⅱ');");` with and without your fixed and it worked the same way for me, I couldn't get the reported error in #4107.
> 
> Hi @MalloD12 Daniel.
> 
> In my tests the problem occurs only when using ´Ⅱ´, not 'Ⅱ': please, observe the difference in the quotes.
> 
> In a certain way, those kinds of quotes should be supported?

Does this mean when using `´Ⅱ´` you get the reported error, but you don't when using `'II'`? To be honest, I don't know whether these quotes should be used or not. What would be the value of supporting those other quotes? I'm just asking because I think I have never seen SQL examples using them. 

@filipelautert @nvoxland do you guys have any other opinion about this?

jccampanero:
> > > @jccampanero I tried to execute `SqlParser.parse("INSERT INTO a VALUES ('Ⅱ');");` with and without your fixed and it worked the same way for me, I couldn't get the reported error in #4107.
> > 
> > 
> > Hi @MalloD12 Daniel.
> > In my tests the problem occurs only when using ´Ⅱ´, not 'Ⅱ': please, observe the difference in the quotes.
> > In a certain way, those kinds of quotes should be supported?
> 
> Does this mean when using `´Ⅱ´` you get the reported error, but you don't when using `'II'`? To be honest, I don't know whether these quotes should be used or not. What would be the value of supporting those other quotes? I'm just asking because I think I have never seen SQL examples using them.
> 
> @filipelautert @nvoxland do you guys have any other opinion about this?

Hi @MalloD12.

I completely agree with you, I think that type of quotes don't make sense, they are useless. As I said before, when I implemented the fix, I assumed that the error was present; only when you mentioned that you were unable to reproduce the issue I realized that in fact it is only present when using that type of quotes.

In any case, I think the change still could be valuable: the range of Unicode symbols included are rare but they could may sense in certain applications like those related to math, etc.

Another important point is understand why the parser accepts the symbol if it is not directly supported in the grammar: perhaps, one good test could be inserting the character in an actual database system and see if the symbol is selected as it should.

In any way, as I said, I agree with you, the issue probably shouldn't be considered as such but as an improvement instead, if applicable.

MalloD12:
> > > > @jccampanero I tried to execute `SqlParser.parse("INSERT INTO a VALUES ('Ⅱ');");` with and without your fixed and it worked the same way for me, I couldn't get the reported error in #4107.
> > > 
> > > 
> > > Hi @MalloD12 Daniel.
> > > In my tests the problem occurs only when using ´Ⅱ´, not 'Ⅱ': please, observe the difference in the quotes.
> > > In a certain way, those kinds of quotes should be supported?
> > 
> > 
> > Does this mean when using `´Ⅱ´` you get the reported error, but you don't when using `'II'`? To be honest, I don't know whether these quotes should be used or not. What would be the value of supporting those other quotes? I'm just asking because I think I have never seen SQL examples using them.
> > @filipelautert @nvoxland do you guys have any other opinion about this?
> 
> Hi @MalloD12.
> 
> I completely agree with you, I think that type of quotes don't make sense, they are useless. As I said before, when I implemented the fix, I assumed that the error was present; only when you mentioned that you were unable to reproduce the issue I realized that in fact it is only present when using that type of quotes.
> 
> In any case, I think the change still could be valuable: the range of Unicode symbols included are rare but they could may sense in certain applications like those related to math, etc.
> 
> Another important point is understand why the parser accepts the symbol if it is not directly supported in the grammar: perhaps, one good test could be inserting the character in an actual database system and see if the symbol is selected as it should.
> 
> In any way, as I said, I agree with you, the issue probably shouldn't be considered as such but as an improvement instead, if applicable.

Hi @jccampanero, I have tried today your suggestion about trying to insert this character into the DB and couldn't reproduce it either. Let me discuss this with the team and I'll let you know it. 

Thanks,
Daniel.

jccampanero:
Of course @MalloD12  Daniel, as you need.

Thank you for your support in testing this and other contributions, it is great.

MalloD12 (approved):
**Approved**.

Thank you, @jccampanero. After discussing with the team even though we haven't been able to reproduce it and add some test to it, we are ok to move forward with these changes as they look simple and to not hurt being there. I appreciate your  contribution and your assistance to help me reproducing and testing this change.  

Daniel.

XDelphiGrl (approved):
This PR adds the unicode character encodings for upper and lowercase Roman numerals to Liquibase's JavaCC SQL grammar parser. This may address the [reported bug](https://github.com/liquibase/liquibase/issues/4107) but after several tries, no one could reproduce it. However, adding this character range to the grammar parser is benign (and may help).
* Functional tests passing.
* Test harness failures are due to an environmental issue (Titan did not start); this can be disregarded for the purposes of this PR.

APPROVED

jccampanero:
> **Approved**.
> 
> Thank you, @jccampanero. After discussing with the team even though we haven't been able to reproduce it and add some test to it, we are ok to move forward with these changes as they look simple and to not hurt being there. I appreciate your contribution and your assistance to help me reproducing and testing this change.
> 
> Daniel.

Sorry for the late reply @MalloD12 Daniel.

Thank you very much, I really appreciate it.

MalloD12:
> > **Approved**.
> > Thank you, @jccampanero. After discussing with the team even though we haven't been able to reproduce it and add some test to it, we are ok to move forward with these changes as they look simple and to not hurt being there. I appreciate your contribution and your assistance to help me reproducing and testing this change.
> > Daniel.
> 
> Sorry for the late reply @MalloD12 Daniel.
> 
> Thank you very much, I really appreciate it.
> 
> Thank you very much you too @XDelphiGrl.

No worries @jccampanero. Thank you for all your contributions, they are very appreciated it.

Daniel.

filipelautert (approved):

PR #4115, opened 2023-04-11
PD-3256: STRICT=true description

adrian-velonis1:
https://datical.atlassian.net/browse/PD-3256

Providing more useful description for STRICT global config; minor tweaks to other descriptions

### DIFF
--- a/liquibase-cli/src/test/groovy/liquibase/integration/commandline/LiquibaseCommandLineTest.groovy
+++ b/liquibase-cli/src/test/groovy/liquibase/integration/commandline/LiquibaseCommandLineTest.groovy
@@ -204,7 +204,7 @@ Global Options
 
   -h, --help                 Show this help message and exit
 
-      --headless=PARAM       Force liquibase to think it has no access to a
+      --headless=PARAM       Force Liquibase to think it has no access to a
                                keyboard
                              DEFAULT: false
                              (defaults file: 'liquibase.headless', environment
@@ -351,8 +351,8 @@ Global Options
                                'LIQUIBASE_OUTPUT_LINE_SEPARATOR')
 
       --preserve-schema-case=PARAM
-                             Should liquibase treat schema and catalog names as
-                               case sensitive?
+                             If true, Liquibase treats schema and catalog names
+                               as case sensitive
                              DEFAULT: false
                              (defaults file: 'liquibase.preserveSchemaCase',
                                environment variable:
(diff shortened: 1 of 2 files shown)


liquibase-core/src/main/java/liquibase/GlobalConfiguration.java
```diff
@@ -174,13 +174,13 @@ public class GlobalConfiguration implements AutoloadedConfigurations {
...
         HEADLESS = builder.define("headless", Boolean.class)
-                .setDescription("Force liquibase to think it has no access to a keyboard")
+                .setDescription("Force Liquibase to think it has no access to a keyboard")
                 .setDefaultValue(false)
                 .setCommonlyUsed(true)
                 .build();
 
         STRICT = builder.define("strict", Boolean.class)
-                .setDescription("Be stricter on allowed Liquibase configuration and setup?")
+                .setDescription("If true, restricts Liquibase configuration and setup: requires changeset author to be non-empty and requires LIQUIBASE_* environment variable names to be valid.")
```
nvoxland, line 183: The strict setting also requires settings in the defaultsFile to be valid as well. I wonder if even now it's a lot to list the specifics of what strict mode does in this help description and instead we should make a docs page and point to that? At some point there will be enough strict details we'll have to, so maybe now?

I'd suggest giving a more "why" answer than the details of what it specifically does. Like: `If true, Liquibase enforces certain best practices and proactively looks for common errors. For more information see https://docs.liquibase.com/strict` and then have that strict page give details and examples about the properties file, env variables, the author tag, and whatever else we add down the road

adrian-velonis1: @nvoxland Your description works for me! I'm (slowly) creating individual pages for all the global arguments to provide more detailed usage/behavior descriptions and I can definitely make one for `strict` soon.

Though this begs the question of why we have so many disparate behaviors going through one argument. For example, if a user wants to be strict about envar names, but doesn't want to be strict on empty `author` values, shouldn't we support that via multiple distinct arguments? (Even if it's not a best practice, it's already not really a best practice to disable this in the first place.)

adrian-velonis1:
New docs page: https://github.com/Datical/liquibase-docs/blob/master/Content/parameters/strict.htm

https://datical.atlassian.net/browse/PD-3309

StevenMassaro (approved):

filipelautert:
Hi @adrian-velonis1 - is this one still valid?

adrian-velonis1:
@filipelautert Yes this is still valid AFAIK!

filipelautert:
> @filipelautert Yes this is still valid AFAIK!

Thanks @adrian-velonis1 ! I fixed the merge conflicts, could you take a last look to see if it still ok? Then I can merge it.

filipelautert (approved):

filipelautert:
@adrian-velonis1 Ouch, tests needs to be fixed as in the last PR you created and Steve reviewed.

PR #4119, opened 2023-04-12
Switch to distribute jars generated by Java 11 allowing Jakarta cdi 4 upgrade

filipelautert:
## Impact

## Description

cdi version 5 requires Java 11, but liquibase still supports Java 8. . In order to use it some changes were made:
* added an exclusion so jakarta-cdi module is skipped when building using Java 8
* switched distributing jars from the ones created by Java 8 to the ones built by Java 11 (target still 8), so jakarta-cdi module is included

With that we are able to accomplish the dependabot requirements:
* Upgrade CDI to version 4 and Weld to 5
* Drops Java 8 support

Relevant comments from dependabot issue: https://github.com/liquibase/liquibase/pull/3891#issuecomment-1460343043


StevenMassaro (approved):

.github/workflows/build.yml
```diff
@@ -185,9 +185,15 @@ jobs:
         uses: stCarolas/setup-maven@v4.5
         with:
           maven-version: ${{ env.MAVEN_VERSION }}
+      - name: Build & Test Java 8
+        if: ${{ matrix.java == 8}}
+        run: |
+          mvn -B "-Dbuild.repository.owner=${{ needs.setup.outputs.thisRepositoryOwner }}" "-Dbuild.repository.name=${{ needs.setup.outputs.thisRepositoryName }}" "-Dbuild.branch=${{ needs.setup.outputs.thisBranchName }}" "-Dbuild.number=${{ github.run_number }}" "-Dbuild.commit=${{ needs.setup.outputs.thisSha }}" "-DtrimStackTrace=false" -pl '!liquibase-dist,!liquibase-cdi-jakarta' clean test package surefire-report:report
       - name: Build & Test
```
StevenMassaro, line 192: Can you change this step to be called something like "Build & Test all other Java versions"

filipelautert: I'm calling it non-jdk-8 to be compliant with the pom profile

liquibase-cdi-jakarta/src/main/java/liquibase/integration/jakarta/cdi/CDIBootstrap.java
```diff
@@ -85,11 +78,6 @@ public boolean isAlternative() {
                 return false;
             }
 
-            @Override
```
XDelphiGrl, line 88: May I inquire as to why this method has been removed, @filipelautert ?

filipelautert: It was removed because it was removed upstream - so it's "overridable" anymore.

XDelphiGrl: Thanks for the explanation, @filipelautert.

XDelphiGrl (approved):
PR updates Weld and the Jakarata Enterprise CDI API dependencies for `liquibase-cdi-jakarta`. Weld is now `5.1.0.Final` and the CDI API is `4.0.1`. The associated GitHub workflows are updated to build the Liquibase Jakarta CDI plugin with Java 11 as required by CDI 5.
* No testing required.

APPROVED

PR #4137, opened 2023-04-14
GitHub action concurrency

filipelautert:
## Impact

## Description

That change aims to save us some resources. With this settings, when a second push is done to a branch it will cancel the previous build if it still running. Example:

* Build 1 was executing until.... -> https://github.com/liquibase/liquibase/actions/runs/4797020625
* Build 2 started running and cancelled it -> https://github.com/liquibase/liquibase/actions/runs/4797026004/jobs/8533500060
* The canceling message at build 1 says: `Canceling since a higher priority waiting request for 'Build and Test-refs/heads/github-action-concurrency' exists` . We can notice that this is the key defined at **build.yml -> concurrency -> groups:** `${{ github.workflow }}-${{ github.event.pull_request.number || github.ref }}` 

## Things to be aware of

With our current github build flow it will only stop the builds on core. If we think it is a good tool we can replicate it to liquibase-pro-tests and pro too.

Kudos to @abrackx for finding this setting.


abrackx (approved):
:shipit:

abrackx (changes requested):
Don't cancel on master branch

abrackx (approved):
Not 100% sure this will **not** cancel the builds on master (we intend to have a build for each commit on that branch), but we can always revert this if it ends up not working. 🚢

Issue #4048, opened 2023-03-28
ERROR:  permission denied for sequence (in unrelated schema)


joaocc:
### Search first

- [X] I searched and no similar issues were found

### Description

When using loadData (in xml changelog) to load CSV data into tables in schemaA, liquibase hits an error due to lack of permissions on an unrelated schema of the same database. (postgresql 15.1)

```
[3480] ERROR:  permission denied for sequence some_sequence_in_schema2
[3480] STATEMENT:  
```

the query in question is
```
SELECT NS.NSPNAME AS SCHEMA_NAME,
	TD.RELNAME AS TABLE_NAME,
	PA.ATTNAME AS COLUMN_NAME,
	COALESCE(PG_SEQUENCE_LAST_VALUE(C.OID::REGCLASS) + S.SEQINCREMENT, S.SEQSTART) AS START_VALUE,
	S.SEQINCREMENT AS INCREMENT_BY
FROM PG_CLASS C
JOIN PG_SEQUENCE S ON C.OID = S.SEQRELID
JOIN PG_NAMESPACE NS ON C.RELNAMESPACE = NS.OID
JOIN PG_DEPEND D ON C.OID = D.OBJID
JOIN PG_CLASS TD ON TD.OID = D.REFOBJID
JOIN PG_ATTRIBUTE PA ON PA.ATTRELID = TD.OID
AND PA.ATTNUM = D.REFOBJSUBID
WHERE C.RELKIND = 'S'
	AND D.DEPTYPE = 'a'
```
It seems like Liquibase is searching for a sequence in all the schemas, which may be blocked due to lack of permissions

### Steps To Reproduce

Unable to provide info to repro due to code confidentiality clauses

the loadData statement is as follows:
```
        <loadData
                catalogName="cat"
                commentLineStartsWith="#"
                encoding="UTF-8"
                file="seed.test/file1.tsv"
                quotchar="&quot;"
                relativeToChangelogFile="false"
                separator="\t"
                tableName="some_table_in_schema1"
                usePreparedStatements="false">
            <column name="id1" type="numeric"/>
            <column name="ts1" type="timestamp"/>
            <column name="ts2" type="timestamp" defaultValueDate="NULL"/>
            <column name="data1" type="STRING"/>
        </loadData>
```
the corresponding table is similar to this:
```
CREATE  TABLE some_table_in_schema1 (
	id1              bigserial  NOT NULL  ,
[...]

PR #4140, opened 2023-04-15
low-skilled attempt at fixing #4048

joaocc:
## Impact

- [X ] Bug fix (non-breaking change which fixes expected existing functionality)
 
## Description
https://github.com/liquibase/liquibase/issues/4048 description and discussion indicate that (at least in postgres) the search for the sequences referring to a specific column should be restricted to the schema the column resides in.

## Things to be aware of

Injected the schema name from the column in the function. Renamed a private method that is used only here.

## Things to worry about

Please note that this was not tested, as I am not an active java dev, and am only trying to help in advancing a possible solution for the simple issue.

### DIFF
--- a/liquibase-core/src/main/java/liquibase/snapshot/jvm/ColumnAutoIncrementService.java
+++ b/liquibase-core/src/main/java/liquibase/snapshot/jvm/ColumnAutoIncrementService.java
@@ -44,14 +44,14 @@ public class ColumnAutoIncrementService {
      * @param snapshot snapshot data used to store cache information.
      * @return Map with the sequence name and auto increment details
      */
-    public Map<String, Column.AutoIncrementInformation> obtainSequencesInformation(Database database, DatabaseSnapshot snapshot) {
+    public Map<String, Column.AutoIncrementInformation> obtainSequencesInformation(Database database, Schema schema, DatabaseSnapshot snapshot) {
         if (autoIncrementColumns != null) {
             return autoIncrementColumns;
         } else {
             autoIncrementColumns = new ConcurrentHashMap<>();
             Executor executor = Scope.getCurrentScope().getSingleton(ExecutorService.class).getExecutor("jdbc", database);
             try {
-                String query = this.getQueryForDatabase(database);
+                String query = this.getQueryForDatabaseAndSchema(database, schema);
                 List<Map<String, ?>> rows = executor.queryForList(new RawSqlStatement(query));
                 for (Map<String, ?> row : rows) {
                     String schemaName = (String) row.get("SCHEMA_NAME");
@@ -74,12 +74,13 @@ public class ColumnAutoIncrementService {
         return autoIncrementColumns;
     }
 
-    private String getQueryForDatabase(Database database) throws LiquibaseException {
+    private String getQueryForDatabaseAndSchema(Database database, Schema schema) throws LiquibaseException {
         if (database instanceof MSSQLDatabase) {
             return "SELECT object_schema_name(object_id) AS schema_name, " +
                     "object_name(object_id) AS table_name, name AS column_name, " +
                     "CAST(seed_value AS bigint) AS start_value, " +
                     "CAST(increment_value AS bigint) AS increment_by " +
+                    ()
                     "FROM sys.identity_columns";
         } else if (database instanceof PostgresDatabase) {
             int version = 9;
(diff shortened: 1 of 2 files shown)


liquibase-core/src/main/java/liquibase/snapshot/jvm/ColumnAutoIncrementService.java
```diff
@@ -116,7 +117,7 @@ private String getQueryForDatabase(Database database) throws LiquibaseException
                         "    JOIN pg_depend d ON c.oid = d.objid " +
                         "    JOIN pg_class td ON td.oid = d.refobjid " +
                         "    JOIN pg_attribute pa ON  pa.attrelid=td.oid AND pa.attnum=d.refobjsubid " +
-                        "WHERE c.relkind = 'S' AND d.deptype = 'a'";
+                        "WHERE c.relkind = 'S' AND d.deptype = 'a' AND ns.nspname = '" + schema.getName() + "'";
```
nvoxland, line 120: There are a couple other `if` branches we should make sure are respecting the new schema name

nvoxland:

Thanks for the start to the fix, @joaocc! It points out a good spot to fix, not just for this permission issue but also as a performance improvement in general. 

I added a line comment that we should look at applying the extra `where` clause to the other schemas, but otherwise the fix seems good. If there are problems with the query or how the info gets read, we should catch it with the existing tests
 
Questions I have:
- None

Potential risks:
- Test that the case of the schema is correctly handled. I think it's coming from the snapshot so it should be cased as the database expects, but need to double check.

- Nothing

MalloD12:
Hi, @joaocc! Could you please give write permissions to your repo? I was trying to push a code change extension from what you have already done here, and want to share this PR with some members of the team to make sure it works correctly too for MSSQL.

joaocc:
> Hi, @joaocc! Could you please give write permissions to your repo? I was trying to push a code change extension from what you have already done here, and want to share this PR with some members of the team to make sure it works correctly too for MSSQL.

Sure. Added @MalloD12 already. Please let me know if anyone else needs access. Thx

MalloD12:
> > Hi, @joaocc! Could you please give write permissions to your repo? I was trying to push a code change extension from what you have already done here, and want to share this PR with some members of the team to make sure it works correctly too for MSSQL.
> 
> Sure. Added @MalloD12 already. Please let me know if anyone else needs access. Thx

Thanks @joaocc!

MalloD12:
Hey @XDelphiGrl, 

As I modified also MSSQL query in the `getQueryForDatabaseAndSchema()` method. Would you mind testing in this DB as well this issue is not happening there. The steps I was using to reproduce this error were:

```
1. Create another schema
2. Create a test user
3. Create a table in the new schema created (one of the column should be serial/bigserial type)
4. Gran new user permissions to default schema (I was using `public` in  Postgres) in MSSQL.
 Command Executed in Posgtres: `GRANT ALL ON ALL TABLES IN SCHEMA "public" TO <test user>`
5. Then executed the changelog with two changesets (without specifying any schema on them) one for creating another simple table and one to perform a simple LoadData
```
For 5. I was deploying below change sets:
```
<changeSet id="1" author="mallod">
        <createTable tableName="products">
            <column name="id" type="int" autoIncrement="true" />
            <column name="name" type="varchar(250)" />
            <column name="price" type="number(8,2)" />
        </createTable>
    </changeSet>

    <changeSet id="2" author="mallod">
        <loadData tableName="products" file="changelog/test_data.csv"
                  encoding="UTF-8"
                  quotchar="&quot;"
                  relativeToChangelogFile="false"
                  usePreparedStatements="false">
            <column name="id" />
            <column name="name" />
            <column name="price" />
        </loadData>
```

Please let me know if something is not clear enough and I'll do my best to clarify it.

Thanks,
Daniel.

joaocc:
Thanks.
Not sure if we will be able to test in the next days, but we have a clearer understanding of what seems to happen.
We have a set of migration scripts that deploy several schemas into a single DB, each with access from a separate user.
If we deploy in a specific order, everything is ok.
If we deploy in a different order (s-1 using user-1 and then s-2 using user-2), one of the migrations on s-2 will do a query across all schemas, which will encounter a sequence in s-1, for which user-2 has no permission.
We solved this by giving user-2 permissions on s-1 to modify the sequence.
Of course, ideally this should not be needed, as per the rationale of the issue.
Hope this helps.

MalloD12 (approved):
**Approved**.

Thank you for submitting this PR @joaocc. My apologies for taking a bit longer to complete review on this one.  Code change you provide to us looks good to me, I have applied the same query change you did into the postgres version lower than 10 and mssql which is intended to improve performance a bit. I have also manually tested this change (team helped me to test it on mssql), and I have added an postgres integration test with the same steps I was performing manually.

Thanks,
Daniel.

filipelautert (approved):

suryaaki2 (approved):

filipelautert:
Hello @joaocc - hopefully we will be releasing until the end of the month 🤞

PR #4145, opened 2023-04-17
DAT-14194

sayaliM0412:
## Impact

## Description

- Upgrade-all-github-runner-images-to-Ubuntu-22.04
- Github runner images were upgraded to Ubuntu 22.04 on `core/build.yml`, but all the other scripts remain on Ubuntu 20.04 
- There is a known issue: Ubuntu 22.04 is missing the command `rename.ul`, used at script re-version.sh, which only runs at release time.

### DIFF
--- a/.github/util/re-version.sh
+++ b/.github/util/re-version.sh
@@ -95,7 +95,7 @@ do
   rm -rf $workdir/rebuild
 
   cp $workdir/$jar $outdir
-  rename 0-SNAPSHOT $version $outdir/$jar
+  mv -v 0-SNAPSHOT $version $outdir/$jar
 done
 
 ## Test jar structure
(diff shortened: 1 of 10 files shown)


.github/util/re-version.sh
```diff
@@ -75,7 +75,9 @@ do
   (cd $workdir/finalize-jar && jar cfm $workdir/$jar $workdir/tmp-manifest.mf .)
 
   cp $workdir/$jar $outdir
-  rename.ul 0-SNAPSHOT $version $outdir/$jar
+  RENAME_SNAPSHOTS=$(ls $outdir/$jar | sed -e "s/0-SNAPSHOT/$version/g")
```
jnewton03, line 78: @sayaliM0412 there's no chance the version will ever have a `/` in it, right?  I've been bitten by using that particular delimiter with sed before.

sayaliM0412: hmm, it's replacing `0-SNAPSHOT` with `$version` globally. 
🤔

filipelautert: I believe we will never have a version like 4.21/0 unless it's a typo - so I believe it's fine.

filipelautert (approved):

PR #4148, opened 2023-04-18
maven-plugin: Support increasing maven log level

mensinda:
## Impact

## Description

This PR adds a new property to the maven plugin, which allows increasing the log level with a new `logLevel` property.

Our primary use case / problem is that liquibase has *a lot* of INFO level logs. For instance, the `Loading artifacts into URLClassLoader` section is already bigger than the rest of the non-liquibase maven output for some of our projects. To solve this we want to *only* increase the log level of the liquibase plugin (so no `mvn --quiet`). Sadly, maven does not support changing the log level for a specific plugin inside the `pom.xml`. Thus, this PR is required.

## Things to be aware of

- Impact on the codebase: relatively minimal, since there is only one new class, and the logger is changed at a central location.

## Things to worry about

None

## Additional Context

I am aware that there already exists the [maven SLF4J logging configuration](https://maven.apache.org/maven-logging.html). However this does not fit our need since this is a *maven installation* and not *project* configuration.


nvoxland:

I'm a bit torn on what I think on this PR, and don't have an off-hand answer so I'll move it along the process for more discussion.

Like you said, Maven has their built-in functionality for managing the logs and my base theory is that Liquibase should be simply plugging into the existing logging system and not trying to fight it or change it. That if we try to go our own way with the logging, it ends up with strange edge cases due to conflicts in configurations. At the same time, I get that the maven logging config has limitations on how it can be configured and how well it's documented. 

I also wonder if the root problem of `Our primary use case / problem is that liquibase has a lot of INFO level logs` is really best solved by looking at what is at info level and fixing that? The info level should be "what a regular user would expect as CLI output" aka "reasonably significant messages that will make sense to end users and system administrators" and a lot of that sounds like it fails that test.

So I'll open it up to the larger group:
- Does fixing the log level of a few messages solve the problem best?
- Do we want to do more than rely on what maven has configured for logging?
- Is this implementation the right answer to the problems? How is it handling edge cases like "what if people are also configuring things like levels and channels in slf4j?" or "how does it work with simpler `-X` type log-related flags?" etc?

mensinda:
> Does fixing the log level of a few messages solve the problem best?

This would partially solve my problem:

First of all, here would be my wishlist for logs that should just be debug logs for all scenarios: 

- The `Loading artifacts into URLClassLoader` logs [here](https://github.com/liquibase/liquibase/blob/v4.21.1/liquibase-maven-plugin/src/main/java/org/liquibase/maven/plugins/MavenUtils.java#L124). Actually, all the info logs in this class could IMHO be downgraded to debug. My reasoning: If you are at the point to consider checking the classpath, you would have probably used `mvn -X` already.
- The big liquibase ASCII art log also does not add anything (even the liquibase version is already logged by maven).
- The `Reading resource:` [logged here](https://github.com/liquibase/liquibase/blob/1832407ac6e8b9c7f86c2ba9f7758a983059b7a9/liquibase-standard/src/main/java/liquibase/changelog/DatabaseChangeLog.java#L744) -- if you are wondering why your changeset did not execute use debug.

Additionally, the log level of `Running Changeset:` [here](https://github.com/liquibase/liquibase/blob/1832407ac6e8b9c7f86c2ba9f7758a983059b7a9/liquibase-standard/src/main/java/liquibase/changelog/visitor/UpdateVisitor.java#L53C1-L56), `ChangeSet XXX ran successfully in` [here](https://github.com/liquibase/liquibase/blob/1832407ac6e8b9c7f86c2ba9f7758a983059b7a9/liquibase-standard/src/main/java/liquibase/changelog/ChangeSet.java#L737), and everything in between (`Table FOO created`, etc.) should be configurable. This is because with our setup, we generate the test database for our unittests with liquibase from scratch. We use the same liquibase scripts that we also use for our production DB. Thus, over the years we have accumulated **a lot** of changesets wich will be **all** executed each `mvn clean verify` run. This adds a lot of noice (the lines of liquibase loggs exceed the lines of test logs by a factor of at least 10).

Essentially, my goal is **for our maven test DB workflow** that liquibase only loggs a fixed number of lines, regardless of how big the classpath is and how many changesets are executed. However, changing the changeset log level for other use-cases (incremental DB updates) would certainly be a mistake.

liquibase-maven-plugin/src/main/java/org/liquibase/maven/plugins/MavenLog.java
```diff
@@ -0,0 +1,152 @@
...
+        private final int level;
+
+        Level(int level) {
+            this.level = level;
+        }
+
+        public boolean isAtLeast(Level other) {
+            return this.level <= other.level;
+        }
+    }
```
StevenMassaro, line 23: Is there a reason this is using a duplicate implementation of what is available in `java.util.logging.Level`?

StevenMassaro:
> @StevenMassaro from MDC point of view, is this change ok?

I don't think this should have an impact. However, if someone were to decrease the log level, like to SEVERE, they may miss out on some MDC logging that they ordinarily would get with INFO/DEBUG/etc. However, that's intended functionality.

liquibase-maven-plugin/src/main/java/org/liquibase/maven/plugins/MavenLog.java
mensinda: Yes:

- maven only supports the 4 log levels, and there is no `DEBUG` and `ERROR` in `java.util.logging.Level`
- The `isAtLeast` (or something equivalent) method also does not exist.

It would be possible to use `java.util.logging.Level`, but then `ERROR` would have to be renamed to `SEVERE` and `DEBUG` to `FINE.*`. This would be less user friendly, since maven uses `debug` and `error` names. Additionally, all the other levels would have to be filtered out. So, using `java.util.logging.Level` would result in even more code.

StevenMassaro: Thanks for the explanation, looks like you made the right decision here.

filipelautert (approved):
After all the reviews here this PR seems good to be moved along.

rberezen:
This pull request introduces a new property to the liquibase-maven-plugin, enabling the ability to change the log level using the 'logLevel' property.

- Tested manually.

APPROVED

rberezen (approved):

PR #4152, opened 2023-04-18
DAT-12945: support JSON logging in maven plugin

StevenMassaro:
## Impact


suryaaki2 (approved):

filipelautert (approved):
Some integration tests rely on maven , so as they are passing it is fine.

nvoxland (commented):
I didn't go with "request changes", but I worry about this direction of going our own way with custom logging rather than relying on what maven is providing us. This will likely make some use cases easier, but also break other ones that people expect to work.

So we should make sure we're really thinking through the implications of this change in strategy

StevenMassaro:
> I didn't go with "request changes", but I worry about this direction of going our own way with custom logging rather than relying on what maven is providing us. This will likely make some use cases easier, but also break other ones that people expect to work.
> 
> So we should make sure we're really thinking through the implications of this change in strategy before merging this

Good points. Since this is not an exclusive solution, we're going to go forward with this solution and come back to revisit the slf4j approach if it seems that it is desired by users. We'll be logging a new ticket for that and waiting for feedback from users.

PR #4169, opened 2023-04-24
Improve liquibase-core pom.xml handling (DAT-14546)

nvoxland:
## Impact

## Description

Liquibase-core pom now correctly builds its pom within maven rather than than using a custom "relase.pom.xml".

This should confuse maven less in "install" and "deploy" operations

## Things to be aware of

- Ensure the pom file deployed to github package manager matches what was released in 4.21.1 
- Ensure the pom file directly deployed matches the one in the jar

## Things to worry about

- Nothing


filipelautert:
Btw @nvoxland , would it fix our Maven 3.9.0 build issues related to javadocs?

sayaliM0412 (approved):

liquibase-cli/pom.xml
```diff
@@ -23,6 +23,8 @@
             <groupId>info.picocli</groupId>
             <artifactId>picocli</artifactId>
             <version>4.7.1</version>
+            <scope>compile</scope>
+            <optional>true</optional>
```
StevenMassaro, line 27: Is pico really an optional dependency? Seems pretty non-optional to me.

liquibase-core/pom.xml
```diff
@@ -108,26 +122,69 @@
                 </executions>
             </plugin>
 
-           <plugin>
-               <groupId>org.jacoco</groupId>
-               <artifactId>jacoco-maven-plugin</artifactId>
-               <executions>
-                   <execution>
-                       <id>report-aggregate</id>
-                       <phase>test</phase>
```
filipelautert, line 117: Great, this fixes Maven 3.9.0 build.

filipelautert (approved):

liquibase-cli/pom.xml
nvoxland: The reason that picocli got marked as optional is because we don't want liquibase-core to have that as a dependency. It currently isn't there at all, because it is optional for liquibase-core.
Ideally it would come through as optional vs. not there at all and my hope was that marking it as optional in cli would bubble through to liquibase-core that way, but it just doesn't show up at all.
So it's in sort of a "I wish it worked this way" state, but also very easy for people to accidentally break by removing that "optional" flag in liquibase-cli (since that would make sense) and then suddenly we are releasing liqubiase-core with a hard dependency on picocli which we don't want.
So I pushed a change that makes it more explicit that liquibase-core doesn't include picocli as a dependency.

StevenMassaro: I think the change d7307d98439c9ae17accd96d557e9ffbf530d5e3 makes your intention much clearer. Thanks!

suryaaki2 (approved):

PR #4171, opened 2023-04-24
[DAT-6608] Refactors DropAll command as CommandStep

filipelautert:
## Impact

## Description

DAT-6608 -  Refactors DropAll command as CommandStep

* Refactored to DropAllCommandStep. 
* Removed HUB
* DropAllCommand.java has been calling the wrong wrapper for 2 years and no one noticed that (calling `dropAllInternal `when the correct was `internalDropAll`) . Removing it.

## Things to notice
* Removed changelogFile parameter - it was used only for HUB. I prefered to remove this command line parameter so we don't cause any confusion, as users may think it may be used somehow by dropAll but in the end it is not used.


liquibase-standard/src/main/java/liquibase/command/core/DropAllCommandStep.java
```diff
@@ -1,49 +1,37 @@
...
+public class DropAllCommandStep extends AbstractCommandStep {
 
     public static final String[] COMMAND_NAME = {"dropAll"};
 
-    public static final CommandArgumentDefinition<String> USERNAME_ARG;
-    public static final CommandArgumentDefinition<String> PASSWORD_ARG;
-    public static final CommandArgumentDefinition<String> URL_ARG;
-    public static final CommandArgumentDefinition<String> DEFAULT_CATALOG_NAME_ARG;
-    public static final CommandArgumentDefinition<String> DEFAULT_SCHEMA_NAME_ARG;
+    private static final Logger LOG = Scope.getCurrentScope().getLog(DropAllCommandStep.class);
```
StevenMassaro, line 22: I want to say that getting the log as a class level static variable might be problematic for structured logging, because it might execute before we've had a chance to set up structured logging. I am not positive of this. Was this just a convenience change?

StevenMassaro (approved):

liquibase-standard/src/main/java/liquibase/command/core/DropAllCommandStep.java
filipelautert: That's good to know, let me fix it then. But there is sonar minor warning for this, 
so I it was force of habit -> https://rules.sonarsource.com/java/RSPEC-1312 . In web project it's common use to declare it as static too.

StevenMassaro: Well... I am not positive of this. Maybe it's worth experimenting with at some point.

filipelautert: No problems, and thanks for noticing that.

suryaaki2 (approved):

PR #4179, opened 2023-04-26
Sybase JDBC: Detecting current and some historic driver versions

mkarg:
## Impact

## Description

* Added the JDBC URI prefix and driver class name of SQL Anywhere 17 (taken from the original JDBC driver JAR) to allows users of the current version of the SQL Anywhere driver to use Liquibase more easily.
* Added some historic JDBC URI prefix and driver class names of previous releases of SQL Anywhere (taken from the former SQL Anyhwere lead architect's web site), to allow users of some older driver version to use Liquibase more easily.

## Things to be aware of

It is impossible to supper *all* historic versions of the JDBC driver, due to the following choices Sybase made:
* There is no historic information published for drivers older than v9.0.2.
* Driver v9.0.2..10.0.0 incorrectly used the previd `jdbc:odbc`, which originally was reserved by Sun's JDBC-ODBC bridge. Liquibase has no possibility to detect if Sun's bridge is to be used, or if Sybase's driver is to be used.
* Driver v11.0.1 shares the same JDBC URI prefix with later drivers, but uses a different driver class. Liquibase has no possibility to detect if the old or the new driver is to be used. The old driver is more than one decade old, so most users most likely will use the newer driver.

## Things to worry about

The SQL Anywhere Driver since v12 is supporting JDBC 4, so all modern drivers should work out of the box without further extending the list created by this PR.

## Additional Context

Website of Jason Hinsperger: https://blogs.sap.com/2014/05/02/connecting-to-sql-anywhere-using-jdbc-2/.


nvoxland:

Thanks for the fix! The code is what I'd expect it to be, and only impacts sybase-related code

Questions I have:
- None

Potential risks:
- None

- Nothing

filipelautert (approved):

XDelphiGrl:
@mkarg, thank you for the PR and super-extra-thanks for the excellent description. Your extensive research is appreciated!

@rberezen

suryaaki2 (approved):

PR #4193, opened 2023-04-28
DAT-14642 - Prevents Liquibase from incorrectly attempting to redeploy changesets  during checksum upgrade

filipelautert:
## Impact

## Description

Fix handling of ChangeLogHistoryServiceFactory when dealing with checksum upgrades.


wwillard7800 (commented):
Looks good

suryaaki2 (approved):

liquibase-standard/src/main/java/liquibase/Liquibase.java
```diff
@@ -1220,6 +1220,7 @@ public void checkLiquibaseTables(boolean updateExistingNullChecksums, DatabaseCh
         changeLogHistoryService.init();
         if (updateExistingNullChecksums) {
             changeLogHistoryService.upgradeChecksums(databaseChangeLog, contexts, labelExpression);
+            ChangeLogHistoryServiceFactory.getInstance().resetAll();
```
nvoxland, line 1223: This feels like a fix at the wrong level. The "resetAll" is sort of a "this code knows something the history service doesn't" hack when code outside the historyservice messes with the underlying data. But if you had just called "upgradeChecksums" the history service should know about what's going on.

That makes it feel like the actual bug is a problem with caching or something in the history service. Or, from doing a bit of debugging with daniel, a problem with how we're managing the history service singleton. 

There are possibly other places in our code and for sure places in other peoples code where they are calling methods on the history service and expecting it to work correctly. If the problem is that we're not seeing it work correctly unless we start calling resetAll, we can't be expecting us or others to fix their code, we should fix the service.

nvoxland (changes requested):

XDelphiGrl:
Passing functional tests for Postgres are here: [https://github.com/liquibase/liquibase-pro-tests/actions/runs/4872052466](https://github.com/liquibase/liquibase-pro-tests/actions/runs/4872052466). The issue was that there is a test validation fix for the `calculate-checksum` tests that was not in liquibase-pro-tests master when the build triggered. This run used the test PR with the validation fix and this PR branch of Liquibase.

PR #4197, opened 2023-05-01
Consume CORE and PRO packages from GitHub Package Manager (DAT-14615)

sayaliM0412:
## Impact

## Description

1. Consume CORE and PRO packages from GitHub Package Manager
2. Moving the logic of branch-build to build.yml so as to deploy the package which has successful builds to GPM
3. Adding a few dependencies to main pom.xml to like `snakeyaml`, `maven-resources-plugin` to handle the copying of project resources to the output directory, `maven-install-plugin` to install `tar.gz` 
4. Flatten the pom for maven-plugin due to shaded logic
5. Changes to java classes for `osgi framework` errors

## Things to worry about

Maybe `liquibase-maven-plugin` would throw error when merged for missing dependencies


filipelautert (approved):

.github/workflows/build.yml
```diff
@@ -355,22 +353,15 @@ jobs:
...
-          path: download/liquibase-commercial
-          repo: liquibase/liquibase-pro
+      # Download liquibase-commercial from GPM
 
       - name: Install Built Modules
         run: |
           (find . -name *-SNAPSHOT.jar -exec mvn -B org.apache.maven.plugins:maven-install-plugin:3.0.0-M1:install-file -Dfile={} \;)
           (find . -name *-SNAPSHOT-sources.jar -exec mvn -B org.apache.maven.plugins:maven-install-plugin:3.0.0-M1:install-file -Dclassifier=sources -Dfile={} \;)
-          mvn -B org.apache.maven.plugins:maven-install-plugin:3.0.0-M1:install-file -Dfile=download/liquibase-commercial/liquibase-commercial-0-SNAPSHOT.jar
+          mvn org.apache.maven.plugins:maven-dependency-plugin:3.5.0:get -Dartifact=org.liquibase:liquibase-commercial:0-SNAPSHOT -DrepoUrl=https://maven.pkg.github.com/liquibase/liquibase-pro
```
nvoxland, line 362: Version 0-SNAPSHOT shouldn't be on GPM to download this way? And since it is available from GPM which maven understands how to auto-download from you shouldn't have to manually get and install it at all.

So I think this line plus the next 2 are not needed at al anymore? Instead have a `liquibase-commercial.version` property which can be set in any of the maven calls that will be using liquibase-commercial?

pom.xml
```diff
@@ -75,6 +75,16 @@
         <argLine></argLine>
     </properties>
 
+    <repositories>
```
nvoxland, line 78: I don't think we want to add liquibase-pro's repo to the public source's pom? That is something we can uniquely do and so we shouldn't be setting everyone in the world to be using it by default. Instead add it to the settings.xml on the build server. 

Also on there we probably want/need a different id than `github` since that will maybe be expected to be the current projects github package repo vs. a one for a different dependency?

nvoxland (changes requested):

jnewton03:
@sayaliM0412 I think we'll need a desk check on this.  It sounds like this would not pull from GPM, is that correct?

sayaliM0412:
> @sayaliM0412 I think we'll need a desk check on this. It sounds like this would not pull from GPM, is that correct?

For now, only for [package](https://github.com/liquibase/liquibase/blob/e732cf53b95a503c68f3902134c40114a93420cf/.github/workflows/build.yml#L329) job, it would not, until we make changes to other extensions to look into GPM and not build artifact page.

jnewton03 (approved):

.github/workflows/build.yml
```diff
@@ -175,31 +174,34 @@ jobs:
...
+
       - name: Set up Maven
         uses: stCarolas/setup-maven@v4.5
         with:
           maven-version: ${{ env.MAVEN_VERSION }}
       - name: Build & Test Java 8
         if: ${{ matrix.java == 8}}
         run: |
-          mvn -B "-Dbuild.repository.owner=${{ needs.setup.outputs.thisRepositoryOwner }}" "-Dbuild.repository.name=${{ needs.setup.outputs.thisRepositoryName }}" "-Dbuild.branch=${{ needs.setup.outputs.thisBranchName }}" "-Dbuild.number=${{ github.run_number }}" "-Dbuild.commit=${{ needs.setup.outputs.thisSha }}" "-DtrimStackTrace=false" -pl '!liquibase-dist,!liquibase-cdi-jakarta' clean test package surefire-report:report
+          mvn -B "-Dbuild.repository.owner=${{ needs.setup.outputs.thisRepositoryOwner }}" "-Dbuild.repository.name=${{ needs.setup.outputs.thisRepositoryName }}" "-Dbuild.branch=${{ needs.setup.outputs.thisBranchName }}" "-Dbuild.number=${{ github.run_number }}" "-Dbuild.commit=${{ needs.setup.outputs.thisSha }}" "-DtrimStackTrace=false" -pl '!liquibase-dist,!liquibase-cdi-jakarta' clean test package surefire-report:report -DskipTests
```
StevenMassaro, line 194:
```suggestion
          mvn -B "-Dbuild.repository.owner=${{ needs.setup.outputs.thisRepositoryOwner }}" "-Dbuild.repository.name=${{ needs.setup.outputs.thisRepositoryName }}" "-Dbuild.branch=${{ needs.setup.outputs.thisBranchName }}" "-Dbuild.number=${{ github.run_number }}" "-Dbuild.commit=${{ needs.setup.outputs.thisSha }}" "-DtrimStackTrace=false" -pl '!liquibase-dist,!liquibase-cdi-jakarta' clean test package surefire-report:report
```

.github/workflows/build.yml
```diff
@@ -175,31 +174,34 @@ jobs:
...
       - name: Build & Test Java non-jdk-8
         if: ${{ matrix.java != 8}}
         run: |
-          mvn -B "-Dbuild.repository.owner=${{ needs.setup.outputs.thisRepositoryOwner }}" "-Dbuild.repository.name=${{ needs.setup.outputs.thisRepositoryName }}" "-Dbuild.branch=${{ needs.setup.outputs.thisBranchName }}" "-Dbuild.number=${{ github.run_number }}" "-Dbuild.commit=${{ needs.setup.outputs.thisSha }}" "-DtrimStackTrace=false" -pl '!liquibase-dist' clean test package surefire-report:report
+          mvn -B "-Dbuild.repository.owner=${{ needs.setup.outputs.thisRepositoryOwner }}" "-Dbuild.repository.name=${{ needs.setup.outputs.thisRepositoryName }}" "-Dbuild.branch=${{ needs.setup.outputs.thisBranchName }}" "-Dbuild.number=${{ github.run_number }}" "-Dbuild.commit=${{ needs.setup.outputs.thisSha }}" "-DtrimStackTrace=false" -pl '!liquibase-dist' clean test package surefire-report:report -DskipTests
```
StevenMassaro, line 198:
```suggestion
          mvn -B "-Dbuild.repository.owner=${{ needs.setup.outputs.thisRepositoryOwner }}" "-Dbuild.repository.name=${{ needs.setup.outputs.thisRepositoryName }}" "-Dbuild.branch=${{ needs.setup.outputs.thisBranchName }}" "-Dbuild.number=${{ github.run_number }}" "-Dbuild.commit=${{ needs.setup.outputs.thisSha }}" "-DtrimStackTrace=false" -pl '!liquibase-dist' clean test package surefire-report:report
```

.github/workflows/build.yml
```diff
@@ -338,36 +347,63 @@ jobs:
...
+                "id": "liquibase-pro",
+                "username": "liquibot",
+                "password": "${{ secrets.LIQUIBOT_PAT }}"
+              },
+              {
+                "id": "liquibase",
+                "username": "liquibot",
+                "password": "${{ secrets.LIQUIBOT_PAT }}"
+              }
+            ]
```
StevenMassaro, line 400: I would like to try removing these and seeing if the build still succeeds. I know that @sayaliM0412 and I tried that a couple times, but I do think that this is unnecessary, and I am a proponent of having a complete definition of the build in the pom, rather than in the build.yml. We should be able to take our pom over to another build system (think Jenkins), and it should just work.

I am willing to try removing these from the build.yml and seeing if the build passes. Alternatively we could put this in a new story, but it is so small that I'd like to see it done now, personally.

liquibase-maven-plugin/pom.xml
```diff
@@ -148,29 +145,61 @@
...
+<!--                                <id>default-deploy</id>-->
+<!--                                <phase>deploy</phase>-->
+<!--                                <goals>-->
+<!--                                    <goal>deploy-file</goal>-->
+<!--                                </goals>-->
+<!--                            </execution>-->
+<!--                        </executions>-->
+<!--                    </plugin>-->
+<!--                </plugins>-->
+<!--            </build>-->
```
StevenMassaro, line 202: Please remove this if it is not needed.

pom.xml
```diff
@@ -72,23 +90,24 @@
         <dependency.spock.version>1.3-groovy-2.4</dependency.spock.version>
         <junit.version>4.13.2</junit.version>
         <trimStackTrace>false</trimStackTrace>
+        <liquibase-pro.version>0-SNAPSHOT</liquibase-pro.version>
```
StevenMassaro, line 93: This doesn't appear to be used anywhere, and if it is not, it should be removed.

liquibase-maven-plugin/pom.xml
```diff
@@ -30,11 +58,6 @@
             <groupId>javax.xml.bind</groupId>
             <artifactId>jaxb-api</artifactId>
             <version>2.3.1</version>
-        </dependency>
-        <dependency>
-            <groupId>org.apache.maven</groupId>
-            <artifactId>maven-plugin-api</artifactId>
-            <version>${targetMavenVersion}</version>
```
StevenMassaro, line 37: Why was this removed?

liquibase-maven-plugin/pom.xml
```diff
@@ -45,22 +68,14 @@
             <scope>provided</scope>
         </dependency>
 
-        <dependency>
-            <groupId>org.apache.maven</groupId>
-            <artifactId>maven-artifact</artifactId>
-            <version>${targetMavenVersion}</version>
-            <scope>provided</scope>
-        </dependency>
```
StevenMassaro, line 53: Why was this removed?

StevenMassaro (changes requested):

pom.xml
sayaliM0412: I think we are using it in build.yml in step:  Publish tar.gz package to GPM

liquibase-maven-plugin/pom.xml
sayaliM0412: I think it was done as a part of this commit -> [here](https://github.com/liquibase/liquibase/commit/868d09e7868873358f22ad6127f82fbd1620dc81#diff-56e64c6e32ac9a2a4dd259dd9c572ef3342344f575bc38296d5647c7f5b2c341) to flatten the pom.

liquibase-maven-plugin/pom.xml
sayaliM0412: same for this . I think it was done as a part of this commit -> [here](https://github.com/liquibase/liquibase/commit/868d09e7868873358f22ad6127f82fbd1620dc81#diff-56e64c6e32ac9a2a4dd259dd9c572ef3342344f575bc38296d5647c7f5b2c341) to flatten the pom

StevenMassaro (approved):

suryaaki2 (approved):

.github/workflows/build-branch.yml
```diff
@@ -39,6 +39,5 @@ jobs:
       - name: Publish package
         run: |
           mvn -B clean deploy -DskipTests=true
```
StevenMassaro, line 41:
```suggestion
          mvn -B clean deploy -pl '!liquibase-dist' -DskipTests=true
```

PR #4202, opened 2023-05-02
For SQL Server, if a delimiter has not been specified, then add one. DAT-13998

wwillard7800:
## Impact

## Description

For SQL Server, if a delimiter has not been specified, then add one.

https://learn.microsoft.com/en-us/sql/t-sql/language-elements/transact-sql-syntax-conventions-transact-sql?redirectedfrom=MSDN&view=sql-server-ver16

`Transact-SQL statement terminator. Although the semicolon isn't required for most statements in this version of SQL Server, it will be required in a future version.`

### DIFF
--- a/liquibase-standard/src/main/java/liquibase/changelog/ChangeSet.java
+++ b/liquibase-standard/src/main/java/liquibase/changelog/ChangeSet.java
@@ -714,8 +714,7 @@ public class ChangeSet implements Conditional, ChangeLogChild {
 
                         addSqlMdc(change, database, false);
 
-                        List<SqlVisitor> sqlChangeVisitors = addSqlVisitorsIfNecessary(database, change);
-                        database.executeStatements(change, databaseChangeLog, sqlChangeVisitors);
+                        database.executeStatements(change, databaseChangeLog, sqlVisitors);
                         log.info(change.getConfirmationMessage());
                         if (listener != null) {
                             listener.ran(change, this, changeLog, database);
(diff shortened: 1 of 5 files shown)


liquibase-standard/src/main/java/liquibase/changelog/ChangeSet.java
```diff
@@ -776,6 +776,22 @@ public ExecType execute(DatabaseChangeLog databaseChangeLog, ChangeExecListener
         return execType;
     }
 
+    //
+    // If this is for MSSQL and we have an AbstractSqlChange
+    // then we add an appending SQL visitor that will add
+    // on the end delimiter if necessary
+    //
+    private List<SqlVisitor> addSqlVisitorsIfNecessary(Database database, Change change) {
+        if (! (database instanceof MSSQLDatabase && change instanceof AbstractSQLChange) || ((AbstractSQLChange)change).getEndDelimiter() != null) {
```
nvoxland, line 785: This isn't just an AbstractSQLChange thing, is it? We're wanting all changes on mssql to have an end delimter? Removing that part of the `if` makes it more globally applicable and so less surprising to users, and also keeps us from adding more dependencies from database/change generic code like ChangeSet to implementation details.

Also, maybe better to change this from a "if not this case, just return getSqlVisitors()" and instead have it as "if this case, add to getSqlVisitors()"? Maybe easier to read now, and easier to maintain later if/when we want to add other visitors for other databases?

Finally, in theory the ChangeSet shouldn't have dependencies on MSSQL either, since that's more of an implementation. Maybe instead of adding this MSSQL-specific if block here, the MSSQLDatabase.executeStatements() should have the logic of adding the delimiter?

filipelautert (approved):

suryaaki2 (approved):

Issue #4184, opened 2023-04-27
"addDefaultValue" with attributes "schemaName" and "defaultValueSequenceNext" creates invalid checksum on rollback


moberwasserlechner:
### Search first

- [X] I searched and no similar issues were found

### Description

Using the "addDefaultValue" in combination with "schemaName" and "defaultValueSequenceNext" on PostgreSQL database results invalid checksums on rollback.

```
Caused by: liquibase.exception.ValidationFailedException:
```
Stacktrace
```
        at liquibase.changelog.DatabaseChangeLog.validate(DatabaseChangeLog.java:357)
        at liquibase.Liquibase$10.run(Liquibase.java:1503)
        at liquibase.Scope.lambda$child$0(Scope.java:187)
        at liquibase.Scope.child(Scope.java:196)
        at liquibase.Scope.child(Scope.java:186)
        at liquibase.Scope.child(Scope.java:165)
        at liquibase.Liquibase.runInScope(Liquibase.java:2714)
        at liquibase.Liquibase.rollback(Liquibase.java:1481)
        at liquibase.integration.commandline.Main.doMigration(Main.java:1922)
        at liquibase.integration.commandline.Main$1.lambda$run$0(Main.java:412)
        at liquibase.Scope.lambda$child$0(Scope.java:187)
        at liquibase.Scope.child(Scope.java:196)
        at liquibase.Scope.child(Scope.java:186)
        at liquibase.Scope.child(Scope.java:165)
        at liquibase.integration.commandline.Main$1.run(Main.java:411)
        at liquibase.integration.commandline.Main$1.run(Main.java:228)
        at liquibase.Scope.child(Scope.java:196)
        at liquibase.Scope.child(Scope.java:172)
        at liquibase.integration.commandline.Main.run(Main.java:228)
        at liquibase.command.AbstractCliWrapperCommandStep.run(AbstractCliWrapperCommandStep.java:32)
        at liquibase.command.CommandScope.execute(CommandScope.java:196)
```

### Steps To Reproduce

Using this changelog we wanted to rename a table and its sequence.
```
  <changeSet id="API-101-rename_postgres_table" author="Michael Oberwasserlechner">
    <renameTable oldTableName="api_old_postgres_table" newTableName="api_new_table" schemaName="non_public_schema" />
    <renameSequence oldSequenceName="api_old_postgres_table_id_seq" newSequenceName="api_new_table_id_seq" schemaName="non_public_schema" />
[...]

PR #4203, opened 2023-05-02
Fix addDefaultValue checksum calculation when providing schema name

filipelautert:
## Impact

## Description

Make sure we create new objects instead of modifying the existing ones when processing addDefaultValue change type.

## Things to be aware of

It changes checksum value so it must be merged in the same release as https://github.com/liquibase/liquibase/pull/3914 .


XDelphiGrl (approved):
In the very specific case of a `addDefaultValue` changeset that specifies the `schemaName` and `defaultValueSequenceNext`, `update` and `rollback` commands were calculating different checksums. This PR addresses this discrepancy for newly deployed changesets; existing deployed changesets will continue to throw a validation error until this PR is combined with the overall checksum update from v8 to v9. Both the checksum upgrade and this fix will be available in the same release.
* Manual testing passed [https://github.com/liquibase/liquibase/issues/4184#issuecomment-1533556760](https://github.com/liquibase/liquibase/issues/4184#issuecomment-1533556760).
* Functional test suite passing.
* The SQL Server test harness failures are due to a known test validation issue addressed in a different PR; these failures can be disregarded for the purposes of this PR.

APPROVED

Issue #4129, opened 2023-04-13
Properties are not being substituted


joserebelo:
### Search first

- [X] I searched and no similar issues were found

### Description

Properties are not being substituted when used on changes using the `${parameter}` format - they get passed to the changes as the literal string `"${parameter}"`.

git bisect tracked the issue to 6f8a113ad

### Steps To Reproduce

Set a parameter on the defaults file:

```
parameter.a_parameter=asd
```

Use the parameter on a change:

```xml
<changeSet author="joserebelo" id="test.change">
    <createTable tableName="${a_parameter}">
        <column name="user_id" type="VARCHAR(40)"/>
    </createTable>
</changeSet>
```

### Expected/Desired Behavior

It was expected that a table with the name `"asd"` would have been created. However, a table with the literal name `"${a_parameter}"` gets created.

The same issue applies to args in the `executeCommand` change and params on customChanges.

### Liquibase Version

4.21.0

### Database Vendor & Version

PostgreSQL 9.6

### Liquibase Integration

cli

### Are you willing to submit a PR?

- [ ] I'm willing to submit a PR (Thank you!)

astiob:
Came here while debugging the same issue. Some more info:

Works when specifying properties on the command line as `-Dname=value`. Fails when specifying them in the defaults file (liquibase.properties) as `parameter.name=value`.

The failure affects both the SQL generated (as evidenced by `liquibase update-sql` run against a blank database/schema) and the changeset checksum (as evidenced by the failures we’re getting from `update` and by manual `liquibase calculate-checksum`).

gallipi:
Version 4.21.1 is also affected.
Database: Oracle 19c

Liquibase throws the exception `Oracle non supported sql92 token`  when trying to parse the property.

alexandre-cartapanis:
Same with 4.21.1 and postgresql 14

nvoxland:
It's only when you are setting properties from the `liquibase.properties` file, correct?  Or are there other ways anyone is setting them and it's also not working?

XDelphiGrl:
I validated the fix for this issue using a changelog that has changesets that use parameters defined in the changelog, parameters defined in the liquibase.properties file and a table that uses a combination of changelog and liquibase.properties parameters. I tested using the CLI against a Postgres 12.6; the Liquibase build is [https://github.com/liquibase/liquibase/actions/runs/4874060226](https://github.com/liquibase/liquibase/actions/runs/4874060226).

[...]

PR #4204, opened 2023-05-03
Load properties from defaults-file fixing variable replacement for refactored commands

filipelautert:
## Impact

## Description

Before command refactoring, `LiquibaseCommandLine `would execute` Main class` that has a complex way of loading properties values from files (it starts at method` parseDefaultPropertyFiles()` and goes on). 
After command refactoring, `LiquibaseCommandLine `is calling the commands directly without passing by `Main` - so this logic has been lost and properties values were not being provided when `DatabaseChangeLog` is created.

But as `LiquibaseCommandLine` already loads the default-file as a provider into `LiquibaseConfiguration `, instead of copying all the logic from Main this PR just uses the already existing information and populates the parameters from there.

## Additional Context

Fixes bug introduced by command refactoring.

### DIFF
--- a/liquibase-standard/src/main/java/liquibase/command/core/helpers/DatabaseChangelogCommandStep.java
+++ b/liquibase-standard/src/main/java/liquibase/command/core/helpers/DatabaseChangelogCommandStep.java
@@ -8,6 +8,9 @@ import liquibase.changelog.ChangeLogHistoryServiceFactory;
 import liquibase.changelog.ChangeLogParameters;
 import liquibase.changelog.DatabaseChangeLog;
 import liquibase.command.*;
+import liquibase.configuration.ConfigurationValueProvider;
+import liquibase.configuration.LiquibaseConfiguration;
+import liquibase.configuration.core.DefaultsFileValueProvider;
 import liquibase.database.Database;
 import liquibase.exception.LiquibaseException;
 import liquibase.lockservice.LockService;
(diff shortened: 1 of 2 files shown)


suryaaki2:
@filipelautert Can we write any unit tests?

wwillard7800 (approved):

XDelphiGrl:
Rerunning test-harness on develop with the fix for the MSSQL tests here: [https://github.com/liquibase/liquibase-test-harness/actions/runs/4873636204](https://github.com/liquibase/liquibase-test-harness/actions/runs/4873636204)

liquibase-standard/src/main/java/liquibase/command/core/helpers/DatabaseChangelogCommandStep.java
```diff
@@ -124,10 +128,25 @@ public void cleanUp(CommandResultsBuilder resultsBuilder) {
      * Add java property arguments to changelog parameters
      * @param changeLogParameters the changelog parameters to update
      */
-    public void addJavaProperties(ChangeLogParameters changeLogParameters) {
+    private void addJavaProperties(ChangeLogParameters changeLogParameters) {
```
XDelphiGrl, line 131: Will changing this from `public` to `private` represent an API-breaking change?

filipelautert: Should not break anything.. but I rolled back the change so we don't miss the focus here :)

suryaaki2 (approved):

XDelphiGrl (approved):
Liquibase changelog parameters act as variables, allowing Liquibase to substitute different values based on the parameter. These parameters can be defined inline in the changelog or configured as `parameter.paramName:paramValue` in the liquibase.properties file. This PR fixes a bug in property substitution when parameters are defined in the liquibase.properties file.
* Manual validation passed [https://github.com/liquibase/liquibase/issues/4129#issuecomment-1533428373](https://github.com/liquibase/liquibase/issues/4129#issuecomment-1533428373)


APPROVED

PR #4210, opened 2023-05-04
Fix null md5sum issues during checksum upgrades

filipelautert:
## Impact

## Description

When updating checksums make sure to use the stored file path instead of the generated one as we changed the way that we store paths during the years.

### DIFF
--- a/liquibase-standard/src/main/java/liquibase/changelog/AbstractChangeLogHistoryService.java
+++ b/liquibase-standard/src/main/java/liquibase/changelog/AbstractChangeLogHistoryService.java
@@ -74,8 +74,6 @@ public abstract class AbstractChangeLogHistoryService implements ChangeLogHistor
                         Scope.getCurrentScope().getLog(getClass()).fine(
                                 "Updating null or out of date checksum on changeSet " + changeSet + " to correct value"
                         );
-                        // to allow switching between saving normalized or configured path
-                        changeSet.setStoredFilePath(ranChangeSet.getStoredChangeLog());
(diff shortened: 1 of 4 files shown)


MalloD12 (approved):
**Approved**.

This is looking good to me. We have been manually testing this fix  with Filipe and seems to be working ok. We have also discussed and agree to create a separate ticket for adding some test coverage to make sure `StoreFilePath` is correctly set, in order to move forward this is and work on some other failing commands.

XDelphiGrl (commented):
In Liquibase versions prior to 4.x, DATABASE.FILENAME values were calculated differently. In particular, the changelog filepath from Spring Boot or other integrations were entered into the DATABASE.FILENAME value with `classpath:` prepended to the filename.  Liquibase 4.x normalized filepaths; these newer versions of Liquibase enter the changelog filepath from Spring Boot or other integrations without the prepending `classpath:`.  This caused a bug where the upgrade from v8 to v9 checksums set DATABASE.FILENAME to NULL for rows with `classpath:`. Once set to NULL, those values would remain NULL, behaving as if the changesets were no longer in the changelog.

This PR addresses the bug due to the conflict in previous path standardization and checksum upgrades.

* Manually tested to ensure proper v9 checksums for DATABASE.FILENAME values prepended with `classpath:`. 

APPROVEDas if the changesets were no longer in the changelog.

This PR addresses the bug due to the conflict in previous path standardization and checksum upgrades.

* Manually tested to ensure proper v9 checksums for DATABASE.FILENAME values prepended with `classpath:`. 

APPROVED

PR #4212, opened 2023-05-04
DAT-14686

sayaliM0412:
Cleanup branch builds for other extensions
## Impact

## Description

In GitHub actions we currently are only cleaning liquibase-core packages. We want to be able to clean other snapshot packages such as “Intergation-test, jakarta,cdi” as well.

### DIFF
--- a/.github/workflows/cleanup-branch-builds.yml
+++ b/.github/workflows/cleanup-branch-builds.yml
@@ -6,9 +6,23 @@ on:
     branches:
       - '**DAT-**'
 jobs:
+  setup:
+    runs-on: ubuntu-22.04
+    outputs:
+      packages_to_delete: ${{ steps.get_packages.outputs.packages_to_delete }}
+    steps:
+      - id: get_packages 
+        run: |
+          packages_to_delete=($(mvn -B exec:exec -Dexec.executable=echo -Dexec.args='###MODULE_GAV### ${project.groupId}:${project.artifactId}' | grep '###MODULE_GAV### ' | cut -f2 -d' ' ))
+          echo '::set-output packages_to_delete=${packages_to_delete}'
+
   delete-package:
+    needs: [ setup ]
     name: Delete Github Package for Branch
     runs-on: ubuntu-22.04
+    strategy:
+      matrix:
+        value: ${{fromJson(needs.setup.outputs.packages_to_delete)}}
     steps:
       # Get version id(s) based on version name
     - uses: castlabs/get-package-version-id-action@v2.1
@@ -22,7 +36,7 @@ jobs:
         # Name of the package.
         # Defaults to an empty string.
         # Required if `package-version-ids` input is not given.
-        package-name: org.liquibase.liquibase-core
+        package-name: ${{ matrix.value }}
 
         # The number of latest versions to keep.
         # This cannot be specified with `num-old-versions-to-delete`. By default, `num-old-versions-to-delete` takes precedence over `min-versions-to-keep`.


jnewton03 (approved):
This looks okay, but we definitely want to test it.  We'll want to make sure there is always at least 1 version available in packages.  For example, since the trigger for this cleanup is "On Release" then once we release liquibase, is that going to delete all versions of liquibase-extension-testing from github packages?  If so, then builds will start to fail because they can no longer find that package.

filipelautert (approved):
I think it's fine, but I'm adding Steve as reviewer as he always has good insights about changes in those flows .

sayaliM0412:
> @sayaliM0412 This looks okay, but we definitely want to test it. We'll want to make sure there is always at least 1 version available in packages. For example, since the trigger for this cleanup is "On Release" then once we release liquibase, is that going to delete all versions of liquibase-extension-testing from github packages? If so, then builds will start to fail because they can no longer find that package.

I think with "Delete artifact based off of branch and commit SHA" step in `cleanup-master-builds.yml`, I think it will only delete the packages which have been built at workflow "Build Main/Master SNAPSHOT" in `build-main.yml`, due to the fact that it's on `cleanup-master-builds.yml` runs on `types: [ published ]` . 
Also, we have not mentioned to delete these packages '^DAT-.*-SNAPSHOT$'.

.github/workflows/cleanup-branch-builds.yml
```diff
@@ -6,30 +6,28 @@ on:
...
-        # This cannot be specified with `num-old-versions-to-delete`. By default, `num-old-versions-to-delete` takes precedence over `min-versions-to-keep`.
-        # When set to 0, all deletable versions will be deleted.
-        # When set greater than 0, all deletable package versions except the specified number will be deleted.
-        min-versions-to-keep: 0
-
-        # Can be a single package version id, or a comma separated list of package version ids.
-        # Defaults to an empty string.
-        package-version-ids: "${{ steps.version.outputs.ids }}"
+      run: |
+        packages_to_delete=($( mvn -B exec:exec -Dexec.executable=echo -Dexec.args='###MODULE_GAV### ${project.groupId}.${project.artifactId}' | grep '###MODULE_GAV### ' | cut -f2 -d' ' ))
```
StevenMassaro, line 21: Is there an alternative way of generating this list besides this call? I fear that this will be difficult to debug.

.github/workflows/cleanup-branch-builds.yml
```diff
@@ -6,30 +6,28 @@ on:
...
+        echo $packages_to_delete
+        for package_name in "${packages_to_delete[@]}"
+        do
+          for version_id in ${{ steps.version.outputs.ids }}
+          do
+            curl -X DELETE \
+            -H "Accept: application/vnd.github+json" \
+            -H "Authorization: Bearer ${{ secrets.LIQUIBOT_LIQUIBASE_PAT }}"\
+            -H "X-GitHub-Api-Version: 2022-11-28" \
+            https://api.github.com/user/packages/maven/$package_name/versions/$version_id
```
StevenMassaro, line 31: Is it possible to continue to use the existing actions/delete-package-versions@v3 action? Seems like that might be a bit more user-friendly than a curl call here.

.github/workflows/cleanup-branch-builds.yml
sayaliM0412: I think, the main issue/concern here is not getting the list of packages but to ensure the package versions we are deleting are the right ones.

.github/workflows/cleanup-branch-builds.yml
sayaliM0412: Unfortunately this action : actions/delete-package-versions@v3 only takes a single package and not a "list of packages"

StevenMassaro: I guess I am just a tad afraid of the complexity here. Have we looked to see if any other similar github actions can be used here?

delete-package-versions can "Ignore version(s) from deletion through regex"

StevenMassaro:
I could be making a bigger deal out of the complexity here than I need to. One more thing I wanted to share, this looks like it could be useful:

https://jordanbeandev.com/how-to-write-a-loop-in-a-github-action-yaml-file/

.github/workflows/cleanup-branch-builds.yml
sayaliM0412: for the ignore versions we previously had put `'^DAT-.*-SNAPSHOT$'`. In the script I am deleting `master-SNAPSHOT` and `$git_sha` we get from "based off of branch and commit SHA"

sayaliM0412:
@StevenMassaro : I do have my reservations as well, I will talk with Jake about those and maybe can come up with a better solution or tweak the existing one.

StevenMassaro (approved):

jnewton03 (approved):
LGTM 🚀

sayaliM0412:
Had a talk with @suryaaki and I will merge this PR after the merge freeze is lifted.

suryaaki2 (approved):

PR #4217, opened 2023-05-05
execute-sql command refactor (DAT-6609)

abrackx:
## Impact

## Description
Refactors `execute-sql` to use command framework. Removes internal and deprecated `execute-sql` commands. Adds command tests.

## Things to be aware of
`execute-sql` should not touch `Main` during cli execution now.


StevenMassaro (approved):

liquibase-standard/src/main/java/liquibase/integration/commandline/Main.java
```diff
@@ -1516,18 +1516,6 @@ protected void doMigration() throws Exception {
                 outputWriter.write(result);
                 outputWriter.flush();
                 return;
-            } else if (COMMANDS.EXECUTE_SQL.equalsIgnoreCase(command)) {
```
filipelautert, line 1519: Don't we still need the ExecuteSql somewhere in Main for backwards compatibility?

abrackx: Hm, you may be right. Since we don't have any other APIs besides main exposing this and we don't have a maven mojo or even a `Liquibase#executeSql` method, `Main` was probably the only way to utilize this feature outside of the CLI. I'll add this back in. 

- [ ] Revert main changes, add comment explaining the preferred method is to directly use command framework

filipelautert: Last week I saw a comment from community people that were migrating from 3.x and they were calling Main from their code in a kind of Liquibase wrapper  🙃

filipelautert (approved):

PR #4262, opened 2023-05-15
DAT-6632: refactor snapshot-reference command

StevenMassaro:
## Impact

### DIFF
--- a/liquibase-standard/src/main/java/liquibase/integration/commandline/Main.java
+++ b/liquibase-standard/src/main/java/liquibase/integration/commandline/Main.java
@@ -1533,19 +1533,6 @@ public class Main {
                 outputWriter.write(output);
                 outputWriter.flush();
                 return;
-            } else if (COMMANDS.SNAPSHOT_REFERENCE.equalsIgnoreCase(command)) {
-                CommandScope snapshotCommand = new CommandScope("internalSnapshot");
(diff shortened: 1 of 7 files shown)


liquibase-standard/src/main/java/liquibase/integration/commandline/Main.java
```diff
@@ -1532,19 +1532,6 @@ protected void doMigration() throws Exception {
                 String output = (String) results.getResult("output");
                 outputWriter.write(output);
                 outputWriter.flush();
-                return;
```
filipelautert, line 1535: Shouldn't we keep this in Main for backward compatibility?

filipelautert (approved):

liquibase-standard/src/main/java/liquibase/integration/commandline/Main.java
StevenMassaro: I am open to keeping it. It is in a protected method, and not in the usual Liquibase class in which we have kept around deprecated methods. For those reasons I thought deleting it would be safe(r). What do you think @filipelautert ?

filipelautert: I saw some people calling update method from main in their code. I really doubt someone may be using "SnapshotReference" ... 
So maybe we can drop it and if someone complains about it we can share the new steps.

suryaaki2 (approved):

PR #929, opened 2019-10-22
Fix #CORE-3493 Inconsistent resolution of ChangeLogParameters in liqu…

Holly-Marie:
Solution to the bug report CORE-3493.

I tested it with my changesets and it works as expected.
If required I could look into writing a few unit tests for this change as well.

┆Issue is synchronized with this [Jira Story](https://datical.atlassian.net/browse/LB-44) by [Unito](https://www.unito.io/learn-more)
┆Fix Versions: Liquibase 3.10.2

PR #4267, opened 2023-05-16
Changing the order to get the parameters, this way we can use inherit…

airon-assustadus:
…s fields

## Impact

## Description

Getting the local parameters from the last input instead the first one, this way we can override it and be able to use inherits

Scenary:
We want to have inherits, here is the base changeLog: `0000-BaseCreation.xml`

```xml
<databaseChangeLog
                            http://www.liquibase.org/xml/ns/dbchangelog/dbchangelog-4.3.xsd" logicalFilePath="base.xml">

    <changeSet author="${table.author}" id="createTable_${table.name}" logicalFilePath="create_table_${table.name}.xml">
        <createTable schemaName="${table.schema}" tableName="${table.name}">
            <column name="id" type="bigint" autoIncrement="true">
                <constraints primaryKey="true"/>
            </column>
            <column name="createDate" type="datetime">
            </column>
            <column name="updateDate" type="datetime">
                <constraints nullable="true"/>
            </column>
            <column name="removeDate" type="datetime">
                <constraints nullable="true"/>
            </column>
        </createTable>
    </changeSet>
</databaseChangeLog>
```

Here is the changeLog first table creation: `0001-account-bank-enum.xml`

```xml
<databaseChangeLog
                            http://www.liquibase.org/xml/ns/dbchangelog/dbchangelog-4.3.xsd" logicalFilePath="base.xml">

      <property name="table.schema" value="public" global="false"/>
      <property name="table.name" value="account_bank_enum" global="false"/>
      <property name="table.author" value="AironTeixeira" global="false"/>

       <include file="0000-BaseCreation.xml" relativeToChangelogFile="true"/>
</databaseChangeLog>
```

Here is the changeLog second table creation: `0002-financial-institution-enum.xml`
```xml
<databaseChangeLog
                            http://www.liquibase.org/xml/ns/dbchangelog/dbchangelog-4.3.xsd" logicalFilePath="base.xml">

      <property name="table.schema" value="public" global="false"/>
      <property name="table.name" value="financial_institution_enum" global="false"/>
      <property name="table.author" value="AironTeixeira" global="false"/>

       <include file="0000-BaseCreation.xml" relativeToChangelogFile="true"/>
</databaseChangeLog>
```

Here is the master file:

```xml
<databaseChangeLog
                                           http://www.liquibase.org/xml/ns/dbchangelog/dbchangelog-4.3.xsd">

    <include file="0001-account-bank-enum.xml" relativeToChangelogFile="true"/>
    <include file="0002-financial-institution-enum.xml" relativeToChangelogFile="true"/>


</databaseChangeLog>
``` 

Without my change the second file was never been executed


airon-assustadus:
Hello @nvoxland 

Yes, instead First wins it willl be change to last wins like css does.
Can we move ahead with it?

kevin-atx:
@nvoxland says that the goal is to verify that the global="false" behavior is functioning as designed, or not. The goal is to determine if this feature is broken and subsequently verify the fix.

suryaaki2 (approved):

XDelphiGrl:
@airon-assustadus, thank you for the excellent description and including changelogs to demonstrate the issue! As a QA, I appreciate this very much!

CC @rberezen

XDelphiGrl (approved):
This PR allows inheritance of changelog properties. Changelog property inheritance is useful in cases where you wish to create multiple objects with the same basic structure and substitute only the differences using changelog properties, for example, a table with the same columns and datatypes but with different names or on different schemas. 
* Functional tests had a couple of failures due to a GitHub API Rate limit and a Postgres connection error. Neither of these issues is related to this PR and they can be disregarded.
* Test harness execution passed.

APPROVED

martinsefcik:
@nvoxland I think it was breaking change (not mentioned) in release 4.12.0 where support for local properties visibility in included script was intentionally removed (it was introduced in #2656 and intention can be seen e.g. on the [following line](https://github.com/liquibase/liquibase/pull/2656/files#diff-3a3e312df551992c399205e47b782d18119d917f09552e9b95ca8f0a9399b32bR27)) 

Anyway this line:

> * Local properties are  only available in the change log that they are defined in -- not even in changelogs "included" by the file that defines the property.

is still there in the code so if you returned support for local properties visibility in included scripts back in this PR then I guess this line should be removed/changed because otherwise it is confusing (@XDelphiGrl , @suryaaki2).

When we can expect this PR will be released? As this not mentioned breaking change introduced in 4.12.0 is blocking us to finish upgrade of our project to Spring Boot 3/3.1 where Liquibase 4.20.0 is used and it is not possible to downgrade Liquibase to 4.15.0 and below due to incompatibility with `LiquibaseAutoConfiguration`.

filipelautert:
Hello @martinsefcik - this will be included in the next release that will be out hopefully next Monday.

boolivar:
Unfortunately this is also a breaking change and not the fix for #2656.

[Documentation](https://docs.liquibase.com/concepts/changelogs/property-substitution.html) clearly says:
> Once a property has been set, it cannot be changed. Also, only the first definition is used, others are skipped.

This change effectively inverts logic by picking last value for local properties e.g. for
```xml
<property name="table.name" value="account_bank_enum" global="false"/>
<property name="table.name" value="financial_institution_enum" global="false"/>
```
`table.name` property will be `financial_institution_enum` instead of expected `account_bank_enum`.

You might be wondering why somebody would write properties that way? Let's take a look at more useful example:
```xml
<property name="table.name" value="account_bank_enum" global="false" dbms="oracle"/>
<property name="table.name" value="financial_institution_enum" global="false"/>
```
According to the documentation `table.name` should be `account_bank_enum` for oracle and `financial_institution_enum` for any other db but now it is overriden by latest value.

Hierarchical search for local properties was introduced in liquibase v3.10.2 by #929 and was broken in v4.12.0 by #2656.

So my proposal is to bring hierarchical property search back. I can create a PR.

cc @nvoxland

filipelautert:
@boolivar your example is indeed very useful, and it shows that this fix went way beyond the local scope for variables. Please feel free to submit a pr to fix it !

boolivar:
@filipelautert any concerns on PR ^

PR #4271, opened 2023-05-16
Fix Custom ChangeExecListener Class Cast Exception

filipelautert:
## Impact

## Description

The code  assumes that we have a DefaultChangeExecListener, both for reset() and also for the structured logging changes - but should be using the ChangeExecListener interface. Changed the code to allow that producing fewer output if the class is different from DefaultChangeExecListener . This kind of "problem" may only happen if someone is providing a custom class.


StevenMassaro (approved):
I think we might do this casting in other classes, can you do a search for that?

suryaaki2 (approved):

PR #4278, opened 2023-05-17
refactor unexpected changesets command (DAT-6637)

StevenMassaro:
## Impact

### DIFF
--- a/liquibase-integration-tests/src/test/resources/liquibase/extension/testing/command/unexpectedChangesets.test.groovy
+++ b/liquibase-integration-tests/src/test/resources/liquibase/extension/testing/command/unexpectedChangesets.test.groovy
@@ -30,7 +30,7 @@ Optional Args:
   username (String) Username to use to connect to the database
     Default: null
   verbose (Boolean) Verbose flag
-    Default: null
+    Default: false
 """
 
     run "Happy path", {


liquibase-standard/src/main/java/liquibase/Liquibase.java
```diff
@@ -1178,70 +1178,26 @@ public void reportStatus(boolean verbose, Contexts contexts, LabelExpression lab
...
-                        out.append("     ").append(ranChangeSet.toString()).append(StreamUtil.getLineSeparator());
-                    }
-                }
-            }
-
-            out.flush();
-        } catch (IOException e) {
-            throw new LiquibaseException(e);
-        }
+        // todo need to fill out command here
```
StevenMassaro, line 1200: I need to do this part! Oops

liquibase-standard/src/main/java/liquibase/Liquibase.java
```diff
@@ -1178,71 +1178,38 @@ public void reportStatus(boolean verbose, Contexts contexts, LabelExpression lab
...
     public void reportUnexpectedChangeSets(boolean verbose, String contexts, Writer out) throws LiquibaseException {
         reportUnexpectedChangeSets(verbose, new Contexts(contexts), new LabelExpression(), out);
     }
 
+    @Deprecated
     public void reportUnexpectedChangeSets(boolean verbose, Contexts contexts, LabelExpression labelExpression,
                                            Writer out) throws LiquibaseException {
         changeLogParameters.setContexts(contexts);
         changeLogParameters.setLabels(labelExpression);
-        checkLiquibaseTables(false, getDatabaseChangeLog(true), null, null);
```
filipelautert, line 1217: Seems we lost checkLiquibaseTables when moving the code to the command. In fact a lot of steps use it, couldn't we create a helper for it?

StevenMassaro: The command calls that method, here: https://github.com/liquibase/liquibase/blob/master/liquibase-standard/src/main/java/liquibase/command/core/helpers/DatabaseChangelogCommandStep.java#L85

filipelautert (approved):

liquibase-standard/src/main/java/liquibase/Liquibase.java
filipelautert: Thanks, I missed that.

PR #4306, opened 2023-05-24
Fixes #3625 JAVA_HOME Path when parenthesis are present

### LINKED ISSUE #3625: Could not use liquibase cli on windows if there is a parenthesis in path
## Environment

**Liquibase Version**: 4.18.0

**Database Vendor & Version**: n/a

**Operating System Type & Version**: Windows 10 Pro

**Infrastructure Type/Provider**: n/a

## Description

When I try to run liquibase under cmd on Windows I get an error.
After investigation it seems this is due to the parenthesis in path.

## Steps To Reproduce

1. Download Liquibase CLI > Archive Download (zip)
2. Unzip the archive in the folder "Program Files (x86)"
3. Run liquibase

## Actual Behavior

```console
C:\Program Files (x86)\liquibase-4.18.0>liquibase
\liquibase-4.18.0\\jre was unexpected at this time.
```

## Expected/Desired Behavior

We must see liquibase banner.

JulienMa94:
## Impact

## Description

**Fixes #3625 JAVA_HOME Path when parenthesis are present**

Inside the liquibase.bat file the JAVA_HOME could not be processed due to parenthesis. The issue is fixed when the path is wrapped by quotes.

Users can now start liquibase also from directories which includes parenthesis in their path.

### DIFF
--- a/liquibase-dist/src/main/archive/liquibase.bat
+++ b/liquibase-dist/src/main/archive/liquibase.bat
@@ -14,7 +14,7 @@ if NOT "%JAVA_HOME%" == "" set JAVA_HOME=%JAVA_HOME:"=%
 
 rem set JAVA_HOME to local jre dir if not set
 if exist "%LIQUIBASE_HOME%\jre" if "%JAVA_HOME%"=="" (
-    set JAVA_HOME=%LIQUIBASE_HOME%\jre
+    set JAVA_HOME="%LIQUIBASE_HOME%\jre"
 )
 
 if NOT "%JAVA_HOME%" == "" if not exist "%JAVA_HOME%" (


kevin-atx:
@filipelautert This change was made to the Windows batch file. In addition to this change, could you check and see if a similar change needs to be made to the Linux script file? Are there any other places in the code where this change makes sense?

filipelautert:
@kevin-atx all fine, Linux version already has the double quotes .

@JulienMa94 - thanks for the PR! With this fix, am I right to say that we should do the same around line 31 too? 

`if NOT "%JAVA_HOME%" == "" set JAVA_PATH=%JAVA_HOME%\bin\java`

JulienMa94:
> @JulienMa94 - thanks for the PR! With this fix, am I right to say that we should do the same around line 31 too? 
> `if NOT "%JAVA_HOME%" == "" set JAVA_PATH=%JAVA_HOME%\bin\java`

@filipelautert Makes sense to me too. I will do the change the coming days and test it.

JulienMa94:
@filipelautert After testing my changes I think that line 31 > `if NOT "%JAVA_HOME%" == "" set JAVA_PATH=%JAVA_HOME%\bin\java` should not be changed.

I tryed starting the liquibase cli from a path with containing parenthesis & also changed my JAVA_HOME systemvariable to a path with parenthesis and could start liquibase / see the banner with options.

Inside the screenshot you can see the LIQUIBASE_HOME, JAVA_HOME, JAVA_PATH variables printed to the powershell before the liquibase banner appears.

image

So I assume that the JAVA_PATH is set correctly and that line 33 already uses the correct qoutes > `"%JAVA_PATH%" %JAVA_OPTS% -jar "%LIQUIBASE_HOME%\internal\lib\liquibase-core.jar" %*`.

filipelautert (approved):
Thanks for the fix @JulienMa94

rberezen:
This pull request addresses an issue where parentheses in the file path prevent the Liquibase CLI from running on Windows. I have verified that the fix works on Windows.

Thank you @JulienMa94!

Approved.

rberezen (approved):

tati-qalified (approved):

filipelautert:
Hello @JulienMa94 - we had to revert this PR as it caused https://github.com/liquibase/liquibase/issues/4695 . I tried a fix but it didn't work. Do you have any ideas around that?

JulienMa94:
Hi @filipelautert just read your message and looked into the releated pull requests / work. At the moment I don't have any ideas what could cause / fix these issue.

JulienMa94:
Hi @filipelautert, 

maybe found a fix for the parenthese issue. In the following [Stackoverflow thread](https://stackoverflow.com/questions/7883169/how-to-escape-variables-with-parentheses-inside-if-clause-in-a-batch-file) it is suggested to encapsulate the whole set command with double quoutes to allow special characters inside a variable like so:
`
@echo off
set a=some value with (parentheses) inside
if 1 == 1 (
    set "PATH=%a%"
)
`

So we could try
`set "JAVA_HOME=%LIQUIBASE_HOME%\jre"` 
instead of 
`set JAVA_HOME="%LIQUIBASE_HOME%\jre"`

Maybe this could be a possible fix and also does not create other issues.

Greetings
Julien

filipelautert:
Thanks for the research @JulienMa94 ! We could create a new PR with this change and then test team can help us to test it on various windows environments.

JulienMa94:
Hi @filipelautert,
I can submit a PR in roughly 2 weeks I am not able to work until the end of September. If this fix can wait I am happy to contribute.

Greetings
Julien

filipelautert:
@JulienMa94 -> https://github.com/liquibase/liquibase/pull/4985

PR #4307, opened 2023-05-24
Fixes #3661

### LINKED ISSUE #3661: Specifying precision in fixed-precision integers fails when executing on Postgres
## Environment

**Liquibase Version**: 4.9.1

**Liquibase Integration & Version**: maven

**Liquibase Extension(s) & Version**: liquiibase-core 4.9.1

**Database Vendor & Version**: PostgreSQL 14.5.0

**Operating System Type & Version**: Mac OS 13.1

**Infrastructure Type/Provider**: n/a

## Description

We have several old changesets which executed on H2 and mySQL without issue but there are change sets which declare a column type to be `bigint(20)`. This parameter is deprecated and unused in modern versions of mySQL and Liquibase correctly ignores the parameter when executing against H2, but Liquibase generates invalid SQL against Postgres, which does not support the parameter at all.

## Steps To Reproduce

```yaml
databaseChangeLog:
  - changeSet:
      id: "create-revision-info-table"
      author: fraferna
      changes:
        - createTable:
            tableName: revinfo
            columns:
              - column:
                  name: id
                  type: int(11)
                  autoIncrement: true
                  constraints:
                    primaryKey: true
                    nullable: false
              - column:
                  name: timestamp
                  type: bigint(20)
                  defaultValueComputed: null
```

This change set is executed by Spring Data Liquibase when starting up a Spring Boot application. It executes successfully against H2 1.4.200 and mySQL 8.0.30-commercial. Against a Postgres 14.5.0 instance running in a Docker container on my local machine, it generates an error

## Actual Behavior

```
org.springframework.beans.factory.BeanCreationException: Error creating bean with name 'liquibase' defined in class path resource [org/springframework/boot/autoconfigure/liquibase/LiquibaseAutoConfiguration$LiquibaseConfiguration.class]: Invocation of init method failed; nested exception is liquibase.exception.LiquibaseException: liquibase.exception.MigrationFailedException: Migration failed for change set db/changelog/1.0/create-tables-1.0.yml::create-revision-info-table::fraferna:
     Reason: liquibase.exception.DatabaseException: ERROR: syntax error at or near "("
  Position: 102 [Failed SQL: (0) CREATE TABLE vrr01dev.revinfo (id INTEGER GENERATED BY DEFAULT AS IDENTITY NOT NULL, timestamp BIGINT(20), CONSTRAINT revinfo_pkey PRIMARY KEY (id))]
```

## Expected/Desired Behavior

I would expect Liquibase to generate a statement with `timestamp BIGINT` rather than including the invalid parameter to the column type on Postgres.

farion:
## Impact

## Description

Fixes #3661. In case of Postgresql a definition of BIGINT with precision (e.g. bigint(20)) won't create broken SQL anymore. For Postgresql precision is removed from those definitions.

## Things to be aware of

Implementation was done like it was done for other databases (e.g. MySQL)

## Additional Context

Added a unit test.


liquibase-standard/src/main/java/liquibase/datatype/core/BigIntType.java
```diff
@@ -54,16 +54,15 @@ public DatabaseDataType toDatabaseDataType(Database database) {
             return new DatabaseDataType("BIGINT");
         }
         if (database instanceof PostgresDatabase) {
-            if (isAutoIncrement()) {
-                if (((PostgresDatabase) database).useSerialDatatypes()) {
-                    return new DatabaseDataType("BIGSERIAL");
+            if (isAutoIncrement() && ((PostgresDatabase) database).useSerialDatatypes()) {
+                return new DatabaseDataType("BIGSERIAL");
+            } else {
```
filipelautert, line 59: Before this else would execute only if it was autoIncrement, but now it always executes for postgresql. Is it expected?

StevenMassaro (commented):
I reviewed this and nothing jumps out at me as particularly troubling (besides what Filipe has pointed out), but I am not familiar enough with the code to give it an approve.

liquibase-standard/src/main/java/liquibase/datatype/core/BigIntType.java
farion: Yes this is correct. For Posgres is it never allowed to use the default behavior (using BIGINT with size)

filipelautert: Thanks!

rberezen:
This pull request addresses an error that occurs when executing a change set that declares a column type as bigint with precision (e.g., bigint(20)) on PostgreSQL, resulting in a failure. The fix properly handles the issue by converting it to the bigint type with no precision.

- New unit test added.
- No additional testing required.

APPROVED

rberezen (approved):

PR #4308, opened 2023-05-24
Fix issue #3317 - StringIndexOutOfBoundsException if date is set to 1970-01-01

### LINKED ISSUE #3317: StringIndexOutOfBoundsException if date is set to 1970-01-01
When running liquibase on a host that has no date set (and it is therefore set to 1970-01-01, e.g. because the battery of the RTC is empty), the AbstractChangeLogHistoryService.generateDeploymentId method throws an exception:

```
 java.lang.StringIndexOutOfBoundsException: begin -4, end 6, length 6<CR><LF>
                at java.base/java.lang.String.checkBoundsBeginEnd(Unknown Source)<CR><LF>
                at java.base/java.lang.String.substring(Unknown Source)<CR><LF>
                at java.base/java.lang.String.substring(Unknown Source)<CR><LF>
                at liquibase.changelog.AbstractChangeLogHistoryService.generateDeploymentId(AbstractChangeLogHistoryService.java:131)<CR><LF>
                at liquibase.Liquibase.lambda$update$1(Liquibase.java:229)<CR><LF>
                at liquibase.Scope.lambda$child$0(Scope.java:180)<CR><LF>
                at liquibase.Scope.child(Scope.java:189)<CR><LF>
                at liquibase.Scope.child(Scope.java:179)<CR><LF>
                at liquibase.Scope.child(Scope.java:158)<CR><LF>
                at liquibase.Liquibase.runInScope(Liquibase.java:2405)<CR><LF>
                ... 14 more<CR><LF>
```

The root cause can be found in the implementation of generateDeploymentId:

```
String dateString = String.valueOf(new Date().getTime());
this.deploymentId = dateString.substring(dateString.length() - 10);
```
The method assumes that the time in milliseconds converted to a string results in a string that has at least length 10. If the date is close to the start of the era (1970-01-01) this may not be the case and the substring call fails with an exception because the index is negative.

Mika2147:
## Impact

## Description

Fixes #3317 where `indexOutOfBounds` exception was raised if the date was too close to 1/1/1970 and Milliseconds since then had less than 10 digits. 

The solution was to fill up the resulting 10-digit deployment id with leading zeros.


liquibase-standard/src/main/java/liquibase/changelog/AbstractChangeLogHistoryService.java
```diff
@@ -132,7 +132,8 @@ public void resetDeploymentId() {
     public void generateDeploymentId() {
         if (this.deploymentId == null) {
             String dateString = String.valueOf(new Date().getTime());
-            this.deploymentId = dateString.substring(dateString.length() - 10);
+            this.deploymentId = dateString.length() > 9 ? dateString.substring(dateString.length() - 10) :
+                    String.format("%010d", dateString);
```
MalloD12, line 136: This new piece is not working for me, when setting my local time to `1/1/1970 1:00:00` when this code gets executed I'm getting a `d != java.lang.String` error, because the format is not totally accurate as `d` flag expects for a number, but we are passing over a string. 

I tried something like this:

```
if (this.deploymentId == null) {
            long dateTime = new Date().getTime();
            String dateString = String.valueOf(dateTime);
            DecimalFormat decimalFormat = new DecimalFormat("0000000000");

            this.deploymentId = dateString.length() > 9 ? dateString.substring(dateString.length() - 10) :
                    decimalFormat.format(dateTime);
        }
```
and it worked for me. 

Would you mind updating this code, or if you prefer I can do it?

Thanks,
Daniel.

MalloD12:
Hi @Mika2147, 

Thank you for taking ownership of #3317 and implementing a fix for it. I left you a review comment to it.

Thanks,
Daniel.

MalloD12:
Hi @Mika2147,

Have you had a chance to look at my review comment? 

Thanks,
Daniel.

Mika2147:
> Hi @Mika2147,
> 
> Have you had a chance to look at my review comment?
> 
> Thanks, Daniel.

I didn't had the time to fix it yet, but will look at it soon.

Mika2147:
@MalloD12 

I implemented the fix you suggested.

MalloD12 (approved):
**Approved**.

Not sure how frequent is to get a date which produces a time value with less than 10 digits, but it's good to prevent that case. 

Again thank you @Mika2147 for creating this PR, and I appreciate you address my review comment. 

Build and all tests checks have been successfully executed. It has been also manually tested. 


### Things to be aware of:
- None

### Things to worry about:
- None

filipelautert (approved):

rberezen (approved):

PR #4311, opened 2023-05-24
[DAT-14885] Upgrade to v9 after running liquibase update command

filipelautert:
## Impact

## Description

Implementation Details
Add a new visitor class ChecksumsUpgradeVisitor to have the logic to upgrade to v9

Expected result:
After running “update” command all the v8 checksums associated with the changesets that are “checked” should be now v9 (in other words, only changesets which match the supplied filters (ie labels, contexts, dbms, etc) will have their checksums upgraded)

### DIFF
--- a/liquibase-standard/src/main/java/liquibase/changelog/visitor/UpgradeCheckSumVisitor.java
+++ b/liquibase-standard/src/main/java/liquibase/changelog/visitor/UpgradeCheckSumVisitor.java
@@ -17,6 +17,10 @@ import liquibase.statement.core.UpdateChangeSetChecksumStatement;
 
 import java.util.Set;
 
+/**
+ * ChangeSetVisitor that will upgrade the checksum version and calculation from previous versions to this Liquibase checksum version .
+ * It should not downgrade checksums neither recalculate existing ones.
+ */
 public class UpgradeCheckSumVisitor implements ChangeSetVisitor {
 
     private final Database database;
@@ -36,8 +40,8 @@ public class UpgradeCheckSumVisitor implements ChangeSetVisitor {
         Executor executor = Scope.getCurrentScope().getSingleton(ExecutorService.class).getExecutor("jdbc", database);
 
         ChangeLogHistoryService changeLogService = ChangeLogHistoryServiceFactory.getInstance().getChangeLogService(database);
-        if (changeLogService.isDatabaseChecksumsCompatible() || changeSet.getStoredCheckSum() == null ||
-                changeSet.getStoredCheckSum().getVersion() == ChecksumVersions.latest().getVersion()) {
+        if (changeLogService.isDatabaseChecksumsCompatible() ||
+                (changeSet.getStoredCheckSum() != null &&  changeSet.getStoredCheckSum().getVersion() >= ChecksumVersions.latest().getVersion())) {
             return;
         }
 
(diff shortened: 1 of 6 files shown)


liquibase-standard/src/main/java/liquibase/changelog/visitor/UpgradeCheckSumVisitor.java
```diff
@@ -0,0 +1,57 @@
...
+    @Override
+    public void visit(ChangeSet changeSet, DatabaseChangeLog databaseChangeLog, Database database,
+                      Set<ChangeSetFilterResult> filterResults) throws LiquibaseException {
+        Executor executor = Scope.getCurrentScope().getSingleton(ExecutorService.class).getExecutor("jdbc", database);
+
+        ChangeLogHistoryService changeLogService = ChangeLogHistoryServiceFactory.getInstance().getChangeLogService(database);
+        if (changeLogService.isDatabaseChecksumsCompatible() || changeSet.getStoredCheckSum() == null ||
+                changeSet.getStoredCheckSum().getVersion() == ChecksumVersions.latest().getVersion()) {
+            return;
+        }
```
StevenMassaro, line 42: Why do we return if the stored checksum is null?

liquibase-standard/src/main/java/liquibase/exception/UnsupportedChecksumVersionException.java
```diff
@@ -1,10 +1,15 @@
...
 public class UnsupportedChecksumVersionException extends RuntimeException {
 
     private static final long serialVersionUID = -229400973681987065L;
 
     public UnsupportedChecksumVersionException(int i) {
-        super("Unsupported Checksum version: " + i);
+        super(String.format("Liquibase detected an unknown checksum calculation version %s in the DatabaseChangelog table. " +
+                "This can be caused by a previous operation on this changelog with a newer version than your current %S release. " +
+                "Please update your current Liquibase, or clear checksums on changesets you are certain you need " +
+                "to deploy.", i, LiquibaseUtil.getBuildVersionInfo()));
```
StevenMassaro, line 13: I don't think we should even mention clear checksums.

```suggestion
        super(String.format("Liquibase detected an unknown checksum calculation version %s in the DatabaseChangelog table. " +
                "This can be caused by performing operations on this changelog/database with a newer version of Liquibase than your current %S release. " +
                "Please update your current Liquibase to perform this operation against this database..", i, LiquibaseUtil.getBuildVersionInfo()));
```

filipelautert: This error message was created based on the suggestion from DAT-14876 . Maybe we can discuss it there and then update the message? Meanwhile I"ll make it simple, thanks!

liquibase-standard/src/main/java/liquibase/changelog/visitor/UpgradeCheckSumVisitor.java
filipelautert: Hm, right. If it's null we should upgrade it, makes sense!

liquibase-standard/src/main/java/liquibase/command/core/AbstractUpdateCommandStep.java
```diff
@@ -108,16 +108,19 @@ public void run(CommandResultsBuilder resultsBuilder) throws Exception {
...
-        // FIXME this is not working - I believe we need to reload DatabaseChangelog too.
-        changeLogService.reset();
-        changeLogService.init();
-        changeLogService.getRanChangeSets(true);
-        changeLogService.upgradeChecksums(databaseChangeLog, contexts, labels);
+        ChangeLogIterator runChangeLogIterator = new ChangeLogIterator(databaseChangeLog,
+                new ContextChangeSetFilter(contexts),
+                new LabelChangeSetFilter(labelExpression),
+                new DbmsChangeSetFilter(database),
+                new IgnoreChangeSetFilter());
```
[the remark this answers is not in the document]
filipelautert: Good idea, done!

StevenMassaro (approved):

PR #4318, opened 2023-05-26
[DAT-14885] - UpdateVisitor upgrade

filipelautert:
## Impact

## Description
Changes updateVisitor so it is the one responsible by upgrading checksums.


liquibase-standard/src/main/java/liquibase/changelog/visitor/UpdateVisitor.java
```diff
@@ -65,15 +113,14 @@ public void visit(ChangeSet changeSet, DatabaseChangeLog databaseChangeLog, Data
...
-            execType = ChangeSet.ExecType.RERAN;
+        if (!Objects.equals(runStatus, RunStatus.NOT_RAN) && Objects.equals(execType, ExecType.EXECUTED)) {
+            execType = ExecType.RERAN;
         }
         fireRan(changeSet, databaseChangeLog, database, execType);
         addAttributesForMdc(changeSet, execType);
         // reset object quoting strategy after running changeset
         this.database.setObjectQuotingStrategy(previousStr);
         this.database.markChangeSetExecStatus(changeSet, execType);
-        this.database.commit();
```
StevenMassaro, line 76: Why did you remove this?

liquibase-standard/src/main/java/liquibase/changelog/visitor/UpdateVisitor.java
```diff
@@ -49,11 +62,46 @@ public Direction getDirection() {
...
+        if (isAccepted) {
+            executeAcceptedChange(changeSet, databaseChangeLog, database);
+        } else if ((oldChecksum == null || oldChecksum.getVersion() < ChecksumVersions.latest().getVersion())) {
+            upgradeCheckSumVersionForAlreadyExecutedOrNullChange(changeSet, database, oldChecksum);
+        }
+
+        this.database.commit();
+    }
+
+    private static CheckSum updateCheckSumIfRequired(ChangeSet changeSet) {
```
StevenMassaro, line 78: Javadoc here please.

liquibase-standard/src/main/java/liquibase/changelog/visitor/UpdateVisitor.java
```diff
@@ -49,11 +62,46 @@ public Direction getDirection() {
...
+        CheckSum oldChecksum = null;
+        if (changeSet.getStoredCheckSum() != null && changeSet.getStoredCheckSum().getVersion() < ChecksumVersions.latest().getVersion()) {
+            oldChecksum = changeSet.getStoredCheckSum();
+            changeSet.clearCheckSum();
+            changeSet.setStoredCheckSum(changeSet.generateCheckSum(ChecksumVersions.latest()));
+        }
+        return oldChecksum;
+    }
+
+    private static void upgradeCheckSumVersionForAlreadyExecutedOrNullChange(ChangeSet changeSet, Database database, CheckSum oldChecksum) throws DatabaseException {
```
StevenMassaro, line 88: Javadoc here please.

liquibase-standard/src/main/java/liquibase/changelog/visitor/UpdateVisitor.java
```diff
@@ -49,11 +62,46 @@ public Direction getDirection() {
...
         Executor executor = Scope.getCurrentScope().getSingleton(ExecutorService.class).getExecutor("jdbc", database);
         if (! (executor instanceof LoggingExecutor)) {
+            Scope.getCurrentScope().getUI().sendMessage(String.format("Upgrading checksum for Changeset %s from %s to %s.",
+                    changeSet, (oldChecksum != null? oldChecksum.toString() : "<null>"), changeSet.getStoredCheckSum().toString()));
+        }
+        Scope.getCurrentScope().getSingleton(ExecutorService.class).getExecutor("jdbc", database)
+                .execute(new UpdateChangeSetChecksumStatement(changeSet));
+    }
+
+    private void executeAcceptedChange(ChangeSet changeSet, DatabaseChangeLog databaseChangeLog, Database database)
```
StevenMassaro, line 98: Javadoc here please.

liquibase-standard/src/main/java/liquibase/changelog/visitor/UpdateVisitor.java
```diff
@@ -27,17 +31,26 @@ public class UpdateVisitor implements ChangeSetVisitor {
...
     /**
      * @deprecated - please use the constructor with ChangeExecListener, which can be null.
      */
     @Deprecated
     public UpdateVisitor(Database database) {
         this.database = database;
     }
-    
+
+    @Deprecated
```
StevenMassaro, line 44: Can you explain how this class should be instantiated with this newly deprecated constructor?

StevenMassaro (approved):
I just realized that there is one Update command that doesn't extend from Abstract Update. Can you verify that this other command (I think it's update-one-changeset) will also get this change?

liquibase-standard/src/main/java/liquibase/changelog/visitor/UpdateVisitor.java
filipelautert: It's on the calling method, line 74.

liquibase-standard/src/main/java/liquibase/changelog/visitor/UpdateVisitor.java
filipelautert: Thinking a bit more, I believe we don't need to deprecate it. I added a validation for `shouldRunChangeSetFilter `so it will behave the same when this constructor is used.

PR #4321, opened 2023-05-30
DAT-6585 Apt-based Linux package

jandroav:
## Impact

## Description

feat(build): add JReleaser support to build and package Liquibase CLI… tool

This commit adds support for `jdeb` to build and package the `deb` package for liquibase. It also upload it to a new created APT repo hosted in AWS S3: `repo.liquibase.com`


jreleaser.yml
```diff
@@ -0,0 +1,58 @@
...
+
+assemble:
+  jlink:
+    liquibase-jink:
+      active: ALWAYS
+      exported: false
+      imageName: 'liquibase-{{projectVersion}}'
+      executable: liquibase
+      targetJdks:
+        - path: '{{jdkPathPrefix}}/zulu17LinuxIntel/{{jdkFilePrefix}}-linux_x64'
```
aalmiray, line 32: Custom JDKs must be explicitly provisioned, JReleaser will not perform this task.
Have a look at https://github.com/jreleaser/helloworld-java-jpackage for more information.

jandroav: Thanks @aalmiray It was just a first test. Also since the `liquibase` final artifact is already generated in the build process (a `.tgz` file under https://github.com/liquibase/liquibase/tree/master/liquibase-dist/target) do you think it is possible to just use `jrealease` (packager) to create `deb`/`rpm` packages based on that `.tgz` file? I was tryting to find an example using `maven` so I can ommit using the `jreleaser.yml` file as well and just configure `jrealease` in the main `pom.xml`.

aalmiray: It's definitely possible to use the `.tgz` but its contents must be explicitly expanded and mapped to the JReleaser configuration. Perhaps it would be a good idea to support archives (zips and tars) as sources for JReleaser's `jlink` assembler.

aalmiray: Filed https://github.com/jreleaser/jreleaser/issues/1399

jandroav: I opened this as well https://github.com/jreleaser/jreleaser/issues/1400

nvoxland:
Yes, @jandroav, `liquibase-core.jar` is the complete jar file and what is actually shipped in the .tar.gz. `liquibase-cli` is just a subset of that used for internal build dependency management only. 

It might be good to move the logic here not just to liquibase-core but to liquibase-dist? `liquibase-dist` is the module in charge of creating the overall CLI installer.

I also wonder if we shouldn't be using jlink in the first place? Because Liquibase is designed to be extensible by others, we don't actually know the complete list of java modules people will be using in order to create a custom java runtime -- if we do, people who run the apt package with an extension that uses a module that didn't make the cut will get failures that they aren't expecting. Perhaps better to just be adding java as a dependency of the .deb file so it gets reused or newly installed as needed?

Even if you want to use jlink, you'll want to make sure you're building the .deb file from what is extracted from the .tar.gz file created in `liquibase-dist`. We'd want to make sure what's included in that distribution is the same as all our other distributions and the tar.gz file is the canonical base for all that. Besides the launcher bash script, it has other files needed to run liquibase such as the additional .jar files to use at runtime, local documentation, example/getting-started content, and license info which we'd need.

We are no longer bothering to build the exploded-directory version of the .tar.gz file using the assembly plugin in liquibase-dist. The reason for that is because in the assembly we want to make sure executable bits are set on the right files and those get lost in the exploded directory on windows. But if it makes it easier for you, it's easy to re-enable with the `formats` section of assembly-bin-pro.xml and assembly-bin.xml

liquibase-dist/src/main/archive/liquibase.sh
```diff
@@ -0,0 +1,6 @@
+#!/bin/sh
```
filipelautert, line 1: Maybe this file should be called something like `liquibase-env.sh`?  First time I saw it I thought it was a wrapper to liquibase command.

liquibase-dist/src/main/archive/liquibase.sh
```diff
@@ -0,0 +1,6 @@
...
+
+export LIQUIBASE_HOME=/opt/liquibase
+export PATH=$PATH:$LIQUIBASE_HOME
+source /etc/profile.d/
```
filipelautert, line 5: The file is installed inside `/etc/profile.d/` . Does it need to source it?

liquibase-dist/src/main/archive/liquibase.sh
jandroav: 👍🏻

liquibase-dist/src/main/archive/liquibase.sh
jandroav: Nop

filipelautert (approved):

aalmiray:
https://github.com/jreleaser/jreleaser/issues/1399 was fixed and released in JReleaser `1.7.0`.

In the future, JReleaser will also support creating `.deb` and `.rpm` artifacts independently from `jpackage that is, without bundling a Java runtime. This will work similarly as the jdeb maven plugin seen in this PR.

jnewton03 (approved):

PR #4361, opened 2023-06-08
[DAT-14970] FormattedSqlChangeLogParser refactoring

vitaliimak:
## Impact

## Description

For creating `FormattedMongoshChangeLogParser` we need to move the common "ForamattedChangeLogParser" logic into the base class. Please take a look at the discussion in DAT-13151 for more information.

It is still In progress. However, I would like to get feedback about this refactoring.

In general, I have moved all common logic into `FormattedChangeLogParser` class and created a few abstract methods that should be implemented in derived SQL or Mongosh classes. E.g. the comment characters are different for SQL and JS, so they should be declared in the relevant classes.


vitaliimak:
Hello @filipelautert @wwillard7800 @KushnirykOleh!
Could you look at this PR?

filipelautert:
Hello @vitaliimak - at a first look it seems good. Seems that all public methods were kept so it's not breaking apis, and the abstract methods are simple and easy to understand. 
The Abstract class is quite big (as the original one was). Do you think it would be possible to split it and get rid of this god class?

filipelautert:
> However, if you have some ideas on how to make bigger refactoring we can discuss it.

Hello! As discussed so far I've no better ideas, so let's leave as is. Thanks!

filipelautert (approved):
Approved once functional tests are passing for it.

PR #4363, opened 2023-06-08
Added additional setup method to test class outside command tests DAT-14751

wwillard7800:
Test-only code modification


wwillard7800:
Should not be implemented here

PR #4409, opened 2023-06-20
DAT-15030

sayaliM0412:
## Impact

## Description

Enforce labels on every PR. Cannot merge until a label is added to the PR.

### DIFF
--- a/.github/pull_request_template.md
+++ b/.github/pull_request_template.md
@@ -5,9 +5,9 @@
 - [ ] Enhancement/New feature (adds functionality without impacting existing logic)
 - [ ] Breaking change (fix or feature that would cause existing functionality to change)
  
-## Mandatory Labels to add to a PR
+<!--  Mandatory Labels to add to a PR
 
-<!-- At least one of the labels must be added to the PR before it's merged. If no label is provided the workflow will fail and you will not be able to merge the PR. After the label is added it re-runs the `Pull Request Labels / label (pull_request)` and gives a green check. 
+At least one of the labels must be added to the PR before it's merged. If no label is provided the workflow will fail and you will not be able to merge the PR. After the label is added it re-runs the `Pull Request Labels / label (pull_request)` and gives a green check. 
 
 `skipReleaseNotes`   - Don't show up on the Draft Release Notes page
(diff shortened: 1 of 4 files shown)


.github/pull_request_template.md
```diff
@@ -25,6 +25,20 @@ A clear and concise description of the change being made.
...
+
+`skipReleaseNotes`   - Don't show up on the Draft Release Notes page
+`notableChanges`     - Any notable changes
+`TypeEnhancement`    - New features
+`TypeTest`           - New Test features
+`TypeBug`            - bug fixes
+`breakingChanges`    - any breaking changes
+`APIBreakingChanges` - any API breaking changes
+`sdou`               - Security, Driver and Other Updates -dependabot PR's
+`newContributors`    - New Contributors
```
abrackx, line 40: I don't think we need a new visible section for this in the template since the label will be attached to the pr. Can we put this all in a comment? If it's not as visible that way maybe it should be moved higher in the template (but still in a comment imo).

abrackx (changes requested):

.github/pull_request_template.md
filipelautert: Makes sense. Maybe we don't even need this info in the template?

abrackx: IMO I think we **should** have the info in the template so a community user doesn't experience any more friction than they need to to get a green build. I just don't think it's useful when I'm looking at the PR description as a reviewer. As a creator of a PR I do see the value.

.github/workflows/label-pr.yml
```diff
@@ -0,0 +1,19 @@
...
+  pull_request:
+    types: [opened, labeled, unlabeled, synchronize, reopened]
+jobs:
+  label:
+    runs-on: ubuntu-latest
+    permissions:
+      issues: write
+      pull-requests: write
+    steps:
+      - name: Label PR
```
abrackx, line 12: Can we name this step `Validate PR Labels?` or something similar? The name is a bit misleading as is.

abrackx (approved):
🚀 Awesome!

PR #4447, opened 2023-06-28
allow sql checks without an expected result as precondition

MichaelKern-IVV:
## Impact

## Description

Not only queries should be allowed to be used as SQL checks, but also commands.
For example, for DB2luw the command SET ROLE is used to check whether a role is assigned to the connected user or not.
If the role was not assigned, the command is aborted with an error. There is no result beyond that.

See https://www.ibm.com/docs/en/db2/11.5?topic=statements-set-role for details


MalloD12:
Hi @MichaelKern-IVV, 

I've started testing doing a bit of manual testing, plus trying to add some tests, and today after working a bit on this with @filipelautert we thought maybe not be the best option to remove `expectedResult`, because it will allow running anything inside like for example something like an `INSERT` could be run without any issues which I think it wouldn't be the ideal case. 

I'm thinking maybe it would be a better and safe option to write a new nested precondition type like `sqlChange` but which doesn't expect a specific result and allow us to execute any type of command.

What do you think about it?

Thanks,
Daniel.

MichaelKern-IVV:
Hi @MalloD12,

I think that is a good idea.

Michael

MichaelKern-IVV:
Hallo @MalloD12,
I have added the FaultlessSqlExecutionPrecondition. I am not sure if the name is so well done. If not, please change it as you like.

MalloD12:
> Hallo @MalloD12, I have added the FaultlessSqlExecutionPrecondition. I am not sure if the name is so well done. If not, please change it as you like.

Hi @MichaelKern-IVV,

Thank you for adding these new changes, I'll be looking into it today. If you have any chance, could you please add any unit tests for this?

Thanks,
Daniel.

MalloD12:
Hey @MichaelKern-IVV,

I've been looking at your changes which look good to me, but I've been thinking about whether we really need this precondition type or not. I was thinking if this is not something we can also implement with the current support liquibase provides us by for example defining at the top of our changelog a new changeset with a raw SQL to execute for which we can add the `failOnError` attribute set to `false` as displayed below:

```
<changeSet  author="liquibase-docs"  id="sql-example" failOnError="false">
  <sql>
SET ROLE 
   </sql>
</changeSet>
```
If there is any specific reason I'm happy to move forward with this, but I think it would be nice to add a more customized message when throwing an error when the execution command fails for some reason:
```
catch (DatabaseException e) {
            throw new PreconditionErrorException(e, changeLog, this);
```
I can make those changes if you wish. Also, if we are ok with this new precondition type I think we should update its name to something like `SqlWithoutExpectedValPrecondition` or maybe something like `SqlCommandExecutionPrecondition`.

Please let me know what you think about all this. In the meantime, I'll double-check this with someone else from the team.

Thanks,
Daniel.

MichaelKern-IVV:
Hi @MalloD12 ,

that's a good idea, I will switch to raw SQL with the runAlways option. Thanks for your effort. So I will close the request,

Michael

MalloD12:
Hi @MichaelKern-IVV,

Do you mean having both options `runAlways=true` and `failOnError=false`, right? 

Thanks,
Daniel.

MichaelKern-IVV:
@MalloD12

yes

Issue #4448, opened 2023-06-28
DeploymentId not populated after waiting for lock


abrackx:
### Description

#4427 introduced a call in AbstractUpdateCommandStep to LockServiceFactory to lock the database if it is not locked prior to an update. 
```
if(!isDBLocked) {
       LockServiceFactory.getInstance().getLockService(database).waitForLock();
}
```
This call to `waitForLock()` calls `aquireLock()` which then calls `Scope.getCurrentScope().getSingleton(ChangeLogHistoryServiceFactory.class).resetAll();`. This resets the ChangelogHistoryService, clearing the generated deploymentId. 

### Steps To Reproduce

Using the latest build on master run the command `liquibase update`. 

### Expected/Desired Behavior

DeploymentId should be populated when running update. 

### Liquibase Version

4.23.1-SNAPSHOT

PR #4450, opened 2023-06-28
Fix deploymentId not populating after locking

wwillard7800:
## Impact

## Description
Fix issue: #4448 
I spent a while trying to implement a `FastCheckCommandStep` class because the root cause really is `we want to not lock the database if fastcheck reports we are up to date`, but ending up stumbling on a number of issues relating to the different usages of the `LockServiceCommandStep`. Because the lock service step is so "low level" it doesn't really work to add the necessary parameters to support a fast check. I thought about implementing a `FastCheckLockServiceCommandStep` but just resolved I was getting in too deep and should be able to fix this in a few lines of code for now.

I'm not sure if `FastCheckLockServiceCommandStep` is the right move for the future or if we should choose some other route.

### DIFF
--- a/liquibase-integration-tests/src/test/groovy/liquibase/command/UpdateCommandsIntegrationTest.groovy
+++ b/liquibase-integration-tests/src/test/groovy/liquibase/command/UpdateCommandsIntegrationTest.groovy
@@ -58,6 +58,10 @@ class UpdateCommandsIntegrationTest extends Specification {
         resultSet.next()
         resultSet.getInt(1) == 1
 
+        def detailsResultSet = h2.getConnection().createStatement().executeQuery("select DEPLOYMENT_ID from databasechangelog")
+        detailsResultSet.next()
+        assert detailsResultSet.getString(1) != null : "No deployment ID found for the update"
+
         def rsTableExist = h2.getConnection().createStatement().executeQuery("select count(1) from example_table")
         rsTableExist.next()
         rsTableExist.getInt(1) == 0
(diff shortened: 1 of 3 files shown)


filipelautert (approved):
Thanks! That's not the first time that this happens.. if I'm right before 4.21.0 we had a similar issue.

MalloD12:
Hey @wwillard7800, I'm seeing there is another call of:

```
Scope.getCurrentScope().getSingleton(ChangeLogHistoryServiceFactory.class).getChangeLogService(database).generateDeploymentId();
```
in **AbstractFutureRollbackCommandStep**. Should that one also be removed or is not used? I don't have all this command process super clear, but thought it might be worth checking with you.

Thanks,
Daniel.

wwillard7800:
> Hey @wwillard7800 / @abrackx , I'm seeing there is another call of:
> 
> ```
> Scope.getCurrentScope().getSingleton(ChangeLogHistoryServiceFactory.class).getChangeLogService(database).generateDeploymentId();
> ```
> 
> in **AbstractFutureRollbackCommandStep**. Should that one also be removed or is not used? I don't have all this command process super clear, but thought it might be worth checking with you.
> 
> Thanks, Daniel.

I believe that this call needs to be here.  @abrackx do you agree?

filipelautert (approved):

abrackx:
> Hey @wwillard7800, I'm seeing there is another call of:
> 
> ```
Scope.getCurrentScope().getSingleton(ChangeLogHistoryServiceFactory.class).getChangeLogService(database).generateDeploymentId();
> ```
> 
> in **AbstractFutureRollbackCommandStep**. Should that one also be removed or is not used? I don't have all this command process super clear, but thought it might be worth checking with you.
> 
> Thanks, Daniel.

Good catch but I believe this is irrelevant to the bug. `future-rollback-<count/tag/etc>-sql` commands only yield rollback scripts and do not touch (nor operate on) deployment id. As far as I know, the only times deployment id really come into play is when running an update.

PR #4452, opened 2023-06-28
Fix checksum upgrade changes preventing useless updates and using AbstractChangeLogHistoryService for extensions compatibility

filipelautert:
## Impact

## Description

New method liquibase.changelog.visitor.UpdateVisitor#updateCheckSumIfRequired returns null if the change set has a stored checkSum, but it should return the old checksum value.
So everytime we get here we are updating the checksums for already executed checksums with the same value. It's not breaking anything but performance will be affected.

Steps To Reproduce
Run update using a changelog with 1 changeset
Add a new changeset to changelog
Run update SQL: first changelog will have the checksum updated and the second will execute just fine.
ex:

```
UPDATE public.databasechangelog SET MD5SUM = '9:56da2f293105bdbf802515b0581600f1' WHERE ID = '2' AND AUTHOR = 'fl' AND FILENAME = 'changelogs/changelog/simple_changelog.xml';

INSERT INTO public.databasechangelog (ID, AUTHOR, FILENAME, DATEEXECUTED, ORDEREXECUTED, MD5SUM, DESCRIPTION, COMMENTS, EXECTYPE, CONTEXTS, LABELS, LIQUIBASE, DEPLOYMENT_ID, TAG) VALUES ('2.5', 'fl', 'changelogs/changelog/simple_changelog.xml', NOW(), 4, '9:9a2f4bc8afbf0c9587b25b2ffd922d2e', 'tagDatabase', '', 'EXECUTED', 'v8', NULL, 'DEV', '7973169078', 'rb1');
```

### DIFF
--- a/liquibase-standard/src/main/java/liquibase/changelog/visitor/UpdateVisitor.java
+++ b/liquibase-standard/src/main/java/liquibase/changelog/visitor/UpdateVisitor.java
@@ -20,7 +20,6 @@ import liquibase.exception.MigrationFailedException;
 import liquibase.executor.Executor;
 import liquibase.executor.ExecutorService;
 import liquibase.executor.LoggingExecutor;
-import liquibase.integration.commandline.LiquibaseCommandLineConfiguration;
 import liquibase.statement.core.UpdateChangeSetChecksumStatement;
 
 import java.util.Objects;
@@ -81,9 +80,8 @@ public class UpdateVisitor implements ChangeSetVisitor {
      * @return oldChecksum the former checksum
      */
     private static CheckSum updateCheckSumIfRequired(ChangeSet changeSet) {
-        CheckSum oldChecksum = null;
-        if (changeSet.getStoredCheckSum() == null || changeSet.getStoredCheckSum().getVersion() < ChecksumVersion.latest().getVersion()) {
-            oldChecksum = changeSet.getStoredCheckSum();
+        CheckSum oldChecksum = changeSet.getStoredCheckSum();
+        if (oldChecksum == null || oldChecksum.getVersion() < ChecksumVersion.latest().getVersion()) {
             changeSet.clearCheckSum();
             changeSet.setStoredCheckSum(changeSet.generateCheckSum(ChecksumVersion.latest()));
         }
--- a/liquibase-standard/src/main/java/liquibase/changelog/visitor/UpdateVisitor.java
+++ b/liquibase-standard/src/main/java/liquibase/changelog/visitor/UpdateVisitor.java
@@ -3,12 +3,9 @@ package liquibase.changelog.visitor;
 import liquibase.ChecksumVersion;
 import liquibase.Scope;
 import liquibase.change.CheckSum;
-import liquibase.changelog.ChangeLogHistoryService;
-import liquibase.changelog.ChangeLogHistoryServiceFactory;
-import liquibase.changelog.ChangeSet;
+import liquibase.changelog.*;
 import liquibase.changelog.ChangeSet.ExecType;
 import liquibase.changelog.ChangeSet.RunStatus;
-import liquibase.changelog.DatabaseChangeLog;
 import liquibase.changelog.filter.ChangeSetFilterResult;
 import liquibase.changelog.filter.ShouldRunChangeSetFilter;
 import liquibase.database.Database;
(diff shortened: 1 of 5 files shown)


abrackx (approved):

MalloD12 (approved):
Looks ok to me. Is it possible to add a unit test for this or at this stage better to rely on manual testing and let QA team add some automated test for this?

rberezen:
@filipelautert @MalloD12 Added functional test here: https://github.com/liquibase/liquibase-pro-tests/pull/869/files
Also, tested manually. Looks good to me. But I checked only described steps to reproduce. I have done any other checksum testing. Do you think we need to verify any edge cases or other 'risky' areas? 
@obovsunivskyii @Tamelianovych @TymofiiKritsak could you please share your expert opinion too?

rberezen (approved):

liquibase-standard/src/main/java/liquibase/changelog/visitor/UpdateVisitor.java
```diff
@@ -99,8 +94,13 @@ private static void upgradeCheckSumVersionForAlreadyExecutedOrNullChange(ChangeS
             Scope.getCurrentScope().getUI().sendMessage(String.format("Upgrading checksum for Changeset %s from %s to %s.",
                     changeSet, oldChecksum, changeSet.getStoredCheckSum()));
         }
-        Scope.getCurrentScope().getSingleton(ExecutorService.class).getExecutor("jdbc", database)
-                .execute(new UpdateChangeSetChecksumStatement(changeSet));
+        ChangeLogHistoryService changeLogService = Scope.getCurrentScope().getSingleton(ChangeLogHistoryServiceFactory.class).getChangeLogService(database);
+        if (changeLogService instanceof AbstractChangeLogHistoryService) {
```
nvoxland, line 98: Doing `if instanceof` is usually a bad code smell. The big reason for OOP is to be making the calling code not care about the implementation details and only tell the interface what it wants to do.

In this case, nothing outside of the HistoryService implementations should know or care about UpdateChangeSetChecksumStatement, that should just be an implementation detail of some subclasses of the interface. From the standpoint of this UpdateVisitor, it should just be calling a descriptive method of what it wants from whatever the HistoryService is, like `changeLogService.replaceChecksum(changeSet)` and it's up to the implementation to do what you want.

That is a new method for the interface so you have to decide if you want that to be a breaking API change, or if you want to put a default implementation on the interface. Maybe one that executes the UpdateChangeSetChecksumStatement?

The job of the HistoryService is to wrap ALL the logic for managing the history because that is how extensions control it. You can't even assume a HistoryService is even built around statements, which is why the logic needs to be around a method not on a Generator.

But, you should really try to avoid breaking API changes, and default interface impementations are your friends for that. Like with 4.23 you maybe could have added defaults for the new getRanChangeSets(boolean) and isDatabaseChecksumsCompatible() that call off to the old methods and/or returned "fallback" values like "false"?

filipelautert: Thanks for the review @nvoxland ! 
I tried moving it to ChangeLogHistoryService but then I need another method to get the database, and this one would need to be abstract breaking the interface:

```
    Database getDatabase();

    default void replaceChecksum(ChangeSet changeSet) throws DatabaseException {
        Scope.getCurrentScope().getSingleton(ExecutorService.class).getExecutor("jdbc", getDatabase()).execute(new UpdateChangeSetChecksumStatement
                (changeSet));
        getDatabase().commit();
        reset();
    }
```

We can also leave it as is now just using UpdateChangeSetChecksumStatement and don't touch ChangeLogHistoryService, but seems it is hard for extensions to implement it. I don't know which is worse.

TymofiiKritsak:
Hi @rberezen, I believe we have enough coverage with functional tests for now. Thus, maybe we should add more tests for some `ignored` changesets or changesets with `runAlways` and/or `runOnChange`?
CC @Tamelianovych @obovsunivskyii

obovsunivskyii:
> runAlways

we could do this for both for runAlways and runOnChange

liquibase-standard/src/main/java/liquibase/changelog/visitor/UpdateVisitor.java
nvoxland: True, you would need that getDatabase() do do the default. I don't think you can/should rely on the `UpdateChangeSetChecksumStatement` for updating the checksums, it needs to be a method on the HistoryService. 

And you don't want to add `getDatabase()` to the HistoryService interface because I don't want to be forcing/requiring a single Database on the HistoryService like that. You want to keep the interface as clean and generic as possible.

So it does seem like it is a new method that's needed which unfortunately will then be a breaking API change, unless there's a default implementation that you can come up with that doesn't require any other methods to be added. Having the default be a no-op wouldn't be a valid default, would it?

Tamelianovych:
Hello @rberezen I think we need to retest the new workflow we added here https://github.com/liquibase/liquibase/issues/3549 to make sure it still has a correct behavior

liquibase-standard/src/main/java/liquibase/changelog/visitor/UpdateVisitor.java
filipelautert: The default no-op is a good idea! I added it to the other method too. What do you think now?

PR #3616, opened 2022-12-23
Fix CheckSum generation issues + improvements 

MalloD12:
## Impact

- [x] Bug fix (non-breaking change which fixes expected existing functionality)
- [x] Enhancement/New feature (adds functionality without impacting existing logic)
- [ ] Breaking change (fix or feature that would cause existing functionality to change)
 
## Description

Checksum generation logic has been updated to fix some issues like avoiding considering some characters like white spaces, indentation, new line, etc when generating a checksum. Also, as part of these changes we will start ignoring contexts-filters, labels and dbms, as well as some specific properties for some objects (for example CreateProcedure, CreateView, etc) like path, encoding, relativeToChangeLog, etc. 

Fixes #3549 

## Things to be aware of
This PR cannot be merged until the [Liquibase internal] DAT-13285 is complete.

## Additional Context

- Nothing

PR #4453, opened 2023-06-28
Implements equals method on AbstractChange to make sure that we can compare Changes correctly

filipelautert:
## Impact

## Description

Commit https://github.com/liquibase/liquibase/pull/3616/commits/279b1b35d9c5f984d2f8542b792c870fd71c3ec0# that is part of checksums upgrade moved the dbms logic from Changeset to DbmsFilter. But it happens that Change object and all of the inheriting classes to not implement equals method. So when it calls collection.removeAll(object) it is ending up removing all changes that are of the specific type (for instance AbstractSqlChange), and not only the expected change. 

So if you execute a sample changelog like the one below in a `H2 `database:

```
    <changeSet id="0" author="fl" >
        <sql>create table anydb(id boolean default true, status varchar(2))</sql>
        <sql dbms="oracle">create table oraculo(id boolean default true, status varchar(2))</sql>
        <dropTable tableName="anydb" />
    </changeSet>
```

The expected result would be to create anydb table then drop it; but it is just dropping the table.

This PR fixes this issue.

### DIFF
--- a/liquibase-standard/src/main/java/liquibase/change/AbstractChange.java
+++ b/liquibase-standard/src/main/java/liquibase/change/AbstractChange.java
@@ -868,4 +868,22 @@ public abstract class AbstractChange extends AbstractPlugin implements Change {
         return description;
     }
 
+    @Override
+    public boolean equals(Object o) {
+        if (this == o) return true;
+        if (o == null || getClass() != o.getClass()) return false;
+        if (!super.equals(o)) return false;
+        if (o instanceof DbmsTargetedChange && this instanceof DbmsTargetedChange) {
+            if (!((DbmsTargetedChange)o).getDbms().equals(((DbmsTargetedChange)this).getDbms())) {
+                return false;
+            }
+        }
+        AbstractChange that = (AbstractChange) o;
+        return Objects.equals(changeSet, that.changeSet);
+    }
+
+    @Override
+    public int hashCode() {
+        return Objects.hash(super.hashCode(), changeSet);
+    }
 }
--- a/liquibase-standard/src/main/java/liquibase/changelog/ChangeSet.java
+++ b/liquibase-standard/src/main/java/liquibase/changelog/ChangeSet.java
@@ -960,7 +960,7 @@ public class ChangeSet implements Conditional, ChangeLogChild {
      * Method created to remove changes from a changeset
      * @param collection
      */
-    public void removeAllChanges(Collection<?> collection) {
+    public void removeAllChanges(Collection<Change> collection) {
         this.changes.removeAll(collection);
     }
 


abrackx (commented):
Can we add that example as a test case?

liquibase-standard/src/main/java/liquibase/change/AbstractChange.java
```diff
@@ -868,4 +868,21 @@ public String getDescription() {
...
     }
 
+    @Override
+    public boolean equals(Object reference) {
+        if (this == reference) return true;
+        if (reference instanceof DbmsTargetedChange && this instanceof DbmsTargetedChange &&
+            !((DbmsTargetedChange)reference).getDbms().equals(((DbmsTargetedChange)this).getDbms())) {
+            return false;
+        }
+        if (reference == null || getClass() != reference.getClass()) return false;
```
wwillard7800, line 878: Does this null check need to happen right after the same-object check to avoid an NPE?  The instanceof call will allow it a null.

I know that is an edge case.

wwillard7800 (approved):

MalloD12:
Hey @rberezen I have added a test to ensure a case like the one Filipe mentioned above doesn't happen again. We only limited this scope to H2, because we just want to test a dbms filtered change that is not deployed in any other given DB (H2 in this case). Are you ok with this or think this should be more generic and be added for example in **AbstractIntegrationTest** or somewhere else?

liquibase-standard/src/main/java/liquibase/change/AbstractChange.java
filipelautert: I simulated it here, but "reference instanceof DbmsTargetedChange" returns false so the second if will bail out on the first condition and won't call the getDbms methods.

rberezen (changes requested):
hi @filipelautert 
fails for me when 

```
    <changeSet id="0" author="fl" >
        <sql dbms="oracle">create table anydb(id boolean default true, status varchar(2))</sql>
        <sql dbms="oracle">create table oraculo(id boolean default true, status varchar(2))</sql>
        <dropTable tableName="anydb" />
    </changeSet>
```

```
Unexpected error running Liquibase: java.lang.StackOverflowError

java.lang.StackOverflowError
	at java.base/java.util.TreeMap$KeyIterator.<init>(TreeMap.java:1259)
	at java.base/java.util.TreeMap.keyIterator(TreeMap.java:1107)
	at java.base/java.util.TreeMap$KeySet.iterator(TreeMap.java:1120)
	at java.base/java.util.AbstractCollection.addAll(AbstractCollection.java:351)
	at java.base/java.util.TreeSet.addAll(TreeSet.java:312)
	at java.base/java.util.TreeSet.<init>(TreeSet.java:160)
	at liquibase.AbstractExtensibleObject.getAttributes(AbstractExtensibleObject.java:41)
	at liquibase.util.StringUtil.join(StringUtil.java:390)
	at liquibase.AbstractExtensibleObject.describe(AbstractExtensibleObject.java:238)
	at liquibase.AbstractExtensibleObject.equals(AbstractExtensibleObject.java:259)
	at liquibase.change.AbstractChange.equals(AbstractChange.java:879)
	at java.base/java.util.Objects.equals(Objects.java:77)
```

rberezen:
> Hey @rberezen I have added a test to ensure a case like the one Filipe mentioned above doesn't happen again. We only limited this scope to H2, because we just want to test a dbms filtered change that is not deployed in any other given DB (H2 in this case). Are you ok with this or think this should be more generic and be added for example in **AbstractIntegrationTest** or somewhere else?

I'm fine with that. Could you pleas add a scenario for this changeset?  Thanks in advance
```
    <changeSet id="0" author="fl" >
        <sql dbms="oracle">create table anydb(id boolean default true, status varchar(2))</sql>
        <sql dbms="oracle">create table oraculo(id boolean default true, status varchar(2))</sql>
        <dropTable tableName="anydb" />
    </changeSet>
```

liquibase-integration-tests/src/test/java/liquibase/dbtest/h2/H2IntegrationTest.java
```diff
@@ -189,6 +190,22 @@
         Assert.assertEquals(insertedValue.toString(), queryResult.get(1).get(colName));
     }
 
+    @Test
+    public void makeSureDbmsFilteredChangeIsNotDeployed() throws Exception {
+        clearDatabase();
+
+        Liquibase liquibase = createLiquibase("changelogs/h2/complete/sql.change.dbms.filtered.should.not.be.deployed.changelog.xml");
+        liquibase.update();
```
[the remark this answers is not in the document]
rberezen: @MalloD12 could you take a looks on this please?

liquibase-integration-tests/src/test/java/liquibase/dbtest/h2/H2IntegrationTest.java
```diff
@@ -189,6 +190,22 @@ public void testRollbackWithoutContext() throws Exception {
...
```
rberezen, line 198: @MalloD12 if `liquibase.update();` fails, test will also fail, right?

liquibase-integration-tests/src/test/java/liquibase/dbtest/h2/H2IntegrationTest.java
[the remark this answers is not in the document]
MalloD12: Added a new `runUpdate()` method in the AbstractIntegrationTest class to run the update command without having to use the deprecated one anymore.

MalloD12:
> > Hey @rberezen I have added a test to ensure a case like the one Filipe mentioned above doesn't happen again. We only limited this scope to H2, because we just want to test a dbms filtered change that is not deployed in any other given DB (H2 in this case). Are you ok with this or think this should be more generic and be added for example in **AbstractIntegrationTest** or somewhere else?
> 
> I'm fine with that. Could you pleas add a scenario for this changeset? Thanks in advance
> 
> ```
>     <changeSet id="0" author="fl" >
>         <sql dbms="oracle">create table anydb(id boolean default true, status varchar(2))</sql>
>         <sql dbms="oracle">create table oraculo(id boolean default true, status varchar(2))</sql>
>         <dropTable tableName="anydb" />
>     </changeSet>
> ```

Added this new changeset, but I removed the `dropTable` change of it, in this case, it shouldn't be there unless we want to make sure the test fails trying to remove `anydb` table. If that's the case, I'm already covering that part with a `try-catch` by executing a `select` against that table. 

Please have a look at it and let me know if you are ok with the changes made.

Thanks,
Daniel.

liquibase-integration-tests/src/test/java/liquibase/dbtest/h2/H2IntegrationTest.java
MalloD12: I think it would depend on the failure it returns, if there is a validation error but it doesn't stop the execution of the update then it will try to run the rest of the tests, but if for some reason `update` execution is stopped by an exception test will fail as well.

rberezen (approved):
LGTM! Thanks @MalloD12 and @filipelautert!

Issue #3446, opened 2022-11-07
Multithread build on maven fails with goal updateSQL


piotrwielgolaski-tomtom:
## Environment

**Liquibase Version**: 4.17.1

**Liquibase Integration & Version**: maven
**Liquibase Extension(s) & Version**: 

**Database Vendor & Version**: offline

**Operating System Type & Version**: Linux

**Infrastructure Type/Provider**: <AWC, GCS, Azure, VM, etc>

## Description

My app is composed of multi modules, I want to get build fast, so I use mvn with option `-T 1C`
When I run updateSQL goal on many modules it fails with errors like this

```
[ERROR] Failed to execute goal org.liquibase:liquibase-maven-plugin:4.17.1:updateSQL (liquibase-install-4-tests) on project xxx-liquibase: Cannot end scope ucuxobkxqd when currently at scope gdwnihdevm -> [Help 1]
[ERROR] Failed to execute goal org.liquibase:liquibase-maven-plugin:4.17.1:updateSQL (liquibase-install-4-tests) on projectyyy-liquibase: Cannot end scope xgzdkkhbwv when currently at scope psiywodkzr -> [Help 1]
```

## Actual Behavior
Fails to build project

## Expected/Desired Behavior
Work with option `-T 1C`

I wonder if using https://github.com/liquibase/liquibase/blob/master/liquibase-core/src/main/java/liquibase/ThreadLocalScopeManager.java would not help this.
I dont see any way in maven to control it.

piotrwielgolaski-tomtom:
it is even more problematic, as when I want to skip liquidbase `-Dliquibase.skip=true` it is still failing

piotrwielgolaski-tomtom:
@FBurguer  I could try to work on PR, but I wonder if you would be ok with changing ScopeManager by default or maybe it should be configuration option?

nvoxland:
@piotrwielgolaski-tomtom I'd be OK with you changing the liquibase-maven-plugin to use the ThreadLocalScopeManager instead of SingletonScopeManager.

In the normal maven case, that shouldn't and in the multithreaded maven case I think the ThreadLocal setup should always make sense. If people find use cases where that doesn't make sense we can reconsider a config setting at that point.

Or, if you find that the way maven manages multi-threading makes more sense to have a different ScopeManager you can do that too.

Forcing ThreadLocalScopeManager (or a different one) on everyone which has otherwise not been used a lot is a bit of risk, but if there are problems it should be immediately obvious.

[...]

PR #4461, opened 2023-07-02
support concurrent build in maven plugin

piotrwielgolaski-tomtom:
## Impact

## Description

This is remaining part of https://github.com/liquibase/liquibase/issues/3446

To make liquid base work with maven concurrent build (-T 1C)

It uses scope manager, but it need to be set in static so before any plugins action kick in.
Also it should not inherit any scope, so each plugin execution can start on clean scope isolated from another.

### DIFF
--- a/liquibase-standard/src/main/java/liquibase/Scope.java
+++ b/liquibase-standard/src/main/java/liquibase/Scope.java
@@ -113,21 +113,7 @@ public class Scope {
     }
 
     public static void setScopeManager(ScopeManager scopeManager) {
-        Scope currentScope = getCurrentScope();
-        if (currentScope == null) {
-            currentScope = new Scope();
-        }
-
-        try {
-            currentScope = scopeManager.init(currentScope);
-        } catch (Exception e) {
-            Scope.getCurrentScope().getLog(Scope.class).warning(e.getMessage(), e);
-        }
-        scopeManager.setCurrentScope(currentScope);
-
         Scope.scopeManager = scopeManager;
-
-
     }
 
     /**
--- a/liquibase-standard/src/main/java/liquibase/command/core/helpers/AbstractOutputWriterCommandStep.java
+++ b/liquibase-standard/src/main/java/liquibase/command/core/helpers/AbstractOutputWriterCommandStep.java
@@ -17,7 +17,7 @@ import java.util.List;
 
 public abstract class AbstractOutputWriterCommandStep extends AbstractHelperCommandStep implements CleanUpCommandStep {
 
-    private static OutputStreamWriter outputStreamWriter;
+    private OutputStreamWriter outputStreamWriter;
 
     @Override
     public List<Class<?>> providedDependencies() {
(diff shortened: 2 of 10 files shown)


piotrwielgolaski-tomtom:
@filipelautert I noticed issue with test, fixed it now

piotrwielgolaski-tomtom:
@filipelautert any chance to get this under review?

trico-tillster:
@piotrwielgolaski-tomtom I still see this issue with latest version 4.23 using -T2. Will your PR fix this? Thank you
`Error:  Failed to execute goal org.liquibase:liquibase-maven-plugin:4.23.0:update (liquibase) on project ordering-db-schema: Cannot end scope ztupxdjtqp when currently at scope root`

piotrwielgolaski-tomtom:
> @piotrwielgolaski-tomtom I still see this issue with latest version 4.23 using -T2. Will your PR fix this? Thank you

 Biased on my testing, yes it should, but keep in mind that I focused on offline database use case. 
 anyway for sure it will bring us closer for all cases.

trico-tillster:
@piotrwielgolaski-tomtom I've patched liquibase with this fix and I can confirm that the scope issue regarding maven concurrency that I was observing in 4.23.0 is fixed. Thank you
@filipelautert Thank you for picking this up and move it forward.

liquibase-standard/src/main/java/liquibase/Scope.java
```diff
@@ -113,21 +113,7 @@ public static Scope getCurrentScope() {
     }
 
     public static void setScopeManager(ScopeManager scopeManager) {
-        Scope currentScope = getCurrentScope();
```
filipelautert, line 116: Besides the "default value" is removed here, the getCurrentScope later on has similar logic to make sure that we don't have a null scope.

liquibase-standard/src/main/java/liquibase/command/core/helpers/AbstractOutputWriterCommandStep.java
```diff
@@ -17,7 +17,7 @@
 
 public abstract class AbstractOutputWriterCommandStep extends AbstractHelperCommandStep implements CleanUpCommandStep {
 
-    private static OutputStreamWriter outputStreamWriter;
+    private OutputStreamWriter outputStreamWriter;
```
filipelautert, line 20: Thanks, this static was incorrect .

filipelautert (approved):

liquibase-standard/src/main/java/liquibase/Scope.java
piotrwielgolaski-tomtom: It is important to not inject common scope for scope manager, as all child will inherit from it, which you don't want if scope manager should have isolated scope.

piotrwielgolaski-tomtom:
@rberezen it looks it is waiting for your approval, please find time to take a look at.

filipelautert:
Hello @piotrwielgolaski-tomtom - our build process is going through some refactoring and it is blocking @rberezen . Once it is done/fixed he will be able to finish the review here... Thanks for your patience!

rberezen (approved):

PR #4476, opened 2023-07-07
[DAT-15258] Makes sure that plugins that implements change are supporting the current database.

### LINKED ISSUE #4377: Prevent database-specific subclasses to interfere with executions targeting other databases
### Search first

### Description

# Subclassing use cases

The Neo4j extensions subclasses existing core changes for different purposes.

One of the purpose is to "alias" an existing element name.
For instance, `liquibase.ext.neo4j.change.CypherChange` subclasses `liquibase.change.core.RawSQLChange` so that users can write `<neo4j:cypher>...</neo4j:cypher>` instead of `<sql>...</sql>` (since SQL does not make much sense in Neo4j context).

Similarly, subclassing is also used to "alias" existing attribute names.
`liquibase.ext.neo4j.change.InsertNodeChange` subclasses `liquibase.change.core.InsertDataChange` to introduce the `labelName` attribute, as an "alias" to the RDBMS-centric `tableName` attribute: users can write `<neo4j:insert labelName="Person">...` instead of `<insert tableName="Person">...`.

Finally, subclassing is used to give complete control to the extension over what gets executed.
This is the case of `liquibase.ext.neo4j.change.LoadGraphDataChange`, subclassing `liquibase.change.core.LoadDataChange`. Here, the core logic of serializing the CSV data remains in core, but the statement creation is completely customized in the extension for Neo4j.

# Extension interference issue

The actual issue was first reported [here](https://github.com/liquibase/liquibase-neo4j/issues/413).

The problem occurs when database-specific change subclasses are available on the classpath (with a `liquibase.servicelocator.PrioritizedService#PRIORITY_DATABASE` priority) and Liquibase runs against another database (without any `dbms` filter).

In the issue above, the Neo4j extension is in the classpath and the user wants to run Liquibase against an Oracle database.

The execution path is something like this:

1.  `liquibase.command.core.helpers.DatabaseChangelogCommandStep#getDatabaseChangeLog` - which resolves the change log and its change sets
2. `liquibase.changelog.DatabaseChangeLog#handleChildNode(liquibase.parser.core.ParsedNode, liquibase.resource.ResourceAccessor, java.util.Map<java.lang.String,java.lang.Object>)` - which handles the deserialization of change log elements (like change sets and their changes)
3. `liquibase.changelog.ChangeSet#toChange` which handles the creation of the change of a given change set
4. `liquibase.plugin.AbstractPluginFactory#getPlugin`

The last listed method sorts all possible classes compatible with the requested change type by their declared priority and picks the first one.

The problem is that approach is that the class selection does not consider the target database at all.

In the issue linked above, `liquibase.ext.neo4j.change.LoadGraphDataChange` is selected as the best candidate for running `<loadData>` against Oracle and this fails, given Oracle does not support Cypher statements ([at least for now](https://github.com/oracle/pgql-lang)).

A more subtle issue is that `liquibase.ext.neo4j.change.InsertNodeChange` is also selected for `<insert>` and this specific subclass defines (well, used to define) the extra attribute `labelName` with a non-null default value (`""` instead of `null`). 
This results in a different change checksum than the one that was persisted during the first run, since the persisted checksum was obtained through `liquibase.change.core.InsertDataChange`.

### Steps To Reproduce

Do something similar to https://github.com/liquibase/liquibase-neo4j/blob/98594d36e27ab4cf775304084410829425bdfe8b/src/test/groovy/liquibase/ext/neo4j/e2e/MySQLMigrationIT.groovy#L36-L64 with the Neo4j extension on the classpath (at version `4.22.0.1`).

### Expected/Desired Behavior

Once this is fixed, I expect database-specific changes to not interfere with executions that do not target the databases they support (as defined by `liquibase.change.Change#supports`).

### Liquibase Version

all

### Database Vendor & Version

_No response_

### Liquibase Integration

core

### Liquibase Extensions

_No response_

_No response_

### Additional Context

I took a quick stab at the problem: https://github.com/liquibase/liquibase/pull/4364.
My approach consisted in propagating the `Database` object to `liquibase.plugin.AbstractPluginFactory#getPlugin` and pick the first class that actually supports the database.

This solved the problem but there is a major issue.

`liquibase.plugin.AbstractPluginFactory#getPlugin` is quite central to core, and adding a `Database` parameter there ends up requiring to spread that change to too many places, like [`liquibase.parser.ChangeLogParser#parse`](https://github.com/liquibase/liquibase/pull/4364/files#diff-c81f711b7d78cdbf5efc920189930dbe69baa82175d397601bc89ad486980f25R27).

Not only is it a breaking change, it is also not a desirable one. Change log parsers should not need to know anything about target databases. Their role is just to deserialize file representation of change log into their Java `ChangeLog` counterpart.

To quote @nvoxland on the topic, here is what he said on the approach that should likely be followed:

> The longer term fix is probably around a different separation between the database-independent logic in the change logic and the database-dependent logic elsewhere.
There are many ways to do that, and it will possibly be some level of breaking API changes in there. Like if the Change type really can't be based on the database, is it better to enforce that with having generateStatements not take a database call?  Or do we want Change to be selectable based on the database and so the ChangelogParser needs to parse out to ChangeDefintions or something which hold what the user asked for but isn't database specific? Or what will work best?
Maybe even should have more of a "pipeline" style API like we have for the Command classes where there isn't a specific Change implementation for each change at all, but instead ChangeSteps or something that attach themselves to the "loadData" change?
 

filipelautert:
## Impact

May Fix #4377 .


liquibase-standard/src/main/java/liquibase/plugin/AbstractPluginFactory.java
```diff
@@ -65,8 +67,15 @@ protected T getPlugin(final Object... args) {
             }
         }
 
-        if (applicable.size() == 0) {
+        if (applicable.isEmpty()) {
             return null;
+        } else if (applicable.size() == 1) {
+            return applicable.iterator().next();
+        } else if (getPluginClass().equals(Change.class)) {
```
filipelautert, line 74: I don't like this equals here , but I can't think of something better to address this problem.

StevenMassaro: I think the equals is okay, but it doesn't belong in the plugin factory. It should go in the `ChangeFactory`. Perhaps you need to add a new abstract method called `filter` or something like that, and provide a simple no-op implementation in the plugin factory (for backwards compatibility) and move this logic to the `ChangeFactory` class.

suryaaki2 (approved):

liquibase-standard/src/main/java/liquibase/plugin/AbstractPluginFactory.java
filipelautert: Thanks! As I was working with protected methods, I added a new protected one that returns all plugins.

StevenMassaro (approved):

liquibase-standard/src/main/java/liquibase/changelog/visitor/ValidatingVisitor.java
```diff
@@ -168,6 +169,52 @@ public void visit(ChangeSet changeSet, DatabaseChangeLog databaseChangeLog, Data
...
+    /**
+     * MongoDB's extension was incorrectly messing with CreateIndex checksum when the extension was added to the lib folder
+     * but a database other than mongodb was used. This method checks:
+     * * is it a CreateIndex change?
+     * * are we not using mongo?
+     * * do we have mongo extension loaded?
+     * * If I use CreateIndex from mongo extension, does the checksum matches?
+     * If everything matches than we fix the checksum on the database and say it's fine to continue.
+     */
+    private boolean isItAnOldButCorrectChecksumVersionGeneratedByABuggyExtension(ChangeSet changeSet, RanChangeSet ranChangeSet, DatabaseChangeLog databaseChangeLog) {
```
filipelautert, line 181: @StevenMassaro I added this new hack here to handle and fix incorrect checksums instead of just failing validation.

StevenMassaro: Oh this is yucky. Is this a product requirement?

filipelautert: Yes, this here -> https://datical.atlassian.net/browse/DAT-15258?focusedCommentId=128295

StevenMassaro: I don't think this code should be added to the Liquibase codebase. This is hard to test, and will have edge cases that we cannot imagine now. I fear that adding this could break some part of Liquibase for the majority of users who do not fall into this very narrow window of users we're trying to fix for. Additionally, I don't want to set a precedent that we will be adding in all kinds of hacky workarounds (no offense to your fix or code) to fix problems caused by other extensions.

The proper solution to a problem like this is education and documentation in the event that someone falls into the unlikely situation of meeting all of these criteria.

StevenMassaro (changes requested):

liquibase-standard/src/main/java/liquibase/changelog/visitor/ValidatingVisitor.java
filipelautert: @StevenMassaro @suryaaki2 I replaced the magic code by a warning message. It is working like that:

```
     1 changesets check sum
          changelogs/changelog/simple_changelog.xml::3::fl4 was: 9:6189b7cbca33bf5dec62fd48dd75c473 but is now: 9:1ef3d63cb4491a88e7d50778226c49ca
          Checksum value 9:6189b7cbca33bf5dec62fd48dd75c473 for CreateIndexChange probably was created by an old version of mongodb extension and needs to be fixed manually, changelogs/changelog/simple_changelog.xml::3::fl4
```

wwillard7800 (approved):

liquibase-standard/src/main/java/liquibase/changelog/visitor/ValidatingVisitor.java
StevenMassaro: Are we able to add some more instructions about how to fix it manually to that message?

PR #4483, opened 2023-07-11
Add ability to set labels and contexts in included changelogs DAT-15202

wwillard7800:
## Impact

## Description

Add ability to set labels and contexts in included changelogs.  Refactored `DatabaseChangeLog` class to allow `findResources` method to be called publicly.


liquibase-standard/src/main/java/liquibase/util/StringUtil.java
```diff
@@ -33,6 +34,27 @@ public static String trimToEmpty(String string) {
...
+        if (newParts.length == 1) {
+            return newParts;
+        }
+        String[] editedParts = new String[newParts.length];
+        AtomicInteger i = new AtomicInteger();
+        Arrays.stream(newParts).forEach(label-> {
+            editedParts[i.getAndIncrement()] = label.trim();
+        });
+        return editedParts;
+    }
```
StevenMassaro, line 56: This method already exists on line 421. Please use that one instead and remove this.

PR #4185, opened 2023-04-27
Adding support for Postgresql generated columns.

filipelautert:
## Impact

- [ ] Bug fix (non-breaking change which fixes expected existing functionality)
- [X] Enhancement/New feature (adds functionality without impacting existing logic)
- [ ] Breaking change (fix or feature that would cause existing functionality to change)
 
## Description

Adding support for Postgresql generated columns that were introduced in Postgresql 12 

* https://www.postgresql.org/docs/current/ddl-generated-columns.html
* Using information from pgjdbc driver: https://github.com/pgjdbc/pgjdbc/blob/ed209a6eee8611d1ed68b36096031268dc170914/pgjdbc/src/main/java/org/postgresql/jdbc/PgDatabaseMetaData.java#L1704

PR #4499, opened 2023-07-15
Fixed #4489: SQL Anywhere incorrectly reports VARCHAR(n CHAR) as VARCHAR(n)

mkarg:
## Description
Fixes #4489 

## Impact

- [x] Bug fix (non-breaking change which fixes expected existing functionality)
- [ ] Enhancement/New feature (adds functionality without impacting existing logic)
- [ ] Breaking change (fix or feature that would cause existing functionality to change)
 

## Description

Cause: There is no way to detect the size unit in the JDBC standard.

Solution: Using a native query to detect the size unit.

## Things to be aware of

N/A

## Things to worry about

N/A

## Additional Context

N/A

PR #4507, opened 2023-07-19
Fixing Issue #4429: SQL Anywhere: generateChangeLog fails when column's default value is an expression (e. g. Unexpected error running Liquibase: Unknown boolean value: if "mycol" is null then 0 else 1 endif)

### LINKED ISSUE #4429: SQL Anywhere: `generateChangeLog` fails when column's default value is an expression (e. g. `Unexpected error running Liquibase: Unknown boolean value: if "mycol" is null then 0 else 1 endif`)
### Search first

### Description

In SQL Anywhere, columns can have default values which are *no literals but expressions*, for example `if "mycol" is null then 0 else 1 endif`. Apparently Liquibase cannot support expressions but only literals, as the error message says:
```
Unexpected error running Liquibase: Unknown boolean value: if "mycol" is null then 0 else 1 endif
```
This is a showstopper for generating change log from existing SQL Anywhere database.

### Steps To Reproduce

```sql
CREATE TABLE A (
  c1 int,
  c2 bit NULL COMPUTE (if "c1" is null then 0 else 1 endif)
)
```
```bash
liquibase --changeLogFile=mydatabase_changelog.xml generateChangeLog
```

### Expected/Desired Behavior

Liquibase successfully creates change log including the column default value expression `if "mycol" is null then 0 else 1 endif"`.

### Liquibase Version

master

### Database Vendor & Version

SAP SQL Anywhere 17.0.11

### Liquibase Integration

N/A

### Liquibase Extensions

N/A

Windows 10 Pro (German)

### Additional Context

N/A

mkarg:
## Impact

Fixes #4429 

## Description

The cause of the reported problem was a combination of three original shortcomings, all of them get fixed by this PR:
* SQL Anywhere can defer foreign key constraints, but uses a different SQL syntax to define it.
* Adding support for SQL Anywhere generated columns.
* Unknown boolean value: if mycol is null then 0 else 1 endif

## Things to be aware of

As there is no native way to transport the information "this is a generated column" inside of Liquibase, this PR introduces the same workaround for SQL Anyhwere that was introduced in https://github.com/liquibase/liquibase/pull/4185 for PostgreSQL: Misusing the handling of computed *defaults*. As the word `COMPUTE` unambiguously indicates this misuse to the parser, this should not have negative side effects other than the fact that the changelog is specific to SQL Anywhere.

## Things to worry about

Misusing fields is not nice and counterintuitive. It would be good if we would have an original solution (hence: a separate changelog attribute) for generated columns.

## Additional Context

See `COMPUTE` clause in [SQL Anywhere Manual](https://help.sap.com/docs/SAP_SQL_Anywhere/93079d4ba8e44920ae63ffb4def91f5b/816ce9fb6ce210148da3ff81f8580ad0.html).


mkarg:
This PR is in draft-mode because:
* It essentially depends upon another bug fix which not yet is merged into `master`: https://github.com/liquibase/liquibase/pull/4499
* While the distribution bundle was manually tested to work correctly, the (anyways `@Ignore`d) unit test `testGeneratedColumn()` is failing its sole assertion.

mkarg:
I rebased on current master to resolve merge conflicts.

filipelautert (approved):

suryaaki2 (approved):

rberezen (approved):

Issue #4514, opened 2023-07-20
`generateChangelog` creates `defaultValue='null'` if LONG VARCHAR has no default value


mkarg:
### Search first

- [X] I searched and no similar issues were found

### Description

A `LONG VARCHAR` column that has no default value, or where the default values is set to the SQL keyword `NULL` (i. e. *not* the string literal `NULL`), `generateChangelog` produces `defaultValue="null"`, effectively leading to the string literal `null` being found in all inserted columns. In fact, it should produce `defaultValueComputed="NULL"` to prevent this.

### Steps To Reproduce

`CREATE TABLE T (C LONG VARCHAR DEFAULT NULL)`
`liquibase generateChangelog`

### Expected/Desired Behavior

`defaultValueComputed="NULL"`

### Liquibase Version

master

### Database Vendor & Version

SAP SQL Anywhere 17.0.11

### Liquibase Integration

N/A

### Liquibase Extensions

N/A

### OS and/or Infrastructure Type/Provider

Windows 10 Pro (German)

### Additional Context

Reproduced with SQL Anywhere, assuming this will happen on *any* RDBMS.

### Are you willing to submit a PR?

- [X] I'm willing to submit a PR (Thank you!)

PR #4515, opened 2023-07-20
Fixed: LONG VARCHAR DEFAULT NULL instead of LONG VARCHAR DEFAULT 'NULL'

mkarg:
## Impact

Fixes https://github.com/liquibase/liquibase/issues/4514.

## Description

This PR implements for `LONG VARCHAR` and `LONG NVARCHAR` the similar solution as implemented for `BLOB`, but with the distinction for single-quoted strings.

## Things to be aware of

The fix assumes that the RDBMS is compliant with the description found at https://docs.oracle.com/javase/8/docs/api/java/sql/DatabaseMetaData.html#getColumns-java.lang.String-java.lang.String-java.lang.String-java.lang.String-:
>COLUMN_DEF String => default value for the column, which should be interpreted as a string **when the value is enclosed in single quotes** (may be `null`)

## Additional Context

N/A

### DIFF
--- a/liquibase-standard/src/main/java/liquibase/util/SqlUtil.java
+++ b/liquibase-standard/src/main/java/liquibase/util/SqlUtil.java
@@ -237,11 +237,11 @@ public abstract class SqlUtil {
             } else if (typeId == Types.JAVA_OBJECT) {
                 return new DatabaseFunction(stringVal);
             } else if (typeId == Types.LONGNVARCHAR) {
-                return stringVal;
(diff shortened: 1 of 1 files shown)


mkarg:
This is a showstopper at least for SQL Anywhere.

filipelautert (approved):

rberezen:
This PR fixes an issue when `liquibase generateChangelog` creates defaultValue='null' if LONG VARCHAR or LONG NVARCHAR has no default value.

- @tati-qalified Manual testing is required  

APPROVED

rberezen (approved):

PR #4516, opened 2023-07-20
DAT-15470

sayaliM0412:
## Impact

resolving vulnerability in the build.yml file for caching dependencies

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -325,8 +325,14 @@ jobs:
       - name: Built Code Cache
         uses: actions/cache@v3.3.1
         with:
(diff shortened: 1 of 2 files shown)


jnewton03:
@sayaliM0412 is this the only cache action where we need this change or should we apply it elsewhere within this yaml file?

sayaliM0412:
> @sayaliM0412 is this the only cache action where we need this change or should we apply it elsewhere within this yaml file?

none of the other action has keys in it's /target

jnewton03 (approved):
looks good!  after its merged lets make sure those files aren't being cached any longer.

Issue #4512, opened 2023-07-20
SQL Anywhere: `generateChangelog` produces `DOUBLE(53)`, but `update` tells the JDBC driver doesn't like the length suffix `(53)`.


mkarg:
### Search first

- [X] I searched and no similar issues were found

### Description

We are doing `generateChangelog` from a SQL Anywhere database, and get the data type `DOUBLE(53)` in the generated changelog as a result. `validate` says this is nice, but `update` tells us that SQL Anywhere's JDBC driver does not support the length restriction `(53)`.

### Steps To Reproduce

Using SQL Anywhere:
`CREATE TABLE T (C DOUBLE)`
`liquibase generateChangelog`
`liquibase validate`
`liquibase update`

### Expected/Desired Behavior

Depends on how Liquibase is intended to work:
* Either `liquibase generateChangelog` should create `DOUBLE` instead of `DOUBLE(53)`.
* Or `liquibase update` should use `DOUBLE` when it is asked to create `DOUBLE(53)`.

### Liquibase Version

master

### Database Vendor & Version

SAP SQL Anywhere 17.0.11

### Liquibase Integration

N/A

### Liquibase Extensions

N/A

### OS and/or Infrastructure Type/Provider

Windows 10 Pro (German)

### Additional Context

N/A

### Are you willing to submit a PR?

- [X] I'm willing to submit a PR (Thank you!)

PR #4520, opened 2023-07-21
Fixed: SQL Anywhere reports fancy precision with DOUBLE, SMALLINT and TINYINT

mkarg:
## Impact

Fixes https://github.com/liquibase/liquibase/issues/4512 for `DOUBLE` but also for `SMALLINT` and `TINYINT`.
 
## Description

SQL Anywhere (inadvertently) reports rather fancy precision values for `DOUBLE`, `SMALLINT` and `TINYINT`.

"Inadvertently" because this RDBMS product is unable to customize the precision of these datatypes at all, so it makes no sense to report purely RDBMS-specific internal details. These details are documented or confirmed to be stable *nowhere* in the SQL Anywhere manuals. Hence, this information is *unstable* as it might change at any time without any further note.

"fancy" because the reported precision is not the actual precision, or it is not measured in the unit proposed by the ANSI standard or used by other RDBMS. Hence, this information is *useless*, as it is not the truth or misunderstood by the consumer.

This PR proposes a solution which effectively fixes two ends:
* Producer Side: `generateChangelog` will *not report* the fancy values, to prevent confusion on the consumer side.
* Consumer Side: `update` will *ignore* reported values, as SQL Anywhere's syntax parse otherwise refuses to process the DDL statement.

## Things to be aware of

N/A

## Things to worry about

N/A

## Additional Context

N/A


nvoxland (approved):

mkarg:
This is a showstopper at least for SQL Anywhere.

filipelautert (approved):
Thanks, everything here  only applies to Sql Anywhere.

suryaaki2 (approved):

rberezen (approved):

Issue #4530, opened 2023-07-24
SQL Anywhere generates "NO ACTION" where actually "SET DEFAULT" is configured


mkarg:
### Search first

- [X] I searched and no similar issues were found

### Description

`liquibase generateChangelog` produces `NO ACTION` in the changelog when a foreign key is configured as `SET DEFAULT` in SQL Anywhere.

### Steps To Reproduce

```sql
CREATE TABLE T (
  A INT DEFAULT 1 PRIMARY KEY,
  B INT,
  FOREIGN KEY (B) REFERENCES T (A) ON DELETE SET DEFAULT ON UPDATE SET DEFAULT
)
```
```bash
liquibase generateChangelog
```

### Expected/Desired Behavior

```xml
<addForeignKeyConstraint … onDelete="SET DEFAULT" onUpdate="SET DEFAULT" …/>
```

### Liquibase Version

master

### Database Vendor & Version

SAP SQL Anywhere 17.0.11

### Liquibase Integration

N/A

### Liquibase Extensions

N/A

### OS and/or Infrastructure Type/Provider

Windows 10 Pro (German)

### Additional Context

N/A

### Are you willing to submit a PR?

- [X] I'm willing to submit a PR (Thank you!)

mkarg:
There is no bug in Liquibase. The bug is in the JDBC driver sajdbc4.jar from SAP. I have reported this bug to SAP. Nevertheless, I will provide a PR providing a workaround.

Issue #4522, opened 2023-07-21
SQL Anywhere does not support "NO ACTION"


mkarg:
### Search first

- [X] I searched and no similar issues were found

### Description

`update` using `onDelete='NO ACTION'` and `onUpdate='NO ACTION'` fails on SQL Anywhere with the error message, that the syntax `NO ACTION` is unknown.

The reason is that SQL Anywhere does not support "no action". A foreign key in SQL Anywhere MUST support actions for both, `ON UPDATE` and `ON DELETE`, where *not specifying* an action is the same as `RESTRICT`.

### Steps To Reproduce

use `onDelete='NO ACTION'` in changelog

apply changelog using `liquibase update`

### Expected/Desired Behavior

Liquibase `validate` reports that `NO ACTION` is not supported.

Liquibase `update` applies SQL Anywhere's *implicit* default action, i. e. `RESTRICT`. While this is not what the author of the changelog wanted, it prevents the changelog from failing and is *safer* than any other action.

### Liquibase Version

master

### Database Vendor & Version

SAP SQL Anywhere 17.0.11

### Liquibase Integration

N/A

### Liquibase Extensions

N/A

### OS and/or Infrastructure Type/Provider

Windows 10 Pro (German)

### Additional Context

N/A

### Are you willing to submit a PR?

- [X] I'm willing to submit a PR (Thank you!)

PR #4531, opened 2023-07-24
No action

mkarg:
## Impact

Fixes  https://github.com/liquibase/liquibase/issues/4530
Fixes https://github.com/liquibase/liquibase/issues/4522

## Description

SQL Anywhere does not support `NO ACTION` (but supports `RESTRICT`, and uses `RESTRICT` as the default action if no action is provided), but its JDBC driver `sajdbc4.jar` reports `importedKeyNoAction` instead of `importedKeySetDefault`.

This PR changes Liquibase's behavior in three ways to deal with this situation in the least annoying way:
* If a changelog contains `NO ACTION` then for SQL Anyway *no* action is explicitly configured, resulting in effectively `RESTRICT` getting used. A validation warning is provided in that case.
* Liquibase will not generate changelogs containing `NO ACTION` when snapshotting SQL Anywhere, as that would be wrong; instead, `SET DEFAULT` is used, as this is correct.

## Things to be aware of

N/A

## Things to worry about

N/A

## Additional Context

N/A

### DIFF
--- a/liquibase-standard/src/main/java/liquibase/snapshot/jvm/ForeignKeySnapshotGenerator.java
+++ b/liquibase-standard/src/main/java/liquibase/snapshot/jvm/ForeignKeySnapshotGenerator.java
@@ -224,6 +224,10 @@ public class ForeignKeySnapshotGenerator extends JdbcSnapshotGenerator {
             if (jdbcType == DatabaseMetaData.importedKeyCascade) {
                 return ForeignKeyConstraintType.importedKeyCascade;
             } else if (jdbcType == DatabaseMetaData.importedKeyNoAction) {
+                if (database instanceof SybaseASADatabase) {
+                    //SQL Anywhere doesn't support NO ACTION, but reports it instead of SET DEFAULT
+                    return ForeignKeyConstraintType.importedKeySetDefault;
+                }
                 return ForeignKeyConstraintType.importedKeyNoAction;
             } else if (jdbcType == DatabaseMetaData.importedKeyRestrict) {
                 if (database instanceof MSSQLDatabase) {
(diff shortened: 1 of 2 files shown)


mkarg:
This is a showstopper for SQL Anywhere.

liquibase-standard/src/main/java/liquibase/snapshot/jvm/ForeignKeySnapshotGenerator.java
```diff
@@ -224,6 +224,10 @@ protected ForeignKeyConstraintType convertToForeignKeyConstraintType(Integer jdb
             if (jdbcType == DatabaseMetaData.importedKeyCascade) {
                 return ForeignKeyConstraintType.importedKeyCascade;
             } else if (jdbcType == DatabaseMetaData.importedKeyNoAction) {
+                if (database instanceof SybaseASADatabase) {
```
filipelautert, line 227: Seems we are missing an import here:

```
Error:  Failed to execute goal org.apache.maven.plugins:maven-compiler-plugin:3.11.0:compile (default-compile) on project liquibase-standard: Compilation failure
Error:  /home/runner/work/liquibase/liquibase/liquibase-standard/src/main/java/liquibase/snapshot/jvm/ForeignKeySnapshotGenerator.java:[227,41] cannot find symbol
Error:    symbol:   class SybaseASADatabase
Error:    location: class liquibase.snapshot.jvm.ForeignKeySnapshotGenerator
```

mkarg: Good catch! This is fixed in https://github.com/liquibase/liquibase/pull/4531/commits/7a2824f1e51c4f0e6de7673111d28375feecb17b.

In fact, I missed to rebuild after rebasing my fix on the master branch. As (unfortunately) the Liquibase project is not providing automatic PR builds to let a robot detect this, I have to perform (time-consuming) full builds of each separate bug fix manually, which is a bit annoying. 🙄

filipelautert (approved):
SQL Anywhere specific changes.

suryaaki2 (approved):

rberezen (approved):

PR #4533, opened 2023-07-24
decode URL encoded path in DirectoryPathHandler (DAT-15154)

StevenMassaro:
## Impact

### DIFF
--- a/liquibase-standard/src/main/java/liquibase/resource/DirectoryPathHandler.java
+++ b/liquibase-standard/src/main/java/liquibase/resource/DirectoryPathHandler.java
@@ -1,7 +1,12 @@
 package liquibase.resource;
 
+import liquibase.Scope;
+
 import java.io.*;
+import java.nio.charset.StandardCharsets;
 import java.nio.file.*;
+
+import static java.net.URLDecoder.*;
 
 /**
  * {@link PathHandler} that converts the path into a {@link DirectoryResourceAccessor}.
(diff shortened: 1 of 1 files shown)


abrackx (approved):

liquibase-standard/src/main/java/liquibase/resource/DirectoryPathHandler.java
```diff
@@ -33,6 +38,13 @@ public ResourceAccessor getResourceAccessor(String root) throws FileNotFoundExce
         root = root
                 .replace("file:", "")
                 .replace("\\", "/");
+        try {
+            // Because this method is passed a URL from liquibase.resource.ClassLoaderResourceAccessor.configureAdditionalResourceAccessors,
+            // it should be decoded from its URL-encoded form.
+            root = decode(root, StandardCharsets.UTF_8.name());
+        } catch (UnsupportedEncodingException e) {
+            Scope.getCurrentScope().getLog(getClass()).fine("Failed to decode path " + root + "; continuing without decoding.", e);
```
filipelautert, line 46: But if it is a `file://`  url should we not suppress this error message as we did the replace above?

StevenMassaro: Line 44 is primarily about removing things like `%20` from the `root` and replacing them with a space. I think keeping the error message could be useful.

filipelautert (approved):

PR #4534, opened 2023-07-24
Fix database test system external driver load.

filipelautert:
## Impact

## Description

If you try to load an external driver Liquibase will throw an exception saying that it could not load java.sql.DriverManager that is a standard java class.  This PR fix the issue.


mkarg (approved):
Thanks for fixing this!

MalloD12 (approved):
**Approved**. 

Change looks ok to me. Integrations tests shows `DatabaseTestSystem` keeps working fine with this change as well as it works fine loading external drivers.

PR #4537, opened 2023-07-25
SQL Anywhere: Switching off view definition checks to allow creation of dependent views even if dependency does not yet exist.

### LINKED ISSUE #4536: SQL Anywhere fails to create view-on-a-view if changes are performed in wrong order
### Search first

### Description

SQL Anywhere can define views on tables, but also views on views. At creation of the view it checks if the source table or source view does really exists. Whether this check succeeds is a race condition among the view creations: If V1 is a `SELECT` on V2, and V2 is created *after* V1, then `update` fails to create `V1`. Unfortunately, `generateChangelog` does not specify any particular order of the changes when reading them from a SQL Anywhere database, so `V1` *is* tried to get created before `V2` - hence `update` fails.

### Steps To Reproduce

```sql
CREATE TABLE T (C INT);
CREATE VIEW V2 AS SELECT * FROM T;
CREATE VIEW V1 AS SELECT * FROM V2;
```
```
liquibase generateChangelog
```
```
liquibase update
```

### Expected/Desired Behavior

Either `generateChangelog` produces a changelog specifying the needed sequence (V2 before V1), *or* `update` finds a way to silently resolve this situation.

### Liquibase Version

master

### Database Vendor & Version

SAP SQL Anywhere 17.0.11

### Liquibase Integration

N/A

### Liquibase Extensions

N/A

Windows 10 Pro (German)

### Additional Context

N/A

mkarg:
Fixes #4536 

## Impact

## Description

This PR introduces the exact same solution as SQL Anywhere's native schema-creation tool `dbunload` is using in the same situation: It temporarily switches off the check whether the dependencies found in the view definition already do exist.

See https://help.sap.com/docs/SAP_SQL_Anywhere/61ecb3d4d8be4baaa07cc4db0ddb5d0a/813bf68b6ce21014b288ca2a8ab530f9.html. While it says, this option is just for the use within `reload.sql` and `dbunload`, the fact that it is documented proofs that SAP is pretty aware of the fact that there is *no other* solution to the problem, and the solution *must* be usable outside of the `reload.sql` file. In fact, `reload.sql` in a nutshell could be seen as the output of `update-sql`, so our use is rather *valid*.

## Things to be aware of

SQL Anywhere will not complain anymore about *any* missing dependencies when creating views. Maybe someone *wants* `update` to fail? OTOH if he *wants* such a check, such a check would be a **new feature** for `validate`. IMHO it is a misuse of `update` to rely on its existing bugs, *and* the original SAP tool does *neither* provides such a checking feature!

## Things to worry about

Paraphrasing SAP's documentation, this should not get used *uncoditionally* with *all* `CREATE VIEW`, but *only* with the ones created by `generateChangelog`. So a 1:1 copy of SAP's native solution would *not* be to enable that option *in `update`*, but to create a `<sql dbms='asany'>` change by `generateChangelog`. OTOH that would render the changelog SQL specific, *and* it would not solve the problem when applying changelos to SQL Aynwhere created by other DBMSs.

IMHO the current PR is the best option for most users. If users actively complain about the solution, we could later add a PR with a new option to switch it off (and reveal the bug again).

## Additional Context

N/A


filipelautert:
@mkarg this one is having a test failure for asany as it is not expecting the set view:

```Error:  Failures: 
Error:    VerifyChangeClassesTest.compareGeneratedSqlWithExpectedSqlForMinimalChangesets:141 Unexpected difference in /home/runner/work/liquibase/liquibase/liquibase-standard/src/test/java/liquibase/verify/saved_state/compareGeneratedSqlWithExpectedSqlForMinimalChangesets/createView/asany.sql
Original:
[-- Database: asany
-- Change Parameter: selectQuery=SELECT id, name FROM person WHERE id > 10
-- Change Parameter: viewName=v_person
CREATE VIEW v_person AS SELECT id, name FROM person WHERE id > 10;]
New state:
[-- Database: asany
-- Change Parameter: selectQuery=SELECT id, name FROM person WHERE id > 10
-- Change Parameter: viewName=v_person
SET TEMPORARY OPTION force_view_creation='ON';
CREATE VIEW v_person AS SELECT id, name FROM person WHERE id > 10;]```

mkarg:
> @mkarg this one is having a test failure for asany as it is not expecting the set view:

@filipelautert Oops, sorry, seems I missed to push one changed file. Uploaded it a minute ago. Good catch, thank you! :-)

filipelautert (approved):
SQL anywhere specific changes.

suryaaki2 (approved):

rberezen (approved):

PR #4564, opened 2023-07-28
remove useless check

filipelautert:
## Impact

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -33,11 +33,6 @@ jobs:
     name: Check if Build should be done
     runs-on: ubuntu-22.04
     steps:
-      - name: Supress builds for PRs from branch github-action-*
-        if: github.event.pull_request && contains(github.head_ref, 'github-action')
-        run: |
-          echo "not building PRs with source branch github-action-*"
-          exit 1
       - name: Emit failure code for unsafe fork PRs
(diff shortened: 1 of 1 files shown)


sayaliM0412 (approved):
LGTM !!

StevenMassaro:
This was originally added as a security measure, I think. Why was it removed?

filipelautert:
@sayaliM0412 and I were trying to understand why it would be a security feature and could not find a case where it would prevent a insecure build, as the next check (for safeToBuild label) is the one preventing community pr's from being built. I believe it was ther just to make sure that we never execute any github-action from community?

Or maybe we could revert it but adding another check for the source repo like
```
 if: github.event.pull_request && contains(github.head_ref, 'github-action') && && github.event.pull_request.head.repo.full_name != 'liquibase/liquibase'
```

StevenMassaro:
> @sayaliM0412 and I were trying to understand why it would be a security feature and could not find a case where it would prevent a insecure build, as the next check (for safeToBuild label) is the one preventing community pr's from being built. I believe it was ther just to make sure that we never execute any github-action from community?
> 
> Or maybe we could revert it but adding another check for the source repo like
> 
> ```
>  if: github.event.pull_request && contains(github.head_ref, 'github-action') && && github.event.pull_request.head.repo.full_name != 'liquibase/liquibase'
> ```

I can't find anything in Jira that explains why it was added either. Looks like Andre added it, who no longer works here.

PR #4395, opened 2023-06-18
Improve update summary usability

EvaristeGalois11:
## Impact

- [ ] Bug fix (non-breaking change which fixes expected existing functionality)
- [ ] Enhancement/New feature (adds functionality without impacting existing logic)
- [x] Breaking change (fix or feature that would cause existing functionality to change)
 
## Description

Hi, I changed a bit the way the summary logs works because it was a bit ugly the way it rendered in a Spring application for example. I made a few small changes that are somewhat unrelated to this PR so let me know if you prefer to split them on their own proper PR.

This is what I've done:

- Skipped changesets are propagated to the changelog that includes them. I've done this mostly because I wanted to see the skipped changesets summary but couldn't quite understand why it wasn't produced. It seems that changesets included in other changelogs with the include or includeAll directive aren't considered for the summary recap. The fix is very simple this is why I'm including it in this PR, but if you think it isn't necessary or it needs more polishing I can branch it to a separate PR.
- Clean up the summary recap logged. In general it isn't necessary to include new lines in logged string because log aggregators can take care of collecting and presenting logs in the best way possible. This is why I removed the new line from the logged string, but still preserving them in the strings outputted to the standard output where new lines could be necessary.
- Remove the outputted summary recap unless the VERBOSE option is selected. This will partially mitigate https://github.com/liquibase/liquibase/issues/2396#issuecomment-1534159037 in the sense that by default the summary isn't spammed on the standard output anymore, but one have to opt in with the VERBOSE option. I don't know if it's considered a breaking changes, but this is the reason I selected the breaking checkbox above.
- Add the showSummary parameter to Maven and Spring.

## Things to be aware of

I didn't manage to run the tests so I don't know what it's broken and needs to be fixed. I would ask you to help a bit understanding what tests should be changed and if it's necessary to add new ones thanks.

## Additional Context

Sorry if it's a bit messy as a PR but it started with something totally unrelated and I ended up with a somewhat semi functional PR.

PR #4574, opened 2023-07-29
Implement showSummaryOutput parameter

EvaristeGalois11:
## Impact

## Description

Introduce a parameter `showSummaryOutput` that controls where to output the update summary generated. Possible values are `LOG`, `CONSOLE` and `ALL` (default to maintain backward compatibility).

## Things to be aware of
This is just a first attempt at tackling the issue to test if my approach is right. Stuff like the `showDetailTable` is very hard to crack because it has already some logic of differentiation between what to output on log and what on the output stream, so I left it alone for now. Moreover I didn't add the parameter to Maven or Spring because I fear to make a big mess of a merge with #4395.

## Additional Context

See the discussion under #4395 for more context


liquibase-integration-tests/src/test/resources/liquibase/extension/testing/command/update.test.groovy
```diff
@@ -42,6 +42,8 @@ Optional Args:
     OBFUSCATED
   showSummary (UpdateSummaryEnum) Type of update results summary to show.  Values can be 'off', 'summary', or 'verbose'.
     Default: SUMMARY
+  showSummaryOutput (UpdateSummaryOutputEnum) Where to report the results summary. Values can be 'log', 'console', or 'all'.
```
MalloD12, line 45: I would update this description to something like:
```
showSummaryOutput (UpdateSummaryOutputEnum) Summary output to report update summary results. Values can be 'log', 'console', or 'all'.
   Default: ALL
```

liquibase-standard/src/main/java/liquibase/util/ShowSummaryUtil.java
```diff
@@ -280,4 +278,29 @@ private static Writer createOutputWriter(OutputStream outputStream) throws IOExc
         String charsetName = GlobalConfiguration.OUTPUT_FILE_ENCODING.getCurrentValue();
         return new OutputStreamWriter(outputStream, charsetName);
     }
+
+    private static void writeMessage(String message, UpdateSummaryOutputEnum showSummaryOutput, OutputStream outputStream) throws LiquibaseException {
+        if (showSummaryOutput != UpdateSummaryOutputEnum.CONSOLE) {
```
MalloD12, line 283: Maybe something minor, but I would prefer seeing something like this:
```
switch(showSummaryOutput) {
  case UpdateSummaryOutputEnum.CONSOLE:
    writeToOutput(outputStream, message);
    break;
  case UpdateSummaryOutputEnum.LOG:
   writeToLog(message);
    break;
  default:
    writeToLog(message);
    writeToOutput(outputStream, message);
}
```

liquibase-integration-tests/src/test/resources/liquibase/extension/testing/command/update.test.groovy
EvaristeGalois11: Thank you, I've updated the description in all the place where it was used.

liquibase-standard/src/main/java/liquibase/util/ShowSummaryUtil.java
EvaristeGalois11: Done!

MalloD12:
Hi @EvaristeGalois11,

Thank you for your latest changes. I have been looking at these PR changes and they look good to me. I've been performing a bit of manual testing and the results look as expected for each of the supported values. 

The only thing I would like to add to this PR to move forward is some automated tests to make sure for example when `LOG` option is set for `showSummaryOutput` then update summary results should only be written in a log file. I started working on this yesterday, but couldn't make it works yet. 

Thanks,
Daniel.

MalloD12:
Hey @EvaristeGalois11,

I pushed a few tests I added to validate these changes. If you don't mind, could you please look at them and share any feedback? Or if you see they can improve in some way, please feel free to make any update you think necessary. For me, the rest of the things look good.

Just to let you, I'll wait until approving this PR since there is a build issue some members of the team are working on.  Once, that's resolved I'll approve it and move it to "Code Review" state so others can review it and test it as well. 

Thanks,
Daniel.

EvaristeGalois11:
Hi @MalloD12, I don't have much to say except that your tests look very good! :D

Regarding a more unrelated topic I took a look at #4584 and I think it's a very cool idea to default to `LOGGER` for `uiService` when using Spring, so I tried doing the same adding `showSummaryOutput` to `SpringLiquibase` with a default value of `LOG`. Let me know what you think about this and if I can do anything else for this PR or #4395.

MalloD12:
> 

HI @EvaristeGalois11,

Yeap, that sounds good to me.  

One thing I forgot to mention yesterday about this change which I think we should fix is when `showSummary=VERBOSE ` and `showSummaryOutput=CONSOLE` we are still displaying the below line in the log:

```
INFO [liquibase.util]    'changelog/xml/modifyChange.xml::2::mallod' : Context does not match 'test'
```

And that's because of this:

```
String skippedMessage = String.format("   '%s' : %s", st.getChangeSet().toString(), consumer.getMessage());
Scope.getCurrentScope().getLog(ShowSummaryUtil.class).info(skippedMessage);
```

that is part of `showDetailTable()` method. I think we might need to implement a similar logic that the one you did in `showSummary()` to depending in the output type set log this accordingly. 

Thanks,
Daniel.

EvaristeGalois11:
Hi @MalloD12, I thought a bit about the `showDetailTable` method and here is what I've done: 
```
String skippedMessage = String.format("   '%s' : %s", st.getChangeSet().toString(), consumer.getMessage());
Scope.getCurrentScope().getLog(ShowSummaryUtil.class).info(skippedMessage);
```
and
```
List<Integer> widths = new ArrayList<>();
widths.add(60);
widths.add(40);
Writer writer = createOutputWriter(outputStream);
TableOutput.formatOutput(table, widths, true, writer);
```
are basically showing the same thing just formatted in a different way, so it makes sense to me to group them together inside the `shouldPrintDetailTable` if. This all works out, but it has the property that the `skippedMessages` are considered the same of the detail table and they need the `VERBOSE` flag to be showed. It is admittedly different from before, but more coherent in my opinion.

I didn't update any test because I'm not sure if this is the right way to solve this problem, so I'm waiting until your feedback to update them. Thanks!

MalloD12:
Hi @EvaristeGalois11,

I've been reviewing your latest changes on Friday and they look good to me. From our side tests are ok, but I would like to we can get a clean run from some other test suites that belong to some other of our internal teams before approving this. Also, as you can see here one of the build checks is failing and some members of the team are actively working on this. I'll keep you posted on how these things go.

Again, very much appreciate all your work here!

Thanks,
Daniel.

MalloD12 (approved):
**Approved**.

Again, thank you @EvaristeGalois11 for submitting this PR and my apologies for the delay caused by the build issues. 

Code changes look good to me. It has been manually tested, plus some integration tests have been added to make sure show-summary output is redirected to the chosen output (`log`, `console`, or `both`), fixed some failing tests, etc.

Due to the build issues the team has found a workaround to be able to run build and execute all the checks. They all have been run successfully. 

Note: [PD-3649](https://datical.atlassian.net/browse/PD-3649) internal doc ticket has been created to reflect these changes in Liquibase docs. 

### Things to be aware of:
- None

### Things to worry about:
- None

filipelautert (approved):

wwillard7800 (approved):

suryaaki2 (approved):

rberezen (approved):

PR #4577, opened 2023-07-31
SQL Anywhere: Recompile all views after last change of update

### LINKED ISSUE #4575: SQL Anywhere fails to SELECT from view when update modified underlying table
### Search first

### Description

SQL Anywhere marks views as "invalid" when the underlying table changes. This is getting auto-fixed as soon as the view is SELECTed. Unfortunately this auto-fixing fails in some cases. As an effect, a SELECT on such a view results in an SQL error. The manual recommends to simply recompile all views in that case. Due to that limitation of SQL Anywhere, the original SQL Anywhere schema modification tool (`dbunload`) recompiles *all* views after the end of the last change *unconditionally* and ignores any failures, as this *never* induces *any* harm. It makes sense to mimic this behavior in Liquibase to overcome the situation that a *successful* update leads to an *unusable* database.

### Steps To Reproduce

```sql
CREATE TABLE T(c1 INTEGER, c2 INTEGER);
CREATE VIEW V1 AS SELECT * FROM T;
CREATE VIEW V2 AS SELECT * FROM V1;
```
```xml
<changeSet author="mkarg" id="fail-on-select">
    <dropColumn columnName="c1" tableName="T"/>
</changeSet>
```
`liquibase update`

### Expected/Desired Behavior

Liquibase should recompile all views after the last change was applied.

### Liquibase Version

master

### Database Vendor & Version

SAP SQL Anywhere 17.0.11

### Liquibase Integration

N/A

### Liquibase Extensions

N/A

Windows 10 Pro (German)

### Additional Context

N/A

mkarg:
## Impact

Fixes #4575 

## Description

The solution for #4575 simply is to execute the command `sa_recompile_views(1)` (see [SAP SQL Anywhere Manual](https://help.sap.com/docs/SAP_SQL_Anywhere/93079d4ba8e44920ae63ffb4def91f5b/8176e5da6ce21014b8619107e7d3d779.html)) once after the very last change was executed by `update`, which is exactly the same command that the original native tools from SAP execute to resolve this situation. This command never induces any risk, so unconditional execution is no problem in any case (which is why SAP's native tools *do* execute it unconditionaly).

## Things to be aware of

To make the code a little cleaner I extended the `Database` interface by a new `default` method, i. e. it implies no harm to existing code, but it opens the possibility for *all* RDBMS to execute some useful code in similar situations.

## Things to worry about

Despite the causeless warning in the SAP manual, as this is exactly the original solution applied by the original native SAP tools, there is absolutely nothing to worry about.

## Additional Context

N/A


MalloD12 (approved):
**Approved**.

Code changes looks good to me. Change made on update logic does not have any impact as this is only doing something for `SybaseASADatabase`.

Thank you @mkarg!

suryaaki2 (approved):

rberezen (approved):

liquibase-standard/src/main/java/liquibase/command/core/AbstractUpdateCommandStep.java
```diff
@@ -89,6 +89,8 @@ public void run(CommandResultsBuilder resultsBuilder) throws Exception {
                 ShowSummaryUtil.showUpdateSummary(databaseChangeLog, getShowSummary(commandScope), statusVisitor, resultsBuilder.getOutputStream());
             });
 
+            database.afterUpdate();
```
filipelautert, line 92: @StevenMassaro what do you think about this solution here? I"ve been thinking of options to handle this "afterUpdate" but I'm out of ideas.

StevenMassaro: What about putting this in the ChangeExecListener class? There is already a `ran` method there that you could key into.

mkarg: A `ChangeExecListener` will be invoked *per Changeset* by definition...
>An implementation of ChangeExecListener can be called by UpdateVisitor or RollbackVisitor for each changeset that is actually run.

 ...so it would be wrong to add an event handling method to it that is run after *the last* Changeset.

mkarg: A solution could be to rephrase the JavaDoc of `ChangeExecListener`. @filipelautert How to proceed?

filipelautert: @mkarg So I tried to get a working implementation using the ChangeExecListener approach but it was becoming a big refactoring to accommodate a simple update like this one = it is really meant to be executed for each changeset as the javadoc says, cause ChangeExecListener is called by UpdateVisitor that is called by changelogIterator and so on.
So let's go with your solution !
I just did a small refactoring to use the already existing sql executor method instead of creating a new one - could you confirm if it's fine for you?

filipelautert:
@mkarg I did a small refactoring to use the already existing sql executor method instead of creating a new one - could you confirm if it's fine for you? Then I'll get it merged for next release.

mkarg:
> @mkarg I did a small refactoring to use the already existing sql executor method instead of creating a new one - could you confirm if it's fine for you? Then I'll get it merged for next release.

LGTM. :-)

PR #4598, opened 2023-08-03
SQL Anywhere allows two foreign keys to differ only by name

### LINKED ISSUE #4597: SQL Anywhere allows two foreign keys to differ only by name
### Search first

### Description

SQL Anywhere allows to foreign keys to differ only by name.

Liquibase currently differs foreign keys only by name if it has no columns, or ignores the names.

As a result, two identical foreign keys found in an existing database will end up as just one foreign key produced by `liquibase generateChangelog`. This must not happen. It is always up to the database administrator to design this, as long as SQL Anywhere allows it. It is not Liquibase's job to change the design "on the fly".

### Steps To Reproduce

```sql
CREATE TABLE T (
  pk INT PRIMARY KEY,
  FOREIGN KEY fk1 (pk) REFERENCES T (pk),
  FOREIGN KEY fk2 (pk) REFERENCES T (pk)
)
```
```bash
liquibase generateChangelog
```

### Expected/Desired Behavior

Generated changelog must include *both* foreign keys, fk1 *and* fk2.

### Liquibase Version

master

### Database Vendor & Version

SAP SQL Anywhere 17.0.11

### Liquibase Integration

N/A

### Liquibase Extensions

N/A

Windows 10 Pro (German)

### Additional Context

N/A

mkarg:
## Impact

Fixes #4597 

## Description

Two foreign keys with different names are by definition *different* foreign keys on all RDBMS, not just on SQL Anywhere. It is the administrator's sole discretion to design two identical foreign keys with different names.

## Things to be aware of

This change affects *all* RDBMS, not just SQL Anywhere. It changes Liquibase's general behavior in such a way that the names of foreign keys *always* are considered when comparing two foreign keys. While these keys compare as "being equal", they are not "identical", hence `compare` is not be changed, while `equals` is.

## Things to worry about

N/A

## Additional Context

N/A

### DIFF
--- a/.github/workflows/build-branch.yml
+++ b/.github/workflows/build-branch.yml
@@ -69,6 +69,8 @@ jobs:
     permissions:
       contents: read
       packages: write
+    env:
+      GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
     steps:
       - uses: actions/checkout@v3
         with:
@@ -90,7 +92,6 @@ jobs:
       - name: Version Artifact
         run: |
            mvn versions:set "-DnewVersion=${{ needs.setup.outputs.thisBranchName }}-SNAPSHOT"
-           #mvn versions:set "-DnewVersion=${{ inputs.branchName }}-SNAPSHOT"
 
       # Publish to GitHub Packages
       - name: Publish package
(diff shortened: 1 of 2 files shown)


mkarg:
This is a showstopper for SQL Anywhere, as the decision of the administrator is not obeyed by Liquibase: The foreign key gets lost silently, but subsequent SQL might refer to its (now missing) name.

filipelautert (approved):
@rberezen all tests passed here -> https://github.com/liquibase/liquibase/pull/4777/checks

rberezen (approved):

suryaaki2 (approved):

PR #4621, opened 2023-08-08
Fixed: Issue #4620 - "SQL Anywhere DOES support CREATE OR REPLACE VIEW, despite what the source code says!"

### LINKED ISSUE #4620: SQL Anywhere DOES support CREATE OR REPLACE VIEW, despite what the source code says!
### Search first

### Description

The source code says that SQL Anywhere does not support CREATE OR REPLACE VIEW, and "thinks" it must skip handling the "OR REPLACE" case. In fact, simply removing this code proofs that SQL Anywhere works pretty well still (I tested it with SQL Anywhere 17.0.11, hence latest. I correctly handles CREATE VIEW just as CREATE OR REPLACE VIEW, and it works correctly with `replaceIfExists` when `true`, when `false` and when omitted. The manual proofs that this is officially supported: https://help.sap.com/docs/SAP_SQL_Anywhere/93079d4ba8e44920ae63ffb4def91f5b/816d37a36ce21014bf6cacd0588a42f3.html. Hence the is no need for *any* special treatment; we can simply remove the special cases here.

NB: It is true that `OR REPLACE` was not existing in *very, very, very* old releases of SQL Anywhere, but for at least ten years it is officially supported now. We do not have to take care for *such* old versions.

### Steps To Reproduce

```xml
<createView fullDefinition="false" viewName="MyView" replaceIfExists="true">SELECT 1</createView>
```
```bash
liquibase update-sql
```

### Expected/Desired Behavior

```sql
CREATE OR REPLACE VIEW MyView AS SELECT 1
```

### Liquibase Version

master

### Database Vendor & Version

SAP SQL Anywhere 17.0.11

### Liquibase Integration

N/A

### Liquibase Extensions

N/A

Windows 10 Pro (German)

### Additional Context

N/A

mkarg:
## Impact

## Description

Fixes #4620 

As SQL Anywhere supports OR REPLACE since more than a decade, we can safely remove all the special treatment for `replaceIfExists=true` in the source code. I used it for several hours with that change and it just works fine out-of-the-box in all cases, with `fullDefinition=true/false`, etc. Apparently this code was just a leftover from the very ancient days and is not needed since at least 10 years.

https://help.sap.com/docs/SAP_SQL_Anywhere/93079d4ba8e44920ae63ffb4def91f5b/816d37a36ce21014bf6cacd0588a42f3.html

## Things to be aware of

Very, very, very outdated versions of SQL Anywhere actually did not support OR REPLACE.

## Things to worry about

Do not use SQL Anywhere unsupported releases, i. e. v17.x, available since approx. a decade.

## Additional Context

N/A

### DIFF
--- a/liquibase-standard/src/main/java/liquibase/parser/AbstractFormattedChangeLogParser.java
+++ b/liquibase-standard/src/main/java/liquibase/parser/AbstractFormattedChangeLogParser.java
@@ -28,7 +28,7 @@ public abstract class AbstractFormattedChangeLogParser implements ChangeLogParse
     private static final ResourceBundle coreBundle = getBundle("liquibase/i18n/liquibase-core");
     private static final String EXCEPTION_MESSAGE = coreBundle.getString("formatted.changelog.exception.message");
 
-    protected final String FIRST_LINE_REGEX = String.format("%s\\s*liquibase formatted.*", getSingleLineCommentSequence());
+    protected final String FIRST_LINE_REGEX = String.format("^\\s*%s\\s*liquibase\\s*formatted.*", getSingleLineCommentSequence());
     protected final Pattern FIRST_LINE_PATTERN = Pattern.compile(FIRST_LINE_REGEX, Pattern.CASE_INSENSITIVE);
 
     protected final String PROPERTY_REGEX = String.format("\\s*%s[\\s]*property\\s+(.*:.*)\\s+(.*:.*).*", getSingleLineCommentSequence());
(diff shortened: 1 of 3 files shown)


MalloD12 (approved):
This looks good to me. Made a minor update on one of the test assertions related to this change.

There are still some build errors (not related with this PR) and functional and test harness test suites cannot be executed, but I double checked on both DB supported matrix and there is not SQL Anywhere execution we need to worry about.

filipelautert (approved):

suryaaki2 (approved):

rberezen (approved):

filipelautert (approved):

PR #4625, opened 2023-08-08
Handle leading spaces in formatted SQL changelog  DAT-15221

wwillard7800:
## Impact


liquibase-standard/src/main/java/liquibase/parser/core/formattedsql/FormattedSqlChangeLogParser.java
```diff
@@ -40,7 +40,7 @@ protected String getEndMultiLineCommentSequence() {
 
     @Override
     protected String getSingleLineCommentSequence() {
-        return "\\-\\-";
+        return "^\\s*\\-\\-";
```
vitaliimak, line 43: The liquibase-pro-mongodb extension overrides `getSingleLineCommentSequence()` method for FormattedMongoshChangelog. If this approach is approved, the extension will need to add regex to the overridden method too.
Wouldn't it be better to add regex `^\\s*` at the beginning of `FIRST_LINE_REGEX` string?

E.g.:
`protected final String FIRST_LINE_REGEX = String.format("^\\s*%s\\s*liquibase\\s*formatted.*", getSingleLineCommentSequence());`
It shouldn't force the extension to rewrite `getSingleLineCommentSequence()`.

wwillard7800: Good catch, and I agree.

vitaliimak (approved):

suryaaki2 (approved):

PR #4664, opened 2023-08-11
Fixed: #4662 - PostgreSQL: Doesn't support (n CHAR) syntax, but only (n) syntax.

### LINKED ISSUE #4662: PostgreSQL: Doesn't support (n CHAR) syntax, but only (n) syntax.
### Search first

### Description

Some RDBMS support units like CHAR in the length specifier, so schemas do exists using CHARACTER(n CHAR) or VARCHAR(n CHAR) units.

Unfortunately this leads to an SQL exception when applying such a schema on PostgreSQL, as that RDBMS does not support the CHAR specifier.

### Steps To Reproduce

```xml
<createTable tableName="T">
  <column name="c" type="VARCHAR(20 CHAR)"/>
</createTable)
```
```
liquibase update
```

### Expected/Desired Behavior

PostgreSQL ignores the `CHAR` unit, as it implicitly interprets the value as characters.

### Liquibase Version

master

### Database Vendor & Version

PostgreSQL 15.3

### Liquibase Integration

N/A

### Liquibase Extensions

N/A

Debian 12.2.0-14

### Additional Context

N/A

mkarg:
Fixes #4662 

## Impact

## Description

This PR cuts away the CHAR ending when it is found in the sole parameter, and the value's pattern is "the word CHAR follows a number". This is correct, as PostgreSQL expects the number to be given in characters, but just does not understand the subsequent word "CHAR".

## Things to be aware of

N/A

## Things to worry about

I noticed that special handling was in place producing "CHARACTER VARYING" when "CHARACTER (2147483647)" was found. I kept that solution and extended it to apply to "CHARACTER (2147483647 CHAR)", too. Nevertheless, IMHO that special handling is wrong for "CHARACTER" (while it is correct for "CHARACTER VARYING") as it produces a VARCHAR on the database, not a CHAR. As this failure existed before this PR, it is not the task of this PR to resolve that bug. Nevertheless, it makes sense to fix that in a separate PR eventually.

## Additional Context

N/A

### DIFF
--- a/liquibase-standard/src/main/java/liquibase/datatype/core/CharType.java
+++ b/liquibase-standard/src/main/java/liquibase/datatype/core/CharType.java
@@ -41,13 +41,19 @@ public class CharType extends LiquibaseDataType {
             type.addAdditionalInformation(getAdditionalInformation());
             return type;
         } else if (database instanceof PostgresDatabase){
-            if ((getParameters() != null) && (getParameters().length == 1) && "2147483647".equals(getParameters()[0]
-                .toString())) {
-                DatabaseDataType type = new DatabaseDataType("CHARACTER");
-                type.addAdditionalInformation("VARYING");
-                return type;
+            final Object[] parameters = getParameters();
+            if ((parameters != null) && (parameters.length == 1)) {
+                // PostgreSQL only supports (n) syntax but not (n CHAR) syntax, so we need to remove CHAR.
+                final String parameter = parameters[0].toString().replaceFirst("(?<=\\d+)\\s*(?i)CHAR$", "");
+                // PostgreSQL uses max. length implicitly if no length is provided to CHARACTER VARYING, so we can spare it.
+                if ("2147483647".equals(parameter)) {
+                    DatabaseDataType type = new DatabaseDataType("CHARACTER");
+                    type.addAdditionalInformation("VARYING");
+                    return type;
+                }
+                parameters[0] = parameter;
+                return new DatabaseDataType("CHARACTER", parameters);
             }
-            return super.toDatabaseDataType(database);
         } else if (database instanceof H2Database) {
             if (getRawDefinition().toLowerCase(Locale.US).contains("large object")) {
                 return new DatabaseDataType("CHARACTER LARGE OBJECT");
--- a/liquibase-standard/src/main/java/liquibase/datatype/core/CharType.java
+++ b/liquibase-standard/src/main/java/liquibase/datatype/core/CharType.java
@@ -52,7 +52,9 @@ public class CharType extends LiquibaseDataType {
                     return type;
                 }
                 parameters[0] = parameter;
-                return new DatabaseDataType("CHARACTER", parameters);
+                DatabaseDataType type = new DatabaseDataType(this.getName().toUpperCase(Locale.US), parameters);
+                type.addAdditionalInformation(this.getAdditionalInformation());
+                return type;
             }
         } else if (database instanceof H2Database) {
             if (getRawDefinition().toLowerCase(Locale.US).contains("large object")) {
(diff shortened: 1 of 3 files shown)


filipelautert:
Hello @mkarg ! Some tests are failing could you help to handle them? Thanks

mkarg:
> Hello @mkarg ! Some tests are failing could you help to handle them? Thanks

Hello @filipelautert! Thank you for telling me. I was PTO and have a lot of urgent tasks this week, but I will look into these failures ASAP.

filipelautert:
Hello @mkarg - the tests are failing because for Postgres we expect the type `varchar(255)` , but for this PR it is being generated as `character varying(255)` . Is it expected?

mkarg:
> Hello @mkarg - the tests are failing because for Postgres we expect the type `varchar(255)` , but for this PR it is being generated as `character varying(255)` . Is it expected?

Thank you, @filipelautert, for picking this up, I am still very busy with other projects at the moment.

Indeed, PostgreSQL does not know a type named `varchar` but only knows the type named `character varying` (see https://www.postgresql.org/docs/current/datatype-character.html). Hence it is correct to generate `character varying(255)`, while it is incorrect to expect `varchar(255)` on PostgreSQL.

BTW, this PR does not change the generated type name. Hence I assume some different PR broke the test while this PR was on sitting in the review queue.

filipelautert (approved):

suryaaki2 (approved):

liquibase-standard/src/main/java/liquibase/datatype/core/CharType.java
```diff
@@ -41,13 +41,21 @@ public DatabaseDataType toDatabaseDataType(Database database) {
...
             return type;
         } else if (database instanceof PostgresDatabase){
-            if ((getParameters() != null) && (getParameters().length == 1) && "2147483647".equals(getParameters()[0]
-                .toString())) {
-                DatabaseDataType type = new DatabaseDataType("CHARACTER");
-                type.addAdditionalInformation("VARYING");
+            final Object[] parameters = getParameters();
+            if ((parameters != null) && (parameters.length == 1)) {
+                // PostgreSQL only supports (n) syntax but not (n CHAR) syntax, so we need to remove CHAR.
+                final String parameter = parameters[0].toString().replaceFirst("(?<=\\d+)\\s*(?i)CHAR$", "");
```
rberezen, line 47: @mkarg when I check "(?<=\\d+)\\s*(?i)CHAR$" I receive an error that says "* A quantifier inside a lookbehind makes it non-fixed width"
do you think it may result in any problems?

rberezen: @mkarg I wanted to kindly ask if you could take a moment to review the comment I've left in the pull request. Your input and feedback would be greatly appreciated.

mkarg: Sorry for the delay. I will look into this tomorrow.

mkarg: Sorry for the delay. I checked the regex and cannot reproduce your problem in OpenJDK, as OpenJDK's parser is pretty happy with the quantifier and works perfectly well. Apparently the parser you are using is not compliant with the standardized API found in https://docs.oracle.com/javase/8/docs/api/java/util/regex/Pattern.html.

So I do not see a need to change my PR, actually.

rberezen: @mkarg thank you!

rberezen (approved):

adrian-velonis1:
No docs needed—no change to user experience.

PR #4668, opened 2023-08-13
Expand variables in the --sql argument to executeSql Github issue 4666

### LINKED ISSUE #4666: Env vars not being substituted in `execute-sql --sql-file` command
### Search first

### Description

In the `execute-sql` command, environment variables are not substituted in SQL files that are passed as a parameter. This does work in a "regular" setup when using a changelog file, but I don't want to use a changelog file.

### Steps To Reproduce

If I run:
```shell
export SOME_ENV_VAR=foo
liquibase --sql-file=some-file.sql
```
where `some-file.sql` looks like:
```sql
INSERT INTO some_table (value)
VALUES (${SOME_ENV_VAR});
```
The `${SOME_ENV_VAR}` doesn't get substituted.

### Expected/Desired Behavior

Environment variables are substituted in the `execute-sql` command as well.

### Liquibase Version

4.23.1

### Database Vendor & Version

Oracle

### Liquibase Integration

CLI

### Liquibase Extensions

_No response_

Ubuntu 20.04

### Additional Context

After looking at the Liquibase source code (for the first time, so bear with me please), it looks like it might be possible to refactor the `liquibase.changelog.ExpressionExpander#expandExpressions` method(s) to work without a `DatabaseChangeLog`. That parameter currently only seems to be required for [exception message purposes](https://github.com/liquibase/liquibase/blob/66e46b7372b85b5a3729c668bdf333989f030238/liquibase-standard/src/main/java/liquibase/changelog/ExpressionExpander.java#L61C101-L61C101) (it's also used to [retrieve a value](https://github.com/liquibase/liquibase/blob/66e46b7372b85b5a3729c668bdf333989f030238/liquibase-standard/src/main/java/liquibase/changelog/ExpressionExpander.java#L52C91-L52C91) with the `liquibase.changelog.ChangeLogParameters#getValue` method, but the `changeLog` parameter may be `null` there, so it's not actually required). This issue would be that that method depends on `ChangeLogParameters`, which the `liquibase.command.core.ExecuteSqlCommandStep` currently does not have of course. I see two options:
1. The `ExecuteSqlCommandStep` also constructs a `ChangeLogParameters` instance (but without a database set), and fills it with env vars like [so](https://github.com/liquibase/liquibase/blob/66e46b7372b85b5a3729c668bdf333989f030238/liquibase-standard/src/main/java/liquibase/changelog/ChangeLogParameters.java#L53C80-L53C80).
2. Make `ChangeLogParameters` extend an abstract class/interface that defines a `getValue` method. A new class could be created that extends the same abstract class/interface, but merely serves the purpose to store env vars.

I think option 1 is easier, but it feels more dirty as the whole idea of the `ExecuteSqlCommandStep` is that there is no change log, so constructing a `ChangeLogParameters` instance feels off. Option 2 would be a bit more work, but might make more sense in terms of class design. 

I would be willing to submit a PR for either option, but as I said before, I have very little experience with Liquibase itself, let alone contributing to Liquibase. If some people/a main contributor would confirm either option is the way to go, I'd definitely be willing to give it a shot and open a PR.

wwillard7800:
## Impact

Fixes #4666 


filipelautert (approved):

rberezen:
@wwillard7800 there are some test failures, would you mind to take a look?

tati-qalified (approved):

rberezen (approved):

PR #4680, opened 2023-08-15
Allow both include and includeAll to appear in modifyChangeSets tag

wwillard7800:
## Impact


suryaaki2 (approved):

suryaaki2:
@filipelautert Is this an XSD change?Do we need to run by anyone from CS team?

wwillard7800:
> @filipelautert Is this an XSD change?Do we need to run by anyone from CS team?

It is an XSD change, but it will not affect functionality.  The code has always been designed to accept both `include` and `includeAll` in the `modifyChangeSets` tag.  Users will need to upgrade their XSD version in order if they want both.

filipelautert (approved):
Before you had to use include or includeAll - now you can use both at same time. All fine, this isn't a breaking change .

PR #4697, opened 2023-08-18
Replacing special constants specific to SQL Anywhere by commonly supported JDBC Escape Functions

### LINKED ISSUE #4678: SQL Anywhere: CURRENT TIMESTAMP as an alias for CURRENT_TIMESTAMP
### Search first

### Description

While most RDBMS know `CURRENT_TIMESTAMP`, SQL Anywhere does not. Instead, it solely knows `CURRENT TIMESTAMP` (note the missing underscore replaced by blank).

For improved portability of changelogs, Liquibase should handle `CURRENT TIMESTAMP` as an alias for `CURRENT_TIMESTAMP`.

### Steps To Reproduce

```sql
CREATE TABLE T (c TIMESTAMP CURRENT TIMESTAMP)
```
```bash
liquibase generateChangelog
```

### Expected/Desired Behavior

```xml
<createTable name="T">
  <column name="c" type="TIMESTAMP" defaultValueComputed="CURRENT_TIMESTAMP"/>
</createTable>
```

### Liquibase Version

master

### Database Vendor & Version

SAP SQL Anywhere 17.0.11

### Liquibase Integration

N/A

### Liquibase Extensions

N/A

Windows 10 Pro (German)

### Additional Context

N/A

mkarg:
Closes #4678 

## Impact

## Description

Despite the proposal in #4678 I decided to no simply turn `CURRENT TIMESTAMP` into `CURRENT_TIMESTAMP`, but to turn it into JDBC Escape Functions, and to also do that with `CURRENT TIME`, `CURRENT DATE` and `CURRENT USER`. The justification is:
* The problem is not specific to timestamps, but covers all the `CURRENT` special keywords, hence it makes sense to cover not only `CURRENT_TIMESTAMP` but also `CURRENT_DATE`, `CURRENT_TIME` and `CURRENT_USER`.
* CURRENT_TIMESTAMP etc. might be an ANSI SQL standard, but not all databases do support it (e. g. SQL Anywhere itself fails to parse `CURRENT_TIME` and `CURRENT_DATE`, but accepts `CURRENT_TIMESTAMP` and `CURRENT_USER`). As Liquibase is JDBC based, it makes much more sense, to let SQL Anywhere produce JDBC Escape Functions instead, as those are understood by *all* JDBC drivers by definition.

## Things to be aware of

This solely effects the production of change sets, not its consumption, and it only effects SQL Anywhere.
This does not fix the case where a different RDBMS produces a change set, and SQL Anywhere shall consume it.
In that case, still SQL Anywhere will fail.
I deliberately have not resolved *that* case, as IMHO it would make much more sense that *all* RDBMS produce JDBC Escape Functions eventually, as the current N:M support (all RDBMS must have special cases for all other RDBMS) makes not much sense in a world that has JDBC Escape Funktions to boild this down into N:JDBC:M.

## Things to worry about

Maybe some customers do not like the JDBC Escape Syntax, so we might like to introduce a `--produce-jdbc-escape-syntax` switch having a default of `false`.

## Additional Context

N/A


filipelautert (approved):
SQL Anywhere specific changes.

rberezen (approved):

suryaaki2 (approved):

PR #4705, opened 2023-08-21
DAT-15338 Refactor connection string patterns to allow extensions to extend they

vitaliimak:
## Impact


StevenMassaro (approved):
This is outside the scope of this work, but this class would do with a refactor to use the service locator. This would allow plugins to provide their own database-specific filter patterns without requiring changes here.

filipelautert (approved):

vitaliimak:
Hi @StevenMassaro, if it is possible I'd like to do it. It looks like a much better approach than hardcoding new extension patterns inside the core codebase. The current fix cannot affect extension before the next liquibase release, so we do have time to make refactoring.
Do you agree @filipelautert?

vitaliimak:
Hello @StevenMassaro, @filipelautert 
Could you look at this variant of solving the issue?

StevenMassaro (approved):
Looks great, just needs some javadoc. This implementation will go a long way in improving the extensibility of Liquibase. Nice job!

filipelautert (approved):

filipelautert (approved):
After adding javadocs as Steve commented this will be fine!

vitaliimak:
Hello @filipelautert @StevenMassaro 
I've made more changes than only adding Javadoc. 
I've replaced the ConnectionPatterns interface with the abstract class, as it should store patterns. The abstract class is a better place for holding data than an interface. It allowed me to decrease fields visibility and add "setter" methods.
I've added Javadoc only for "setter" methods (public void addJdbcBlankPatterns(Map.Entry<Pattern, Pattern> jdbcBlankPatterns)), as API clients should use them to add more patterns. The getter methods are used only in the `JdbcConnection` class in the liquibase-core, I don't know if I need to add Javadoc to them as well.
What do you think?

liquibase-standard/src/main/java/liquibase/database/jvm/JdbcConnection.java
```diff
@@ -609,7 +580,11 @@ public boolean supportsBatchUpdates() throws DatabaseException {
         }
     }
 
-    private static class PatternPair {
+    /**
+     *  @deprecated this class has been moved to {@link ConnectionPatterns}, use that one instead.
+     */
+    @Deprecated
+    public static class PatternPair {
```
StevenMassaro, line 587: Can this class be deleted? If it was private before, it will not be a breaking change.

StevenMassaro (approved):
Looks good!

liquibase-standard/src/main/java/liquibase/database/jvm/JdbcConnection.java
vitaliimak: Oh! I missed that it was private. I'll remove it. Thanks!

PR #4707, opened 2023-08-21
DAT-15602

sayaliM0412:
## Impact

- update: run `branch-build.yml` on every PR commit
- update: run on every merge to `master`

### DIFF
--- a/.github/workflows/build-branch.yml
+++ b/.github/workflows/build-branch.yml
@@ -2,15 +2,11 @@ name: Build Branch SNAPSHOT
 
 on:
   workflow_dispatch:
-  push:
-    branches-ignore:
-      - master
   pull_request_target:
     types:
       - opened
       - reopened
       - synchronize
-      - labeled
 
 env:
   MAVEN_VERSION: '3.9.2'
--- a/.github/workflows/build-branch.yml
+++ b/.github/workflows/build-branch.yml
@@ -2,9 +2,6 @@ name: Build Branch SNAPSHOT
 
 on:
   workflow_dispatch:
-  push:
-    branches:
-      - master
   pull_request_target:
     types:
       - opened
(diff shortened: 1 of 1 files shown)


.github/workflows/build-branch.yml
```diff
@@ -3,14 +3,13 @@ name: Build Branch SNAPSHOT
 on:
   workflow_dispatch:
   push:
-    branches-ignore:
+    branches:
```
StevenMassaro, line 6: Doesn't this file https://github.com/liquibase/liquibase/blob/master/.github/workflows/build-main.yml cover the master build? Doesn't that make this a duplicate of that build?

.github/workflows/build-branch.yml
```diff
@@ -3,14 +3,13 @@ name: Build Branch SNAPSHOT
...
       - master
   pull_request_target:
     types:
       - opened
       - reopened
       - synchronize
-      - labeled
```
StevenMassaro, line 13: I don't think this should be removed. I misunderstood your initial question on slack.

StevenMassaro (changes requested):

.github/workflows/build-branch.yml
sayaliM0412: you are right. Run `master` on every push to `build-main` and run `build-branch` on every `pull_request_target` types specified

.github/workflows/build-branch.yml
sayaliM0412: just a note that if we have all of the above options enabled, it would trigger 2 runs when a PR is made, which is when we add a `label` and `opened` a PR.

StevenMassaro: I wonder if this will have an impact on adding the safetobuild label that @filipelautert does when reviewing community PRs.

sayaliM0412: When a safeToBuild label is added, it would trigger the build. I tested it on a community PR.

StevenMassaro: Okay sounds like it is safe to remove then.

StevenMassaro (approved):

PR #4728, opened 2023-08-24
fix: Add integration-test and sonar as dependencies for the package job

jandroav:
## Impact
 
## Description

chore(build.yml): Add integration-test and sonar as dependencies for the package job to ensure they are executed before packaging the artifacts

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -329,7 +329,7 @@ jobs:
   package:
     name: Package Artifacts
     runs-on: macos-latest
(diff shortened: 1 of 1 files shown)


sayaliM0412:
> I'd prefer we not do this. The build is already quite slow, and adding these dependencies will slow it down significantly more.

I agree with @StevenMassaro . We would like to separate the Sonar Scan entirely as @jnewton03 has mentioned in the past

sayaliM0412:
closing this PR

PR #4750, opened 2023-08-25
Added default method isFolder to Resource interface DAT-15388

wwillard7800:
Adding a new Resource interface method `isFolder` with a default implementation.  This will allow the S3Resource implementation to skip over folders that are returned during `getAll`.

## Impact

### DIFF
--- a/liquibase-standard/src/main/java/liquibase/resource/Resource.java
+++ b/liquibase-standard/src/main/java/liquibase/resource/Resource.java
@@ -30,6 +30,18 @@ public interface Resource {
     boolean isWritable();
 
     /**
+     *
+     * Default implementation
+     *
+     * @param   possibleFolder    Path to check for a folder
+     * @return  boolean
+     *
+     */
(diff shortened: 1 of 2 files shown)


liquibase-standard/src/main/java/liquibase/resource/Resource.java
```diff
@@ -29,6 +29,18 @@ public interface Resource {
...
+     *
+     * Default implementation
+     *
+     * @param   possibleFolder    Path to check for a folder
+     * @return  boolean
+     *
+     */
+    default boolean isFolder(String possibleFolder) {
+       return false;
+    }
```
StevenMassaro, line 42: Since this effectively has no implementation for anything other than S3, isn't this a bit of a concern? Someone might use this method thinking that it will actually tell them if something is a folder, but it will only work for S3.

wwillard7800:
Closing this PR, as the solution can be handled completely in the S3 extension.

PR #4762, opened 2023-08-26
Fetch ran change sets before updating check sums

### LINKED ISSUE #4460: Checksums are not actually upgraded when `FILENAME` starts with `classpath:`
### Search first

### Description

When a plain vanilla Spring Boot + Liquibase app is used with a file `src/main/resources/db/changelog/db.changelog-master.yaml` that contains changesets, then the persisted `FILENAME` will be `classpath:/db/changelog/db.changelog-master.yaml`.

When upgrading to Liquibase 4.23.0, then Liquibase will log that it is going to upgrade the checksum and will also log that it has successfully done so. However, that's not actually the case. All rows in `DATABASECHANGELOG` with a `FILENAME` starting with `classpath:` will not actually be updated.

The problem is that the `UPDATE` statement generated by Liquibase contains the condition `FILENAME = 'db/changelog/db.changelog-master.yaml'` which lacks the `classpath:`.

That the number of affected rows of the `UPDATE` statement is not checked to be actually 1 could also be considered as a problem.

### Steps To Reproduce

I have a full reproducer here: [hpoettker/liquibase-reproducer](https://github.com/hpoettker/liquibase-reproducer)

It is a plain vanilla Spring Boot app with `src/main/resources/db/changelog/db.changelog-master.yaml` containing

```yaml
databaseChangeLog:
  - changeSet:
      id: 0
      author: me
      changes:
        - sql:
            sql: CREATE TABLE `some_table` (`id` BIGINT)
```

The problem can be reproduced by first starting the app with e.g. Liquibase 4.17.2 and then Liquibase 4.23.0.

### Expected/Desired Behavior

The expected behavior is that all checksum are actually upgraded.

Furthermore, I'd expect that Liquibase fails fast when it tries to upgrade a checksum but the corresponding update statement does not actually affect any row.

### Liquibase Version

4.23.0

### Database Vendor & Version

MySQL 8.0.33

### Liquibase Integration

Spring Boot

### Liquibase Extensions

_No response_

Ubuntu

### Additional Context

_No response_

hpoettker:
## Impact

## Description

Fixes #4460.

The fix ensures that the field `storedFilePath` for each `changeSet` in `UpdateChangeSetChecksumGenerator` is set. Then, the `WHERE` condition of the produced `UPDATE` statements matches the database entry for the respective change set.

### DIFF
--- a/liquibase-integration-tests/src/test/groovy/liquibase/command/core/UpgradeChecksumVersionIntegrationTest.groovy
+++ b/liquibase-integration-tests/src/test/groovy/liquibase/command/core/UpgradeChecksumVersionIntegrationTest.groovy
@@ -14,6 +14,7 @@ import liquibase.extension.testing.testsystem.spock.LiquibaseIntegrationTest
 import liquibase.statement.core.RawSqlStatement
 import spock.lang.Shared
 import spock.lang.Specification
+import spock.lang.Unroll
 
 @LiquibaseIntegrationTest
 class UpgradeChecksumVersionIntegrationTest extends Specification{
@@ -34,11 +35,12 @@ VALUES('1', 'your.name', '$changelogfile', '2023-05-31 14:33:39.108', 1, 'EXECUT
 """))
     }
 
+    @Unroll
     def "update command should upgrade all checksums when no filters supplied" () {
         def changesetFilepath = "changelogs/common/checksum-changelog.xml"
         final ChangeLogHistoryService changeLogService = ChangeLogHistoryServiceFactory.getInstance().getChangeLogService(mysql.getDatabaseFromFactory())
         changeLogService.init()
-        insertDbclRecord(changesetFilepath)
+        insertDbclRecord(storedFilepathPrefix + changesetFilepath)
 
         when:
         def ranChangeSets = getRanChangesets(changeLogService)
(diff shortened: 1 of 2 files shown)


hpoettker:
@filipelautert Do you mean the integration tests for Oracle and MariaDB?

That is indeed curious. Both tests are failing when creating the sequence `seqtestdatatypeonall` with

- Oracle: `CREATE SEQUENCE LBUSER.seqtestdatatypeonall AS int`
- MariaDB: `CREATE SEQUENCE lbcat.seqtestdatatypeonall AS int`

I'd say that the `AS int` might work on PostgreSQL but is indeed invalid syntax for Oracle or MariaDB.

I don't see how my proposed change could cause such a problem. I also don't see how I can easily reproduce the test setup for the integration tests.

hpoettker:
I found the yaml to get the integration tests working locally. 😄 

The original PR was indeed erroneous. Using the constructor of `ChangeLogIterator` that takes the list of ran change sets seemed like an elegant idea at first sight. But it really does something much more specific than setting the stored file path in change sets.

I've also amended `UpgradeChecksumVersionIntegrationTest` with a test to reproduce the issue that this PR addresses.

filipelautert (approved):

filipelautert (approved):
Nice, good idea to fix it on ValidatingVisitor. In fact the Visitor is becoming a "fix it all" Visitor - something to think about .

@rberezen functional tests are passing here -> https://github.com/liquibase/liquibase-pro-tests/actions/runs/6038104842

suryaaki2 (approved):

rberezen (approved):

PR #4834, opened 2023-09-04
Optimise the logic for tableIsEmpty check constraint 

### LINKED ISSUE #4734: Optimise the logic for verifying if a table is empty
### Search first

### Description

Originally metioned [here](https://github.com/liquibase/liquibase/issues/1549).

Currently, to check if a table is empty, Liquibase is using `JdbcExecutor.queryForInt(..)`, which isn't efficient for tables with a large number of rows. Furthermore, if a table contains more rows than the maximum Integer number, the run will fail with an exception.

### Steps To Reproduce

Create a changeSet that includes the TableIsEmpty precondition, which will call the `check` function in the RowCountPrecondition class.

If the table that the changeSet is being applied to has over Integer.MAX_VALUE rows, the run will fail with an exception.

### Expected/Desired Behavior

The method for checking if a table is empty should be something similar to checking if there is a first row or not, instead of counting all of them and seeing if the number of rows is 0.

### Liquibase Version

All

### Database Vendor & Version

All

### Liquibase Integration

All

### Liquibase Extensions

_No response_

_No response_

### Additional Context

_No response_

filipelautert:
## Impact

## Description

Optimise the logic for verifying if a table is empty: instead of selecting count the table, uses exists. 

## Things to be aware of

SQL Ansi for all databases query.

## Things to worry about

NA

## Additional Context

Added tests. Fixes #4734

### DIFF
--- a/liquibase-integration-tests/src/test/resources/changelogs/common/common.tests.changelog.xml
+++ b/liquibase-integration-tests/src/test/resources/changelogs/common/common.tests.changelog.xml
@@ -484,8 +484,20 @@
         </createTable>
     </changeSet>
     <changeSet id="sampleCSVData1" author="nvoxland">
+        <preConditions>
+            <tableIsEmpty tableName="csvdata"/>
+        </preConditions>
         <loadData tableName="csvdata" file="changelogs/sample.data1.csv"/>
     </changeSet>
+
+    <changeSet id="tableIsEmptyShouldFail" author="flautert">
+        <preConditions onFail="MARK_RAN">
+            <tableIsEmpty tableName="csvdata"/>
+        </preConditions>
+        <!-- if table is dropped then next changesets will fail, thus it is an indirect check -->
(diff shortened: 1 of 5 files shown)


suryaaki2 (approved):

rberezen (approved):
@tati-qalified could you please check if tableIsEmpty still works as expected? 
Also created https://datical.atlassian.net/browse/PD-3741 cc @filipelautert 
https://docs.liquibase.com/concepts/changelogs/preconditions.html#available-preconditions

rberezen:
Fix for the test failure https://github.com/liquibase/liquibase-pro-tests/pull/969

tati-qalified:
@rberezen the precondition is working as expected with an xml format changelog. I wasn't able to find how to implement the precondition for other formats, if it's even supported, so I didn't test those cases.

tati-qalified (approved):

PR #4897, opened 2023-09-12
Fixes support for special characters in schema name

### LINKED ISSUE #1640: Liquibase failed when hyphen char in schema name
Hi

As I understand according to [CORE-1371](https://liquibase.jira.com/browse/CORE-1371) this problem was fixed in liquibase 3.0.3

We are using the latest 4.2.2 however still getting a similar problem, here is the output log:

```
Starting Liquibase at 12:26:49 (version 4.2.2 #36 built at 2020-12-09 20:07+0000)
Unexpected error running Liquibase: liquibase.exception.DatabaseException: liquibase.exception.DatabaseException: 
Connection could not be created to jdbc:mysql://<***skipped-host***>:3360/timetable-solver-service_nspace25_INT_1329?createDatabaseIfNotExist=true with driver com.mysql.cj.jdbc.Driver.  
You have an error in your SQL syntax; check the manual that corresponds to your MySQL server version for the right syntax to use near '-solver-service_nspace25_INT_1329' at line 1
For more information, please use the --logLevel flag
[2021-01-20 12:26:52] SEVERE [liquibase.integration] Unexpected error running Liquibase: liquibase.exception.DatabaseException: liquibase.exception.DatabaseException: Connection could not be created to jdbc:mysql://<***skipped-host***>.com:3360/timetable-solver-service_nspace25_INT_1329?createDatabaseIfNotExist=true with driver com.mysql.cj.jdbc.Driver.  You have an error in your SQL syntax; check the manual that corresponds to your MySQL server version for the right syntax to use near '-solver-service_nspace25_INT_1329' at line 1
```

the schema name is defined by this pattern:
`schema-name: ${spring.application.name}_${ENV_SPACE}_${DB_SCHEMA_POSTFIX}`
where: 
`${spring.application.name}` = Service name
`${ENV_SPACE}` = Kubernetis space
`${DB_SCHEMA_POSTFIX}` = Git branch name

This approach was working for many services until we faced a case where service names contain a hyphen in the name.
We obviously can rename this parameter and replace all hyphens with underscores, but still... it should work with hyphens according to CORE-1371 fix

Any ideas?

ahus1:
## Impact

## Description

Previously the `DatabaseUtils` class didn't add quotes for schema names where it should have done it. Due to that, schema names with a dash (`-`) were not working. 

This fixes it so schema names with a dash now work. 

I checked it manually for PostgreSQL and it works as expected for me. 
This affects several databases looking both at the original issue and the code, and I applied the change for all of them.

Fixes #1640

## Things to be aware of

Looking at the code of `escapeObjectName` it will only add quotes where needed, so the existing behavior should not be affected.

## Things to worry about

I didn't add a new test. A pointer to the test base would be appreciated.

### DIFF
--- a/liquibase-standard/src/main/java/liquibase/database/core/DatabaseUtils.java
+++ b/liquibase-standard/src/main/java/liquibase/database/core/DatabaseUtils.java
@@ -8,6 +8,7 @@ import liquibase.exception.DatabaseException;
 import liquibase.executor.Executor;
 import liquibase.executor.ExecutorService;
 import liquibase.statement.core.RawSqlStatement;
+import liquibase.structure.core.Catalog;
 import liquibase.structure.core.Schema;
 import liquibase.util.StringUtil;
 
@@ -45,7 +46,7 @@ public class DatabaseUtils {
                     if (GlobalConfiguration.PRESERVE_SCHEMA_CASE.getCurrentValue()) {
                         finalSearchPath = ((PostgresDatabase) database).quoteObject(defaultSchemaName, Schema.class);
                     } else {
-                        finalSearchPath = defaultSchemaName;
+                        finalSearchPath = database.escapeObjectName(defaultSchemaName, Schema.class);
                     }
 
                     if (StringUtil.isNotEmpty(searchPath)) {
@@ -66,6 +67,7 @@ public class DatabaseUtils {
                 if (schema == null) {
                     schema = defaultSchemaName;
                 }
+                database.escapeObjectName(schema, Schema.class);
                 executor.execute(new RawSqlStatement("SET CURRENT SCHEMA "
                         + schema));
             } else if (database instanceof MySQLDatabase) {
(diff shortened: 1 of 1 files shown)


MalloD12:
Hi @ahus1,

Thank you for taking the time to create a PR for #1640. Code changes look good to me.

I've been trying to reproduce this issue against master and couldn't reproduce it. I tried to deploy a changeset to create a table having a database whose name contains `-` and also having a schema name that also contains `-`. I first didn't specify  the schemaName on the create table change:

Screenshot 2023-09-29 at 1 16 37 PM
 
And then I included it:
Screenshot 2023-09-29 at 1 18 47 PM

As you can on the above screenshots both executions were successful. Am I missing any steps to reproduce this issue? Could you please share yours?

Regarding testing, I was thinking for this case we could add a new integration test in the **AbstractIntegrationTest** class where basically we will simply execute an update of a new changelog we create. This changelog will only contain a changeset like the one below:

```
changeSet id="schemaNameHyphenTest" author="mallod">
        <sql>
            create schema "this-is-a-schema-name-test"
        </sql>
        <createTable tableName="defaultValueTest" schemaName="this-is-a-schema-name-test">
            <column name="id" type="int"/>
            <column name="intA" type="int"/>
        </createTable>
</changeSet>
```

This integration test will be executed for several databases as displayed in the below screenshot:
Screenshot 2023-09-29 at 1 51 11 PM

I haven't tried this yet, so I'm not sure if there might be any particular DB that does not support this way of creating a schema.

Thanks,
Daniel.

MalloD12:
Hi @ahus1,

Could you please share any additional details that can help me to reproduce this issue and then validate this fix? Once, that's done then we can move forward with this fix through our review process.

Thanks,
Daniel.

MalloD12:
Moving this PR to "In Discussion" until we get back some feedback from the above comments.

lbadam:
I'm going to close this PR as we can't reproduce and haven't heard back in a while about the issue. If it is still an issue on newer versions (currently 4.29.2) please reopen.

PR #4906, opened 2023-09-13
remove labelled trigger from build.yml

StevenMassaro:
## Impact

## Description

For each standalone OSS PR, we're currently triggering two builds and cancelling one. Sometimes this means that you don't get any build results in the PR, [like here](https://github.com/liquibase/liquibase/pull/4904). I don't think we need this trigger on build.yml any longer, now that we've removed the safeToBuild label logic.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -18,7 +18,6 @@ on:
       - opened
       - reopened
       - synchronize
-      - labeled
 
   push:
     branches:


jnewton03 (approved):

sayaliM0412 (approved):

filipelautert (approved):
Let's give it a try!

PR #4925, opened 2023-09-18
CommandTests: copy files on the fly, not at startup

StevenMassaro:
## Impact

## Description

This changes the behavior of the `CommandTests.copyResource` method to copy the test files right before test execution, rather than copying them all at test framework startup. This is advantageous for two reasons:
- when running a subset of the command tests, the tests will run faster because there are fewer unrelated copy actions occurring
- each test is more atomic, and will rely less on the copied test artifacts from a previous test run


wwillard7800:
I think we also have the ability to copy resources just-in-time by setting up a closure, if I'm not mistaken.

filipelautert (approved):

PR #4959, opened 2023-09-27
Remove unnecessary conditions in IndexComparator

asashour:
This PR removes two unnecessary conditions, which always evaluate to `true`, as detected by IntelliJ.

The reason is: in line 71, if any of them is 0, then the method return.

And since they are the result of `list.size()`, they can not be negative.

And because of the previous two points, both must be more than 0, and so the conditions are not needed.


MalloD12 (approved):
Approved.

Code change makes sense to me. Build and all tests check have been successfully executed.
Screenshot 2023-10-02 at 7 10 47 PM


### Things to be aware of:
- None

### Things to worry about:
- None

filipelautert (approved):

PR #4997, opened 2023-10-02
DAT-16016

sayaliM0412:
## Impact

create draft release for push `master` and `release-**`
NOTE: use `refs/heads/release-**` if `release-**` does not work

### DIFF
--- a/.github/workflows/release-drafter.yml
+++ b/.github/workflows/release-drafter.yml
@@ -2,9 +2,13 @@ name: Release Drafter
 
 on:
   workflow_dispatch:
+  #This job will be triggered when changes are pushed to the master branch, and it will also be triggered when a branch is created with a name that starts with 'refs/heads/releases-'
   push:
     branches:
-      - master
+      push:
+        branches:
+          - master
+          - 'release-**'
     
 permissions:
   contents: read
@@ -19,7 +23,13 @@ jobs:
         pull-requests: write
     runs-on: ubuntu-latest
     steps:
-      # Drafts your next Release notes as Pull Requests are merged into the default branch
-      - uses: release-drafter/release-drafter@v5
+      - name: Check branch name
+        id: branch_check
+        #-qE : quiet mode Search for a regular expression pattern in the input text using extended(E) regular expressions
+        run: echo "is_release_branch=$(echo $GITHUB_REF | grep -qE 'refs/heads/(release-|master)' && echo true || echo false)" >> "$GITHUB_ENV"
+
+      - name: Draft Release
+        if: steps.branch_check.outputs.is_release_branch == 'true'
+        uses: release-drafter/release-drafter@v5
         env:
           GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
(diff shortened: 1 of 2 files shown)


.github/workflows/release-drafter.yml
```diff
@@ -2,9 +2,13 @@ name: Release Drafter
...
 on:
   workflow_dispatch:
+  #This job will be triggered when changes are pushed to the master branch, and it will also be triggered when a branch is created with a name that starts with 'refs/heads/releases-'
   push:
     branches:
-      - master
+      push:
+        branches:
+          - master
+          - 'release-**'
```
filipelautert, line 11: so we call it a rule... release branches needs to start with "release-"

.github/workflows/release-drafter.yml
```diff
@@ -19,7 +23,13 @@ jobs:
         pull-requests: write
     runs-on: ubuntu-latest
     steps:
-      # Drafts your next Release notes as Pull Requests are merged into the default branch
-      - uses: release-drafter/release-drafter@v5
+      - name: Check branch name
+        id: branch_check
+        #-qE : quiet mode Search for a regular expression pattern in the input text using extended(E) regular expressions
+        run: echo "is_release_branch=$(echo $GITHUB_REF | grep -qE 'refs/heads/(release-|master)' && echo true || echo false)" >> "$GITHUB_ENV"
```
filipelautert, line 29: That's a strange expression..... what is it doing?

sayaliM0412: it would check whether the value of $GITHUB_REF contains either "refs/heads/release-" or "refs/heads/master." If it does, it sets the is_release_branch environment variable to "true." Otherwise, it sets it to "false." The result is then appended to the $GITHUB_ENV environment, making it available for subsequent steps in the GitHub Actions workflow. I am still unsure if it would match with just `release-**` or does it need `refs/heads/(release-|master)`. So I decided to keep it `refs/heads/(release-|master)` . I should make that change in on:push: branches: as well.

.github/workflows/release-drafter.yml
sayaliM0412: I think having a set rule for branch name would be a better standard practise to have. Please let me know if you feel otherwise or want the branch-name to start with something else.

filipelautert (approved):

jnewton03 (approved):

.github/workflows/release-drafter.yml
```diff
@@ -19,7 +21,13 @@ jobs:
...
+
+      - name: Draft Release
+        if: steps.branch_check.outputs.is_release_branch == 'true'
+        uses: release-drafter/release-drafter@v5
         env:
-          GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
\ No newline at end of file
+          GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
```
jandroav, line 33: And what if `steps.branch_check.outputs.is_release_branch` is false? No draft release for master?

sayaliM0412: The `if` condition above will check if the branch name starts with `release-**` or `master`. And set it to true. The condition will be false only if the branch name is something other than `release-**` or `master`. But my concern now is, we cannot create 2 Draft release notes with same tag. And what if we have both the branches present.

jandroav: got it. we can create 2 drafts with the same tag only if we have 2 branches.

this will work:
master - tagA
release - tagA

this wont work:
master - tagA
master - tagA

sayaliM0412:
running successfully : https://github.com/liquibase/liquibase-pro/actions/runs/6405977339

PR #5012, opened 2023-10-04
DAT-15755 - Sonar issues fix

MalloD12:
## Impact

### DIFF
--- a/liquibase-maven-plugin/src/main/java/org/liquibase/maven/plugins/AbstractLiquibaseMojo.java
+++ b/liquibase-maven-plugin/src/main/java/org/liquibase/maven/plugins/AbstractLiquibaseMojo.java
@@ -791,7 +791,7 @@ public abstract class AbstractLiquibaseMojo extends AbstractMojo {
                     Database database = null;
                     try {
                         if (databaseConnectionRequired()) {
-                            database = getDatabase(mavenClassLoader);
+                            database = createDatabase();
                             DbUrlConnectionCommandStep.logMdc(url, database);
                             try (Liquibase liquibase = createLiquibase(database)) {
 
@@ -858,7 +858,15 @@ public abstract class AbstractLiquibaseMojo extends AbstractMojo {
         }
     }
 
-    protected Database getDatabase(ClassLoader mavenClassLoader) throws DatabaseException {
+    /**
+     *  Creates a database implementation setting up all its properties and returns a Database object.
+     *
+     * @return Database
+     * @throws DatabaseException
+     * @throws MojoExecutionException
+     */
+    protected Database createDatabase() throws DatabaseException, MojoExecutionException {
+        ClassLoader mavenClassLoader = getClassLoaderIncludingProjectClasspath();
         String dbPassword = (emptyPassword || (password == null)) ? "" : password;
         String driverPropsFile = (driverPropertiesFile == null) ? null : driverPropertiesFile.getAbsolutePath();
         Database database = CommandLineUtils.createDatabaseObject(mavenClassLoader,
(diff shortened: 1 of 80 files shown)


liquibase-maven-plugin/src/main/java/org/liquibase/maven/plugins/AbstractLiquibaseMojo.java
```diff
@@ -875,6 +858,28 @@ public void execute() throws MojoExecutionException, MojoFailureException {
         }
     }
 
+    protected Database getDatabase(ClassLoader mavenClassLoader) throws DatabaseException {
```
filipelautert, line 861: I think this method could be called something like `createDatabase` and some javadoc would be nice. When I saw `getDatabase ` being used in a `try with resources`  in `AbstractLiquibaseUpdateMojo `at first i tough it was a property getter.

liquibase-maven-plugin/src/main/java/org/liquibase/maven/plugins/AbstractLiquibaseUpdateMojo.java
```diff
@@ -53,10 +55,11 @@ protected void printSettings(String indent) {
     getLog().info(indent + "number of changes to apply: " + changesToApply);
   }
 
-  protected void handleUpdateException(LiquibaseException exception) throws LiquibaseException {
-    try {
+  protected void handleUpdateException(LiquibaseException exception) throws LiquibaseException, MojoExecutionException {
+    ClassLoader mavenClassLoader = getMavenArtifactClassLoader();
```
filipelautert, line 59: Could this ClassLoader getter be moved to `getDatabase/createDatabase` ?

liquibase-maven-plugin/src/main/java/org/liquibase/maven/plugins/LiquibaseHistoryMojo.java
```diff
@@ -14,10 +16,16 @@ public class LiquibaseHistoryMojo extends AbstractLiquibaseMojo {
 
     @Override
     protected void performLiquibaseTask(Liquibase liquibase) throws LiquibaseException {
-      CommandScope historyCommand = new CommandScope(InternalHistoryCommandStep.COMMAND_NAME);
-
-      historyCommand.addArgumentValue(InternalHistoryCommandStep.DATABASE_ARG, getLiquibase().getDatabase());
-
-      historyCommand.execute();
+        ClassLoader mavenClassLoader;
```
filipelautert, line 19: Same as above - could this ClassLoader getter be moved to `getDatabase/createDatabase` ? And here we have a try/catch, above we don't.

filipelautert (commented):
Overral the removal os duplicated code + usage of try with resources seems good, just some small comments above.

liquibase-maven-plugin/src/main/java/org/liquibase/maven/plugins/AbstractLiquibaseMojo.java
MalloD12: Yeah, sure. I created this PR yesterday especially to see if there was something broken I could figure out through the functional tests. What you are proposing here makes total sense to me and I'll apply it now. 

Thanks,
Daniel.

liquibase-standard/src/main/java/liquibase/parser/AbstractFormattedChangeLogParser.java
```diff
@@ -314,7 +314,8 @@
                     } else {
                         try {
                             long ignoreCount = Long.parseLong(ignoreLinesMatcher.group(1));
-                            while (ignoreCount > 0 && reader.readLine() != null) {
+                            String newLine;
```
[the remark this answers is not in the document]
filipelautert: Do we need this new line if we re not using it?

liquibase-standard/src/main/java/liquibase/snapshot/jvm/PrimaryKeySnapshotGenerator.java
```diff
@@ -58,7 +58,7 @@ protected DatabaseObject snapshotObject(DatabaseObject example, DatabaseSnapshot
 
                 String ascOrDesc = row.getString("ASC_OR_DESC");
                 Boolean descending = "D".equals(ascOrDesc) ? Boolean.TRUE : "A".equals(ascOrDesc) ? Boolean.FALSE : null;
-                boolean computed = descending != null && descending;
+                boolean computed = descending != null;
```
filipelautert, line 61: This test seems wrong. Descending is being used to hold 3 values here: true, false or null. 
d = true; a = false; else null . Maybe should be changed to an` if ` for clarity

filipelautert (changes requested):

liquibase-standard/src/main/java/liquibase/snapshot/jvm/PrimaryKeySnapshotGenerator.java
MalloD12: Done.

filipelautert (approved):

suryaaki2 (approved):

PR #5043, opened 2023-10-10
Improve support for DB2[z] empty table precondition

MichaelKern-IVV:
## Impact

## Description

Adds statements for DB2 to the TableIsEmptyGenerator


MalloD12 (approved):
This looks ok to me. I manually test it against a DB2 instance by deploying a create table change and then the below changeset:

```
<changeSet id="tableIsEmptyShouldFail" author="flautert">
            <tableIsEmpty tableName="csvdata"/>
        </preConditions>
        <!-- if table is dropped then next changesets will fail, thus it is an indirect check -->
        <dropTable tableName="csvdata" />
    </changeSet>
```

Which dropped the `csvdata` table successfully. See below:
Screenshot 2023-10-13 at 9 03 35 PM


Thanks @MichaelKern-IVV!

filipelautert (approved):

suryaaki2 (approved):

rberezen (approved):

liquibase-standard/src/main/java/liquibase/sqlgenerator/core/TableIsEmptyGenerator.java
```diff
@@ -29,11 +29,16 @@ public ValidationErrors validate(TableIsEmptyStatement tableIsEmptyStatement, Da
...
+        if (database instanceof Db2zDatabase) {
+            return String.format("SELECT COUNT(1) FROM SYSIBM.SYSDUMMY1 WHERE EXISTS (SELECT * FROM %s)", tableName);
+        }
+        if (database instanceof OracleDatabase || database instanceof MySQLDatabase) {
             return String.format("SELECT COUNT(1) FROM DUAL WHERE EXISTS (SELECT * FROM %s)", tableName);
-        } else if (database instanceof FirebirdDatabase) {
+        }
+        if (database instanceof FirebirdDatabase) {
             return String.format("SELECT COUNT(1) FROM RDB$DATABASE WHERE EXISTS (SELECT * FROM %s)", tableName);
         } else if (database instanceof DB2Database) {
```
filipelautert, line 43: Hey @MichaelKern-IVV ! I was working on some test fixes do DB2 and added this part of code to check table is empty (using this dummy table). Which one would be better to use in DB2: values(0) or dummy?

filipelautert (changes requested):

liquibase-standard/src/main/java/liquibase/sqlgenerator/core/TableIsEmptyGenerator.java
MichaelKern-IVV: Hi @filipelautert, I apologize for the late reply.

It's a question of taste. Personally, I find values(0) somewhat more elegant because this solution does not require a dummy table. On the other hand, DB2z only supports the variant with sysibm.sysdummy1.

That is why I have chosen different variants for DB2z and DB2luw, but - as I said - this is a matter of personal taste. I would also be fine with the dummy table solution for both DB2 versions, please adapt the change as you find it more suitable.

filipelautert (approved):

PR #5053, opened 2023-10-11
add option to show hidden arguments in help output

StevenMassaro:
## Impact

## Description

If true, all command arguments marked as hidden will be shown in the help output, ignoring the hidden flag. NOTE, due to the order of value provider loading at such an early point in Liquibase startup, you MUST set this as a environment variable. Command line parameters will not be recognized.


StevenMassaro:
> Changes are fine! I'm just trying to understand what would be the use for that?

We often add hidden properties as we work on features in pro that aren't ready to be released. People will ask me "what is the right formatting to use this parameter in the liquibase.properties file". Without this change, I need to unhide the properties, do a build, copy the help output, then rehide the properties. With this fix, I can just expose the hidden properties quickly without needing to do a whole build. Believe it or not, I find myself in this situation very often.

PR #5073, opened 2023-10-18
Run tests using Java 21

filipelautert:
## Impact

## Description

Profile non-jdk8 was used to upgrade mockito libraries as we still need to compile and test using Java 8.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -153,20 +153,7 @@ jobs:
       matrix:
         os: [ ubuntu-22.04 ]
         java: [ 8, 11, 17 ]
-        exclude:
-          # exclude non-java 8 on macos and windows builds
-          - os: macos-latest
-            java: 11
-          - os: windows-2019
-            java: 11
-          - os: macos-latest
-            java: 17
-          - os: windows-2019
-            java: 17
-          - os: macos-latest
-            java: 21
-          - os: windows-2019
-            java: 21
+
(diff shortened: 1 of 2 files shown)


filipelautert:
Caused by: java.lang.IllegalArgumentException: Java 21 (65) is not supported by the current version of Byte Buddy which officially supports Java 20 (64) - update Byte Buddy or set net.bytebuddy.experimental as a VM property

filipelautert:
Our bytebuddy version is old (1.12.19) because our Mockito version is old (4.10) .  We can try to upgrade bytebuddy by itself and see if it works.
We can't upgrade mockito to 5.x series that uses the new bytebuddy as it requires java 11  - this would break our java 8 build.

StevenMassaro:
> Our bytebuddy version is old (1.12.19) because our Mockito version is old (4.10) . We can try to upgrade bytebuddy by itself and see if it works. We can't upgrade mockito to 5.x series that uses the new bytebuddy as it requires java 11 - this would break our java 8 build.

Why are we still building against Java 8 if it's end of life?

StevenMassaro (approved):

filipelautert:
> Why are we still building against Java 8 if it's end of life?

@StevenMassaro seems people still using Java 8 (even pro customers?) so we want to make sure that we are compatible with that. But I agree that we should drop support for it soon (Liquibase 5?) .

PR #5087, opened 2023-10-19
Handles snapshot of Oracle temporary tables + ignore queue tables

filipelautert:
## Impact

### DIFF
--- a/liquibase-standard/src/main/java/liquibase/snapshot/JdbcDatabaseSnapshot.java
+++ b/liquibase-standard/src/main/java/liquibase/snapshot/JdbcDatabaseSnapshot.java
@@ -1243,7 +1243,8 @@ public class JdbcDatabaseSnapshot extends DatabaseSnapshot {
                             "c.COMMENTS as REMARKS, A.tablespace_name as tablespace_name, CASE WHEN A.tablespace_name = " +
                             "(SELECT DEFAULT_TABLESPACE FROM USER_USERS) THEN 'true' ELSE null END as default_tablespace " +
                             "from ALL_TABLES a " +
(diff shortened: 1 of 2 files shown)


rberezen (approved):

liquibase-standard/src/main/java/liquibase/snapshot/JdbcDatabaseSnapshot.java
```diff
@@ -1243,7 +1243,8 @@ private List<CachedRow> queryOracle(CatalogAndSchema catalogAndSchema, String ta
                             "c.COMMENTS as REMARKS, A.tablespace_name as tablespace_name, CASE WHEN A.tablespace_name = " +
                             "(SELECT DEFAULT_TABLESPACE FROM USER_USERS) THEN 'true' ELSE null END as default_tablespace " +
                             "from ALL_TABLES a " +
-                            "join ALL_TAB_COMMENTS c on a.TABLE_NAME=c.table_name and a.owner=c.owner ";
+                            "join ALL_TAB_COMMENTS c on a.TABLE_NAME=c.table_name and a.owner=c.owner " +
+                            "left outer join ALL_QUEUE_TABLES q ON a.TABLE_NAME = q.QUEUE_TABLE and a.OWNER = q.OWNER ";
```
hel1e, line 1247: You don't actually filter out the queue-tables here as far as I can see. You will need to also add a where clause:
`where q.QUEUE_TABLE is null`

filipelautert: whoops, that's why it was so easy :) Thanks for the review - I'll create another pr with the fix.

PR #5131, opened 2023-10-26
Allows to set older checksum versions for testing purposes

filipelautert:
## Impact


filipelautert:
@StevenMassaro what do you think about this implementation? 
Looking at that now I think that maybe it's a bit too much and I could use only the scope parameter instead of this whole Plugin/Factory stuff.

StevenMassaro:
> @StevenMassaro what do you think about this implementation? Looking at that now I think that maybe it's a bit too much and I could use only the scope parameter instead of this whole Plugin/Factory stuff.

That does make sense. I was previously approaching this from a perspective of "I don't want it to be easy for users to intentionally/unintentionally set their checksum version to an old one". A CLI param would be easy. Scope is equally as difficult as a plugin, I'd say. I'm fine with you switching it.

liquibase-integration-tests/src/test/groovy/liquibase/command/core/UpgradeChecksumVersionIntegrationTest.groovy
```diff
@@ -335,4 +336,54 @@ VALUES('2', 'fl', '$changesetFilepath', '2023-09-29 14:33:39.112', 2, 'EXECUTED'
...
+        when:
+        def ranChangeSets = getRanChangesets(changeLogService)
+
+        then:
+        ranChangeSets.size() == 2
+        ranChangeSets.forEach({ rcs -> assert rcs.getLastCheckSum().getVersion() == 8 })
+
+        when:
+        def scopeSettings = [
+                (Scope.Attr.latestChecksumVersion.name()): ChecksumVersion.V8
```
fbiville, line 366: would be interesting to test it with a version other than 8, since the persisted change sets already are at version 8.

liquibase-integration-tests/src/test/groovy/liquibase/command/core/UpgradeChecksumVersionIntegrationTest.groovy
```diff
@@ -335,4 +336,54 @@ VALUES('2', 'fl', '$changesetFilepath', '2023-09-29 14:33:39.112', 2, 'EXECUTED'
...
     }
+
+
+    @Unroll
+    def "Calculate change using v8 forced calculator" () {
+        def changesetFilepath = "changelogs/h2/checksum/dbms-filter-changelog.xml"
+        final ChangeLogHistoryService changeLogService = Scope.getCurrentScope().getSingleton(ChangeLogHistoryServiceFactory.class).getChangeLogService(h2.getDatabaseFromFactory())
+        changeLogService.init()
+        h2.execute(new RawSqlStatement("""
+INSERT INTO DATABASECHANGELOG
```
fbiville, line 347: could you manually instantiate change sets and insert them through ChangeLogHistoryService, within the scope of an older checksum version instead of running raw SQL?

filipelautert: I added the` manually generate v7 checksum` test below this one, what do you think?

fbiville (approved):

filipelautert:
@rberezen now it's ready for review.

rberezen (approved):

PR #5134, opened 2023-10-27
Makes sure that precondition onErrorMessage and onFailMessage are logged when an error happens

filipelautert:
## Impact

## Description

Currently onErrorMessage and onFailMessage are logged only if the log-level property is set to something higher than info. As they are custom error messages they should be logged everytime an error occurs.

### DIFF
--- a/liquibase-standard/src/main/java/liquibase/changelog/visitor/ValidatingVisitor.java
+++ b/liquibase-standard/src/main/java/liquibase/changelog/visitor/ValidatingVisitor.java
@@ -14,12 +14,16 @@ import liquibase.precondition.FailedPrecondition;
 import liquibase.precondition.core.PreconditionContainer;
 import liquibase.util.StringUtil;
 import liquibase.util.ValidatingVisitorUtil;
+import lombok.Getter;
 
 import java.util.*;
 
+@Getter
 public class ValidatingVisitor implements ChangeSetVisitor {
 
     private final List<String> invalidMD5Sums = new ArrayList<>();
+    private String failedPreconditionsMessage = null;
+    private String errorPreconditionsMessage = null;
     private final List<FailedPrecondition> failedPreconditions = new ArrayList<>();
     private final List<ErrorPrecondition> errorPreconditions = new ArrayList<>();
     private final Set<ChangeSet> duplicateChangeSets = new HashSet<>();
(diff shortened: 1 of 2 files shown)


rberezen (approved):

rberezen:
@tati-qalified could you validate the fix please? TIA!

tati-qalified:
**It's working for me, with no issues.**
I tested this using the CLI and a simple changelog:

```
databaseChangeLog:
  - preConditions:
    - onError: HALT
    - onFail: HALT
    - onErrorMessage: ---- this is an error ----
    - onFailMessage: ---- this is a failure ----
    - sqlCheck:
          expectedResult: 1
          sql: SELECT 1 from myTable
  -  changeSet:  
      id:  1  
      author:  nvoxland  
      preConditions:
        - onError: HALT
        - onFail: HALT
        - onErrorMessage:  ---- this is an error ----
        - onFailMessage: ---- this is a failure ----
        - foreignKeyConstraintExists:
            foreignKeyName: columna12345
            foreignKeyTableName: json_table
      changes:  
        -  createTable:  
            tableName:  newtable  
            columns:  
              -  column:  
                  name:  id  
                  type:  int  
                  autoIncrement:  true  
                  constraints:  
                    primaryKey:  true  
                    nullable:  false  
              -  column:  
                  name:  firstname  
                  type:  varchar(50)  
              -  column:  
                  name:  lastname  
                  type:  varchar(50)  
                  constraints:  
                    nullable:  false  
              -  column:  
                  name:  state  
                  type:  char(2)  
```

- First I tested only the preconditions for the entire changelog. I commented out the preconditions inside the changeset. 
- After that, I uncommented the changeset preconditions and commented the changelog preconditions.
- Then I used both the changelog and changeset preconditions.

For all three scenarios I followed these steps:
1. Use a precondition that will generate an error
2. Run `liquibase update`
3. Run `liquibase update --log-level= ` with each log level value
4. Do the same with a precondition that will fail
5. Run `liquibase update` with a precondition that will work - no errors nor failures

I tested around 6 different preconditions to be sure it worked in general. I don't believe it's necessary to test each one, though I can do it if someone thinks I should.

tati-qalified:
Apart from that, there's an error message that isn't too user-friendly:

When trying to run a changelog with an empty sqlCheck
```
databaseChangeLog:
  - preConditions:
    - onError: HALT
    - onFail: HALT
    - onErrorMessage: ---- this is an error ----
    - onFailMessage: ---- this is a failure ----
    - sqlCheck:
```
The error message is:
```
Unexpected error running Liquibase: Migration failed for changeset changelog.yml::1::nvoxland:
     Reason: java.lang.NullPointerException: Cannot invoke "String.replaceFirst(String, String)" because the return value of "liquibase.precondition.core.SqlPrecondition.getSql()" is null
```

It should probably say something along the lines of "sqlCheck attributes sql and expectedResult are required"

PR #5141, opened 2023-10-30
Fix Liquibase changeset relative path check

filipelautert:
## Impact

## Description

Liquibase path handling has changed over time leading to valid paths not being accepted thus breaking changesets. This fix aims to check for all possible path variations that Liquibase used over time.

### DIFF
--- a/liquibase-standard/src/main/java/liquibase/changelog/RanChangeSet.java
+++ b/liquibase-standard/src/main/java/liquibase/changelog/RanChangeSet.java
@@ -4,8 +4,8 @@ import liquibase.ChecksumVersion;
 import liquibase.ContextExpression;
 import liquibase.Labels;
 import liquibase.change.CheckSum;
-import liquibase.integration.commandline.LiquibaseCommandLineConfiguration;
 
+import java.nio.file.Paths;
 import java.util.Date;
 
 /**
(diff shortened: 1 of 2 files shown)


tati-qalified:
I've tested this and I'm getting what I **think** is the expected result. Do let me know if I'm assuming anything wrong.

**Steps taken ([here are the files for reference](https://github.com/liquibase/liquibase/files/13227768/pr5141_test.zip)):**
Using a directory structure like this:
```
folder1
|
+-- folder2
|     |
|     +-- change2.xml
|
+-- folder3
|     |
|     +-- change3.xml
|
+-- change1.xml
```

With changesets set up so that:
- change1 links to change2
- change2 links to change3

with relativeToChangeLogFile=true

For each of the Liquibase versions 4.19.0 through to 4.24.0, I repeated these steps:
1. In `change3.xml`, edited the changeset Id and table name so they're unique
2. Set the changeLogFile to `change1.xml`
3. Ran `liquibase update`


Then I used the PR code and ran these steps:
1.  In `change3.xml`, edited the changeset Id and table name so they match one of the previous runs
2. Ran `liquibase update`
3. Verify that the changeset is **not** re-run



**Results:**
When using the PR code, Liquibase correctly recognises a changeset as already executed, no matter if the file path is `folder1/folder2/../folder3/change3.xml` or `folder1/folder3/change3.xml`.

**Here's what might or might not be correct:**
If, using this same file and directory structure, we point `change1.xml` directly to `change3.xml`, still using relativeToChangeLogFile=true, it's not recognised as the same, even though it's technically the same path.

tati-qalified (approved):

rberezen (approved):

PR #5165, opened 2023-11-06
Do not warn when env var matches argument alias (DAT-16117)

abrackx:
## Impact

## Description
Do not warn on unrecognized LIQUIBASE_ environment variables when the environment variable matches an alias.


abrackx:
> Are there any tests for this code? Could you add to them or create one for this case?
> 
> Also, @wwillard7800 , didn't you make a ticket for this recently? I could be imagining things.

There's none for this particular case since it's integrating the configuration with the env var provider. But yeah, the test cases in this area are very light. I'll see what I can do.

filipelautert (approved):

wwillard7800 (approved):

PR #5182, opened 2023-11-09
Invoke database.commit only if needed

filipelautert:
## Impact

## Description
Invoke database.commit only if something was executed

When running over large changesets , every commit means a request and for cloud databases it means a new request to the cloud that can take up some time.


suryaaki2 (approved):

suryaaki2:
@filipelautert Can we write some tests to test this scenario?

Tylorjg (approved):
I was able to test the build and it was able to deploy a single changeset on a DBCL (that has 1000) and it did not cause hundreds of commits being sent to the database, and the runtime was milliseconds from the end of the query of the changelog to deploying the changeset.

Fix worked in this scenario.

Issue #4464, opened 2023-07-04
PostgreSQL generated columns get incorrect syntax during diffChangeLog operation


LonwoLonwo:
### Search first

- [X] I searched and no similar issues were found

### Description

This is a continuation of this ticket: https://github.com/liquibase/liquibase/issues/3494

Now in the latest version, table comparison with DDL like:

```sql
CREATE TABLE people (
    height_cm numeric,
    height_in numeric GENERATED ALWAYS AS (height_cm / 2.54) STORED
);
```

works perfectly.

but this table with DDL

```sql
CREATE TABLE users (
	fullname text NOT NULL GENERATED ALWAYS AS ((surname || ', '::text) || forename) STORED,
	surname text NOT NULL,
	forename text NOT NULL,
	CONSTRAINT pk_users_2 PRIMARY KEY (fullname)
);
```

will be transformed during the diffChangeLog command to

```sql
CREATE TABLE users (
"fullname" TEXT 'GENERATED ALWAYS AS GENERATED ALWAYS AS ((surname || '''', ''''::text) || forename) STORED STORED' NOT NULL,
"surname" TEXT NOT NULL,
"forename" TEXT NOT NULL,
CONSTRAINT "pk_users_2" PRIMARY KEY ("fullname")
);
```

And this syntax is incorrect. Probably, the connection here is with the column type TEXT, not numeric, as in the "people" table.

### Steps To Reproduce

- Create two empty schemas.
- Create a table in the first schema:

```sql
CREATE TABLE users (
	fullname text NOT NULL GENERATED ALWAYS AS ((surname || ', '::text) || forename) STORED,
	surname text NOT NULL,
	forename text NOT NULL,
	CONSTRAINT pk_users_2 PRIMARY KEY (fullname)
);
```

- Compare two schemas.

### Expected/Desired Behavior

I want to see the original table syntax for generated columns in the comparison result.

### Liquibase Version

4.23

### Database Vendor & Version

PostgreSQL 12 and higher

### Are you willing to submit a PR?

- [ ] I'm willing to submit a PR (Thank you!)

tati-qalified:
Hello @LonwoLonwo , I was able to replicate this issue in Liquibase version 4.23, and I also verified that this is not a problem in version 4.21 (which is the previous version, as 4.22 was deprecated).
I will be forwarding this to the development team. Thanks for letting us know!

LonwoLonwo:
Works well on the latest version!

PR #5202, opened 2023-11-13
4464 fix generated default column value using in PostgreSQL and Oracle for char/clob data types

LonwoLonwo:
## Impact

`TypeBug`            - bug fixes

## Description

So, I'm trying to fix the case from https://github.com/liquibase/liquibase/issues/4464

The first part of this PR - is avoiding duplication of the `GENERATED ALWAYS AS` part.
And second - an attempt to not add quotes around the function `GENERATED ALWAYS AS` for character/clob columns.
I understand that this is a terrible solution - trying to guess by the string beginning - if this default value is a function or not.
But I do not see other good options.
I tried to add to the ClobType#objectToSql, but it is not general enough.
So, I am open for discussion.

## Things to be aware of

I'm not sure about the test part, also.

## Things to worry about

Can users use the "GENERATED ALWAYS AS" syntax just for default values? I really do not know.


liquibase-standard/src/main/java/liquibase/snapshot/jvm/ColumnSnapshotGenerator.java
```diff
@@ -39,6 +39,7 @@ public class ColumnSnapshotGenerator extends JdbcSnapshotGenerator {
     private static final Pattern POSTGRES_NUMBER_VALUE_PATTERN = Pattern.compile(POSTGRES_NUMBER_VALUE_REGEX);
 
     private static final String MYSQL_DEFAULT_GENERATED = "DEFAULT_GENERATED";
+    private static final String POSTGRES_GENERATED_ALWAYS = "GENERATED ALWAYS AS ";
```
MalloD12, line 42: If I'm not wrong this is also supported also by Oracle. I think maybe we can rename this constant and use it for both cases. Have you tried this in Oracle as well?

LonwoLonwo: So, PostgreSQL and Oracle cases are close, but I see some differences.

1. Different syntax. For PostgreSQL, it will be `GENERATED ALWAYS AS (value) STORED`. So, we have this STORED at the end.
2. It Looks like Oracle returns us the default value without "()", but PostgreSQL returns with:

2023-12-22 19_43_20-DBeaver  C__Users_anvolka_work_DbeaverP_dbeaver-workspace-idea  - C__Users_anvol

2023-12-22 19_39_18-DBeaver  C__Users_anvolka_work_DbeaverP_dbeaver-workspace-idea  - C__Users_anvol

3. We have different conditions for column flags. 
It is `"YES".equals(columnMetadataResultSet.get("VIRTUAL_COLUMN")` for Oracle.
And `"YES".equals(columnMetadataResultSet.get("IS_GENERATEDCOLUMN"))` for PostgreSQL.

So, I can add all these conditions as flags in one merged method. But I am unsure.

MalloD12: I think we shouldn't merge conditions, I would leave it as it's now.

liquibase-standard/src/main/java/liquibase/snapshot/jvm/ColumnSnapshotGenerator.java
```diff
@@ -515,14 +519,14 @@ protected Object readDefaultValue(CachedRow columnMetadataResultSet, Column colu
...
-                columnMetadataResultSet.set(COLUMN_DEF_COL, new DatabaseFunction("null"));
+                columnMetadataResultSet.set(COLUMN_DEF_COL, new DatabaseFunction(NULL_VALUE));
             }
         }
 
         if ((database instanceof OracleDatabase) && (columnMetadataResultSet.get(COLUMN_DEF_COL) == null)) {
             columnMetadataResultSet.set(COLUMN_DEF_COL, columnMetadataResultSet.get("DATA_DEFAULT"));
 
-            if ((columnMetadataResultSet.get(COLUMN_DEF_COL) != null) && "NULL".equalsIgnoreCase((String)
+            if ((columnMetadataResultSet.get(COLUMN_DEF_COL) != null) && NULL_VALUE.equalsIgnoreCase((String)
```
MalloD12, line 529: I think it would look a bit better if we replace this `NULL_VALUE` check with something like:
```
StringUtil.equalsWordNull((String)
                    columnMetadataResultSet.get(COLUMN_DEF_COL))
```

liquibase-standard/src/main/java/liquibase/util/SqlUtil.java
```diff
@@ -114,6 +112,12 @@ public static Object parseValue(Database database, Object val, DataType type) {
             }
         }
 
+        if ((database instanceof PostgresDatabase) && (liquibaseDataType instanceof CharType || liquibaseDataType instanceof ClobType)
```
MalloD12, line 115: I think we can add Oracle to the database check of this condition:

```(database instanceof PostgresDatabase || database instanceof OracleDatabase)```  &&...

File: liquibase-integration-tests/src/test/java/liquibase/dbtest/pgsql/PostgreSQLIntegrationTest.java [hunk 4]
```diff
@@ -268,6 +268,27 @@ public void testGeneratedColumn() throws Exception {
 
     }
 
+    @Test
+    public void testGeneratedClobColumn() throws Exception {
+        assumeNotNull(getDatabase());
```
MalloD12, line 273: I think we can probably add a similar test to `OracleIntegrationTest` considering the differences you stated in your differences findings comment.

LonwoLonwo: Ok, I will try to do it.

liquibase-standard/src/main/java/liquibase/snapshot/jvm/ColumnSnapshotGenerator.java
LonwoLonwo: O! I see. I will fix it.

liquibase-standard/src/main/java/liquibase/util/SqlUtil.java
LonwoLonwo: Ok, for the future. But, for now, we will never be here because of the conditions above:

```java
else if (!stringVal.matches("\\d+\\.?\\d*")) {
                //not just a number
                return new DatabaseFunction(stringVal);
```

liquibase-integration-tests/src/test/java/liquibase/dbtest/pgsql/PostgreSQLIntegrationTest.java
LonwoLonwo: Ok, I've added a new test for Oracle. But, unfortunately, I can't build the test part, so I can't test my test :(
Please take a look at it.

liquibase-standard/src/main/java/liquibase/snapshot/jvm/ColumnSnapshotGenerator.java
LonwoLonwo: Added

liquibase-standard/src/main/java/liquibase/util/SqlUtil.java
LonwoLonwo: Added

liquibase-integration-tests/src/test/java/liquibase/dbtest/pgsql/PostgreSQLIntegrationTest.java
MalloD12: I can't test that locally either, but let's see how it goes in GH and then we can see whether it requires any change or not. 

Thanks,
Daniel.

MalloD12: It seems there is something wrong with the generated SQL executed. I'll ask someone's help who can help me with oracle to review this and I'll let you know.

LonwoLonwo: Thank you!

liquibase-integration-tests/src/test/java/liquibase/dbtest/oracle/OracleIntegrationTest.java
```diff
@@ -225,7 +225,7 @@ public void testChangeLogGenerationForTableWithGeneratedColumn() throws Exceptio
         String textToTest = "GENERATED ALWAYS AS (QTY * PRICE)";
 
         Scope.getCurrentScope().getSingleton(ExecutorService.class).getExecutor("jdbc", getDatabase()).execute(new RawSqlStatement(
-            String.format("CREATE TABLE GENERATED_COLUMN_TEST(QTY INT, PRICE INT, TOTALVALUE INT %s);", textToTest)));
```
LonwoLonwo: Oops.

filipelautert: Almost there :)
The assertion failed as it is not able to find our sql fragment. The generated SQL below is:
```
<column defaultValueComputed="GENERATED ALWAYS AS (&quot;QTY&quot;*&quot;PRICE&quot;)" name="TOTALVALUE" type="NUMBER(*, 0)"/>
```
So in my next commit I'm removing `"` from virtColumnDef variable.. and seems it works. I wonder if this is the best solution @LonwoLonwo  @MalloD12 , but at least it is working now :)

MalloD12 (approved):
**Approved**.

Thanks @LonwoLonwo for this PR, for applying the required changes, and for your patience. Code changes look good to me. Build and all tests have been successfully executed.

It would be good if we can test running the Oracle test added in some different version than the one use for `OracleIntegrationTest` suite (`oracle-xe:18-slim`). @rberezen could you please do that when you review this PR? This test I'm talking about was failing on the mentioned version, but using the generated SQL and executing it in a newer version was successfully executed, so I would like to make sure that now it pass on version 18 it also works in newer versions.

filipelautert (approved):

rberezen (approved):

PR #5221, opened 2023-11-15
Added some expression evaluation dependencies  DAT-16164

wwillard7800:
## Impact


filipelautert (approved):

StevenMassaro (changes requested):
Since this dependency is only used in Pro right now, can you shade the dependency into the pro jar and not add it as a separate jar in the dist tar.gz?

wwillard7800:
> Since this dependency is only used in Pro right now, can you shade the dependency into the pro jar and not add it as a separate jar in the dist tar.gz?

I bet I can.  I was just being expedient.

wwillard7800:
Now shading MVEL into the Pro .jar

PR #5223, opened 2023-11-15
Improve empty check style by using isEmpty String method

### LINKED ISSUE #3718: Code Cleanup - Improve empty check style  by using isEmpty String method
## Description

Improve empty string check by making use of isEmpty string method to make code more readable. Below is an example:

`!"".equals(clause.toString().trim())`

## Steps To Reproduce

This is a type enhancement issue, **NO** steps to reproduce here.

## Actual Behavior
Having comparisons like the one below in the code:
`!"".equals(clause.toString().trim())`

## Expected/Desired Behavior
Do string empty checks by doing something like `clause.toString().trim().isEmpty()`

- None

## Additional Context
- None

MalloD12:
- Replaced literal empty string comparison by using isEmpty string method.
- Test update.

Fixes #3718 

## Impact


liquibase-standard/src/main/java/liquibase/serializer/core/xml/XMLChangeLogSerializer.java
```diff
@@ -127,15 +127,15 @@ public <T extends ChangeLogChild> void write(List<T> children, OutputStream out)
         }
 
         for (Map.Entry<String, String> entry : shortNameByNamespace.entrySet()) {
-            if (!"".equals(entry.getValue())) {
+            if (!entry.getValue().isEmpty()) {
```
filipelautert, line 130: It's unlikely but it is possible to have a null value in here.

liquibase-standard/src/main/java/liquibase/serializer/core/xml/XMLChangeLogSerializer.java
```diff
@@ -127,15 +127,15 @@ public <T extends ChangeLogChild> void write(List<T> children, OutputStream out)
...
                 changeLogElement.setAttribute("xmlns:" + entry.getValue(), entry.getKey());
             }
         }
 
 
         StringBuilder schemaLocationAttribute = new StringBuilder();
         for (Map.Entry<String, String> entry : urlByNamespace.entrySet()) {
-            if (!"".equals(entry.getValue())) {
+            if (!entry.getValue().isEmpty()) {
```
filipelautert, line 138: Same, so StringUtil.isNotEmpty should be safer.

liquibase-standard/src/main/java/liquibase/util/StandardSqlParser.java
```diff
@@ -20,7 +20,7 @@ public StringClauses parse(String sqlBlock, boolean preserveWhitespace, boolean
         SimpleSqlGrammar t = new SimpleSqlGrammar(tokenManager);
         try {
             Token token = t.getNextToken();
-            while (!"".equals(token.toString())) {
+            while (!token.toString().isEmpty()) {
```
filipelautert, line 23: I think we can have a null value here.

liquibase-standard/src/test/groovy/liquibase/change/core/LoadDataChangeTest.groovy
```diff
@@ -706,6 +706,9 @@ public class LoadDataChangeTest extends StandardChangeTest {
         assert dCol2 instanceof Timestamp
         ((Timestamp) dCol2).toLocalDateTime() == LocalDateTime.parse("2020-01-02T03:04:05")
         columnValue(sqlStatements[2], Col.bool) == Boolean.TRUE
+        for (Col n : [Col.name, Col.id, Col.date, Col.bool]) {
```
filipelautert, line 709: new tests?

filipelautert (changes requested):

liquibase-standard/src/main/java/liquibase/change/core/LoadDataChange.java
```diff
@@ -350,21 +350,17 @@ public SqlStatement[] generateStatements(Database database) {
                         } else if (columnConfig.getType().equalsIgnoreCase("date")
                                 || columnConfig.getType().equalsIgnoreCase("datetime")
                                 || columnConfig.getType().equalsIgnoreCase("time")) {
-                            if ("NULL".equalsIgnoreCase(value) || "".equals(value)) {
```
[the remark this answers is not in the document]
MalloD12: Initially, the purpose of updating this line was just to replace`"".equals(value)` with `value.isEmpty()` doing this I saw there were some tests that were failing because this was, don't remember exactly if it was when having a `null` or when having an empty string (`''`), so I changed the logic this way, but I think I missed the case of when the value (date) is `null` then set value as null (as it was before). I'm going to update this.

liquibase-snowflake/src/main/java/liquibase/sqlgenerator/core/InsertOrUpdateGeneratorSnowflake.java
```diff
@@ -63,7 +63,7 @@ protected String getElse(Database database) {
     // Copied from liquibase.sqlgenerator.core.InsertOrUpdateGeneratorHsql
     private String convertToString(Object newValue, Database database) {
         String sqlString;
-        if ((newValue == null) || "".equals(newValue.toString()) || "NULL".equalsIgnoreCase(newValue.toString())) {
+        if ((newValue == null) || newValue.toString().isEmpty() || "NULL".equalsIgnoreCase(newValue.toString())) {
```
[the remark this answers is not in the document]
MalloD12: Yeap, I didn't look for more of these occurrences because there are a bunch of them in the project and I was trying to limit the scope to just update the empty string check. I can update this one and also some other occurrences in the code if you think we should address that change in this PR.

liquibase-standard/src/main/java/liquibase/change/core/OutputChange.java
```diff
@@ -91,7 +91,7 @@ public String getSerializedObjectNamespace() {
     @Override
     public Object getSerializableFieldValue(String field) {
         Object value = super.getSerializableFieldValue(field);
-        if ("target".equals(field) && "".equals(value)) {
+        if ("target".equals(field) && value.toString().isEmpty()) {
```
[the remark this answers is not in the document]
MalloD12: You are right. I'll update this one.

liquibase-standard/src/test/groovy/liquibase/change/core/LoadDataChangeTest.groovy
MalloD12: Do you mean adding a test cover for this case separately? I just updated that test that it's verifying some different test data cases, and as I added a new input line I added this check. I can take it to a separate case, although I think the ideal would be to refactor this test case.

filipelautert: No, that's fine!

liquibase-standard/src/main/java/liquibase/util/StandardSqlParser.java
MalloD12: I think in this particular case we should check `token != null && !token.toString().isEmpty()`, because if we use `StringUitl.isNotEmpty(token)` we would allow null and then when processing the token get an NPE. Do you agree?

liquibase-standard/src/main/java/liquibase/change/core/LoadDataChange.java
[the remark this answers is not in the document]
filipelautert: I see.. do we have tests covering that?

liquibase-standard/src/main/java/liquibase/util/StandardSqlParser.java
filipelautert: Makes sense!

filipelautert (changes requested):
almost there!

liquibase-standard/src/main/java/liquibase/change/core/LoadDataChange.java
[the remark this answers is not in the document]
MalloD12: I updated/refactored some tests that are testing the updated code.

liquibase-standard/src/main/java/liquibase/change/core/ExecuteShellCommandChange.java
```diff
@@ -350,7 +350,7 @@
         } else {
             List<String> os = StringUtil.splitAndTrim(StringUtil.trimToEmpty(parsedNode.getChildValue(null, "os",
                     String.class)), ",");
-            if ((os.size() == 1) && ("".equals(os.get(0)))) {
+            if ((os.size() == 1) && (StringUtil.isEmpty(os.get(0)))) {
```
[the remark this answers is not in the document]
filipelautert: Could we use StringUtils here?

liquibase-standard/src/main/java/liquibase/change/core/LoadDataChange.java
```diff
@@ -314,7 +314,7 @@
                         valueConfig.setName(columnName);
                         valueConfig.setAllowUpdate(columnConfig.getAllowUpdate());
 
-                        if (value.isEmpty()) {
+                        if (StringUtil.isEmpty(value)) {
```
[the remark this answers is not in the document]
filipelautert: Could we use StringUtils in this entire PR?

filipelautert (approved):

tati-qalified (changes requested):
Some `StringUtil` uses are leftover, I'll list them to make things easier:

1. `liquibase-snowflake/src/main/java/liquibase/sqlgenerator/core/InsertOrUpdateGeneratorSnowflake.java` lines 8 and 67
2. `liquibase-standard/src/main/java/liquibase/change/core/LoadDataChange.java` lines 91, 344 (first usage), 376, 391, 398
3. `liquibase-standard/src/main/java/liquibase/sqlgenerator/core/InsertOrUpdateGeneratorHsql.java` line 82 (second usage)

MalloD12:
> Some `StringUtil` uses are leftover, I'll list them to make things easier:
> 
> 1. `liquibase-snowflake/src/main/java/liquibase/sqlgenerator/core/InsertOrUpdateGeneratorSnowflake.java` lines 8 and 67
> 2. `liquibase-standard/src/main/java/liquibase/change/core/LoadDataChange.java` lines 91, 344 (first usage), 376, 391, 398
> 3. `liquibase-standard/src/main/java/liquibase/sqlgenerator/core/InsertOrUpdateGeneratorHsql.java` line 82 (second usage)

Hey @tati-qalified,

Those uses haven't been replaced by `StringUtils` because this `commons-lang` library doesn't have a similar method. we might need to implement a new version of this method, but that wasn't part of the scope of this PR. I replaced all the ones Code-Scan was "complaining" about it. 

Thanks,
Daniel.

tati-qalified (approved):

rberezen (approved):

PR #5224, opened 2023-11-15
changes in support of recording dropAll in history table (DAT-16148)

StevenMassaro:
## Impact


liquibase-standard/src/main/java/liquibase/database/Database.java
```diff
@@ -182,6 +196,103 @@ default String getDisplayName() {
...
+                        for (SqlStatement statement : sqlStatements) {
+                            AtomicReference<Sql[]> sqls = new AtomicReference<>(null);
+                            Scope.child(Collections.singletonMap(SqlGeneratorFactory.GENERATED_SQL_ARRAY_SCOPE_KEY, sqls), () -> {
+                                Scope.getCurrentScope().getSingleton(ExecutorService.class).getExecutor("jdbc", this).execute(statement);
+                            });
+                            if (StringUtil.isNotEmpty(completeSql.toString()) && !completeSql.toString().endsWith("; ")) {
+                                completeSql.append("; ");
+                            }
+                            completeSql.append(StringUtil.join(Arrays.stream(sqls.get()).map(Sql::toSql).collect(Collectors.toList()), "; "));
+                        }
```
StevenMassaro, line 265: This is one notable change compared to the same method in `AbstractJdbcDatabase`

liquibase-standard/src/main/java/liquibase/database/Database.java
```diff
@@ -182,6 +196,103 @@ default String getDisplayName() {
...
+            this.commit();
+        }
+    }
+
+    default void addCompleteSqlToScope(StringBuilder completeSql) {
+        AtomicReference<String> sqlsReference = Scope.getCurrentScope().get(COMPLETE_SQL_SCOPE_KEY, AtomicReference.class);
+        if (sqlsReference != null) {
+            sqlsReference.set(completeSql.toString());
+        }
+    }
```
StevenMassaro, line 294: Another notable difference

liquibase-standard/src/main/java/liquibase/command/core/DropAllCommandStep.java
```diff
@@ -83,6 +89,10 @@ public void run(CommandResultsBuilder resultsBuilder) throws Exception {
         resultsBuilder.addResult("statusCode", 0);
     }
 
+    public SnapshotControl getSnapshotControl(CommandScope commandScope, Database database) {
```
filipelautert, line 92: Could we add some javadocs here explaining why we have an empty method? Otherwise I fell someone could try to remove it in the future.

liquibase-standard/src/main/java/liquibase/database/Database.java
```diff
@@ -182,6 +196,103 @@ default String getDisplayName() {
      */
     void dropDatabaseObjects(CatalogAndSchema schema) throws LiquibaseException;
 
+    default void dropDatabaseObjects(CatalogAndSchema schemaToDrop, SnapshotControl snapshotControl) throws LiquibaseException {
+        ObjectQuotingStrategy currentStrategy = this.getObjectQuotingStrategy();
+        this.setObjectQuotingStrategy(ObjectQuotingStrategy.QUOTE_ALL_OBJECTS);
+        try {
+            DatabaseSnapshot snapshot;
+            try {
+                final Set<Class<? extends DatabaseObject>> typesToInclude = snapshotControl.getTypesToInclude();
```
filipelautert, line 205: I see that you added the "null check" for snapshotControl on AbstractJdbcDatabase (creating a new one). But as this method may be called by other pieces of the code now, I think you could move the null check and the new SnapshotControl here.

filipelautert (approved):
Just added 2 small comments for changes.
I remember that 2 or 3 release ago we had a problem where DBCL table wasn't dropped for some extensions. So I think when QA test it they need to test at last mongodb and maybe snowflake to make sure that all tables are still being dropped.  (I'll add thsi comment to jira)

liquibase-standard/src/main/java/liquibase/database/LiquibaseTableNamesFactory.java
```diff
@@ -28,6 +28,7 @@ public List<String> getLiquibaseTableNames(Database database) {
     public void destroy(Database abstractJdbcDatabase) throws DatabaseException {
         for (LiquibaseTableNames generator : generators) {
             generator.destroy(abstractJdbcDatabase);
+            abstractJdbcDatabase.commit();
```
StevenMassaro, line 31: @filipelautert How do you feel about this change? It was necessary to fix a problem I was seeing where the transaction to destroy the history table was abandoned and never committed, but I also know that we fixed an issue recently where we had too many commits being sent and I don't want to add to it if it's the wrong solution.

StevenMassaro: Interestingly, it seems to be a problem isolated only to Postgres, but I don't know why. I seem to remember reading years ago that Postgres does something different regarding commits, or autocommit, or something like that. But I can't find it and I can't remember it.

filipelautert: Correct me if I'm wrong, but this is only called during `dropDatabaseObjects `and it's outside of a loop - in this case is fine.
Previous problem happened because we were doing a commit for every changeset in the changelog file that we read.

And maybe you problem may be related to connection flush or something like that? But anyway 1 commit at the end seems fine .

filipelautert (approved):

PR #5267, opened 2023-11-21
fix #5266: comma separated dbms-attribute for createProcedure-change triggers validation-check

### LINKED ISSUE #5266: createProcedure with multiple comma separated dbms triggers validation-check
### Search first

### Description

According to https://docs.liquibase.com/change-types/create-procedure.html the dbms-attribute in of a createProcedure Change Type should allow comma separated values. It seems, that this is not implemented as such.

In version 4.7 this was working as expected.

### Steps To Reproduce

Create a changeset with a createProcedure-change, which contains dmbs="mysql,mariadb".

### Expected/Desired Behavior

The changeset should not trigger a validation-check.

### Liquibase Version

tested at 4.25.0, but probably all versions 4.8 - 4.25

### Database Vendor & Version

mysql,mariadb

### Liquibase Integration

spring boot

### Liquibase Extensions

_No response_

ubuntu

### Additional Context

I had a look at the code.

`CreateProcedureChange.validate(database)` calls `DatabaseList.validateDefinitions(String definition, vErrors)` with the string provided in the dbms-attribute. It seams, that the valdiateDefinitions-method is supposed to be called with an already split dbms-string.

I want to fix this issue by moving the implementation of `DatabaseList.validateDefinitions(String definition, vErrors)` in a private method `DatabaseList.validateDefinition(String definition, vErrors)` (singular). The public `DatabaseList.validateDefinitions` would then call `toDbmsSet(String definition)` to split the given definition. Then I can call `DatabaseList.validateDefinitions(Collection<String> definitions, vErrors)`, which in turn calls the private `DatabaseList.validateDefinition(String definition, vErrors)``.

I have prepared a fix and will create a pull request.

jclohmann:
## Impact

## Description

As described in the issue #5266, I split the definition string before calling the existing validation-code.

Fixes #5266


liquibase-standard/src/test/java/liquibase/database/DatabaseListTest.java
```diff
@@ -43,4 +50,25 @@ public void databaseMatchesDbmsDefinition() {
         assertFalse(DatabaseList.definitionMatches("!h2,mysql", new H2Database(), false));
         assertTrue(DatabaseList.definitionMatches("!h2,mysql", new MySQLDatabase(), false));
     }
+
```
MalloD12, line 53: Do you mind instead of adding this test here we create a new `DatabaseListTest` class into the database directory from the Groovy sources? Unless it's necessary or because of time constraints we tried to add tests in Groovy format, so if you want to migrate this one/both tests there that would be nice. Otherwise, I can do it.


In case you are not familiar with how to parameterize a test in Spock here's an [example](https://github.com/liquibase/liquibase/blob/4edcddc2d735cc08738a9781aeccfefde9627ee9/liquibase-standard/src/test/groovy/liquibase/changelog/DatabaseChangeLogTest.groovy#L787).

MalloD12 (approved):
**Approved**. 

Dbms attribute has been fixed for stored logic change objects. This is not only applicable for `CreateProcedure`, it also affects `CreateTrigger`, `CreateFunction`, etc. I tested it with CreateProcedure and CreateTrigger and both are working as expected now.

### Things to be aware of:
- None

### Things to worry about:
- None

liquibase-standard/src/test/java/liquibase/database/DatabaseListTest.java
jclohmann: Done. Are you okay with the test names?

filipelautert (approved):

PR #5268, opened 2023-11-21
Rework update summary to support reporting, add new update report fields (DAT-16207)

abrackx:
## Impact

## Description
Reworks ShowSummaryUtil to provide an actual object that captures the show summary output so we can use it later. Add additional fields to UpdateReportParameters. 

## Things to be aware of
Not sure if my approach to the ShowSummaryUtil is the best but it feels like the most straightforward without mucking around in the scope or the mdc.

### DIFF
--- a/liquibase-standard/src/main/java/liquibase/util/ShowSummaryUtil.java
+++ b/liquibase-standard/src/main/java/liquibase/util/ShowSummaryUtil.java
@@ -1,14 +1,14 @@
 package liquibase.util;
 
-import liquibase.GlobalConfiguration;
-import liquibase.Scope;
-import liquibase.UpdateSummaryEnum;
-import liquibase.UpdateSummaryOutputEnum;
+import liquibase.*;
 import liquibase.changelog.ChangeLogIterator;
 import liquibase.changelog.ChangeSet;
 import liquibase.changelog.ChangeSetStatus;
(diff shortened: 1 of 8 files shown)


liquibase-standard/src/main/java/liquibase/util/ShowSummaryUtil.java
```diff
@@ -361,4 +381,14 @@ private static void writeToLog(String message) {
                 .filter(s -> !StringUtil.isWhitespace(s))
                 .forEach(Scope.getCurrentScope().getLog(ShowSummaryUtil.class)::info);
     }
+
+    /**
+     * Container to handle sharing update summary message between different services
+     */
+    @Data
+    @Beta
+    public static class UpdateSummaryDetails {
```
abrackx, line 390: Let me know if you think this approach is legitimate, I'll throw this class into it's own file and clean things up a bit more if so.

StevenMassaro: Seems good to me

StevenMassaro (approved):

filipelautert (approved):

wwillard7800 (approved):
Looks good to me.

suryaaki2 (approved):

PR #5293, opened 2023-11-28
Disable Derby shutdown in Spring Boot context

tleipzig:
## Impact

`TypeBug`

## Description

After liquibase changelog is applied to a Derby embedded database in Spring Boot context, database is shut down. Spring Boot application cannot run any further, as there is no connection available anymore. Sometimes it still works as the shutdown command isn't executed relyably.

Logic adapted from liquibase.integration.servlet.GenericServletListener.

## Additional Context

Not sure why shutdown enabled is the default for Derby databases. Maybe worth to change the default, maybe worth allowing the setting to be overwritten with a property as stated here, but it doesn't work.
https://stackoverflow.com/questions/47314248/why-derby-connection-closed-after-liquibase-update


filipelautert (approved):
Makes sense as we do the same for jee environments. Thanks for the PR @tleipzig !

MalloD12 (approved):

PR #4635, opened 2023-08-09
generateChangelog optionally creates runOnChange=true and replaceIfExists=true for createView changes

mkarg:
Closes #4634 

## Impact

- [ ] Bug fix (non-breaking change which fixes expected existing functionality)
- [x] Enhancement/New feature (adds functionality without impacting existing logic)
- [ ] Breaking change (fix or feature that would cause existing functionality to change)
 

## Description

This PR introduces two new options:
* `--run-on-change-types=createView,createProcedure` will set `runOnChange="true"` for each `changeSet` which contains *solely* `createView` and `createProcedure` changes. For backwards compatibility, the default is an empty list.
* `--replace-if-exists=createView,createProcedure` will set `replaceIfExists="true"` for each `createView` and `createProcedure` change. For backwards compatibility, the default is an empty list. 

## Things to be aware of

N/A

## Things to worry about

N/A

## Additional Context

N/A

PR #5304, opened 2023-11-30
Fix for DAT-15716 :: Include 'OR REPLACE' instruction for a view when generate-changelog/diff-changelog command are executed

MalloD12:
`USE_OR_REPLACE_OPTION` command argument added to control whether we want to include `OR REPLACE` SQL instruction for a VIEW when executing a `generate-changelog` or `diff-changelog` command.

Integration tests were added for both commands.

## Impact

## Things to be aware of
- None

## Things to worry about
- None


liquibase-standard/src/main/java/liquibase/GlobalConfiguration.java
```diff
@@ -240,9 +243,15 @@ public class GlobalConfiguration implements AutoloadedConfigurations {
...
         ALWAYS_DROP_INSTEAD_OF_REPLACE = builder.define("alwaysDropInsteadOfReplace", Boolean.class)
                 .setDescription("If true, drop and recreate a view instead of replacing it.")
+                .setDefaultValue(false)
                 .setValueHandler(ValueHandlerUtil::booleanValueHandler)
                 .build();
 
+        USE_OR_REPLACE_OPTION = builder.define("useOrReplaceOption", Boolean.class)
+                .setDescription("If true, will add 'OR REPLACE' option to the given stored logic change object for example, create view, create procedure, etc.)")
+                .setDefaultValue(false)
+                .build();
```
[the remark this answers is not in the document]
MalloD12: done

liquibase-cli/src/test/groovy/liquibase/integration/commandline/LiquibaseCommandLineTest.groovy
```diff
@@ -425,6 +426,15 @@ Global Options
                              (defaults file: 'liquibase.uiService', environment
                                variable: 'LIQUIBASE_UI_SERVICE')
 
+      --use-or-replace-option=PARAM
+                             If true, will add 'OR REPLACE' option to the given
+                               stored logic change object for example, create
+                               view, create procedure, etc.)
```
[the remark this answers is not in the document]
MalloD12: I haven't updated this one yet, I'm trying to check this one with Mario. I think there is no relationship between these two configurations If I'm not wrong `ALWAYS_DROP_INSTEAD_OF_REPLACE` is mostly used for `update` command (for example, when changing the structure of a view and we don't want to get an error in the try of updating a new version of it). Whereas the new `USE_OR_REPLACE_OPTION` configuration is used for `generate-changelog` and `diff-changelog` commands.

liquibase-standard/src/main/java/liquibase/sqlgenerator/core/CreateViewGenerator.java
```diff
@@ -114,7 +114,7 @@ public Sql[] generateSql(CreateViewStatement statement, Database database, SqlGe
                 // 1) It is already in the SQL
                 // 2) The DB2 version is < 10.5
                 //
-                if (!statement.getSelectQuery().toUpperCase().contains("OR REPLACE")) {
+                if ((!statement.getSelectQuery().toUpperCase().contains("OR REPLACE"))) {
```
filipelautert, line 117: The extra parenthesis can be removed here.

liquibase-standard/src/main/java/liquibase/diff/output/changelog/DiffToChangeLog.java
```diff
@@ -340,6 +348,20 @@ public List<ChangeSet> generateChangeSets() {
         return changeSets;
     }
 
+    private void setReplaceIfExistsTrueIfApplicable(Change[] changes) {
+        if(changes !=null && diffOutputControl.isReplaceIfExistsSet()) {
+            for(int i=0; i < changes.length; i++) {
+                if (changes[i] instanceof CreateViewChange) {
```
filipelautert, line 354: Here instead of the two if's you can just check if it is instanceof  ReplaceIfExists then call the method as both classes are implementing the interface, ie:

```
if (changes[i] instanceof ReplaceIfExists ) {
 ((ReplaceIfExists ) changes[i]).setReplaceIfExists(true);
}
```

filipelautert (changes requested):
Please also add the two @Override requested by codeQL

liquibase-cli/src/test/groovy/liquibase/integration/commandline/LiquibaseCommandLineTest.groovy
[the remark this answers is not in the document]
StevenMassaro: We should probably explain that in the description, at least in the description of the property that you're adding.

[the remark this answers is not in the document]
MalloD12: Well, I finally removed that global configuration and instead added it as a command argument of both (`generate-changelog` and `diff-changelog`) commands respectively. We should also reflect this in the docs.

filipelautert (approved):

MalloD12:
Guys, as discussed in [DAT-15716](https://datical.atlassian.net/browse/DAT-15716) I have added support in this PR for using `useOrReplaceOption` property in `mvn` for `generate-changelog` and `diff-changelog`. Could you please have a look at those changes?

Thanks,
Daniel.

filipelautert (approved):
All fine

liquibase-standard/src/main/java/liquibase/command/core/DiffChangelogCommandStep.java
```diff
@@ -43,6 +46,11 @@ public class DiffChangelogCommandStep extends AbstractCommandStep {
                 .addAlias("contexts")
                 .description("Changeset contexts to generate")
                 .build();
+        USE_OR_REPLACE_OPTION = builder.argument("useOrReplaceOption", Boolean.class)
+                .description("If true, will add 'OR REPLACE' option to the create view change object")
+                .defaultValue(false)
+                .setValueHandler(ValueHandlerUtil::booleanValueHandler)
+                .build();
```
StevenMassaro, line 53: It isn't necessary for this PR, but instead of repeating the argument, you can make a new class that extends `AbstractArgumentCommandStep`, and then specify that as a `requiresDependency` here and in every other class which requires this argument. This helps to reduce the code duplication.

StevenMassaro (approved):

liquibase-standard/src/main/java/liquibase/command/core/DiffChangelogCommandStep.java
MalloD12: I created an abstract class with some common behavior that `GenerateChangelogCommandStep` and `DiffChangelogCommandStep` both use, I was thinking once that [PR](https://github.com/liquibase/liquibase/pull/4635/files) is merged to move this new argument there, and then in these two command classes do this:
```
builder.addArgument(AbstractChangelogCommandStep. USE_OR_REPLACE_OPTION).build();
```
What do you think about it?

Thanks,
Daniel.

StevenMassaro: That sounds good. I don't think that's been done before, but it sounds simpler.

MalloD12: It is! @filipelautert showed me yesterday that was already applied to `RollbackCommandStep`, so I applied the same to GenerateChangelog and DiffChangelog command step classes.

PR #5315, opened 2023-12-05
Allow DiffResults to be reused in pipeline after running DiffChangelogCommandStep (DAT-16374)

abrackx:
## Impact

## Description
When generating missing primary keys, set relation back to original relation once we have generated changes with it to allow reuse of DiffResults in subsequent command steps.

## Things to be aware of
Pro is consuming the DiffResult generated from DiffCommandStep and DiffChangelogCommandStep. The modification `backingIndex.getRelation().setSchema(schema.getCatalogName(), schema.getSchemaName());` breaks the `DiffResult#changedObjects` field if we generate a primary key column by updating the `key` of `changedObjects` to be something different than what the snapshot associated with the DiffResult expects.

## Things to worry about
This code has existed in this form since around 2015-2018. I'm a bit scared to touch it.


abrackx:
Trying to add a test for this now but I'd like to get comments on this change as soon as possible so marking as ready for review.

StevenMassaro (approved):
I have no objections to this change, but it's also a soft approval because I don't really know the impact of it. We might just have to rely on functional tests for this one.

wwillard7800 (approved):

MalloD12 (approved):

filipelautert (approved):
Better too safe than sorry!

Issue #3643, opened 2023-01-05
primaryKeyExists precondition crashes when used with H2 2.x


teppok:
## Environment

**Liquibase Version**:

At least 4.9.1 through 4.18.0

**Liquibase Integration & Version**: <Pick one: CLI, maven, gradle, spring boot, servlet, etc.>

Spring boot 2.7.7

**Liquibase Extension(s) & Version**: 

**Database Vendor & Version**:

H2 2.1.214

**Operating System Type & Version**:

Windows 10 20H2

**Infrastructure Type/Provider**: <AWC, GCS, Azure, VM, etc>

## Description

When using primaryKeyExists precondition without tableName attribute, with H2 2.1.214 database, H2 causes Liquibase to crash with an exception. 

## Steps To Reproduce

An example changelog:
```
<databaseChangeLog
        xmlns="http://www.liquibase.org/xml/ns/dbchangelog"
        xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
        xsi:schemaLocation="
        http://www.liquibase.org/xml/ns/dbchangelog
        http://www.liquibase.org/xml/ns/dbchangelog/dbchangelog-3.3.xsd">


    <changeSet id="test-1" author="tk">
        <createTable tableName="test">
            <column name="id" type="bigint">
                <constraints nullable="false"/>
            </column>
        </createTable>
    </changeSet>

    <changeSet id="test-2" author="tk">
        <preConditions onFail="MARK_RAN">
            <not>
                <primaryKeyExists
                        schemaName="public"
                        primaryKeyName="test_pkey"/>
                <!-- works:
                <primaryKeyExists
                        schemaName="public"
                        tableName="test"
                        primaryKeyName="test_pkey"/>
                -->

            </not>
        </preConditions>

        <addPrimaryKey
                tableName="test"
                schemaName="public"
                constraintName="test_pkey"
                columnNames="id"/>
    </changeSet>
</databaseChangeLog>
```

From the spec, it seems that this is valid pair of changesets, However, when ran on H2, this fails with
```
org.springframework.beans.factory.BeanCreationException: Error creating bean with name 'liquibase' defined in org.springframework.boot.autoconfigure.liquibase.LiquibaseAutoConfiguration$LiquibaseConfiguration: Invocation of init method failed; nested exception is liquibase.exception.LiquibaseException: liquibase.exception.MigrationFailedException: Migration failed for change set db/changelog/changelog-master.xml::test-2::tk:
[...]

PR #5323, opened 2023-12-06
Updated PrimaryKeyExistsPrecondition to require table name for H2 databases

filipelautert:
## Impact

## Description

This PR updates the PrimaryKeyExistsPrecondition in order to throw a friendly message when no table name is provided while using H2 database. This change has been done because the H2 database driver requires a table name to be specified for searching a primary key.

## Doc notes

This precondition fails on databases which don't support the null as third parameter Documentation is required on  https://docs.liquibase.com/concepts/changelogs/preconditions.html mentioning that due to JDBC driver limitations, is some databases such as H2 the tableName is required.

Fixes https://github.com/liquibase/liquibase/issues/3643


rberezen (approved):

rberezen:
@MalloD12 all good, thanks! 

> Code changes looks good to me. I have add some other DBs so precondition returns the same given error and added some integration tests.
> 
> @rberezen Would you mind looking at the test cases added and see if they look good enough for you?
> 
> Thanks, Daniel.

liquibase-integration-tests/src/test/java/liquibase/dbtest/AbstractIntegrationTest.java
```diff
@@ -1310,6 +1308,27 @@ public void verifyIndexIsCreatedWhenAssociatedWithPropertyIsSetAsNone() throws D
...
+            errorMsg = e.getMessage();
+        }
+        finally {
+            clearDatabase();
+        }
+
+        if(!(database instanceof H2Database || database instanceof MySQLDatabase || database instanceof HsqlDatabase
+                || database instanceof SQLiteDatabase || database instanceof DB2Database)) {
+            Assert.assertTrue(errorMsg.isEmpty());
+        }
```
filipelautert, line 1329: Hi @MalloD12 ! Instead of replicating the test code 5 times, would not be better to add an validation here for the expected error message too?

MalloD12: Hi @filipelautert, I wouldn't add to the same test to validate two possible scenarios in the same test, but yes I can add a new test with a similar if logic and in that case assert error message is present in that case.

filipelautert: Nice, otherwise if the message is changed we wil need to fix 5 tests instead of just one ;)

MalloD12: Done

adrian-velonis1:
Docs internal: https://datical.atlassian.net/browse/PD-4392

PR #5391, opened 2023-12-14
fix SQLFileChange.generateCheckSum() to calculate checksum without property substitution

### LINKED ISSUE #5389: SQLFileChange calculates checksum after substituting properties
### Search first

### Description

According to https://docs.liquibase.com/concepts/changelogs/property-substitution.html:
`Note that Liquibase obtains the checksum of a sql changeset after substituting any properties you specify. However, it obtains the checksum of a sqlFile changeset before substituting the properties in the external SQL file.`
but this does not appear to be true in 4.22+. The following changeset has different checksums in different schemas:
```
<?xml version="1.0" encoding="utf-8"?>
<changeSet author="jglass" id="sqlFile">
    <sqlFile path="001_test.sql" relativeToChangelogFile="true" splitStatements="false"/>
</changeSet>
```
001_test.sql:
```
select * from ${database.liquibaseSchemaName}.customer;
```

### Steps To Reproduce

A unit test can demonstrate the problem. Add this file to the liquibase source code at _liquibase-standard/src/test/resources/com/example-2/fileWithSchemaNameProperty.sql_:
```
select * from ${database.liquibaseSchemaName}.customer;
```
and add this function to _liquibase-standard/src/test/groovy/liquibase/change/core/SQLFileChangeTest.groovy_:
```
    def noReplacementOfPropertiesInChecksum() throws Exception {
        when:
        def changelog = new DatabaseChangeLog("com/example/changelog.xml")
        def change1 = new SQLFileChange()
        change1.path = "com/example-2/fileWithSchemaNameProperty.sql"
        change1.relativeToChangelogFile = false
        def changeLogParameters1 = new ChangeLogParameters()
        changeLogParameters1.set("database.liquibaseSchemaName", "schema1")
        def changeSet1 = new ChangeSet("x", "y", false, false, null, null, null, changelog)
        changeSet1.setChangeLogParameters(changeLogParameters1)
        change1.setChangeSet(changeSet1)
 
        def sqlForSchema1 = change1.getSql().trim()
        def checksum1 = change1.generateCheckSum()

        def change2 = new SQLFileChange()
        change2.path = "com/example-2/fileWithSchemaNameProperty.sql"
        change2.relativeToChangelogFile = false
        def changeLogParameters2 = new ChangeLogParameters()
        changeLogParameters2.set("database.liquibaseSchemaName", "schema2")
        def changeSet2 = new ChangeSet("x", "y", false, false, null, null, null, changelog)
        changeSet2.setChangeLogParameters(changeLogParameters2)
        change2.setChangeSet(changeSet2)

        def sqlForSchema2 = change2.getSql().trim()
        def checksum2 = change2.generateCheckSum()

        then:
        assertEquals("select * from schema1.customer;", sqlForSchema1)
        assertEquals("select * from schema2.customer;", sqlForSchema2)
        assertEquals(checksum1, checksum2)
    }
```

### Expected/Desired Behavior

An SQLFileChange object with an external SQL file should return the same checksum value even if the properties referenced in that file are changed.

### Liquibase Version

4.25.0

### Database Vendor & Version

any

### Liquibase Integration

any

### Liquibase Extensions

none

any

### Additional Context

_No response_

jglass524:
## Impact

## Description

Add an overload of SQLFileChange.getSql() that has a parameter that indicates whether property substitution should be done on the contents of the SQL file and fix SQLFileChange.generateCheckSum() to call that overloaded method so that property substitution is not done.

Fixes #5389 

## Things to be aware of

The existing SQLFileChange.getSql() method continues to return the contents of the SQL file with property substitution done so as to minimize the impact of introducing the overloaded method.


filipelautert:
Thanks for the pr @jglass524 , this fix makes a lot of sense! I'm just thinking that people that already have checksums calculated using the current method will face checksums problems. 
In method `ValidatingVisitorUtil.isChecksumIssue` we do workarounds  to make sure that fixes like this won't break the checksum calculation for current users.  Method `validateMongoDbExtensionIssue` is a good example: we try to validate the failed checksum using the "broken" approach, and if it matches we update the checksum to the new value making everything transparent to the end user. 
Do you think you could implement something like that to fix the checksums using property substitution too? If not I can help with that but I'll be on pto until next year.

jglass524:
I will look into a fix for ValidatingVisitorUtil.isChecksumIssue, thanks!

jglass524:
I'm not sure whether ValidatingVisitorUtil.validateSqlFileChangeAndExpandExpressions is supposed to call changeLogService.replaceChecksum(changeSet). validateMongoDbExtensionIssue does, and that's the method you pointed out as an example, but it's not clear to me why validateAbstractSqlChangeV8ChecksumVariant doesn't

jglass524:
is there anything to do in the case where the changelog parameters have changed since the checksum was stored? I can't think of anything but I'm not familiar with all of liquibase's features

filipelautert:
Hello @jglass524 ! You are right about `validateAbstractSqlChangeV8ChecksumVariant `: is not updating the checksum here because in the next steps of the flow Liquibase is going to upgrade it to v9. 
But the `validateMongoDbExtensionIssue` updates it as the problem was happening for checksums **v8 and v9** when we discovered it, so it needs to fixed anyway. Finally, the idea behind fixing it is that this part of the code is triggered only once for a changeset - so in the next update the usual checksum validation works and this heavy weight validation is never reached.

filipelautert (approved):
@jglass524 thanks for this fix, it's really appreciated.

@rberezen Some testing notes: this PR fixes the problem that v9 checksum is calculated after variable substitution for sqlFile (it was before with v8) .
This bug existed from 4.22.0 to 4.25.1 . 

It also will fix the checksum when the property value still the same. If you change the property value before running it will fail as before.

So this PR works when:
* I jump from before 4.22.0 to here, all fine
* I don't change a property value when upgrading from any versions + it fixes the checksum

It will fail in the same case as the previous versions:  when a variable value is changed before running this fix.

Ang going forward the problem is gone.

rberezen (approved):
@jglass524 thank you very much for your contribution! 
Tested, looks good to me. 
Need to merge https://github.com/liquibase/liquibase-pro-tests/pull/1211 tests, after this one is ready.

filipelautert:
> @jglass524 thank you very much for your contribution! Tested, looks good to me. cc @filipelautert we need to merge [liquibase/liquibase-pro-tests#1211](https://github.com/liquibase/liquibase-pro-tests/pull/1211) tests, after this one is ready.

@rberezen done!

@jglass524 thanks again

erik-meuwese-topicus:
With 4.25.1 our v7 and v8 checksums were upgraded to v9 checksums. This change causes the v9 checksums to change for change sets with `<sqlFile path=".....sql" relativeToChangelogFile="true" />`.

PR #5398, opened 2023-12-17
Fix issue #3516 addColumn ignores deleteCascade=true

### LINKED ISSUE #3516: addColumn ignores deleteCascade=true
## Environment

**Liquibase Version**: 4.17.2

**Liquibase Integration & Version**: Library call

**Liquibase Extension(s) & Version**:  N/A

**Database Vendor & Version**: Postgresql 11

**Infrastructure Type/Provider**: AWS

## Description

The following YAML:
```yaml
databaseChangeLog:

  - changeSet:
      id: an_id
      author: an_author
      changes:

        - addColumn:
            tableName: INVITATIONS
            columns:
              - column:
                  name: ORGANISATION_ID
                  type: BIGINT
                  constraints:
                    foreignKeyName: FK_INVITATION_ORGANISATION
                    referencedTableName: ORGANISATIONS
                    referencedColumnNames: ORGANISATION_ID
                    deleteCascade: true
```

does not generate the expected `ON DELETE CASCADE` in the foreign key creation.

It should. The documentation for [addColumn](https://docs.liquibase.com/change-types/add-column.html) references the documentation for [column](https://docs.liquibase.com/concepts/changelogs/attributes/column.html) which documents the [Constraints sub-tag](https://docs.liquibase.com/concepts/changelogs/attributes/column.html#Constraints%20sub-tag:~:text=Constraint%20definitions-,Constraints%20sub%2Dtag,-The%20%3Cconstraints%3E) as containing `deleteCascade`.

andreiyusupau:
Handling of deleteCascade config added to AddColumnGenerator.

## Impact

## Description
Fixes #3516 

## Things to be aware of
Added condition to handle the constraint config with `deleteCascade=true` in `AddColumnGenerator`. Added corresponding test.

## Things to worry about
1) I don't know is it ok to refactor existing code (for example adding a builder to AddForeignKeyConstraintStatement), ready to do it if it's appropriate to be done within the bugfix.
2) Should the constants be used? Didn't found a constant for "CASCADE" , but this String seems be to duplicated in many places through the code.

### DIFF
--- a/.github/workflows/build-azure-uber-jar.yml
+++ b/.github/workflows/build-azure-uber-jar.yml
@@ -113,7 +113,7 @@ jobs:
           cd liquibase-pro/liquibase-azure-deps && mvn -B clean package
 
       - name: Archive Modules
-        uses: actions/upload-artifact@v3
+        uses: actions/upload-artifact@v4
         with:
           name: liquibase-pro-azure-artifacts
           path: |
--- a/.github/workflows/build-azure-uber-jar.yml
+++ b/.github/workflows/build-azure-uber-jar.yml
@@ -113,7 +113,7 @@ jobs:
           cd liquibase-pro/liquibase-azure-deps && mvn -B clean package
 
       - name: Archive Modules
-        uses: actions/upload-artifact@v4
+        uses: actions/upload-artifact@v3
         with:
           name: liquibase-pro-azure-artifacts
           path: |
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -301,19 +301,19 @@ jobs:
           cp $commercialJarM2Location artifacts-named/liquibase-commercial-${{ needs.setup.outputs.thisBranchName }}.jar
 
       - name: Archive Packages
-        uses: actions/upload-artifact@v3
+        uses: actions/upload-artifact@v4
         with:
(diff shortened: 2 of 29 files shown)


filipelautert (approved):
Do all the databases that supports CASCADE uses the same syntax?

MalloD12:
> Do all the databases that supports CASCADE uses the same syntax?

Good point, I think I'll update the provided test and make sure he executed against some different DBs.

Thanks,
Daniel.-

MalloD12:
@filipelautert @andreiyusupau I just updated the existent test parameterized by database (and 'NULL'  string depending on if the given database adds that keyword if column accepts null value for it).

Test findings:

- **Sybase**: does not complain if we set `addForeignKeyConstraintStatement.setOnDelete("CASCADE")` for it, but when generating SQL statement it doesn't add the `"ON DELETE CASCADE"` instruction.
- **Sqlite**: does not complain either if we set `addForeignKeyConstraintStatement.setOnDelete("CASCADE")` for it, but when generating SQL statement it adds `"ON DELETE CASCADE"` instruction, but there are some other SQL statements as well, so we might need to tweak the test a bit more to make it pass for this particular case. I don't think it's worth, but if you think we should add it we can figure this out.

Thanks,
Daniel.

PR #5415, opened 2023-12-21
Restore and deprecate API methods used by extensions.

filipelautert:
## Impact

## Description

PR #4635 added new parameters to some methods in CommandLineUtils, but those methods are used by some extensions and maybe people that uses Liquibase. This PR restore the old methods signatures passing default values for the new parameters and deprecate them.

Example failure: https://github.com/liquibase/liquibase-vertica/actions/runs/7285058778/job/19851361305#step:8:2090


MalloD12 (approved):
Makes sense, thanks Filipe!

PR #5446, opened 2024-01-08
Proposed fix for #5442 - command "unexpectedChangesets" always reports at least 1 unexpected changeset with initial tag in database

### LINKED ISSUE #5442: command "unexpectedChangesets" always reports at least 1 unexpected changeset with initial tag in database
### Search first

### Description

Performing "tag" on an empty database creates a "liquibase-internal" changeset.
"unexpectedChangesets" fails to disregard this changeset and always reports at least 1 unexpected changeset.

### Steps To Reproduce

Step 1: start with empty database
Step 2: tag "mytag"
Step 3: execute 1 changeset
Step 4: run "unexpectedChangesets" with no changes made after step 3
Step 5: get 1 unexpected changeset reported back

### Expected/Desired Behavior

get **no** unexpected changeset reported back

### Liquibase Version

4.25.1

### Database Vendor & Version

Oracle

### Liquibase Integration

-core lib in java project

### Liquibase Extensions

- none -

Win 11

### Additional Context

_No response_

Tylorjg:
## Impact

## Description

Issue #5442 outlines an odd behavior where the `unexpected-changesets` command shows the `tag` changeset as unexpected. If using the `tag` command itself, a changeset is not added to the changelog, but an internal Liquibase changeset will be added to the DBCL table. This would cause the `unexpected-changesets `to always show the internal Liquibase changeset as unexpected. Side note, the `generate-changelog` does not create the changeset in a changelog if it exists in the DBCL, so there isn't a way to make the tag expected.

This fix adds a function that is called first to determine if the file path is the **liquibase-internal** changeset. If it is, it removes it from the unexpected list. 

Fixes #5442 

## Things to be aware of

I couldn't find a way to capture the file path for the ran changeset in the RanChangesSet.java file, so I did the same call that occurs in the isSameAs function (occurs in the isSamePath function technically), so it could be refactored to be more efficient. 

## Things to worry about

There isn't any additional tests associated with the change. I have found that there is a file unexpectedChangesets.test.groovy, but was unsure how to get the `tag` command to execute first. I would be happy to write test cases/modify the existing ones.

## Additional Context

N/A

### DIFF
--- a/liquibase-standard/src/main/java/liquibase/changelog/RanChangeSet.java
+++ b/liquibase-standard/src/main/java/liquibase/changelog/RanChangeSet.java
@@ -121,7 +121,7 @@ public class RanChangeSet {
                 && isSamePath(changeSet.getFilePath());
     }
 
-    public boolean isSameAsIgnoreLiquibaseInternalChangeset(ChangeSet changeSet) {
+    public boolean isSameAsOrIsLiquibaseInternalChangeset(ChangeSet changeSet) {
         String ranChangesetPath = DatabaseChangeLog.normalizePath(this.getChangeLog());
         return ranChangesetPath.equalsIgnoreCase("liquibase-internal") || isSameAs(changeSet);
     }
--- a/liquibase-standard/src/main/java/liquibase/changelog/RanChangeSet.java
+++ b/liquibase-standard/src/main/java/liquibase/changelog/RanChangeSet.java
@@ -117,7 +117,7 @@ public class RanChangeSet {
 
     public boolean isSameAs(ChangeSet changeSet) {
         String normalizedFilePath = DatabaseChangeLog.normalizePath(this.getChangeLog());
-        return normalizedFilePath.equalsIgnoreCase("liquibase-internal") || (this.getId().equalsIgnoreCase(changeSet.getId())
+        return (this.getId().equalsIgnoreCase(changeSet.getId())
                 && this.getAuthor().equalsIgnoreCase(changeSet.getAuthor())
                 && isSamePath(changeSet.getFilePath(), normalizedFilePath));
     }
--- a/liquibase-standard/src/main/java/liquibase/changelog/RanChangeSet.java
+++ b/liquibase-standard/src/main/java/liquibase/changelog/RanChangeSet.java
@@ -119,7 +119,11 @@ public class RanChangeSet {
         return this.getId().equalsIgnoreCase(changeSet.getId())
                 && this.getAuthor().equalsIgnoreCase(changeSet.getAuthor())
                 && isSamePath(changeSet.getFilePath());
+    }
 
+    public boolean isSameAsIgnoreLiquibaseInternalChangeset(ChangeSet changeSet) {
+        String ranChangesetPath = DatabaseChangeLog.normalizePath(this.getChangeLog());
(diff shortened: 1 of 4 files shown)


MalloD12 (commented):
Hi @Tylorjg,

Thank you for adding submitting this PR to fix #5442. I was thinking maybe we can only update `isSameAs()` by also checking whether `filepath` is equals to `"liquibase-internal" `. Something like this:

```
public boolean isSameAs(ChangeSet changeSet) {
        return isSamePath("liquibase-internal") || (
                this.getId().equalsIgnoreCase(changeSet.getId())
                && this.getAuthor().equalsIgnoreCase(changeSet.getAuthor())
                && isSamePath(changeSet.getFilePath()));
    }
```
Then we don't need to create a new method nor update `ExpectedChangesVisitor` class.

Regarding tests, I think it would be nice if we can add some integration test [here](https://github.com/liquibase/liquibase/tree/master/liquibase-integration-tests/src/test/groovy/liquibase/command/core) by creating a new class (UnexpectedChangesetsCommandStepTest), for adding this test you can also use `CommandUtil` test class where you have a `runTag()` and probably you can also add a `runUnexpectedChangeset()` to validate the given issue.

Please let me know if you have any question/concerns.

Thanks,
Daniel.

MalloD12 (approved):
**Approved**.

Thanks @Tylorjg for submitting this PR and addressing my review comment. I have manually tested this change, plus also adding a basic integration test to validate your fix against MySQL in this case, but it could against any DB.

Build and test checks have been successfully executed (there are some functional test failing, but that doesn't seem to be related with this change.

filipelautert (approved):

rberezen (changes requested):
@Tylorjg @filipelautert @MalloD12 
The fix introduces a regression. So, if you execute the following commands:

1. liquibase tag tag1
2. liquibase update
3. liquibase history

you will observe that it appears as though nothing was deployed; the databasechangelog will be empty, but the deployed objects will exist in the database.
image
image

MalloD12:
@Tylorjg would you mind looking at the regression issue @rberezen mentioned above?

Thanks,
Daniel.

MalloD12:
Code changes look good to me, thank you for taking care of this regression issue @Tylorjg.

@rberezen would you mind trying again when you have a chance?

Thanks,
Daniel.

rberezen (approved):

liquibase-standard/src/main/java/liquibase/changelog/RanChangeSet.java
```diff
@@ -116,10 +116,10 @@ public String toString() {
     }
 
     public boolean isSameAs(ChangeSet changeSet) {
-        return this.getId().equalsIgnoreCase(changeSet.getId())
+        String normalizedFilePath = DatabaseChangeLog.normalizePath(this.getChangeLog());
```
filipelautert, line 119: Hey @Tylorjg ! I just noticed those changes to RanChangeSet (moving the normalizedFilePath from isSamePath to isSameAs methods. Seems they were part of some other tests but are not necessary here anymore - so could we rollback them ? The idea of calling normalize path at method isSamePath is that if author or id doesn't match them we never reach this expensive normalization.

filipelautert (changes requested):

filipelautert (approved):

PR #5457, opened 2024-01-11
prevent multiple databases in test classes

StevenMassaro:
## Impact

## Description

Due to a combination of the logic in `liquibase-extension-testing` used to determine whether a particular test class matches the specified databases, and the fact that the integration tests run in a matrix in GHA, test classes with more than one `DatabaseTestSystem` specified will not run on any platform in GHA. This PR prevents someone from accidentally creating a test class that has more than one `DatabaseTestSystem`.


abrackx (approved):

filipelautert (approved):
As it is building and tests are working seems we don't have this condition yet, so it's a good thing to prevent it.

Issue #5458, opened 2024-01-11
Context-Filter is not working with logical expressions


MalloD12:
### Search first

- [X] I searched and no similar issues were found

### Description

When using a logical expressions AND/OR in the context filter field of a changeset is not working. 

### Steps To Reproduce

Having a changeset like the one below:

```
<changeSet id="TAG-v1.3.0" author="dev-team" contextFilter="(JIRA-1 or TEST)">
        <tagDatabase tag="v1.3.0"/>
    </changeSet>
    <include file="reference_create-table.xml" relativeToChangelogFile="true" contextFilter="(JIRA-1 or TEST)"/>
```

and then executing the below command:

```
liquibase update --context-filter="JIRA"
```

Actual Result: context filtering is not working and both changesets are deployed.

### Expected/Desired Behavior

According to documentation:
```
In Liquibase 4.23.0+, you can specify a context using @ in addition to AND, OR, !, and parentheses in the changesets. This requires you to designate a context or label explicitly when Liquibase is run for the changeset to run. The attribute context: @test causes a changeset to not run if Liquibase runs without any contexts provided.
```

context filtering should filter both changesets and not deploy any of them.

Docs Ref: https://docs.liquibase.com/concepts/changelogs/attributes/contexts.html

### Liquibase Version

4.25.1

### Liquibase Integration

CLI

### Additional Context

None

### Are you willing to submit a PR?

- [ ] I'm willing to submit a PR (Thank you!)

nvoxland:
I think this is behaving as expected?

The context-filter attribute is setting the expression to filter contexts, but on the CLI you're specifying a context-filter too not a context. If you run `liquibase update --context=JIRA` does it work?

MalloD12:
> I think this is behaving as expected?
> 
> The context-filter attribute is setting the expression to filter contexts, but on the CLI you're specifying a context-filter too not a context. If you run `liquibase update --context=JIRA` does it work?

Hi @nvoxland,


Sorry for my late response, I missed your response on this issue. No, running `liquibase update --context=JIRA` didn't work and that's why I raised this issue. If I didn't get it wrong the command execution you mentioned should work with the changeset example I provided on this issue. 

Thanks,
Daniel.

filipelautert:
[...]

PR #5477, opened 2024-01-17
Labels & Context filters refactoring

### LINKED ISSUE #1813: Label expression using AND/OR does not work with inherited labels
When a label expression is provided that uses AND or OR, the label filter does not correctly handle inherited labels.

Example:

* master.xml:
`<include labels="a" file="include.xml"/>`
* include.xml:
    `<changeSet labels="b" ... />`
* label expression from CLI:
    `--labels="a and b"`

Line 45 of LabelChangeSetFilter.java treats inherited and normal labels as entirely separate and for the above example executes the following check which will never be true:
    `((a and b) matches b) and ((a and b) matches a)`

From the documentation I think that it should be combining both normal and inherited labels into a single collection and matching the label expression against that collection.

https://github.com/liquibase/liquibase/blob/256675bca43a089e32c940e10ed8ed7a5f9d66f6/liquibase-core/src/main/java/liquibase/changelog/filter/LabelChangeSetFilter.java#L45

### LINKED ISSUE #5175: Labels defined in Include tag are not taken into account when filtering
### Search first

### Description

When I define a label in a Include tag, then I can't filter by them from the terminal with the modifier "@".

### Steps To Reproduce

I have two files:

000-changelog-root.xml
```xml
<databaseChangeLog xmlns="http://www.liquibase.org/xml/ns/dbchangelog" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
                   xsi:schemaLocation="http://www.liquibase.org/xml/ns/dbchangelog http://www.liquibase.org/xml/ns/dbchangelog/dbchangelog-4.24.xsd">

    <changeSet id="TAG-v1.3.0" author="dev-team">
        <tagDatabase tag="v1.3.0"/>
    </changeSet>
    <include file="000_create-table.xml" relativeToChangelogFile="true" labels="JIRA-1"/>

</databaseChangeLog>
```

000_create-table.xml
```xml
<databaseChangeLog xmlns="http://www.liquibase.org/xml/ns/dbchangelog" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
                   xsi:schemaLocation="http://www.liquibase.org/xml/ns/dbchangelog http://www.liquibase.org/xml/ns/dbchangelog/dbchangelog-4.24.xsd">

    <changeSet id="001" author="dev-team">
        <createTable tableName="book">
            <column name="id" type="bigint">
                <constraints nullable="false" primaryKey="true" primaryKeyName="pk_book"/>
            </column>
            <column name="name" type="varchar(255)"/>
        </createTable>
    </changeSet>
</databaseChangeLog>
```

Note: JIRA-1 label is defined in the <include...> of file `000-changelog-root.xml`.

When I execute the command `liquibase update`, the rows of DATABASECHANGELOG has defined correctly this value in the field `LABELS`.

The problem is when you want to use the atributte `--labels` or `--label-filter`.

If you run `liquibase update --labels="@JIRA-1"` or `liquibase rollback --labels="@JIRA-1"` no changesets are found.

If instead of define the label in the tag `<include...>` you define the label in every changeSet (`<changeSet id="001" author="dev-team" label="JIRA-1">`) the commands are executed correctly.

### Expected/Desired Behavior

The commands should make the same if the label is defined in `<include...>` tag or in `<changeSet...>` tag.

### Liquibase Version

4.24.0

### Database Vendor & Version

any

### Liquibase Integration

_No response_

### Liquibase Extensions

_No response_

_No response_

### Additional Context

_No response_

MalloD12:
## Impact

## Description
Fixes #1813

Tickets discovered:

- https://github.com/liquibase/liquibase/issues/5458
- Doc update required: https://datical.atlassian.net/browse/PD-4084


filipelautert:
@MalloD12 this one has been stalled since January.. do you plan to still working on that?

MalloD12:
Hey @filipelautert,

This refactoring is pretty big and could be a breaking change. It would still require some more testing and probably some extra code change. That's why since it doesn't seem to be something urgent I leave it in the backlog. Do you think this is something worth we should plan and addressing in some upcoming sprint?

 Thanks,
Daniel.

MalloD12:
Closing this PR since several changes/refactorings have been made in the last time around labels/contexts that have fixed the original issue.

Issue #5334, opened 2023-12-08
dev experience: provide maven wrapper for all maven-based liquibase projects


rursprung:
### Search first

- [X] I searched and no similar issues were found

### Description

it'd be great if you could add a [maven wrapper](https://maven.apache.org/wrapper/) to all the maven-based liquibase projects.
this way, getting started is as easy as `git clone ...; cd ...; ./mvnw build` and you don't need to first figure out which maven version you need and then install that globally.

the benefit for you is also that all developers will use the same version and you thus don't have to deal with issues where a developer just had the wrong maven version running.

### Steps To Reproduce

see above

### Expected/Desired Behavior

see above: the maven wrapper should be available in all liquibase repositories which are based on maven.

### Additional Context

in gradle based projects this has been common/standard for a long time: i haven't seen a single gradle-based project which didn't come with a wrapper. it just makes life so much easier for everyone involved.

### Are you willing to submit a PR?

- [ ] I'm willing to submit a PR (Thank you!)

kevin-atx:
@rursprung - this sounds like a good idea. Would you be willing to submit a PR for this? If so, would you want/need any guidance from a Liquibase developer?

rursprung:
i don't really have time for this right now (certainly not to do it for all your repos), which is why i explicitly didn't select that i'd be willing to provide PRs for it. also, i don't even have maven installed (which is why i asked for it in the first place) and have never done it (virtually all java-based projects i deal with have migrated to gradle a long time ago), though i doubt that it's hard to do (it certainly is easy with gradle 😄).

kevin-atx:
That sounds good - thanks for the quick response.  :)

filipelautert:
Hello @rursprung 
that's an interesting idea, thanks! Icreated PR https://github.com/liquibase/liquibase/pull/5482 by runnning "mvn wrapper:wrapper" and adding all the generated files. Seems that's it, nothing else?

rursprung:
thanks! i've added some comment on the PR

PR #5482, opened 2024-01-18
Adding maven wrapper

filipelautert:
## Impact

Adding maven wrapper support.

DOCS notes: fix https://contribute.liquibase.com/code/get-started/env-setup/ to get rid of step 2 and update step 3

Partially fixes https://github.com/liquibase/liquibase/issues/5334 .


mvnw
rursprung, line 1: you need to flag this as executable, you can run `git update-index --chmod=+x mvnw` and then commit it to do so

(under linux it should pick up the mode from the file system, under windows you have to do it explicitly in git with the command above, as windows does not have this concept)

rursprung:
also, you'll most likely want to change CI to use the wrapper and document that people can/should use the wrapper (in whichever dev. documentation you currently say that one should manually install maven first - if you even say that anywhere)

filipelautert:
Replaced by https://github.com/liquibase/liquibase/pull/5486

Issue #5474, opened 2024-01-17
Lexical error in 4.25.1 for string containing backslash and a unicode character


instagibb:
### Search first

- [X] I searched and no similar issues were found

### Description

I have a change set with an insert statement that contains a string with a backslash `\` and the unicode character `≤` (\u2264)

This works fine in `4.25.0` but in `4.25.1` I get the following error:
```
Caused by: liquibase.util.grammar.TokenMgrError: Lexical error at line 190, column 9.  Encountered: '\u2264' (8804),
	at liquibase.util.grammar.SimpleSqlGrammarTokenManager.getNextToken(SimpleSqlGrammarTokenManager.java:1410)
SimpleSqlGrammarTokenManager.java:1410
	at liquibase.util.grammar.SimpleSqlGrammar.getNextToken(SimpleSqlGrammar.java:113)
SimpleSqlGrammar.java:113
	at liquibase.util.StandardSqlParser.parse(StandardSqlParser.java:42)
StandardSqlParser.java:42
	at liquibase.util.StringUtil.processMultiLineSQL(StringUtil.java:68)
StringUtil.java:68
	at liquibase.change.AbstractSQLChange.generateStatements(AbstractSQLChange.java:295)
```

I can trigger this with a number of unicode characters but the one specifically from my changeset is `≤`

I assume it is related to #5108 which mentions `\` but it is not clear how to proceed

### Steps To Reproduce

A simple changeset like the following is enough

```sql
INSERT INTO MY_TABLE (NAME, VALUE) VALUES ('test', 'This has a \ (slash) and some unicode  symbol ≤ (u2264)');
```

### Expected/Desired Behavior

The Lexical error should not occur

### Liquibase Version

4.25.1

### Database Vendor & Version

PostgreSQL 15.3 (Debian 15.3-1.pgdg110+1)

### Liquibase Integration

Spring Boot

### OS and/or Infrastructure Type/Provider

MacOS

### Are you willing to submit a PR?

- [ ] I'm willing to submit a PR (Thank you!)

wmcshane:
Not sure if this is useful, however, I found that it would always produce the error if a unicode character with a higher hex value than UTF-16 (hex) 07EA was used. Note, if the string with the high hex value unicode didn't contain a backslash then the error wouldn't occur.

tati-qalified:
Hi @instagibb, thank you for reporting this issue. I was able to reproduce it, as well as what @wmcshane and @mk-github-77 reported.
[...]

PR #5506, opened 2024-01-25
fix: Parse utf8 symbols

### LINKED ISSUE #5483: Liquibase 4.25.1 complains about a certain UTF-8 character in PL/SQL code
### Search first

### Description

I assume it is related to https://github.com/liquibase/liquibase/pull/5474

When deploying our codebase with the help of Liquibase, an error is currently occurring with a certain package. The package can no longer be deployed and Liquibase terminates with an error message:

[2024-01-19 11:58:17] HEAVILY [liquibase.integration] Lexical error at line 489, column 73. Encountered: '\u00f7' (247),
liquibase.util.grammar.TokenMgrError: Lexical error at line 489, column 73. Encountered: '\u00f7' (247),
	at liquibase.util.grammar.SimpleSqlGrammarTokenManager.getNextToken(SimpleSqlGrammarTokenManager.java:1410)

Unexpected error running Liquibase: Lexical error at line 489, column 73. Encountered: '\u00f7' (247),

liquibase.util.grammar.TokenMgrError: Lexical error at line 489, column 73. Encountered: '\u00f7' (247),
	at liquibase.util.grammar.SimpleSqlGrammarTokenManager.getNextToken(SimpleSqlGrammarTokenManager.java:1410)

There were no problems with this package in the previous version. There have been no changes to this package since the last Liquibase version.

### Steps To Reproduce

The following ChangeSets should reproduce the error:

-- changeset.xml
<databaseChangeLog xmlns="http://www.liquibase.org/xml/ns/dbchangelog" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" xsi:schemaLocation="http://www.liquibase.org/xml/ns/dbchangelog http://www.liquibase.org/xml/ns/dbchangelog/dbchangelog-latest.xsd">

    <changeSet id="murks" runOnChange="true">
        <sqlFile path="murks.pks" relativeToChangelogFile="true" endDelimiter="\n/" />
        <sqlFile path="murks.pkb" relativeToChangelogFile="true" endDelimiter="\n/" />
    </changeSet>

</databaseChangeLog>

-- file: murks.pks - utf-8
create or replace package murks as

  procedure foobar;

end murks;
/

-- file: murks.pkb - utf-8
create or replace package body murks as

  procedure foobar
  is
    l_pat varchar2(200) := q'[-~+±_–—/\‚,(){}[]<>‹.#&''"`´’!?¿:;*|²³½¾^÷@¯£=$Þ°§·¦¢…%¶]';
    l_rep varchar2(200) := q'[                                                        ]';
  begin
    -- do nothing;
    null;
  end foobar;

end murks;
/

-- This char should be be the problem: ÷

### Expected/Desired Behavior

No errors when providing correct code with UTF-8 strings.

### Liquibase Version

4.25.1

### Database Vendor & Version

Oracle 19c

### Liquibase Integration

CLI

### Liquibase Extensions

_No response_

MacOS 14.2.1 - Sonoma / Linux

### Additional Context

_No response_

filipelautert:
## Impact

## Description

This PR expands our mapping of UTF-8 symbols to "anything between \u00A1-\uFF65 that wasn't specified as a letter before" . This should prevent errors of unknow UTF-8 and fix https://github.com/liquibase/liquibase/issues/5474 and fix #5483 

Of course this assumes that what we don't know is a symbol, not a letter so it should be parsed as a standalone character as explained below.

## SimpleSqlGrammar.jj basic explanation

The general goal for the jj file is to break the passed string up into tokens, where the tokens are the individual potentially-multi-letter things that are treated as stand-alone things/keywords.
Like if your string is "the cat eats" you have tokens of "the" "cat" and "eats".
For SQL, we basically have strings, numbers, and operators and we need to figure out what is what from the original string. Like "a + b" is ["a", "+", "b"] because each of those are stand-alone things. And if they had put "a+b" they get the same tokens.
Or "1+3" is [1, "+", 3] where we know 1 and 3 are numbers. If it is "13+31" you'd get [13, "+", 31] also as numbers.
But If it is "a1+b3" then even though there are numbers in there, those are now "words" because they have letters in there too, so ["a1", "+", "b3"]

So the COMPLEX_IDENTIFIERS on line 71 defines what makes those more complex things that can be letters, numbers, and some other punctuation like " and ' and $ etc. which still make up those "whole string" .
What is not included in those COMPLEX_IDENTIFIERS though is the SYMBOLS because that is the list of things which need to stand along regardless. Like in "a1+b3", + is a symbol because it's always a separate token whether it is "a1+b3" or "a1 + b3"

So what would end up happening if À ends up a symbol instead of a letter, then "Àx + Ày" would parse to [À, x, +, À, y] instead of ["Àx", "+", "Ày"] like it should

The grammar's job is just to break the original string into tokens, which the rest of the code then consumes. So the testing with it is making sure the letters are correctly grouped into tokens still .


liquibase-standard/src/main/javacc/liquibase/util/grammar/SimpleSqlGrammar.jj
```diff
@@ -511,155 +511,7 @@ TOKEN: /* symbols */
...
-    "\uFF1A",
-    "\uFF1B",
-    "\uFF1F",
-    "\uFF20",
-    "\uFF3B"-"\uFF3D",
-    "\uFF3F",
-    "\uFF5B",
-    "\uFF5D",
-    "\uFF5F"-"\uFF65"
+    "\u00A1"-"\uFF65" /* everything else */
```
filipelautert, line 514: Could we do this @nvoxland ?

nvoxland: The big goal is to make sure the list of "tokens" are:
- not letters
- not numbers
- not control characters
- not other weird things that aren't actually symbols/tokens

But if the tests are passing without having that list, the broader range may be OK because of how the generated code works because the letters and numbers are defined first?

For example, in the original list we started with "\u00A1"-"\u00BF", and then skipped to u037E. But the new list has everything starting u00A1 so it includes what comes after  u00BF aka u00C0 which is À and a letter not a symbol.

If we have a test that is checking for a something like À which is now in the list of "tokens" but it gets parsed correctly still because it's also in the list of "letters" and the final answer for it is "it's a letter" than we should be OK to be more broad in list since we can say "everything in this range, since it's already excluding what was defined for letters & numbers". 

But: make sure that the new, big "\u00A1"-"\uFF65" range doesn't have things that are not letters or numbers but also aren't really "tokens"

filipelautert: Thanks ! I added some new tests to make sure that those conditions are covered

instagibb:
I have tested this fix with my test case from #5474 and it appears to have fixed the issue

suryaaki2 (approved):

rberezen (approved):

PR #5513, opened 2024-01-26
fix nullability for columns in AddColumnStatement for SQLite

cornerman:
## Impact

## Description

An AddColumnChange with a non-nullable column currently creates a nullable column in SQLite.

Example changelog:
```yml
- addColumn:
    tableName: Person
    columns:
    - column:
        constraints:
          nullable: false
        name: blub
        type: text
```

This currently produces code like this (the `blub` column is created as nullable):
```
ALTER TABLE Person RENAME TO Person_temporary;
CREATE TABLE Person (id INTEGER CONSTRAINT PK_PERSON PRIMARY KEY AUTOINCREMENT NOT NULL, name TEXT NOT NULL, blub TEXT);
INSERT INTO Person (id,name) SELECT id,name FROM Person_temporary;
DROP TABLE Person_temporary;
REINDEX Person;
```

This PR fixes how ColumnConstraints are initialized inside the `AddColumnStatementSQLite`.

### DIFF
--- a/liquibase-standard/src/main/java/liquibase/change/ColumnConfig.java
+++ b/liquibase-standard/src/main/java/liquibase/change/ColumnConfig.java
@@ -71,22 +71,7 @@ public class ColumnConfig extends AbstractLiquibaseSerializable {
         }
 
 
-        if (columnSnapshot.getDefaultValue() != null) {
-            Object defaultValue = columnSnapshot.getDefaultValue();
-            if (defaultValue instanceof Boolean) {
-                setDefaultValueBoolean((Boolean) defaultValue);
-            } else if (defaultValue instanceof Number) {
-                setDefaultValueNumeric(defaultValue.toString());
-            } else if (defaultValue instanceof SequenceNextValueFunction) {
-                setDefaultValueSequenceNext((SequenceNextValueFunction) defaultValue);
-            } else if (defaultValue instanceof DatabaseFunction) {
-                setDefaultValueComputed((DatabaseFunction) defaultValue);
-            } else if (defaultValue instanceof Date) {
-                setDefaultValueDate((Date) defaultValue);
-            } else {
-                setDefaultValue(defaultValue.toString());
-            }
-        }
+        setDefaultValueObject(columnSnapshot.getDefaultValue());
         setDefaultValueConstraintName(columnSnapshot.getDefaultValueConstraintName());
 
         boolean nonDefaultConstraints = false;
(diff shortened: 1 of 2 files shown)


liquibase-standard/src/main/java/liquibase/change/ColumnConfig.java
```diff
@@ -473,6 +458,30 @@ public Object getValueObject() {
...
+
+    /**
+     * Set the default value according to its type.
+     */
+    public ColumnConfig setDefaultValueObject(Object defaultValue) {
+        if (defaultValue != null) {
+            if (defaultValue instanceof Boolean) {
+                setDefaultValueBoolean((Boolean) defaultValue);
+            } else if (defaultValue instanceof Number) {
+                setDefaultValueNumeric(defaultValue.toString());
```
cornerman, line 470: This was copied from the constructor, but might make sense to use this:
```
setDefaultValueNumeric((Number) defaultValue);
```

cornerman:
Alternatively, we could consider switching to the `ALTER TABLE ADD COLUMN`-syntax for sqlite (>= 3.35). See: https://www.quickprogrammingtips.com/sqlite/how-to-add-or-remove-columns-from-sqlite-tables.html

liquibase-standard/src/main/java/liquibase/change/ColumnConfig.java
cornerman:
```suggestion
                setDefaultValueNumeric((Number) defaultValue);
```

MalloD12: Yeap, we can do that.

MalloD12:
Hi @cornerman,

Thank you for your contribution with this PR. Code changes look good to me, the only thing it would be nice to add is some unit test to validate this SQL change. Would you mind adding some?

Thanks,
Daniel.

cornerman:
Hi @MalloD12,

I am not exactly sure, what I should test here. Afaik, this needs an integration test with sqlite, because you need metadata. For the ColumnConfig constructor - code was just moved - so the current tests should succeed as before.

Please let me know, what you would like to see here :)

MalloD12:
> Hi @MalloD12,
> 
> I am not exactly sure, what I should test here. Afaik, this needs an integration test with sqlite, because you need metadata. For the ColumnConfig constructor - code was just moved - so the current tests should succeed as before.
> 
> Please let me know, what you would like to see here :)

Hi @cornerman,

Yeap, you are right. This would require an integration test, not a unit test. Would you mind adding it? Please let me know if you have any question.

Thanks,
Daniel.

MalloD12:
Hi @cornerman,

Just checking if you are keen on adding some integration tests so we can move forward with this PR. If you have any questions or need some assistance with it please let me know.

Thanks,
Daniel.

cornerman:

I would still recommend to merge this PR, because the current implementation for sqlite is wrong.

PR #5518, opened 2024-01-26
fix DropColumnChange for SQLite when handling single column

cornerman:
## Impact

## Description

Using DropColumn with a single columns does not delete the column and just recreates the table as is.

Example:
```yml
- changeSet:
    changes:
    - dropColumn:
        columnName: age
        tableName: Person
```

Produces:
```sql
ALTER TABLE Person RENAME TO Person_temporary;
CREATE TABLE Person (id INTEGER CONSTRAINT PK_PERSON PRIMARY KEY AUTOINCREMENT NOT NULL, name TEXT NOT NULL, age INTEGER);
INSERT INTO Person (id,name,age) SELECT id,name,age FROM Person_temporary;
DROP TABLE Person_temporary;
REINDEX Person;
```

This does not actually remove anything. So this PR fixes the case where only a single column is dropped.


MalloD12:
Hi @cornerman,

Also, thank you for this PR. Code change looks good to me, but again it would be nice if we could add some unit tests. Can you please do that? 

One other thing I notice when looking at this PR is the definition `isMultiple()` which I think it's not completely accurate:
```
 private boolean isMultiple() {
        return (columns != null) && !columns.isEmpty();
    }
```
I think it should be:
```
 private boolean isMultiple() {
        return (columns != null) && columns.size() > 1;
    }
```

Thanks,
Daniel.

cornerman:
Hej @MalloD12,

Agree that the `isMultiple` is confusing, but I am not sure, we should do the change to check for cardinality, because the method is used to check whether to check the single column or the columns list. So, I would prefer to not do that refactoring in this PR. 

For further refactoring: Generally, it would make sense to only have a list and treat single element lists differently, instead of having to track two potential column sources - what should actually happen if both are defined.

For the tests, can you point me to the existing tests for the sqlite path of the DropColumnChange? I did not find any. There are some minor tests checking something with a MockDatabase, but that seems a bit difficult with all the `isInstanceOf` checks in the code - and all its code paths.

MalloD12:
> Hej @MalloD12,
> 
> Agree that the `isMultiple` is confusing, but I am not sure, we should do the change to check for cardinality, because the method is used to check whether to process the single column or the columns list. So, I would prefer to not do that refactoring in this PR.
> 
> For further refactoring: Generally, it would make sense to only have a list and treat single element lists differently, instead of having to track two potential column sources - what should actually happen if both are defined.
> 
> For the tests, can you point me to the existing tests for the sqlite path of the DropColumnChange? I did not find any. There are some minor tests checking something with a MockDatabase, but that seems a bit difficult with all the `isInstanceOf` checks in the code - and all its code paths.

Hi @cornerman,

My apologies for taking too long to provide any feedback, for some reason I missed the notifications of your PRs. 

I think we can update it because from what I can see when it's used it will iterate over a given collection of columns, and also this is a `private` method only used in the upadted class.  

Regarding tests, there are some DropColumnChange unit tests, but I think the right place to add some of this would be `SQLiteIntegrationTest`. If you can add any into that class or you can create a new `SQLiteIntegrationTest` under the groovy directory if you feel comfortable with Spock (this option would be preferable since at some point we aim to migrate unit and integration tests using Spock).

Please let me know if you need any help with tests.

Thanks,
Daniel.

cornerman:

I would still recommend to merge this PR, because the current implementation for sqlite is wrong.

MalloD12:
Hey @tati-qalified,

Would you mind testing this SQLite fix?

Thanks,
Daniel.

tati-qalified:
@MalloD12 this is what is being generated when running `liquibase update-sql` with this fix and the changelog:

```
databaseChangeLog:
  - changeSet:
      id: 1
      author: tati
      changes:
        - createTable:
            tableName: person
            columns:
              - column:
                  name: id
                  type: int
                  constraints:
                    primaryKey: true
              - column:
                  name: firstname
                  type: varchar(50)
              - column:
                  name: lastname
                  type: varchar(50)
                  constraints:
                    nullable: false

  - changeSet:
      id: 2
      author: tati
      changes:
        - dropColumn:
            columnName: lastname
            tableName: person
```

```
CREATE TABLE
    person (
        id INTEGER NOT NULL,
        firstname VARCHAR(50),
        lastname VARCHAR(50) NOT NULL,
        CONSTRAINT PK_PERSON PRIMARY KEY (id)
    );

-- Changeset changelog.yml::2::tati
-- WARNING The following SQL may change each run and therefore is possibly incorrect and/or invalid:
ALTER TABLE person
RENAME TO person_temporary;

CREATE TABLE
    person ();

DROP TABLE person_temporary;

REINDEX person;
```

Now all columns are being dropped. This is not yet fixed.

MalloD12:
Thank you, @tati-qalified! Since @cornerman is not keen to keep working and move forward with this PR I think we can just leave in the current state until someone else wants to continue with this work.

Thanks,
Daniel!

PR #5519, opened 2024-01-26
add SetNullableGenerator for SQLite

cornerman:
## Impact

## Description

Currently, we do not support set-nullable changes on SQLite. This PR implements this similar to how we add/rename/drop columns.

We first create a new temporary table with the nullability change, then insert the content of the existing table into it and remove the old table.
For dropping the not-null constraint, this should always work. For creating the not-null constraint, this might fail depending on the data, which seems acceptable to me.


MalloD12:
Hi @cornerman,

This one also overall looks good to me, but if you can add some unit tests that would be great.

There is one thing to fix in this PR and it's that we need to override `getPriority()` method as shown below:

```
@Override
    public int getPriority() {
        return PRIORITY_DATABASE;
    }
```

Thanks,
Daniel.

MalloD12:
Hi @cornerman,

Have you had a chance to update this PR?

Thanks,
Daniel.

cornerman:

lbadam:
@cornerman that's helpful. Thanks! 

sql-lite and that ecosystem is something we will push into more as it matures, but it will take some time 👍

mariochampion:
I am closing this PR and Issue as it is no longer being worked by cornerman, and is not yet a priority for the Liquibase team. thanks.

PR #5527, opened 2024-01-29
expand values for url, username, password that use secrets vaults in maven plugin (DAT-16431)

StevenMassaro:
## Impact

### DIFF
--- a/liquibase-maven-plugin/src/main/java/org/liquibase/maven/plugins/AbstractLiquibaseMojo.java
+++ b/liquibase-maven-plugin/src/main/java/org/liquibase/maven/plugins/AbstractLiquibaseMojo.java
@@ -1,9 +1,13 @@
 package org.liquibase.maven.plugins;
 
-import liquibase.*;
+import liquibase.GlobalConfiguration;
+import liquibase.Liquibase;
+import liquibase.Scope;
+import liquibase.ThreadLocalScopeManager;
 import liquibase.changelog.visitor.ChangeExecListener;
 import liquibase.changelog.visitor.DefaultChangeExecListener;
 import liquibase.command.core.helpers.DbUrlConnectionCommandStep;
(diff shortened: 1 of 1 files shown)


liquibase-maven-plugin/src/main/java/org/liquibase/maven/plugins/AbstractLiquibaseMojo.java
```diff
@@ -1176,6 +1193,11 @@ private Object getDefaultValue(Field field) throws IllegalAccessException {
     }
 
     private void setFieldValue(Field field, String value) throws IllegalAccessException {
+        try {
+            value = Scope.getCurrentScope().getSingleton(ConfiguredValueModifierFactory.class).override(value);
+        } catch (Exception e) {
+            getLog().warn("Failed to modify value for field " + field.getName(), e);
+        }
```
StevenMassaro, line 1200: @filipelautert Should I be catching the exception here? I think not, like if someone uses an `aws-secrets` secrets vault but doesn't have their AWS creds set up, the execution should fail, just like in the CLI.

filipelautert: If you don't catch it here, what would be the impact for the current behavior of the plugin? If none then just remove it :D

filipelautert (changes requested):

liquibase-maven-plugin/src/main/java/org/liquibase/maven/plugins/AbstractLiquibaseMojo.java
StevenMassaro: Thinking about it, there would be no impact. It should definitely be removed. I'll do that.

suryaaki2 (approved):

filipelautert (approved):

PR #5530, opened 2024-01-30
Use static final Pattern to avoid regexp compilation each time replaceAll is called

turbanoff:
## Impact

## Description

Improve performance of `liquibase.Liquibase#getDatabaseChangeLog()` a bit. I've noticed that pattern compilation takes some time in my profile.

## Additional Context

String.replaceAll implementation:
https://github.com/openjdk/jdk/blob/fd8adf308357355bd33916ad80e2328c35434e5a/src/java.base/share/classes/java/lang/String.java#L3110
```
    public String replaceAll(String regex, String replacement) {
        return Pattern.compile(regex).matcher(this).replaceAll(replacement);
    }
```

image

### DIFF
--- a/liquibase-standard/src/main/java/liquibase/change/AbstractChange.java
+++ b/liquibase-standard/src/main/java/liquibase/change/AbstractChange.java
@@ -28,6 +28,7 @@ import java.lang.reflect.Method;
 import java.lang.reflect.Modifier;
 import java.lang.reflect.Type;
 import java.util.*;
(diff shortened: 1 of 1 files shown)


MalloD12 (approved):
**Approved**.

LGTM.

filipelautert (approved):

filipelautert (commented):
Thanks for the enhancement @turbanoff !

rberezen (approved):

PR #5548, opened 2024-02-01
Allow override of snapshot control creation for snapshotReference command DAT-16683

wwillard7800:
## Impact


liquibase-standard/src/main/java/liquibase/command/core/SnapshotCommandStep.java
```diff
@@ -53,27 +53,13 @@ public void adjustCommandDefinition(CommandDefinition commandDefinition) {
         commandDefinition.setShortDescription("Capture the current state of the database");
     }
 
-    private CatalogAndSchema[] parseSchemas(Database database, String... schemas) {
```
filipelautert, line 56: This method was duplicated? Nice catch

filipelautert (approved):

PR #5566, opened 2024-02-05
chunk long strings on insert for Oracle (DAT-16774)

StevenMassaro:
## Impact

## Description

Oracle forbids values longer than 4000 characters on an insert statement into a clob type column. This PR introduces a workaround which chunks the long string into 4000 character chunks.

## Things to worry about

If the string chunking code doesn't handle special characters properly, it is possible that you could end up with a string chunk that is longer than 4000 characters, causing an exception. However, this was also the behavior prior to this fix, so the impact is neutral in this case.


liquibase-standard/src/main/java/liquibase/datatype/core/CharType.java
```diff
@@ -1,18 +1,22 @@
...
+import liquibase.integration.commandline.LiquibaseCommandLineConfiguration;
 import liquibase.statement.DatabaseFunction;
 import liquibase.util.StringUtil;
 
 import java.math.BigInteger;
 import java.util.Arrays;
+import java.util.List;
 import java.util.Locale;
 
 @DataTypeInfo(name="char", aliases = {"java.sql.Types.CHAR", "bpchar", "character"}, minParameters = 0, maxParameters = 1, priority = LiquibaseDataType.PRIORITY_DEFAULT)
```
filipelautert, line 22: The issue mentions CLOB, but the change is being done to char (and varchar by extension). Should it also be applied to clob or the description could be changed?

filipelautert (changes requested):

liquibase-standard/src/main/java/liquibase/datatype/core/CharType.java
StevenMassaro: Liquibase treats it like a varchar insert, not a clob, because the Java data type is a string. This line is where Liquibase infers the type to be a CharType, despite the insert going into a clob column: https://github.com/liquibase/liquibase/blob/master/liquibase-standard/src/main/java/liquibase/sqlgenerator/core/InsertGenerator.java#L73

suryaaki2 (approved):

PR #5581, opened 2024-02-08
Handle stored logic with parameters in the dependency graph DAT-15747

wwillard7800:
## Impact

## Description
This change handles procedures and functions which have parameters.  The name of these objects that comes back from the dependency query does not include the actual parameter names, just the types.  This makes the dependency sorting ignore these objects and just put them at the bottom of the list, instead of in their correct place.


liquibase-standard/src/main/java/liquibase/diff/output/changelog/DiffToChangeLog.java
```diff
@@ -456,8 +454,10 @@ private List<DatabaseObject> sortObjects(final String type, Collection<DatabaseO
                                 schemaName = obj.getSchema().getName();
                             }
 
-                            String name = schemaName + "." + obj.getName();
-                            if (dependencyOrder.contains(name)) {
+                            String objectName = obj.getName();
+                            String name = schemaName + "." + objectName;
```
filipelautert, line 458: Nice, now we can have the same object in different schemas.

filipelautert (approved):

PR #5597, opened 2024-02-14
Format spring package.

asashour:
The current code sometimes uses 2-space indentation and 4-space one.

I think it makes sense to format the spring package (for now) for 4-space indentation, as a preparation to work on a PR for that package.

Please let me know if 2-space one is preferred.

### DIFF
--- a/liquibase-standard/src/main/java/liquibase/integration/spring/MultiTenantSpringLiquibase.java
+++ b/liquibase-standard/src/main/java/liquibase/integration/spring/MultiTenantSpringLiquibase.java
@@ -6,6 +6,8 @@ package liquibase.integration.spring;
 import liquibase.Scope;
 import liquibase.exception.LiquibaseException;
 import liquibase.logging.Logger;
+import lombok.Getter;
+import lombok.Setter;
 import org.springframework.beans.factory.InitializingBean;
 import org.springframework.context.ResourceLoaderAware;
 import org.springframework.core.io.ResourceLoader;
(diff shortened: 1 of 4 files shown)


MalloD12 (approved):
**Approved**.

Formatting changes look good to me. I removed a deprecated set method and its related get method as well, since they were not used and there was a codeQL scan issue reported (due to an unused argument).

No builds or test checks issues reported.

liquibase-standard/src/main/java/liquibase/integration/spring/MultiTenantSpringLiquibase.java
```diff
@@ -76,252 +79,253 @@ public class MultiTenantSpringLiquibase implements InitializingBean, ResourceLoa
...
+        liquibase.setDefaultSchema(defaultSchema);
+        liquibase.setLiquibaseSchema(liquibaseSchema);
+        liquibase.setLiquibaseTablespace(liquibaseTablespace);
+        liquibase.setDatabaseChangeLogTable(databaseChangeLogTable);
+        liquibase.setDatabaseChangeLogLockTable(databaseChangeLogLockTable);
+        return liquibase;
+    }
+
+
+    public String getJndiBase() {
```
filipelautert, line 183: As we are reviewing this class those methods could be replaced by lombok's @Setter and @Getter .

filipelautert (changes requested):

liquibase-standard/src/main/java/liquibase/integration/spring/MultiTenantSpringLiquibase.java
asashour: Done, the setters and getters which have JavaDoc are preserved.

filipelautert (approved):

liquibase-standard/src/main/java/liquibase/integration/spring/MultiTenantSpringLiquibase.java
filipelautert: Thanks @asashour !

PR #5609, opened 2024-02-20
fix dbclHistoryEnabled property from liquibase.properties file in maven (DAT-16881)

StevenMassaro:
## Impact

### DIFF
--- a/liquibase-maven-plugin/src/main/java/org/liquibase/maven/plugins/AbstractLiquibaseMojo.java
+++ b/liquibase-maven-plugin/src/main/java/org/liquibase/maven/plugins/AbstractLiquibaseMojo.java
@@ -654,7 +654,7 @@ public abstract class AbstractLiquibaseMojo extends AbstractMojo {
      *
      * @parameter property="liquibase.dbclHistoryEnabled"
      */
-    @PropertyElement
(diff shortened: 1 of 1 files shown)


suryaaki2 (approved):

liquibase-maven-plugin/src/main/java/org/liquibase/maven/plugins/AbstractLiquibaseMojo.java
```diff
@@ -654,7 +654,7 @@ public void setPassword(String password) throws Exception {
      *
      * @parameter property="liquibase.dbclHistoryEnabled"
```
filipelautert, line 655: Would changing this javadoc parameter do the same thing?

filipelautert (approved):

liquibase-maven-plugin/src/main/java/org/liquibase/maven/plugins/AbstractLiquibaseMojo.java
StevenMassaro: Interestingly, no. It matches on the name the Java field, or the key, not the property name. This is code that Liquibase wrote, which honestly could probably be improved in other ways.

Issue #5426, opened 2023-12-29
Multiple schema upgrade failed when done through multithreading


CG49071:
### Search first

- [X] I searched and no similar issues were found

### Description

Getting following error when run  liquibase update with @Async 
29-Dec-2023 07:39:41.507 SEVERE  liquibase.command.null Could not release lock
        liquibase.exception.LockException: liquibase.exception.DatabaseException: java.lang.RuntimeException: Cannot end scope bsbacsmwja when currently at scope root
                at liquibase.lockservice.StandardLockService.releaseLock(StandardLockService.java:407)
                at liquibase.command.core.helpers.LockServiceCommandStep.cleanUp(LockServiceCommandStep.java:50)
                at liquibase.command.CommandScope.execute(CommandScope.java:227)
                at liquibase.Liquibase.tag(Liquibase.java:1015)
                at org.springframework.aop.framework.ReflectiveMethodInvocation.invokeJoinpoint(ReflectiveMethodInvocation.java:198)
                at org.springframework.aop.framework.ReflectiveMethodInvocation.proceed(ReflectiveMethodInvocation.java:163)
                at org.springframework.aop.interceptor.AsyncExecutionInterceptor.lambda$invoke$0(AsyncExecutionInterceptor.java:115)
                at org.springframework.aop.interceptor.AsyncExecutionAspectSupport.lambda$doSubmit$3(AsyncExecutionAspectSupport.java:278)
                at java.base/java.util.concurrent.CompletableFuture$AsyncSupply.run(Unknown Source)
                at java.base/java.util.concurrent.ThreadPoolExecutor.runWorker(Unknown Source)
                at java.base/java.util.concurrent.ThreadPoolExecutor$Worker.run(Unknown Source)
                at java.base/java.lang.Thread.run(Unknown Source)
        Caused by: liquibase.exception.DatabaseException: java.lang.RuntimeException: Cannot end scope bsbacsmwja when currently at scope root
                at liquibase.executor.jvm.ChangelogJdbcMdcListener.query(ChangelogJdbcMdcListener.java:60)
                at liquibase.lockservice.StandardLockService.releaseLock(StandardLockService.java:372)
                ... 19 more
        Caused by: java.lang.RuntimeException: Cannot end scope bsbacsmwja when currently at scope root
                at liquibase.Scope.exit(Scope.java:231)
                at liquibase.Scope.child(Scope.java:195)
                at liquibase.Scope.child(Scope.java:169)
[...]

Issue #4314, opened 2023-05-26
Facing issue while trying to save the data concurrently for liquibase version 4.22.0


06Siddhi:
### Search first

- [X] I searched and no similar issues were found

### Description

Facing issue while trying to save the data concurrently like saving data for 30 user thread concurrently at a time, success ratio is 20, 17 or sometimes 15.
Facing this issue on version 4.22.0. I've moved to use Scope.setScopeManager(new ThreadLocalScopeManager()) but now see an error due to scopeId mismatch.

**Below is the error in the logs:
,"Caused By": { "liquibase.exception.LiquibaseException: java.lang.RuntimeException: Cannot end scope wxhajsndor when currently at scope root" : {"Liquibase.java:1610":"liquibase.Liquibase.runInScope","Liquibase.java:229":"liquibase.Liquibase.update","Liquibase.java:213":"liquibase.Liquibase.update,"Caused By": { "java.lang.RuntimeException: Cannot end scope wxhajsndor when currently at scope root" : {"Scope.java:241":"liquibase.Scope.exit","Scope.java:205":"liquibase.Scope.child","Scope.java:193":"liquibase.Scope.child","Scope.java:172":"liquibase.Scope.child","Liquibase.java:1605":"liquibase.Liquibase.runInScope"}}}}**

### Steps To Reproduce

Tried saving data using Jmeter for 30 data in database
I've already set Scope.setScopeManager(new ThreadLocalScopeManager()) inside the try block 

### Expected/Desired Behavior

Should be able to save all the data concurrently in database

### Liquibase Version

4.22.0

### Are you willing to submit a PR?

- [ ] I'm willing to submit a PR (Thank you!)

gombasp:
Same issue here. Upgraded from 4.3.1 to 4.22.0. We saw the same "cannot end scope..." issues and also the message saying that 4.22.0 has some issues. So we tried with 4.21.1 but the "cannot end scope" problem was still there. In our case it comes when we try to apply patches. As @06Siddhi also noted the Scope.setScopeManager(new ThreadLocalScopeManager()) does not seem to solve the issue. 
App is running on several pods in k8s on GCP. 
Please advise.

nvoxland:
We've been working on improving TreadLocalScopeManager lately. Sorry to hear you're still having problems with it.

Is either of you able to post a stacktrace (they are shown with `--log-level=info` ) ?

filipelautert:
[...]

Issue #4173, opened 2023-04-24
LockServiceFactory and ChangeLogHistoryServiceFactory cannot maintain state with concurrent use


jamey-clari:
### Search first

- [X] I searched and no similar issues were found

### Description

Even in Liquibase 4.20.0, both `LockServiceFactory` and `ChangeLogHistoryServiceFactory` maintain state in a way that is inherently broken during concurrent calls to `Liquibase.*` APIs.  Our JVMs communicate with numerous DBs and may concurrently attempt to run `Liquibase.update` on multiple (or even the same) DB depending on which threads are trying to access and initialize connection pools for.  This generally works okay BUT problems could arise.  Consider the following:

Thread A calls `Liquibase.update` to apply a changelog to a DB.
The update call acquires a `LockService` via `LockServiceFactory.getInstance().getLockService(database)`.

While Thread A is updating its DB Thread B makes the same `Liquibase.update` call for a different DB.
This update call acquires the different instance of `LockService` via `LockServiceFactory.getInstance().getLockService(database)` but both instances of `LockService` are now registered with the current instance of `LockServiceFactory` (`LockServiceFactory.instance`).

Whichever thread reaches the end of the `Liquibase.update` operation first will call `Liquibase.resetServices()` which calls `LockServiceFactory.getInstance().resetAll()` which sets `LockServiceFactory.instance` to null.  This means that if the second thread makes any call to `LockServiceFactory.getInstance().getLockService(database)` anywhere in the `Liquibase.update` call chain after this point it will get a new instance of `LockServiceFactory` and thus a new instance of `LockService`.

Perhaps the code has been audited to prevent this call pattern for `LockService` (i.e. `LockService`s are only obtained once in a given context and then passed as needed) but this seems dicey.

`ChangeLogHistoryServiceFactory` has exactly the same sort of state management mechanism and this I KNOW is used to obtain `ChangeLogHistoryService` instances at various points in the `Liquibase.update` code path and which will lose state if the `ChangeLogHistoryServiceFactory.instance` is replaced by another thread.

Can the `LockService` and `ChangeLogHistoryService` be bound to the Scope so that, at least using `ThreadLocalScopeManager`, they can be (more) thread-safe?

### Steps To Reproduce

    @Test
    public void testLiquibaseConcurrency() throws Exception
    {
        final ExecutorService exec = Executors.newSingleThreadExecutor();
        final CountDownLatch updateStarted = new CountDownLatch(1);
[...]

Issue #2433, opened 2022-01-28
Cannot end scope when currently at scope on HealthIndicator


vasicvuk:
## Environment

**Liquibase Version**: 3.6.0

**Liquibase Integration & Version**: Spring Boot 2.6.3, Maven

**Liquibase Extension(s) & Version**: 

**Database Vendor & Version**:
PostgreSQL 11.6

**Operating System Type & Version**:
Linux

## Description

When calling Liquibase inside of HealthIndicator (Spring boot) I get following error after few calls:

```
2022-01-28 08:45:39 [Error] Exception while checking health for database
liquibase.exception.LiquibaseException: java.lang.RuntimeException: Cannot end scope locsrufyjn when currently at scope qemsvrvsvc
liquibase.exception.LiquibaseException: java.lang.RuntimeException: Cannot end scope locsrufyjn when currently at scope qemsvrvsvc
at liquibase.Liquibase.runInScope(Liquibase.java:2409)
at liquibase.Liquibase.listUnrunChangeSets(Liquibase.java:1977)
at liquibase.Liquibase.listUnrunChangeSets(Liquibase.java:1968)
at com.asseco.bpm.rest.health.FlywayHealthIndicator.health(FlywayHealthIndicator.java:104)
at org.springframework.boot.actuate.health.HealthIndicator.getHealth(HealthIndicator.java:37)
at org.springframework.boot.actuate.health.HealthEndpointWebExtension.getHealth(HealthEndpointWebExtension.java:87)
at org.springframework.boot.actuate.health.HealthEndpointWebExtension.getHealth(HealthEndpointWebExtension.java:45)
at org.springframework.boot.actuate.health.HealthEndpointSupport.getContribution(HealthEndpointSupport.java:130)
at org.springframework.boot.actuate.health.HealthEndpointSupport.getAggregateContribution(HealthEndpointSupport.java:141)
at org.springframework.boot.actuate.health.HealthEndpointSupport.getContribution(HealthEndpointSupport.java:126)
at org.springframework.boot.actuate.health.HealthEndpointSupport.getHealth(HealthEndpointSupport.java:95)
at org.springframework.boot.actuate.health.HealthEndpointSupport.getHealth(HealthEndpointSupport.java:66)
at org.springframework.boot.actuate.health.HealthEndpointWebExtension.health(HealthEndpointWebExtension.java:73)
[...]

PR #5617, opened 2024-02-21
fix: uses ThreadLocal to prevent concurrent modification on ScopeManager

filipelautert:
## Impact

## Description

Issues  #5426 #2433 #4173  #4314  presents the same error:  "Cannot end scope bsbacsmwja when currently at scope root" . 
The issue was narrowed down to Scope class handling of ScopeManager:

1. Thread "A" starts and creates a root scope. By doing so, Scope.getCurrentScope() returns "root"
2. Thread "A" starts a child scope "ABC". Now, Scope.getCurrentScope() returns "ABC", and parent is root (parent is saved in a non-static variable, while currentScope is static)
3. Thread "A"  is suspended and Thread "B" starts
4. Thread "B" starts.  By doing so, Scope.getCurrentScope() returns "ABC" (**hey! that's already wrong**)
5.  Thread "B" starts a child scope "DEF". Now, Scope.getCurrentScope() returns "DEF" and parent is "ABC"
6. Thread "B"  is suspended, "A"  resumes
7. Thread "A" does it's job and ends the scope.  The last step of exit is restoring the former Scope, so thread "A" sets currentScope to root (remember, parent is local and current is static) .
8. Thread "B" resumes - now working on curentScope "root" . It does it work and tries to finish. It tries to compare scopeId that was saved in a local variable,  but then "B" gets message "Cannot end scope DEF (local variable)  when currently at scope root (statically changed) " - RuntimeException 

We prevent this issue by using a ThreadLocal object that ensures that each thread will have it's own ScopeManager , so "A" won't break "B" as it cannot mess with it's ScopeManager. 

This issue has always been around, but seems it started to show up more as child scopes are being used for blocking tasks (that allows threads switching) such as database execution in ChangelogJdbcMdcListener . In the example above, A and B were suspended waiting for a database query to run.

### DIFF
--- a/liquibase-standard/src/main/java/liquibase/Scope.java
+++ b/liquibase-standard/src/main/java/liquibase/Scope.java
@@ -70,7 +70,7 @@ public class Scope {
 
     public static final String JAVA_PROPERTIES = "javaProperties";
 
-    private static ScopeManager scopeManager;
+    private static final ThreadLocal<ScopeManager> scopeManager = new ThreadLocal<>();
 
     private final Scope parent;
     private final SmartMap values = new SmartMap();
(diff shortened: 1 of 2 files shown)


StevenMassaro (approved):
Nice job identifying the root cause here.

Is it possible to write a unit test for this? Seems like the steps you provided in the description should be pretty straightforward to write a test for.

suryaaki2 (approved):

rberezen (approved):
functional test failure is expected and will be fixed with merging new s3 extension

mches:
@filipelautert Nice fix. Would you suppose this makes `ThreadLocalScopeManager` effectively obsolete? Perhaps it should be deprecated?

filipelautert:
@mches makes sense as now we are always using it, and in a simpler way.

filipelautert:
@mches https://github.com/liquibase/liquibase/pull/5780 , thanks for the reminder!

PR #5626, opened 2024-02-23
Add support for SnakeYaml to be able to parse timestamps correctly when timezone is set or not

### LINKED ISSUE #5499: System timezone changes generated checksums for YAML change logs that doesn't specify timezone for TIMESTAMP columns
### Search first

### Description

If a YAML change log  contains an insert for a TIMESTAMP column but doesn't specify the timezone, Liquibase generates different checksums when  system timezone changes.

### Steps To Reproduce

1. Create a change log which contains an TIMESTAMP column insert but doesn't specify timezone
2. Run change log 
3. Change system timezone to something other than current value
4. Run change log once more
5. Change log fails due to different checksums

### Expected/Desired Behavior

System timezone have no effect on Liquibase checksums when timezone is not specified in YAML change logs

### Liquibase Version

4.23.0

### Database Vendor & Version

PostgresSQL 9.6.17

### Liquibase Integration

spring boot

### Liquibase Extensions

_No response_

VM

### Additional Context
under the hoods snakeyaml uses system timezone as default when timezone is missing. 

Example change log 
```yaml
databaseChangeLog:
  - changeSet:
      id: 0.0.1
      author: johndoe
      changes:
      - createTable:
          tableName: example
          columns:
            - column:
                name: id
                type: VARCHAR(255)
                constraints:
                  primarykey: true
                  unique: true
                  nullable: false
            - column:
                name: created_at
                type: TIMESTAMP(3)
                constraints:
                  nullable: false

      - insert:
          tableName: example
          columns:
            - column:
                name: id
                value: example1
            - column:
                name: created_at
                value: 2018-03-09 08:41:31.000
```

MalloD12:
## Impact

## Description

A new ConstructYamlTimestamp class was created to identify whether a date time should be returned with the timezone included or not.

Fixes #5499 

## Things to be aware of
- None

## Things to worry about
- None

## Additional Context
N/A


liquibase-standard/src/main/java/liquibase/parser/core/yaml/CustomConstructYamlTimestamp.java
StevenMassaro, line 1: Can you point out what part of this class is different from the class it is extending?

MalloD12: Sure, the updated/added parts from the original code are these two if blocks:

```
if (timezoneh_s != null) {
                String time = timezonem_s != null ? ":" + timezonem_s : "00";
                timeZone = TimeZone.getTimeZone("GMT" + timezoneh_s + time);
                calendar = Calendar.getInstance(timeZone);
                ;
            } else {
                calendar = Calendar.getInstance();
            }
```
and 
```
 if (timeZone == null) {
                return new java.text.SimpleDateFormat("yyyy-MM-dd HH:mm:ss").format(calendar.getTime());
            } else {
                return calendar.getTime();
            }
```

StevenMassaro: Is this a change that would be beneficial to make to SnakeYaml itself? I think that if possible, that would be more preferable. I am concerned that as the SnakeYaml class we're extending changes, we won't get the changes ourselves because we override so much of that class.

MalloD12: Steve, do you mean we should create an issue or request the SnakeYaml team to make this change we have done here? I think it's true if they make any sort of change to the `ConstructYamlTimestamp` we won't get it as we override it, although I'm not sure if that could be any time soon. If you guys agree I can create an issue for this change on their end, but let's move ahead with this change and once they fix it on their end we can revert this change. How does it sound to you?

StevenMassaro: Sounds reasonable. I was suggesting that we make a PR on their repo, not an issue, since we know what the change should be.

MalloD12: I am not able to create a PR on their repo, but I created this [issue](https://bitbucket.org/snakeyaml/snakeyaml/issues/1085/parsed-datetime-always-contains-timezone) for them with the proposed changes we added in this PR. In the meantime, as mentioned above I think we can move forward with this PR and revert these changes once they apply the fix for it (if they do).

Thanks,
Daniel.

StevenMassaro: Thanks Daniel!!

StevenMassaro (approved):

liquibase-standard/src/main/java/liquibase/parser/core/yaml/CustomConstructYamlTimestamp.java
```diff
@@ -0,0 +1,83 @@
...
+        ScalarNode scalar = (ScalarNode) node;
+        String nodeValue = scalar.getValue();
+        Matcher match = YMD_REGEXP.matcher(nodeValue);
+        if (match.matches()) {
+            String year_s = match.group(1);
+            String month_s = match.group(2);
+            String day_s = match.group(3);
+            calendar = Calendar.getInstance(TimeZone.getTimeZone("UTC"));
+            calendar.clear();
+            calendar.set(Calendar.YEAR, Integer.parseInt(year_s));
```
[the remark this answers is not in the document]
filipelautert: @MalloD12 - could you review those codeql alerts?

filipelautert (changes requested):

MalloD12:
Hey @filipelautert,

This is pretty much all snakeyaml code. I don't think it's worth spending time on it, since we are going to revert/update this code soon (when they get their code updated).

Thanks,
Daniel.

filipelautert (approved):

rberezen (approved):
Test failure is expected and will be fixed after sync with master. Tested manually, LGTM

PR #5653, opened 2024-03-01
Deployment ID enhancements and execution property

jasonlyle88:
## Impact

## Description

This PR is created in response to challenges encountered in the creation development of #5598.

The purpose of this PR is to create a deployment ID that does have to be changed/generated multiple times and is available early on in the execution timeline.

This will allow the implementation of an execution property as setup in #5598 to be created for the deployment ID.

However, the deployment ID was generated later in the process to be a time-based number that was obtained after a database lock was obtained to guarantee uniqueness. This PR aims to change the deployment ID from a time-based number to a UUIDv4 to ensure uniqueness. The UUID (36 characters) does not fit in the existing databasechangelog.deployment_id column, so the column is updated in this PR to be a length of 36 (and is updated for both creation of the table and update of existing tables).

For a full discussion of why this came to be, the discussion on #5598 can be referenced.

## Things to be aware of

- Decided to use UUIDv4 to guarantee uniqueness as this is a universally accepted solution to this problem.
- The databasechangelog table is updated in this PR (just the length of an existing column being expanded).
- This PR leverages the changes from #5617 that guarantee SCOPE to be thread safe.
- I went through and updated references to the now deprecated AbstractChangeLogHistoryService deployment ID methods throughout the code

## Additional Context

This is created in draft because this PR is blocked until #5598 is merged.


jasonlyle88:
@filipelautert Obviously this is early, since it will need the changes for #5598 to do the execution property part, but I believe this implements the UUID changes we discussed. If you could take a look and let me know what you think, I would appreciate it! Thanks!

jasonlyle88:
@filipelautert This PR is now ready since #5598 was merged. Let me know what you think!

jasonlyle88:
@filipelautert This PR has been ready for 3 months but I haven't heard anything. Is there something going on here?

@MalloD12 I know I've worked with you recently and had you respond, not sure if @filipelautert isn't available?

MalloD12:
Hi @jasonlyle88,

Correct, Filipe is actually on leave (he'll be returning soon).  I can have a look at this, but it could probably be around the end of this week or next week. I think for some reason it gets lost in our backlog. I'll talk about it tomorrow with the team on our community planning, but we will be looking at it very soon. 

Thanks Jason!

Daniel.

MalloD12 (approved):
**Approved**.

Code changes look good for us. We have discussed about this PR with Filipe and we think this is not a breaking change and it is ok to move on. Some tests were fixed because scope id was returned and we were not expecting it in some assertions.

Thanks again for another PR, Jason!

Daniel.

liquibase-standard/src/main/java/liquibase/Scope.java
```diff
@@ -118,6 +119,7 @@ public static Scope getCurrentScope() {
 
             rootScope.values.put(Attr.serviceLocator.name(), serviceLocator);
             rootScope.values.put(Attr.osgiPlatform.name(), ContainerChecker.isOsgiPlatform());
+            rootScope.values.put(Attr.deploymentId.name(), UUID.randomUUID().toString());
```
filipelautert, line 122: UUID, nice!

liquibase-standard/src/main/java/liquibase/changelog/StandardChangeLogHistoryService.java
```diff
@@ -234,15 +234,34 @@ public void init() throws DatabaseException {
                     getLabelsSize() + ")", null));
             }
 
+            boolean deploymentIdColumnNotRightSize = false;
             if (!hasDeploymentIdColumn) {
                 executor.comment("Adding missing databasechangelog.deployment_id column");
                 statementsToExecute.add(new AddColumnStatement(getLiquibaseCatalogName(), getLiquibaseSchemaName(),
-                    getDatabaseChangeLogTableName(), "DEPLOYMENT_ID", "VARCHAR(10)", null));
+                    getDatabaseChangeLogTableName(), "DEPLOYMENT_ID", charTypeName + "(36)", null));
```
filipelautert, line 241: Just a note: `charTypeName` actually is "`VARCHAR`" too.

filipelautert (approved):

rberezen (approved):

tati-qalified:
Hi @jasonlyle88, there is a problem with checksums and previously deployed changesets for the .sql format. Here are the steps to reproduce:
1. With an earlier Liquibase version, run `liquibase update` using a sql changelog. I used 4.30.0.
2. Change to this PR's build and run `liquibase update` again.

Checksums should be recognized as already run so no changesets should be executed. Instead, there will be a checksum difference, for example:
```
     1 changesets check sum
          changes/changelog.sql::2::nvoxland was: 9:929eaf28de0edb099433043592c3d424 but is now: 9:03ebbf5914f26b447a1a16595a96221b
```

This is only for previously deployed changesets, and only for sql formats.

Could you look into it?
Thanks!

tati-qalified (approved):

PR #5658, opened 2024-03-05
User lombok Getter and Setter.

asashour:
This is a step to use lombok Getter and Setter, as hinted in another PR.

Not all potential changes are done, because the PR would be huge. Please advise if this is fine, or project-wide changes should also be added.


MalloD12 (approved):
**Approved**.

This looks ok to me. If you are planning to keep doing these changes for some other files I would recommend you to go in some specific order, maybe in an upcoming PR to keep adding Lombok annotations to the remaining `liquibase-standard` module files, and once finish with this module (which is the bigger) then continue with others.

Thanks,
Daniel.

filipelautert (approved):

rberezen (approved):

PR #5660, opened 2024-03-05
Default implementation of handleInvalidEmptyPreconditionCase added in AbtractFormattedChangeLogParser to avoid breaking extensions

MalloD12:
## Impact

## Description

Recently with the merge of this [PR](https://github.com/liquibase/liquibase/pull/5456/) the team has noticed this is breaking extensions, so just move the code implementation of `handleInvalidEmptyPreconditionCase` to the abstract class which should fix this issue if no extension overrides this method.

### DIFF
--- a/liquibase-standard/src/main/java/liquibase/parser/AbstractFormattedChangeLogParser.java
+++ b/liquibase-standard/src/main/java/liquibase/parser/AbstractFormattedChangeLogParser.java
@@ -14,6 +14,7 @@ import liquibase.exception.ChangeLogParseException;
 import liquibase.resource.ResourceAccessor;
 import liquibase.util.StreamUtil;
 import liquibase.util.StringUtil;
+import org.apache.commons.lang3.NotImplementedException;
 
 import java.io.BufferedReader;
 import java.io.IOException;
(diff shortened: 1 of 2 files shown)


vitaliimak:
Thanks, @MalloD12 for such a quick fix!
It should work, but, I'm wondering if it is possible to leave the abstract class without SQL-specific implementation. The commercial MongoDB extension does not use `sql-check`, `table-exists`, and so on.
Is it possible to write something like this?
```
protected void handleInvalidEmptyPreconditionCase(ChangeLogParameters changeLogParameters, ChangeSet changeSet, Matcher preconditionMatcher) throws ChangeLogParseException {
    throw new NotImplementedException();
}
```
And left SQL-specific code in the `FormattedSqlChangeLogParser` class.

StevenMassaro (approved):

MalloD12:
> protected void handleInvalidEmptyPreconditionCase(ChangeLogParameters changeLogParameters, ChangeSet changeSet, Matcher preconditionMatcher) throws ChangeLogParseException {
>     throw new NotImplementedException();
> }

Ohh, I see. I get it now. Sure, I can do that @vitaliimak.

vitaliimak (approved):

liquibase-standard/src/main/java/liquibase/parser/AbstractFormattedChangeLogParser.java
```diff
@@ -610,6 +609,10 @@ protected InputStream openChangeLogFile(String physicalChangeLogLocation, Resour
         return resourceAccessor.getExisting(physicalChangeLogLocation).openInputStream();
     }
 
+    protected void handleInvalidEmptyPreconditionCase(ChangeLogParameters changeLogParameters, ChangeSet changeSet, Matcher preconditionMatcher) throws ChangeLogParseException {
+        throw new NotImplementedException();
```
filipelautert, line 613: Please add a message explaining what we reached, like: 
```
throw new NotImplementedException("Invalid empty precondition found");
```

filipelautert (changes requested):

liquibase-standard/src/main/java/liquibase/parser/AbstractFormattedChangeLogParser.java
MalloD12: Done

filipelautert (approved):

PR #4628, opened 2023-08-08
DAT -15524 Clear-Checksums implementation using command scope

MalloD12:
## Impact

- [ ] Bug fix (non-breaking change which fixes expected existing functionality)
- [x] Enhancement/New feature (adds functionality without impacting existing logic)
- [ ] Breaking change (fix or feature that would cause existing functionality to change)
 

## Description

- Implementation of `clearChecksums()` using command scope (part of Command Refactoring implemented recently).
- Deprecated `clearChecksums()` from Liquibase class.
- Integration test added to validate MD5Sum column is set to `NULL` after `clear-checksums` command is executed.

Fixes #4456 

## Things to be aware of
- None

## Things to worry about
- None

## Additional Context
- Nothing

PR #5665, opened 2024-03-07
fix: ClearChecksums should not throw Exception in empty databases

### LINKED ISSUE #5319: When spring.liquibase.clear-checksums property of spring boot is set, when the target database is empty, it throws exception, and SpringLiquibase.afterPropertiesSet doesn't run the changelogs
### Search first

### Description

So I want to submit two bugs, but didn't want to open two issues because two can be reproduced from the same repo that I will provide

### Steps To Reproduce

First bug: When `spring.liquibase.clear-chekcsums` is set to true, initializing on an empty database throws exception in spring boot application on application startup
Steps:
1. Clone repo: https://github.com/fatihdogmus/afterpropertiesset-bug
2. Run `docker compose up -d`. This will initialize a bitnami alpine 15.4 postgres image and bind the data to ~/DockerVolumes/liquibase-bug, since we need persistent storage for the next bug
3. Run the application
4. You will see exception: `encountered during context initialization - cancelling refresh attempt: org.springframework.beans.factory.BeanCreationException: Error creating bean with name 'liquibase' defined in class path resource [org/springframework/boot/autoconfigure/liquibase/LiquibaseAutoConfiguration$LiquibaseConfiguration.class]: liquibase.exception.DatabaseException: liquibase.exception.DatabaseException: ERROR: relation "public.databasechangelog" does not exist
  Position: 8 [Failed SQL: (0) UPDATE public.databasechangelog SET MD5SUM = NULL]`

Second bug: This is a bit trickier. After DB initialization, when we want to reset the database programatically with `SpringLiquibase`, by first dropping everything, the changes are not run when `spring.liquibase.clear-chekcsums` is set to true

Steps:
1. Clone the same repo and get DB working with docker. I tried with empty database, so it would be wise to remove anything from previous DB.
2. First, set `spring.liquibase.clear-chekcsums` to false, otherwise the application doesn't start
3. Stop the application, set `spring.liquibase.clear-chekcsums` to true and run the application. After this, the DB is in correct state.
4. Send a GET request to `http://localhost:8087/resetDatabase`, which is this code:
```java
@RestController
public class ResetDatabaseController {

    private final SpringLiquibase springLiquibase;

    public ResetDatabaseController(final SpringLiquibase springLiquibase) {
        this.springLiquibase = springLiquibase;
    }

    @GetMapping("/resetDatabase")
    public String resetDatabase() throws LiquibaseException {
        springLiquibase.setClearCheckSums(false);
        springLiquibase.setDropFirst(true);

        springLiquibase.afterPropertiesSet();

        springLiquibase.setClearCheckSums(true);
        springLiquibase.setDropFirst(false);

        return "Database has been reset successfully";
    }
}

```
5. You will see Database is up to date, no changesets to execute command in the logs, after all the objects are dropped. 
 And the database won't be updated.

Strangely, this bug doesn't happen if the application is started with `spring.liquibase.clear-chekcsums` set to false. I don't know what the problem is, I set it false on my code anyway so I don't know why it would behave differently.

Both problems didn't happen before we migrated to spring boot 3.2.0 from 3.1.4

### Expected/Desired Behavior

For first bug to not throw exceptions and second bug to run the changes.

### Liquibase Version

4.24.0

### Database Vendor & Version

PostgreSQL 15.4

### Liquibase Integration

Spring Boot

### Liquibase Extensions

_No response_

Macos

### Additional Context

_No response_

I can provide a PR, but I don't know what the problem is. I've looked at the source code, but couldn't find the problem, since I'm not very familiar with the code base. If you can find the problem and direct me to it, I can submit a PR.

filipelautert:
## Impact

After https://github.com/liquibase/liquibase/pull/4628 clearChecksums started to raise exceptions in empty databases. This PR fixes this behavior.

Fixes #5319

### DIFF
--- a/liquibase-standard/src/main/java/liquibase/command/core/ClearChecksumsCommandStep.java
+++ b/liquibase-standard/src/main/java/liquibase/command/core/ClearChecksumsCommandStep.java
@@ -3,7 +3,10 @@ package liquibase.command.core;
 import liquibase.Scope;
 import liquibase.changelog.ChangeLogHistoryService;
 import liquibase.changelog.ChangeLogHistoryServiceFactory;
-import liquibase.command.*;
+import liquibase.command.AbstractCommandStep;
+import liquibase.command.CommandDefinition;
+import liquibase.command.CommandResultsBuilder;
+import liquibase.command.CommandScope;
 import liquibase.database.Database;
 import liquibase.lockservice.LockService;
 
@@ -31,9 +34,10 @@ public class ClearChecksumsCommandStep extends AbstractCommandStep {
         CommandScope commandScope = resultsBuilder.getCommandScope();
         final Database database = (Database) commandScope.getDependency(Database.class);
 
-        ChangeLogHistoryService changeLogService = ChangeLogHistoryServiceFactory.getInstance().getChangeLogService(database);
+        ChangeLogHistoryService changeLogHistoryService = Scope.getCurrentScope().getSingleton(ChangeLogHistoryServiceFactory.class).getChangeLogService(database);
+        changeLogHistoryService.init();
         Scope.getCurrentScope().getLog(getClass()).info(String.format("Clearing database change log checksums for database %s", database.getShortName()));
-        changeLogService.clearAllCheckSums();
+        changeLogHistoryService.clearAllCheckSums();
     }
 
     @Override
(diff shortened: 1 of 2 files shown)


liquibase-standard/src/main/java/liquibase/command/core/ClearChecksumsCommandStep.java
```diff
@@ -31,9 +34,10 @@ public void run(CommandResultsBuilder resultsBuilder) throws Exception {
         CommandScope commandScope = resultsBuilder.getCommandScope();
         final Database database = (Database) commandScope.getDependency(Database.class);
 
-        ChangeLogHistoryService changeLogService = ChangeLogHistoryServiceFactory.getInstance().getChangeLogService(database);
+        ChangeLogHistoryService changeLogHistoryService = Scope.getCurrentScope().getSingleton(ChangeLogHistoryServiceFactory.class).getChangeLogService(database);
+        changeLogHistoryService.init();
```
filipelautert, line 38: Init was missing.

rberezen (approved):

PR #5668, opened 2024-03-07
load driver from global argument as well as command argument (DAT-16381)

StevenMassaro:
## Impact

### DIFF
--- a/liquibase-standard/src/main/java/liquibase/command/core/helpers/DbUrlConnectionCommandStep.java
+++ b/liquibase-standard/src/main/java/liquibase/command/core/helpers/DbUrlConnectionCommandStep.java
@@ -7,9 +7,11 @@ import liquibase.command.CleanUpCommandStep;
 import liquibase.command.CommandArgumentDefinition;
 import liquibase.command.CommandResultsBuilder;
 import liquibase.command.CommandScope;
+import liquibase.configuration.ConfiguredValue;
 import liquibase.database.Database;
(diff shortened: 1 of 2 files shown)


suryaaki2 (approved):

filipelautert (approved):

StevenMassaro:
@filipelautert and @suryaaki2 

I am rerequested your review on this PR because I changed the type of the `driver` arg from `Class` to `String`. I don't consider this a breaking change for two reasons. First, this argument wasn't being used in Liquibase until this PR fixed that. Second, `String` should allow more compatibility than `Class`. This was the motivation for the change, because regression tests were failing with a class not found exception when the type was `Class`.

StevenMassaro:
@filipelautert @suryaaki2 

Notably though, this still _could be_ a breaking change if someone had an extension which was expecting this parameter to be of type `Driver`.

I want your opinions on whether this change is permissible and if not, we can discuss how to avoid making this a breaking change.

filipelautert:
> Notably though, this still _could be_ a breaking change if someone had an extension which was expecting this parameter to be of type `Driver`.
> 
> I want your opinions on whether this change is permissible and if not, we can discuss how to avoid making this a breaking change.

Hopefully they are not using it - the only mentions of driver that I remember in extensions are in  `extends Database` class . And  as now we have the nightly builds for extensions, the easiest way to test it would be to merge and watch the nightly builds for fails.

filipelautert (approved):

PR #5670, opened 2024-03-07
5507-performance_improvements

### LINKED ISSUE #5507: Performance degradation when switching from v3.6.3 to v4.25.1
### Description

Using the same changelogs and database/schema liquibase after upgrade to version 4.25.1 is a few times slower, e.g. 33 minutes instead of 8 minutes.

* First I waited for checksums upgrade, then I executed liquibase once again without any new migrations in version 4.25.1 to compare performance between liquibase versions.
* In my setup multiple Liquibase objects are executed in parallel migrating different DB schemas and DB hosts.
* ThreadLocalScopeManager is registered before starting migration.

I managed to profile liquibase 4.25.1 and I have two observations:
* There are a lot of calls to AbstractMapConfigurationValueProvider#getProvidedValue. Threads are waiting on a lock. In this critical section, among other things, EnvironmentValueProvider is called and it looks like computing hash of its map may be time consuming, which may be relevant in the context of multiple other threads waiting.
* There are child scopes spawned in high numbers. Scope initialization takes time, mainly scope id generation.

Could you take a look at this problem?

### Expected/Desired Behavior

No performance degradation.

### Liquibase Version

3.6.3 => 4.25.1

### Database Vendor & Version

MySQL

### Liquibase Integration

Called directly, programmatically. 

### Liquibase Extensions

-

Linux

filipelautert:
## Impact

## Description

feat: use commons-lang3 library to improve random string generation by ~30% .

partially fix #5507

### DIFF
--- a/liquibase-standard/src/main/java/liquibase/util/StringUtil.java
+++ b/liquibase-standard/src/main/java/liquibase/util/StringUtil.java
@@ -6,6 +6,7 @@ import liquibase.Scope;
 import liquibase.changelog.ChangeSet;
 import liquibase.parser.LiquibaseSqlParser;
 import liquibase.parser.SqlParserFactory;
+import org.apache.commons.lang3.RandomStringUtils;
 
 import java.io.UnsupportedEncodingException;
 import java.nio.charset.StandardCharsets;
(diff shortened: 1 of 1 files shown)


rberezen (approved):

MalloD12 (approved):
LGTM

PR #5686, opened 2024-03-13
add ignore property to includeAll XSD (DAT-14921)

StevenMassaro:
## Impact

## Description

The [ignore code has existed for a long time for includeAll](https://github.com/liquibase/liquibase/blob/master/liquibase-standard/src/main/java/liquibase/changelog/DatabaseChangeLog.java#L504), but it was not part of the XSD.


abrackx (approved):
Should we add an xml test for this or do you think it's appropriately covered already?

StevenMassaro:
> Should we add an xml test for this or do you think it's appropriately covered already?

I can add a test, good idea.

suryaaki2 (approved):

PR #5687, opened 2024-03-13
fix: addColumn changes are not respecting Datatypes definitions

filipelautert:
## Impact

## Description

When running a diffChangelog, a create table the column uses the Datatypes classes to define the column types, but an addColumn change would simply dump the database text.

This PR fixes addColumn by making it behave as createTable.

### DIFF
--- a/liquibase-standard/src/main/java/liquibase/diff/output/changelog/core/MissingColumnChangeGenerator.java
+++ b/liquibase-standard/src/main/java/liquibase/diff/output/changelog/core/MissingColumnChangeGenerator.java
@@ -5,6 +5,9 @@ import liquibase.change.Change;
 import liquibase.change.ConstraintsConfig;
 import liquibase.change.core.AddColumnChange;
 import liquibase.database.Database;
+import liquibase.datatype.DataTypeFactory;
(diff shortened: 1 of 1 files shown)


liquibase-standard/src/main/java/liquibase/diff/output/changelog/core/MissingColumnChangeGenerator.java
```diff
@@ -63,9 +66,10 @@ public Change[] fixMissing(DatabaseObject missingObject, DiffOutputControl contr
         AddColumnConfig columnConfig = createAddColumnConfig();
         columnConfig.setName(column.getName());
 
-        String dataType = column.getType().toString();
-
-        columnConfig.setType(dataType);
+        LiquibaseDataType ldt = DataTypeFactory.getInstance().from(column.getType(), referenceDatabase);
```
MalloD12, line 69: Would it be better to extract this updated code into a new method, so then we can unit-test it? Otherwise, we can only test just the whole thing?

rberezen (approved):

MalloD12 (approved):

PR #5707, opened 2024-03-18
[DAT-16135] add new scope flag for modifyChangeSet flow

KushnirykOleh:
## Impact

## Description
For some parsers, (like mongosh parsers) we need to know when modifyChangeSet flow is happening, as we can't run mongo instructions without native executor so `runWith` is required. But in modifyChangeSet flow `runWith` is added to the changeset later on, after parsing. 
Adding modifyChangeSets flag to the Scope allows parsers to read it and skip runWith validation knowing it's inner changeset and it gets 'runWith'  attribute later.

### DIFF
--- a/liquibase-standard/src/main/java/liquibase/changelog/DatabaseChangeLog.java
+++ b/liquibase-standard/src/main/java/liquibase/changelog/DatabaseChangeLog.java
@@ -31,6 +31,7 @@ import liquibase.resource.ResourceAccessor;
 import liquibase.servicelocator.LiquibaseService;
 import liquibase.util.FileUtil;
 import liquibase.util.StringUtil;
+import org.apache.commons.collections4.map.SingletonMap;
 
 import java.io.IOException;
 import java.io.InputStream;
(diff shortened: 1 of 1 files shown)


wwillard7800 (approved):
Looks good.

liquibase-standard/src/main/java/liquibase/changelog/DatabaseChangeLog.java
```diff
@@ -924,11 +924,12 @@ public boolean include(String fileName,
         if (isRelativePath) {
             try {
                 fileName = resourceAccessor.get(this.getPhysicalFilePath()).resolveSibling(fileName).getPath();
-                fileName = normalizePath(Paths.get(fileName).toString());
             } catch (IOException e) {
                 throw new UnexpectedLiquibaseException(e);
             }
         }
+        final String normalizedFilePath = normalizePath(Paths.get(fileName).toString());
```
StevenMassaro, line 931: Is there a reason you moved this from its original location?

KushnirykOleh: had to make it final to pass it to the runner

StevenMassaro: As long as moving it outside of the `if(isrelativePath)` block is okay, then I am fine with this.

StevenMassaro (approved):

filipelautert (approved):

PR #5735, opened 2024-03-26
Adds CommandOverrideProcessor to validate CommandOverride usage at compile time

abrackx:
## Impact

## Description
Adds new annotation processor to detect invalid usage of the `@CommandOverride` annotation. 
TODO: 
- [ ] Do not print multiple messages, this means the annotation processor is running more than once
- [ ] Re-use code, this is currently duplicating the validation found in `CommandFactory`. 
- [ ] Validate that the new service file is actually needed
- [ ] Get this to build in the pipeline

## Things to be aware of
This will (intentionally) break builds. 

## Things to worry about
This can impact lombok. I don't think it does in its current form, but it can.

### DIFF
--- a/liquibase-standard/pom.xml
+++ b/liquibase-standard/pom.xml
@@ -170,6 +170,25 @@
 
         <plugins>
             <plugin>
+                <groupId>org.apache.maven.plugins</groupId>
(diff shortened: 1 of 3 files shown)


filipelautert:
@abrackx can we close this PR?

PR #5766, opened 2024-04-04
Use lombok getter and setter.

asashour:
## Impact

## Description
Similar to #5658, this PR users Lombok getters and setters.

## Things to be aware of

All changes were done automatically by IntelliJ.

Only the methods without JavaDoc were included, as it seems there are bugs in IntelliJ to handle `@param` or `@return`, or for overloaded setters, which would be reported to JetBrains.

An exception is the removal of the JavaDoc of CreateIndexChange.[unique](https://github.com/liquibase/liquibase/blob/master/liquibase-standard/src/main/java/liquibase/change/core/CreateIndexChange.java#L176-L178), as I think it doesn't add much value.


MalloD12 (approved):
LGTM.

Build and tests have been successfully executed.

filipelautert (approved):

rberezen (approved):

PR #5767, opened 2024-04-04
Ensure a new-line at the end of text files.

asashour:
## Impact

## Description

Text files should end by a new-line, as per [POXIS standard](https://stackoverflow.com/a/729795/184201).


MalloD12 (approved):
**Approved**.

LGTM. I only had to update a test assertion from a test that was failing because of the addition of the new line added in one test resource used.

Other than that, I think we are good to go with this as I think functional test failures are not related to these changes.

filipelautert (approved):

php-coder:
BTW in order to prevent such issues in future, I'd suggest to add `.editorconfig` file with `insert_final_newline = true` directive. See https://editorconfig.org for details.

PR #5777, opened 2024-04-07
Prevent error when unlocking the changelog in MySQL affects zero rows and useAffectedRows is true. Fixes #5502

### LINKED ISSUE #5502: When  append useAffectedRows=true to the jdbc URL, liquibase cannot update the changeloglockdatabasechangeloglock properly
### Search first

### Description

I am using a MySQL database，when  append useAffectedRows=true to the jdbc URL, liquibase cannot update the changeloglockdatabasechangeloglock properly.

0 rows were updated instead of the expected 1 row using executor liquibase.executor.jvm.JdbcExecutor there are [{COUNT(*)=1}] rows in the table

### Steps To Reproduce

append eAffectedRows=true to the jdbc URL.
spring.datasource.url=jdbc:mysql://ip:3306/your_db_name?allowPublicKeyRetrieval=true&useAffectedRows=true

### Expected/Desired Behavior

Liquibase can execute normally, regardless of whether there is a new changeset

### Liquibase Version

4.24

### Database Vendor & Version

mysql 8.3

### Liquibase Integration

Maven springboot3.2.2

### Liquibase Extensions

_No response_

Windows10

### Additional Context

_No response_

mpvvliet:
Fixes #5502 

## Impact

## Description

Whenever the JDBC driver for MySQL is configured with `useAffectedRows` and the changeset being processed has been processed before, the changelog unlocking mechanism returns zero rows affected and causes an error. This fix detects this situation and prevents an error from being raised.

## Things to be aware of

Only relevant for MySQL.

## Things to worry about

Any other scenario that also causes the unlocking mechanism to return zero rows affected will be treated as a success. I am unaware of any scenarios breaking this change but maybe I'm overlooking something.


liquibase-standard/src/main/java/liquibase/database/jvm/JdbcConnection.java
```diff
@@ -288,6 +288,10 @@ public boolean getAutoCommit() throws DatabaseException {
         }
     }
 
+    public boolean getUseAffectedRows() throws DatabaseException {
+        return getURL().contains("useAffectedRows=true");
```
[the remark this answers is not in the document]
mpvvliet: Makes sense since it only applies there. I've updated the code.

liquibase-standard/src/main/java/liquibase/database/core/MySQLDatabase.java
```diff
@@ -678,4 +678,9 @@ public boolean supportsCreateIfNotExists(Class<? extends DatabaseObject> type) {
     public boolean supportsDatabaseChangeLogHistory() {
         return true;
     }
+
+    public boolean getUseAffectedRows() throws DatabaseException {
```
filipelautert, line 682: @mpvvliet could you add the `@Override` annotation here? codeql is complaining about it.

filipelautert (approved):
Thanks for the @mpvvliet !

liquibase-standard/src/main/java/liquibase/database/core/MySQLDatabase.java
mpvvliet: I think you are referring to an old codeql scan, this method method does not override any method. 🤔

filipelautert: haha true, it's gone now. Sorry!

rberezen (approved):

PR #5781, opened 2024-04-09
Fix (MSSQL): default end delimiter applied when there is an endDelimiter applied for a given chanset

### LINKED ISSUE #5603: Semicolon added to end of SQL Server stored procedure when running.
### Search first

### Description

this started in  

### Steps To Reproduce

The zip contains a trivial example of a single stored procedure, a changelog.json containing that file, and the powershell script to deploy the changelog to a SQL Server database
[LBSemicolonError.zip](https://github.com/liquibase/liquibase/files/14319068/LBSemicolonError.zip)

You will have to change the powershell script to include a valid username, password, database name.

### Expected/Desired Behavior

The script looks like the following, there is no semicolon defined at the end. 

```
-- liquibase formatted sql
--changeset mtagliaf:procedures.App.TestProcedure.20240129 runOnChange:true stripComments:false splitStatements:true endDelimiter:;;;;;
create or alter procedure App.TestProcedure
as
    select top 5 *
        from
            dbo.CustomerMaster
        where
            TerritoryID = 2
            and IsActive = 1
;;;;;
```
When the script is deployed, there is a semicolon added at the end of the last line in the SQL Server definition of the procedure

### Liquibase Version

4.26

### Database Vendor & Version

SQL Server 2019

### Liquibase Integration

CLI

### Liquibase Extensions

_No response_

Windows 11

### Additional Context

This is a problem with our initial deployment because when we deploy our scripts to an empty database to compare with our production database, the added semicolon means that every view/procedure/function comes up as different from production.

MalloD12:
## Impact

## Description

When `endDelimiter` is specified in a changeset no default endDelimiter should be added to the SQL visitors to be applied. 

Fixes #5603
 

## Things to be aware of
- `AppendSqlIfNotPresentVisitor` will be only applied when there is no endDelimiter set for a changeset. 

## Things to worry about
- Not really, this is only applicable for MSSQL and it will only be applied if there is not endDelimiter specified.

## Additional Context
- None

### DIFF
--- a/liquibase-standard/src/main/java/liquibase/database/core/MSSQLDatabase.java
+++ b/liquibase-standard/src/main/java/liquibase/database/core/MSSQLDatabase.java
@@ -3,6 +3,7 @@ package liquibase.database.core;
 import liquibase.CatalogAndSchema;
 import liquibase.GlobalConfiguration;
 import liquibase.Scope;
+import liquibase.change.AbstractSQLChange;
 import liquibase.change.Change;
 import liquibase.changelog.DatabaseChangeLog;
 import liquibase.database.AbstractJdbcDatabase;
@@ -114,7 +115,12 @@ public class MSSQLDatabase extends AbstractJdbcDatabase {
 
     @Override
     public void executeStatements(Change change, DatabaseChangeLog changeLog, List<SqlVisitor> sqlVisitors) throws LiquibaseException {
-        super.executeStatements(change, changeLog, addSqlVisitors(sqlVisitors));
+        if (change instanceof AbstractSQLChange && StringUtil.trimToNull(((AbstractSQLChange) change).getEndDelimiter()) != null) {
+            super.executeStatements(change, changeLog, sqlVisitors);
+        }
+        else {
+            super.executeStatements(change, changeLog, addSqlVisitors(sqlVisitors));
+        }
     }
 
     //
--- a/liquibase-standard/src/main/java/liquibase/database/core/MSSQLDatabase.java
+++ b/liquibase-standard/src/main/java/liquibase/database/core/MSSQLDatabase.java
@@ -115,7 +115,8 @@ public class MSSQLDatabase extends AbstractJdbcDatabase {
 
     @Override
     public void executeStatements(Change change, DatabaseChangeLog changeLog, List<SqlVisitor> sqlVisitors) throws LiquibaseException {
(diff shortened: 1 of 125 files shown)


liquibase-standard/src/main/java/liquibase/database/core/MSSQLDatabase.java
```diff
@@ -114,7 +115,12 @@ public MSSQLDatabase() {
 
     @Override
     public void executeStatements(Change change, DatabaseChangeLog changeLog, List<SqlVisitor> sqlVisitors) throws LiquibaseException {
-        super.executeStatements(change, changeLog, addSqlVisitors(sqlVisitors));
+        if (change instanceof AbstractSQLChange && StringUtil.trimToNull(((AbstractSQLChange) change).getEndDelimiter()) != null) {
+            super.executeStatements(change, changeLog, sqlVisitors);
```
filipelautert, line 119: In this case a sqlvisitor for the provided end delimiter is already defined?

filipelautert (changes requested):

filipelautert (approved):

MalloD12:
@filipelautert @wwillard7800: I updated the logic added in this PR to avoid not appending the `;` at the end when it should. Logic now is:

```
if (change instanceof AbstractSQLChange) {
            String endDelimiter = StringUtil.trimToNull(((AbstractSQLChange) change).getEndDelimiter());
            if (endDelimiter != null && !endDelimiter.contentEquals(";")) {
                super.executeStatements(change, changeLog, sqlVisitors);
            } else {
                super.executeStatements(change, changeLog, addSqlVisitors(sqlVisitors));
            }
        }
        else {
                super.executeStatements(change, changeLog, addSqlVisitors(sqlVisitors));
            }
```
What this will do is, if endDelimiter is set, but it's a `;` it should append it at the end as it was happening, but if it's not a `;` it won't append it at the end. I'm not 100% sure of this, for example, if `endDelimiter` set is `/` what should we do in that case, should it be appended anyway? Your feedback will be very appreciated.

@rberezen: I think this should fix the current issue, but anyway this would be pending on whether this is an appropriate fix or not. 

Thanks,
Daniel.

filipelautert (approved):

rberezen (approved):
Looks good to me, tests passed. 
cc @filipelautert @MalloD12

PR #5800, opened 2024-04-15
If the generated SQL for a formatted SQL changelog is multi-line, then set splitStatements to false

wwillard7800:
When output for a diffChangelog is to a formatted SQL changelog, then split statements should be set to false if the generated SQL contains multiple lines.  We want JDBC to execute this SQL as one statement.

## Impact


wwillard7800:
This PR will not work in all changes.  I am closing it.

PR #5806, opened 2024-04-16
DAT-17457

sayaliM0412:
fix: `.github/workflows/run-tests.yml`: run snowflake tests only if PR is merged to master

## Impact

### DIFF
--- a/.github/workflows/run-tests.yml
+++ b/.github/workflows/run-tests.yml
@@ -292,7 +292,7 @@ jobs:
           echo "TH_SNOW_URL=jdbc:snowflake://http://snowflake.localhost.localstack.cloud:4566" >> $GITHUB_ENV
 
       - name: Run Tests
-        if: ${{ github.event_name == 'push' && github.ref == 'refs/heads/master' && matrix.testSystem == 'snowflake' }}
+        if: github.ref == 'refs/heads/master' || matrix.testSystem != 'snowflake'
         run: ./mvnw -B clean verify -DtrimStackTrace=false -Dliquibase.sdk.testSystem.test=${{ matrix.testSystem }} -Dliquibase.sdk.testSystem.acceptLicenses=${{ matrix.testSystem }} -Dliquibase.sdk.testSystem.snowflake.url=${{ env.TH_SNOW_URL }} -Dliquibase.sdk.testSystem.snowflake.username=${{ env.TH_DB_ADMIN }} -Dliquibase.sdk.testSystem.snowflake.password=${{ env.TH_DB_PASSWD }} -Dtest=*IntegrationTest,*ExecutorTest -DfailIfNoTests=false -Dsurefire.failIfNoSpecifiedTests=false
 
       - name: Archive Test Results
--- a/.github/workflows/run-tests.yml
+++ b/.github/workflows/run-tests.yml
@@ -254,7 +254,7 @@ jobs:
           cache: 'maven'
 
       - name: Set up JDK 11 # get rid of this after https://github.com/snowflakedb/snowflake-jdbc/issues/589 is fixed
-        if: ${{ matrix.testSystem == 'snowflake' }}
+        if: github.ref == 'refs/heads/master' && matrix.testSystem == 'snowflake'
         uses: actions/setup-java@v4
         with:
           java-version: '11'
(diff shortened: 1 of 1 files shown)


jnewton03 (approved):

StevenMassaro (changes requested):
This is going to fail on PR builds. The Snowflake tests also need to be disabled on PR runs.

sayaliM0412:
> This is going to fail on PR builds. The Snowflake tests also need to be disabled on PR runs.

added a condition to Skip integration test if testSystem is 'snowflake'  `if: ${{ matrix.testSystem != 'snowflake' }}`

StevenMassaro (changes requested):
I don't think you want to comment out snowflake from the matrix. Otherwise it won't run on master.

.github/workflows/run-tests.yml
```diff
@@ -230,9 +230,10 @@ jobs:
           - postgresql
           - sqlite
           - firebird
-          #- snowflake
+          - snowflake
     needs: build_tests
     timeout-minutes: 30
+    if: ${{ matrix.testSystem != 'snowflake' }}
```
StevenMassaro, line 236: This is still going to prevent the tests from running on Snowflake on master.

StevenMassaro (changes requested):

.github/workflows/run-tests.yml
sayaliM0412: I think we can apply the same logic like in the below ones, what do you think?
`if: ${{ github.event_name == 'push' && github.ref == 'refs/heads/master' && matrix.testSystem == 'snowflake' }}`

StevenMassaro: That doesn't look right to me. You need to conditionally skip the tests run on line 295: https://github.com/liquibase/liquibase/pull/5806/files#diff-7314d0ebbd2e9537ae4889316745b4fd2fa43cb86275c9caae18a86ba228b642R295 when the branch is not master and the test system is snowflake.

sayaliM0412: @StevenMassaro : made the change. I think this needs to be merged to master to check some if it runs only on master and not on a PR. 
cc: @filipelautert

.github/workflows/run-tests.yml
```diff
@@ -292,6 +292,7 @@ jobs:
           echo "TH_SNOW_URL=jdbc:snowflake://http://snowflake.localhost.localstack.cloud:4566" >> $GITHUB_ENV
 
       - name: Run Tests
+        if: ${{ github.event_name == 'push' && github.ref == 'refs/heads/master' && matrix.testSystem == 'snowflake' }}
```
StevenMassaro, line 295: I don't think this is what you want either, because this is going to run the tests ONLY on master and ONLY on snowflake.

You want something like:

`if: master || !snowflake`

Unrelated, but, will the push event trigger when a PR is merged to master? I am unfamiliar.

StevenMassaro (changes requested):

.github/workflows/run-tests.yml
sayaliM0412: hmm, it should be what I had previously ie `pull_request`

StevenMassaro: I don't think you need any condition related to the `event_name`. The condition I put in my previous comment should be sufficient. Run only on master or, if not on master, run only if not in the snowflake part of the matrix.

sayaliM0412: we do need to take into consideration if its' a PR or a merge to master though, hence the `event_name`

StevenMassaro: Wouldn't `github.ref` take care of that?

sayaliM0412: what you mentioned translates to `if: github.ref == 'refs/heads/master' || matrix.testSystem != 'snowflake'`

sayaliM0412: I think we should test it out

.github/workflows/run-tests.yml
```diff
@@ -254,7 +254,7 @@ jobs:
           cache: 'maven'
 
       - name: Set up JDK 11 # get rid of this after https://github.com/snowflakedb/snowflake-jdbc/issues/589 is fixed
-        if: ${{ matrix.testSystem == 'snowflake' }}
+        if: ${{ github.event_name == 'push' && github.ref == 'refs/heads/master' && matrix.testSystem == 'snowflake' }}
```
StevenMassaro, line 257: Should you leave the `github.event_name == 'push'` here and on the below two steps?

.github/workflows/run-tests.yml
```diff
@@ -292,6 +292,7 @@ jobs:
           echo "TH_SNOW_URL=jdbc:snowflake://http://snowflake.localhost.localstack.cloud:4566" >> $GITHUB_ENV
 
       - name: Run Tests
+        if: github.ref == 'refs/heads/master' || matrix.testSystem != 'snowflake'
```
StevenMassaro, line 295: Does this need the `${{`? I don't know enough about GHA to know if this is valid without it.

sayaliM0412: yes, forgot in a hurry!!

.github/workflows/run-tests.yml
sayaliM0412: I still think it should be `pull_request` all the way.

StevenMassaro (approved):
I have no idea if the `event_name` stuff is right, but everything else looks good to me. We'll definitely need to test this after merging.

PR #5814, opened 2024-04-17
refactor ValidatingVisitor to use factory pattern, error if strict is set and defaults file does not exist (DAT-15920)

StevenMassaro:
## Impact

## Description

Refactor the `ValidatingVisitor` into the factory pattern to support pro usages.


StevenMassaro:
@filipelautert @suryaaki @wwillard7800 @abrackx 

I would appreciate some opinions on the breaking changes introduced in this PR. Are they problematic? If so, should I see if I can reduce the number of breaking changes (at the cost of making the new interface have a not ideal class name).

filipelautert:
> @filipelautert @suryaaki @wwillard7800 @abrackx
> 
> I would appreciate some opinions on the breaking changes introduced in this PR. Are they problematic? If so, should I see if I can reduce the number of breaking changes (at the cost of making the new interface have a not ideal class name).

@StevenMassaro I searched in our extensions and seems no one overrides or uses it. It makes sense as ValidatingVisitor is at the core of liquibase. Perhaps adding javadoc to the new interface ValidatingVisitor providing directions on how to use it would be enough?

wwillard7800:
> > @filipelautert @suryaaki @wwillard7800 @abrackx
> > I would appreciate some opinions on the breaking changes introduced in this PR. Are they problematic? If so, should I see if I can reduce the number of breaking changes (at the cost of making the new interface have a not ideal class name).
> 
> @StevenMassaro I searched in our extensions and seems no one overrides or uses it. It makes sense as ValidatingVisitor is at the core of liquibase. Perhaps adding javadoc to the new interface ValidatingVisitor providing directions on how to use it would be enough?

I agree with @filipelautert .  I think if we provide instructions on how to get their customized `ValidatingVisitor` into play, then we are good.

filipelautert (approved):

StevenMassaro:
This approach is not ideal. Closing in favor of a new approach.

StevenMassaro:
@filipelautert Can you rereview this? I changed the implementation to not be a breaking change. This should improve things considerably.

abrackx (approved):

filipelautert (approved):
Well, that's an even better approach!

PR #5816, opened 2024-04-17
Add tests and enhance error messaging when using loadData with clob and blob files (DAT-16836)

abrackx:
## Impact

## Description
Add tests and enhance error messaging when using loadData with clob and blob files.


liquibase-standard/src/main/java/liquibase/change/core/LoadDataChange.java
```diff
@@ -394,11 +394,17 @@ public SqlStatement[] generateStatements(Database database) {
                                 valueConfig.setValue(value);
                                 needsPreparedStatement = true;
                             } else {
-                                valueConfig.setValueBlobFile(value);
-                                needsPreparedStatement = true;
+                                // Previously in the fallthrough case we would use `setValueBlobFile` to set the blob file to `value`.
+                                // When executing the statement we would then try to find a file with the filename of the contents of `value`.
+                                // I removed this behaviour as I can't understand how it was expected to work in the first place and replaced
+                                // it with an exception. Reading anything besides binary causes downstream parsing errors anyway.
+                                throw new UnexpectedLiquibaseException(String.format("Error parsing value on line %d: '%s'  is not a valid blob value.", lineNumber, value));
```
StevenMassaro, line 401: What is the purpose of this change here? I'm having trouble following it.

abrackx: The previous code would fail trying to load a file with the filename of the contents of `value`. 

If I changed this else branch to set the value like I did in the `clob` case then once we try to insert the data:
```java
valueConfig.setValue(value);
needsPreparedStatement = true;
``` 
we will get an error `java.sql.BatchUpdateException: ORA-01465: invalid hex number` when attempting to insert anything other than base64 values. 

If I _also_ set the type in this branch (making it identical to the base64 branch): 
```java
valueConfig.setType(columnConfig.getType());
valueConfig.setValue(value);
needsPreparedStatement = true;
```
then we will throw an error: `java.lang.IllegalArgumentException: Illegal base64 character 20`. 

From what I can tell this else branch always lead to errors. So I just made it a bit more explicit imo. Let me know if you think there's a better way to word that in this comment.

StevenMassaro: Are you actually trying to insert base64 data? Because I swear we did that during our zoom call together and it worked, and we only had an error when the test assertions used the wrong method on the ResultSet.

abrackx: That was when we were actually trying to insert base64 data that I got that casting error from the assertion. We were still in the `base64` path in that case (at least IIRC!!!)

abrackx: This conversation is no longer relevant.

filipelautert (approved):

PR #5842, opened 2024-04-24
optimize DBCL checksum query and set fetch size on all statements (DAT-17267)

StevenMassaro:
## Impact

### DIFF
--- a/liquibase-standard/src/main/java/liquibase/changelog/StandardChangeLogHistoryService.java
+++ b/liquibase-standard/src/main/java/liquibase/changelog/StandardChangeLogHistoryService.java
@@ -245,14 +245,12 @@ public class StandardChangeLogHistoryService extends AbstractChangeLogHistorySer
             }
 
             SqlStatement databaseChangeLogStatement = new SelectFromDatabaseChangeLogStatement(
-                    new SelectFromDatabaseChangeLogStatement.ByNotNullCheckSum(),
+                    new SelectFromDatabaseChangeLogStatement.ByCheckSumNotLike(ChecksumVersion.latest().getVersion()),
                     new ColumnConfig().setName("MD5SUM"));
             List<Map<String, ?>> md5sumRS = ChangelogJdbcMdcListener.query(getDatabase(), ex -> ex.queryForList(databaseChangeLogStatement));
 
-            if (!md5sumRS.isEmpty()) {
-                //check if any checksum is not using the current version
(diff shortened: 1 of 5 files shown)


wwillard7800 (approved):

filipelautert (approved):

liquibase-standard/src/main/java/liquibase/changelog/StandardChangeLogHistoryService.java
```diff
@@ -245,14 +245,12 @@ public void init() throws DatabaseException {
             }
 
             SqlStatement databaseChangeLogStatement = new SelectFromDatabaseChangeLogStatement(
-                    new SelectFromDatabaseChangeLogStatement.ByNotNullCheckSum(),
+                    new SelectFromDatabaseChangeLogStatement.ByCheckSumNotLike(ChecksumVersion.latest().getVersion()),
```
StevenMassaro, line 248: @filipelautert Do you think this should be checking `IS NOT NULL AND MD5SUM LIKE '...`, in other words, is the `IS NOT NULL` part important, which existed in the old version?

filipelautert: > @filipelautert Do you think this should be checking `IS NOT NULL AND MD5SUM LIKE '...`, in other words, is the `IS NOT NULL` part important, which existed in the old version?

@StevenMassaro That's true, good catch! After a clearChecksums everything will be null and new line 253 will get a different result than old line 254 .

filipelautert (changes requested):

liquibase-standard/src/main/java/liquibase/statement/core/SelectFromDatabaseChangeLogStatement.java
```diff
@@ -71,8 +81,23 @@ public ByTag(String tagName) {
             this.tagName = tagName;
         }
 
-        public String getTagName() {
-            return tagName;
+        @Override
+        public String generateSql(Database database) {
+            return " WHERE "+database.escapeColumnName(null, null, null, "TAG")+"='" + getTagName() + "'";
```
filipelautert, line 86: Could you replace those string concatenations by String.format ? in this file and in 
 SelectFromDatabaseChangeLogStatement.java  too!

filipelautert (approved):

PR #5850, opened 2024-04-25
Change visibility of useful methods to support custom checks (DAT-16671)

wwillard7800:
## Impact


liquibase-dist/src/main/assembly/assembly-bin-common.xml
filipelautert, line 1: Did this file just have indentation fixes ?

wwillard7800: I wanted to get the file formatted before I made my real change.  There should be another commit after this one.

liquibase-dist/pom.xml
```diff
@@ -27,11 +27,12 @@
 		<db2.version>11.5.9.0</db2.version>
 		<firebird.version>5.0.4.java8</firebird.version>
 		<liquibase-pro.version>master-SNAPSHOT</liquibase-pro.version>
+		<liquibase-scripting.version>0-SNAPSHOT</liquibase-scripting.version>
```
filipelautert, line 30: Should not be master-SNAPSHOT too?

liquibase-dist/pom.xml
```diff
@@ -346,7 +347,27 @@
...
+                    <version>${liquibase-pro.version}</version>
+				</dependency>
+			</dependencies>
+		</profile>
+
+		<profile>
+			<id>liquibase-scripting</id>
+			<activation>
+				<property>
+            <name>liquibase.scripting</name>
```
filipelautert, line 359: Could you fix the indentation here?

filipelautert (changes requested):

liquibase-dist/pom.xml
wwillard7800: This is confusing to me, too.  The artifact that is being produced right now is `0-SNAPSHOT`.  I think that once devops is finished with their build work it should be `master-SNAPSHOT`.

liquibase-dist/pom.xml
wwillard7800: Yes.

filipelautert (changes requested):
We need to review this one with Devops because of Bigquery work to make sure that both are implementing the same solution.

filipelautert:
See DAT-18103

filipelautert (approved):

PR #5853, opened 2024-04-26
use Apache Commons where possible, instead of homegrown solutions (DAT-17303)

StevenMassaro:
## Impact

### DIFF
--- a/liquibase-standard/src/main/java/liquibase/util/StringUtil.java
+++ b/liquibase-standard/src/main/java/liquibase/util/StringUtil.java
@@ -1028,8 +1028,6 @@ public class StringUtil {
 
     /**
      * <p>Splits a camel-case string into words based on the came casing.
-     * <p>
-     * This code originated from the StringUtils class of <a href="https://github.com/apache/commons-lang">commons-lang</a>
      *
      * @param str the String to split, may be {@code null}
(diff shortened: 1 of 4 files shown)


liquibase-standard/src/main/java/liquibase/util/StringUtil.java
```diff
@@ -580,7 +526,7 @@ public static boolean isAscii(String string) {
      * @return true if 7 bit-clean, false otherwise.
      */
     public static boolean isAscii(char ch) {
-        return ch < 128;
+        return StringUtils.isAsciiPrintable(Character.toString(ch));
```
StevenMassaro, line 529:
```suggestion
        // This cannot use Apache Commons here because Commons uses different (correct) criteria to determine if a character is ascii
        return ch < 128;
```

liquibase-standard/src/main/java/liquibase/util/StringUtil.java
```diff
@@ -712,7 +655,7 @@ public static boolean isWhitespace(CharSequence string) {
         if (string == null) {
```
filipelautert, line 655: Same here.. should we deprecate this entire class and say "USe StringUtils from commons instead"?

filipelautert (changes requested):

PR #5869, opened 2024-05-02
Update diff summary results in mdc to not be recalculated (DAT-17096)

abrackx:
## Impact

## Description
If there were multiple usages of the standard diff generator during a command's pipeline the diff results would jump around unexpectedly. This change generates the number a single time during the diff's lifecycle, impacting diff and diff-changelog command's mdc.


filipelautert (approved):

StevenMassaro (approved):
Is there a test that can be added for this?

PR #5883, opened 2024-05-06
Issue 5654 includeall min max depth fixes

### LINKED ISSUE #5654: The maxDepth and minDepth attributes of includeall don't work as documented
### Search first

### Description

The documentation states that the depth parameters should indicate the distance from the included path, but the actual behavior is to calculate depth relative to the search path.  For example, if you had Liquibase code in `src/main/resources/liquibase`, and you wanted to include files in `src/main/resources/liquibase/include` to a maxDepth of 0 to exclude further subdirectories, no changes would be included because the calculated depth of the files in the included path would be 5, which is more than the specified max of 0.

I've traced the issue to ApbstractPathResourceAccessor's meetsSearchCriteria method, which is supposed to look at an included file to see if it should be processed.  There seems to be 3 issues in the code:
1.  "depth" is calculated as the file's parent's name count - the root name count.  It should be the file's parent's name count - the included path's name count.  This would cause depth to be relative to the path and not the working directory, as documented.
2. The `depth < minDepth` comparison should be `depth <= minDepth`
3. I don't see any code that compares the file's depth to the maxDepth, so I think that attribute might not work at all.

### Steps To Reproduce

I don't have a current reproduction - I see the problem in the unit tests for the Liquibase Groovy DSL as I try to get it working in newer versions of Liquibase.

### Expected/Desired Behavior

I expect the depth to be relative to the included path, as documented.  I think the documentation makes sense, and is the desired behavior, especially when you have changelogs in src/main/resources, and mix running Liquibase on the comand line (or via a build tool plugin) and having Spring Boot run it via the classpath. 

### Liquibase Version

4.21+

### Database Vendor & Version

All

### Liquibase Integration

All

### Liquibase Extensions

_No response_

All

### Additional Context

_No response_

jasonlyle88:
## Impact

## Description

Fixes #5654 

Depth calculations were not being calculated off the target path provided in includeAll but rather the path of the changelog including the includeAll.

Code updated to address incorrect depth calculation and MUCH more testing was added

### DIFF
--- a/liquibase-integration-tests/src/test/resources/changelogs/common/common.tests.changelog.xml
+++ b/liquibase-integration-tests/src/test/resources/changelogs/common/common.tests.changelog.xml
@@ -696,6 +696,8 @@
 
     <includeAll path="changelogs/common/pathincluded/" minDepth="0" maxDepth="1"/>
 
+    <include relativeToChangelogFile="true" file="depthboundincludeall/changelog.xml" />
+
     <include file="includerelative\pathinclude1.changelog.xml" relativeToChangelogFile="true"/> <!-- try with dos style path -->
 
     <changeSet id="autocommitTest" author="nvoxland" runInTransaction="false">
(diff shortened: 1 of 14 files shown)


filipelautert (approved):
Thanks for the fix @jasonlyle88 !

rberezen (approved):

PR #5888, opened 2024-05-07
normalize paths in DatabaseChangeLog using commons-io (DAT-17592)

StevenMassaro:
## Impact

## Description

Using `Paths.get` will fail on Windows with certain types of paths, like an `s3://` path. This PR changes the behavior to first use Commons-IO to try to normalize the path, then fail back to `Paths.get` in the event that commons doesn't work as desired.


filipelautert (approved):
I/O!

filipelautert (changes requested):
You also need to update the README.TXT with commons-io

filipelautert (approved):

Issue #5878, opened 2024-05-05
Path normalization breaks classpath-relative paths with a custom ResourceAccessor


andrus:
### Search first

- [X] I searched and no similar issues were found

### Description

I am trying again to upgrade Bootique.io framework to Liquibase 4.x from 3.x. An earlier issue #3105 was fixed in LB 4.17, and helped us to move forward. However, there are still a few places where the base path of a resource is altered, so the relative resources can not be properly resolved inside a custom ResourceAccessor. The culprit is stripping the `classpath:` prefix during path normalization. So this change to `DatabaseChangeLog.normalizePath(..)` fixes everything for me:
```
--- a/liquibase-standard/src/main/java/liquibase/changelog/DatabaseChangeLog.java
+++ b/liquibase-standard/src/main/java/liquibase/changelog/DatabaseChangeLog.java
@@ -1028,10 +1028,6 @@ public class DatabaseChangeLog implements Comparable<DatabaseChangeLog>, Conditi
             return null;
         }

-        if (filePath.startsWith("classpath:")) {
-            filePath = filePath.substring("classpath:".length());
-        }
-
         if (filePath.contains("\\")) {
             filePath = filePath.replace("\\", "/");
         }
```

But I suspect such a change may break things elsewhere, so not sending a PR just yet 😀 Instead, let me point to some of the call spots for `normalizePath(..)` that resulted in broken relative resources, and maybe someone can suggest a proper solution, such as an alternative normalize method that preserves the "protocol" prefix in paths:

* https://github.com/liquibase/liquibase/blob/d956d5f9d9377e9d4205a82238a247a2965b4e21/liquibase-standard/src/main/java/liquibase/parser/core/yaml/YamlChangeLogParser.java#L46
* https://github.com/liquibase/liquibase/blob/d956d5f9d9377e9d4205a82238a247a2965b4e21/liquibase-standard/src/main/java/liquibase/parser/core/xml/AbstractChangeLogParser.java#L20
* https://github.com/liquibase/liquibase/blob/d956d5f9d9377e9d4205a82238a247a2965b4e21/liquibase-standard/src/main/java/liquibase/changelog/DatabaseChangeLog.java#L969

### Steps To Reproduce

This test on branch 123 fails: https://github.com/bootique/bootique-jdbc/blob/123/bootique-jdbc-junit5-derby/src/test/java/io/bootique/jdbc/junit5/derby/DerbyTester_Liquibase_RelativeIT.java

```
[...]

PR #5894, opened 2024-05-08
do not strip "classpath:" when normalizing the path

andrus:
## Impact

This may or may not be a breaking change.
 

## Description

This PR addresses the issue #5878 . It ensures that path "normalization" never strips the `classpath:` prefix when creating `DatabaseChangeLog` for child resources. The solution is submitted per @tati-qalified advise. 

_A user-oriented description:_

This PR introduces an alternative behavior for sub-resource path resolution, namely if a root resource path starts with the `classpath:` prefix, it would preserve that prefix when resolving sub-resource locations. This behavior is activated with `liquibase.preserveClasspathPrefixInNormalizedPaths` property (it must be set to `true`), or via `LIQUIBASE_PRESERVE_CLASSPATH_PREFIX_IN_NORMALIZED_PATHS` variable or `--preserve-classpath-prefix-in-normalized-paths` CLI option. This feature allows to implement custom `ResourceAccessors` that work with abstract URLs without being directly tied to the filesystem, such as the one in Bootique.io. 

## Things to be aware of

The altered method is static and is widely used across Liquibase. This may or may not be a breaking change.

## Things to worry about

I don't fully understand the consequences of NOT stripping `classpath:` prefix when resolving relative paths for other implementations of ResourceAccessor. Quite possibly some of them will need to be updated as well. But it seems logical: in a hierarchy of resources, child resource resolution should follow the same algorithm as the root resource. 

## Additional Context

See #5878


MalloD12:
Hi @andrus,

How are you doing? We very much appreciate you trying to fix #5878 and create this PR for it, but we can't merge it as it is. This PR is breaking some tests, but besides that, I have been discussing it with the team and we think this is a breaking change and we don't want to change the current behavior. 

I think what we can do is define a new global configuration (you can have a look at `src/main/java/liquibase/GlobalConfiguration.java` to see some existent examples) and depending on the value set for this configuration then we can decide to strip/don't strip `classpath:` from the path. I think this way we would prevent breaking somebody else code.

If you have any questions/concerns about it please let us know and me or someone else from the team would be keen to help with you. Last thing, if you are keen to do address these modifications could you please add some unit tests to test the code changes made.

Thanks,
Daniel.

andrus:
Hi @MalloD12 , I suspected we can't just change the policy globally. The suggested idea of doing it conditionally via GlobalConfiguration property should work. Not sure how hard it will be to write a test on the LB site, but I will definitely be happy to test it on Bootique side (and I already have unit tests there).

MalloD12:
You can try adding the tests also in Liquibase repo and I can help with it, or you can share the scenarios you would like to add and I can assist you on adding them. 

Thanks,
Daniel.

andrus:
Let me poke around

andrus:
@MalloD12,  I updated the PR. Now it has two commits:

* c0d203b91310c610c9b5c880995fe4c72fc2d5a5 - a unit test that shows how relative paths do not work for a certain type of dynamic ResourceAccessors
* 94950cad34921f62bf1d5940b6f9e635e1148e76 - my original "fix" that will be replaced with GlobalConfiguration property as you suggested.

Hope this helps with that property implementation.

liquibase-standard/src/main/java/liquibase/changelog/DatabaseChangeLog.java
```diff
@@ -1071,10 +1071,6 @@ public static String normalizePath(String filePath) {
             return null;
         }
 
-        if (filePath.startsWith("classpath:")) {
```
MalloD12, line 1074: I think we should leave this code as it is for now.

MalloD12:
Hi @andrus,

I've been looking at this PR and it's still in a not mergeable state.  By removing:

```java
if (filePath.startsWith("classpath:")) {
    filePath = filePath.substring("classpath:".length());
}
``` 
will change and break the current behavior which is what we are trying to avoid at the moment. I think it is a reasonable approach to create a new resource accessor class as you are doing and handling there the decision to according the global configuration we were discussing previously decide whether we want to remove or not the `classpath:` prefix from the path when normalizing.

The global configuration I was talking about would be something like:

```java 
INCLUDE_CLASSPATH_PREFIX = builder.define("includeClasspathPrefix", Boolean.class)
                .setDescription("If false Liquibase will remove 'classpath:' prefix when normalizing a resource path. By default, path will keep the classpath prefix")
                .setDefaultValue(true)
                .build();
```

Thanks,
Daniel.

andrus:
Yes, this is what I said in my comment. The second commit in the PR is for information purposes so to speak. Anyways, I will remove it shortly to avoid any confusion.

andrus:
Just updated the PR:

* Removed the changes from `DatabaseChangeLog`
* Moved the test `ResourceAccessor` into a top-level class as requested in the previous review.

MalloD12:
Hey @andrus,

Are you keen to add the global configuration and make the necessary changes to the new resource accessor class you have already implemented?

Thanks,
Daniel.

andrus:
I can give it a shot. But unfortunately the check can't be done inside a `ResourceAccessor` (otherwise I wouldn't have a need to patch Liquibase). The check will have to go in the `DatabaseChangeLog.normalizePath(..)` method, and depending on the config value, it would either strip the `classpath:` prefix or not. Is that ok?

andrus:
Hi @MalloD12 , I just updated the PR to conditionally use `preserveClasspathPrefixInNormalizedPaths` GlobalConfiguration property, with default being the current behavior. Also extended the tests to verify multiple levels of nesting and check the outcome with different values of the above property.

Please let me know what you guys think. Looking forward to inclusion of this in one of the upcoming LB releases, as using LB 3.x has become untenable for the Bootique community. 😅

andrus:
Hi @MalloD12 , wonder if you had a chance to take a look at this?

MalloD12:
Hi @andrus,

Thank you for all your updates on this PR. I'll start looking at it tomorrow. 

Thanks,
Daniel.

liquibase-standard/src/test/java/liquibase/resource/UrlDrivenResourceAccessorTest.java
```diff
@@ -0,0 +1,168 @@
...
+                true,
+                accessor,
+                new ContextExpression(),
+                new Labels(),
+                false,
+                DatabaseChangeLog.OnUnknownFileFormat.FAIL));
+    }
+
+    @Test
+    public void nestedChangeSet_RelativePath_Allowed() throws LiquibaseException {
```
MalloD12, line 67: Hi @andrus,

Build is currently failing because of this test that is failing with the below message:

```java
Error:  Errors: 
Error:    UrlDrivenResourceAccessorTest.nestedChangeSet_RelativePath_Allowed:72 � Liquibase Error parsing classpath:liquibase/resource/changelog2.yml : Illegal char <:> at index 9: classpath:liquibase/resource/subresources/child-changelog1.yml
```

As you can see on the execution details this test is only failing on Windows. Would it be possible for you to fix it?

Thanks,
Daniel.

andrus: Yeah, I can reproduce the issue on Windows now. The culprit is [this code](https://github.com/liquibase/liquibase/blob/master/liquibase-standard/src/main/java/liquibase/changelog/DatabaseChangeLog.java#L983) :

```java
fileName = resourceAccessor.get(this.getPhysicalFilePath()).resolveSibling(fileName).getPath();
fileName = normalizePath(Paths.get(fileName).toString());
```

Surprisingly, `Paths.get("classpath:some/path)` works on MacOS, but (less surprisingly) fails on Windows with the error above. Trying to figure out if there's a non-invasive solution.

MalloD12: Cool, thanks for taking the time to look at it.

Daniel.

andrus: Ok, just pushed 19b7ac1d746428ceafe182efee3b32dce0bdac7a with a rather defensive fix ensuring that path normalization via `java.nio.file.Path` can work even for `classpath:` URLs.

liquibase-standard/src/main/java/liquibase/changelog/DatabaseChangeLog.java
```diff
@@ -1106,7 +1108,7 @@ public static String normalizePath(String filePath) {
         we fall back to path normalization using Paths.get(), which might fail on Windows.
          */
         if (normalized == null) {
-            normalized = Paths.get(filePath).normalize().toString();
+            normalized = normalizePathViaPaths(filePath, true);
```
MalloD12, line 1111: Hey @andrus,

Sonar scan is reporting an issue here, because  `normalizePathViaPaths()` can return a null value and it could ends up in NPE with the following code execution. I think we would need to add a `NULL` check to avoid that happening.

MalloD12: Not sure if you are able to see issues reported by Sonar in the following [link](https://sonarcloud.io/project/issues?issueStatuses=OPEN%2CCONFIRMED&pullRequest=5894&id=liquibase).

Thanks,
Daniel.

andrus: It only returns null when filePath is null, and there's already a check for that at the top of `normalizePath`. Any non-null path will result in a non-null return. So while Sonar is formally correct, it is being too paranoid :) 

None the less, I just pushed a change with a null check

MalloD12 (approved):
Code changes looks good to me. Thannk you, @andrus for submitting this PR and addressing all the proposed review comments. 

Thanks,
Daniel.

andrus:
@MalloD12 thanks for the review and suggestions! Looks like we need one more approval Who can help with that?

MalloD12:
Hey @andrus, yeap someone else from the team needs to review this as well. Another Dev and QA team will review it, we are releasing a patch this week, so I would think if everything is ok for the rest of the team this change will be part of our next release (4.30.0).

Thanks,
Daniel.

filipelautert (approved):
Thanks for the improvement @andrus !

filipelautert:
@MalloD12 could you create a documentation ticket for this PR?

liquibase-standard/src/test/java/liquibase/resource/UrlDrivenResourceAccessor.java
```diff
@@ -0,0 +1,103 @@
...
+
+        return Collections.singletonList(resolveAsUri(resourceId));
+    }
+
+    private String resolveAsClasspath(String resourceId) {
+        String path = resourceId.substring(CLASSPATH_URL_PREFIX.length());
+
+        // classpath URLs must not start with a slash. This does not work with ClassLoader.
+        // TODO: should we silently strip the leading path?
+        if (path.length() > 0 && path.charAt(0) == '/') {
```
tati-qalified, line 79: Should we be checking for a backslash as well?

andrus: This is a _test_ accessor , so it works with a small set of known inputs.

andrus: To add to that, the input is a URL, not a filesystem path, and as such is always a `/`

tati-qalified (approved):

MalloD12:
Hi @andrus,

Could you please do us the favor of updating this PR description with the addition and usage of the newly introduced flag, so our docs team can update our docs accordingly? 

Thanks,
Daniel.

MalloD12:
[PD-5137](https://datical.atlassian.net/browse/PD-5137) has been created for docs update.

Thanks,
Daniel.

rberezen (approved):

andrus:
@MalloD12 

> Could you please do us the favor of updating this PR description with the addition and usage of the newly introduced flag, so our docs team can update our docs accordingly?

Just added a paragraph to the top description.

andrus:
Hi folks. Do we need any extra approvals to get this merged?

PR #5619, opened 2024-02-21
Improve snapshot extensibility for NoSQL databases

fbiville:
## Impact

- [ ] Bug fix (non-breaking change which fixes expected existing functionality)
- [x] Enhancement/New feature (adds functionality without impacting existing logic)
- [ ] Breaking change (fix or feature that would cause existing functionality to change)

Pro companion PR: https://github.com/liquibase/liquibase-pro/pull/1605
Pro-tests companion PR: https://github.com/liquibase/liquibase-pro-tests/pull/1322
 
## Things to be aware of

* Fixes SnapshotGeneratorChain
* Improves Database.supports methods

Having a generic method as 
```
 default boolean supports(Class<? extends DatabaseObject> object) {
```

allows extensions (specially NoSql ones) to implement validations as 

```
supports(Table.class) && supports(Catalog.class) 
```

without changing core to add a new "supportsTables" method.

## Things to worry about
TODO

PR #5897, opened 2024-05-10
chore: invoke deprecated methods in default supports method for backwards compatibility

filipelautert:
### TLDR: 
Invoke deprecated methods in default supports method to ensure that extensions are not broken, but one day when those are removed it will return only true

### Description
The new support `method` was introduced in PR https://github.com/liquibase/liquibase/pull/5619 so specially NoSQL databases can check for other supports as `Collections` without changing Database class and adding a `supportsCollections()` method.
In core all the locations were `supportsSequences()` existed were changed to supports(Sequence) , and all *Database classes had this new method implemented. But extensions don't have it, so they could return the wrong value. For instance, BigQuery doesn't support sequences but would start to return true to core.
With this fix we continue to return what is expected.

## Impact

### DIFF
--- a/liquibase-standard/src/main/java/liquibase/database/Database.java
+++ b/liquibase-standard/src/main/java/liquibase/database/Database.java
@@ -25,10 +25,7 @@ import liquibase.sqlgenerator.SqlGeneratorFactory;
 import liquibase.statement.DatabaseFunction;
 import liquibase.statement.SqlStatement;
 import liquibase.structure.DatabaseObject;
-import liquibase.structure.core.ForeignKey;
-import liquibase.structure.core.Index;
-import liquibase.structure.core.PrimaryKey;
-import liquibase.structure.core.UniqueConstraint;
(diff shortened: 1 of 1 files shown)


StevenMassaro (approved):
Code here looks fine but I don't know any of the context around this change.

filipelautert:
> Code here looks fine but I don't know any of the context around this change.

@StevenMassaro the new support `method` was introduced so specially NoSQL databases can check for other supports as `Collections` without changing Database class and adding a `supportsCollections()` method.
In core all the locations were `supportsSequences()` existed were changed to supports(Sequence) , and all *Database classes had this new method implemented. But extensions don't have it, so they could return the wrong value. For instance, BigQuery doesn't support sequences but would start to return true to core.
With this fix we continue to return what is expected.

I'm adding this info to the PR description, thanks!

MalloD12 (approved):
LGTM

PR #5903, opened 2024-05-13
[DAT-16377] Verify if supports method is implemented in a change to prevent unexpected behaviors

filipelautert:
## Impact

## Description
- Created a constant to dev environment
- Verify if the supports method is implemented in the change if it's not part of the default liquibase changes and performs actions according to the defined action level

## Things to be aware of
- DEV will always throw exception to prevent changes being created without supports method


liquibase-standard/src/main/java/liquibase/change/ChangeFactory.java
```diff
@@ -120,6 +134,45 @@ public Change create(String name) {
...
+     * Verify if the supports method is implemented in the change if it's not part of the default liquibase changes
+     */
+    private void verifySupportsMethodImplementation(Set<Change> plugins) {
+        if (GlobalConfiguration.SUPPORTS_METHOD_VALIDATION_LEVELS.getCurrentValue().equals(SupportsMethodValidationLevelsEnum.OFF)) {
+            return;
+        }
+        //we only verify supports method if this is not part of the default liquibase changes
+        for (Change plugin : plugins) {
+            String packageName = plugin.getClass().getPackage().getName();
+            if (packageName.startsWith("liquibase.change") || packageName.startsWith("com.datical.liquibase.ext")) {
```
filipelautert, line 147: Should I keep pro package here or should we verify it too?

suryaaki2 (approved):

liquibase-standard/src/main/java/liquibase/change/ChangeFactory.java
wwillard7800: I think that is okay to keep it out.

wwillard7800 (approved):

liquibase-cli/src/test/groovy/liquibase/integration/commandline/LiquibaseCommandLineTest.groovy
```diff
@@ -430,6 +430,16 @@ Global Options
                                supportPropertyEscaping', environment variable:
                                'LIQUIBASE_SUPPORT_PROPERTY_ESCAPING')
 
+      --supports-method-validation-levels=PARAM
```
StevenMassaro, line 433: This should be `--supports-method-validation-level`, not `levels`.

liquibase-standard/src/main/java/liquibase/change/ChangeFactory.java
StevenMassaro: In fact I think you _should_ remove it since it's an extension.

liquibase-cli/src/test/groovy/liquibase/integration/commandline/LiquibaseCommandLineTest.groovy
```diff
@@ -430,6 +430,16 @@ Global Options
...
+                             Controls the level of validation performed on the
+                               supports method of Change classes. Options are
+                               OFF, WARN, FAIL.
+                             DEFAULT: WARN
```
StevenMassaro, line 437: Should the default be FAIL so that we notice if we violate this rule when we run tests in all of the extensions? If we just warn, we probably won't ever actually notice until runtime, at which point it will be too late.

liquibase-standard/src/main/java/liquibase/change/ChangeFactory.java
```diff
@@ -120,6 +134,43 @@ public Change create(String name) {
...
+                            plugins.remove(plugin);
+                            break;
+                        case FAIL:
+                            throw new UnexpectedLiquibaseException(String.format(SUPPORTS_METHOD_REQUIRED_MESSAGE, plugin.getClass().getName()));
+                        default:
+                            break;
+                    }
+                }
+            } catch (NoSuchMethodException e) {
+                throw new UnexpectedLiquibaseException(String.format(SUPPORTS_METHOD_REQUIRED_MESSAGE, plugin.getClass().getName()));
```
StevenMassaro, line 169: Why are you throwing an exception from here?

StevenMassaro (changes requested):
Can you add tests for this?

liquibase-standard/src/main/java/liquibase/change/ChangeFactory.java
filipelautert: This part of the code will never be reached as the "supports" method will always be there as it is defined in the interface - but reflection throws the checked exception so I just added this extra throws.

liquibase-cli/src/test/groovy/liquibase/integration/commandline/LiquibaseCommandLineTest.groovy
StevenMassaro: Please do and put it in a future sprint.

liquibase-standard/src/main/java/liquibase/change/ChangeFactory.java
```diff
@@ -120,6 +134,43 @@ public Change create(String name) {
...
+    private void verifySupportsMethodImplementation(Set<Change> plugins) {
+        if (GlobalConfiguration.SUPPORTS_METHOD_VALIDATION_LEVEL.getCurrentValue().equals(SupportsMethodValidationLevelsEnum.OFF)) {
+            return;
+        }
+        //we only verify supports method if this is not part of the default liquibase changes
+        for (Change plugin : plugins) {
+            String packageName = plugin.getClass().getPackage().getName();
+            if (packageName.startsWith("liquibase.change")) {
+                continue;
+            }
```
StevenMassaro, line 149: Do we even want this? Thinking more about this, don't we want all changes to be verified? If so, then I think this method just introduces another way for an extension to accidentally override the supports method and us not notice.

liquibase-standard/src/main/java/liquibase/change/ChangeFactory.java
```diff
@@ -120,6 +134,43 @@ public Change create(String name) {
...
+
+            try {
+                // if the supports method is not implemented in the plugin show the wraning according to the defined level
```
StevenMassaro, line 152:
```suggestion
                // if the supports method is not implemented in the plugin show the warning according to the defined level
```

StevenMassaro (approved):

adrian-velonis1:
[DOC] https://datical.atlassian.net/browse/PD-4956

PR #5905, opened 2024-05-13
[DAT-17502] added FK support

KushnirykOleh:
## Impact

## Description
Added classes to support FK for BigQuery. `BigQueryDropForeignKeyConstraintGenerator` is not needed as default implementation works fine


vitaliimak (approved):

liquibase-bigquery/src/main/java/liquibase/snapshot/jvm/BigQueryForeignKeySnapshotGenerator.java
```diff
@@ -0,0 +1,128 @@
...
+                .append("KCU.COLUMN_NAME as FOREIGN_KEY_COLUMN, ")
+                .append("CCU.TABLE_CATALOG as PRIMARY_KEY_TABLE_CATALOG, ")
+                .append("CCU.TABLE_SCHEMA as PRIMARY_KEY_TABLE_SCHEMA, ")
+                .append("CCU.TABLE_NAME as PRIMARY_KEY_TABLE, ")
+                .append("CCU.COLUMN_NAME as PRIMARY_KEY_COLUMN ")
+                .append(String.format("FROM %1$s.%2$s.TABLE_CONSTRAINTS as TC JOIN %1$s.%2$s.CONSTRAINT_COLUMN_USAGE as CCU on " +
+                                "TC.CONSTRAINT_NAME=CCU.CONSTRAINT_NAME JOIN %1$s.%2$s.KEY_COLUMN_USAGE as KCU on KCU.CONSTRAINT_NAME=TC.CONSTRAINT_NAME ",
+                        searchSchema, systemSchema))
+                .append(String.format("WHERE TC.TABLE_NAME='%s' AND TC.table_schema='%s' AND TC.table_catalog='%s' AND TC" +
+                        ".constraint_type='FOREIGN KEY' AND TC.CONSTRAINT_NAME='%s'", searchTableName, searchSchema, searchCatalog, fkName))
```
filipelautert, line 60: A PreparedStatement could be used here to set those 4 parameters.

liquibase-bigquery/src/main/java/liquibase/snapshot/jvm/BigQueryForeignKeySnapshotGenerator.java
```diff
@@ -0,0 +1,128 @@
...
+        }
+        if (foundObject instanceof Table) {
+            Table table = (Table) foundObject;
+            Database database = snapshot.getDatabase();
+            Schema schema = table.getSchema();
+
+            CatalogAndSchema catalogAndSchema = (new CatalogAndSchema(schema.getCatalogName(), schema.getName())).customize(database);
+            String jdbcSchemaName = database.correctObjectName(((AbstractJdbcDatabase) database).getJdbcSchemaName(catalogAndSchema), Schema.class);
+            String query = String.format("SELECT CONSTRAINT_NAME FROM %s.%s.TABLE_CONSTRAINTS WHERE table_name='%s' AND table_schema='%s' AND " +
+                            "table_catalog='%s' AND CONSTRAINT_TYPE='FOREIGN KEY';",
```
filipelautert, line 117: Another PreparedStatement for table.getName(), schema.getName(), schema.getCatalogName()

filipelautert (changes requested):
Overall is fine. I'm just asking for PreparedStatements were possible as we are trying to improve this on core.

liquibase-bigquery/src/main/java/liquibase/snapshot/jvm/BigQueryForeignKeySnapshotGenerator.java
```diff
@@ -0,0 +1,133 @@
...
+
+        if (!results.isEmpty()) {
+            ForeignKey foreignKey = null;
+            for (Map<String, ?> resultMap : results) {
+                Table foreignKeyTable = new Table().setName(Objects.toString(resultMap.get("FOREIGN_KEY_TABLE")));
+
+                String foreignKeyName =
+                        Optional.ofNullable((String) resultMap.get("CONSTRAINT_NAME"))
+                                .map(s -> s.replace(foreignKeyTable.getName() + ".", ""))
+                                .orElse("null");
```
vitaliimak, line 72: Should it be `null`? Not a String `"null"`.

filipelautert (approved):

liquibase-bigquery/src/main/java/liquibase/snapshot/jvm/BigQueryForeignKeySnapshotGenerator.java
KushnirykOleh: if we found FK constraint but it doesn't have a name, (what is unlikely in the first place) we can put it's name as 'null' string or null. 
Null can cause NPE, 'null' should just appear in shapshot as the FK name. 
@filipelautert what's your opinion?

filipelautert: Core doesn't account for this case - it always expect a fk name to be found. So maybe something like "undefined" would be better, as "null" may confuse people. What do think?

vitaliimak (approved):

filipelautert (approved):

PR #5917, opened 2024-05-14
Add support to disable fast-check

### LINKED ISSUE #5784: AbstractUpdateCommandStep.upToDateFastCheck global cache does not detect DB recreation
### Search first

### Description

Hi,

We have a relatively unique case where we have a custom multi tenant datasource and we run liquibase when the tenant logs in.
We create and configure SpringLiquibase and invoke afterPropertiesSet.

Our QA jobs are recreating the DB between runs and the issue is that Liquibase is not applied again.

From what I understand, this is because Liquibase uses an internal cache. 
There is a singleton CommandFactory which holds an instance of all CommandSteps (including UpdateCommandStep)
https://github.com/liquibase/liquibase/blob/master/liquibase-standard/src/main/java/liquibase/command/CommandFactory.java#L244

The AbstractUpdateCommandStep holds an upToDateFastCheck map that remembers if Liquibase was already applied (on that JDBC URL with these changesets, etc.):
https://github.com/liquibase/liquibase/blob/master/liquibase-standard/src/main/java/liquibase/command/core/AbstractUpdateCommandStep.java#L89

This cache exists as long as JVM is up.

Liquibase detects that the dbchangelog tables do not exist and recreate them. I believe this is the place to also invalidate all the upToDateFastCheck cache entries that are relevant to that URL.

### Steps To Reproduce

Attached a sample project.
1. Run failsOnSecondRun test once so it will create dbchangelog with checksums - it should pass.
2. Run failsOnSecondRun test again and it will fail.

[upToDateFastCheck.zip](https://github.com/liquibase/liquibase/files/14924175/upToDateFastCheck.zip)

### Expected/Desired Behavior

If Liquibase understands that databasechangelog tables are missing and creates them, it should also invalidate all the upToDateFastCheck cache entries that are relevant to that URL.

### Liquibase Version

4.24.0 and 4.26.0

### Database Vendor & Version

PostgreSQL

### Liquibase Integration

Custom spring boot

### Liquibase Extensions

_No response_

Ubuntu

### Additional Context

_No response_

MalloD12:
## Impact

## Description

Fixes #5784

### DIFF
--- a/liquibase-standard/src/main/java/liquibase/Liquibase.java
+++ b/liquibase-standard/src/main/java/liquibase/Liquibase.java
@@ -37,6 +37,8 @@ import liquibase.serializer.ChangeLogSerializer;
 import liquibase.structure.DatabaseObject;
 import liquibase.util.LoggingExecutorTextUtil;
 import liquibase.util.StringUtil;
+import lombok.Getter;
+import lombok.Setter;
 
 import java.io.IOException;
 import java.io.PrintStream;
@@ -66,6 +68,10 @@ public class Liquibase implements AutoCloseable {
     private ChangeExecListener changeExecListener;
     private final DefaultChangeExecListener defaultChangeExecListener = new DefaultChangeExecListener();
     private final Map<String, Boolean> upToDateFastCheck = new HashMap<>();
+
+    @Getter
+    @Setter
+    private Boolean fast_check_disabled = Boolean.FALSE;
 
     /**
      * Creates a Liquibase instance for a given DatabaseConnection. The Database instance used will be found with {@link DatabaseFactory#findCorrectDatabaseImplementation(liquibase.database.DatabaseConnection)}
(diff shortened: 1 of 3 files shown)


liquibase-standard/src/main/java/liquibase/Liquibase.java
```diff
@@ -67,6 +69,10 @@ public class Liquibase implements AutoCloseable {
     private final DefaultChangeExecListener defaultChangeExecListener = new DefaultChangeExecListener();
     private final Map<String, Boolean> upToDateFastCheck = new HashMap<>();
 
+    @Getter
+    @Setter
+    private Boolean fast_check_disabled = Boolean.FALSE;
```
MosheElisha, line 74: Consider using positive - `fastCheckEnabled` to make it easier to understand if it is enabled or disabled.

liquibase-standard/src/main/java/liquibase/integration/spring/SpringLiquibase.java
```diff
@@ -254,6 +254,7 @@ public void afterPropertiesSet() throws LiquibaseException {
                             final Connection c = getDataSource().getConnection();
                             liquibase = createLiquibase(c);
                             generateRollbackFile(liquibase);
+                            liquibase.setFast_check_disabled(Boolean.TRUE);
```
MosheElisha, line 257: Should we change the default? This might cause a *small* performance hit, right?

liquibase-standard/src/main/java/liquibase/Liquibase.java
MalloD12: Yeah, I can rename it if we want to move forward with this PR. I thought calling it this way (`fast_check_disabled`) because by default this configuration would stay active, but yes I'm fine renaming it.

PR #5924, opened 2024-05-15
Optimize parsing big xml files

asolntsev:
## Impact

## Description

When running a big integration test suite that runs LiquiBase multiple times, I realized that LB spends a lot of time on generating the same method names again and again (e.g. given property name `name`, LB generates Strings `isName`, `getName` and `setName` many-many times).

I suggest an optimization that computes them only once and holds in a Map. 
Should make parsing big XML files faster and consume less memory.

I believe this change doesn't affect any functionality (besides the above-mentioned performance improvement).

### DIFF
--- a/liquibase-standard/src/main/java/liquibase/util/ObjectMethods.java
+++ b/liquibase-standard/src/main/java/liquibase/util/ObjectMethods.java
@@ -1,0 +1,45 @@
+package liquibase.util;
+
+import static java.util.Locale.ROOT;
+import static java.util.stream.Collectors.toMap;
+
+import java.lang.reflect.Method;
+import java.util.Arrays;
+import java.util.HashMap;
+import java.util.Map;
+import java.util.function.BinaryOperator;
+
+final class ObjectMethods {
+
+  private static final BinaryOperator<Method> ignoreMultipleMethodsWithSameName = (first, second) -> first;
+  private final Map<String, Method> writeMethods;
+  private final Map<String, Method> readMethods;
+
+  ObjectMethods(Method[] methods) {
+    readMethods = new HashMap<>(methods.length);
+    readMethods.putAll(find(methods, 0, "get"));
+    readMethods.putAll(find(methods, 0, "is"));
+    writeMethods = find(methods, 1, "set");
+  }
+
+  private Map<String, Method> find(Method[] methods, int parametersCount, String prefix) {
+    return Arrays.stream(methods)
+      .filter(m -> m.getParameterTypes().length == parametersCount)
+      .filter(m -> m.getName().startsWith(prefix))
+      .collect(toMap(m -> propertyName(m.getName(), prefix), m -> m, ignoreMultipleMethodsWithSameName));
+  }
+
+  private String propertyName(String methodName, String prefix) {
+    int length = prefix.length();
+    return methodName.substring(length, length + 1).toLowerCase(ROOT) + methodName.substring(length + 1);
+  }
+
+  Method getReadMethod(String propertyName) {
+    return readMethods.get(propertyName);
+  }
+
+  Method getWriteMethod(String propertyName) {
+    return writeMethods.get(propertyName);
+  }
+
+}
--- a/liquibase-standard/src/main/java/liquibase/util/ObjectMethods.java
+++ b/liquibase-standard/src/main/java/liquibase/util/ObjectMethods.java
@@ -15,6 +15,10 @@ final class ObjectMethods {
   private final Map<String, Method> writeMethods;
   private final Map<String, Method> readMethods;
 
+  ObjectMethods(Class klass) {
+    this(klass.getMethods());
+  }
+
   ObjectMethods(Method[] methods) {
     readMethods = new HashMap<>(methods.length);
     readMethods.putAll(find(methods, 0, "get"));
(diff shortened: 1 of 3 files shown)


liquibase-standard/src/main/java/liquibase/util/ObjectMethods.java
```diff
@@ -0,0 +1,45 @@
...
+  private Map<String, Method> find(Method[] methods, int parametersCount, String prefix) {
+    return Arrays.stream(methods)
+      .filter(m -> m.getParameterTypes().length == parametersCount)
+      .filter(m -> m.getName().startsWith(prefix))
+      .collect(toMap(m -> propertyName(m.getName(), prefix), m -> m, ignoreMultipleMethodsWithSameName));
+  }
+
+  private String propertyName(String methodName, String prefix) {
+    int length = prefix.length();
+    return methodName.substring(length, length + 1).toLowerCase(ROOT) + methodName.substring(length + 1);
```
MalloD12, line 34: I'm not very familiar with this piece of the code. Still, I was wondering whether we should be uppercasing the given string as it was before (`.toUpperCase(Locale.ENGLISH)`) or should lowercase and this implementation does (`.toLowerCase(ROOT)`). I'm not sure if it could have any side effects, I'll check with the team.

MalloD12:
Hi @asolntsev,

Thank you for creating this improvement PR, it's very appreciated it by the team.  I left you a small question/comment I'll be also checking with the team and I'll let you know. 

In general, this PR looks good to me. I think I would only ask you if you can add some unit tests to `ObjectUtilTest` class for this. 

Last thing, do you have a comparison from your use case that shows us how this improvement is behaving in numbers (how it's now vs how it was previously)?

Thanks,
Daniel.

asolntsev:
Thank you Daniel!

> add some unit tests to ObjectUtilTest class

1. Initially, I didn't add any tests because I didn't add or change any functionality. The existing tests already cover the functionality of `ObjectUtil` class.
2. Now I aded dedicated unit-tests for the new class ObjectMethods. 

> do you have a comparison from your use case that shows us how this improvement is behaving in numbers

On my project and my laptop (MacBook pro M1), I run in parallel threads a suite of integration tests that run LiquiBase multiple times to initialize DB (DB is local in docker container).

Execution time of integration test decreased by ~1 minute:

* BEFORE: 8m 38s
* AFTER:  7m 48s

liquibase-standard/src/main/java/liquibase/util/ObjectMethods.java
MalloD12: Hey Andrei, yeap but what I meant is previously the below code:
```
String getMethodName = "get" + propertyName.substring(0, 1).toUpperCase(Locale.ENGLISH)
            + propertyName.substring(1);
```
would return `NAME` for a get/is method name.

asolntsev: @MalloD12 No, it returns `getName`.
Method `toUpperCase` is called only for the first letter after "get".

MalloD12: Hi @asolntsev, 

My bad you are right. Is there a reason or just an implementation agreement you made for returning `name`, instead of `Name`?

liquibase-standard/src/main/java/liquibase/util/ObjectMethods.java
```diff
@@ -0,0 +1,49 @@
...
+  private static final BinaryOperator<Method> ignoreMultipleMethodsWithSameName = (first, second) -> first;
+  private final Map<String, Method> writeMethods;
+  private final Map<String, Method> readMethods;
+
+  ObjectMethods(Class klass) {
+    this(klass.getMethods());
+  }
+
+  ObjectMethods(Method[] methods) {
+    readMethods = new HashMap<>(methods.length);
```
MalloD12, line 23: what does it mean `parametersCount = 0` / `1`?

liquibase-standard/src/main/java/liquibase/util/ObjectMethods.java
asolntsev: > Is there a reason or just an implementation agreement you made for returning name, instead of Name?

@MalloD12 Sorry, I don't understand your last question. Given `methodName = "getName"`, method `propertyName("getName", "get")` returns string `name`. It's the property.

liquibase-standard/src/main/java/liquibase/util/ObjectMethods.java
asolntsev: Look at constructor:
```java
    find(methods, 0, "get");
    find(methods, 0, "is");
    find(methods, 1, "set");
```

`get` an `is` methods have 0 parameters (they are called "getters"), 
while `set` methods have 1 parameter (they are called "setters).

MalloD12: I know what getters/setters are but I didn't realize it was about the number of arguments each of these methods type could have. I thought it could be another filtering regarding types.

It's clear now!

Thank you!

liquibase-standard/src/main/java/liquibase/util/ObjectMethods.java
MalloD12: I was asking if there was a reason why we are returning `name` and not `Name`.

MalloD12 (approved):
**Approved**.

Looks good to me. Most of the build and test checks are looking good (except for functional tests that are failing because of a non-related reason, same for test-harness but they should be ok in the next run).

filipelautert (approved):
That's a nice improvement and code modernization, thanks @asolntsev !

rberezen (approved):

PR #5932, opened 2024-05-20
use Commons-IO where possible (DAT-17634)

StevenMassaro:
## Impact


liquibase-standard/src/main/java/liquibase/io/WriterOutputStream.java
filipelautert, line 1: I created this class ~1 year ago. I think we can safely remove it as if someone is using they are doing it completely wrong.

filipelautert (changes requested):

filipelautert (approved):

filipelautert:
@StevenMassaro whoops, approved but build is broken ...

PR #5812, opened 2024-04-17
DAT-16825 :: RawSqlStatement refactoring

MalloD12:
## Impact

- [ ] Bug fix (non-breaking change which fixes expected existing functionality)
- [x] Enhancement/New feature (adds functionality without impacting existing logic)
- [ ] Breaking change (fix or feature that would cause existing functionality to change)

PR #5956, opened 2024-05-28
Fix RollbackCommandStepStructuredLoggingTest integration test

MalloD12:
## Impact

## Description

After merging #5812 a PRO integration test (`test structured logging for rollback command with a sqlfile rollback`) started failing and it was because for some reason with the merged changes endDelimiter was getting lost in the middle of the statement processing.  By specifying statement `endDelimiter` value when creating an `UnparsedSql` object the failing test started passing again.


MalloD12:
If we are ok with this PR, we need to revert this [PRO-PR](https://github.com/liquibase/liquibase-pro/pull/1750).

StevenMassaro (approved):
This looks to make sense to me, since it now matches what we have in the RawSqlGenerator. I'd like to hear Filipe's thoughts though.

filipelautert (approved):
Makes sense, good catch!

PR #5961, opened 2024-05-30
Make the message be more accurate when an exception occurs during database close

wwillard7800:
I am unable to replicate the original issue or come up with a way to cause an exception on database close. I have added unit/integration tests to simulate the problem and fix.

## Impact

### DIFF
--- a/liquibase-integration-tests/src/test/groovy/liquibase/snapshot/JdbcDatabaseConnectionCloseTest.groovy
+++ b/liquibase-integration-tests/src/test/groovy/liquibase/snapshot/JdbcDatabaseConnectionCloseTest.groovy
@@ -1,0 +1,75 @@
+package liquibase.snapshot
+
+import liquibase.Scope
+import liquibase.database.AbstractJdbcDatabase
+import liquibase.database.Database
+import liquibase.database.jvm.JdbcConnection
+import liquibase.exception.DatabaseException
+import liquibase.extension.testing.testsystem.DatabaseTestSystem
+import liquibase.extension.testing.testsystem.TestSystemFactory
+import liquibase.extension.testing.testsystem.spock.LiquibaseIntegrationTest
+import liquibase.logging.core.BufferedLogService
+import liquibase.parser.ChangeLogParserConfiguration
+import spock.lang.Shared
+import spock.lang.Specification
+
+import java.sql.Connection
+import java.util.logging.Level
+
+@LiquibaseIntegrationTest
+class JdbcDatabaseConnectionCloseTest extends Specification {
+    @Shared
+    public DatabaseTestSystem h2 = Scope.currentScope.getSingleton(TestSystemFactory).getTestSystem("h2")
+
+    def "Test Exception"() {
+        when:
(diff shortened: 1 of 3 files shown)


suryaaki2 (approved):

liquibase-integration-tests/src/test/groovy/liquibase/snapshot/JdbcDatabaseConnectionCloseTest.groovy
```diff
@@ -0,0 +1,75 @@
...
+class JdbcDatabaseConnectionCloseTest extends Specification {
+    @Shared
+    public DatabaseTestSystem h2 = Scope.currentScope.getSingleton(TestSystemFactory).getTestSystem("h2")
+
+    def "Test Exception"() {
+        when:
+        JdbcConnection connection = new TestJdbcConnection(h2.getConnection())
+        Database database = h2.getDatabaseFromFactory()
+        database.setConnection(connection)
+        database.close()
```
MalloD12, line 29: why closing the connection will throw an exception? If that's the case can we rename the test to something like "validate exception is thrown when connection is closed"?

liquibase-integration-tests/src/test/java/liquibase/helper/DBTest.java
```diff
@@ -0,0 +1,74 @@
...
+        try {
+            oracleDCN.run();
+        } catch (Exception ex) {
+            ex.printStackTrace();
+        }
+    }
+
+    private void run() throws Exception {
+        OracleConnection conn = connect();
+        Statement stmt= conn.createStatement();
```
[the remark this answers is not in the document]
MalloD12: is it not codeQL right here and should we check `conn` is not null?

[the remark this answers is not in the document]
wwillard7800: This was code that I did not intend to commit.  I will remove it.

liquibase-integration-tests/src/test/groovy/liquibase/snapshot/JdbcDatabaseConnectionCloseTest.groovy
wwillard7800: Yes, thank you for noticing that.

MalloD12 (approved):
LGTM

filipelautert (approved):

liquibase-standard/src/main/java/liquibase/database/AbstractJdbcDatabase.java
```diff
@@ -168,6 +167,16 @@ public boolean getAutoCommitMode() {
...
 
+    /**
+     * Used for TEST only
+     *
+     * @param previousAutoCommit
+     *
+     */
+    public void setPreviousAutoCommit(Boolean previousAutoCommit) {
+        this.previousAutoCommit = previousAutoCommit;
+    }
```
StevenMassaro, line 178: I am pretty sure that the spock tests (or possibly groovy) let you access private fields and set their value with only a warning.

PR #5966, opened 2024-05-31
Add support for NULL placeholders on loadData

### LINKED ISSUE #3627: Option to translate empty string values to null for loadData
## Environment

**Liquibase Version**: 4.18

**Database Vendor & Version**:  Snowflake

**Operating System Type & Version**:  Linux

**Infrastructure Type/Provider**: GitHub Actions

## Description

When using [loadData](https://docs.liquibase.com/change-types/load-data.html) to load a `csv` file, empty values are loaded as an _empty_ string, rather than a _null_.

While I am aware that we could change the `csv` file to supply the value `NULL`, and then that will be converted to _null_, it seems odd to me that an _empty string_ is the default behavior.

Adding an extra optional property like `treatEmptyAsNull: true` would maintain backwards compatibility.

davidecavestro:
## Impact

## Description

Improve data load supporting a way to optionally set placeholders for NULL values.
This provides a way to define - column by column - a specific string that will be converted into a sql NULL reference when found on a CSV cell.

## Things to be aware of

A new _nullPlaceholder_ attribute has been added to the column definition.
The pre-existing null handling detection is still in place as the default behaviour - when _nullPlaceholder_ is not set - hardcoded to translate any `"null"` string (case-insensitively) into a NULL sql ref.

Added some automated tests covering
- default behaviour (_nullPlaceholder_ not set)
- _nullPlaceholder_ set to empty string
- _nullPlaceholder_ set to non-empty string

## Things to worry about

This implementation is limited to _loadData_ and doesn't affect _loadUpdateData_ though it could probably be ported.

## Additional Context

Fixes #3627


MalloD12 (approved):
**Approved**.

Code changes and unit tests added look good to me. Build and test checks have been successfully executed.

Thank you, @davidecavestro for implementing these changes.

filipelautert (approved):

rberezen:
@davidecavestro Would you mind resolving the Git conflict?

rberezen (approved):

PR #1826, opened 2021-05-06
Prevent silent failures when two "databaseChangeLog" tags are present

danielthegray:
## Environment
**Liquibase Version**: all

**Liquibase Integration & Version**: all

**Liquibase Extension(s) & Version**: all

**Database Vendor & Version**: all

**Operating System Type & Version**: all

## Pull Request Type

* [ ] Bug fix (non-breaking change which fixes an issue.)
* [x] Enhancement/New feature (non-breaking change which adds functionality)
* [ ] Breaking change (fix or feature that would cause existing functionality to change)

## Description
If someone makes a mistake and copy/pastes a databaseChangeLog block in,
and makes the file have two tags, the SnakeYAML parser silently covers
up the first instance with the second one, which will cause the first
block of changes to be silently omitted.

This can be very problematic and cause hard-to-debug errors, which is
why I have added a small FilterInputStream interceptor to check the
stream as it is fed into the YAML parser, for a duplicate tag. Doing it
with the FilterInputStream like this prevents the efficiency losses of
opening the file twice.

## Steps To Reproduce
You can look at the added test; essentially, just create a changelog file which has two "databaseChangeLog" tags.

## Actual Behavior
It will silently to the wrong thing (i.e. skip over the first group of changesets).

## Expected/Desired Behavior
An error should be thrown to highlight the fact that there are two databaseChangeLog tags/keys.

## Fast Track PR Acceptance Checklist:

* [x] Build is successful and all new and existing tests pass
* [x] Added [Unit Test(s)](https://liquibase.jira.com/wiki/spaces/LB/pages/1274937609/How+to+Write+Liquibase+Core+Unit+Tests)
* [ ] Added [Integration Test(s)](https://liquibase.jira.com/wiki/spaces/LB/pages/1276608569/How+to+Write+Liquibase+Core+Integration+Tests)
* [ ] Added [Test Harness Test(s)](https://github.com/liquibase/liquibase-test-harness/pulls)
* [ ] Documentation Updated

## Need Help?
Come chat with us on our [discord channel](https://discord.com/channels/700506481111597066/700506481572839505)

PR #5976, opened 2024-06-03
Prevent silent failures when two "databaseChangeLog" tags are present (fixes #5963)

### LINKED ISSUE #5963: Prevent duplicate databasechangelog properties without breaking when having duplicated sql ones
### Search first

### Description

We have recently merged #1826 PR which is preventing duplicated `databasechangelog` properties in a YAML changelog, but this change made the case of having duplicated sql properties like the case displayed below:

```yaml
databaseChangeLog:
- changeSet:
    id: 4-addPartyParrotsTableRunWithSQLPlus
    author: Liquibase Pro User
    labels: external,sqlChange,rollback
    runWith: sqlplus
    changes:
    - sql:
      splitStatements: false
      runWith: sqlplus
      sql: >
       CREATE TABLE partyparrots(
           partyparrot_id NUMBER,
           first_name VARCHAR2(50) NOT NULL,
           last_name VARCHAR2(50) NOT NULL,
           PRIMARY KEY(partyparrot_id)
        );
    - rollback:
        - sql:
            sql: DROP TABLE PARTYPARROTS;
```

I think it would be good (if this not a BREAKING change) to rename the inner `sql:` property to something like `sql_statement`, `sql_body`, or something like that. That way the fix provided in #1826 would be still valid and wouldn't be breaking this other case.  

### Steps To Reproduce

Deploying a changeset like the one provided in the description.

### Expected/Desired Behavior

Do not allow duplicate properties, but without breaking other object's structures like the SQL object.

### Liquibase Version

_No response_

### Database Vendor & Version

_No response_

### Liquibase Integration

_No response_

### Liquibase Extensions

_No response_

_No response_

### Additional Context

_No response_

danielthegray:
## Impact

## Description
If someone makes a mistake and copy/pastes a databaseChangeLog block in, and makes the file have two tags, the SnakeYAML parser silently covers up the first instance with the second one, which will cause the first block of changes to be silently omitted.

This can be very problematic and cause hard-to-debug errors, which is why I have added a small FilterInputStream interceptor to check the stream as it is fed into the YAML parser, for a duplicate tag. Doing it with the FilterInputStream like this prevents the efficiency losses of opening the file twice.

I've removed the `@Ignore` flag from the test, which should now validate that the invalid file correctly throws a syntax error.

## Things to be aware of

I did it with a InputStream interceptor to not need to fully load the changeset into memory for parsing, but just check it "on-the-fly" as it flows by. This way no additional memory limitation is imposed on Liquibase and the performance profile should be the same as before.

## Things to worry about

I'm guessing it could have some breaking impact if there was a Liquibase script that was silently failing before but is now throwing the syntax error... maybe we want to disable the syntax check for already processed Liquibase changesets?

## Additional Context

I previously did this change in PR #1826 but it was suggested that I change it to just disallow duplicate keys, which has clearly not worked. I am re-submitting my initial change. Yes, it is a bit less elegant, but it will catch the duplicate databaseChangeLog case and give a warning instead of silently failing. This should fix the follow-up issue #5963.

### DIFF
--- a/liquibase-standard/src/main/java/liquibase/parser/core/yaml/YamlChangeLogParser.java
+++ b/liquibase-standard/src/main/java/liquibase/parser/core/yaml/YamlChangeLogParser.java
@@ -99,10 +99,77 @@ public class YamlChangeLogParser extends YamlParser implements ChangeLogParser {
         }
     }
 
+    /**
+     * A simple "interceptor" InputStream to ensure that the YAML stream doesn't contain two "databaseChangeLog" tags,
+     * which would otherwise cause silent omissions if something is copy/pasted wrongly into the file.
+     */
+    static class DoubleChangelogCheckerInputStream extends FilterInputStream {
+
+        byte[] databaseChangeLog = "databaseChangeLog".getBytes(StandardCharsets.UTF_8);
+        int scanPos;
+        boolean databaseChangeLogSeen;
+
+        DoubleChangelogCheckerInputStream(InputStream in) {
+            super(in);
+            scanPos = 0;
+            databaseChangeLogSeen = false;
+        }
+
+        private void processByte(int nextByte) {
+            // scanPos will be -1 if we aren't in a position where "databaseChangeLog" would cause a problem
+            // i.e. the beginning of a file or of a line
+            if (nextByte == 10) {
+                // a new line means we should expect to see it again
+                scanPos = 0;
+            } else if (scanPos > -1 && nextByte == databaseChangeLog[scanPos]) {
+                scanPos++;
+                if (scanPos == databaseChangeLog.length) {
+                    if (databaseChangeLogSeen) {
+                        throw new IllegalStateException("Two 'databaseChangeLog' tags found in a single file!");
+                    } else {
+                        databaseChangeLogSeen = true;
+                        scanPos = -1;
+                    }
+                }
+            } else {
+                // no point in scanning since the string was not found
+                scanPos = -1;
+            }
+        }
+
+        @Override
+        public int read() throws IOException {
+            int nextByte = super.read();
+            processByte(nextByte);
+            return nextByte;
+        }
+
+        @Override
+        public int read(byte[] b) throws IOException {
+            int numBytes = super.read(b);
+
+            for (int i=0; i<numBytes; i++) {
+                processByte(b[i]);
+            }
+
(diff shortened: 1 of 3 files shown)


kevin-atx:
@mariochampion suggests we add this functionality behind the `--strict` flag and validate the formatting of the changelog, but leave the current behavior intact.

MalloD12:
Hi @danielthegray,

Hey, I just want to check with you whether you are okay with applying Mario's suggestion of using  `--strict` global configuration. Please let us know if there is anything we can help you with or if you have any questions/concerns about it.

Thanks,
Daniel.

lbadam:
@filipelautert can we close this as a duplicate of #1826 ?

filipelautert:
@lbadam I"m not sure as @MalloD12 partially reverted the fix on https://github.com/liquibase/liquibase/pull/5962 .

MalloD12:
Hey @lbadam,

We should keep this one open for implementation of this duplicate tags validation using strict as discussed and suggested by @mariochampion. As @filipelautert mentioned we reverted  #1826 with #5962 because it was fixing `databaseChangeLog` tags issue, but it was making some functional tests fail for a case having two `sql` tags as shown in the below example:

```
changes:
    - sql:
      splitStatements: false
      runWith: sqlplus
      sql: >
       CREATE TABLE partyparrots(
           partyparrot_id NUMBER,
           first_name VARCHAR2(50) NOT NULL,
           last_name VARCHAR2(50) NOT NULL,
           PRIMARY KEY(partyparrot_id)
        );
```

I opened #5963 for that. I think we can close that one as it should be covered when fixing this one.  

What do you think @lbadam?

Thanks,
Daniel.

lbadam:
@MalloD12 that makes sense. I think closing #5963 would be good to reduce duplicate issues.

MalloD12:
Hi @danielthegray,

I hope you are doing very well. I just want to follow up with you to see if you are keen on applying strict flag changes to this PR so we can move forward with it.  

Thanks,
Daniel.

filipelautert:
@MalloD12 did you implement this in another PR recently?

MalloD12:
> @MalloD12 did you implement this in another PR recently?

Do you mean #6827? However, that one is not for duplicated `<databaseChangeLog>` tag; it's to prevent any other tag from being set as the root element of a changelog.

PR #5980, opened 2024-06-06
DAT-17768 - Add new directory internal/ext to classpath

filipelautert:
Changes in liquibase OSS code to add internal/ext directory as a library directory


filipelautert:
> What do you think about calling it `internal/extensions`? I find `internal/ext` to be confusing. Like internal AND external?

Sure! Originally it was "extensions", but during conversations we got lazy to say it and shortened for ext.

StevenMassaro (approved):

PR #5982, opened 2024-06-07
DAT-17770 Package BQ commercial jar inside liquibase tar ball & zip

jandroav:
## Impact


.github/workflows/create-release.yml
```diff
@@ -12,6 +12,11 @@ on:
         description: 'RunId of liquibase/liquibase artifacts to attach'
         required: true
 
+env:
+  MAVEN_VERSION: '3.9.2'
```
filipelautert, line 16: I think we don't need maven version + stCarolas down there as we are only getting dependencies with maven. we can use the one provided by image.

filipelautert (changes requested):

PR #5989, opened 2024-06-10
DAT-17896 Trigger functional and harness tests by request

rberezen:
## Impact

### DIFF
--- a/.github/workflows/run-functional-tests.yml
+++ b/.github/workflows/run-functional-tests.yml
@@ -18,7 +18,7 @@ on:
 
 jobs:
   functional-tests-build:
-    if: ${{ github.event.label.name == 'runFunctionalTests' }}
+    if: ${{ contains(github.event.pull_request.labels.*.name, 'runFunctionalTests') }}
     runs-on: ubuntu-latest
     steps:
       - uses: actions/checkout@v4
--- a/.github/workflows/run-functional-tests.yml
+++ b/.github/workflows/run-functional-tests.yml
@@ -32,7 +32,7 @@ jobs:
       - uses: actions/checkout@v4
 
       - name: Check runFunctionalTests label
-        if: contains(github.event.pull_request.labels.*.name, 'runFunctionalTests')
+        if: "!contains(github.event.pull_request.labels.*.name, 'runFunctionalTests')"
         run: |
           echo "Add the runFunctionalTests label to the PR to trigger the Functional Tests workflow."
           exit 1
--- a/.github/workflows/run-functional-tests.yml
+++ b/.github/workflows/run-functional-tests.yml
@@ -16,6 +16,7 @@ on:
 
 jobs:
   functional-tests-build:
+    if: ${{ github.event.label.name == 'runFunctionalTests' }}
     runs-on: ubuntu-latest
     steps:
       - uses: actions/checkout@v4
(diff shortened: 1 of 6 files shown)


MalloD12:
Does this mean the test execution flow will now change and if we want to trigger Test-harness and Functional Test execution we need to add their respective labels, right? Does this apply to contributor's PRs as well as internal PRs?

Thanks,
Daniel.

rberezen:
> Does this mean the test execution flow will now change and if we want to trigger Test-harness and Functional Test execution we need to add their respective labels, right? Does this apply to contributor's PRs as well as internal PRs?
> 
> Thanks, Daniel.

yes, sir.
and should be the same procedure for contributor's PRs
but unfortunately I can't test this in branch

.github/workflows/run-functional-tests.yml
```diff
@@ -29,6 +31,12 @@ jobs:
     steps:
       - uses: actions/checkout@v4
 
+      - name: Check runFunctionalTests label
+        if: contains(github.event.pull_request.labels.*.name, 'runFunctionalTests')
+        run: |
+          echo "Add the runFunctionalTests label to the PR to trigger the Functional Tests workflow."
+          exit 1
```
StevenMassaro, line 38: What is this supposed to do?

MalloD12:
I have a few more questions:

- Why are functional tests and test harness triggered without the required tags? Is that what you mentioned it wasn't possible to test in this branch? @rberezen 
- I see most of the checks are being executed twice using `pull_request` and `pull_request_target`. What does it mean each of these? Is this the expected behavior running checks for each of them? @jandroav @sayaliM0412 

Thanks,
Daniel.

.github/workflows/run-functional-tests.yml
rberezen: ah, good catch
changed to "!contains...'
so if no appropriate label - fail step and job

StevenMassaro (approved):
Looks fine to me. You could move the "Check runTestHarnessTests label" above the checkout step if you cared to.

rberezen:
> I have a few more questions:
> 
> * Why are functional tests and test harness triggered without the required tags? Is that what you mentioned it wasn't possible to test in this branch? @rberezen
> * I see most of the checks are being executed twice using `pull_request` and `pull_request_target`. What does it mean each of these? Is this the expected behavior running checks for each of them? @jandroav @sayaliM0412
> 
> Thanks, Daniel.

1. yes, can't test in branch
2. looking at it, removed   pull_request: types: [labeled] where it is not necessary...checking

filipelautert:
> Looks fine to me. You could move the "Check runTestHarnessTests label" above the checkout step if you cared to.

@rberezen +1 to @StevenMassaro comments. It will save us some extra ticks.

rberezen:
> I have a few more questions:
> 
> * Why are functional tests and test harness triggered without the required tags? Is that what you mentioned it wasn't possible to test in this branch? @rberezen
> * I see most of the checks are being executed twice using `pull_request` and `pull_request_target`. What does it mean each of these? Is this the expected behavior running checks for each of them? @jandroav @sayaliM0412
> 
> Thanks, Daniel.

@MalloD12 should be better now, no more extra run for unit/integrations tests

PR #5993, opened 2024-06-11
Move ChangeLogParameters helper methods to ChangeLogParameters class (DAT-17883)

abrackx:
## Impact

## Description
Moves both `addJavaProperties` and `addDefaultFileProperties` from the `DatabaseChangelogCommandStep` class to the `ChangeLogParameters` class. These method are useful and should be the standard way to fetch these properties even if our command pipeline isn't explicitly using DatabaseChangelogCommandStep.


MalloD12:
This looks ok to me. Just one question, would it not be useful to have this two methods in a `CommandStepUtil` class? These two methods seem to be just used in `DatabaseChangelogCommandStep` class, but there is one similar implementation for `addJavaPropertiesToChangelogParameters()` in `LiquibaseCommandLine` (maybe we can use the same implementation there as well).

abrackx:
> This looks ok to me. Just one question, would it not be useful to have this two methods in a `CommandStepUtil` class? 
What about sticking it on the `ChangeLogParameters` class itself? That makes sense to me. I'll have to leave the java props one inside this class since it's public, but can change it to call the new one. 
> there is one similar implementation for `addJavaPropertiesToChangelogParameters()` in `LiquibaseCommandLine` (maybe we can use the same implementation there as well).
That's true! But that one looks slightly different and deals with calling out to the `Main` class. That makes me nervous, idk if I want to mess with that one too. 🤣

StevenMassaro (approved):

StevenMassaro:
Are there places where this is currently done manually and should be changed to instead use these newly public methods?

abrackx:
@MalloD12 @StevenMassaro I moved these into the `ChangeLogParameters` class because that seemed even better for discoverability.

MalloD12 (approved):
LGTM

filipelautert (approved):

StevenMassaro (approved):

PR #5994, opened 2024-06-12
DAT-17896 Trigger functional tests when merging to master

rberezen:
## Impact

### DIFF
--- a/.github/workflows/run-functional-tests.yml
+++ b/.github/workflows/run-functional-tests.yml
@@ -21,7 +21,7 @@ jobs:
     runs-on: ubuntu-latest
     steps:
       - name: Check runFunctionalTests label
-        if: "!contains(github.event.pull_request.labels.*.name, 'runFunctionalTests')"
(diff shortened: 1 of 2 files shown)


jandroav (approved):

rberezen:
> Shouldn't this be `||` not `&&`?

I think with || tests will be triggered on PR everytime and will ignore label condition

StevenMassaro:
> > Shouldn't this be `||` not `&&`?
> 
> I think with || tests will be triggered on PR everytime and will ignore label condition

With `&&` will they ever get triggered on master? There won't be any labels on master.

StevenMassaro:
Maybe it should be:

```
isMaster || (!isMaster && hasLabels)
```

rberezen:
> Maybe it should be:
> 
> ```
> isMaster || (!isMaster && hasLabels)
> ```

so in my case we want to exit workflow when !isMaster and when !hasLabels
in other cases we skip the step 'name: Check runTestHarnessTests label' and execute tests

StevenMassaro (approved):
I totally missed the point

sayaliM0412 (approved):

jnewton03 (approved):

PR #5995, opened 2024-06-12
DAT-17704

sayaliM0412:
feat: `.github/workflows/run-tests.yml`:  fail the build if test count is outside the acceptable range of +/- 1% of the expected count. The values will always be updated in Github variable when there is a merge to master so there is no need to add the failure logic to notify in slack for test count. 

## Impact


.github/workflows/run-tests.yml
```diff
@@ -309,12 +377,12 @@ jobs:
...
 
   sonar:
-    needs: [ build_tests, integration-test ]
+    needs: [ setup, build_tests, integration-test ]
     uses: liquibase/build-logic/.github/workflows/sonar-test-scan.yml@v0.7.8
     with:
       thisBranchName: ${{ needs.setup.outputs.thisBranchName }}
-      thisSha: ${{ needs.setup.outputs.thisSha }}
-      liquibaseBranchName: ${{ needs.setup.outputs.liquibaseBranchName }}
+      thisSha: ${{ needs.setup.outputs.latestMergeSha }}
```
filipelautert, line 384: Are those changes expected?

.github/workflows/run-tests.yml
```diff
@@ -174,6 +174,55 @@ jobs:
...
+          $TEST_COUNT = (Get-ChildItem -Recurse -Filter "*.txt" | Select-String -Pattern 'Tests run: (\d+)' | ForEach-Object { $_.Matches.Groups[1].Value } | Measure-Object -Sum).Sum
+          Write-Output "Total Unit tests run: $TEST_COUNT"
+          if ($TEST_COUNT -eq 0) {
+            Write-Output "No Unit tests were run."
+            exit 1
+          }
+          echo "TEST_COUNT_JAVA_WINDOWS=$TEST_COUNT" | Out-File -Append -Encoding utf8 -FilePath $env:GITHUB_ENV
+
+      - name: Notify slack if Test Count decreases
+        if: env.TEST_COUNT_JAVA_WINDOWS && env.TEST_COUNT_JAVA_WINDOWS < 5999 || env.TEST_COUNT_JAVA_UBUNTU && env.TEST_COUNT_JAVA_UBUNTU < 5981 || env.TEST_COUNT_JAVA_MACOS && env.TEST_COUNT_JAVA_MACOS < 5981
```
filipelautert, line 218: Having those hardcoded counts here may be a problem, as we will need to update them from time to time. 
Ideally we would use an action similar to https://github.com/marketplace/actions/set-persistent-value , but who knows where it stores information. 
As an option, maybe we could store this info in a text file and update it when it runs on master? 
Any other ideas @StevenMassaro ?

filipelautert (changes requested):

.github/workflows/run-tests.yml
sayaliM0412: the values are not getting populated here : https://github.com/liquibase/liquibase/actions/runs/9566205290/job/26374745886#step:1:44. But I think I'll delete it for this PR.

.github/workflows/run-tests.yml
sayaliM0412: the action https://github.com/marketplace/actions/set-persistent-value is not well maintained. I'd prefer using json instead.

jandroav: we could store the values in github secrets

StevenMassaro (changes requested):
I agree with the others. This really needs to do two things:
- check how many tests were run
- if it's less (maybe by a threshold, but that could come later) than usual (on master), then fail
- if it's more, when it gets merged to master, it needs to update the expected count in github secrets (or wherever it is stored)

StevenMassaro:
You might be able to leverage this existing action from the marketplace to accomplish this: https://github.com/EnricoMi/publish-unit-test-result-action

You'd need to look into it a bit more. We used to use this action but got rid of it because it was annoying. It was removed in 4442bfa03202240e5108e54f6fd7a939bd6129ab. Perhaps it could be readded but only configured to comment and fail the build if the number of tests decreases.

sayaliM0412:
> You might be able to leverage this existing action from the marketplace to accomplish this: https://github.com/EnricoMi/publish-unit-test-result-action
> 
> You'd need to look into it a bit more. We used to use this action but got rid of it because it was annoying. It was removed in [4442bfa](https://github.com/liquibase/liquibase/commit/4442bfa03202240e5108e54f6fd7a939bd6129ab). Perhaps it could be readded but only configured to comment and fail the build if the number of tests decreases.

I think I like the script I used as it gives the exact output without any issues and is easy to understand.

StevenMassaro:
I would personally prefer the build to fail, rather than notify via slack if the test counts go down.

Also, what is the procedure for updating the test counts?

sayaliM0412:
> I would personally prefer the build to fail, rather than notify via slack if the test counts go down.
> 
> Also, what is the procedure for updating the test counts?

updating test_counts here: https://github.com/liquibase/liquibase/pull/5995/files#diff-1e14c970a1e66d52c97fb8708517e0a2ea284cc5bc9e8bcb5914c9b524c7ec4d
I think I can make the build fail, no issues there. But I thought Slack is what we like, therefore went with slack.

StevenMassaro:
> > I would personally prefer the build to fail, rather than notify via slack if the test counts go down.
> > Also, what is the procedure for updating the test counts?
> 
> updating test_counts here: https://github.com/liquibase/liquibase/pull/5995/files#diff-1e14c970a1e66d52c97fb8708517e0a2ea284cc5bc9e8bcb5914c9b524c7ec4d I think I can make the build fail, no issues there. But I thought Slack is what we like, therefore went with slack.

Maybe we need a new ticket, but I think the master build should update the test counts on each build. Otherwise, as time goes on, the numbers will drift further and further apart.

We should check with @filipelautert but my vote is build failure, not slack notification.

sayaliM0412:
> > > I would personally prefer the build to fail, rather than notify via slack if the test counts go down.
> > > Also, what is the procedure for updating the test counts?
> > 
> > 
> > updating test_counts here: https://github.com/liquibase/liquibase/pull/5995/files#diff-1e14c970a1e66d52c97fb8708517e0a2ea284cc5bc9e8bcb5914c9b524c7ec4d I think I can make the build fail, no issues there. But I thought Slack is what we like, therefore went with slack.
> 
> Maybe we need a new ticket, but I think the master build should update the test counts on each build. Otherwise, as time goes on, the numbers will drift further and further apart.
> 
> We should check with @filipelautert but my vote is build failure, not slack notification.

if the test count gets updated the test_count file would need an update as well. But let's fail the build if test count goes down for now

filipelautert:
> > > > I would personally prefer the build to fail, rather than notify via slack if the test counts go down.
> > > > Also, what is the procedure for updating the test counts?
> > > 
> > > 
> > > updating test_counts here: https://github.com/liquibase/liquibase/pull/5995/files#diff-1e14c970a1e66d52c97fb8708517e0a2ea284cc5bc9e8bcb5914c9b524c7ec4d I think I can make the build fail, no issues there. But I thought Slack is what we like, therefore went with slack.
> > 
> > 
> > Maybe we need a new ticket, but I think the master build should update the test counts on each build. Otherwise, as time goes on, the numbers will drift further and further apart.
> > We should check with @filipelautert but my vote is build failure, not slack notification.
> 
> if the test count gets updated the test_count file would need an update as well. But let's fail the build if test count goes down for now

So if I got it right, things would work like that for branch builds: 
1. test count go down -> fail
2. tests count go up -> update json file with the new value

This way master build counts should never change. And if one day we need to remove tests we can just update the json file with the new number and the branch will build correctly.

sayaliM0412:
> > > > > I would personally prefer the build to fail, rather than notify via slack if the test counts go down.
> > > > > Also, what is the procedure for updating the test counts?
> > > > 
> > > > 
> > > > updating test_counts here: https://github.com/liquibase/liquibase/pull/5995/files#diff-1e14c970a1e66d52c97fb8708517e0a2ea284cc5bc9e8bcb5914c9b524c7ec4d I think I can make the build fail, no issues there. But I thought Slack is what we like, therefore went with slack.
> > > 
> > > 
> > > Maybe we need a new ticket, but I think the master build should update the test counts on each build. Otherwise, as time goes on, the numbers will drift further and further apart.
> > > We should check with @filipelautert but my vote is build failure, not slack notification.
> > 
> > 
> > if the test count gets updated the test_count file would need an update as well. But let's fail the build if test count goes down for now
> 
> So if I got it right, things would work like that for branch builds:
> 
> 1. test count go down -> fail
> 2. tests count go up -> update json file with the new value
> 
> This way master build counts should never change. And if one day we need to remove tests we can just update the json file with the new number and the branch will build correctly.

Yes Filipe , that's right. I think having a call helped understand things clearer. 
For example, I have a PR:
1.      I add a test
2.     I update the test_count.json
3.     I run from my branch

It wont affect any other branches or builds.

StevenMassaro:
> > > > > > I would personally prefer the build to fail, rather than notify via slack if the test counts go down.
> > > > > > Also, what is the procedure for updating the test counts?
> > > > > 
> > > > > 
> > > > > updating test_counts here: https://github.com/liquibase/liquibase/pull/5995/files#diff-1e14c970a1e66d52c97fb8708517e0a2ea284cc5bc9e8bcb5914c9b524c7ec4d I think I can make the build fail, no issues there. But I thought Slack is what we like, therefore went with slack.
> > > > 
> > > > 
> > > > Maybe we need a new ticket, but I think the master build should update the test counts on each build. Otherwise, as time goes on, the numbers will drift further and further apart.
> > > > We should check with @filipelautert but my vote is build failure, not slack notification.
> > > 
> > > 
> > > if the test count gets updated the test_count file would need an update as well. But let's fail the build if test count goes down for now
> > 
> > 
> > So if I got it right, things would work like that for branch builds:
> > 
> > 1. test count go down -> fail
> > 2. tests count go up -> update json file with the new value
> > 
> > This way master build counts should never change. And if one day we need to remove tests we can just update the json file with the new number and the branch will build correctly.
> 
> Yes Filipe , that's right. I think having a call helped understand things clearer. For example, I have a PR:
> 
> 1. ```
>     I add a test
>    ```
> 2. ```
>    I update the test_count.json
>    ```
> 3. ```
>    I run from my branch
>    ```
> 
> It wont affect any other branches or builds.

I have two issues with this approach.

1. This will create merge conflicts on almost every single branch.
2. Most developers will forget to update this JSON file.

This is why I think this should be automated on the master branch, or some other approach that we come up with.

sayaliM0412:
> > > > > > > I would personally prefer the build to fail, rather than notify via slack if the test counts go down.
> > > > > > > Also, what is the procedure for updating the test counts?
> > > > > > 
> > > > > > 
> > > > > > updating test_counts here: https://github.com/liquibase/liquibase/pull/5995/files#diff-1e14c970a1e66d52c97fb8708517e0a2ea284cc5bc9e8bcb5914c9b524c7ec4d I think I can make the build fail, no issues there. But I thought Slack is what we like, therefore went with slack.
> > > > > 
> > > > > 
> > > > > Maybe we need a new ticket, but I think the master build should update the test counts on each build. Otherwise, as time goes on, the numbers will drift further and further apart.
> > > > > We should check with @filipelautert but my vote is build failure, not slack notification.
> > > > 
> > > > 
> > > > if the test count gets updated the test_count file would need an update as well. But let's fail the build if test count goes down for now
> > > 
> > > 
> > > So if I got it right, things would work like that for branch builds:
> > > 
> > > 1. test count go down -> fail
> > > 2. tests count go up -> update json file with the new value
> > > 
> > > This way master build counts should never change. And if one day we need to remove tests we can just update the json file with the new number and the branch will build correctly.
> > 
> > 
> > Yes Filipe , that's right. I think having a call helped understand things clearer. For example, I have a PR:
> > 
> > 1. ```
> >     I add a test
> >    ```
> > 2. ```
> >    I update the test_count.json
> >    ```
> > 3. ```
> >    I run from my branch
> >    ```
> > 
> > It wont affect any other branches or builds.
> 
> I have two issues with this approach.
> 
> 1. This will create merge conflicts on almost every single branch.
> 2. Most developers will forget to update this JSON file.
> 
> This is why I think this should be automated on the master branch, or some other approach that we come up with.

we can setup a Github Environment variable for the number of tests_count for each initially. And update that number whenever there is a merge in master branch ?

StevenMassaro (approved):

StevenMassaro (approved):
Looks good to me! We'll have to test it in master.

PR #6008, opened 2024-06-14
DAT-16650 Automate OWASP scanner

jandroav:
## Impact


.github/workflows/owasp-scanner.yml
```diff
@@ -0,0 +1,52 @@
...
+
+
+jobs:
+  scan:
+    runs-on: ubuntu-22.04
+    steps:
+
+     - name: Download OWASP Dependency-Check
+       run: |
+        wget https://github.com/jeremylong/DependencyCheck/releases/download/v9.2.0/dependency-check-9.2.0-release.zip
```
abrackx, line 35: I see dependencycheck has [a maven plugin](https://github.com/jeremylong/DependencyCheck?tab=readme-ov-file#maven-plugin) is there a reason we're downloading the release zip instead of using that?

filipelautert: I think it's because @jandroav automated the manual process that we did before, and it was done using this external tool. But @abrackx got a really good point! Should be add a new profile and just run it?
I think it would be easier to track owasp versions this ways as dependabot would maintain it for us.

jandroav: Yeah, good point. but where to execute it?

jandroav: also keep in mind that this workflow can work with every other future component like extensions without adding maven plugins to anywhere

filipelautert: Makes sense for me.

filipelautert (approved):

.github/workflows/create-release.yml
```diff
@@ -39,17 +39,24 @@ jobs:
       - run: |
           echo "Creating version ${{ steps.collect-data.outputs.version }} from ${{ steps.collect-data.outputs.branch }} with artifacts from build ${{ steps.collect-data.outputs.runId }} "
           
+  owasp-scanner:
+    needs: [ setup ]
+    uses: liquibase/build-logic/.github/workflows/owasp-scanner.yml@DAT-16650
```
filipelautert, line 44: Just remember to remove the DAT reference here befopre merging.

filipelautert (approved):

.github/workflows/create-release.yml
jandroav: done

PR #6010, opened 2024-06-14
Skip harness and functional tests instead of failing workflow, one trigger for harness and functional to avoid multiple triggers

rberezen:
## Impact

## Description

@wwillard7800 suggested not to fail the workflow but rather skip it, and I think he is right. Another change in the PR is to add one label, runFunctional&HarnessTests, because if we add separate labels like runFunctionalTests and runHarnessTests, the second added label will generate an extra test run for the first label.

### DIFF
--- a/.github/workflows/run-functional-tests.yml
+++ b/.github/workflows/run-functional-tests.yml
@@ -19,7 +19,7 @@ on:
 jobs:
   functional-tests-build:
     runs-on: ubuntu-latest
-    if: (github.event_name != 'pull_request' || github.event_name != 'pull_request_target') ||  github.event.label.name == 'runFunctional&HarnessTests'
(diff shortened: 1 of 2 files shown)


rberezen:
@wwillard7800 suggested not to fail the workflow but rather skip it, and I think he is right. Another change in the PR is to add one label, runFunctional&HarnessTests, because if we add separate labels like runFunctionalTests and runHarnessTests, the second added label will generate an extra test run for the first label.

wwillard7800 (approved):

StevenMassaro (approved):

filipelautert (approved):

sayaliM0412 (approved):
LGTM !

jandroav (approved):

PR #6026, opened 2024-06-19
Bump Spring-core version to 5.3.37

filipelautert:


filipelautert:
@MalloD12 dependabot always tries to upgrade this package to version 6.x and because of that it is missing the 5.x series upgrades.

MalloD12 (approved):

MalloD12:
@filipelautert makes a lot of sense to me. Yeah, we have been rejecting some other dependabot PRs for Spring 6 but I personally didn't realize we should keep upgrading to newer versions of Spring 5. Thank you for bringing this up!

PR #6036, opened 2024-06-22
feat: allow setting row-level dependency tracking in the create table statement on Oracle Database

### LINKED ISSUE #6035: Support for ROWDEPENDENCIES in CREATE TABLE statements
### Search first

### Description

I have the need to use `row-level dependency tracking` in my Oracle tables. This behavior is set at the creation of the table and cannot be modified.

Currently there is no support to do that in Liquibase (that i know of) and i don't want to have to go back to writing plain SQL changesets.

This feature is enabled by adding the keyword `ROWDEPENDENCIES` to the CREATE TABLE statement.

Example:

```sql
CREATE TABLE TEST_TABLE (
    "ID" NUMBER(38,0) NOT NULL
) ROWDEPENDENCIES;
```

Here is a link to the Oracle documentation: https://docs.oracle.com/en/database/oracle/oracle-database/19/sqlrf/CREATE-TABLE.html#GUID-F9CE0CC3-13AE-4744-A43C-EAC7A71AAAB6__CJAEEGDA

Sugestion for the feature:

```xml
<createTable tableName="test_table" rowDependencies="true">
</createTable>
```

> PS: I've already implemented this in my fork and i'll be submiting a request for review soon.

### Steps To Reproduce

Its a feature request.

### Expected/Desired Behavior

Expect to be able to create my tables in Oracle with the ROWDEPENDENCIES appended to the statement.

### Liquibase Version

4.28.0

### Database Vendor & Version

Oracle 19+

### Liquibase Integration

Maven, Spring Boot

### Liquibase Extensions

_No response_

Ubuntu

### Additional Context

_No response_

matheusviegas:
## Impact

## Description

Adding support to set row-level dependency tracking in the create table statement on Oracle Database.

Fixes #6035

### DIFF
--- a/liquibase-standard/src/main/java/liquibase/database/Database.java
+++ b/liquibase-standard/src/main/java/liquibase/database/Database.java
@@ -656,14 +656,6 @@ public interface Database extends PrioritizedService, AutoCloseable {
     }
 
     /**
-     * Does the database support the "ROWDEPENDENCIES" syntax for table creation?
-     * @return true if the "ROWDEPENDENCIES" syntax is supported for table creation, false otherwise.
-     */
-    default boolean supportsCreateRowDependencies() {
-        return false;
-    }
-
-    /**
      * Does the particular database implementation support the database changelog history feature and associated
      * table?
      * @return true if supported, false otherwise
--- a/liquibase-standard/src/main/java/liquibase/database/Database.java
+++ b/liquibase-standard/src/main/java/liquibase/database/Database.java
@@ -656,6 +656,15 @@ public interface Database extends PrioritizedService, AutoCloseable {
     }
 
     /**
+     * Does the database support the "ROWDEPENDENCIES" syntax?
+     * @param type the DatabaseObject type to be checked.
+     * @return true if the "ROWDEPENDENCIES" syntax is supported, false otherwise.
+     */
+    default boolean supportsCreateRowDependencies(Class<? extends DatabaseObject> type) {
+        return false;
+    }
+
+    /**
      * Does the particular database implementation support the database changelog history feature and associated
      * table?
      * @return true if supported, false otherwise
--- a/liquibase-standard/src/main/java/liquibase/database/Database.java
+++ b/liquibase-standard/src/main/java/liquibase/database/Database.java
@@ -656,11 +656,10 @@ public interface Database extends PrioritizedService, AutoCloseable {
     }
 
     /**
-     * Does the database support the "ROWDEPENDENCIES" syntax?
-     * @param type the DatabaseObject type to be checked.
-     * @return true if the "ROWDEPENDENCIES" syntax is supported, false otherwise.
+     * Does the database support the "ROWDEPENDENCIES" syntax for table creation?
+     * @return true if the "ROWDEPENDENCIES" syntax is supported for table creation, false otherwise.
      */
-    default boolean supportsCreateRowDependencies(Class<? extends DatabaseObject> type) {
+    default boolean supportsCreateRowDependencies() {
         return false;
     }
 
--- a/liquibase-standard/src/main/java/liquibase/database/core/OracleDatabase.java
+++ b/liquibase-standard/src/main/java/liquibase/database/core/OracleDatabase.java
@@ -273,11 +273,6 @@ public class OracleDatabase extends AbstractJdbcDatabase {
     }
 
     @Override
-    public boolean supportsCreateRowDependencies() {
-        return true;
-    }
-
-    @Override
     public boolean isReservedWord(String objectName) {
         return reservedWords.contains(objectName.toUpperCase());
     }
--- a/liquibase-standard/src/main/java/liquibase/database/core/OracleDatabase.java
+++ b/liquibase-standard/src/main/java/liquibase/database/core/OracleDatabase.java
@@ -273,6 +273,11 @@ public class OracleDatabase extends AbstractJdbcDatabase {
     }
 
     @Override
+    public boolean supportsCreateRowDependencies(Class<? extends DatabaseObject> type) {
+        return true;
+    }
+
+    @Override
     public boolean isReservedWord(String objectName) {
         return reservedWords.contains(objectName.toUpperCase());
     }
--- a/liquibase-standard/src/main/java/liquibase/database/core/OracleDatabase.java
+++ b/liquibase-standard/src/main/java/liquibase/database/core/OracleDatabase.java
@@ -273,7 +273,7 @@ public class OracleDatabase extends AbstractJdbcDatabase {
     }
 
     @Override
-    public boolean supportsCreateRowDependencies(Class<? extends DatabaseObject> type) {
+    public boolean supportsCreateRowDependencies() {
         return true;
     }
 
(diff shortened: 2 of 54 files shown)


liquibase-standard/src/main/java/liquibase/database/core/OracleDatabase.java
```diff
@@ -272,6 +272,11 @@ public boolean supportsInitiallyDeferrableColumns() {
         return true;
     }
 
+    @Override
+    public boolean supportsCreateRowDependencies(Class<? extends DatabaseObject> type) {
```
MalloD12, line 276: Do we need the`type` parameter for this method?  If we do, we should check whether it's a Table type. Otherwise, we can pass any other type here and it will never return `true`.

liquibase-standard/src/test/java/liquibase/sqlgenerator/core/CreateTableGeneratorTest.java
```diff
@@ -1255,4 +1255,17 @@ public void testWithCreateIfNotExists() {
             }
```
MalloD12, line 1255: I think it would be nice if we could also add a test changeset into `https://github.com/liquibase/liquibase/blob/master/liquibase-integration-tests/src/test/resources/changelogs/oracle/complete/root.changelog.xml` with a create table change that include `rowDependencies` property. That way, at least we can make sure whether liquibase will process and deploy it successfully or not.

MalloD12:
@tati-qalified, would you mind to manually test this change when you have some time?

Thanks,
Daniel.

liquibase-standard/src/test/java/liquibase/sqlgenerator/core/CreateTableGeneratorTest.java
matheusviegas: Sure, i've added a changeset to the `root.changelog.xml` file for Oracle.

liquibase-standard/src/main/java/liquibase/database/Database.java
```diff
@@ -655,6 +655,15 @@
         return false;
     }
 
+    /**
+     * Does the database support the "ROWDEPENDENCIES" syntax?
+     * @param type the DatabaseObject type to be checked.
+     * @return true if the "ROWDEPENDENCIES" syntax is supported, false otherwise.
+     */
+    default boolean supportsCreateRowDependencies(Class<? extends DatabaseObject> type) {
```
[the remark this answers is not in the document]
matheusviegas: Parameter removed.

MalloD12 (approved):
**Approved**.

Code changes look good to me. 

Thank you, @matheusviegas!

tati-qalified:
Hi @matheusviegas, I tested this using the CLI and I wasn't able to get the extra column on the created table. I used the changeset you added to the tests.

Could you specify the steps you took when you tested it, so that I can verify if I'm missing something?
Thanks!

matheusviegas:
Hi @tati-qalified, thanks for taking the time to test this changes!

The changes that i've made in this PR don't create a extra column, but instead adds a modifier to the created table that makes Oracle track dependencies at row level.

If you query the metadata for a table created with the default configuration in Oracle from the `all_tables` view with this SQL:

```sql
select dependencies from all_tables where lower(table_name) = lower('TABLE_NAME_HERE');
```

The return value of the column `dependencies` in this query should be `DISABLED` (which is the default behavior in Oracle).

For a table created with `rowDependencies=true` in the changeset, when you run the query above, it should return `ENABLED`.

If you need more information, let me know.

tati-qalified:
@matheusviegas thank you for clarifying!
I was able to effectively test it, and in general it works as expected.

First I'll give a quick summary and then go into specifics of what didn't work and other suggestions.

format \ db  | oracle | h2 | postgres
-- | -- | -- | --
.xml | ✅ | ✅ | ✅
.yml | ✅ | ✅ | ✅
.json | ✅ | ✅ | ✅
.sql | ❌ | ❌ | ❌

The only format that doesn't work is .sql files: using `--changeset tester:1 rowDependencies:true` doesn't fail, but it doesn't include `rowDependencies` at the end of the statement.

For the dbs that aren't Oracle, I checked that `rowDependencies` **wasn't** included.

________________________________

On another note, looking at the docs I found that there's an opposite clause, `noRowDependencies`. I think this would be very useful for rollbacks and for `alterTable` statements. Do you think you could implement something with this?

Thank you!
Tatiana

matheusviegas:
Hi @tati-qalified!

About the test results for .sql files.. could you help me replicate the test that y've made? I have never used .sql files in Liquibase nor the CLI. So if you could help me replicate this test, that woul help a lot.

About the opposite clause `NOROWDEPENDENCIES`, its the default behavior in Oracle when not specified, so i didn't include explicitly in the generated statement when rowDependencies=false in the changeset, in order to not alter the statement unless necessary.

Regarding rollbacks and alterTable, this behavior `ROWDEPENDENCIES` is set at table creation and cannot be modified with alter table statements.

In the issue where i approach this subject, i explain a little about the problem i'm trying to solve with this PR. Here is the link: https://github.com/liquibase/liquibase/issues/6035

If you need more information, let me know.

tati-qalified:
@matheusviegas after discussing with the team, we're comfortable with keeping this PR as is, without adding the attribute to sql changesets.

This looks good to me! Thank you again for your contribution to Liquibase :)

tati-qalified (approved):

liquibase-standard/src/test/resources/liquibase/change/ChangeDefinitionTest.yaml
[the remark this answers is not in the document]
matheusviegas: Sure, i've altered this file to specify that is supported only for Oracle so far.

```yml
rowDependencies boolean 
Description: When true, creates the table with ROWDEPENDENCIES
Supported: oracle
```

Thanks for testing this PR!

liquibase-standard/src/main/java/liquibase/database/Database.java
```diff
@@ -655,6 +655,14 @@ default boolean supportsCreateIfNotExists(Class<? extends DatabaseObject> type)
         return false;
     }
 
+    /**
+     * Does the database support the "ROWDEPENDENCIES" syntax for table creation?
+     * @return true if the "ROWDEPENDENCIES" syntax is supported for table creation, false otherwise.
+     */
+    default boolean supportsCreateRowDependencies() {
```
filipelautert, line 662: I guess we don't need this new supports method as this is only supported by Oracle and this class already has plenty of `supports` methods.  Just adding `instanceof OracleDatabase` where it would be called is enough.

liquibase-standard/src/main/java/liquibase/sqlgenerator/core/CreateTableGenerator.java
```diff
@@ -403,6 +403,11 @@ public Sql[] generateSql(CreateTableStatement statement, Database database, SqlG
         if ((database instanceof MySQLDatabase) && (statement.getRemarks() != null)) {
             sql += " COMMENT='" + database.escapeStringForDatabase(statement.getRemarks()) + "' ";
         }
+
+        if (statement.isRowDependencies() && database.supportsCreateRowDependencies()) {
```
filipelautert, line 407: Replace by instance of as in line 403 would be fine .

filipelautert (changes requested):

liquibase-standard/src/main/java/liquibase/database/Database.java
matheusviegas: Done.

liquibase-standard/src/main/java/liquibase/sqlgenerator/core/CreateTableGenerator.java
matheusviegas: Done.

filipelautert (approved):
Thanks for the PR @matheusviegas !

rberezen (approved):

adrian-velonis1:
[DOCUMENTATION] https://datical.atlassian.net/browse/PD-5283

PR #6039, opened 2024-06-25
command argument aliases that can be printed to help (DAT-17870)

StevenMassaro:
## Impact

## Description

Add ability for command argument aliases to be printed to the help output.


liquibase-cli/src/main/java/liquibase/integration/commandline/LiquibaseCommandLine.java
```diff
@@ -976,14 +977,17 @@ private void addSubcommand(CommandDefinition commandDefinition, CommandLine root
                     if (def.isRequired()) {
                         description = "[REQUIRED] " + description;
                     }
+                    if (!def.getForcePrintedAliases().isEmpty()) {
+                        String pluralized = def.getForcePrintedAliases().size() > 1 ? "aliases" : "alias";
+                        description = pluralized + ": '" + StringUtils.join(def.getForcePrintedAliases().stream().map(LiquibaseCommandLine::convertArgumentNameToKabobCase).collect(Collectors.toList()), "', '") + "'\n" + description;
```
abrackx, line 982: Can you use `Collectors.joining`?

abrackx (approved):
Nice, lgtm!

liquibase-cli/src/main/java/liquibase/integration/commandline/LiquibaseCommandLine.java
StevenMassaro: Good call

liquibase-standard/src/main/java/liquibase/command/CommandArgumentDefinition.java
```diff
@@ -28,16 +27,66 @@ public class CommandArgumentDefinition<DataType> implements Comparable<CommandAr
...
+     *  of
+     *  but
+     *  can be explicitly with
+     * .
+     */
+    @Getter
     private String defaultValueDescription;
+    /**
+     * -- GETTER --
+     *  Function for converting values set in underlying
```
filipelautert, line 76: Those auto-generated comments are broken - better to get rid or fix them. For instance, this one should be
```
    /**
     * Function for converting values set in underlying {@link liquibase.configuration.ConfigurationValueProvider}s into the
     * type needed for this command.
     */
```

filipelautert (changes requested):

liquibase-standard/src/main/java/liquibase/command/CommandArgumentDefinition.java
StevenMassaro: That's weird. I just assumed that the lombok plugin knew what it was doing. Nice catch. I just pushed up changes to the 3 that appear to be broken. I can also completely revert this if needed.

filipelautert (approved):

Issue #6040, opened 2024-06-25
OSGi header missing for ValidatingVisitorGenerator


steinarb:
### Search first

- [X] I searched and no similar issues were found

### Description

When bumping the liquibase version from 4.27.0 to 4.28.0, the integration test failed with the following stack trace:
```
2024-06-25T21:24:02,473 | ERROR | features-3-thread-1 | SampleDbLiquibaseRunner          | 87 - no.priv.bang.karaf.liquibase.sample - 4.27.1.SNAPSHOT | Error creating sampleapp test database schema
liquibase.exception.CommandExecutionException: java.lang.NullPointerException: Cannot invoke "liquibase.changelog.visitor.ValidatingVisitorGenerator.generateValidatingVisitor(java.util.List)" because "generator" is null
	at liquibase.command.CommandScope.execute(CommandScope.java:257) ~[?:?]
	at no.priv.bang.karaf.liquibase.runner.LiquibaseClassPathChangeLogRunner.lambda$applyLiquibaseChangelist$0(LiquibaseClassPathChangeLogRunner.java:56) ~[?:?]
	at liquibase.Scope.child(Scope.java:200) ~[?:?]
	at liquibase.Scope.child(Scope.java:176) ~[?:?]
	at no.priv.bang.karaf.liquibase.runner.LiquibaseClassPathChangeLogRunner.applyLiquibaseChangelist(LiquibaseClassPathChangeLogRunner.java:53) ~[?:?]
	at no.priv.bang.karaf.sample.db.liquibase.test.SampleDbLiquibaseRunner.applyLiquibaseChangelist(SampleDbLiquibaseRunner.java:69) ~[?:?]
	at no.priv.bang.karaf.sample.db.liquibase.test.SampleDbLiquibaseRunner.prepare(SampleDbLiquibaseRunner.java:52) ~[?:?]
	at org.ops4j.pax.jdbc.config.impl.DataSourceRegistration.<init>(DataSourceRegistration.java:88) ~[?:?]
	at org.ops4j.pax.jdbc.config.impl.DataSourceConfigManager.lambda$null$4(DataSourceConfigManager.java:95) ~[?:?]
	at org.ops4j.pax.jdbc.config.impl.ServiceTrackerHelper$1.addingService(ServiceTrackerHelper.java:132) ~[?:?]
[...]

PR #5498, opened 2024-01-23
Add missing OSGI settings

filipelautert:
## Impact

- [X] Bug fix (non-breaking change which fixes expected existing functionality)
- [ ] Enhancement/New feature (adds functionality without impacting existing logic)
- [ ] Breaking change (fix or feature that would cause existing functionality to change)
 

## Description
 Fixes https://github.com/liquibase/liquibase/issues/5357

PR #6045, opened 2024-06-27
Add osgi.serverloader manifest header entries for liquibase.changelog.visitor.ValidatingVisitorGenerator #6040 fix

steinarb:
## Impact

## Description

Liquibase 4.28.0 failing under OSGi because the MANIFEST.MF of liquibase-core does not contain and for liquibase.changelog.visitor.ValidatingVisitorGenerator (similar to what was fixed in https://github.com/liquibase/liquibase/pull/5498 )

## Things to be aware of

The bug only takes place under OSGi (with the osgi.serviceloader loading SPI services).

The fix also only has any effect under OSGi.

### DIFF
--- a/liquibase-standard/pom.xml
+++ b/liquibase-standard/pom.xml
@@ -270,7 +270,8 @@
                             osgi.serviceloader; osgi.serviceloader=liquibase.resource.PathHandler,
                             osgi.serviceloader; osgi.serviceloader=liquibase.report.ShowSummaryGenerator,
                             osgi.serviceloader; osgi.serviceloader=liquibase.parser.LiquibaseSqlParser,
-                            osgi.serviceloader; osgi.serviceloader=liquibase.changeset.ChangeSetService
+                            osgi.serviceloader; osgi.serviceloader=liquibase.changeset.ChangeSetService,
+                            osgi.serviceloader; osgi.serviceloader=liquibase.changelog.visitor.ValidatingVisitorGenerator
                         </Provide-Capability>
                         <Require-Capability>
                             osgi.extender; filter:="(osgi.extender=osgi.serviceloader.registrar)",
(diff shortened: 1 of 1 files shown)


filipelautert (approved):
@steinarb thanks again! Do you think we can write some kind of test to make sure we don't break it anymore?

steinarb:
>>>>> filipe ***@***.***>:

> @steinarb thanks again! Do you think we can write some kind of test to make sure we don't break it anymore?

I think it's possible and will give it a shot, but I will wait until the
release of karaf 4.4.7, because there is a fix to Apache Aries SPI Fly
in there (make SPI Fly start at a low OSGi start-level, like its
documentation recommends).

The test will consist of two maven modules: one maven module defining a
pax jdbc-config DataSource together with a PreHook OSGi component to run
liquibase change lists, and the other module a pax exam/karaf test to
try loading in OSGi).

If I can make a test that will run with liquibase 4.27.0 and fail on
4.28.0, I have nailed it.

Where in the liquibase tree should such a test live, if I were to
provide a PR??

filipelautert:
> Where in the liquibase tree should such a test live, if I were to provide a PR??

@steinarb thanks! And well... as seems you will need Karaf dependency it would be interesting to add it under, around here -> https://github.com/liquibase/liquibase/tree/master/liquibase-integration-tests/src/test . We prefer groovy tests, but if it's not possible a java integration one would be fine too.

steinarb:
>>>>> filipe ***@***.***>:

>> Where in the liquibase tree should such a test live, if I were to provide a PR??
> @steinarb thanks! And well... as seems you will need Karaf dependency it would be interesting to add it under, around here -> https://github.com/liquibase/liquibase/tree/master/liquibase-integration-tests/src/test . We prefer groovy tests, but if it's not possible a java integration one would be fine too. 

The test itself can possibly live there? (it requires some Junit 4
dependencies because pax exam has not been ported to Junit 5)

But I need to build the artifact being loaded outside of this maven
project (I have experiemented and I need some artifacts deployed to
~/.m2/repository during the maven install phase. So either the tests
have to run at a later phase than install (hard and with unforseen side
effects) or the artifact being deployed has to be built into a different
maven module and referred to with maven dependencies)

I tried to do it in a single module maven project, but no luck.

rberezen (approved):

filipelautert:
> But I need to build the artifact being loaded outside of this maven project (I have experiemented and I need some artifacts deployed to ~/.m2/repository during the maven install phase. So either the tests have to run at a later phase than install (hard and with unforseen side effects) or the artifact being deployed has to be built into a different maven module and referred to with maven dependencies) I tried to do it in a single module maven project, but no luck.

So maybe it would need to be a separated project that builds against liquibase master version?

I'll merge this PR but we can continue the  discussion here.

PR #6046, opened 2024-06-27
feat: add current java version to Banner

filipelautert:
## Impact

Adding current java version to Liquibase banner in a subtle manner so this information is available to help debug support issues.

Sample output:
image


StevenMassaro (approved):
Good idea

PR #6073, opened 2024-07-04
fix mysql datetime type according to the column size

### LINKED ISSUE #6069: ReadDataType for Mysql datetime
### Search first

### Description

`CommandScope commandScope = new CommandScope("diffChangeLog");`

When executing this code to create a changeset.xml file by comparing two database tables and applying the file, the following error occurs:
```
Caused by: liquibase.exception.DatabaseException: Invalid default value for 'reg_ymdt' [Failed SQL: (1067) CREATE TABLE create_ssht (reg_ymdt datetime DEFAULT CURRENT_TIMESTAMP(3) on update CURRENT_TIMESTAMP(3) NULL)]
	at liquibase.executor.jvm.JdbcExecutor$ExecuteStatementCallback.doInStatement(JdbcExecutor.java:476)
```

Upon checking, it seems that the issue arises because the actual MySQL type is `datetime(3)`, but the changeset.xml is created as follows. `type="datetime`
```
<createTable tableName="create_ssht">
    <column defaultValueComputed="CURRENT_TIMESTAMP(3) on update CURRENT_TIMESTAMP(3)" name="reg_ymdt" type="datetime"/>
</createTable>
```

#### Step 1:
Create the schema with mysql database
```
CREATE TABLE `create_ssht` (
   `reg_ymdt` datetime(3) DEFAULT CURRENT_TIMESTAMP(3) ON UPDATE CURRENT_TIMESTAMP(3)
 ) ENGINE=InnoDB DEFAULT CHARSET=utf8mb4 COLLATE=utf8mb4_bin
```

#### Step 2:
`CommandScope commandScope = new CommandScope("snapshot");`
Generate changeset.xml file with the given command.

#### Step 3:
`CommandScope updateScope = new CommandScope("update");`
Apply the changeset.xml file to the MySQL database using the following command.

### Expected/Desired Behavior

It should be type="datetime(3)" instead of type="datetime".

### Liquibase Version

4.28.0

### Database Vendor & Version

Mysql & 8.0.35

### Liquibase Integration

gradle, spring boot

### Liquibase Extensions

_No response_

_No response_

### Additional Context

_No response_

diqksrk:
## Impact

## Description
Fixes #6069

In the case of MySQL, the `datetime` and `timestamp` types are affected by the column size. The type, such as datetime, datetime(1), or datetime(2), is determined according to the column size, not decimal_digits.

Therefore, I am submitting a PR that determines the type of datetime based on the column size as follows

Here are some additional materials.
https://forums.mysql.com/read.php?3,506409,506409
image

https://dev.mysql.com/doc/refman/8.4/en/storage-requirements.html
image


liquibase-standard/src/main/java/liquibase/database/core/MySQLDatabase.java
```diff
@@ -273,6 +271,25 @@ public int getDatabasePatchVersion() throws DatabaseException {
 
     }
 
+    public Integer getTimeStampFSPWith(int columnSize) {
```
tati-qalified, line 274: I think this could be improved in terms of efficiency. I propose adding an "if" statement and then just returning columnSize%10.

It would be something along the lines of:
```
public Integer getTimeStampFSPWith(int columnSize) {
    if (columnSize > 20 && columnSize < 27) {
        return columnSize % 10
    }
    return 0
}
```

I'm unaware of the lower and upper limits on the columnSize, so it might not need both verifications.

diqksrk: Thank you very much for your advice.
I tested the method and confirmed that it works correctly.
After changing to this method, I proceeded to commit again.

liquibase-standard/src/test/groovy/liquibase/snapshot/jvm/ColumnSnapshotGeneratorTest.groovy
```diff
@@ -49,6 +50,22 @@ class ColumnSnapshotGeneratorTest extends Specification {
         dataType.getTypeName() == "varchar"
     }
 
+    def "ReadDataType specifies column size for mysql datetime data fsp"() {
```
MalloD12, line 53: I think we can customize a bit this test and make it lightly generic by parameterizing `database`, `columnSize`, and `fps`. Then we can list different scenarios with different values, and test different limits and DBs. At this point, this change is applicable for MySQL and also for MariaDB (which extends from MySQL) and the test result would be the same for either of them, but if there is any future change or any method override it would be nice to be testing the two DBs separately. If you need some help with Spock details we can assist you with it.

liquibase-standard/src/main/java/liquibase/snapshot/jvm/ColumnSnapshotGenerator.java
```diff
@@ -501,6 +501,10 @@
                 );
             }
 
+            if (database instanceof MySQLDatabase) {
+                type.setColumnSize(((MySQLDatabase)database).getTimeStampFSPWith(columnSize));
```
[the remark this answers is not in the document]
MalloD12: I think it would be nice to add a null check on this `if` condition.

MalloD12:
@rberezen - @tati-qalified: I think besides the comment I added about the existent test of this PR we are good with the test coverage of this change. I was thinking additionally that we can add an integration test to generate a snapshot and then validate type is what we expect, but maybe it's not worth. What do you think?

@diqksrk: if i'm not wrong this is only applicable for `datetime`. But what about time and timestamp or any other type that we can specify a `FSP`? Should we not consider making a type distinction in the given method (`getTimeStampFSPWith`)?
And one more thing, considering we are returning the FSP I was wondering whether we should change this method name to something like `getFSPFromTimeType`?

Thank you, guys!

rberezen:
> @rberezen - @tati-qalified: I think besides the comment I added about the existent test of this PR we are good with the test coverage of this change. I was thinking additionally that we can add an integration test to generate a snapshot and then validate type is what we expect, but maybe it's not worth. What do you think?
> 
> @diqksrk: if i'm not wrong this is only applicable for `datetime`. But what about time and timestamp or any other type that we can specify a `FSP`? Should we not consider making a type distinction in the given method (`getTimeStampFSPWith`)? And one more thing, considering we are returning the FSP I was wondering whether we should change this method name to something like `getFSPFromTimeType`?
> 
> Thank you, guys!

I think we do not need extra tests here.

diqksrk:
> @rberezen - @tati-qalified: I think besides the comment I added about the existent test of this PR we are good with the test coverage of this change. I was thinking additionally that we can add an integration test to generate a snapshot and then validate type is what we expect, but maybe it's not worth. What do you think?
> 
> @diqksrk: if i'm not wrong this is only applicable for `datetime`. But what about time and timestamp or any other type that we can specify a `FSP`? Should we not consider making a type distinction in the given method (`getTimeStampFSPWith`)? And one more thing, considering we are returning the FSP I was wondering whether we should change this method name to something like `getFSPFromTimeType`?
> 
> Thank you, guys!

First, I apologize for the delayed response.

In MySQL, the date types that can have an FSP value are TIME, DATETIME, and TIMESTAMP. This can be confirmed in the following document: [MySQL Documentation on Fractional Seconds](https://dev.mysql.com/doc/refman/8.4/en/fractional-seconds.html).

Among these, the types mapped to java.sql.Types.TIMESTAMP = 93 are DATETIME and TIMESTAMP.

As you mentioned, the data type corresponding to java.sql.Types.TIME = 92 is TIME. Both types use FSP to calculate the Column Size.

The way to calculate column size is divided into:

1. TIME
2. DATETIME and TIMESTAMP

Currently, I have implemented the calculation method for the second part. As you suggested, I will modify the method to accept a variable indicating whether it is TIME or DATETIME/TIMESTAMP, and then return the column size accordingly for each type.

I will also investigate and correct any error cases related to the first part 'TIME', and commit the changes.

Thank you always for your kind guidance!

diqksrk:
@MalloD12 
As you suggested, I have modified the method to accept jdbctype as a variable.

I will investigate if there are any error cases related to Time, and if any are found, I will register a new issue and inform you.

Thank you.

liquibase-standard/src/main/java/liquibase/snapshot/jvm/ColumnSnapshotGenerator.java
```diff
@@ -501,6 +501,10 @@ protected DataType readDataType(CachedRow columnMetadataResultSet, Column column
                 );
             }
 
+            if (database instanceof MySQLDatabase) {
+                type.setColumnSize(((MySQLDatabase) database).getFSPFromTimeType(columnSize, Types.TIMESTAMP));
```
MalloD12, line 505: We need to add a null check for `columnSize` here. That's being reported as an alert from our CodeQL scan. Could you please do that?

Thanks,
Daniel.

liquibase-standard/src/test/groovy/liquibase/snapshot/jvm/ColumnSnapshotGeneratorTest.groovy
MalloD12: This would be also nice to do, but it's not mandatory. I can do it if you don't mind.

Thanks,
Daniel.

diqksrk:
@MalloD12 
I have added a null check for columnSize. Additionally, if it is null, it will be set to the default value.

For example, in the case of DATETIME, it can be DATETIME(0), DATETIME(1), DATETIME(2), and so on up to DATETIME(6). Here, it will be set to the default value of 0

MalloD12 (approved):
**Approved**.

Thanks, @diqksrk for this PR and for addressing our review comments. Code changes look good to me. I made a little change to the test you provided here and make it parameterized.

Thanks,
Daniel.

filipelautert (approved):

tati-qalified (approved):
Verified that this works correctly for all 4 changelog formats.
If the existing table has a column of type `datetime`, with no precision specified, the generated changelog will show `datetime(0)`, which is equivalent.

Issue #1777, opened 2021-03-28
Logger API compatibility broke starting in 3.6.0


garydgregory:
## Environment
**Liquibase Version**: 3.6.0 through 4.3.2

**Liquibase Integration & Version**:  All

**Liquibase Extension(s) & Version**: `liquibase.logging.core.AbstractLogger`

**Database Vendor & Version**: N/A

**Operating System Type & Version**: N/A

## Description
Since 3.5.5 there have been several API compatibility breaks in the logging APIs:

* 3.6 removed the buildMessage() function
* 4.0 split Logger and LoggerService apart
* 4.0 introduced the new ServiceLoader setup requirement
* Perhaps others

One place specifically that this is being felt is in the log4j-2 plugin. The goal of the plugin is to switch the logger used from the default to log4j. An example configuration is like this:

```
               <dependency>
                       <groupId>org.apache.logging.log4j</groupId>
                       <artifactId>log4j-api</artifactId>
                       <version>2.14.0</version>
                       <scope>provided</scope>
               </dependency>
               <dependency>
                       <groupId>org.apache.logging.log4j</groupId>
                       <artifactId>log4j-core</artifactId>
                       <version>2.14.0</version>
                       <scope>provided</scope>
               </dependency>
                <dependency>
                       <groupId>org.apache.logging.log4j</groupId>
                       <artifactId>log4j-liquibase</artifactId>
                       <version>2.14.0</version>
                </dependency>
                <dependency>
                       <groupId>org.liquibase</groupId>
                       <artifactId>liquibase-core</artifactId>
                       <version>4.3.1</version>
                </dependency>
```

* See https://forum.liquibase.org/t/liquibase-logging-with-log4j2/5202
* See https://github.com/apache/logging-log4j2/commits/release-2.x

From a log4j standpoint, they don't know what version of Liquibase they are running against and would like to work automatically with any version.

## Steps To Reproduce
* Clone https://github.com/apache/logging-log4j2/commits/release-2.x
* Change the Liquibase version in the root `pom.xml`
* Run `mvn clean install` to install the snapshot log4j libraries
* Watch compilation break: `mvn clean compile -pl log4j-liquibase`

## Expected/Desired Behavior
Don't break compilation.

JonBetts:
[...]

PR #6080, opened 2024-07-08
RawSqlStatement Liquibase PRO Refactoring 

MalloD12:
## Impact

## Description

Functional tests were failing because of the changes made on [#1777](https://github.com/liquibase/liquibase-pro/pull/1777), so we found out these changes were required to fix those failing tests.


liquibase-standard/src/main/java/liquibase/executor/jvm/JdbcExecutor.java
```diff
@@ -188,7 +188,27 @@ public void execute(final SqlStatement sql, final List<SqlVisitor> sqlVisitors)
     private void setParameters(final PreparedStatement pstmt, final RawParameterizedSqlStatement sql) throws SQLException {
         final List<Object> parameters = sql.getParameters();
         for (int i = 0; i < parameters.size(); i++) {
-            pstmt.setObject(i + 1, parameters.get(i));
+            Object parameter = parameters.get(i);
+            if(parameter instanceof ArrayList){
+                int finalI = i;
+                ((ArrayList<?>) parameter).forEach(param -> {
+                    try {
+                        if (param instanceof String) {
```
filipelautert, line 196: The same logic is being repeated inside here and in the else.. could you create a private method to handle it and call it in both places?

filipelautert (changes requested):

PR #6094, opened 2024-07-11
[DAT-17993] Improve diffChangelog between MSSQL and Oracle

filipelautert:
* SQL Server int* types without precision docs: https://learn.microsoft.com/en-us/sql/t-sql/data-types/int-bigint-smallint-and-tinyint-transact-sql?view=sql-server-ver16
* Oracle int to number(*,) behavior: https://docs.oracle.com/en/database/oracle/oracle-database/18/sqlrf/Data-Types.html#GUID-7690645A-0EE3-46CA-90DE-C96DF5A01F8F, https://asktom.oracle.com/ords/asktom.search?tag=datatype-coversion, https://forums.oracle.com/ords/apexds/post/why-does-integer-type-change-to-number-when-alter-a-table-5269, https://stackoverflow.com/questions/54164445/i-choose-int-for-column-datatype-but-number-is-saved, 
* Only oracle supports PK validate parameter, so we should use it to compare only when we have 2 Oracle instances


liquibase-standard/src/main/java/liquibase/diff/output/changelog/core/ChangedPrimaryKeyChangeGenerator.java
```diff
@@ -123,4 +118,18 @@ public Change[] fixChanged(DatabaseObject changedObject, ObjectDifferences diffe
...
+        if (clusteredDiff != null && ((clusteredDiff.getReferenceValue() == null) || (clusteredDiff.getComparedValue() == null))) {
+            differences.removeDifference("clustered");
+        }
+
+        // as only oracle supports PK validate, we will use it only for Oracle vs Oracle comparisons
+        if (differences.getDifference("validate") != null &&
+            (!(comparisonDatabase instanceof OracleDatabase) || (!(referenceDatabase instanceof OracleDatabase)))) {
+            differences.removeDifference("validate");
+        }
+    }
```
StevenMassaro, line 134: Maybe make some global constants for "clustered" and "validate".

StevenMassaro (approved):

suryaaki2 (approved):

StevenMassaro:
Are there any tests which cover these changes? If not, can one or some be added?

mmelnik:
@filipelautert Hi, Filipe,

I've recently switched to liquibase-core 4.29.2 from 4.27.0 in my project and found that Oracle diff calculation is broken because of changes in this pull reuqest, especially because of this implementation of `OracleDatabase.correctObjectName()`:
```
@Override
public String correctObjectName(String objectName, Class<? extends DatabaseObject> objectType) {
    // Internally Oracle converts int/integer to number, so when we need to extract and compare it we need number too
    if (objectType.equals(Column.class) && StringUtils.startsWithIgnoreCase(objectName, "int")) {
        return "NUMBER(*, 0)";
    }
    return super.correctObjectName(objectName, objectType);
}
```
This method automatically converts all column names that starts from int*, like "internalPhoneNumber" to "NUMBER(*, 0)" and brakes Metadata comparison so diff generation is incorrect.

Exact thing is `StringUtils.startsWithIgnoreCase(objectName, "int")` - I confused a little bit why we're unconditionally change `objectName` which can be the column name itself and not the COLUMN TYPE name?

I might be wrong, but this looks like bug to me. Could you or someone from the team help me to clarify that, please?

PR #6099, opened 2024-07-12
chore: include slf4j-simple and pin version to allow dependabot to upgrade it for us

filipelautert:
Include one slf4j implementation as sqlite requires slf4j, so we add simple to prevent log warning messages.

We also add it here to be able to manage the version for it pin version to allow dependabot to upgrade it for us.


StevenMassaro:
We could silence this slf4j message by setting the reporter level to error: https://www.slf4j.org/apidocs/org/slf4j/helpers/Reporter.html

rberezen:
@filipelautert could you please fix this? 
`Comparing the contents of the tar.gz with the expected contents (using baseline file in liquibase-dist directory) to ensure that no new files were accidentally added. A failure here indicates that a new file was added to the tar.gz, and it should either be rectififed or added to the baseline.`

rberezen (approved):

filipelautert:
> We could silence this slf4j message by setting the reporter level to error: https://www.slf4j.org/apidocs/org/slf4j/helpers/Reporter.html

As discussed this is only available in master branch/v2 and we still on 1.7.x because the driver requires this version to work :/

PR #6109, opened 2024-07-17
DAT-18191 and DAT-18192 - Fix classpath loading and incorrect configuration for addForeignKeyConstraint

filipelautert:
The calls being modified here should not be in those locations as they happen before all classloading is complete, thus causing incorrect evaluations of available classes.
 
To be more specific, DatabaseFactory is called before classpath parameter is evaluated by command line, causing extension that provide support to other databases that are pointed out by "classpath" parameter to not be loaded.

Fix https://github.com/liquibase/liquibase-docs/issues/105

### DIFF
--- a/liquibase-standard/src/main/java/liquibase/change/core/AddForeignKeyConstraintChange.java
+++ b/liquibase-standard/src/main/java/liquibase/change/core/AddForeignKeyConstraintChange.java
@@ -2,7 +2,6 @@ package liquibase.change.core;
 
 import liquibase.change.*;
 import liquibase.database.Database;
-import liquibase.database.DatabaseFactory;
 import liquibase.exception.UnexpectedLiquibaseException;
 import liquibase.snapshot.SnapshotGeneratorFactory;
 import liquibase.statement.SqlStatement;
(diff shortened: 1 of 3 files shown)


liquibase-standard/src/main/java/liquibase/change/core/AddForeignKeyConstraintChange.java
```diff
@@ -55,24 +51,6 @@ public class AddForeignKeyConstraintChange extends AbstractChange {
     private String onUpdate;
     private String onDelete;
 
-
-    @Override
-    protected String[] createSupportedDatabasesMetaData(
```
filipelautert, line 60: This was totally wrong and was causing https://github.com/liquibase/liquibase-docs/issues/105

liquibase-standard/src/test/resources/liquibase/change/ChangeDefinitionTest.yaml
```diff
@@ -108,13 +108,13 @@ addForeignKeyConstraint: |
     Required For: all
   deferrable boolean 
     Description: Defines whether the foreign key is deferrable
-    Supported: asany,edb,oracle,postgresql,sqlite
```
filipelautert, line 111: The removed code on AFKCC class was incorrectly adding sqlite here.

suryaaki2 (approved):

StevenMassaro (approved):

PR #6119, opened 2024-07-23
chore: upgrade installer jdk version to 21.0.4+7

filipelautert:


filipelautert:
@rberezen we need to test installers generation.

rberezen:
> @rberezen we need to test installers generation.

@tati-qalified what os can you check?

tati-qalified:
@rberezen Windows and maybe Ubuntu

rberezen:
> @rberezen Windows and maybe Ubuntu

could you please check Windows, I'll check the rest
https://github.com/liquibase/liquibase/actions/runs/10251637098 - installers should be availabe here when job is finished

rberezen:
> > @rberezen Windows and maybe Ubuntu
> 
> could you please check Windows, I'll check the rest https://github.com/liquibase/liquibase/actions/runs/10251637098 - installers should be availabe here when job is finished

okay, we need to have liquibase-pro branch with the same name to build installers 
I'll create one and trigger build again

tati-qalified:
@rberezen something strange happens - I had v4.25.0 in my path environment variable, and when I ran the installer with the option to update my existing version, it for some reason updated v4.23.0, which I didn't have linked anywhere. 
It didn't change the value in the path, it just added a **new** variable with the new path.
Any ideas on why this could be happening?

rberezen:
@tati-qalified honestly, have no idea, try to reproduce again but make some clean up on your environment (remove old versions, env variables, etc)

tati-qalified:
@rberezen here are the installer options:
image



When I use "update the existing installation" and tick the "add liquibase to path" box, it doesn't check if it's already in the path. It will add it again instead of modifying the old one:
image



As the previous version is higher up on the list, that's the one that will be considered by the PC.
I think this beats the purpose of updating the existing installation

rberezen:
> @rberezen here are the installer options: image
> 
> When I use "update the existing installation" and tick the "add liquibase to path" box, it doesn't check if it's already in the path. It will add it again instead of modifying the old one: image
> 
> As the previous version is higher up on the list, that's the one that will be considered by the PC. I think this beats the purpose of updating the existing installation

agree, but I think this is not introduced in this PR and should be fixed as a separate issue
cc @filipelautert

tati-qalified (approved):

PR #6124, opened 2024-07-24
chore: re-enable releasing from master branch

filipelautert:
Msater branch name must be sed'ed too


abrackx (approved):
We should investigate if in the future can we just use `mvn versions:set` instead of this sed stuff 😮‍💨

filipelautert:
> We should investigate if in the future can we just use `mvn versions:set` instead of this sed stuff 😮‍💨

Agreed @abrackx ! Everytime we need to use sed/regex to fix a problem usually means we are doing something wrong ...

jandroav (approved):

rberezen (approved):

jandroav:
The issue is re-versioning stuff once everything is packed in the zip/tar...it would be more than investigating the `mvn versions:set`command :-(

PR #6128, opened 2024-07-24
'MANUAL' MySQL reserved word added

### LINKED ISSUE #6033: Missing reserved keywords for MySQL 8.4
### Search first

### Description

We use `manual` as a column name.
When updating from MySQL 8.0 to 8.4, this is now a reserved keyword but Liquibase does not recognize it as such and doesn't quote it in table creation statements when using the default `objectQuotingStrategy`, leading to an error.

### Steps To Reproduce

Use the `manual` keyword as column name in a Liquibase changeset for `createTable`, using the default `objectQuotingStrategy`.

### Expected/Desired Behavior

Liquibase should quote reserved keywords even when `objectQuotingStrategy=LEGACY`

### Liquibase Version

4.28.0

### Database Vendor & Version

MySQL 8.4

### Liquibase Integration

_No response_

### Liquibase Extensions

_No response_

_No response_

### Additional Context

_No response_

MalloD12:
## Impact

## Description

Fixes #6033 

without adding `manual` to the MySQL reserved word list even if we set `objectQuotingStrategy=LEGACY` it led to an error. Adding this new reserved word (available from MySQL from v8.4) to the list of keywords fix this issue.


liquibase-standard/src/main/java/liquibase/database/AbstractJdbcDatabase.java
```diff
@@ -946,9 +946,20 @@ protected boolean mustQuoteObjectName(String objectName, Class<? extends Databas
         if (isCatalogOrSchemaType(objectType) && preserveCaseIfRequested() == CatalogAndSchema.CatalogAndSchemaCase.ORIGINAL_CASE) {
             return true;
         }
+        addMySQLReservedWordIfApplicable("MANUAL");
```
filipelautert, line 949: Instead of adding it inside `mustQuoteObjectName `, I think it's better to add a new method to Database interface and call it where it's required. The problem of adding it here is that it may produce side effects to `mustQuoteObjectName ` - you call a method to check quotes and gets something added to the reserved words.

filipelautert (changes requested):

liquibase-standard/src/main/java/liquibase/database/AbstractJdbcDatabase.java
MalloD12: Hey Filipe, I agree with you that having this logic in `mustQuoteObjectName()` method was not in the right place. I was looking for options for a new method, but I thought moving this logic `AbstractjdbcDatabase.addReservedWords()` was good enough, so I moved it there. 

Could you please have a look at it again and please let me know if you think this is not the right approach?

Thanks,
Daniel.

filipelautert (approved):

rberezen (approved):

tati-qalified (approved):

liquibase-standard/src/main/java/liquibase/database/AbstractJdbcDatabase.java
```diff
@@ -180,6 +180,7 @@ public void setPreviousAutoCommit(Boolean previousAutoCommit) {
     @Override
     public final void addReservedWords(Collection<String> words) {
         reservedWords.addAll(words);
+        addMySQLReservedWordIfApplicable("MANUAL");
```
filipelautert, line 183: @MalloD12  Sorry, I just noticed that this was added to AbstractJdbcDatabase.java . Could you override this method in MysqlDatabase class and move this logic there?

filipelautert (changes requested):

liquibase-standard/src/main/java/liquibase/database/AbstractJdbcDatabase.java
MalloD12: Do you think it's wrong to have this logic there? Even if it's only applicable for `MySQL`? Do you mean overriding `addReservedWords()` this is a final method, so I wouldn't be able to override it for `MySQL` unless we want to remove the `final` keyword, but I think that would force us to implement this method for other DBs as well.

filipelautert: We can remove the final here as it was added only to [make sonar happy ](https://github.com/liquibase/liquibase/blob/3c57492ddc779de35b6e894e2b779fcdfa8b9bbe/liquibase-standard/src/main/java/liquibase/database/core/MariaDBDatabase.java#L21)  (and that's strange) . 
The main reason to move it to `MysqlDatabase` is to have  it in a place that makes more sense to find it later: it is confusing to find a `Mysql` reserved word in the super class if you already have a specialized class for Mysql, right?

filipelautert (changes requested):

liquibase-standard/src/main/java/liquibase/database/AbstractJdbcDatabase.java
MalloD12: Hey @filipelautert,

Just so you know- I have updated this PR as suggested. With changes made, there were some test failures because of `PERIOD` reserved word for MariaDB. I was looking for this word, and didn't see it's a `MySQL` reserved  word, I only found it's a reserved word if MariaDB is used in `Oracle` mode but it seems that's only applicable for v10.3.6 (see [here](https://mariadb.com/kb/en/reserved-words/)).

Do you think it's fine if we proceed this way?

Thanks,
Daniel.

filipelautert: I found that we missed a call to super in this refactoring - that's why the tests were failing. I added the call and reverted the tests, now it should build fine. Thanks for the refactoring!

filipelautert:
@tati-qalified we did some code refactoring here, could you validate this one again?

liquibase-standard/src/main/java/liquibase/database/AbstractJdbcDatabase.java
MalloD12: ohh true, my bad! Thanks for noticing it!

filipelautert (approved):

tati-qalified (approved):

PR #6130, opened 2024-07-24
DAT-17797

sayaliM0412:
feat: `.github/workflows/fossa.yml`: remove unwanted Curl calls and use custom instead of git as `Project Locator` 
(I am using a test bucket for now so as not to mess with the already existing bucket)

PRO PR: https://github.com/liquibase/liquibase-pro/pull/1874
uploaded report to s3: https://github.com/liquibase/liquibase-pro/actions/runs/10084308226
## Impact


jnewton03:
@sayaliM0412 is this still a valid PR?

sayaliM0412:
Doesnt seem like it. I think we should use do something like this for OSS and PRO:
https://github.com/liquibase/build-logic/blob/main/.github/workflows/generate-upload-fossa-report.yml.
I know there were issues generating reports as mentioned here: https://datical.atlassian.net/wiki/spaces/DEV/pages/4162224176/OSS+PRO+Fossa+Report+Generation
But we can use what we did in Enterprise to overcome the issues

PR #6143, opened 2024-07-30
Truncate Oracle PK names based on DB version

### LINKED ISSUE #6142: Oracle primary key size based on version
### Search first

### Description

The `OracleDatabase::generatePrimaryKeyName` method is always truncating the auto generated primary key name to 30 characters.
It should instead make use of the `OracleDatabase::getIdentifierMaximumLength` method to get the right lenght for the identifier, something like this:

```
@Override
public String generatePrimaryKeyName(String tableName) {
	int maximumLength = getIdentifierMaximumLength();
	String primaryKeyName = "PK_" + tableName.toUpperCase(Locale.US);
	if (primaryKeyName.length() > maximumLength) {
		return primaryKeyName.substring(0, maximumLength);
	} else {
		return primaryKeyName;
	}
}
```

Would it be ok if I raise a PR to address this?

### Steps To Reproduce

To reproduce this, you can run liquibase against an Oracle version grater than 12, with a long enough `databaseChangeLogLockTableName`, something longer than 30 characters.
The primary key name will be truncated to 30 characters.

### Expected/Desired Behavior

Having Oracle version grater than 12 and a long enough `databaseChangeLogLockTableName`, the primary key name should be truncated to 128 characters.

### Liquibase Version

_No response_

### Database Vendor & Version

Oracle version grater than 12

### Liquibase Integration

_No response_

### Liquibase Extensions

_No response_

_No response_

### Additional Context

_No response_

m4tt30c91:
Truncate Oracle Primary Key names based on DataBase version. Oracle till version 12.1 supports primary keys up to 30 characters, while newer versions support iderntifiers up to 128 characters long.

## Impact

## Description

Fixes #6142 

The change solve the issue by making use of the `OracleDatabase::getIdentifierMaximumLength` to get the maximum length the Oracle supports for its identifiers.


liquibase-standard/src/test/java/liquibase/database/core/OracleDatabaseTest.java
```diff
@@ -151,5 +152,89 @@ public void getDateLiteral_datetime_invalid() {
     public void getDateLiteral_unsupported() {
         assertEquals("UNSUPPORTED:123", database.getDateLiteral("123"));
     }
+
+    @Test
```
MalloD12, line 156: Hi @m4tt30c91,

We very much appreciate you adding several scenarios to test this change. One small suggestion about it is instead of adding a few very similar tests, adding only one and parameterizing it would look better. To do you can add something like this:

```java
private static Stream<Arguments> method() {
        return Stream.of(
                Arguments.of("SOME_TABLE_WITH_MORE_THAN_30_CHARACTERS", "PK_SOME_TABLE_WITH_MORE_THAN_3", OracleDatabase.ORACLE_12C_MAJOR_VERSION, "Should truncate primary key to 30 characters"),
                ...
               Arguments.of()
        );
    }

    @ParameterizedTest
    @MethodSource("method")
    @Test
    public void shouldTruncatePrimaryKeyNameCorrectlyDependingOnDBVersion(String tableName, String expectedPrimaryKeyName, int databaseVersion, String assertMessage) throws Exception {
        final Database database = getDatabase();
        final Database spyDatabase = Mockito.spy(database);
        Mockito.when(spyDatabase.getDatabaseMajorVersion()).thenReturn(databaseVersion);
        Mockito.when(spyDatabase.getDatabaseMinorVersion()).thenReturn(1);
        String primaryKeyName = spyDatabase.generatePrimaryKeyName(tableName);
        assertEquals(expectedPrimaryKeyName, primaryKeyName, assertMessage);
    }
```

How does it sound to you?

m4tt30c91: Hi @MalloD12 ,

Yeap thanks for the suggestion, I did a new commit to simplify the tests.
Let me know if that is ok :)

MalloD12 (approved):
**Approved**.

Code changes look good to me. Thank you, @m4tt30c91 for submitting this PR and addressing my suggestion so quickly.

Thanks,
Daniel.

liquibase-standard/src/main/java/liquibase/database/core/OracleDatabase.java
```diff
@@ -256,12 +256,12 @@ protected String getAutoIncrementClause(final String generationType, final Boole
 
     @Override
     public String generatePrimaryKeyName(String tableName) {
-        if (tableName.length() > 27) {
-            //noinspection HardCodedStringLiteral
-            return "PK_" + tableName.toUpperCase(Locale.US).substring(0, 27);
+        int maximumLength = getIdentifierMaximumLength();
```
filipelautert, line 259: It's interesting that this method was added back in 2017 and never used.

filipelautert (approved):
Thanks for the PR @m4tt30c91 !

liquibase-standard/src/main/java/liquibase/database/core/OracleDatabase.java
m4tt30c91: Yeap, probably the method was designed as part of some improvements and unfortunately forgot.

rberezen (approved):

filipelautert:
Thanks for the PR @m4tt30c91 , merging it to the next "major" release (aka 4.30.0)

Issue #6122, opened 2024-07-23
Diff-Command does not find missing columns since 4.24.0


ErrorProne:
### Search first

- [X] I searched and no similar issues were found

### Description

We use a small test to detect schema changes which are missing in the changelogs.
For this we populate a H2 from DD, another one with liquibase and run the diff command.
Since 4.24.0 the diff started to not pick up any columns.

This looks like a heavy impact which makes me think that I must have overlooked something. But I've read, and re-read the changelogs, issues and forums and I can't find anything. I'm very sorry in advanced if I've missed something.

### Steps To Reproduce

I wrote a dirty spring boot reproducer: https://github.com/ErrorProne/liquibase-reproducer-missing-columns
The test `LiquibaseMissingFieldNotDetectedApplicationTests.testLiquibaseAndHibernateSchemaPopulationsAreTheSame` succeeds with 4.23.0 and fails with 4.24.0.

### Expected/Desired Behavior

I expected the test diff to still pick up the missing column `FINDME` from the `DummyEntity` and create a changeset for it.

### Liquibase Version

>= 4.24.0

### Database Vendor & Version

H2

### Liquibase Integration

spring boot

### Are you willing to submit a PR?

- [X] I'm willing to submit a PR (Thank you!)

Pyenb:
Possibly related to #5943

tati-qalified:
Hi @ErrorProne, thank you for reporting this bug and for providing a reproducer. I have been able to replicate it with Liquibase versions 4.24.0 and 4.29.0, so it's still an issue.

@Pyenb I'm not entirely sure if it's the same problem, as the reporter said the following:

> The diff command show the difference correctly but the diffChangeLog doesn't.

However in this case it's the diff command that's not working correctly.

Would any of you be willing to submit a PR? Our development team will be available to provide any guidance needed.

Thank you,
Tatiana

ErrorProne:
> Would any of you be willing to submit a PR? Our development team will be available to provide any guidance needed.

@tati-qalified Sure, I'll have a look

ErrorProne:
That was not the place at all I expected the problem to be: https://github.com/liquibase/liquibase/pull/4052/files#diff-a0076d9bb727dcbec183152bec633f6fc4dfdc414835260f6a876aba91dc5f8fR404

This change breaks the diff if DATABASE_TO_LOWER=true is specified for the H2 connection, which is a thing if you, for example run in PostgreSQL compatibility mode.

[...]

PR #6153, opened 2024-07-31
[6022] Keep original schema and catalog case

ErrorProne:
H2 allows users to specify the behavior for casing. This is important when certain databases are mimicked. If the schema and catalog always are rewritten to UPPER_CASE, as it is the default Database.java behavior, situations may occur in which the schema and/or catalog name no longer match the actual database object.

To handle this schema and catalog are now always returned in the original case.

## Impact

## Description

- Fix for https://github.com/liquibase/liquibase/issues/6122
- Handle H2s ability to adapt the casing for schemas and catalogs
- Changed H2 to always  use original case for catalog and schema

### DIFF
--- a/liquibase-standard/src/main/java/liquibase/database/core/H2Database.java
+++ b/liquibase-standard/src/main/java/liquibase/database/core/H2Database.java
@@ -411,6 +411,11 @@ public class H2Database extends AbstractJdbcDatabase {
     }
 
     @Override
+    public CatalogAndSchema.CatalogAndSchemaCase getSchemaAndCatalogCase() {
+        return CatalogAndSchema.CatalogAndSchemaCase.ORIGINAL_CASE;
+    }
+
+    @Override
     public String escapeObjectName(String objectName, final Class<? extends DatabaseObject> objectType) {
         if (objectName == null) {
             return null;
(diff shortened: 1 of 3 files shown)


ErrorProne:
The failed tests look like they may be related to my change. I'll check them later.

ErrorProne:
@filipelautert 
I've no idea. It feels like the handling of quoting and casing are too entangled and there are too many combinations of how H2 may behave. My case is one of them, but the fix would maybe create new problems on another end or combination.

Maybe it is better suited for an addition in the doc, that if H2 with postgres (and maybe other) compatibility mode is run the 
liquibase.preserveSchemaCase must be set to true.

filipelautert:
> @ErrorProne probably the issues you are facing were introduced by PR #4052 . But in the PR there is a hint in this comment: [#4052 (comment)](https://github.com/liquibase/liquibase/pull/4052#issuecomment-1756134319)
> Do you think it makes sense to "fix" your issues?

@tati-qalified could you help to reproduce this issue and this workaround?

tati-qalified:
@filipelautert as there is already a way to obtain this behaviour, do you think we should implement the fix, or change the docs?

filipelautert:
> @filipelautert as there is already a way to obtain this behaviour, do you think we should implement the fix, or change the docs?

Updating the docs seems a better approach, as changing this behavior now may break current H2 setup for other users.

lbadam:
@filipelautert I agree. @tati-qalified can you create a ticket with instructions for the docs?

PR #5704, opened 2024-03-15
Fix SQL execution issue with endDelimiter '/' and splitStatements

MalloD12:
## Impact

- [x] Bug fix (non-breaking change which fixes expected existing functionality)
- [ ] Enhancement/New feature (adds functionality without impacting existing logic)
- [ ] Breaking change (fix or feature that would cause existing functionality to change)
 

## Description

Fixes #5359

Issue #5359, opened 2023-12-12
[V4.25.0] Regression in SQL execution with endDelimiter and splitStatements


adamjones0101:
### Search first

- [X] I searched and no similar issues were found

### Description

In a changeset an sql statement is added.  splitStatements= true and endDelimiter = "/"
if there are 2 statements in the sql which have an "/" as the last character then the execution fails with a "missing or invalid option" error, ORA-00922 in Oracle DB

### Steps To Reproduce

Include the following in a changeset:

<sql dbms="oracle" endDelimiter="/" splitStatements="true" stripComments="false>
   CREATE USER mytestuser1 IDENTIFIED BY "abcde12345"  PROFILE DEFAULT ACCOUNT UNLOCK/
   GRANT CREATE TABLE TO mytestuser1/
</sql>

### Expected/Desired Behavior

in version 4.9.1 this works correctly but in 4.25.0 it appears that the entire sql is executed as a single statement.

### Liquibase Version

4.25.0

### Are you willing to submit a PR?

- [ ] I'm willing to submit a PR (Thank you!)

tati-qalified:
Hi @adamjones0101, thank you for reporting this issue.
I was able to verify that it is replicable and it was introduced by Liquibase 4.15.0, possibly with #3118.

I'll be forwarding this to the development team so they can evaluate how to move forward.

Thanks,
Tatiana

MalloD12:
Hi @adamjones0101,

As stated by @tati-qalified this behavior has been updated with #3118, and it is working as designed. For you to make this SQL execute successfully you need to update it in the below way:
```
<sql dbms="oracle" endDelimiter="/" splitStatements="true" stripComments="false">
CREATE USER mytestuser1 IDENTIFIED BY "abcde123" PROFILE DEFAULT ACCOUNT UNLOCK
/
GRANT CREATE TABLE TO mytestuser1
/
</sql>
```
There shouldn't be any space either at the beginning or the end of each sentence and `/` should be in the next line. As specified by @nvoxland, this is to avoid splitting a statement at the wrong spot. 

I'll close this issue for now, but if you have any additional feedback or you are keen to propose an enhancement for this you are more than welcome to do it.

Thanks,
Daniel.

adamjones0101:
I guess that there's no concern that nearly every liquibase changelog on the planet has been broken.
We are going to have to modify hundreds if not thousands of changesets to accommodate this backward incompatible change.

MalloD12:
[...]

PR #3118, opened 2022-07-27
Improved handling of endDelimiter="/"

nvoxland:
## Impact

- [X] Bug fix (non-breaking change which fixes expected existing functionality)
- [ ] Enhancement/New feature (adds functionality without impacting existing logic)
- [ ] Breaking change (fix or feature that would cause existing functionality to change)
 
## Description

If people put endDelimiter="/", they really mean the oracle-style syntax of:

```
statement 1
/
statement2
/
```

and we should treat that as such and not actually use "/" as a simple delimiter. Otherwise, something like:

```
create table x id int default 2/4
/
```
splits in the wrong spot.

Fixes #2688

## Things to be aware of

- Only impacts split logic when endDelimiter is /
- The default end delimiter already handles `/` like this, this change just extends that to explicitly specifying endDelimiter="/"

## Things to worry about

- Nothing

PR #2135, opened 2021-10-01
Fix #2133 and preserve CORE-2971 (Liquibase endDelimiter does not work after upgrading from 3.10.x to 4.x)

Saucistophe:
## Environment
**Liquibase Version**: Latest (4.5.1?)

**Liquibase Integration & Version**: core

## Pull Request Type
* [x] Bug fix (non-breaking change which fixes an issue.)
* [ ] Enhancement/New feature (non-breaking change which adds functionality)
* [ ] Breaking change (fix or feature that would cause existing functionality to change)

## Description
Fixes #2133 while preserving (and testing) CORE-2971.
It's a tad less aggressive regarding end delimiters.

## Steps To Reproduce
All steps are clearly listed in the Github issue.

## Fast Track PR Acceptance Checklist:
* [x] Build is successful and all new and existing tests pass
* [x] Added [Unit Test(s)](https://liquibase.jira.com/wiki/spaces/LB/pages/1274937609/How+to+Write+Liquibase+Core+Unit+Tests)
* [ ] Added [Integration Test(s)](https://liquibase.jira.com/wiki/spaces/LB/pages/1276608569/How+to+Write+Liquibase+Core+Integration+Tests)
* [ ] Added [Test Harness Test(s)](https://github.com/liquibase/liquibase-test-harness/pulls)
* [ ] Documentation Updated

- - -
#### Links
* Fixed Issue: https://github.com/liquibase/liquibase/issues/2133
* Test Results: https://github.com/liquibase/liquibase/pull/2135/checks

#### Testing
* Steps to Reproduce: https://github.com/liquibase/liquibase/issues/2133 while not impacting https://liquibase.jira.com/browse/CORE-2971
* Guidance:
  * Impact: changes the handling of delimiters in parsed sql, whether it is in formatted sql or a sql/sqlFile tag

#### Dev Verification
Reviewed code and verified that automated tests now check the described problem

┆Issue is synchronized with this [Jira Bug](https://datical.atlassian.net/browse/LB-2186) by [Unito](https://www.unito.io)

PR #6157, opened 2024-08-01
fix: EndDelimiter not working as expected since 4.29

### LINKED ISSUE #6150: EndDelimiter '/' not working as expected since 4.29
### Search first

### Description

We are deploying SQL and PLSQL scripts on an Oracle Database via Liquibase. We use ‘/’ as endDelimiter because it’s compatible with native SQL and PLSQL.

Since 4.29 this is some kind of broken. Before only a single "/" as first and only character of a new line was recognized as endDelimiter. Now also a "/" at the end of a line is interpreted as endDelimiter. This is not compatible with native SQL anymore.

### Steps To Reproduce

Code similar to the following was leading to an error because of false code splitting

```
-- liquibase formatted sql
-- changeset abc:ABC_PROC runOnChange:true stripComments:false endDelimiter:/
-- comment Initial Creation
CREATE OR REPLACE PROCEDURE ABC_PROC…IS
…
BEGIN
…
i := j /
 2;
…
END;
/
```

### Expected/Desired Behavior

"/" should only interpreted as endDelimiter if it's the only character on a new line.

### Liquibase Version

4.29

### Database Vendor & Version

Oracle 19

### Liquibase Integration

CLI

### Liquibase Extensions

_No response_

Docker

### Additional Context

_No response_

filipelautert:
## Description

Consider that you have sql endDelimiter = "/" , and the following SQL:
```
a = var1 / 
var2;
```

How Liquibase should handle it? Should we split or divide? It can be understood as:

```
a = var1 / var2; --  'division behavior'
```
or
```
a = var1 ;   
var2; -- 'strict behavior'
```

Historically, we had:

* Until v4.9.0 - Division behavior
* v4.9.0 (2022.03.17) - changed to Strict behavior - https://github.com/liquibase/liquibase/pull/2135
* v4.15.0 (2022.08.04) - rolled back to Division behavior - https://github.com/liquibase/liquibase/pull/3118
* v4.29.0 (2024-07-28) - changed to Strict behavior again! - https://github.com/liquibase/liquibase/pull/5704

Versions v4.9.0 to v4.14.x (only 5 months) were `strict` while all other versions are `division`. And with 4.29.0 we restored `strict`  because of issue https://github.com/liquibase/liquibase/issues/5359 (this is a case where an upgrade is being done from Liquibase 4.9.1).

As we have 2 different behaviors we decided to support both by using the `strict` flag. So: 

* Division behavior has been suported for longer and it's the default behavior
* Strict behavior is optional and can be enabled by using the `--scrict` parameter ( https://docs.liquibase.com/parameters/strict.html ) .

This is being documented so we don't change it again.

Fixes #6150

### DIFF
--- a/liquibase-integration-tests/src/test/java/liquibase/dbtest/oracle/OracleIntegrationTest.java
+++ b/liquibase-integration-tests/src/test/java/liquibase/dbtest/oracle/OracleIntegrationTest.java
@@ -239,7 +239,6 @@ public class OracleIntegrationTest extends AbstractIntegrationTest {
     @Test
     public void verifySlashReplacementOnStrictMode() throws DatabaseException {
         assumeNotNull(getDatabase());
-        clearDatabase();
         try {
             Scope.child(GlobalConfiguration.STRICT.getKey(), true, () -> {
                 Database database = getDatabase();
@@ -250,8 +249,6 @@ public class OracleIntegrationTest extends AbstractIntegrationTest {
             });
         } catch (Exception e) {
             Assert.fail("Should not fail. Reason: " + e.getMessage());
-        } finally {
-            clearDatabase();
         }
     }
 }
--- a/liquibase-integration-tests/src/test/java/liquibase/dbtest/oracle/OracleIntegrationTest.java
+++ b/liquibase-integration-tests/src/test/java/liquibase/dbtest/oracle/OracleIntegrationTest.java
@@ -23,6 +23,7 @@ import liquibase.sql.visitor.AbstractSqlVisitor;
 import liquibase.statement.core.RawParameterizedSqlStatement;
 import liquibase.structure.core.Index;
 import org.junit.Assert;
+import org.junit.Ignore;
 import org.junit.Test;
 
 import java.io.ByteArrayOutputStream;
@@ -230,7 +231,9 @@ public class OracleIntegrationTest extends AbstractIntegrationTest {
     }
 
     @Test
+    @Ignore
     public void verifySlashReplacementOnStrictMode() throws DatabaseException {
+        assumeNotNull(getDatabase());
         clearDatabase();
         try {
             Scope.child(GlobalConfiguration.STRICT.getKey(), true, () -> {
--- a/liquibase-integration-tests/src/test/java/liquibase/dbtest/oracle/OracleIntegrationTest.java
+++ b/liquibase-integration-tests/src/test/java/liquibase/dbtest/oracle/OracleIntegrationTest.java
@@ -23,7 +23,6 @@ import liquibase.sql.visitor.AbstractSqlVisitor;
 import liquibase.statement.core.RawParameterizedSqlStatement;
 import liquibase.structure.core.Index;
 import org.junit.Assert;
-import org.junit.Ignore;
 import org.junit.Test;
 
 import java.io.ByteArrayOutputStream;
--- a/liquibase-integration-tests/src/test/java/liquibase/dbtest/oracle/OracleIntegrationTest.java
+++ b/liquibase-integration-tests/src/test/java/liquibase/dbtest/oracle/OracleIntegrationTest.java
@@ -187,6 +187,12 @@ public class OracleIntegrationTest extends AbstractIntegrationTest {
         }
     }
 
+    @Override
+    @Test
+    public void testDiffExternalForeignKeys() throws Exception {
+        //cross-schema security for oracle is a bother, ignoring test for now
+    }
+
     @Test
     public void verifyIndexIsCreatedWhenAssociatedWithPropertyIsSetAsForeignKey() throws DatabaseException {
         clearDatabase();
@@ -231,7 +237,6 @@ public class OracleIntegrationTest extends AbstractIntegrationTest {
     }
 
     @Test
-    @Ignore
     public void verifySlashReplacementOnStrictMode() throws DatabaseException {
         assumeNotNull(getDatabase());
         clearDatabase();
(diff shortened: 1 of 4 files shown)


liquibase-integration-tests/src/test/java/liquibase/dbtest/oracle/OracleIntegrationTest.java
```diff
@@ -236,4 +235,23 @@ public void testChangeLogGenerationForTableWithGeneratedColumn() throws Exceptio
         String generatedChangeLog = baos.toString();
         assertTrue("Text '" + textToTest + "' not found in generated change log: " + generatedChangeLog, generatedChangeLog.contains(textToTest));
     }
+
+    @Test
+    public void verifySlashReplacementOnStrictMode() throws DatabaseException {
+        assumeNotNull(getDatabase());
```
MalloD12, line 241: We can clean up a bit this test by removing the `clearDatabase()` calls.

liquibase-standard/src/main/java/liquibase/util/StringUtil.java
```diff
@@ -110,8 +110,13 @@
             }
 
             if (isInClause == 0 && splitStatements && (piece instanceof String) && isDelimiter((String) piece, previousPiece, endDelimiter)) {
-                String sentenceWithoutDelimiter = removeEndDelimiterIfItsASlash(endDelimiter, currentString);
-                String trimmedString = sentenceWithoutDelimiter.isEmpty() ? StringUtil.trimToNull(currentString.toString()):StringUtil.trimToNull(sentenceWithoutDelimiter);
+                String trimmedString;
+                if (Boolean.TRUE.equals(GlobalConfiguration.STRICT.getCurrentValue())) {
+                    String sentenceWithoutDelimiter = removeEndDelimiterIfItsASlash(endDelimiter, currentString);
+                    trimmedString = sentenceWithoutDelimiter.isEmpty() ? StringUtil.trimToNull(currentString.toString()) : StringUtil.trimToNull(sentenceWithoutDelimiter);
```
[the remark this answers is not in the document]
MalloD12: We can use `StringUtils` class from `org.apache.commons.lang3`.

liquibase-integration-tests/src/test/java/liquibase/dbtest/oracle/OracleIntegrationTest.java
```diff
@@ -22,6 +23,7 @@
 import liquibase.statement.core.RawParameterizedSqlStatement;
 import liquibase.structure.core.Index;
 import org.junit.Assert;
+import org.junit.Ignore;
```
rberezen, line 26: I guess we do not need this

rberezen:
testing blocked by https://datical.atlassian.net/browse/DAT-18440

filipelautert:
Blocked by https://github.com/liquibase/liquibase-pro/pull/1909 . CC @tati-qalified

rberezen:
@tati-qalified @filipelautert this should be unblocked now

tati-qalified:
@rberezen @filipelautert using `--strict=true` doesn't fail anymore!

This changeset generates the same code, no matter if strict is true or false:
```
-- liquibase formatted sql
-- changeset AUTHOR:CHANGESET_NAME endDelimiter:GO rollbackEndDelimiter:GO runAlways:true runOnChange:true dbms:mssql
declare cursorToProcess cursor local
	read_only forward_only fast_forward
	for
		SELECT * FROM DATABASECHANGELOG;

open cursorToProcess;
```

Generates:
```
-- Changeset mssql.sql::CHANGESET_NAME::AUTHOR
declare cursorToProcess cursor local
	read_only forward_only fast_forward
	for
		SELECT * FROM DATABASECHANGELOG;

open cursorToProcess;
GO
```

I think we're good to go with this

tati-qalified (approved):

rberezen (approved):

rberezen:
@adrian-velonis1 @AMBERMW13 
we need to update this section:
```
Newline behavior with a / end delimiter
In Liquibase 4.8.0 to 4.14.0, using the delimiter / (forward slash) in a PL/SQL block, for example as a division operator, causes Liquibase to fail on Oracle databases:

Caused by: liquibase.exception.LiquibaseException: liquibase.exception.MigrationFailedException: Migration failed for change set filepath/example-changeset.sql::example-id::example-author:
		Reason: liquibase.exception.DatabaseException: ORA-00900: invalid SQL statement
In Liquibase 4.15.0 to 4.28.0, Liquibase fixes this by requiring a newline before the / to treat it as a delimiter. Otherwise, the / is treated as a regular character. However, using a newline in the wrong place can lead to the error message "ORA-00922: missing or invalid option" on Oracle databases.

In Liquibase 4.29.0, Liquibase fixes this by requiring a newline after the / to treat it as a delimiter, not before. Otherwise, the / is treated as a regular character.
```
to

`In Liquibase 4.29.2, the / is treated as a regular character when --strict=true and as a delimiter when  --strict=false or not specified.`
is it correct @filipelautert 
https://docs.liquibase.com/change-types/enddelimiter-sql.html

filipelautert:
> `In Liquibase 4.29.2, the / is treated as a regular character when --strict=true and as a delimiter when --strict=false or not specified.`
> is it correct @filipelautert ?
> https://docs.liquibase.com/change-types/enddelimiter-sql.html

@rberezen yes, this seems a good description.

adrian-velonis1:
[DOC TICKET] https://datical.atlassian.net/browse/PD-5023

PR #6169, opened 2024-08-06
update-to-tag fixes when provided tag doesn't exist or there is not tagDatabaseChange in the changelog

### LINKED ISSUE #6126: update-to-tag runs even when provided tag doesn't exist
### Search first

### Description

When running `liquibase update-to-tag --tag=myTag`, two different non-ideal things can happen:

1. If there is no `tagDatabase` change type in the changelog file, the entire changelog will run anyway. 
2. If the provided tag doesn't exist (for example if it's misspelled), all changes will run.

### **First case:**
1. Have a changelog with no `tagDatabase`, for example
```
<databaseChangeLog
        http://www.liquibase.org/xml/ns/dbchangelog/dbchangelog-latest.xsd">

    <changeSet  id="1"  author="nvoxland">  
        <createTable  tableName="person">  
            <column  name="id"  type="int"  autoIncrement="true">  
                <constraints  primaryKey="true"  nullable="false"/>  
            </column>  
            <column  name="firstname"  type="varchar(50)"/>  
            <column  name="lastname"  type="varchar(50)">  
                <constraints  nullable="false"/>  
            </column>  
            <column  name="state"  type="char(2)"/>  
        </createTable>  
    </changeSet>  

    <changeSet  id="2"  author="nvoxland">  
        <addColumn  tableName="person">  
            <column  name="username"  type="varchar(8)"/>  
        </addColumn>  
    </changeSet>  

</databaseChangeLog>
```
2. Run `liquibase update-to-tag --tag=myTag`

### **Second case:**
1. Have a changelog *with* a `tagDatabase` change type like
```
<databaseChangeLog
    xmlns:pro="http://www.liquibase.org/xml/ns/pro"
        http://www.liquibase.org/xml/ns/dbchangelog/dbchangelog-latest.xsd
        http://www.liquibase.org/xml/ns/dbchangelog-ext http://www.liquibase.org/xml/ns/dbchangelog/dbchangelog-ext.xsd
        http://www.liquibase.org/xml/ns/pro http://www.liquibase.org/xml/ns/pro/liquibase-pro-latest.xsd">

    <changeSet  id="1"  author="nvoxland">  
        <createTable  tableName="person">  
            <column  name="id"  type="int"  autoIncrement="true">  
                <constraints  primaryKey="true"  nullable="false"/>  
            </column>  
            <column  name="firstname"  type="varchar(50)"/>  
            <column  name="lastname"  type="varchar(50)">  
                <constraints  nullable="false"/>  
            </column>  
            <column  name="state"  type="char(2)"/>  
        </createTable>  
    </changeSet>  

    <changeSet id="abc" author="123">
        <tagDatabase tag="myTag" />
    </changeSet>

    <changeSet  id="2"  author="nvoxland">  
        <addColumn  tableName="person">  
            <column  name="username"  type="varchar(8)"/>  
        </addColumn>  
    </changeSet>  

</databaseChangeLog>
```
2. Run `liquibase update-to-tag --tag=differentTag`

### Expected/Desired Behavior

If no `tagDatabase` change type is present in the changelog, or if the provided tag doesn't exist, it should be an immediate error, and nothing should be executed.

### Liquibase Version

All

### Database Vendor & Version

All

### Liquibase Integration

_No response_

### Liquibase Extensions

_No response_

_No response_

### Additional Context

_No response_

MalloD12:
## Impact

## Description

Fixes #6126 

When running liquibase update-to-tag --tag=myTag, two different non-ideal things can happen:

1.  If there is no tagDatabase change type in the changelog file, the entire changelog will run anyway.
2. If the provided tag doesn't exist (for example if it's misspelled), all changes will run.

To keep backward compatibility with some people's usage we have not made any specific change to the current behavior, but what we have done on this PR is to introduce the usage of `STRICT` global configuration for the two above cases. Now, if `STRICT=true` in the first case a `TagDatabaseChange` not found error message will be displayed and in the second case a message letting the user know there is no matching between the command tag argument and the tag specified in the change.


liquibase-standard/src/main/java/liquibase/changelog/StatusChangeLogIterator.java
```diff
@@ -38,6 +41,8 @@ public void run(ChangeSetVisitor visitor, RuntimeEnvironment env) throws Liquiba
         try {
             Scope.child(Scope.Attr.databaseChangeLog, databaseChangeLog, () -> {
                 List<ChangeSet> changeSetList = new ArrayList<>(databaseChangeLog.getChangeSets());
+                TagDatabaseChange tagDatabaseChange = getTagDatabaseChange(changeSetList);
+                boolean thereIsTagDatabaseChange =  tagDatabaseChange != null;
```
filipelautert, line 45: Those 2 new parameters can be moved inside the `if(Strict)` block below so they don't get computed if they aren't used.

filipelautert (changes requested):

liquibase-standard/src/main/java/liquibase/changelog/StatusChangeLogIterator.java
MalloD12: Done

liquibase-standard/src/main/java/liquibase/changelog/StatusChangeLogIterator.java
```diff
@@ -58,6 +61,18 @@ public void run(ChangeSetVisitor visitor, RuntimeEnvironment env) throws Liquiba
                         }
                     }
                 }
+                if(GlobalConfiguration.STRICT.getCurrentValue()) {
```
filipelautert, line 64: Do we need this change to `StatusChangeLogIterator.java` ? It feels like everything could be validated in the Command Step as this code is executed before the status validation -> https://github.com/liquibase/liquibase/pull/6169/files#diff-d71f26aeddfbd41c01479ae9b2fd4ac32f42e76b8ca9c2b5bce22bb2260dbf47R108

filipelautert (changes requested):

liquibase-standard/src/main/java/liquibase/changelog/StatusChangeLogIterator.java
MalloD12: Done

liquibase-standard/src/main/java/liquibase/command/core/UpdateToTagCommandStep.java
```diff
@@ -87,35 +89,64 @@ public UpdateSummaryEnum getShowSummary(CommandScope commandScope) {
...
                     "The tag '%s' was not found in the changelog '%s'. All changesets in the changelog were deployed.%nLearn about options for undoing these changes at https://docs.liquibase.com.",
                     tag, changeLog.getPhysicalFilePath());
             Scope.getCurrentScope().getLog(UpdateToTagCommandStep.class).warning(message);
             Scope.getCurrentScope().getUI().sendMessage("WARNING:  " + message);
             warningMessageShown = true;
         }
+
+        if(GlobalConfiguration.STRICT.getCurrentValue()) {
+            TagDatabaseChange tagDatabaseChange = getTagDatabaseChange(changeLog.getChangeSets());
+            boolean thereIsTagDatabaseChange =  tagDatabaseChange != null;
```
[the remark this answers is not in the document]
MalloD12: Yeah,  I saw it but decide to go with this change as it is, because `found` boolean flag considers both cases together (`tagDatabaseChange` exits and the runtime tag matches with the change's tag).

filipelautert (approved):

tati-qalified (approved):

rberezen:
this is a useful change, but want to notify other stakeholders about this implementation
one thing I'm concerned about is that we put too much on the --strict flag
there might be cases when user might want to use i.e. endDelimiter with --strict=true and update-to-tag --strict=false that will conflict each other
cc @suryaaki @lbadam @mariochampion

rberezen:
ran extra tests https://github.com/liquibase/liquibase-pro-tests/actions/runs/10620473636

rberezen (approved):

lbadam:
@rberezen that's a good point. I do agree that we need to be careful how we continue to add to `strict`. I would like to see us implement a system where we have a generic `strict` for users while also allowing them to toggle individual properties like https://www.typescriptlang.org/tsconfig/#strict

MalloD12:
I agree with @rberezen, I think at some point there would be some feature crashing using `strict` flag. In the past, we had some related discussions where we were discussing whether to overload strict or have separate global configurations that allow us to control some specific behavior, but we decided to keep using strict for this purpose instead of increase more and more this list (of configurations) up to a point it gets unmaintainable. I think something similar is happening now with `strict` flag, I'm thinking maybe we can go in the direction of a middle point of having a few strict flags to control command-specific behaviors, or to control some other properties specific behaviors. For example:

```
--strict-update

--stric-changesetProperties (or some other names, but I'm pointing out here cases like endDelimiter or any other).

...

etc
```

What do you guys think?

PR #6171, opened 2024-08-06
Lombokify (DAT-15687)

StevenMassaro:
## Impact


StevenMassaro:
The dev team does not think that this can be done in a safe and reliable way, given the current state of the Lombok plugin for Intellij.

PR #6179, opened 2024-08-07
Make Scope ID RNG no longer use SecureRandom (fix #6178)

### LINKED ISSUE #6178: Liquibase hangs when trying to apply changesets on a newly created machine
### Search first

### Description

Applying a large amount of Liquibase changesets from a newly created machine (like a fresh VM/Docker container) leaves the process hanging and making very little progress.

### Steps To Reproduce

Launch a Liquibase changeset application (with a large-ish number of changes) from a fresh VM/container, and it will hang for what seems forever, checking the `/proc/sys/kernel/random/entropy_avail` shows that it is regularly at 0 during the hang.

### Expected/Desired Behavior

Liquibase should not hang, it should apply the changesets normally.

### Liquibase Version

4.29

### Database Vendor & Version

_No response_

### Liquibase Integration

_No response_

### Liquibase Extensions

_No response_

Ubuntu (but most Linux would probably be affected too)

### Additional Context

The culprit seems to be 3daa688c6a99ed131687036ba4e40cd743c4b3c8, where the SecureRandom was moved to use the Apache Commons RandomStringUtils. This probably drained the entropy faster than the initial solution. I'm aware that this change was done to improve performance of the Scope ID generation, so I will send a PR that should perform better than SecureRandom and not have this entropy problem. 

danielthegray:
## Impact

## Description
SecureRandom is a robust RNG for use in security-sensitive scenarios (like generating cryptographic/private keys). For something like a random Scope ID, it is overkill, as it's slower, and becomes a potentially blocking call if the machine hasn't been on for long or have enough entropy.

~~I've looked into the original change (in commit 3daa688c6a99ed131687036ba4e40cd743c4b3c8) and it seems to have been done to improve the performance of the Scope ID generation. While it clearly did improve things, it's much better to just use a normal ThreadLocalRandom instance, that will not experience contention issues even under high traffic, and will not have the problem with depending on system entropy.~~

The original change (in commit 3daa688c6a99ed131687036ba4e40cd743c4b3c8) already did this change implicitly, since RandomStringUtils up to version 3.14 was using ThreadLocalRandom. So from Liquibase's perspective, this maintains the improved performance profile without making it subject to the whims of Apache Commons Lang3.

Fixes #6178. I built a local `liquibase-core` JAR and linked into my CLI tool, and the changeset deployment no longer hangs.

## Things to be aware of
~~This makes the Scope ID generation not use a SecureRandom instance anymore, but it should really not be a problem as it should not require cryptographic-level randomness.~~
This is just setting your Scope ID generation to use ThreadLocalRandom again, more explicitly, and in a way that cannot be arbitrarily changed by myopic library maintainers :)

## Things to worry about
~~I think that the performance should actually be much better than the previous implemented fix (using Apache Commons), but I have not tested your use-case. I would recommend that you test it and see what the actual performance impact is.~~

Actually, you did the change initially, to move away from `SecureRandom` to `ThreadLocalRandom`, and the Apache Commons Lang3 team just changed that out from under you in v3.15, not just back to `SecureRandom` but all the way to `SecureRandom.getInstanceStrong()` which will make this block/take forever. So I see nothing more to worry about :)

### DIFF
--- a/liquibase-standard/src/main/java/liquibase/util/StringUtil.java
+++ b/liquibase-standard/src/main/java/liquibase/util/StringUtil.java
@@ -8,12 +8,12 @@ import liquibase.parser.LiquibaseSqlParser;
 import liquibase.parser.SqlParserFactory;
 import org.apache.commons.lang3.ArrayUtils;
 import org.apache.commons.lang3.CharUtils;
-import org.apache.commons.lang3.RandomStringUtils;
 import org.apache.commons.lang3.StringUtils;
 
 import java.io.UnsupportedEncodingException;
 import java.nio.charset.StandardCharsets;
 import java.util.*;
+import java.util.concurrent.ThreadLocalRandom;
 import java.util.regex.Matcher;
 import java.util.regex.Pattern;
 
@@ -768,7 +768,12 @@ public class StringUtil {
      * @return an identifier of the desired length
      */
     public static String randomIdentifier(int len) {
-        return RandomStringUtils.random(len, true, false);
+        final String AB = "ABCDEFGHIJKLMNOPQRSTUVWXYZ";
+
+        StringBuilder sb = new StringBuilder(len);
+        for (int i = 0; i < len; i++)
+            sb.append(AB.charAt(ThreadLocalRandom.current().nextInt(AB.length())));
+        return sb.toString();
     }
 
     /**


filipelautert (approved):
Thanks for the PR @danielthegray !

danielthegray:
I think you'll also want to check your usage of RandomStringUtils in `liquibase-integration-tests/src/test/groovy/liquibase/diffchangelog/DiffChangelogMssqlIntegrationTest.groovy`, as under the new implementation this test could run slower or block, although with just 2 invocations, maybe it won't be too big of a deal, but I would suggest that you just stop using RandomStringUtils.

I have definitely banned this class from my codebase as this irresponsible behavior make this class a foot-gun for future colleagues.

tati-qalified (commented):
The prior version of this method returned strings that contained both upper and lowercase letters, while the new version uses only uppercase.
@filipelautert I think it would be best to add all lowercase letters to `final String AB = "<...>"`. Thoughts?

filipelautert:
> The prior version of this method returned strings that contained both upper and lowercase letters, while the new version uses only uppercase. @filipelautert I think it would be best to add all lowercase letters to `final String AB = "<...>"`. Thoughts?

The original code before using `RandomStringUtils` only used uppercase letters too, so that's not an issue ;)

tati-qalified (approved):

PR #6182, opened 2024-08-08
Use insecure RNG for generating scope IDs

danielthegray:
## Impact

## Description
Apache Commons Lang3 3.15 changed to using SecureRandom instead of ThreadLocalRandom, and due to the various hanging issues that have been reported, a secure()/insecure() API was created to make the choice more explicit.

This would revert the performance improvement that was obtained by moving to RandomStringUtils a few weeks ago.

## Things to be aware of
This involved a version increase to 3.16 of Apache Commons Lang3. It is insane that they published such a breaking behavioral change with a minor version bump, but it is what it is.

## Things to worry about
I can see a risk of people taking a dependency on Liquibase not being on Apache Commons Lang3 3.16 having some errors since the new API to select the insecure RNG is not API-compatible, but just having hard-to-debug random hangs is worse.

### DIFF
--- a/liquibase-standard/src/main/java/liquibase/util/StringUtil.java
+++ b/liquibase-standard/src/main/java/liquibase/util/StringUtil.java
@@ -768,7 +768,7 @@ public class StringUtil {
      * @return an identifier of the desired length
      */
     public static String randomIdentifier(int len) {
-        return RandomStringUtils.random(len, true, false);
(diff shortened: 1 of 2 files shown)


filipelautert:
As discussed in discord, let's go with https://github.com/liquibase/liquibase/pull/6179

PR #6209, opened 2024-08-16
initial add of analytics tracking and supporting classes (DAT-17823)

StevenMassaro:
## Impact

### DIFF
--- a/liquibase-standard/src/main/java/liquibase/analytics/AnalyticsConfiguration.java
+++ b/liquibase-standard/src/main/java/liquibase/analytics/AnalyticsConfiguration.java
@@ -1,0 +1,22 @@
+package liquibase.analytics;
+
+import liquibase.configuration.AutoloadedConfigurations;
+import liquibase.configuration.ConfigurationDefinition;
+
+public class AnalyticsConfiguration implements AutoloadedConfigurations {
+
+    public static final ConfigurationDefinition<AnalyticsOutputDestination> OUTPUT_DESTINATION;
+    public static final ConfigurationDefinition<String> FILENAME;
+
+    static {
+        ConfigurationDefinition.Builder builder = new ConfigurationDefinition.Builder("liquibase.analytics");
+
+        OUTPUT_DESTINATION = builder.define("outputDestination", AnalyticsOutputDestination.class)
+                .setDefaultValue(AnalyticsOutputDestination.CSV)
+                .build();
+
+        FILENAME = builder.define("filename", String.class)
+                .setDefaultValue("./analytics.csv")
+                .build();
+    }
+}
--- a/liquibase-standard/src/main/java/liquibase/analytics/CsvAnalyticsListener.java
+++ b/liquibase-standard/src/main/java/liquibase/analytics/CsvAnalyticsListener.java
@@ -1,0 +1,32 @@
+package liquibase.analytics;
+
+import liquibase.Scope;
+import org.apache.commons.io.FileUtils;
+import org.apache.commons.lang3.StringUtils;
+
+import java.io.File;
+import java.io.IOException;
+import java.nio.charset.StandardCharsets;
+
+public class CsvAnalyticsListener implements UsageAnalyticsListener {
+    @Override
+    public int getPriority() {
+        String filename = AnalyticsConfiguration.FILENAME.getCurrentValue();
+        AnalyticsOutputDestination destination = AnalyticsConfiguration.OUTPUT_DESTINATION.getCurrentValue();
+        if (StringUtils.isNotEmpty(filename) && AnalyticsOutputDestination.CSV.equals(destination)) {
+            return PRIORITY_SPECIALIZED;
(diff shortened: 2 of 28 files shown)


liquibase-standard/src/main/java/liquibase/command/CommandScope.java
```diff
@@ -203,74 +206,96 @@ public CommandResults execute() throws CommandExecutionException {
...
+            Scope.getCurrentScope().addMdcValue(MdcKey.LIQUIBASE_COMMAND_NAME, commandName);
         }
-        CommandResultsBuilder resultsBuilder = new CommandResultsBuilder(this, outputStream);
-        final List<CommandStep> pipeline = commandDefinition.getPipeline();
-        final List<CommandStep> executedCommands = new ArrayList<>();
-        Optional<Exception> thrownException = Optional.empty();
-        validate();
+        Event analyticsEvent = ExceptionUtil.doSilently(() -> {
+            return new Event(commandName);
+        });
```
StevenMassaro, line 215: This event should only be created if telemetry is enabled. This will be handled in a future story.

liquibase-standard/src/main/java/liquibase/analytics/CsvAnalyticsListener.java
```diff
@@ -0,0 +1,32 @@
...
+
+import liquibase.Scope;
+import org.apache.commons.io.FileUtils;
+import org.apache.commons.lang3.StringUtils;
+
+import java.io.File;
+import java.io.IOException;
+import java.nio.charset.StandardCharsets;
+
+public class CsvAnalyticsListener implements UsageAnalyticsListener {
```
StevenMassaro, line 11: This class should be removed, along with its enum once the Segment listener is added.

liquibase-standard/src/main/java/liquibase/analytics/AnalyticsConfiguration.java
```diff
@@ -0,0 +1,22 @@
+package liquibase.analytics;
+
+import liquibase.configuration.AutoloadedConfigurations;
+import liquibase.configuration.ConfigurationDefinition;
+
+public class AnalyticsConfiguration implements AutoloadedConfigurations {
```
StevenMassaro, line 6: This class is largely temporary and will either be completely removed or changed significantly.

liquibase-standard/src/main/java/liquibase/util/LiquibaseLauncherSettings.java
StevenMassaro, line 1: @filipelautert will moving this class from liquibase-cli to liquibase-standard be a breaking change for anyone you can think of?

liquibase-standard/src/main/java/liquibase/util/VersionUtils.java
StevenMassaro, line 1: @filipelautert same question here

liquibase-standard/src/main/java/liquibase/analytics/UsageAnalyticsFactory.java
```diff
@@ -0,0 +1,27 @@
...
+    protected Class<UsageAnalyticsListener> getPluginClass() {
+        return UsageAnalyticsListener.class;
+    }
+
+    @Override
+    protected int getPriority(UsageAnalyticsListener obj, Object... args) {
+        return obj.getPriority();
+    }
+
+    public void handleEvent(Event event) {
```
filipelautert, line 20: Are we planing to add some threading around here to make it non-blocking?

filipelautert (changes requested):

liquibase-standard/src/main/java/liquibase/analytics/UsageAnalyticsFactory.java
StevenMassaro: I am not sure how valuable the threading here would be. The handleEvent call gets called when the `CommandScope#execute` method exits, so it's essentially right before Liquibase shuts down. So it would be in a separate thread but nothing else would be executing at the same time.

My intention is to have a timeout, so it attempts to deliver the telemetry for a set amount of time before giving up.

filipelautert: oh i see, so we are only sending the data after liquibase executes and not at startup. Alright then!

filipelautert (approved):

liquibase-standard/src/main/java/liquibase/util/LiquibaseLauncherSettings.java
filipelautert: It shouldn't be as cli first release to maven was with 4.29.0 and it has dependency issues, so no one should be using it by now.

liquibase-standard/src/main/java/liquibase/util/VersionUtils.java
filipelautert: Same answer here

Issue #5145, opened 2023-10-31
Liquibase loadData fails to insert UUID in mysql


gnu-lorien:
### Search first

- [X] I searched and no similar issues were found

### Description

When using loadData to load a UUID into a column in MySQL setting the type of the column in loadData causes the following failure when inserting the data:

`Caused by: liquibase.exception.DatabaseException: java.sql.SQLException: Incorrect string value: '\xAC\xED\x00\x05sr...' for column 'policy_id' at row 1`

If you change the type of the column in loadData to `STRING` this data is able to load, but that can cause issues for other dbms that need different special handling for uuids. In particular, the documentation here https://docs.liquibase.com/workflows/liquibase-community/working-with-uuids.html implies that using uuid as the conversion type is the correct one.

### Steps To Reproduce

I created a table with initial data on a MySQL database using the following files:

```
databaseChangeLog:
- changeSet:
    author: asayman
    changes:
    - createTable:
        columns:
        - column:
            constraints:
              nullable: false
              primaryKey: true
              unique: false
            name: policy_id
            type: UUID
        tableName: ProductPolicy
    - modifySql:
        dbms: mysql
        replace:
          replace: NVARCHAR(100)
          with: VARCHAR(100) CHARACTER SET UTF8MB4
    - loadData:
        columns:
        - column:
            header: policy_id
            name: policy_id
            type: uuid
        encoding: UTF-8
        file: ../changelog-initial-data/001-product-policy.csv
        quotechar: ''''
        relativeToChangelogFile: true
        separator: ','
        tableName: ProductPolicy
    id: 000-table-productpolicy
```

```
policy_id
1872003c-2b7c-4678-994a-0ff3c24f4797
e231b4aa-3d37-4736-9459-1d6f1e469e33
825bf43d-45fd-4a49-8820-1164bde70830
0fef3af7-efbd-4ef1-91b6-10ea32e02c9f
e1c07c47-59ec-4f14-b6a2-a1879c764d97
f0d1243b-0d3b-4b14-8336-9cd1c4de3176
651e2feb-46b9-4983-b463-dad8e2ff6816
5fdb4863-0872-4b69-9213-9938cd9c138f
```

which results in the following error:

```
####################################################
##   _     _             _ _                      ##
##  | |   (_)           (_) |                     ##
##  | |    _  __ _ _   _ _| |__   __ _ ___  ___   ##
##  | |   | |/ _` | | | | | '_ \ / _` / __|/ _ \  ##
[...]

PR #6212, opened 2024-08-18
[5145] Fix UUID implementation with MySQL to use proper suggested data type

qsevers23:
## Impact

**WIP PR** for #5145 

## Description

This PR fixes the UUID implementation for MySQL databases to properly store them as a BINARY(16) column. Currently, when creating a UUID column, it will be a char(36) due to defaults within `UUIDType`. However, as of MySQL 8, there is [UUID support in the form of binary(16) strings](https://dev.mysql.com/blog-archive/mysql-8-0-uuid-support/). 

## Things to be aware of

I modified the DataTypeFactory class to take into account a per database priority level. Currently, LiquibaseDataTypes can only have one priority for all database types via the DataTypeInfo attribute. Although, depending on what happens to the topic in the section below, this could be reverted.

## Things to worry about

Currently, binary(16) applies to both BlobType and UUIDType under the DataTypeInfo alias section. I could not locate any other types that shared an alias in the same manor. I think there could be some logic added somewhere that would take into account the columnConfig string from the create table script, but I think you lose context of that once you're in a different changeset. 

If reviewers can determine if MySQL should identify binary(16) as just UUID instead of Blob (or instead of trying to support both with some additional magic), then this PR could be simplified a bit.

The DataTypeFactory test that is failing was implemented in this commit: https://github.com/liquibase/liquibase/commit/885490063ffce79064f3a3fdde7f110fc75fc502

However, this is before the binary alias is added to BlobType info attribute

## Additional Context

There are some MySQL functions that I'm not sure could be implemented into the liquibase side, or more of a jdbc thing:
* UUID()
* uuid_to_bin()
* bin_to_uuid()


liquibase-standard/src/main/java/liquibase/datatype/core/UUIDType.java
```diff
@@ -48,6 +48,17 @@ public DatabaseDataType toDatabaseDataType(Database database) {
                 throw new RuntimeException(ex);
             }
         }
+        // Not sure if version specific yet
+//        try {
+//            if (database instanceof MySQLDatabase && (database.getDatabaseMajorVersion() == 8)) {
```
filipelautert, line 53: As per the [blog mentioned below](https://dev.mysql.com/blog-archive/mysql-8-0-uuid-support/) , uuid was introduced in Mysql 8
For MariaDB it was introduced[ in version 10.7 ](https://mariadb.com/kb/en/uuid-data-type/), so this PR could also support MariaDB.

liquibase-standard/src/main/java/liquibase/datatype/core/UUIDType.java
```diff
@@ -59,6 +70,10 @@ protected String otherToSql(Object value, Database database) {
         if (database instanceof MSSQLDatabase) {
             return "'" + value.toString().toUpperCase(Locale.ENGLISH) + "'";
         }
+        // MYSQL displays binary(16) uuids as lowercase: https://dev.mysql.com/blog-archive/mysql-8-0-uuid-support/
+        if (database instanceof MySQLDatabase && !(database instanceof MariaDBDatabase)) {
```
filipelautert, line 74: Seems it is the same for MariaDB.

filipelautert (changes requested):

liquibase-standard/src/main/java/liquibase/datatype/core/UUIDType.java
qsevers23: I think MariaDB is already implemented above on line 40. Is that what you meant?

liquibase-standard/src/test/groovy/liquibase/datatype/DataTypeFactoryTest.groovy
```diff
@@ -212,6 +212,7 @@ class DataTypeFactoryTest extends Specification {
         "mediumblob"                                   | new MySQLDatabase()    | "MEDIUMBLOB"                                   | BlobType      | false
         "mediumtext"                                   | new MySQLDatabase()    | "MEDIUMTEXT"                                   | ClobType      | false
         "real"                                         | new MySQLDatabase()    | "REAL"                                         | FloatType     | false
+        "uuid"                                         | new MySQLDatabase()    | "binary(16)"                                   | UUIDType      | false
```
filipelautert, line 215: This should be BINARY uppercase. When tests are run, `getDatabaseMajorVersion` is called from `AbstractJdbcDatabase` returning `999` .

filipelautert (changes requested):

filipelautert (approved):
Thanks for the improvement @qsevers23 !

qsevers23:
> Thanks for the improvement @qsevers23 !

No problem! However, I don't think this is ready to merge just yet. There is some `loadData` changes that could be added to this PR I think. As this stands right now, it'll blow up. See [this](https://github.com/liquibase/liquibase/issues/5145#issuecomment-2295368817) comment in the GH issue for more details.

I just need some assistance in figuring out where those changes need to be implemented. Spent a good few hours, but couldn't really figure it out. I think I got lost in the un/prepared statement stuff right before it reaches the JDBC layer and don't know all the placeholders without digging in more.

Any help/guidance in the right direction would be greatly appreciated!

filipelautert:
> No problem! However, I don't think this is ready to merge just yet. There is some `loadData` changes that could be added to this PR I think. As this stands right now, it'll blow up when inserting a UUID column for a MySQL db. See [this](https://github.com/liquibase/liquibase/issues/5145#issuecomment-2295368817) comment in the GH issue for more details.
> 
> I just need some assistance in figuring out where those changes need to be implemented. Spent a good few hours, but couldn't really figure it out. I think I got lost in the un/prepared statement stuff right before it reaches the JDBC layer and don't know all the placeholders without digging in more.
> 
> Any help/guidance in the right direction would be greatly appreciated!

@qsevers23 whops, seems I jumped the gun on it.

UUIDType already return the correct LOAD_DATA_TYPE.UUID;  the problem may be around https://github.com/liquibase/liquibase/blob/732dc734735d749c8f3f7541d455a6172fba73a1/liquibase-standard/src/main/java/liquibase/change/core/LoadDataChange.java#L575 
the `fromDescription` method needs to understand that a Binary(16) for Mysql in fact is an UUID. I believe that fixing it here will solve this problem.

qsevers23:
> > No problem! However, I don't think this is ready to merge just yet. There is some `loadData` changes that could be added to this PR I think. As this stands right now, it'll blow up when inserting a UUID column for a MySQL db. See [this](https://github.com/liquibase/liquibase/issues/5145#issuecomment-2295368817) comment in the GH issue for more details.
> > I just need some assistance in figuring out where those changes need to be implemented. Spent a good few hours, but couldn't really figure it out. I think I got lost in the un/prepared statement stuff right before it reaches the JDBC layer and don't know all the placeholders without digging in more.
> > Any help/guidance in the right direction would be greatly appreciated!
> 
> @qsevers23 whops, seems I jumped the gun on it.
> 
> UUIDType already return the correct LOAD_DATA_TYPE.UUID; the problem may be around
> 
> https://github.com/liquibase/liquibase/blob/732dc734735d749c8f3f7541d455a6172fba73a1/liquibase-standard/src/main/java/liquibase/change/core/LoadDataChange.java#L575
> 
> the `fromDescription` method needs to understand that a Binary(16) for Mysql in fact is an UUID. I believe that fixing it here will solve this problem.

@filipelautert I think I may have gotten it figured out! Needed to convert the `UUID` to `byte[]` for the insert. Let me know what you think of the latest changes. Thanks!

liquibase-standard/src/main/java/liquibase/datatype/core/UUIDType.java
```diff
@@ -10,7 +10,7 @@
 
 import java.util.Locale;
 
-@DataTypeInfo(name = "uuid", aliases = { "uniqueidentifier", "java.util.UUID" }, minParameters = 0, maxParameters = 0, priority = LiquibaseDataType.PRIORITY_DEFAULT)
+@DataTypeInfo(name = "uuid", aliases = { "binary(16)", "binary", "uniqueidentifier", "java.util.UUID" }, minParameters = 0, maxParameters = 0, priority = LiquibaseDataType.PRIORITY_DATABASE)
```
filipelautert, line 13: Right now binary is defined in BlobType  - this may cause issues for other databases :/
Maybe you need  to replicate some of this logic around `BlobType:73` ?

filipelautert (changes requested):

liquibase-standard/src/main/java/liquibase/datatype/core/UUIDType.java
qsevers23: Hmm, that is a big change for the tests. Should have ran them locally before pinging you.

I'm not sure if that line in BlobType will help any here, it seems to be a priority coming from the registry located here: `liquibase/datatype/DataTypeFactory.java:184`. There seems to be a datatype priority available in the `DataTypeInfo` attribute, but it's the same across the board (and blows up a lot of tests when you modify it from default). 

There needs to be a priority for a type, for a given database I think. Without the priority change, this just about works but the registry pulls blob, and uuid (in that order) and then just returns on blob because mysql supports it, and it's at index 0 for the registry classes. 

I went ahead and tried to implement a database priority, let me know what you think.

I also removed a test for mapping "BINARY(16)" -> BlobType for MySQL, I'm not sure if anyone would use that column type for anything else in MySQL?

filipelautert:
@qsevers23 an unrelated test failed... this is getting too complicated.
And what if you implement a specific class to handle mysql uuid in the likes of `BinaryTypeSnowflake `? This would apply only to mysql and it would be easier to handle all of this binary/uuid mess in there.

lucas-it:
Hello, is there anything new here? I am currently facing this problem and need this fix.

Thanks!

filipelautert:
@qsevers23 I'll close this PR as https://github.com/liquibase/liquibase/pull/6649 is fixing the related issue.
If you still wish to implement support for binary(16), I think we would need a flag or something like this to turn it on/off to keep backwards compatibility.

PR #6218, opened 2024-08-19
Do not release lock if not locked by this update  DAT-18370

wwillard7800:
## Impact

### DIFF
--- a/liquibase-extension-testing/pom.xml
+++ b/liquibase-extension-testing/pom.xml
@@ -72,7 +72,7 @@
         <dependency>
             <groupId>org.slf4j</groupId>
             <artifactId>slf4j-jdk14</artifactId>
-            <version>2.0.13</version>
+            <version>2.0.16</version>
         </dependency>
 
         <dependency>
--- a/liquibase-extension-testing/pom.xml
+++ b/liquibase-extension-testing/pom.xml
@@ -72,7 +72,7 @@
         <dependency>
             <groupId>org.slf4j</groupId>
             <artifactId>slf4j-jdk14</artifactId>
-            <version>2.0.16</version>
+            <version>2.0.13</version>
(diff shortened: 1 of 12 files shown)


suryaaki2 (approved):

liquibase-standard/src/main/java/liquibase/executor/jvm/JdbcExecutor.java
```diff
@@ -188,7 +188,28 @@ public void execute(final SqlStatement sql, final List<SqlVisitor> sqlVisitors)
     private void setParameters(final PreparedStatement pstmt, final RawParameterizedSqlStatement sql) throws SQLException {
         final List<Object> parameters = sql.getParameters();
         for (int i = 0; i < parameters.size(); i++) {
-            pstmt.setObject(i + 1, parameters.get(i));
+            Object parameter = parameters.get(i);
```
filipelautert, line 191: Seems we are bringing some commits from master here

liquibase-extension-testing/pom.xml
```diff
@@ -72,7 +72,7 @@
         <dependency>
             <groupId>org.slf4j</groupId>
             <artifactId>slf4j-jdk14</artifactId>
-            <version>2.0.13</version>
+            <version>2.0.16</version>
```
filipelautert, line 75: Seems we are bringing some commits from master here too

filipelautert (changes requested):

liquibase-standard/src/main/java/liquibase/executor/jvm/JdbcExecutor.java
wwillard7800: Good catch

filipelautert (approved):

PR #6265, opened 2024-08-27
Dat 18350

wwillard7800:
The solution for this story turned out to be a little more involved than just adding the attribute to `include/includeAll` and propagating the value, because the `logicalFilePath` setting changes the DBCL key by replacing the file path with the logical path. This would cause the addition of another DBCL entry.  This situation would occur if the user just modified the changelog without running `changelogsync`.  

Also, I have modified the SQL changelog parser to calculate a unique changeset ID for new change sets, rather than just hard-coding it to raw.  This change is needed because if the user sets the logical file path attribute, then all change sets would end up with the same key of raw::includeAll::<logical file path>.  If the change set has already been executed, then the ID stays the same.

## Impact


filipelautert (approved):
Nice approach to make sure that we don't break what is already there.

Issue #6061, opened 2024-07-01
CustomTaskRollback rollback method runs twice


ThePersonThat:
### Search first

- [X] I searched and no similar issues were found

### Description

I have exactly the same issue as described here: https://github.com/liquibase/liquibase/issues/5314, but with rollback. 
As far as I investigated the fix has not been added for rollback logic: https://github.com/liquibase/liquibase/blob/master/liquibase-standard/src/main/java/liquibase/change/custom/CustomChangeWrapper.java#L197. Rollback logic is ran twice (while executing "addSqlMdc" and "executeStatements" methods)

### Steps To Reproduce

Create any CustomTaskRollback implementation. It will execute twice.

### Expected/Desired Behavior

Rollback logic must be executed only once

### Liquibase Version

4.26.0

### Are you willing to submit a PR?

- [ ] I'm willing to submit a PR (Thank you!)

tati-qalified:
Hi @ThePersonThat, thank you for reporting this issue.
I found that it had been reported before: #5422. Would you be willing to submit a PR to fix the bug? Our development team is available to provide guidance if necessary.

Thank you!
Tatiana

momosetkn:
this problem was solved by [4.30.0](https://github.com/liquibase/liquibase/releases/tag/v4.30.0)

Issue #5422, opened 2023-12-26
CustomTask ROLLBACK executed twice


marianomerlo:
### Search first

- [X] I searched and no similar issues were found

### Description

This is very similar to what was reported here: https://github.com/liquibase/liquibase/issues/5314 but for the rollback command instead of the update. 

I've gone through all the conversation in there, forked the project used to reproduce #5314 and added the rollback operation to give you a minimal project to reproduce the rollback issue as well: 

https://github.com/marianomerlo/liquibase-issue-custom-task-change

The latest working release for the rollback seems to be 4.19.0. 

### Steps To Reproduce

1. Create any custom change with rollback.
2. Perform a liquibase update
3. Perform a liquibase rollback

### Expected/Desired Behavior

Rollback operation should be executed once

### Liquibase Version

4.25.1

### Database Vendor & Version

PostgresSQL 13.11

### Are you willing to submit a PR?

- [ ] I'm willing to submit a PR (Thank you!)

tati-qalified:
Hi @marianomerlo, thank you for reporting this issue. I was able to replicate it with Liquibase version 4.25.1.

I'll forward this to the development team so they can look into it. If you have any suggestions on how to fix it, you can create a PR and we'll guide you when necessary.

Thank you,
Tatiana

momosetkn:
this problem was solved by 4.30.0

PR #6266, opened 2024-08-27
fix CustomTaskRollback twice call rollback bug

momosetkn:
## Impact

## Description

https://github.com/liquibase/liquibase/issues/5422
https://github.com/liquibase/liquibase/issues/6061

this bug is CustomTaskRollback twice call rollback. 
fix to call once by this PR.

## Things to worry about

I fixed bug. but I think it's better to avoid the global variable Scope.getCurrentScope...because it's implicit.


liquibase-standard/src/main/java/liquibase/change/custom/CustomChangeWrapper.java
```diff
@@ -192,14 +192,17 @@ public SqlStatement[] generateStatements(Database database) {
...
-                throw new RollbackImpossibleException("Unknown rollback type: " + customChange.getClass().getName());
+            if (shouldExecute) {
+                if (customChange instanceof CustomSqlRollback) {
+                    statements = ((CustomSqlRollback) customChange).generateRollbackStatements(database);
+                } else if (customChange instanceof CustomTaskRollback) {
+                    ((CustomTaskRollback) customChange).rollback(database);
+                } else {
+                    throw new RollbackImpossibleException("Unknown rollback type: " + customChange.getClass().getName());
+                }
             }
```
momosetkn, line 206: same to follow code
https://github.com/liquibase/liquibase/blob/v4.29.1/liquibase-standard/src/main/java/liquibase/change/custom/CustomChangeWrapper.java#L163-L174

liquibase-integration-tests/src/test/java/liquibase/dbtest/AbstractIntegrationTest.java
```diff
@@ -565,17 +566,22 @@ public void testUpdateClearUpdate() throws Exception {
 
     @Test
     @SuppressWarnings("squid:S2699") // Successful execution qualifies as test success.
-    public void testRollbackableChangeLog() throws Exception {
+    public synchronized void testRollbackableChangeLog() throws Exception {
```
momosetkn, line 569: Just to be sure add synchronized, for static property executeCallCount and rollbackCallCount.

momosetkn:
@MalloD12 
Run FOSSA Test is failed, What should I do? I don't know the cause for failed github-actions.

MalloD12:
> @MalloD12 Run FOSSA Test is failed, What should I do? I don't know the cause for failed github-actions.

Hi @momosetkn,

Don't worry by this FOSSA check failing, that's a known issue for our devOps team and it's not related to this change.

liquibase-integration-tests/src/test/java/liquibase/dbtest/AbstractIntegrationTest.java
MalloD12: This is a test, I'm not sure to understand why it would need a sync keyword. I don't think this is needed here.

momosetkn: how to care for static property?
maybe, alternative way was check call count the `execute` and `rollback` method with Mockito. I think that's the way to do it politely.

momosetkn: I found alternative way. it is Check ExampleCustomTaskChange with executed `liquibase.getDatabaseChangeLog().getChangeSets()`.
I will try it.

momosetkn:
```java
        List<CustomChange> customChanges = liquibase.getDatabaseChangeLog().getChangeSets().stream().flatMap(changeSet ->
                changeSet.getChanges().stream().filter(change ->
                        change instanceof CustomChangeWrapper &&
                                ((CustomChangeWrapper) change).getCustomChange() instanceof ExampleCustomTaskChange
                ).map(change ->
                        ((CustomChangeWrapper) change).getCustomChange())
        ).collect(Collectors.toList());
        for (CustomChange customChange : customChanges) {
            assertEquals(1, ((ExampleCustomTaskChange)customChange).executeCallCount);
            assertEquals(0, ((ExampleCustomTaskChange)customChange).rollbackCallCount);
        }
```

I wrote this code. but cannot check call-count, actual value is `0`. because, difference ExampleCustomTaskChange object.

MalloD12: I was going to suggest making ExampleCustomTaskChange a singleton class, and then checking `rollbackcount` but I think this won't help with this issue. Let me ask tomorrow one of our team members with more experience/knowledge on custom changes because my guess is rollback is executing twice because there might be something else calling also this same method.

MalloD12: Hi @momosetkn,

I was talking about this with someone else from the team, and we agreed that `synchronized` keyword is not needed there. To validate counts, we think maybe what we can do here is to have a map and save your `ExampleCustomTaskChange` instances there, then save this map in the current scope.  I was thinking of adding something like this in `CustomChangeWrapper`, but I'm unsure if this is the right place. It would probably be if some other custom changes implement something like this, but for now, we can probably do this in the scope you were trying to do this validation.

Thanks,
Daniel.

momosetkn: I see, ThreadLocal(map in the current scope) can resolve synchronous problem.
I fix.

momosetkn: Sorry for the late fix.
I fixed.

momosetkn: @MalloD12 
Sorry for the late fix.
I fixed.

liquibase-integration-tests/src/test/java/liquibase/dbtest/AbstractIntegrationTest.java
```diff
@@ -569,21 +570,38 @@ public void testUpdateClearUpdate() throws Exception {
...
-        liquibase = createLiquibase(rollbackChangeLog);
-        liquibase.update(this.contexts);
+            Map<String, Object> map = Scope.getCurrentScope().get(ExampleCustomTaskChange.SCOPE_ATTR_CALL_COUNT_MAP, Map.class);
+            Integer executeCallCountInteger = (Integer) map.get(ExampleCustomTaskChange.COUNT_MAP_EXECUTE_CALL_COUNT);
+            Integer rollbackCallCount = (Integer)map.get(ExampleCustomTaskChange.COUNT_MAP_ROLLBACK_CALL_COUNT);
+            // call `execute` method
+            assertEquals(1, (int)executeCallCountInteger);
+            assertEquals(0, (int)rollbackCallCount);
+
+            liquibase = createLiquibase(rollbackChangeLog);
```
MalloD12, line 591: Hey @momosetkn we don't need to create several liquibase instances if they are all using the same changelog. I would leave only one of these.

That's my only suggestion to update on this test,

Thanks,
Daniel.

momosetkn: ok, I fixed!

MalloD12 (approved):
**Approved**.

Code changes and test looks good to me.

Thank you, @momosetkn!

filipelautert (approved):
Thanks for the PR @momosetkn !

@wwillard7800 could you review it too?

tati-qalified (approved):
Looks great, thank you for your contribution @momosetkn!

rberezen (approved):

PR #6267, opened 2024-08-27
load telemetry configuration from remote server (DAT-18404)

StevenMassaro:
## Impact


liquibase-standard/src/main/java/liquibase/telemetry/configuration/SegmentTelemetryConfiguration.java
```diff
@@ -1,20 +1,31 @@
...
-        );
+        String url = TelemetryArgs.CONFIG_ENDPOINT_URL.getCurrentValue();
+        AtomicReference<RemoteTelemetryConfiguration> remoteTelemetryConfiguration = new AtomicReference<>();
+        Thread thread = new Thread(() -> {
+            try {
+                InputStream input = new URL(url).openStream();
+                Yaml yaml = new Yaml();
+                remoteTelemetryConfiguration.set(yaml.loadAs(input, RemoteTelemetryConfiguration.class));
+            } catch (Exception e) {
+                Scope.getCurrentScope().getLog(SegmentTelemetryConfiguration.class).fine("Failed to load telemetry configuration from " + url, e);
```
filipelautert, line 23: When this fails, do we provide some kind of default settings? Or telemetry gets disabled? Or we get a `nullpointer` on `get() ` below?

filipelautert (changes requested):

liquibase-standard/src/main/java/liquibase/telemetry/configuration/SegmentTelemetryConfiguration.java
StevenMassaro: This will return a null pointer on the gets below, but those exceptions are swallowed because all exceptions are swallowed, thus effectively disabling telemetry.

filipelautert (approved):

Issue #6281, opened 2024-08-29
liquibase.exception.CommandExecutionException: java.lang.NullPointerException: Cannot invoke "String.replaceAll(String, String)" because "defaultValue" is null


xeagle2:
### Search first

- [X] I searched and no similar issues were found

### Description

NPE is thrown when using a connection to the SAP database.

At class **liquibase.snapshot.jvm.ColumnSnapshotGenerator** (**method: readDefaultValue**), there is **no check defaultValue for null** in a code provided as a snippet below

```
if (database instanceof SybaseASADatabase) {
            String defaultValue = (String) columnMetadataResultSet.get(COLUMN_DEF_COL);

           // SQL Anywhere returns `CURRENT DATE` (without underscore), which no other RDBMS would understand
           defaultValue = defaultValue.replaceAll("(?i)\\bCURRENT\\s+DATE\\b", "{fn CURDATE()}");

           // SQL Anywhere returns `CURRENT TIME` (without underscore), which no other RDBMS would understand
           defaultValue = defaultValue.replaceAll("(?i)\\bCURRENT\\s+TIME\\b", "{fn CURTIME()}");

           // SQL Anywhere returns `CURRENT TIMESTAMP` (without underscore), which no other RDBMS would understand
           defaultValue = defaultValue.replaceAll("(?i)\\bCURRENT\\s+TIMESTAMP\\b", "{fn NOW()}");

           // SQL Anywhere returns `CURRENT USER` (without underscore), which no other RDBMS would understand
           defaultValue = defaultValue.replaceAll("(?i)\\bCURRENT\\s+USER\\b", "{fn USER()}");

           columnMetadataResultSet.set(COLUMN_DEF_COL, defaultValue);

           if ("YES".equals(columnMetadataResultSet.get("IS_GENERATEDCOLUMN"))) {
               Object virtColumnDef = columnMetadataResultSet.get(COLUMN_DEF_COL);
               if ((virtColumnDef != null) && !"null".equals(virtColumnDef)) {
                   columnMetadataResultSet.set(COLUMN_DEF_COL, "COMPUTE (" + virtColumnDef + ")");
               }
           }
        }
```

```
liquibase.exception.CommandExecutionException: java.lang.NullPointerException: Cannot invoke "String.replaceAll(String, String)" because "defaultValue" is null
	at liquibase.command.CommandScope.execute(CommandScope.java:237)
	at liquibase.integration.commandline.CommandRunner.call(CommandRunner.java:55)
[...]

PR #6287, opened 2024-08-30
#6281 fix NPE when default value is null

asolntsev:
Sample:
```java
java.lang.NullPointerException: Cannot invoke "String.replaceAll(String, String)" because "defaultValue" is null
	at liquibase.snapshot.jvm.ColumnSnapshotGenerator.readDefaultValue(ColumnSnapshotGenerator.java:571)
	at liquibase.snapshot.jvm.ColumnSnapshotGenerator.readColumn(ColumnSnapshotGenerator.java:291)
	at liquibase.snapshot.jvm.ColumnSnapshotGenerator.addTo(ColumnSnapshotGenerator.java:204)
	at liquibase.snapshot.jvm.JdbcSnapshotGenerator.snapshot(JdbcSnapshotGenerator.java:78)
	at liquibase.snapshot.SnapshotGeneratorChain.snapshot(SnapshotGeneratorChain.java:49)
```

## Impact

## Description

This PR fixes bug https://github.com/liquibase/liquibase/issues/6281

### DIFF
--- a/liquibase-standard/src/main/java/liquibase/snapshot/jvm/ColumnSnapshotGenerator.java
+++ b/liquibase-standard/src/main/java/liquibase/snapshot/jvm/ColumnSnapshotGenerator.java
@@ -580,19 +580,19 @@ public class ColumnSnapshotGenerator extends JdbcSnapshotGenerator {
 
         if (database instanceof SybaseASADatabase) {
             String defaultValue = (String) columnMetadataResultSet.get(COLUMN_DEF_COL);
+            if (defaultValue != null) {
+                // SQL Anywhere returns `CURRENT DATE` (without underscore), which no other RDBMS would understand
+               defaultValue = defaultValue.replaceAll("(?i)\\bCURRENT\\s+DATE\\b", "{fn CURDATE()}");
 
-           // SQL Anywhere returns `CURRENT DATE` (without underscore), which no other RDBMS would understand
-           defaultValue = defaultValue.replaceAll("(?i)\\bCURRENT\\s+DATE\\b", "{fn CURDATE()}");
+               // SQL Anywhere returns `CURRENT TIME` (without underscore), which no other RDBMS would understand
+               defaultValue = defaultValue.replaceAll("(?i)\\bCURRENT\\s+TIME\\b", "{fn CURTIME()}");
 
-           // SQL Anywhere returns `CURRENT TIME` (without underscore), which no other RDBMS would understand
-           defaultValue = defaultValue.replaceAll("(?i)\\bCURRENT\\s+TIME\\b", "{fn CURTIME()}");
+               // SQL Anywhere returns `CURRENT TIMESTAMP` (without underscore), which no other RDBMS would understand
+               defaultValue = defaultValue.replaceAll("(?i)\\bCURRENT\\s+TIMESTAMP\\b", "{fn NOW()}");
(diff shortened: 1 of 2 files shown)


liquibase-standard/src/main/java/liquibase/snapshot/jvm/ColumnSnapshotGenerator.java
```diff
@@ -574,30 +575,32 @@ protected Object readDefaultValue(CachedRow columnMetadataResultSet, Column colu
...
+               // SQL Anywhere returns `CURRENT USER` (without underscore), which no other RDBMS would understand
+               defaultValue = defaultValue.replaceAll("(?i)\\bCURRENT\\s+USER\\b", "{fn USER()}");
+           }
 
            columnMetadataResultSet.set(COLUMN_DEF_COL, defaultValue);
 
            if (YES_VALUE.equals(columnMetadataResultSet.get(IS_GENERATED_COLUMN))) {
                Object virtColumnDef = columnMetadataResultSet.get(COLUMN_DEF_COL);
-               if ((virtColumnDef != null) && !StringUtil.equalsWordNull(virtColumnDef.toString())) {
+               if ((virtColumnDef != null) && !"null".equalsIgnoreCase(virtColumnDef.toString())) {
```
MalloD12, line 603: I have a question about this change. Was it done because StringUtil is a deprecated class? If that's the reason behind this change I think we should revert it back as it was because we have an open PR to undeprecate `StringUtil` class.

asolntsev: @MalloD12 Probably yes, because of deprecation.
BUT
I don't see any reasons to revert this change. 
Expression `"null".equalsIgnoreCase` is more short and absolutely clear. You don't need to look into documentation to quickly understand what id does.
Definitely better than `StringUtil.equalsWordNull`.

MalloD12: Hi @asolntsev, 

I think it's maybe more a matter of preference.   In the past, we worked on an issue to make this type of replacement by the equivalent `StringUtil` methods (see #3718).

asolntsev: Yes, but in our case we are not replacing `StringUtil` by `StringUtils`.
We are replacing `StringUtil` by a native Java method. Which is (at least) no longer or harder to read.

MalloD12: I very much appreciate your intention to keep improving Liquibase code, but it was a team preference we decided to make in the past to update this kind of conditional as specified in #3718.

Thanks,
Daniel.

asolntsev: Ok, I reverted all StringUtils-related changes and rebased the branch from the latest master.

MalloD12 (approved):
Code changes looks good to me. Thank you @asolntsev for this PR and for addressing our review comments. 

Daniel.

rberezen (approved):

filipelautert (approved):

PR #6286, opened 2024-08-30
#6264 append the cause of the problem to thrown exception

asolntsev:
The most important fix is line 271:
"Could not find/read changelogs from %s directory" - WHY???

## Impact

- [x] Bug fix (non-breaking change which fixes expected existing functionality)
- [ ] Enhancement/New feature (adds functionality without impacting existing logic)
- [ ] Breaking change (fix or feature that would cause existing functionality to change)
 

## Description

This PR fixes problem https://github.com/liquibase/liquibase/issues/6264

More exactly, it adds the error details to thrown exception (which was previously swallowed).

Issue #6264, opened 2024-08-27
includeAll not working when used through a JAR


Uchiha268:
### Search first

- [X] I searched and no similar issues were found

### Description

I have the following file structure - 
```
src
    └── main
        ├── java
        └── resources
            └── db
                └── changelog
                    ├── db.changelog-master.xml
                    └── releases
                        ├── release1.xml
                        ├── release2.xml
```
                        
And within db.changelog-master.xml I am referring to all the changelings in the releases directory in the following way :- `<includeAll path="releases" relativeToChangelogFile="true" errorIfMissingOrEmpty="true"/>`

The liquibase update then runs perfectly while running through IntelliJ IDEA, however when I create a JAR of the same liquibase is unable to find the directory releases.

```
Caused by: java.io.IOException: Could not find/read changelogs from releases directory
	at liquibase.changelog.DatabaseChangeLog.findResources(DatabaseChangeLog.java:866) ~[liquibase-core-4.29.1.jar!/:na]
	... 40 common frames omitted
```

I have tried to change how I refer to the folder by not using relative path but by using the class path as well :-
`<includeAll path="db/changelog/releases" errorIfMissingOrEmpty="true"/>`

However again it runs perfectly through IntelliJ, applying the changelogs but fails while running the JAR of it.

```
Caused by: java.io.IOException: Could not find/read changelogs from db/changelog/releases directory
	at liquibase.changelog.DatabaseChangeLog.findResources(DatabaseChangeLog.java:866) ~[liquibase-core-4.29.1.jar!/:na]
	... 40 common frames omitted
```

I am using liquibase version 4.25.1 but I found the same issue while using the new version, 4.29.1 as well. I have also checked whether the changelogs were part of the jar by extracting the jar created and they are exactly where they should be. And that doesn't also seem to be the problem since when I refer to each changelog file individually in the master using include rather then includeAll, then everything works.

### Steps To Reproduce

The following is how I am running the liquidate update command in the java application - 

```
try {
            CommandScope update = new CommandScope("update");

            update.addArgumentValue("changelogFile", "db/changelog/db.changelog-master.xml");
            update.addArgumentValue("url", URL);
            update.addArgumentValue("username", USER);
            update.addArgumentValue("password", PASSWORD);

            update.execute();
        } catch (CommandExecutionException e) {
            logger.error("Liquibase Update failed with error message: " + e.getMessage());
            throw e;
}
```

### Expected/Desired Behavior

[...]

PR #6288, opened 2024-08-30
#6264 avoid ignoring the caught exception

asolntsev:
## Impact

Similar to https://github.com/liquibase/liquibase/pull/6286, but in many other places.

I found all places where IDEA shows warning "catch block may ignore exception", and fixed many of them.


lbadam:
@asolntsev thanks so much for submitting this PR. We'll review it and get back to you soon.

liquibase-standard/src/main/java/liquibase/executor/jvm/StatementCreatorUtils.java
```diff
@@ -80,7 +82,7 @@ private static void setParameterValueInternal(
                     useSetObject = (ps.getConnection().getMetaData().getDatabaseProductName().contains("Informix"));
                 }
                 catch (Throwable ex) {
-//                    logger.debug("Could not check database product name", ex);
+                    Scope.getCurrentScope().getLog(StatementCreatorUtils.class).fine("Could not check database product name", ex);
```
MalloD12, line 85: Something very minor, but I would keep the same log level (debug) here or at least change it as a `warning`.

MalloD12:
Hhi @asolntsev,

Thank you for submitting another PR for Liquibase. In general, this PR looks good to me. The only thing I have for commenting here is in some cases we are logging an error message in the `catch` block as a `warning` or even maybe as `info` log message. I think maybe we can update these to the right log level. What do you think about it?

Thanks,
Daniel.

liquibase-standard/src/main/java/liquibase/executor/jvm/StatementCreatorUtils.java
asolntsev: @MalloD12 But "fine" is the same as "debug". :)

Look, this is javadoc of `liquibase.logging.Logger`:

image

asolntsev:
> I think maybe we can update these to the right log level

Generally I like the idea :)

But I intentionally didn't change any logic to keep the PR scope as small as possible. 
In some cases, "info" might be totally fine:
1. e.g. if LiquiBase could not get DB version because DB connection has been closed by this moment
2. or if DB doesn't support such a function.

liquibase-standard/src/main/java/liquibase/executor/jvm/StatementCreatorUtils.java
MalloD12: Yeah, according to that class, it's correct what you have done. But I'm not sure why `debug` methods are deprecated here. I'll check with the team. 

Thanks,
Daniel.

MalloD12 (approved):
**Approved**.

PR looks good , I had only a doubt regarding some log level messages but after reviewing with the team we are ok as it is. 

Thank you, @asolntsev for one other PR.

rberezen (approved):

liquibase-standard/src/main/java/liquibase/datatype/core/UUIDType.java
```diff
@@ -42,11 +42,7 @@ public DatabaseDataType toDatabaseDataType(Database database) {
                 return new DatabaseDataType("UUID");
             }
         } catch (DatabaseException e) {
-            try {
```
filipelautert, line 45: lol this one was fun

filipelautert (approved):

PR #6294, opened 2024-09-03
[DAT-18534] chore: suppress UnsupportedClassVersionErrors for Bigquery at info level

filipelautert:
chore: do not log UnsupportedClassVersionErrors for Bigquery at info level - but remain logging them at fine (debug) level.


suryaaki2 (approved):

StevenMassaro (approved):
I'd prefer we not have bigquery code in open source, is there a way to move this logic to the bigquery extension?

filipelautert:
> I'd prefer we not have bigquery code in open source, is there a way to move this logic to the bigquery extension?

sorry, it's not possible - it's not even being loaded. But when we drop liquibase 8 support (Liquibase 5? ) we can remove this code.

PR #6308, opened 2024-09-10
Handle offline connection for Snowflake during diff DAT-18652

wwillard7800:
## Impact


abrackx (approved):
Is this only impacting snowflake? Based on the code change it seems like all dbms would bee impacted.

wwillard7800:
> Is this only impacting snowflake? Based on the code change it seems like all dbms would bee impacted.

Good catch @abrackx .  I should have added this check in the `SnowflakeJdbcExecutor` override of the `execute` method.

PR #6318, opened 2024-09-12
Always add database information to MDC and analytics when running via maven (DAT-18630)

abrackx:
## Impact

## Description
When executing a maven goal the database command scope argument is set up before the DbUrlConnectionCommandStep is executed. This meant that occasionally we didn't actually log all of the relevant database platform information in MDC and analytics. This change logs the mdc information regardless of whether the database was set up in the command step or elsewhere. 

## Things to be aware of
This change removes an if/else that caused the bug and refactors the code to be a bit more compact, but functionally nothing should change.


StevenMassaro (approved):
Nice!

filipelautert (approved):

PR #6321, opened 2024-09-13
put integration details in scope when running via spring (DAT-18605)

StevenMassaro:
## Impact

### DIFF
--- a/liquibase-standard/src/main/java/liquibase/integration/IntegrationDetails.java
+++ b/liquibase-standard/src/main/java/liquibase/integration/IntegrationDetails.java
@@ -1,6 +1,8 @@
 package liquibase.integration;
 
+import lombok.AllArgsConstructor;
 import lombok.Getter;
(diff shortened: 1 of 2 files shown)


liquibase-standard/src/main/java/liquibase/integration/IntegrationDetails.java
StevenMassaro, line 1: @filipelautert - wondering if you have any thoughts on this. I made some changes to this class that are not backwards compatible with previous versions of Liquibase. So if you try to use the new Spring code with an older version of Liquibase (which I think isn't possible...) then this wouldn't work. Is this a risk or do you think this is safe to do?

wwillard7800 (approved):

liquibase-standard/src/main/java/liquibase/integration/IntegrationDetails.java
filipelautert: You are right, this is not possible - so no worries!

PR #6328, opened 2024-09-17
Added a stub attribute to CommandStep interface so commands can be recognized as stubs

wwillard7800:
DAT-18431

This work allows the original check command steps to be marked as stubs.  Any stubs will be removed from the pipeline if a real command is discovered.

## Impact


filipelautert (approved):

StevenMassaro:
I don't think you need this. You could use the existing `@CommandOverride` annotation on the "real" command.

wwillard7800:
> I don't think you need this. You could use the existing `@CommandOverride` annotation on the "real" command.

I'll try to use the annotation so I can back this change out.

liquibase-standard/src/main/java/liquibase/command/CommandFactory.java
```diff
@@ -186,19 +190,41 @@ public SortedSet<CommandDefinition> getCommands(boolean includeInternal) {
...
-
+    private static String findKeyInMapIgnoreCase(String key, Map<String, String[]> map) {
+        for (Map.Entry<String, String[]> mapEntry : map.entrySet()) {
+            String actualKey = mapEntry.getKey();
+            if (actualKey.equalsIgnoreCase(key)) {
+                return actualKey;
+            }
+        }
+        return null;
+    }
```
StevenMassaro, line 201: This method already exists in liquibase.util.CollectionUtil#findKeyInMapIgnoreCase and should not be duplicated here.

StevenMassaro (changes requested):

abrackx:
I know this would be a total rework but did you consider adding a `@CommandStub` annotation? This "allow duplicates" toggle feels a bit buried in the api.

wwillard7800:
> I know this would be a total rework but did you consider adding a `@CommandStub` annotation? This "allow duplicates" toggle feels a bit buried in the api. I won't hold this up to do that but I wanted to throw the idea out.

It would be a good idea, but even then, I think I'd still have the duplicate args problem.  If we have another situation like this, then we should address it.

liquibase-standard/src/main/java/liquibase/command/CommandFactory.java
wwillard7800: I refactored this to just use a case-insensitive Set.

PR #6330, opened 2024-09-17
DAT-18563 :: Unhide tag parameter for UpdateTestingRollback command

MalloD12:
## Impact

## Description

> During testing, @James Bennett discovered that the updateTestingRollback command has a hidden argument for tag.  The argument has always been a part of the command, but was accidentally marked as hidden during command framework refactoring.  It appears to be working as designed but should be tested further after unhiding it.

With this PR, we just unhide `TAG` command argument and update the corresponding tests.


MalloD12:
Related PD ticket: https://datical.atlassian.net/browse/PD-5112

@adrian-velonis1 - @AMBERMW13: I know you guys are already aware of PD-5112, but just tagging you here as well just in case.

rberezen (approved):

filipelautert (approved):

PR #6333, opened 2024-09-18
Log the database platform when running generate changelog in mdc and analytics (DAT-18647)

abrackx:
## Impact


MalloD12 (approved):

StevenMassaro (changes requested):
This will not work, because the logMdc method logs the provided db url as the target url, not the ref url.

PR #6335, opened 2024-09-18
Add analytics toggle to maven plugin (DAT-18633)

abrackx:
## Impact

### DIFF
--- a/liquibase-standard/src/main/java/liquibase/analytics/configuration/AnalyticsArgs.java
+++ b/liquibase-standard/src/main/java/liquibase/analytics/configuration/AnalyticsArgs.java
@@ -8,7 +8,7 @@ import org.apache.commons.lang3.BooleanUtils;
 
 public class AnalyticsArgs implements AutoloadedConfigurations {
 
-    private static final ConfigurationDefinition<Boolean> ENABLED;
+    public static final ConfigurationDefinition<Boolean> ENABLED;
     public static final ConfigurationDefinition<String> CONFIG_ENDPOINT_URL;
     public static final ConfigurationDefinition<Integer> CONFIG_ENDPOINT_TIMEOUT_MILLIS;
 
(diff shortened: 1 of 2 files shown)


liquibase-standard/src/main/java/liquibase/analytics/configuration/AnalyticsArgs.java
```diff
@@ -8,7 +8,7 @@
 
 public class AnalyticsArgs implements AutoloadedConfigurations {
 
-    private static final ConfigurationDefinition<Boolean> ENABLED;
+    public static final ConfigurationDefinition<Boolean> ENABLED;
```
StevenMassaro, line 11: Does this need to be public? I purposefully made it private so that you use the `isAnalyticsEnabled()` method instead, which has additional logic in it.

abrackx: I didn't want to hard code the key for this value, but I suppose I can. It feels better to use `AnalyticsArgs.ENABLED.getKey()` over the string `liquibase.analytics.enabled` IMO, but lmk what you'd prefer.

StevenMassaro: I guess I don't feel strongly about it, I see where you're coming from. We definitely do hardcode strings in other places in the maven plugin (yet another reason it needs to be refactored). I'll leave it up to you about whether you want to hardcode the string in the maven plugin, hardcode the strings in public variables in the AnalyticsArgs class (and then keep the `ENABLED` instance variable private and use those strings in the same class), or keep it public and add a comment so that people know not to use it directly, but to use the other method instead.

MalloD12: I am more in favor of Alex's approach. This kind of strings are indeed hardcoded in many places in the Maven plugin, but it would be maybe one less place to refactor later.

abrackx: Cool. I'll add some javadocs to the `ENABLED` arg that says to use the method for checking analytics enabled.

StevenMassaro (approved):

PR #6341, opened 2024-09-20
Fix sqlite hsqldb and environment variable tests (DAT-18646)

abrackx:
## Impact

## Description
Issue 1: The test `StrictIntegrationTests` would fail if there were pro (or other extension related) environment variables set up outside of the scope of test execution. The change to `EnvironmentVariableProviderForTest` makes it so that we do not read the existing environment variables when utilizing this provider and enables us to successfully run the test without messing with existing environment variables.

Issue 2: When running sqlite integration tests before hsqldb integration tests, the databaseobjectfactory would carry over between test executions. Since sqlite does not support sequences, this would lead to sequences not being detected in hsqldb's integration tests causing them to fail. The change to the teardown method in the sqlite tests resets the factory, allowing sequences to be dropped in subsequent tests.

### DIFF
--- a/liquibase-integration-tests/src/test/java/liquibase/dbtest/sqlite/SQLiteIntegrationTest.java
+++ b/liquibase-integration-tests/src/test/java/liquibase/dbtest/sqlite/SQLiteIntegrationTest.java
@@ -6,6 +6,8 @@ import liquibase.database.DatabaseFactory;
 import liquibase.dbtest.AbstractIntegrationTest;
 import liquibase.exception.ValidationFailedException;
 import liquibase.snapshot.DatabaseSnapshot;
+import liquibase.snapshot.SnapshotGeneratorFactory;
+import liquibase.structure.core.DatabaseObjectFactory;
 import org.junit.Test;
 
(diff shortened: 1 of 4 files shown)


wwillard7800 (approved):

MalloD12 (approved):

liquibase-integration-tests/src/test/java/liquibase/dbtest/sqlite/SQLiteIntegrationTest.java
```diff
@@ -86,6 +88,7 @@ public void testOutputChangeLog() throws Exception {
     @Override
     public void tearDown() throws Exception {
         super.tearDown();
+        DatabaseObjectFactory.getInstance().reset();
```
StevenMassaro, line 91: Should this go somewhere inside the AbstractIntegrationTest? Or wherever we run dropall, which I think is in the interceptor class?

StevenMassaro (approved):

tati-qalified (approved):

PR #6343, opened 2024-09-21
FIX #6342 - Liquibase Output changeset executes twice

### LINKED ISSUE #6342: Liquibase Output change executed twice
### Search first

### Description

When using a change type of `output`, the string to be output gets output twice instead of once.

It appears that with the output type, since it isn't actually a database change it is handled slightly differently than other change types. The `addSqlMdc` function of change generates the change to be done, but this triggers actually writing the output to the screen.

The actual execution of the change then occurs, and output the information a second time.

The `addSqlMdc` should not actually be executing the output.

### Steps To Reproduce

Create a changelog with an output change type

### Expected/Desired Behavior

Output only shows up once

### Liquibase Version

4.29.2

### Database Vendor & Version

_No response_

### Liquibase Integration

_No response_

### Liquibase Extensions

_No response_

_No response_

### Additional Context

_No response_

jasonlyle88:
## Impact

## Description

Fixes and leverages the existing SHOULD_EXECUTE logic to determine if the output change should actually be output to the buffer.

Thanks to @wwillard7800 for identifying the SHOULD_EXECUTE flag scope was not correctly set.

Once correctly set, the output change just checks if it should execute or not before actually executing.

## Things to be aware of

Fixes #6342


jasonlyle88:
@filipelautert @MalloD12 

Hey y'all! Would appreciate some help here on what I thought would be an easy fix. I noticed that the output change type is executing when the MDC logging is occurring and also when the change is actually executed (so the output is duplicated). Then I noticed that there is already code that sets a `SHOULD_EXECUTE` value to false when adding the change to the MDC log (`liquibase.changelog.ChangeSet` class's `addSqlMdc` method has a `scopeValues.put(Change.SHOULD_EXECUTE, Boolean.FALSE);` line).

However... I can't seem to figure out how to leverage this. In the `liquibase.change.core.OutputChange` class, I am modifying the generateStatemetnts method in order to look at the above scope value. I have tried the two following two ways:

```java
Boolean shouldExecute1 = Scope.getCurrentScope().get(Change.SHOULD_EXECUTE, Boolean.class);
Boolean shouldExecute2 = shouldExecuteChange(database);
```

but both are always `null` whether this is in the scope of the MDC logging stage or the execution phase.

What am I missing here? How can I appropriately access that value to avoid the output during the MDC logging stage?

Thank in advance for any help you can provide!

wwillard7800:
I believe I see the problem.  The `addSqlToMdc` method in `ChangeSet` does not properly set the `SHOULD_EXECUTE` flag.  It only sets it for the `generateStatements` call, and it needs to set it for the `getSqlString` call.

Like this:

```
    /**
     *
     * Adds changeset sql to mdc if the change is supported by the database
     * @param change the change to read sql from
     * @param database the database to generate change sql against
     * @param generateRollbackStatements controls generation of rollback sql statements or standard statements sql
     * @throws RollbackImpossibleException if you cannot generate rollback statements
     *
     */
    private String addSqlMdc(Change change, Database database, boolean generateRollbackStatements) throws Exception {
        //
        // If the change is for this Database
        // add a Boolean flag to Scope to indicate that the Change should not be executed when adding MDC context
        //
        if (! change.supports(database)) {
            return null;
        }
        AtomicReference<SqlStatement[]> statementsReference = new AtomicReference<>();
        Map<String, Object> scopeValues = new HashMap<>();
        scopeValues.put(Change.SHOULD_EXECUTE, Boolean.FALSE);
        StringBuilder sqlStatementsMdc = new StringBuilder();
        Scope.child(scopeValues, () -> {
            statementsReference.set(generateRollbackStatements ?
                    change.generateRollbackStatements(database) : change.generateStatements(database));
            sqlStatementsMdc.append(Arrays.stream(statementsReference.get())
                    .map(statement -> SqlUtil.getSqlString(statement, SqlGeneratorFactory.getInstance(), database))
                    .collect(Collectors.joining("\n")));
            Scope.getCurrentScope().addMdcValue(MdcKey.CHANGESET_SQL, sqlStatementsMdc.toString());
        });

        return sqlStatementsMdc.toString();
    }
```

jasonlyle88:
@wwillard7800 That did it, thank you very much!

I'm not sure I fully understand the change here... but I'm also not super familiar with this side of things, so not terribly surprised 😬

Thanks for the assist, I think you actually did more work for this PR than I did!

With that, I believe this PR is ready to be reviewed!

wwillard7800:
> @wwillard7800 That did it, thank you very much!
> 
> I'm not sure I fully understand the change here... but I'm also not super familiar with this side of things, so not terribly surprised 😬
> 
> Thanks for the assist, I think you actually did more work for this PR than I did!
> 
> With that, I believe this PR is ready to be reviewed!

No problem, glad to help.

Basically, for the other change types, having the `SHOULD_EXECUTE` flag be available during `change.generateStatements`  call was sufficient, but for the Output change, the call to `getSqlString` needed the same flag, so I just made it available for both.

MalloD12:
Code changes look good to me. I'm only wondering if we can add a test where we deploy a changelog containing for example:

```xml
<changeSet  author="mallod"  id="outputChangeTest">
        <output  target="STDERR">Make sure you feed the cat</output>
</changeSet>
```

and we make sure there is nothing logged. Otherwise, I think this is something we can manually test as well. @rberezen @tati-qalified, what do you guys think? 

Thanks,
Daniel.

jasonlyle88:
@MalloD12 

Testing is one thing I was unsure about. I wasn't aware of any way to check what is sent to stderr/stdout/whatever buffer and so wasn't sure if there was a good way to test this other than manual tests.

MalloD12:
@jasonlyle88 I was thinking of doing something like this:

```java
 BufferedLogService bufferLog = new BufferedLogService()

        Scope.child([
                (Scope.Attr.logService.name()) : bufferLog,
        ], {
            CommandScope commandScope = new CommandScope(UpdateCommandStep.COMMAND_NAME)
            commandScope.addArgumentValue(UpdateCommandStep.CONTEXTS_ARG, "test2")
            commandScope.addArgumentValue(DbUrlConnectionArgumentsCommandStep.URL_ARG, h2.getConnectionUrl())
            commandScope.addArgumentValue(DbUrlConnectionArgumentsCommandStep.USERNAME_ARG, h2.getUsername())
            commandScope.addArgumentValue(DbUrlConnectionArgumentsCommandStep.PASSWORD_ARG, h2.getPassword())
            commandScope.addArgumentValue(UpdateCommandStep.CHANGELOG_FILE_ARG, <YOUR_CHANGELOG_FILE>)
            commandScope.setOutput(outputStream)
            commandResults = commandScope.execute()
        } as Scope.ScopedRunner)
        then:
        bufferLog.getLogAsString(Level.SEVERE).contains(<LOG_MESSAGE_TO_CHECK>);
```

jasonlyle88:
@MalloD12 I attempted your test as can be seen in the `liquibase-integration-tests/src/test/groovy/liquibase/command/update/UpdateIntegrationTest.groovy` test file. I am getting the following error:

```
The file src/test/resources/changelogs/h2/update/output.xml was not found in the configured search path:
More locations can be added with the 'searchPath' parameter.
```

However, I'm unsure how I can set the `serachPath` parameter given how this is being executed. Any help would be greatly appreciated!

MalloD12:
> @MalloD12 I attempted your test as can be seen in the `liquibase-integration-tests/src/test/groovy/liquibase/command/update/UpdateIntegrationTest.groovy` test file. I am getting the following error:
> 
> ```
> The file src/test/resources/changelogs/h2/update/output.xml was not found in the configured search path:
> More locations can be added with the 'searchPath' parameter.
> ```
> 
> However, I'm unsure how I can set the `serachPath` parameter given how this is being executed. Any help would be greatly appreciated!
> 
> Also, I think I'm on the right path going on what you said, but I'm not positive. If im not, please let me know!

Hey @jasonlyle88,

See the below example of what I do in some tests when I want to use a real changelog file to run an `update` command:

``` java
 given:
        def changelogFileName = "target/test-classes/changelogs/update.changelog.sql"
        def resourceAccessor = new SearchPathResourceAccessor(".,target/test-classes")
        def scopeSettings = [
                (Scope.Attr.resourceAccessor.name()) : resourceAccessor
        ]
        Scope.child(scopeSettings, {
            CommandUtil.runUpdate(db, changelogFileName, "generateChangelogWithEmptyTable", null, null)
        } as Scope.ScopedRunner)
```

As you can see, it needs to define the location where your files will be searched and then use the initialized resource accessor in the same scope that the update command will run. 

Can you try that and let me know if that works for you? If you have any other questions, please let me know.

Thanks,
Daniel.

jasonlyle88:
Hey @MalloD12,

Thank you for the help there! I was able to put that together with other information I've found in previous attempts to capture the changelog execution to get SO close. But I can't quite seem to get the last part.

If you take a look at the integration test, I've got a single changelog with a single changeset that has a single change of an output. I've gotten the integration test to the point where it captures the following in a string:

```
Running Changeset: src/test/resources/changelogs/h2/update/output.xml::test_output_changeset::jlyle
Liquibase: Update has been successful. Rows affected: 1
```

However, this does not include the output from the `output` change. I've confirmed the output from the `output` change is showing up in the console itself. I've also tried all of the different targets for the `output` change type. Yet I still can't seem to figure out a way to capture this output so I can check it.

I also tired using the `outputFile` parameter of the `CommandUtil.runUpdate` method, but that only has the update summary at the end of the run.

I apologize, but I need some more help in trying to figure out how to capture the output here so I can do this test properly.

Thanks in advance!

MalloD12 (approved):
**Approved**.

Code change looks ok for me, integration test has been fixed and all test are now successfully executed.

Thank you, @jasonlyle88!

Daniel.

jasonlyle88:
I updated the integration test to ensure the output only occurs one and only one time (as before it would pass if it output twice)!

MalloD12:
@jasonlyle88 cool, that makes sense to me. 

Thanks,
Daniel.

filipelautert (approved):

rberezen:
@jasonlyle88 Could I kindly ask you to resolve the merge conflict?
cc @filipelautert @MalloD12

MalloD12:
@rberezen merge conflict resolved.

jasonlyle88:
Just getting a chance to take a look at this, thanks for resolving that @MalloD12!

rberezen (approved):

filipelautert (approved):

filipelautert:
Seems that this PR broke some tests in liquibase-pro. Reverting the changes to Changeset class as in https://github.com/liquibase/liquibase/pull/6528 fixes it, but then the new tests added for this fix start to fail...

MalloD12:
Ok, let's revert these changes, and I can work with Jason later on making it work.

PR #6359, opened 2024-09-25
Handle blank change set ID DAT-18687

wwillard7800:
## Impact


liquibase-standard/src/main/java/liquibase/parser/AbstractFormattedChangeLogParser.java
```diff
@@ -717,6 +719,19 @@ protected boolean handleAdditionalLines(DatabaseChangeLog changeLog, ResourceAcc
         return false;
     }
 
+    //
+    // If the change set ID is empty after removing colons and blank spaces then it is invalid
+    //
```
abrackx, line 724: Can you make this an actual javadoc not a comment?

abrackx (approved):

liquibase-standard/src/main/java/liquibase/parser/AbstractFormattedChangeLogParser.java
wwillard7800: Yes

suryaaki2 (approved):

liquibase-standard/src/main/java/liquibase/parser/AbstractFormattedChangeLogParser.java
```diff
@@ -717,6 +719,19 @@
...
+    private void validateChangeSetId(String changeSetId, String line, int count) throws ChangeLogParseException {
+        String parsed = changeSetId.replace(":","").replace(" ","");
+        if (StringUtil.isEmpty(parsed)) {
```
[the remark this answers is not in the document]
filipelautert: Instead of replacing " " we could use `StringUtils.isBlank()` that would handle it for us.

filipelautert (changes requested):

PR #6378, opened 2024-10-01
GenerateChangelogCommandStep must only report having written a Changelog file if it did so

### LINKED ISSUE #6340: GenerateChangelogCommandStep always reports having written a Changelog file
### Search first

### Description

If you have an empty table or your includeObjects/excludeObjects result in no matches - because of a typo, there is no result set to generate a changelog file.
This is both true for SQL ans XML changelog generation.
Haven't tried a JSON variant.

### Steps To Reproduce

Perform a `generateChangeLog` on an empty table.

In the console, you will see

```
Generated changelog written to mytable.oracle.sql
```

But there is no such file.

This is not Oracle-specific.

### Expected/Desired Behavior

If no changelog file was produced, do not state it was generated.

### Liquibase Version

4.29.2

### Database Vendor & Version

_No response_

### Liquibase Integration

_No response_

### Liquibase Extensions

_No response_

_No response_

### Additional Context

_No response_

catull:
## Impact

In all cases, `generate-changelog` reports having written a changelog file.
The check happens for SQL and XML data exports.

However, in some cases, no such file is written.
The check only validates if a parameter was passed with `--changeLogFile=SOME_LOG_FILE_PATH`.

This PR corrects the behaviour.

If a change log file does exist, the message is reported as before.
If the change log file does not exist, it is now reported that no such file was written.

The result is preciser, depending on existence of the log file.

Fixes #6340

`TypeEnhancement`    - New features
`TypeTest`           - New Test features
`TypeBug`            - bug fixes


liquibase-standard/src/main/java/liquibase/command/core/GenerateChangelogCommandStep.java
```diff
@@ -158,9 +158,12 @@ public void run(CommandResultsBuilder resultsBuilder) throws Exception {
                     changeLogWriter.print(outputStream);
                 }
             }
-            if (StringUtil.trimToNull(changeLogFile) != null) {
+            if (!changeLogWriter.generateChangeSets().isEmpty()) {
                 Scope.getCurrentScope().getUI().sendMessage("Generated changelog written to " + changeLogFile);
             }
+            else {
+                Scope.getCurrentScope().getUI().sendMessage(String.format("Changelog not generated. There are not changesets to write to %s changelog", changeLogFile));
```
MalloD12, line 165: There is a small typo here. Message should be:
```
"Changelog not generated. There are no changesets to write to %s changelog"
```

liquibase-integration-tests/src/test/groovy/liquibase/command/core/GenerateChangeLogEmptyIntegrationTest.groovy
```diff
@@ -0,0 +1,99 @@
...
+import liquibase.util.FileUtil
+
+import spock.lang.Shared
+import spock.lang.Specification
+
+import java.nio.file.Files
+import java.nio.file.Path
+
+@LiquibaseIntegrationTest
+class GenerateChangeLogEmptyIntegrationTest extends Specification {
```
MalloD12, line 21: For tests I do have two comments:

1. Let's add a new method for calling `generate-changelog` by specifying `diffTypes` and `includeSchema` args in the `CommandUtil` test class:

```groovy
 private void runGenerateChangelog(DatabaseTestSystem db, String outputFile, String diffTypes) throws CommandExecutionException {
        CommandScope commandScope = new CommandScope(GenerateChangelogCommandStep.COMMAND_NAME)
        commandScope.addArgumentValue(DbUrlConnectionArgumentsCommandStep.URL_ARG, db.getConnectionUrl())
        commandScope.addArgumentValue(DbUrlConnectionArgumentsCommandStep.USERNAME_ARG, db.getUsername())
        commandScope.addArgumentValue(DbUrlConnectionArgumentsCommandStep.PASSWORD_ARG, db.getPassword())
        commandScope.addArgumentValue(GenerateChangelogCommandStep.OVERWRITE_OUTPUT_FILE_ARG, true)
        commandScope.addArgumentValue(GenerateChangelogCommandStep.CHANGELOG_FILE_ARG, outputFile)
        commandScope.addArgumentValue(PreCompareCommandStep.DIFF_TYPES_ARG, diffTypes)
        commandScope.addArgumentValue(DiffOutputControlCommandStep.INCLUDE_SCHEMA_ARG, true)
        OutputStream outputStream = new ByteArrayOutputStream()
        commandScope.setOutput(outputStream)
        commandScope.execute()
    }
```
2.  `CommandUtil.runDropAll(db)` from `cleanup` section is not needed because it's already being called from `LiquibaseIntegrationMethodInterceptor`.

liquibase-standard/src/main/java/liquibase/command/core/GenerateChangelogCommandStep.java
catull: Well spotted, corrected now, will be included in next commit.

catull: Implemented and pushed.

catull (commented):
Hello @MalloD12 

You missed the other two tables:

```
diff --git a/liquibase-integration-tests/src/test/groovy/liquibase/command/core/GenerateChangeLogEmptyIntegrationTest.groovy b/liquibase-integration-tests/src/test/groovy/liquibase/command/core/GenerateChangeLogEmptyIntegrationTest.groovy
index ef21971e8..4a6815f24 100644
--- a/liquibase-integration-tests/src/test/groovy/liquibase/command/core/GenerateChangeLogEmptyIntegrationTest.groovy
+++ b/liquibase-integration-tests/src/test/groovy/liquibase/command/core/GenerateChangeLogEmptyIntegrationTest.groovy
@@ -41,7 +41,7 @@ class GenerateChangeLogEmptyIntegrationTest extends Specification {
 create table "TEST_CHANGELOG_GENERATED" (
   b VARCHAR(10)
 );
-INSERT INTO "TEST" (b) VALUES ('Geronimo!');
+INSERT INTO "TEST_CHANGELOG_GENERATED" (b) VALUES ('Geronimo!');
 COMMIT;
 """)

@@ -54,7 +54,7 @@ COMMIT;
         Files.exists(Path.of(outputFileName))
         def contents = FileUtil.getContents(outputFile)
         contents.contains("""
-INSERT INTO "public"."TEST" ("b") VALUES ('Geronimo!');
+INSERT INTO "public"."TEST_CHANGELOG_GENERATED" ("b") VALUES ('Geronimo!');
 """)

         cleanup:
```

MalloD12:
> Details

Yeap, my bad. Fixing it now!

Thanks,
Daniel

MalloD12:
Hey @catull,

As you can see, I have been working on this PR, trying to fix the tests. I could see there was a confusion regarding changelog generation between an empty table or a table with some data, in any of these cases, changelog should be generated. Something else that I changed from the previous approach is that `generate-changelog` uses databachangelog metadata to recognize which changesets have been deployed in the given DB and need to be part of the generated changelog, that's why I created a changelog and deployed it by running an `update` command. 

After my test refactoring all these tests started passing all the time either by running them from my IDE or by executing them through `mvn` command, but for some reason they fail when running alongside `PostgresThreadingTest. I have created an internal ticket for this, but in the meantime, I commented out the insert changeset and some assertion/test. I think once we can identify the reason for this failure, we can remove that test.

Thanks,
Daniel.

catull:
Could it be that the table is created outside of a changeset ?

I have no problem packing the `create table` and `insert into table` into a regular change set.
That way, there would be a regular databasechange / and lock table.

Bad idea ?

MalloD12:
> Could it be that the table is created outside of a changeset ?
> 
> I have no problem packing the `create table` and `insert into table` into a regular change set. That way, there would be a regular databasechange / and lock table.
> 
> Bad idea ?

I'm not sure what's causing this. I spent some time investigating this, but can't find the reason yet. As I mentioned before I created an internal ticket to see if someone from our team can look at it, but also created the same issue (see #6396) in the community project in case you or someone else is keen to investigate this. 

It's not a recommended practice to pack multiple changes in a single changeset. Liquibase recommends having a 1:1 relationship between them. 

With that said, I think PR is in a good state for moving forward since code changes are ok and test coverage is ok. I left the other test so we can reproduce this issue and see if there is an underlying issue we need to fix in our code base that's causing this failure.

Thanks,
Daniel.

catull:
My use case actually involves no LiquiBase meta tables (DATABASECHANGELOG and LOCK).

The source DB is Oracle 19c, with some 80+ tables in a particular schema.
We are interested in the XML changeset as export file.

Why ?

Well, the target DB is MS SQLserver.
Post-processing SQL changeset INSERTs from Oracle format to SQLserver is no fun.

So we use LiquiBase on both DBs, let it do the right thing for us.

From the viewpoint of LiquiBase, I see that having proper meta table setup does matter.

It does not cover our use case entirely, but then I am not telling you how to manage your product.

I am very grateful that we can use LiquiBase, it saves us a lot of work.

catull:
There is a follow-up issues #6396 to investigate.

MalloD12 (approved):
**Approved**.

Code changes made looks good to me. I also pushed the test changes I made some days ago, as you can see one of the tests is commented because we found an issue with  not sure yet if it's related to `generate-changelog` execution or `LoadData` (see #6396 or [DAT-18821](https://datical.atlassian.net/browse/DAT-18821) for more information about it). Besides the mentioned issue I think this PR is ok to move forward with it.

Thanks,
Daniel.

catull:
Unfortunately, I cannot access https://datical.atlassian.net/browse/DAT-18821, but I believe you.

filipelautert (approved):

tati-qalified (approved):

MalloD12:
Hi @catull,

We have been discussing today about this change with some members of our team, and we think it would be nice to change the logic we added to only create a  changelog file when there are changes to write. Currently, we have implement this logic as:
```java
  if (!changeLogWriter.generateChangeSets().isEmpty()) {
                Scope.getCurrentScope().getUI().sendMessage("Generated changelog written to " + changeLogFile);
            }
  else {
                Scope.getCurrentScope().getUI().sendMessage(String.format("Changelog not generated. There are no changesets to write to %s changelog", changeLogFile));
          }
```

But what @filipelautert notice is that could have a negative impact on performance when there are many changes to write down because the `generateChangeSets()` is already being called in the `print()` method we are calling before this logic. 

I think something we can change this logic by using the `diffResult` object and do something like this:

```java
 if (!areThereChangeObjectsToWrite(diffResult)) {
                Scope.getCurrentScope().getUI().sendMessage(String.format("Changelog not generated. There are no changesets to write to %s changelog", changeLogFile));
            }
 else {
      Scope.getCurrentScope().getUI().sendMessage("Generated changelog written to " + changeLogFile);
  }
```

where:
```java
 private boolean areThereChangeObjectsToWrite(DiffResult diffResult) {
        return (diffResult.getMissingObjects().size() > 2 || !diffResult.getUnexpectedObjects().isEmpty() || !diffResult.getChangedObjects().isEmpty());
  }
```

That `getMissingObjects().size() > 2` is because diffResult is always returning catalog and schema as missing objects. How does it sound to you?

`NOTE`: Something I noticed yesterday is that with the last merges you made, some commits got lost, which made some of the tests added in this PR fail. Please have that in mind if you merge master on this PR again.

Thanks,
Daniel.

MalloD12:
Suggested changes applied. Re-reviewed.

filipelautert (approved):

rberezen (approved):

tati-qalified:
@MalloD12 @filipelautert This same change could be extended to the diff-changelog command, as it's currently not reporting whether the changelog file was written or not. 
Both classes use DiffResult and DiffToChangeLog objects, so maybe we could move the `areThereChangeObjectsToWrite` method to one of those so we don't have redundant code.
Thoughts?

PR #6386, opened 2024-10-02
specify liquibase integration when running through ant (DAT-18608)

StevenMassaro:
## Impact

### DIFF
--- a/liquibase-standard/pom.xml
+++ b/liquibase-standard/pom.xml
@@ -179,6 +179,12 @@
             <artifactId>commons-text</artifactId>
             <version>${commons-text.version}</version>
         </dependency>
+        <dependency>
+            <groupId>org.nanohttpd</groupId>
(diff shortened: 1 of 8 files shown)


filipelautert (approved):

liquibase-standard/pom.xml
```diff
@@ -179,6 +179,12 @@
             <artifactId>commons-text</artifactId>
             <version>${commons-text.version}</version>
         </dependency>
+        <dependency>
```
rberezen, line 182: @StevenMassaro just to check if any license update is required with adding this artifact?

filipelautert: @rberezen the license is a bit different: https://github.com/NanoHttpd/nanohttpd/blob/master/LICENSE.md
But as this is used for testing and we don't ship it then we don't need to update it anywhere.

filipelautert (approved):

PR #6397, opened 2024-10-06
fix: top to bottom (globals first) changelog parameters search

boolivar:
## Impact

## Description

This PR fixes regression introduced in #4267 and #2656 making it possible to inherit local properties from included changelogs while keeping "first wins" behavior for property definitions. Property search starts from global properties and continues for local properties from root changelog down to the current.

## Things to be aware of

It makes sense to check `globalProperties` first and then traverse changesets collecting local properties to list but I decided to collect local and global properties in single list and iterate in reverse order for simplicity.

## Things to worry about

Although it is considered as non-breaking change some expectations might be violated:
starting from v4.23.0 (#4267) local parameters behavior was changed to "last wins" making it possible to "inherit" properties in included changesets by assigning the same `logicalFilePath` for changesets.

**This is no longer possible** since properties in same `logicalFilePath` can't be changed once been set.

In order to inherit local properties they should be defined in distinct changesets as shown in the test changes.

cc @airon-assustadus

## Additional Context

Some details provided in https://github.com/liquibase/liquibase/pull/4267#issuecomment-2380604946

### DIFF
--- a/liquibase-standard/src/main/java/liquibase/changelog/ChangeLogParameters.java
+++ b/liquibase-standard/src/main/java/liquibase/changelog/ChangeLogParameters.java
@@ -14,6 +14,8 @@ import liquibase.structure.core.Schema;
 import liquibase.structure.core.Sequence;
 import liquibase.util.StringUtil;
 import lombok.Getter;
+import org.apache.commons.collections4.iterators.ReverseListIterator;
+import org.apache.commons.lang3.stream.Streams;
 
 import java.util.*;
 import java.util.stream.Collectors;
@@ -29,7 +31,7 @@ import java.util.stream.Collectors;
  * <p>
  * Properties can be defined as "global" or "local". Global properties span all change logs.
  * A global setting configured in an included changelog is still available to all changesets
- * Local properties are  only available in the change log that they are defined in -- not even in changelogs "included" by the file that defines the property.
+ * Local properties are available in the change log that they are defined AND in included changelogs.
  */
 public class ChangeLogParameters {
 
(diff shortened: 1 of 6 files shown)


tati-qalified (approved):

boolivar:
PR adjusted according to the latest [ChangeLogParameters](https://github.com/liquibase/liquibase/pull/6407) changes

martinsefcik:
Any chance to have this PR merged and released? 

Due to this regression introduced in #2656 we had to stay on v4.11.0 in our project.

MalloD12:
Hey @boolivar,

I've been looking at this PR, and overall changes made look good to me, but discussing with the team, these changes could also be a breaking change. We think that at this point, a safer approach could be to add a global configuration to control whether we want to use a reverse list or a regular one, to then put in practice the "last win" or the "first win" approach.

Could it be possible for you to introduce that change? 

Thanks,
Daniel.

MalloD12:
Hi @boolivar,

We are ok with the changes you are trying to introduce with this PR, but the idea of adding a configuration flag to control the ways parameters are found is mainly trying to avoid breaking someone else's code (that might fork the liquibase code and make use of existing logic based on their needs).

Updating docs is something we can surely do pretty quickly, but any code change we think is appropriate, we would need our QA team to spend some time testing it. 

Thanks,
Daniel.

boolivar:
@MalloD12 please take a look at #7032 - this PR doesn't introduce any breaking changes making hierarchical search for local properties with "last-wins" behavior.
Thank you!

boolivar:
Closed in favor of #7032

PR #6407, opened 2024-10-08
 Add filterable flag to ChangeLogParameter class (DAT-18811)

abrackx:
## Impact

## Description
Add new `systemParameters` field to `ChangeLogParameters`. Respect new `filterable` flag when filtering changelog parameters. System parameters such as environment variables do not get filtered as there is no way to configure a label or context for these types of parameters. This enables users to use label filters such as `@some-filter` and still have values replaced from global changelog parameters like environment variables. 

## Things to be aware of
This pr treats every "system" parameter like environment variables as "not filterable". This is respected ONLY when these "system" parameters are set up without any filterable attributes (dbms, label, context).


liquibase-standard/src/test/groovy/liquibase/changelog/ChangeLogParametersTest.groovy
```diff
@@ -145,4 +145,17 @@ class ChangeLogParametersTest extends Specification {
         then:
         expanded == "1234"
     }
+
+    def "System properties set up naively are not filtered when using required labels or contexts"() {
+        when:
+        def database = new MockDatabase()
+
+        System.setProperty("SOME_PROPERTY", "This should not be filtered")
```
StevenMassaro, line 153: You should unset this System property in a cleanup block in this test.

liquibase-standard/src/main/java/liquibase/changelog/ChangeLogParameters.java
StevenMassaro, line 1: Could you add some javadoc around the changes you made to this class? It's already a bit confusing.

StevenMassaro (changes requested):

liquibase-standard/src/main/java/liquibase/changelog/ChangeLogParameters.java
abrackx: Done!

liquibase-standard/src/test/groovy/liquibase/changelog/ChangeLogParametersTest.groovy
abrackx: Done!

liquibase-standard/src/main/java/liquibase/changelog/ChangeLogParameters.java
abrackx: Updated javadocs and added comments!

filipelautert (approved):

Issue #6396, opened 2024-10-04
RuntimeException thrown when 'generate-changelog' command tries to process a insert change for SQL changelog format


MalloD12:
### Search first

- [x] I searched and no similar issues were found

### Description

While running either `PostgresThreadingTest` and `GenerateChangeLogEmptyIntegrationTest` together from `MVN` or the `IDE` it ended up making the below test:

```groovy
def "Should generate changelog file with non-empty table"() {
        given:
        def changelogFileName = "target/test-classes/changelogs/update.changelog.sql"
        def resourceAccessor = new SearchPathResourceAccessor(".,target/test-classes")
        def scopeSettings = [
                (Scope.Attr.resourceAccessor.name()) : resourceAccessor
        ]
        Scope.child(scopeSettings, {
            CommandUtil.runUpdate(db, changelogFileName, "both", null, null)
        } as Scope.ScopedRunner)

        when:
        def outputFileName = 'test/test-classes/output2.postgresql.sql'
        CommandUtil.runGenerateChangelog(db, outputFileName, "tables, data")
        def outputFile = new File(outputFileName)
        def fileContent = FileUtil.getContents(outputFile)

        then:
        fileContent.containsIgnoreCase("CREATE TABLE \"generate_changelog_test_sql\"")
        fileContent.containsIgnoreCase("INSERT INTO \"generate_changelog_test_sql\"")

        cleanup:
        outputFile.delete()
    }
```
`update.changelog.sql`:
```sql
--liquibase formatted sql

--changeset mallod:1 labels:generateChangelogWithEmptyTable,both
CREATE TABLE public.generate_changelog_test_sql (id int not null primary key, name varchar(255));

--changeset mallod:2 labels:both
INSERT INTO public.generate_changelog_test_sql (id, name) VALUES (1, 'Geronimo!');
```

to fail with the below exception:

```java
[ERROR] liquibase.command.core.GenerateChangeLogEmptyIntegrationTest.Should generate changelog file with non-empty table -- Time elapsed: 0.486 s <<< ERROR!
liquibase.exception.CommandExecutionException: java.lang.RuntimeException: java.lang.RuntimeException: liquibase.exception.UnexpectedLiquibaseException: Unable to read file private/var/folders/cm/1krjp23j0j9054m7xfn_gwtc0000gn/T/liquibase-data/generate_changelog_test_sql.csv
	at liquibase.command.CommandScope.execute(CommandScope.java:258)
[...]

PR #6138, opened 2024-07-26
Add support for include columns in indexes

stephenatwell:
## Impact

- [ ] Bug fix (non-breaking change which fixes expected existing functionality)
- [ X ] Enhancement/New feature (adds functionality without impacting existing logic)
- [ ] Breaking change (fix or feature that would cause existing functionality to change)
 

## Description

Added support for specifying columns to be used in the 'Include' clause of indexes, and made generate changelog generates changes that include indexes when run against a database that uses these indexes. Tested against MSSQL.

Fixes #1697 

Here is an example changeset creating two indexes, one with and one without include:
- changeSet:
    id: 1722028836430-8
    author: stephenatwell (generated)
    changes:
    - createIndex:
        columns:
        - column:
            descending: true
            name: id
        - column:
            included: true
            name: name
        indexName: saTestInclude
        tableName: person

## Things to be aware of
column object now has a new boolean that states whether or not the column should go in the 'included' clause.

PR #6422, opened 2024-10-14
MissingDataChangeGenerator: small optimisation, skip null-value columns (issues #6420)

### LINKED ISSUE #6420: MissingDataChangeGenerator: small optimisation, skip null-value columns in export
### Search first

### Description

I noticed that NULL-value colums are exported as such.

Some of the tables have anywhere from 1/3 to more than 1/2 of the columns NULL-ed.

Perhaps the `generate-changelog` shiuld be extended to skip those columns.

The default should remain to preserve NULL-value columns.

### Steps To Reproduce

Create a table with 4 varchar(4):

```sql
create table "TEST" (
  a varchar (4),
  b varchar (4),
  c varchar (4),
  d varchar (4)
);

-- all 4 inserts are equivalent
insert into "TEST" (a)          values ("AAAA");
insert into "TEST" (a, b)       values ("AAAA", null);
insert into "TEST" (a, b, c)    values ("AAAA", null, null);
insert into "TEST" (a, b, c, d) values ("AAAA", null, null);
```

### Expected/Desired Behavior

A simpler export file, both for SQL or XML exports, skipping NULL-value columns.

E.g. I'd like an SQL import to look like this from the above insertions:

```sql
insert into "TEST" (a) values ("AAAA");
insert into "TEST" (a) values ("AAAA");
insert into "TEST" (a) values ("AAAA");
insert into "TEST" (a) values ("AAAA");
```

### Liquibase Version

_No response_

### Database Vendor & Version

_No response_

### Liquibase Integration

_No response_

### Liquibase Extensions

_No response_

_No response_

### Additional Context

_No response_

catull:
Export table data preserving NULL-values, or not.

## Impact

## Description

Have you ever wondered if changelog exports could skip all NULL-values ?
This MR does exactly this.

It implements issue #6420.


catull:
@MalloD12 Implemented the logic for SQL, JSON, YAML and XML serialisation.

MalloD12:
Hey @catull, 

Could you please review the changes made on the serializer, I think they are not yet fixing these cases. For example what has been added in `XMLChangeLogSerializer`:

```java
if (value instanceof ColumnConfig) {
      ColumnConfig columnConfig = (ColumnConfig) value;
      if (!preserveNullValues && columnConfig.isNull()) {
          return;
      }
      node.appendChild(createNode((LiquibaseSerializable) value));
}
```

is broken the below assertion from `serialize_pretty_nestedNodeWithAttributesAndNotExistsIgnored()` (XMLChangeLogSerializerTest)

```java
assertEquals("<createTable catalogName=\"a\"\n" +
                "        schemaName=\"b\"\n" +
                "        tableName=\"c\">\n" +
                "    <column defaultValue=\"x1\" name=\"x\"/>\n" +
                "    <column defaultValue=\"y1\" name=\"y\"/>\n" +
                "</createTable>", out);
```

Because we are not including `defaultValue` in the node as it was behaving previously. Additionally, it would be worth checking other serializers to see if the same issue is happening. At the moment, we don't have any reported issues for those serializers related to this change, but that's because we don't have specific tests for them.

Thanks,
Daniel.

MalloD12:
Moving this PR to `In-Discussion` column until serializer tests are fixed.

Thanks,
Daniel

catull:
Good thing we have test cases for XML serialiser with default values.
I was not considering them in `ColumnConfig#isNull`.

The test `XMLChangeLogSerializerTest#serialize_pretty_nestedNodeWithAttributesAndNotExistsIgnored` works again.

MalloD12:
Hi @catull,

I have a few thing I have observed on the current state of this PR: 

1. I think temporarily we should ignore `GenerateSqlChangeLogPostgresIntegrationTest` tests (we can do this by adding `@Ignore` annotation to each test), because when generating a SQL format changelog file that contains a `Insert` change it results in a runtime exception as described in #6396, so until we fix that case I think we cannot perform a scenario like this one.  
2. `GenerateXmlChangeLogPostgresIntegrationTest` seems to be still not working. For the case of `"Should generate XML changelog excl. NULL-values"` we are expecting to received something like:
```xml
  <insert schemaName="public" tableName="preservation_test">
            <column name="a" value="AA"/>
  </insert>
```
but we are still receiving:
```xml
  <insert schemaName="public" tableName="preservation_test">
          <column name="a" value="AA"/>
          <column name="b" "/>
          <column name="c" />
 </insert>
```
this case corresponds to the case of having `preserve_null = false`, but for the case of  `preserve_null = true` we are also generatingthe same above output:

```xml
          <column name="b" "/>
```
Where are not displaying the null values, I was thinking we were intending to be displaying that. Was that your understanding as well? I'll check with our product team to see what would they be expecting in this case. This same thing is happening on YAML and JSON formats.

On the other hand, regarding serializers test classes, yeap it would be good to add some coverage for the other format types.

Thanks,
Daniel.

MalloD12:
Hey @catull,

I checked earlier with our product team and they said they would also be expecting something like:

```xml
 <column name="b" value=null/> 
``` 
 or potentially:

```xml
 <column name="b" value="null"/>
```

for the case `preserve_null = true`.

Thanks,
Daniel.

catull:
> Hey @catull,
> 
> I checked earlier with our product team and they said they would also be expecting something like:
> 
> ```
>  <column name="b" value=null/> 
> ```
> 
> or potentially:
> 
> ```
>  <column name="b" value="null"/>
> ```
> 
> for the case `preserve_null = true`.
> 
> Thanks, Daniel.

The first one is invalid in XML - unquoted value in a tag -, but tolerated in HTML 4 and below.
It is also not valid in XHTML, a.k.a HTML 5.

So, in order to distinguish Java `null` from a String with value `"null"`, this is what I suggest

Case Java `null`
```xml
  <column name="b" value="null" />
```

Case Java `"null"`

```xml
  <column name="b" value="'null'" />
```

That should do it.

MalloD12:
I'm talking about the case  `<column name="b" value="null" />` or any other option that can preserve the null value in the generated output file for XML, YAML, and any other changelog format type when a column has no value set for it.

Thanks,
Daniel.

catull:
> > In my opinion `<column name="b" value="null" />` looks nicer.
> > Thanks, Daniel.
> 
> For which case ? I corrected my suggestion.
> 
> Are you referring to the case of a SQL NULL value or a Java String with the content "null". Because in Liquibase, there is no distinction, see issue #6506.

Although this is an unrelated change - it should be made all on its own -, I went ahead and made a change.
See 1e4601f1a021448542f171ba79d3172c55746aad

catull:
Dear @filipelautert

I made the suggested changes related to explicitly export `null` values in XML format.
Can you review the changes, please ?

catull:
> Hi @catull,
> 
> I have a few thing I have observed on the current state of this PR:
> 
>     1. I think temporarily we should ignore `GenerateSqlChangeLogPostgresIntegrationTest` tests (we can do this by adding `@Ignore` annotation to each test), because when generating a SQL format changelog file that contains a `Insert` change it results in a runtime exception as described in [RuntimeException thrown when 'generate-changelog' command tries to process a insert change for SQL changelog format #6396](https://github.com/liquibase/liquibase/issues/6396), so until we fix that case I think we cannot perform a scenario like this one.
> 
>     2. `GenerateXmlChangeLogPostgresIntegrationTest` seems to be still not working. For the case of `"Should generate XML changelog excl. NULL-values"` we are expecting to received something like:
> 
> 
> ```
>   <insert schemaName="public" tableName="preservation_test">
>             <column name="a" value="AA"/>
>   </insert>
> ```
> 
> but we are still receiving:
> 
> ```
>   <insert schemaName="public" tableName="preservation_test">
>           <column name="a" value="AA"/>
>           <column name="b" "/>
>           <column name="c" />
>  </insert>
> ```
> 
> this case corresponds to the case of having `preserve_null = false`, but for the case of `preserve_null = true` we are also generatingthe same above output:
> 
> ```
>           <column name="b" "/>
> ```
> 
> Where are not displaying the null values, I was thinking we were intending to be displaying that. Was that your understanding as well? I'll check with our product team to see what would they be expecting in this case. This same thing is happening on YAML and JSON formats.
> 
> On the other hand, regarding serializers test classes, yeap it would be good to add some coverage for the other format types.
> 
> Thanks, Daniel.

@MalloD12 

Since #6396 was closed as it could not be reproduced, I hope this very issue moves on now.

MalloD12:
Hi @catull,

I triggered the tests for this PR and am still seeing some failures as you can see in the below screenshot:

Screenshot 2024-12-30 at 3 16 31 PM

I would recommend you not to keep updating this branch with master, at least not until these tests are fixed, because otherwise if someone else from the team doesn't authorize the tests to be executed you might not be able to see the execution details that can help you to fix the failing ones. 

Thanks,
Daniel.

catull:
> Hi @catull,
> 
> I triggered the tests for this PR and am still seeing some failures as you can see in the below screenshot:
> Screenshot 2024-12-30 at 3 16 31 PM
> 
> I would recommend you not to keep updating this branch with master, at least not until these tests are fixed, because otherwise if someone else from the team doesn't authorize the tests to be executed you might not be able to see the execution details that can help you to fix the failing ones.
> 
> Thanks, Daniel.

Dear @MalloD12

Happy New Year.

It turns out, the logic that establishes which NULL-values have to be considered for output was flawed.
It is fixed now.

Both suites `XMLChangeLogSerializerTest` and `YamlChangeLogSerializerTest` complete without failures.

So does `mvn -DtrimStackTrace=false -P 'skip-integration-tests' -pl '!liquibase-cdi-jakarta' clean test package surefire-report:report`.

MalloD12:
Happy New Year, @catull!

Nice, I'll have a look at it. 

Thanks,
Daniel.

MalloD12:
Hi @catull,

I'm seeing tests are not still in a good state. For example, if you run `GenerateYamlChangeLogPostgresIntegrationTest` you will see:

- "Should generate YAML changelog incl. NULL-values" is passing but this is not accurate:

The reason of it is `preserveNullValues=true` (L31), but as shown in the below test assertion it is not validating in this case columns `b` and `c` have a `null` value set for them, so in this case `null` value is not being preserved.

```yaml
changes:
    - insert:
        columns:
        - column:
            name: a
            value: AA
        - column:
            name: b
        - column:
            name: c
```

- Regarding this other test "Should generate YAML changelog excl. NULL-values" there is undesired content here. The below content with a lot of `null` values set for different ColumnConfig attributes is displayed for each column (`b` and `c`). In this case, we can see `value: null` is displayed among others, but in this case we are setting `preserveNullValues=false` (see L68). This test is failing because the test assertion specified is not expecting to see this. Below an example of it:

```yaml
- !!liquibase.change.ColumnConfig
          startWith: null
          defaultValueConstraintName: null
          constraints: null
          valueDate: null
          valueNumeric: null
          incrementBy: null
          descending: null
          valueBoolean: null
          defaultValueSequenceNext: null
          value: null
          type: null
          defaultValueNumeric: null
          valueBlobFile: null
          defaultValue: null
          autoIncrement: null
          valueClobFile: null
          generationType: null
          defaultValueBoolean: null
          name: b
          encoding: null
          included: null
          defaultOnNull: null
          computed: null
          valueSequenceCurrent: null
          valueSequenceNext: null
          remarks: null
          valueComputed: null
          defaultValueDate: null
          defaultValueComputed: null
```

- Same is happening for `GenerateJsonChangeLogPostgresIntegrationTest`.

- Last thing, I'm facing an additional error here, but I don't think this is related to your PR. I fixed something very similar but tried the same fix, and it didn't work. I'll ask someone on the team who's coming back from holidays on Monday. 

```java
 GenerateChangeLogEmptyIntegrationTest.Should generate changelog file with non-empty table:63 » CommandExecution java.lang.RuntimeException: java.lang.RuntimeException: liquibase.exception.UnexpectedLiquibaseException: Unable to read file private/var/folders/cm/1krjp23j0j9054m7xfn_gwtc0000gn/T/liquibase-data/generate_changelog_test_sql.csv
```

Hope you have a great weekend,
Daniel.

catull:
@MalloD12 
Fixed the 2 cases in `GenerateYamlChangeLogPostgresIntegrationTest` as well as the third case reported against `GenerateJsonChangeLogPostgresIntegrationTest`.

Have a pleasant first week-end of the year.

catull:
@MalloD12 I do not see a failure/exception with `GenerateChangeLogEmptyIntegrationTest`.

MalloD12:
We have resolved the error  of the type:

```java
CommandExecution java.lang.RuntimeException: java.lang.RuntimeException: liquibase.exception.UnexpectedLiquibaseException: Unable to read file private/var/folders/cm/1krjp23j0j9054m7xfn_gwtc0000gn/T/liquibase-data/generate_changelog_test_sql.csv
```

We were facing it, and it should be fine now.  The only thing that we still need to update are the below tests:

- "Should generate SQL changelog excl. NULL-values" (GenerateSqlChangeLogPostgresIntegrationTest)
- "Should generate XML changelog incl. NULL-values" (GenerateXmlChangeLogPostgresIntegrationTest) 

I added the expected result in a comment for each of these tests. With those pending test correction I think we should be fine.

Thanks,
Daniel.

catull:
> We have resolved the error of the type:
> 
> ```java
> CommandExecution java.lang.RuntimeException: java.lang.RuntimeException: liquibase.exception.UnexpectedLiquibaseException: Unable to read file private/var/folders/cm/1krjp23j0j9054m7xfn_gwtc0000gn/T/liquibase-data/generate_changelog_test_sql.csv
> ```
> 
> We were facing it, and it should be fine now. The only thing that we still need to update are the below tests:
> 
>     * "Should generate SQL changelog excl. NULL-values" (GenerateSqlChangeLogPostgresIntegrationTest)
> 
>     * "Should generate XML changelog incl. NULL-values" (GenerateXmlChangeLogPostgresIntegrationTest)
> 
> 
> I added the expected result in a comment for each of these tests. With those pending test correction I think we should be fine.
> 
> Thanks, Daniel.

Thank you, @MalloD12 

I changed the code to correct the tests you mention, thanks for making it easier by introducing the "expected" output as comments.

Now that `GenerateXmlChangeLogPostgresIntegrationTest#"Should generate XML changelog incl. NULL-values"` produces

```xml
        <insert schemaName="public" tableName="preservation_test">
            <column name="a" value="AA"/>
            <column name="b" value="null"/>
            <column name="c" value="null"/>
        </insert>
```
, how is the import going to treat the value of columns `b` and `c` ?

Is it going to try variant A) or B):

```sql
-- A)
INSERT INTO "PRESERVATION_TEST" (a, b, c) VALUES ('AA', null, null);
-- B)
INSERT INTO "PRESERVATION_TEST" (a, b, c) VALUES ('AA', 'null', 'null');
```
?

I fear it is going to try B).

MalloD12:
Hey @catull,

We have a problem here that I noticed yesterday and thought it was just an XMLSerializer issue, but it's not it is a generic issue that we have detected in this case because we have the below  test changeset in one of our XML tests that integration tests uses:

```xml
<changeSet id="indexIncludeTest" author="satwell">
    <createIndex indexName="idx_include_test" tableName="TestChange2">
        <column name="name"/>
        <column name="name2" included="true"/>
    </createIndex>
</changeSet>
```

that showed us this [test failure](https://github.com/liquibase/liquibase/actions/runs/12653322687/job/35258827489?pr=6422). The problem is that when `generate-changelog ` command generates the below changeset it causes this issue:

```xml
<changeSet author="dmall (generated)" id="1736310059802-90">
    <createIndex catalogName="lbcat" indexName="idx_include_test" schemaName="dbo" tableName="TestChange2">
        <column name="name" value="null"/>
        <column name="name2" value="null"/>
    </createIndex>
</changeSet>
```

And this seems to be related to this other [PR](https://github.com/liquibase/liquibase/pull/6138) that is `MSSQL` specific that we have recently merged because if I removed the `included="true"` attribute, the test works. We would need to update our serializers to make them recognize that when creating a node/property for a column it should only add the `value="null"` part if its parent is not a `createIndex`.
 

I hope I'm being clear enough, if I am not please let me know and we can probably discuss about it in discord to have a more fluent conversation. 

Thanks,
Daniel.

catull:
> Hey @catull,
> 
> We have a problem here that I noticed yesterday and thought it was just an XMLSerializer issue, but it's not it is a generic issue that we have detected in this case because we have the below test changeset in one of our XML tests that integration tests uses:
> 
> ```
> <changeSet id="indexIncludeTest" author="satwell">
>     <createIndex indexName="idx_include_test" tableName="TestChange2">
>         <column name="name"/>
>         <column name="name2" included="true"/>
>     </createIndex>
> </changeSet>
> ```
> 
> that showed us this [test failure](https://github.com/liquibase/liquibase/actions/runs/12653322687/job/35258827489?pr=6422). The problem is that when `generate-changelog ` command generates the below changeset it causes this issue:
> 
> ```
> <changeSet author="dmall (generated)" id="1736310059802-90">
>     <createIndex catalogName="lbcat" indexName="idx_include_test" schemaName="dbo" tableName="TestChange2">
>         <column name="name" value="null"/>
>         <column name="name2" value="null"/>
>     </createIndex>
> </changeSet>
> ```
> 
> And this seems to be related to this other [PR](https://github.com/liquibase/liquibase/pull/6138) that is `MSSQL` specific that we have recently merged because if I removed the `included="true"` attribute, the test works. We would need to update our serializers to make them recognize that when creating a node/property for a column it should only add the `value="null"` part if its parent is not a `createIndex`.
> 
> I hope I'm being clear enough, if I am not please let me know and we can probably discuss about it in discord to have a more fluent conversation.
> 
> Thanks, Daniel.

Hello @MalloD12. I am now on discord with the same handle.
You are correc, the error you describe is due to 'preserve-null-values' PR.

I think I know how to fix this.
I am preparing it in a different PR.

liquibase-standard/src/main/java/liquibase/serializer/core/xml/XMLChangeLogSerializer.java
```diff
@@ -185,8 +193,26 @@ public Element createNode(LiquibaseSerializable object) {
...
+        if (object instanceof ColumnConfig) {
+            ColumnConfig cc = (ColumnConfig) object;
+            if (cc.isNullValue() && !cc.hasDefaultValue() && !cc.hasSortOrder() && cc.getConstraints() == null) {
+                node.setAttribute("name", (String) object.getSerializableFieldValue("name"));
+
+                Object type = object.getSerializableFieldValue("type");
+
+                if (null != type) {
+                    node.setAttribute("type", type.toString());
+                } else {
```
MalloD12, line 205: Something else I just noticed is that I don't think this `else` is accurate as it is. I think it should be a separate `IF` block comparing if a value field (similar to what we are getting in L201) is `null`, if it's then we set `node.setAttribute("value", "null");`

catull: I will check it, and perhaps add more tests.

catull: Actually, in the latest change, I onlt have to check if the ColumnConfig only contains the column name alone.
The `type` case is handled by means of `ColumnConfig#isSerializable(boolean)`.

MalloD12:
Hey @catull,

I found that the latest changes on the serializers are not totally accurate. For example, in the case of the XML serializer, it's producing an incorrect  result: 

```xml
<createTable preserveNullValues="true" tableName="str4">
    <column name="col1" type="INT">
    </column>
    <column autoIncrement="true" name="col2" type="INT">
    </column>
    <column name="col3" type="INT"/>
</createTable>
```

This is what I saw in one of the failing tests, and that's wrong because we are not expecting to have a `preserveNullValues` property at a createTable level, so we need to correct that, unless we plan to have this attribute at a higher level and then make it applicable to its child (columns in this case), that could be a possibility but it will require some other additional changes.

liquibase-integration-tests/src/test/resources/liquibase/extension/testing/command/calculateChecksum.test.groovy
```diff
@@ -47,7 +47,7 @@ Optional Args:
         ]
 
         expectedResults = [
-                checksumResult   : "9:10de8cd690aed1d88d837cbe555d1684"
+                checksumResult   : "9:99091c432814de759be3e5f65eececf0"
```
MalloD12, line 50: I think these test data updates are not correct because this generation of a new Checksum value is because of the new property included in `createTable` change, which at the moment is not an expected property there. In my opinion, we should revert this. 

Thanks,
Daniel.

catull: OK, I reverted the changes that brought `preserverNullValues` to the surface.
Now that property is never serialised.  It does appear in the meta-data, like `serializedFields`.

MalloD12 (approved):
**Approved**.

Looking at the overall changes, I think this PR is ready to move on in the review process. Build and tests checks are now successfully executing (see [here](https://github.com/liquibase/liquibase/actions/runs/12737625847/job/35539292771?pr=6422) for more details).

Thank you, @catull for implementing this change and for adressing all our review comments.

Thanks,
Daniel.

catull:
> **Approved**.
> 
> Looking at the overall changes, I think this PR is ready to move on in the review process. Build and tests checks are now successfully executing (see [here](https://github.com/liquibase/liquibase/actions/runs/12737625847/job/35539292771?pr=6422) for more details).
> 
> Thank you, @catull for implementing this change and for adressing all our review comments.
> 
> Thanks, Daniel.

My pleasure, and thanks for your guidance.

filipelautert (approved):

MalloD12:
@AMBERMW13 - @adrian-velonis1: I have created this [ticket](https://datical.atlassian.net/browse/PD-5619) to document this change. 

Thanks,
Daniel.

tati-qalified:
Hi @catull, I've been testing this change manually through the CLI and I've found a few bugs:

- When generating an `xml` file, the attribute `"value"=null` is added to the `<changeSet>`, `<insert>` and `<column>` tags. For the columns, when they're of any non-string type, `diff-changelog` creates them as `valueNumeric`, `valueDate`, etc., so this means the columns end up with two different `value` attributes.
This is most likely coming from [`MissingDataChangeGenerator.java#137`](https://github.com/liquibase/liquibase/pull/6422/files#diff-8945400e2d671d98e4fdc8edf9a13cc349fa1a308a90a0207e5c3ce72453bf4dR137) and onwards, as the code isn't contemplating null values in different types.
When they **are** string types, the actual value is being replaced by `"value"=null`, so all non-null values are lost.
```
 <changeSet author="tfernandez (generated)" id="1737989442471-1" value="null">
        <insert tableName="test_for_nulls" value="null">
            <column name="string" value="null"/>
            <column name="number" value="null"/>
            <column name="date" value="null"/>
            <column name="boolean" value="null" valueBoolean="false"/>
        </insert>
        <insert tableName="test_for_nulls" value="null">
            <column name="string" value="null"/>
            <column name="number" value="null" valueNumeric="23"/>
            <column name="date" value="null" valueDate="2023-04-12T00:00:00"/>
            <column name="boolean" value="null"/>
        </insert>

         (...)
```
This happens with `preserveNullValues` set to both `true` and `false`.

- For `yml` format, null values are skipped successfully. However, when a row of data has all its values set to `null`, there are straight brackets in the `columns` section of the changeset:
```
 - insert:
        columns: []
        tableName: test_for_nulls
```
The same thing happens for `json`, and even though in this case the syntax is correct, it seems unnecessary to add this section
``` 
  {
     "insert": {
       "columns": [
         ],
       "tableName": "test_for_nulls"
      }
   }
```

- Lastly, for `sql` files, null values are not skipped.

Could you look into these issues?
Thank you!
Tatiana

catull:
@filipelautert @MalloD12 
Excuse me, why is this issue closed ?

tati-qalified:
@catull it looks like you closed it
image

catull:
Hmm. I am not aware I did so. Can I reopen it?

tati-qalified:
sure! I can't do it because there are no new commits, so you'll have to commit something and reopen it

filipelautert:
> Hmm. I am not aware I did so. Can I reopen it?

Hey @catull , seems you did not press the close button, but by the "force-push" under that I believe you may have pushed master  branch over your pr branch. As both branches became equal, github closed the PR. 

I think you just need to force-push your branch again then you will be able to reopen.

The same is valid for https://github.com/liquibase/liquibase/pull/6506 .

catull:
Thanks @filipelautert 

I was able to retrieve the branch and rebase/merge master into it.

filipelautert:
Hey @catull , did the comments in https://github.com/liquibase/liquibase/pull/6422#issuecomment-2616510330 were addressed?

catull:
SInce I had many incorrect commit author information.
I backed the original branch to https://github.com/catull/liquibase/tree/implement-issue-6420-orig.

Then I created a new branch with just the modified files vs `master`.

There is no difference in content between
     https://github.com/catull/liquibase/tree/implement-issue-6420
and
    https://github.com/catull/liquibase/tree/implement-issue-6420-orig
.
The length of the history is different, though.

filipelautert:
@MalloD12 here is comment that I mentioned: https://github.com/liquibase/liquibase/pull/6422#issuecomment-2616510330 . Could you and @catull follow up ? After that we are ready to merge.

catull:
@MalloD12 @tati-qalified 

I made 2 small changes, see https://github.com/liquibase/liquibase/pull/6422/commits/f37bc39a2b84f3c9b9232cd6bf5847eb28b531f4 and https://github.com/liquibase/liquibase/pull/6422/commits/ae52fb96820d04a0aac423fb12226b26af7928fd.

I then built a dist and installed it locally, to `/opt/liquibase.snap`.
LIQUIBASE_HOME is set to `/opt/liquibase.snap`.
`which liquibase` gives `/opt/liquibase.snap/liquibase`.

Then I launched an H2 database locally, version 2.4.240, with

```
java \
  -cp /opt/h2/h2-2.4.240/bin/h2-2.4.240.jar \
  org.h2.tools.Server \
  -web \
  -webPort 8082 \
  -webAllowOthers \
  -tcpPort 9092 \
  -tcp \
  -tcpAllowOthers \
  -tcpPassword 123 \
  -tcpDaemon \
  -baseDir /tmp/test/ \
  -ifNotExists
```

and with this liquibase.properties:

```
driver: org.h2.Driver
url: jdbc:h2:tcp://localhost:9092/test
username: sa
password:

schemas: public
author: dbuser

headless: true
strict: true
includeSchema: true

logLevel: warn

liquibase.showBanner: false
liquibase.fileEncoding: utf-8

liquibase.analytics.devOverride: false
liquibase.analytics.enabled: false
liquibase.analytics.logLevel: off

liquibase.command.includeObjects: table:test_for_nulls
liquibase.command.overwriteOutputFile: true
liquibase.command.diffTypes: data
```

I populated the DB with

```sql
create table PUBLIC."test_for_nulls" (
   "string"   varchar2(10),
   "number"   numeric(10),
   "date"     date,
   "boolean"  boolean
);

insert into PUBLIC."test_for_nulls"
values (
    'a string',
    7,
    current_date,
    true
);

insert into PUBLIC."test_for_nulls"
values (
    'a string',
    null,
    null,
    null
);

insert into PUBLIC."test_for_nulls"
values (
    null,
    7,
    null,
    null
);

insert into PUBLIC."test_for_nulls"
values (
    null,
    null,
    current_date,
    null
);

insert into PUBLIC."test_for_nulls"
values (
    null,
    null,
    null,
    false
);

insert into PUBLIC."test_for_nulls"
values ();
```
Finally, I generated all sorts of changelogs: SQL, XML and YAML.
I noticed that the XML schema does not like the all-null columns export:
image

But, it is perfectly legal to do this:

```sql
insert into PUBLIC."test_for_nulls"
values ();
```

If, and only if the table has no mandatory columns.
And in this case, we have such a case.

My export script:

```shell
liquibase generate-changelog \
  --changelog-file changelog.nulls.h2.sql \
  --preserve-null-values true

liquibase generate-changelog \
  --changelog-file changelog.nulls.h2.xml \
  --preserve-null-values true

liquibase generate-changelog \
  --changelog-file changelog.nulls.h2.yaml \
  --preserve-null-values true


liquibase generate-changelog \
  --changelog-file changelog.no-nulls.h2.sql \
  --preserve-null-values false

liquibase generate-changelog \
  --changelog-file changelog.no-nulls.h2.xml \
  --preserve-null-values false

liquibase generate-changelog \
  --changelog-file changelog.no-nulls.h2.yaml \
  --preserve-null-values false
```

So, dear @tati-qalified, can you please re-run your tests based on the latest changes in this feature branch ?
Or confirm, this test setup is sufficient ?

Thank you.

Regards,
Carlo

MalloD12:
Hi @catull.

I've been working throught this PR today and I think we should make a couple of changes:
- Solution is not currently building, you can check any of the executions (for example, [this one](https://github.com/liquibase/liquibase/actions/runs/19316628635/job/55334184712?pr=6422)) because of some failing tests.
- `protected boolean preserveNullValues = true;` added in the **YamlSerializer** (L44) it is not backwards compatible change. This same thing is defined in the **FormattedSqlChangeLogSerializer**, but it's never used.
- XMLSerializer issue,  even though we have defined `protected boolean preserveNullValues = false;` I think it could become a deserialization issue if someone make the following call `xmlSerializer.preserveNullValues(true)`, specially for cases where column is not a varchar type.
 - And lastly, I think it would be nice to add some additional tests, especially around `ColumnConfig.isSerializable()`, deserialization tests, and some generate-changelog test using `preserveNullValues` flag.

Let me know if you need any help with any of the above points.

Thanks,
Daniel.

Issue #6023, opened 2024-06-18
XSD accepts most elements as root XML element


b-gyula:
### Search first

- [X] I searched and no similar issues were found

### Description

Albeit the documentation says `databaseChangeLog` must be the root element in XML format, the system and the XSD allows to use most elements as root XML element like `tagDatabase`.

The following XML works fine and does nothing:
```xml
<tagDatabase tag='test' xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
                   xmlns="http://www.liquibase.org/xml/ns/dbchangelog"
 xsi:schemaLocation="http://www.liquibase.org/xml/ns/dbchangelog http://www.liquibase.org/xml/ns/dbchangelog/dbchangelog-4.16.xsd"/>
```
command `liquibase update` output:
```
Starting Liquibase at 16:25:25 (version 4.16.1 #4571 built at 2022-09-11 16:02+0000)
Liquibase Version: 4.16.1
WARNING: License service not loaded, cannot determine Liquibase Pro license status. Please consider re-installing Liquibase to include all dependencies. Continuing operation without Pro license.
Liquibase command 'update' was executed successfully.
```

### Steps To Reproduce

1. Create a simple xml containing only one of the following elements:
stop, empty, output, renameTable, executeCommand, delete, column, sql, sqlFile, and, not, or, createProcedure, tagDatabase, renameView, dropTable, dropView, dropProcedure, renameColumn, dropColumn, mergeColumns, createSequence, alterSequence, dropSequence, renameSequence, createIndex, addAutoIncrement, addDefaultValue, dropDefaultValue, loadData, loadUpdateData, dbms, tableExists, columnExists, sequenceExists, dropIndex, rollback, update, rowCount, expectedQuotingStrategy, runningAs, addNotNullConstraint, dropNotNullConstraint
like:
```xml
<dbms type='test' xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
                   xmlns="http://www.liquibase.org/xml/ns/dbchangelog"
 xsi:schemaLocation="http://www.liquibase.org/xml/ns/dbchangelog http://www.liquibase.org/xml/ns/dbchangelog/dbchangelog-4.16.xsd"/>
```
2. Excute command `liquibase update`

### Expected/Desired Behavior

[...]

PR #6430, opened 2024-10-16
XSD changes:

b-gyula:
- Do not allow other root element than `databaseChangeLog`
- added documentation for some elements
- Fix: addNotNullConstraint@schemaName + @catalogName was missing
- attributeGroups used only once merged into the elements they are used
- New attributeGroups for common attributes: schemaNameAttribute
- integerExp made based on xsd:long and non-negative
- property@name restricted according to the documentation site
- required string parameters made `nonEmptyString` to avoid SQL error for definitions like
-`rowCount@expectedRows`, `setTableRemarks@remarks` made mandatory

Code changes:
- simple clear error message if the root element is other than `databaseChangeLog`: databaseChangeLog expected as root element
- 'uniqueConstraint' was misspelled## Impact


MalloD12:
Hi @b-gyula,

This PR is still in draft state, is there any additional changes you would like to make here?  Updating XSD is not a trivial thing and is something that would require deeper testing. I think we are making many changes here, and it would be a bit risky moving forward with all this. I think it would be a good idea to split up this PR into smaller pieces. For one side, I think we can address the root element issue, documentation updates for one other, required parameters for another, etc. 

So far, I've seen missing attributes for some specific elements. For example, for `property` element I could see we are not including the following ones:

- relativeToChangelogFile
- errorIfMissing
- context
- contextFilter
- labels
- global

On the other hand, I could see these changes are breaking our integration tests (hence breaking some of our elements for XML changelog format).

- For example,  for `onChangeLogPreconditionErrorOrFail` type I have to add the below values:

```xml
<xsd:simpleType name="onChangeLogPreconditionErrorOrFail">
        <xsd:restriction base="xsd:string">          
            <xsd:enumeration value="MARK_RAN"/>
            <xsd:enumeration value="CONTINUE"/>
        </xsd:restriction>
    </xsd:simpleType>
```

- `Column` complex type is also broken without `mixed="true"`

-  `dropView` is missing `ifExists` attribute.

```xml
<xsd:complexType name="dropView">      
        <xsd:attribute name="ifExists" type="booleanExp"/>
</xsd:complexType>
```
- etc

MalloD12:
Hi @b-gyula,

As mentioned above, I think it would be good to split up this PR into smaller pieces.  We are thinking of taking part of the changes you submitted in this PR to fix #6023. Then, we can keep discussing the other refactoring you are tyring to address here. How does it sound to you?

Thanks,
Daniel.

mariochampion:
hey @MalloD12 lets take the smaller part of the PR we knwo will not cause issues and move that through, and if that means a whole new issue with a comment here to link them that would be fine. I ll leave it up to you how to make that happen. In the mean time i ll move this to FTT column and if we hear back from @b-gyula later we ll revive discussion of the deferred elements of the original PR. thanks!

MalloD12:
@mariochampion, I took a smaller piece of this PR to solve the root element issue and added an additional piece of code that was needed. I thought this was enough, but I was breaking some existing tests that I was updating incorrectly. We discussed it with @filipelautert and agreed probably the best way to fix the root element issue would be to read the XML changelog content as a String and then verify whether the first tag element is or not a `databaseChangeLog` tag, I tried to do that on Friday without any luck yet, but I'll revisit this once I finish with the dashboard task.

Thanks,
Daniel.

b-gyula:
> Hi @b-gyula,
> 
> As mentioned above, I think it would be good to split up this PR into smaller pieces. We are thinking of taking part of the changes you submitted in this PR to fix #6023. Then, we can keep discussing the other refactoring you are tyring to address here. How does it sound to you?
> 
> Thanks, Daniel.

Hi,

I've fixed the errors.
Can you provide the test failing?
How do you want me to split it?


The original file format of Liquibase was XML, so the primary UI for Liquibase is an XSD aware XML IDE / editor. To get the developer experience close to other languages like java for example, the XSD should provide *valid* autocomplete and on spot documentation. It would give huge help for developers
code-completition That's my goal. With my changes I try to make the XSD match the source code requirements and logic as much as possible.
The XSD is a structured format, that is perfectly capable to fulfill these in IDEs like Eclipse, IntelliJ, etc. Plus it can enforce constraints *in the editor* by visually highlighting to help avoid very basic errors like empty `name` attribute in `insert/column` tag, that would give strange error just much later @ runtime like: `Column not found; SQL statement: INSERT INTO PUBLIC.a ("") VALUES (NULL)`.
The XSD validation error would pinpoint the exact row in the file where it fails, whereas from the runtime error it is very difficult to identify the exact place (especially in case of included file)

Regarding the documentation, as I understood the original approach was to apply well designed annotations like `@DatabaseChange` and `@DatabaseChangeProperty` in the source. The whole documentation site was generated from those. Since it was not complete (e.g missing defaults, types) I've revised most of the  source: updated descriptions from the documentation site and from the code logic. Then improved the doc site generator to extract those extra information. Few month later the whole documentation site was restructured and my weeks of work was abandoned. I still use that because, albeit since the documentation site has been redesigned several times since, one still can't find what is the default value e.g `loadData@quotChar` See: https://b-gyula.github.io/liquibase-doc/documentation/changes/
The annotations in the code still does not seem to be maintained, so I thought let's make the XSD as the master for change descriptions. That's why I started compile the documentation into it.
It is a major decision needs to be taken: what should be the primary source of the changelog documentation the source code or the XSD. If the code is, the descriptions should be generated from that into the XSD.
If the XSD, then the documentation site could be at - least partially - generated from it.

MalloD12:
Hi @b-gyula,

Sorry for taking a while to answer; we have been working on some other priorities. Let me list my comments below:

- For the changes made for "Do not allow other root element than databaseChangeLog" we have used part of your code changes (and made some others), and created #6827. I think this is the reason for the merge conflicts we are seeing in this PR now.
-  Regarding the approach for the rest of the changes. I would recommend if we can split the rest of the changes in individual PRs, which would help us provide a faster review and help our QA team to validate the changes. Specially, for the XSD changes, which is a sensitive piece. 
- Regarding documentation, let's rely on the docs site. XSD main purpose is to define and control XML changelog structure. Although we are having an internal discussion with the team on the differences between the docs site and XSD, and how we keep each current.

Thanks,
Daniel.

MalloD12:
Hi @b-gyula,

Have you had a chance to have a look at my previous comment?

Thanks,
Daniel.

PR #6431, opened 2024-10-16
[DRAFT] Enable selection of Checksum hashing algorithm

### LINKED ISSUE #6062: [Feature Request]
### Search first

### Description

Allow Liquibase, by configuration, to use SHA256 instead of MD5.

### Steps To Reproduce

None.  Feature request.

### Expected/Desired Behavior

Allow by configuration to specify which hash digest, MD5 or SHA256, will be used by Liquibase.  Alternately move the changelog hashing to SHA256.

### Liquibase Version

4.28.0

### Database Vendor & Version

NA

### Liquibase Integration

JVM

### Liquibase Extensions

NA

NA

### Additional Context

There are some companies, sadly mine included, that believe MD5 is an agent of evil and even mentioning it in a codebase will instantly cause all security to fail.  Regardless if it is only being used in a way to provide a general hash value to avoid collisions in non-cryptographically-security situations.  Having a config value to indicate that SHA256 should be used in preference to MD5 would mean not having to jump through hoops.

filipelautert:
## Impact

## Description

Reasoning from source issue: There are some companies, that believe MD5 is an agent of evil and even mentioning it in a codebase will instantly cause all security to fail. Regardless if it is only being used in a way to provide a general hash value to avoid collisions in non-cryptographically-security situations. Having a config value to indicate that SHA256 should be used in preference to MD5 would mean not having to jump through hoops.

What this PR does:
- Introduces a new Global Flag , CHECKSUM_ALGORITHM, that allows to switch from default MD5 to SHA1 or SHA256
- Increase the MD5 column from databasechangelog table  to support those new hash sizes
- Refactorings and tests

## Things to be aware of
So far, this PR is not handling migrations. So if an user wants to go from` MD5 to SHAX`, he first needs to run `clearchecksums` **THEN** run an `update`. Same logic to rollback.

Should we add a new `changeCheckSums` command to handle this migration smoothly? Should it be done under the hood?

## Things to worry about

We are not storing the algorithm used to calculate checksum anywhere. So if someone runs liquibase using` -checksumAlgorithm=SHA256 `, and in the next run forgets the parameter there will be a checksum error message. 

Should we improve our current checksum from **VERSION:CHECKSUM** to **VERSION:ALGORITHM:CHECKSUM** , or add a new column to dbcl, or do nothing? some thoughts: 
* adding the algorithm we can have mixed checksum algorithms, so the user can switch from MD5 to SHAX without changing the existing values. 
* But this would be against the whole point of changing the checksum, right?
* This would be mostly useful if the user forgot which algorithm he used, or to remove the requirement of keeping the global flag after migrating everything - but who does it?
* Changing the checksum string would break backwards compatibility, and it means that we would need to bump checksums version again, impacting _everyone_ - even people that is not migrating
* Adding a new column would be less intrusive is this case, as we don't need to bump checksum version
* Doing nothing is already implement, works like a charm

## Additional Context

* Fixes #6062 
* Still need to verify if there is any impact to liquibase-commercial (hopefully not)

### DIFF
--- a/liquibase-standard/src/main/java/liquibase/changelog/StandardChangeLogHistoryService.java
+++ b/liquibase-standard/src/main/java/liquibase/changelog/StandardChangeLogHistoryService.java
@@ -49,6 +49,8 @@ public class StandardChangeLogHistoryService extends AbstractChangeLogHistorySer
     protected static final String LABELS_SIZE = "255";
     protected static final String CONTEXTS_SIZE = "255";
 
+    public static final String MD5_COLUMN_SIZE = "68";
+
     @Override
     public int getPriority() {
         return PRIORITY_DEFAULT;
@@ -130,6 +132,7 @@ public class StandardChangeLogHistoryService extends AbstractChangeLogHistorySer
             boolean hasLiquibase = changeLogTable.getColumn("LIQUIBASE") != null;
             boolean hasContexts = changeLogTable.getColumn("CONTEXTS") != null;
             boolean hasLabels = changeLogTable.getColumn("LABELS") != null;
+            boolean hasSmallChecksum = changeLogTable.getColumn("MD5SUM").getType().getColumnSize() < Integer.parseInt(MD5_COLUMN_SIZE);
             boolean liquibaseColumnNotRightSize = false;
             if (!(this.getDatabase() instanceof SQLiteDatabase)) {
                 DataType type = changeLogTable.getColumn("LIQUIBASE").getType();
@@ -248,6 +251,19 @@ public class StandardChangeLogHistoryService extends AbstractChangeLogHistorySer
                         getLiquibaseSchemaName(), getDatabaseChangeLogTableName()));
                 }
             }
+
+            if (hasSmallChecksum) {
+                executor.comment("Modifying size of databasechangelog.md5sum column");
+                statementsToExecute.add(new ModifyDataTypeStatement(getLiquibaseCatalogName(),
+                        getLiquibaseSchemaName(), getDatabaseChangeLogTableName(), "MD5SUM",
+                        charTypeName + "(" + MD5_COLUMN_SIZE + ")"));
+
+            }
+
+            SqlStatement databaseChangeLogStatement = new SelectFromDatabaseChangeLogStatement(
+                    new SelectFromDatabaseChangeLogStatement.ByCheckSumNotNullAndNotLike(ChecksumVersion.latest().getVersion()),
+                    new ColumnConfig().setName("MD5SUM"));
+            List<Map<String, ?>> md5sumRS = ChangelogJdbcMdcListener.query(getDatabase(), ex -> ex.queryForList(databaseChangeLogStatement));
 
             //check if any checksum is not using the current version
             databaseChecksumsCompatible = getIncompatibleDatabaseChangeLogs().isEmpty();
--- a/liquibase-standard/src/main/java/liquibase/changelog/StandardChangeLogHistoryService.java
+++ b/liquibase-standard/src/main/java/liquibase/changelog/StandardChangeLogHistoryService.java
@@ -132,7 +132,7 @@ public class StandardChangeLogHistoryService extends AbstractChangeLogHistorySer
             boolean hasLiquibase = changeLogTable.getColumn("LIQUIBASE") != null;
             boolean hasContexts = changeLogTable.getColumn("CONTEXTS") != null;
             boolean hasLabels = changeLogTable.getColumn("LABELS") != null;
-            boolean hasSmallChecksum = changeLogTable.getColumn("MD5SUM").getType().getColumnSize() < Integer.parseInt(MD5_COLUMN_SIZE);
+            boolean hasShortChecksum = changeLogTable.getColumn("MD5SUM").getType().getColumnSize() < Integer.parseInt(MD5_COLUMN_SIZE);
             boolean liquibaseColumnNotRightSize = false;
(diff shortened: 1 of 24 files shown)


liquibase-standard/src/main/java/liquibase/util/MD5Util.java
```diff
@@ -1,97 +1,23 @@
...
-
-        String returnString = new String(encodeHex(digestBytes));
-
-        String inputToLog = input;
-        if (inputToLog.length() > 500) {
-            inputToLog = inputToLog.substring(0, 500)+"... [truncated in log]";
-        }
-        Scope.getCurrentScope().getLog(MD5Util.class).fine("Computed checksum for "+inputToLog+" as "+returnString);
-        return returnString;
+        return HashingUtil.compute(input, "MD5");
```
StevenMassaro, line 15: Can you use the ChecksumAlgorithm enum in this class?

liquibase-standard/src/main/java/liquibase/checksums/ChecksumAlgorithm.java
```diff
@@ -0,0 +1,17 @@
...
+public enum ChecksumAlgorithm {
+    MD5("MD5"),
+    SHA1("SHA-1"),
+    SHA256("SHA-256");
+
+    private final String algorithm;
+
+    ChecksumAlgorithm(String algorithm) {
+        this.algorithm = algorithm;
+    }
```
StevenMassaro, line 15: Could use lombok for this if you want

liquibase-standard/src/main/java/liquibase/changelog/StandardChangeLogHistoryService.java
```diff
@@ -244,6 +247,14 @@ public void init() throws DatabaseException {
                 }
             }
 
+            if (hasShortChecksum) {
+                executor.comment("Modifying size of databasechangelog.md5sum column");
+                statementsToExecute.add(new ModifyDataTypeStatement(getLiquibaseCatalogName(),
+                        getLiquibaseSchemaName(), getDatabaseChangeLogTableName(), "MD5SUM",
+                        charTypeName + "(" + MD5_COLUMN_SIZE + ")"));
+
+            }
```
StevenMassaro, line 256: I wonder how many databases support this type of type migration. I don't know offhand, but I'm pretty sure it isn't ALL databases. I think some you have to do some data copying nonsense. Or does Liquibase handle that for you?

liquibase-standard/src/main/java/liquibase/changelog/StandardChangeLogHistoryService.java
```diff
@@ -46,6 +46,8 @@ public class StandardChangeLogHistoryService extends AbstractChangeLogHistorySer
     protected static final String LABELS_SIZE = "255";
     protected static final String CONTEXTS_SIZE = "255";
 
+    public static final String MD5_COLUMN_SIZE = "68";
```
StevenMassaro, line 49: How did you calculate this?

StevenMassaro:
> Should we add a new changeCheckSums command to handle this migration smoothly? Should it be done under the hood?

I would prefer we not add a command for this, as it is unlikely to happen often. More to maintain and go wrong. If we do, it should be an effort separate from this PR. I also think doing it under the hood should be separate from this effort too (but don't think we should do it under the hood either).

> Should we improve our current checksum from VERSION:CHECKSUM to VERSION:ALGORITHM:CHECKSUM , or add a new column to dbcl, or do nothing?

My vote would be to do nothing. Messing with the DBCL DDL is not a lot of fun, so the less we have to change, the better.

lbadam:
This is great @filipelautert!

I agree with @StevenMassaro here: 

> My vote would be to do nothing. Messing with the DBCL DDL is not a lot of fun, so the less we have to change, the better.

One question about this scenario:

> if someone runs liquibase using-checksumAlgorithm=SHA256, and in the next run forgets the parameter there will be a checksum error message

If they see the error and then remember to add `checksumAlgorithm=SHA256` will they be able to proceed without a problem?

filipelautert:
> If they see the error and then remember to add `checksumAlgorithm=SHA256` will they be able to proceed without a problem?

@lbadam right, we do not change anything. Based on the checksum size we can try to guess and give them a hint too: as md5 is `32` chars, SHA1is `40 `and SHA256 is `64 `we can check the size and say something like `Did you mean to use SHA1 as checksum algorithm? Try adding ......`

lbadam:
@filipelautert I think that suggestion would be a really good addition. 

I'm also good with this solution as-is to help users leverage SHA256. Once people are using it, we should get feedback to see if we need to add or change the behavior at all.

liquibase-standard/src/main/java/liquibase/changelog/StandardChangeLogHistoryService.java
filipelautert: SHA256 size is `64 `. So 64 + `:` + version up to 3 digits (thinking big here!)
Originally it was `35`: 32 for MD5  + `:` + 2 version up to 2 digits

jmbharathram:
can I get an update on when this feature might be available to us please?  @mariochampion @filipelautert

filipelautert:
TODO: add this parameter as a Springboot option

filipelautert:
> can I get an update on when this feature might be available to us please? 

@jmbharathram probably later this year in Liquibase 5.0 .

jmbharathram:
@filipelautert - I see that version 5.0.x is out. But this feature isn't supported yet. Is that correct?

PR #5988, opened 2024-06-10
High Level API (aka undeprecate Liquibase class)

filipelautert:
## Impact

This PR undeprecates Liquibase class methods, improves javadocs and removes deprecated dependencies/sonar warnings as requested by community in #5663, #5066, forum and other locations.

Fixes #5663

PR #6432, opened 2024-10-16
Adds customizer to discover and react to unknown change sets

raphw:
## Impact

## Description

Adds a `Customizer` to be used in Spring that allows to discover change sets that are not known to an application. Typically this can be used to prevent an application's startup as described here: https://github.com/spring-projects/spring-boot/issues/42699 as experienced in this scenario:

    1. An application with Liquibase was updated from version 1 to version 2 on Kubernetes.
    2. The application started up and applied a new Liquibase change set.
    3. The startup failed for different reason and an automatic rollback was applied.
    4. Version 1 of the application came back up, seeing the change set of version 2.
    5. The application started but misbehaved in a rather destructive manner due to an unforeseen consequence of the new data model in combination with the old SQL.

## Things to be aware of

This listener has become possible after the `Customizer` API was introduced.

## Things to worry about

I am not entirely sure if this is correct with the details of Liquibase's API, but it worked for my tests.

## Additional Context

This solves an actual production issue that I experienced. The Spring team suggested a convenience configuration might be added to Spring Boot if this feature is maintained within Liquibase.


liquibase-standard/src/main/java/liquibase/integration/spring/UnexpectedChangeSetsCustomizer.java
```diff
@@ -0,0 +1,79 @@
...
+    public void customize(T liquibase) {
+        try {
+            liquibase.getDatabaseChangeLog().getChangeSets().stream().map(changeSet -> new ChangeSetInfo(
+                    changeSet.getId(),
+                    changeSet.getChangeLog().getFilePath(),
+                    changeSet.getAuthor())).forEach(applied::add);
+            UnexpectedChangesetsCommandStep.listUnexpectedChangeSets(
+                liquibase.getDatabase(),
+                liquibase.getDatabaseChangeLog(),
+                new Contexts(),
```
filipelautert, line 37: The contexts and labels are available in [SpringLiquibase ](https://github.com/liquibase/liquibase/blob/master/liquibase-standard/src/main/java/liquibase/integration/spring/SpringLiquibase.java#L76) class.  It there a way to get them here and use the contexts and labels from it?

liquibase-standard/src/main/java/liquibase/integration/spring/UnexpectedChangeSetsCustomizer.java
```diff
@@ -0,0 +1,79 @@
...
+    }
+
+    @Override
+    public void customize(T liquibase) {
+        try {
+            liquibase.getDatabaseChangeLog().getChangeSets().stream().map(changeSet -> new ChangeSetInfo(
+                    changeSet.getId(),
+                    changeSet.getChangeLog().getFilePath(),
+                    changeSet.getAuthor())).forEach(applied::add);
+            UnexpectedChangesetsCommandStep.listUnexpectedChangeSets(
```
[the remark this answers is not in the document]
raphw: Those methods are deprecated, this is why I thought I should avoid them?

liquibase-standard/src/main/java/liquibase/integration/spring/UnexpectedChangeSetsCustomizer.java
raphw: If the customizer is registered as a bean in Spring, it does not receive the `SpringLiquibase` but only the `Liquibase` instance. But one could add a method to adjust.

raphw: I added a way to add `LabelExpressions` and `Contexts`.

liquibase-standard/src/main/java/liquibase/integration/spring/UnexpectedChangeSetsCustomizer.java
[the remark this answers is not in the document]
filipelautert: Ah true, you are were right. But they were "undeprecated" for 4.29.0 -> https://github.com/liquibase/liquibase/pull/5988 .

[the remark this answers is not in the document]
raphw: Indeed, I subsituted.

liquibase-standard/src/main/java/liquibase/integration/spring/UnexpectedChangeSetsCustomizer.java
filipelautert: Hi @raphw - did you mean to push changes to add labels and contexts?

raphw: I did but tried to push to origin and not to my branch, so it failed. Fixed now!

raphw:
Do I need to sign some CA to make this fossa scan step work?

filipelautert:
> Do I need to sign some CA to make this fossa scan step work?

Don't worry, we need to fix this one for contributed PRs.

raphw:
I tested locally, at least. I am not entirely sure if `ContextRefreshedEvent` is the right event, but I think it is. Possibly, one could ask the Spring team for feedback. @wilkinsona, what do you think?

wilkinsona:
As I understand the problem, it could occur in any situation where Liquibase is used and a rollback of an application deployment doesn't also rollback the database change sets such that the old app's left looking at the new schema. With this in mind, I don't see a need to couple this enhancement `SpringLiquibase` and its `Customizer` callback. Could it not be implemented with "plain" Liquibase, thereby broadening its benefit and also making it easier to test?

raphw:
I agree that this might be a better approach. The main reason to bind this to SpringLiquibase was the way it was discovered. The class could however just implement the Consumer interface, too, and offer similar functionality independently of an integration.

This way I could also write tests.

Should I adjust this PR?

filipelautert:
Tests would be nice!
But regarding usage, I can't imagine other use cases for it. As for command line and Maven we already provide the `unexpectedChangesets` command, and if you use liquibase as a library you can use the same code that was added here. Also having unexpected changesets not always means that you have a problem, so we can't take any action for the user besides warning him. But by refactoring and moving some code to `liquibase-standard` you can add any tests that you want !

raphw:
I think more of being able to integrate it with micronaut and quarkus, for example. Those could certainly make use of this if it was plain Liquibase.

raphw:
@filipelautert Here is a new suggestion, that I think offers a better abstraction without requiring the use of Spring. I also added tests.

raphw:
I pushed the suggestion to this branch: https://github.com/raphw/liquibase/tree/integration-change-set

filipelautert:
> I pushed the suggestion to this branch: https://github.com/raphw/liquibase/tree/integration-change-set

@raphw sorry for the delay to answer and thanks for the new branch. Seems a better approach, can you merge it to your master so we have it in this PR?

raphw:
Super. Merged!

raphw:
I am wondering. In:

            liquibase.getDatabaseChangeLog().getChangeSets().stream().map(changeSet -> new ChangeSetInfo(
                    changeSet.getId(),
                    changeSet.getFilePath(),
                    changeSet.getAuthor())).forEach(applied::add);
            liquibase.listUnexpectedChangeSets(contexts, labelExpression).stream().map(changeSet -> new ChangeSetInfo(
                    changeSet.getId(),
                    changeSet.getChangeLog(),
                    changeSet.getAuthor())).forEach(unexpected::add);

Will `changeSet.getFilePath()` be the same as `changeSet.getChangeLog()`? Before I called `changeSet.getChangeSet().getFilePath()`, but I am not sure if that is the same or which one should be used.

filipelautert:
> Will `changeSet.getFilePath()` be the same as `changeSet.getChangeLog()`? Before I called `changeSet.getChangeSet().getFilePath()`, but I am not sure if that is the same or which one should be used.

Yes, they should be always the same , so this fix is fine!

filipelautert (approved):

tati-qalified (approved):
@raphw Thank you for your contribution!

rberezen (approved):

Issue #6424, opened 2024-10-15
createProcedure fails with exception if contains only comment as body


b-gyula:
### Search first

- [x] I searched and no similar issues were found

### Description

If a createProcedure tag contains only block comment as body, then the execution fails with exception:
```
java.lang.StringIndexOutOfBoundsException: String index out of range: 18
	at java.base/java.lang.StringLatin1.charAt(StringLatin1.java:47)
	at java.base/java.lang.String.charAt(String.java:693)
	at liquibase.util.StringUtil.getLastBlockComment(StringUtil.java:944)
	at liquibase.util.StringUtil.stripSqlCommentsAndWhitespacesFromTheEnd(StringUtil.java:1019)
	at liquibase.sqlgenerator.core.CreateProcedureGenerator.removeTrailingDelimiter(CreateProcedureGenerator.java:123)
	at liquibase.sqlgenerator.core.CreateProcedureGenerator.generateSql(CreateProcedureGenerator.java:87)
	at liquibase.sqlgenerator.core.CreateProcedureGenerator.generateSql(CreateProcedureGenerator.java:27)
	at liquibase.sqlgenerator.SqlGeneratorChain.generateSql(SqlGeneratorChain.java:32)
	at liquibase.sqlgenerator.SqlGeneratorFactory.generateSql(SqlGeneratorFactory.java:224)
	at liquibase.util.SqlUtil.getSqlString(SqlUtil.java:360)
	at liquibase.changelog.ChangeSet.lambda$addSqlMdc$3(ChangeSet.java:1617)
	at java.base/java.util.stream.ReferencePipeline$3$1.accept(ReferencePipeline.java:195)
	at java.base/java.util.Spliterators$ArraySpliterator.forEachRemaining(Spliterators.java:948)
	at java.base/java.util.stream.AbstractPipeline.copyInto(AbstractPipeline.java:484)
	at java.base/java.util.stream.AbstractPipeline.wrapAndCopyInto(AbstractPipeline.java:474)
	at java.base/java.util.stream.ReduceOps$ReduceOp.evaluateSequential(ReduceOps.java:913)
	at java.base/java.util.stream.AbstractPipeline.evaluate(AbstractPipeline.java:234)
	at java.base/java.util.stream.ReferencePipeline.collect(ReferencePipeline.java:578)
	at liquibase.changelog.ChangeSet.addSqlMdc(ChangeSet.java:1618)
[...]

PR #6436, opened 2024-10-17
check if the next char out of length

fntz:
## Impact

## Description

fix for https://github.com/liquibase/liquibase/issues/6424

### DIFF
--- a/liquibase-standard/src/main/java/liquibase/util/StringUtil.java
+++ b/liquibase-standard/src/main/java/liquibase/util/StringUtil.java
@@ -943,6 +943,7 @@ public class StringUtil {
             return null;
         }
         String reversedString = StringUtils.reverse(sqlString);
+        int length = reversedString.length();
         int idxClosingLastChar = -1, idxOpeningFirstChar = -1;
         for (int i = 0; i < reversedString.length(); i++) {
             if (idxClosingLastChar < 0) {
@@ -950,9 +951,11 @@ public class StringUtil {
                 char c = reversedString.charAt(i);
                 if (c == '/') {
                     // check the second one
-                    char s = reversedString.charAt(i + 1);
-                    if (s == '*') {
-                        idxClosingLastChar = i;
+                    if (i + 1 < length) {
+                        char s = reversedString.charAt(i + 1);
+                        if (s == '*') {
+                            idxClosingLastChar = i;
+                        }
                     }
                 } else if (!Character.isWhitespace(c)) {
                     // does not look like it ends with block comment, return null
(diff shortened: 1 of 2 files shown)


MalloD12:
Hi Mike (@fntz),

Thank you for submitting a PR to fix #6424! Code changes added look good to me. I would only ask for adding some additional tests for this given issue. I was thinking we could add the changeset example provided in the description of the issue into the `common.tests.changelog.xml` file that contains a list of changesets that integration tests will deploy, that specific file will be executed by some of the different DBs that Liquibase support.

Example:
```xml
<changeSet id="createProcedureWithCommentedBodyTest" author="testUser">
		<createProcedure>
			/*CREATE PROC... */
		</createProcedure>
</changeSet> 
```

Additionally, while manually testing your fix we have seen that for example if the `createProcedure` change object does not contain a body (let's say this is an empty body) the `update` execution is also failing with the below error:

```console
Caused by: liquibase.exception.ValidationFailedException: Validation Failed:
          Must specify either 'path' or a nested procedure text in createProcedure, changelog/xml/production.issue.xml::1::testprocedure
```

I think that's a similar case than this one and it would be good to also fix it, but in my opinion that would be a separate issue. Would you be keen on working on that issue as well?

Thanks,
Daniel.

MalloD12:
Hi @fntz,

I've added a test scenario that contains the `createProcedure` change with an empty SQL body that will be executed by our integration tests. This given changeset is failing for some DBs like `db2`, `hsql`, `oracle`, and `firebird`. Some of the errors are like the one below:

```java
Error:    DB2IntegrationTest>AbstractIntegrationTest.testUpdateTwice:547 » CommandExecution liquibase.exception.LiquibaseException: liquibase.exception.MigrationFailedException: Migration failed for changeset changelogs/common/common.tests.changelog.xml::createProcedureWithOnlyBlockCommentAsBody::mallod:
     Reason: liquibase.exception.DatabaseException: [jcc][10234][10927][4.33.31] SQL passed with no tokens. ERRORCODE=-4462, SQLSTATE=null [Failed SQL: (-4462) ]
```

which means that `creteProcedure` change is trying to execute the SQL body, but since this is a block comment it fails for the mentioned DBs.

MalloD12 (approved):
**Approved**.

Code changes look good to me, this helpful to avoid  `StringIndexOutOfBoundsException` error happening again. Additionally to the unit test added we have added a test scenario that will be executed as part of our Integration tests. This test helped us to identify there are some DBs (like Oracle, HSQL, etc) that doesn't accept an empty/block of comment for the SQL body, because the given DB executor will try to execute that body, and it will fail because there are no SQL token present on it. We have decided to move forward with it leaving on user judgement that just a block of comment is not what some others DBs expect when creating a procedure.

filipelautert (approved):

rberezen (approved):

PR #6446, opened 2024-10-21
Allow for scopemanager to be shared between threads created by liquibase (DAT-18871)

abrackx:
## Impact

## Description
When sending analytics we create a new thread to easily time out the thread in case processing takes too long. This works fine for normal use cases, but in the event we are say, redirecting the logs to a file or using structured logging with our pro extension, we need to check the liquibase configuration and grab information via the scope. This can result in unexpected errors from inside a new thread as the `scopeManager` has not been set up in this new thread. By swapping to a `InheritableThreadLocal`, we automatically transfer this scopevalue into the child thread without having to muck with copying the scope values. 

## Things to worry about
As far as I can tell this should work the same as a normal `ThreadLocal` but allow us to not need to reconfigure our scope when spawning child threads. That said, changing something this fundamental to Liquibase is always a risk. 

When chatting with the development team we came to the conclusion that the only way this change should be noticed is if a user creates a thread within the thread they are running liquibase on and then runs liquibase again. This form of usage seems unlikely.

### DIFF
--- a/liquibase-standard/src/main/java/liquibase/Scope.java
+++ b/liquibase-standard/src/main/java/liquibase/Scope.java
@@ -78,7 +78,7 @@ public class Scope {
 
     public static final String JAVA_PROPERTIES = "javaProperties";
 
-    private static final ThreadLocal<ScopeManager> scopeManager = new ThreadLocal<>();
+    private static final InheritableThreadLocal<ScopeManager> scopeManager = new InheritableThreadLocal<>();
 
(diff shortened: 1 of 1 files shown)


filipelautert (approved):
🤞

StevenMassaro (approved):
Could you update the analytics code, in the two threads that are created, to move the initialization of things inside the threads and remove the comments explaining why it's outside the thread?

PR #6451, opened 2024-10-23
Migrate init project command to core (DAT-15504)

abrackx:
## Impact


liquibase-standard/src/main/java/liquibase/util/StringUtil.java
```diff
@@ -1171,4 +1172,13 @@ public static List<String> splitToChunks(String input, int chunkSize) {
         }
         return chunks;
     }
+
+    /**
+     *
+     * Remove the "file://" prefix from a URI if it exists.
+     *
+     */
+    public static String stripUriPrefix(URI uri) {
```
filipelautert, line 1181: This method name can be misleading, could it be `stripFileUrlPrefix` or something like that?

liquibase-standard/src/main/resources/META-INF/services/liquibase.command.CommandStep
filipelautert, line 1: What? Now `serviceloader-maven-plugin` handles it? That's a bonus change for this PR.

filipelautert (changes requested):

liquibase-integration-tests/src/test/resources/liquibase.sample.properties
```diff
@@ -0,0 +1,66 @@
+####     _     _             _ _
```
filipelautert, line 1: nice

liquibase-standard/src/main/resources/META-INF/services/liquibase.command.CommandStep
abrackx: Yep! It's time to embrace the service loader 😁

liquibase-standard/src/main/java/liquibase/util/StringUtil.java
```diff
@@ -1171,4 +1172,13 @@ public static List<String> splitToChunks(String input, int chunkSize) {
...
+        return uri.toString().replace("file://","");
```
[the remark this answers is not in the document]
abrackx: I've updated this to check for null and return an empty string. I've also migrated this out of `StringUtil` since the only usage was in a single `InitProject` class. It can stay private for now.

liquibase-standard/src/main/java/liquibase/util/StringUtil.java
abrackx: Good call! I've renamed this to `stripFileUriPrefix`. I've also migrated this out of StringUtil since the only usage was in a single InitProject class. It can stay private for now.

filipelautert (approved):

PR #6461, opened 2024-10-25
DAT-17706 - Add checks example files to all format directories

filipelautert:


liquibase-standard/src/main/java/liquibase/snapshot/DatabaseSnapshot.java
```diff
@@ -690,8 +690,6 @@ public ParsedNode serialize() {
 
     /**
      * Used to get and store misc data that should be scoped to the snapshot. Helpful for caching snapshot results.
-     *
-     * @deprecated Will be removed with 4.0
```
filipelautert, line 694: The undeprecation at DatabaseSnapshot.java is not related here, just using the PR to merge it too .

OliaLub (approved):

PR #6472, opened 2024-10-30
 Fail validation for blank changeset ids (DAT-18773)

abrackx:
## Impact

## Description
 If a changeset id is entirely blank (one or more spaces) fail validation.

## Things to be aware of
Do not fail validation for changes that have already been ran. Changes that have been ran with an empty changeset id that are marked as `runOnChange` or `runAlways` WILL fail validation.

### DIFF
--- a/liquibase-standard/src/main/java/liquibase/changelog/visitor/ValidatingVisitor.java
+++ b/liquibase-standard/src/main/java/liquibase/changelog/visitor/ValidatingVisitor.java
@@ -48,7 +48,7 @@ public class ValidatingVisitor implements ChangeSetVisitor {
 
     public ValidatingVisitor(List<RanChangeSet> ranChangeSets) {
         ranIndex = new HashMap<>();
-        for(RanChangeSet changeSet: ranChangeSets) {
+        for (RanChangeSet changeSet : ranChangeSets) {
             ranIndex.put(changeSet.toString(), changeSet);
         }
     }
@@ -67,11 +67,11 @@ public class ValidatingVisitor implements ChangeSetVisitor {
                 preconditions.check(database, changeLog, null, null);
             }
         } catch (PreconditionFailedException e) {
-            Scope.getCurrentScope().getLog(getClass()).warning("Precondition Failed: "+e.getMessage(), e);
+            Scope.getCurrentScope().getLog(getClass()).warning("Precondition Failed: " + e.getMessage(), e);
             failedPreconditionsMessage = e.getMessage();
             failedPreconditions.addAll(e.getFailedPreconditions());
         } catch (PreconditionErrorException e) {
-            Scope.getCurrentScope().getLog(getClass()).severe("Precondition Error: "+e.getMessage(), e);
+            Scope.getCurrentScope().getLog(getClass()).severe("Precondition Error: " + e.getMessage(), e);
             errorPreconditionsMessage = e.getMessage();
             errorPreconditions.addAll(e.getErrorPreconditions());
         } finally {
@@ -94,7 +94,7 @@ public class ValidatingVisitor implements ChangeSetVisitor {
         String key = changeSet.toNormalizedString();
         return ranIndex.get(key);
     }
-        
+
     @Override
     public void visit(ChangeSet changeSet, DatabaseChangeLog databaseChangeLog, Database database, Set<ChangeSetFilterResult> filterResults) throws LiquibaseException {
         if (changeSet.isIgnore()) {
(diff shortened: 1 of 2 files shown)


MalloD12 (approved):

liquibase-standard/src/main/java/liquibase/changelog/visitor/ValidatingVisitor.java
```diff
@@ -184,7 +184,7 @@ protected void validateChange(ChangeSet changeSet, Database database, Change cha
 
     private boolean areChangeSetAttributesValid(ChangeSet changeSet) {
         boolean authorEmpty = StringUtils.isEmpty(changeSet.getAuthor());
```
StevenMassaro, line 186: Should we introduce another validation for the author as well, since it also uses isempty? Maybe in another ticket?

StevenMassaro (approved):

liquibase-standard/src/main/java/liquibase/changelog/visitor/ValidatingVisitor.java
```diff
@@ -184,7 +184,7 @@ protected void validateChange(ChangeSet changeSet, Database database, Change cha
...
-        boolean idEmpty = StringUtils.isEmpty(changeSet.getId());
+        boolean idEmpty = StringUtils.isBlank(changeSet.getId());
```
filipelautert, line 187: @abrackx but this is validation all changesets, including existing ones right? IT's not in the ticket description, but ther eis a comment stating: 
`We will not fix changesets that have been already deployed with empty ChangeSetIDs`
So if the changeset was already applied we should not validate it, otherwise the user will have to fix it and will have a changeset mismatch. 
can we do it here?

filipelautert (changes requested):

liquibase-standard/src/main/java/liquibase/changelog/visitor/ValidatingVisitor.java
MalloD12: I agree. If we apply the same logic for `author`, I think we can remove the strict validation we added some time ago.

liquibase-standard/src/main/java/liquibase/changelog/visitor/ValidatingVisitor.java
abrackx: Good catch! I've changed this validation method to have a `strictIdCheck` flag. 

If the changeset has NOT ran or is about to run (like in the case of `runAlways` or `runOnChange`) we will check this "strictly" aka make sure it's not a space. 
If the changeset has already ran then and is not about to run we will use the previous validation and allow spaces.

I think this makes sense to me, let me know what you guys think @StevenMassaro @filipelautert

filipelautert: @abrackx  This `shouldValidate` flag is interesting, good find! I almost think  that instead of passing it to the method, we could use it in the if:

```
if (shouldValidate && !areChangeSetAttributesValid(changeSet)) {
```

What do you think?

shouldValidate exists since 2013 and the commit message is interesting:

> Don't validate changesets that have already been run.
> Do validate the checksums of all changesets.
> Do validate changesets that should run even if they have run before.

Method `areChangeSetAttributesValid `was introduced ~2 years ago and seems it ignored this premisse from `shouldValidate `

liquibase-standard/src/main/java/liquibase/changelog/visitor/ValidatingVisitor.java
StevenMassaro: @abrackx did you make a ticket here?

abrackx: Created DAT-18956!

liquibase-standard/src/main/java/liquibase/changelog/visitor/ValidatingVisitor.java
abrackx: Updated! Removed the `strictIdCheck` from the `areChangeSetAttributesValid` and added the `shouldValidate` flag to the conditional. Same behavior but now we only validate when the change is actually going to be ran.

liquibase-standard/src/main/java/liquibase/changelog/visitor/ValidatingVisitor.java
```diff
@@ -104,14 +104,14 @@ public void visit(ChangeSet changeSet, DatabaseChangeLog databaseChangeLog, Data
...
             DatabaseList.validateDefinitions(changeSet.getDbmsSet(), validationErrors);
         }
-        changeSet.setStoredCheckSum(ran?ranChangeSet.getLastCheckSum():null);
-        changeSet.setStoredFilePath(ran?ranChangeSet.getStoredChangeLog():null);
+        changeSet.setStoredCheckSum(ran ? ranChangeSet.getLastCheckSum() : null);
+        changeSet.setStoredFilePath(ran ? ranChangeSet.getStoredChangeLog() : null);
         boolean shouldValidate = !ran || changeSet.shouldRunOnChange() || changeSet.shouldAlwaysRun();
 
-        if (!areChangeSetAttributesValid(changeSet)) {
+        if (shouldValidate && !areChangeSetAttributesValid(changeSet)) {
```
filipelautert, line 114: Found the change! the rest is just formatting stuff :)

filipelautert (approved):

PR #6492, opened 2024-11-06
Fix(DAT-18813): classloader configuration in LiquibaseCommandLine

filipelautert:
- Ensure classloader is configured correctly by calling `configureClassLoader` as soon as possible.
- Include all libraries in the new classloader to avoid library conflicts/claspath hell.
- do not use the temporary classloader as a parent, but fall back to the root classloader
- Refactor `LiquibaseLauncher` to store library URLs in a static list for reuse instead of processing them again

## Impact

### DIFF
--- a/liquibase-cli/src/main/java/liquibase/integration/commandline/LiquibaseLauncher.java
+++ b/liquibase-cli/src/main/java/liquibase/integration/commandline/LiquibaseLauncher.java
@@ -1,12 +1,19 @@
 package liquibase.integration.commandline;
 
 import liquibase.Scope;
+import liquibase.resource.DirectoryPathHandler;
+import liquibase.resource.Resource;
+import liquibase.util.LiquibaseLauncherSettings;
 import liquibase.util.StringUtil;
 
 import java.io.File;
+import java.io.IOException;
+import java.io.InputStream;
 import java.net.MalformedURLException;
 import java.net.URL;
 import java.net.URLClassLoader;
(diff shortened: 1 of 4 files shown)


liquibase-cli/src/main/java/liquibase/integration/commandline/LiquibaseLauncher.java
```diff
@@ -116,6 +117,44 @@ public static void main(final String[] args) throws Exception {
...
     }
 
+    protected static ClassLoader configureClassLoader(ClassLoader parentLoader, String[] args, List<URL> libUrls) throws IllegalArgumentException, IOException {
+
+        String classpath = getParameter(LIQUIBASE_CLASSPATH, "classpath", args, true);
+
+        final List<URL> urls = new ArrayList<>(libUrls);
+        if (classpath != null) {
+            String[] classpathSoFar;
+            if (System.getProperties().getProperty("os.name").toLowerCase().startsWith("windows")) {
```
StevenMassaro, line 127: Do you want to use commons here to do the windows check?

StevenMassaro (approved):

liquibase-cli/src/main/java/liquibase/integration/commandline/LiquibaseLauncher.java
filipelautert: I would like to but I can't as it is not loaded yet :(

PR #6506, opened 2024-11-11
Distinguish "null" / "NULL" String values from null (Issue #6504)

### LINKED ISSUE #6504: generateChangeLog: export to SQL of "null" / "NULL" as text values results in "..... VALUES (..., NULL, ...)" instead of " ... VALUES (...., 'null', ...)"
### Search first

### Description

We have a Java `Type` enumeration, with serialised values of "json", "date", "string", "boolean", "serializable", "instant", "integer" and .... "null".

On the DB the corresponding `type_` column has a "CHECK TYPE_ IS NOT NULL" constraint.

We noticed for the `type_` being 'null' in the source DB, those values are output as SQL NULL in an SQL export file.

We need those value to be 'null' in the SQL export file, because they must be turned into `Type.NULL` enumeration.

We cannot export 2 tables from one DB instance to our dev instances because we cannot turn every SQL `NULL` into 'null'.
Some NULL values are legal values.

There is currently no other way to export and load the data due to limited access.
Thus we use LiquiBase for this process.

We'd much rather avoid manual processing of SQL export files; it is time-consuming, involving ten of thousands of lines.

We have located the offending lines [here](https://github.com/liquibase/liquibase/blob/master/liquibase-standard/src/main/java/liquibase/sqlgenerator/core/InsertGenerator.java#L70)

```java
            if ((newValue == null) || "NULL".equalsIgnoreCase(newValue.toString())) {
                sql.append("NULL");
            }
```

These lines must be turned into

```java
            if (newValue == null) {
                sql.append("NULL");
            } else if ("NULL".equalsIgnoreCase(newValue.toString())) {
                sql.append("'").append(newValue).append("'");
           }
```

Will create a pull request next.

### Steps To Reproduce

Create a table with a VARCHAR2(4) column.
Fill in some values into that column, e.g. "null", "NULL", "nuLL".

Run a generateChangeLog on this table.

```sql
create table NULLS_HERE (
    type_ varchar2 (4) NOT NULL
);

INSERT INTO TABLE_NULLS_HERE (type_) VALUES ('null');
INSERT INTO TABLE_NULLS_HERE (type_) VALUES ('NULL');
INSERT INTO TABLE_NULLS_HERE (type_) VALUES ('nuLL');
INSERT INTO TABLE_NULLS_HERE (type_) VALUES ('NUll');
```

### Expected/Desired Behavior

**Expected**
The export must be exactly the one outlined above.

**Effective**

Currently we get this:

```sql
INSERT INTO TABLE_NULLS_HERE (type_) VALUES (NULL);
INSERT INTO TABLE_NULLS_HERE (type_) VALUES (NULL);
INSERT INTO TABLE_NULLS_HERE (type_) VALUES (NULL);
INSERT INTO TABLE_NULLS_HERE (type_) VALUES (NULL);
```

### Liquibase Version

4.29.2

### Database Vendor & Version

Oracle / PostgreSQL

### Liquibase Integration

CLI

### Liquibase Extensions

_No response_

Linux / macOS

### Additional Context

_No response_

catull:
## Impact

## Description

Fixes #6504.

If a string column has the value "null" or "NULL", that column is exported as NULL when exported to SQL.


liquibase-standard/src/main/java/liquibase/sqlgenerator/core/InsertGenerator.java
```diff
@@ -67,8 +68,10 @@ public void generateValues(StringBuilder sql,InsertStatement statement, Database
 
         for (String column : statement.getColumnValues().keySet()) {
             Object newValue = statement.getColumnValues().get(column);
-            if ((newValue == null) || "NULL".equalsIgnoreCase(newValue.toString())) {
+            if (newValue == null) {
```
MalloD12, line 71: Hi @catull,

I've been testing this, and I see that it does not seem to work in all cases. For example, for enumeration types, this wouldn't be valid because it will try to insert a value other than the ones valid for that given enumeration. Not sure, if there is any other type this could also affect, that I haven't realized yet.

liquibase-integration-tests/src/test/groovy/liquibase/diffchangelog/DiffChangelogIntegrationTest.groovy
```diff
@@ -274,4 +274,36 @@ COMMENT ON COLUMN $viewName.$columnName IS '$columnComment';
         commandScope.addArgumentValue(ReferenceDbUrlConnectionCommandStep.REFERENCE_DATABASE_ARG, referenceDatabase)
         commandScope.execute()
     }
+
+    def "Treat 'null' / 'NULL' values not as NULLs"() {
+        given:
+        def outputChangelogFile = "test/classes/test-null.postgres.sql"
```
MalloD12, line 280: Let's avoid outputting a changelog file in SQL format until we fix #6396 (I have started working on that issue), but for now I would try using XML or any other format.

catull: My main motivation is changelog in SQL format.  I can wait for #6396.

liquibase-standard/src/main/java/liquibase/sqlgenerator/core/InsertGenerator.java
catull: Hmm, do you have a test case ?  I am not sure I can follow you.

liquibase-integration-tests/src/test/groovy/liquibase/diffchangelog/DiffChangelogIntegrationTest.groovy
MalloD12: You can ignore my previous comment because I've been trying to reproduce that issue and it seems to be working just fine now.

liquibase-standard/src/main/java/liquibase/sqlgenerator/core/InsertGenerator.java
MalloD12: Yeap, I faced this issue in the `included.changelog.xml` (which is under this path `changelogs/pgsql/complete/included.changelog.xml`). The execution of the below changeset:

```xml
 <changeSet id="6" author="molivasdat">
      <loadData file="datafiles/enum-data.csv" relativeToChangelogFile="true" tableName="sample_enum_table">
          <column header="int_col" name="int_col" type="NUMERIC"/>
          <column header="str_col" name="str_col" type="STRING"/>
          <column header="enum_col" name="enum_col" type="OTHER"/>
      </loadData>
</changeSet>
```
This was making some integration tests fail because of trying to insert a `"NULL"` into an enum type column.

catull: @MalloD12

Can you be a bit more specific ?
There are about ~20 references to `included.changelog.xml`.

I'd like to investigate the test failure you refer to.

Thanks.

catull: At any rate, I found the "bug".

`LoadDataChange#generateStatementsFromRows()` conflates `null` and `"NULL"`.
Both would be used as a key in a `Map`.

In order to disinguish `null` from the String `"NULL"`, I introduce a `Null` class.

Please review this, @MalloD12.

MalloD12: Hey @catull,

Are you able to see the this [execution details](https://github.com/liquibase/liquibase/actions/runs/12287878906/job/34393355812?pr=6506), you can find here what's currently failing.  If I'm not wrong, yes previously, were `PostgreSQLIntegrationTest` failing.

catull:
```
Error:    LoadDataOtherTest.testOtherColumnLoadData:46 expected:<NULL> but was:<null>
```

I do not get this error locally 
image

Same deal with

```
Error:    LoadDataChangeTest.NULL placeholder handling:809 Condition not satisfied:
```

Runs smoothly locally:

image

catull: In fact, the whole test class `LoadDataChangeTest` runs all tests successfully.

catull: image

catull: The problem with OpenJDK 8 is, I cannot even build it on macOS, never have been.

I perform a `mvn clean package`, and I always end up with this:

image

Even after I `rm -fr ~/.m2/repository/jakarta`, the build never succeeds, same result as just shown.
I have stopped using OpenJDK 8 with Liquibase, when I develop with it.
I can succesfully build Liquibase with OpenJDK 11 and higher.

I use the official distribution of Liquibase in production with OpenJDK 8, OpenJDK 17 and OpenJDK 21.

catull: Even this does not help at all: `mvn -DskipTests -Dmaven.javadoc.skip=true clean package`, same error.

catull: This statement is from `changelog.txt`:

```
- (#5810) Bump jakarta.enterprise:jakarta.enterprise.cdi-api from 4.0.1 to 4.1.0 @dependabot
```

That version is from March 2024, when the project `jakarta.enterprise.cdi-api` jumped from JDK 11 -> JDK 17.

So, which trick do you use to build Liquibase under JDK 8 ?

catull: Thanks, @MalloD12 

I had a look at a full test run under Fedora.
I believe I corrected them all now.

I get no errors now with `mvn -DtrimStackTrace=false -pl '!liquibase-cdi-jakarta' test`.

filipelautert:
Hey @catull , to build under java 8 and 11 we skip the jakarta module like this:

`./mvnw -P 'skip-integration-tests'  -pl '!liquibase-cdi-jakarta' clean test package`

liquibase-standard/src/main/java/liquibase/sqlgenerator/core/InsertGenerator.java
```diff
@@ -67,8 +68,10 @@ public void generateValues(StringBuilder sql,InsertStatement statement, Database
 
         for (String column : statement.getColumnValues().keySet()) {
             Object newValue = statement.getColumnValues().get(column);
-            if ((newValue == null) || "NULL".equalsIgnoreCase(newValue.toString())) {
+            if ((newValue == null) || newValue.toString() == null) {
```
MalloD12, line 71: If I'm not wrong, `newValue.toString() == null` could lead us to an NPE because newValue is not a `Null` instance, so if its value is `null`, the `toString()` call will cause an exception here. Shouldn't we update this line to:

```java 
if ((newValue == null) || new Null().equals(newValue)) 
```

MalloD12:
Hi @catull,

I just left you a review comment. I'm seeing a test failure for Postgres integration tests, but this is not related to your PR,  and it is a general issue I'm also facing in #6560.  

So, after resolving the current review comment we would be good to move on with this PR.

Thank you,
Daniel.

liquibase-standard/src/main/java/liquibase/sqlgenerator/core/InsertGenerator.java
MalloD12, line 71: I think the second part of this IF will never be true, I think it would be just if `newValue` were of `Null` type.

MalloD12:
Exactly, I was pointing out that in my first comment that later I deleted. Because, when  `newValue == null` would lead us to that NPE you also saw when calling the `toString()` method. Now with this change, don't you think it's just enough to ask for `if (newValue == null)`, because I think the `newValue instanceof null` doesn't add you any extra validation, it would be different if we do i`f (newValue instanceof Null)` (where `Null` is the type that you created in this PR), but that won't happen because this type is not something that will come from the DB.

catull:
> Exactly, I was pointing out that in my first comment that later I deleted. Because, when `newValue == null` would lead us to that NPE you also saw when calling the `toString()` method. Now with this change, don't you think it's just enough to ask for `if (newValue == null)`, because I think the `newValue instanceof null` doesn't add you any extra validation, it would be different if we do i`f (newValue instanceof Null)` (where `Null` is the type that you created in this PR), but that won't happen because this type is not something that will come from the DB.

You are right with your suggestion, @MalloD12.

I split the `if` into 2 seperate branches, to better count how many `null` and how many `Null` cases do occur.
It is a clear 552 : 0 for pure `null`.

You convinced me now, I am changing the logic it ccordingly.

Thanks for your patience.

catull:
@filipelautert @MalloD12 
Excuse me, why is this issue closed ?

catull:
Thanks @filipelautert 

I was able to retrieve the branch and rebase/merge master into it.

catull:
SInce I had many incorrect commit author information.
I backed the original branch to https://github.com/catull/liquibase/tree/fix-issue-6504-orig.

Then I created a new branch with just the modified files vs `master`.

There is no difference in content between
     https://github.com/catull/liquibase/tree/fix-issue-6504
and
    https://github.com/catull/liquibase/tree/fix-issue-6504-orig
.
The length of the history is different, though.

MalloD12 (approved):
Hi @catull,

This is looking good to me. I have made the following changes:

- I made the updated of replacing `if ((newValue == null) || "NULL".equalsIgnoreCase(newValue.toString()))` by `if ((newValue == null) || (newValue instanceof Null))` in some os the generators (InserOrUpdateGenerator. UpdateGenerator, and InsertOrUpdateGeneratorInformix).

- Update LoadDataChangeTest ->  `new Null()` calls with `Null.getInstance()`. 
 
The rest looks good! Thank you for the changes made here.

@filipelautert are you able to authorize the test execution for this PR? I can't, but I can see it builds successfully locally.

catull:
Thank you very much, @MalloD12, for your continuous support!

filipelautert (approved):

tati-qalified:
Hi @catull, thank you for your contribution!
I see that hsql and postgres tests are failing, could you look into it?
Thank you

filipelautert:
Hello @catull , did you see this comment here? https://github.com/liquibase/liquibase/pull/6506#issuecomment-3558962080 Seems they still failing.

PR #6519, opened 2024-11-14
Fix the XSD adding the nullPlaceholder attribute for loadData column

### LINKED ISSUE #6518: XML validation failure using nullPlaceholder column definition property for loadData
### Search first

### Description

The `nullPlaceholder` property added to fix #3627 is missing from the XSD.
This prevents proper validation of XML based on XSD,  so if liquibase is runi without `--validate-xml-changelog-files=false` it complains for
> Error parsing line 4 column 72 of foo.xml: cvc-complex-type.3.2.2: Attribute 'nullPlaceholder' is not allowed to appear in element 'column'.

### Steps To Reproduce

Validate the following changelog, i.e. at https://www.freeformatter.com/xml-validator-xsd.html 
or use it with `liquibase update` leaving xml validation enabled
```xml
<?xml version="1.0" encoding="UTF-8"?><databaseChangeLog xmlns="http://www.liquibase.org/xml/ns/dbchangelog" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" xsi:schemaLocation="http://www.liquibase.org/xml/ns/dbchangelog-ext http://www.liquibase.org/xml/ns/dbchangelog/dbchangelog-ext.xsd http://www.liquibase.org/xml/ns/dbchangelog http://www.liquibase.org/xml/ns/dbchangelog/dbchangelog-latest.xsd">
  <changeSet author="config-exporter" id="gxi4-2 (loadData tableName=MY)">
    <loadData usePreparedStatements="true" commentLineStartsWith="#" encoding="UTF-8" file="my.csv" quotchar="&quot;" separator="," tableName="MY" >
      <column header="ID" name="ID" type="NUMERIC" nullPlaceholder=""/>
    </loadData>
  </changeSet>
</databaseChangeLog>
```
It will complain for the _nullPlaceholder_ attribute not allowed.

### Expected/Desired Behavior

The XSD validation should pass also for an XML changelog like the one sketched above, that uses the `nullPlaceholder` attribute on `column` element of `loadData`.

### Liquibase Version

4.30

### Database Vendor & Version

_No response_

### Liquibase Integration

_No response_

### Liquibase Extensions

_No response_

_No response_

### Additional Context

In the meantime a workaround is using `--validate-xml-changelog-files=false` or similar. See https://docs.liquibase.com/parameters/validate-xml-changelog-files.html

davidecavestro:
## Impact

## Description

Fixes #6518
Fix the XSD to pass XML validation when using the `nullPlaceholder` on `column` definition for `loadData`.

## Things to be aware of

I simply added the missing attribute to the _latest_ XSD.
Not sure if the `nullPlaceholder` attribute - available from version 4.29 on - should also be added (retroactively) to the versioned XSD for 4.29 and 4.30.

## Things to worry about

The attribute is not mentioned on the docs. I can propose a PR for that. See https://github.com/davidecavestro/liquibase-docs/tree/add-loaddata-column-nullplaceholder


MalloD12:
Hi @davidecavestro,  

Thank you for your contribution to this PR. I have added an integration test to validate this change. Originally, I added the below two changesets:

```xml
<changeSet id="nullPlaceHolderTest_CreateTable" author="mallod">
    <createTable tableName="nullPlaceholderTable">
        <column name="colKey" type="int"/>
        <column name="col2" type="varchar(200)"/>
    </createTable>
</changeSet>

<changeSet id="nullPlaceHolderTest" author="mallod">
    <loadData tableName="nullPlaceholderTable"
              file="nullPlaceHolderTestData.csv"
              relativeToChangelogFile="true">
        <column name="test_col" type="COMPUTED" nullPlaceholder="testNullPlaceholder"/>
        <column name="other_col" type="varchar(100)"/>
    </loadData>
</changeSet>
```
to be executed as part of our integration tests, but some of those tests do not make any validation, we use some of them to validate if a given changeset can be deployed successfully or not. So, I changed that approach and created a test to run an update and then generate a changelog to see the value coming from a given column where I use the `nullPlacerholder` property. For that given column what I got was:

```xml
<column name=\"col2\"/> value="testNullPlaceholder"
```
But that's not what I was expecting, considering column value is `null` I was expecting to see:

```xml
<column name=\"col2\"/>"
```
This will change any time soon after we merge the next two PRs #6422 and #6506. Could you please look at this?

On the other hand, you don't need to worry about documenting this. I'll create an internal ticket for our docs team and they will take care of it. I'll point them to your docs commit, though.

MalloD12:
Hi @davidecavestro,

Yeah, it was a wrong copy-paste I made when using an existing `loadData` changeset I had from one other test. I fixed that, and I'm also seeing the same output you do:

```xml
<changeSet author="mallod (generated)" id="1733149128231-3">
        <insert tableName="nullPlaceholderTable">
            <column name="colKey" valueNumeric="2"/>
            <column name="col2" value="testNullPlaceholder"/>
        </insert>
        <insert tableName="nullPlaceholderTable">
            <column name="colKey" valueNumeric="5"/>
            <column name="col2" value="there"/>
        </insert>
    </changeSet>
```

I'm still thinking it's not correct, if it would be translating the `nullPlaceHolder` into a `null` value then we should be getting something like this:

```xml
<column name=\"col2\"/>"
```

I'll push that `loadData` column name fix and uncomment the test, are you on Slack? Not sure what your time zone is, but maybe we can syn up on Slack and talk about this issue if you wish.

Thanks,
Daniel.

davidecavestro:
> I'll push that `loadData` column name fix and uncomment the test, are you on Slack? Not sure what your time zone is, but maybe we can syn up on Slack and talk about this issue if you wish.

Sure, I'm on Europe/Rome timezone, so now GMT+1.
That said, I've jsut pushed a fix as I'm pretty sure I found the issue: the `nullPlaceholder="testNullPlaceholder"` was set just for the _colKey_ column on the changeset, while the csv had the _testNullPlaceholder_ value just on the _col2_ col.
image

MalloD12:
Good catch, @davidecavestro! I set it up to the wrong column.  Thank you, for fixing it! 

I just added a minor extra assertion for the same column.

Thanks,
Daniel.

MalloD12 (approved):
Thank you, @davidecavestro for creating this PR to add this missing property in the XSD file, and for replying to all my inquiries very quickly.

I changed the test approach for the integration test added. We were getting different output locally that the one from GitHub. Instead of testing this change through a `generate-changelog` output we thought it would be clearer for our purpose (validate `null` value was set in replacement of the `nullPlaceHolder` value set) to do it performing an `update-sql` and validating its output.

Build and all tests are now successfully executed.

filipelautert (approved):

tati-qalified (approved):

rberezen (approved):

PR #6521, opened 2024-11-14
Fixes #3763 tagDatabase tag allows for duplicate tag values

### LINKED ISSUE #3763: tagDatabase tag allows for duplicate tag values
## Environment

**Liquibase Version**: 4.19.0

**Liquibase Integration & Version**: Gradle

**Liquibase Extension(s) & Version**: N/A

**Operating System Type & Version**: Mac OSX 13.1

**Infrastructure Type/Provider**: N/A

## Description

The `<tagDatabase>` tag allows the same tag value to be created with different authors. We use `<tagDatabase>` at the start of a release. If two developers are concurrently making schema changes it's easy enough for both to inadvertently include their `<tagDatabase>` tags at commit/merge, resulting in multiple tags of the same version in databasechangelog. This muddies the waters around rollback via tag.

## Steps To Reproduce

Example XML:
```
    <changeSet id="TAG-12.0" author="dev1">
        <tagDatabase tag="release/12.0"/>
    </changeSet>
    <changeSet id="TAG-12.0" author="dev2">
        <tagDatabase tag="release/12.0"/>
    </changeSet>
```

## Actual Behavior
In this simple example LB creates the tag twice. 

Running Changeset: db.changelog-v12.0.xml::TAG-12.0::dev1
[2023-02-01 11:12:35] INFO [liquibase.changelog] Tag 'release/12.0' applied to database
[2023-02-01 11:12:35] INFO [liquibase.changelog] ChangeSet db.changelog-v12.0.xml::TAG-12.0::dev1 ran successfully in 1ms
Running Changeset: db.changelog-v12.0.xml::TAG-12.0::dev2
[2023-02-01 11:12:35] INFO [liquibase.changelog] Tag 'release/12.0' applied to database
[2023-02-01 11:12:35] INFO [liquibase.changelog] ChangeSet db.changelog-v12.0.xml::TAG-12.0::dev2 ran successfully in 0ms
[2023-02-01 11:12:35] INFO [liquibase.lockservice] Successfully released change log lock

In a more complicated example, dev3 might have tagged the database, made other changes and then dev4 might have inadvertently added their own same tag and made their changes. 

## Expected/Desired Behavior

My feeling is that one shouldn't be able to tag the database with the same value, or minimally there should be some warning. I'm not sure what the value in allowing the current behavior would be.

Example of a more complex scenario this could impact.
image

## Additional Context
Thanks for your consideration.

manunin:
## Impact

## Description

Fixed duplicate tags in the **databasechangelog** table per recommendations from issue #3763:

- For the **update** command: Before applying a tag changeset, the system checks **databasechangelog** for any existing rows with the same tag. If found, the existing tag value is cleared.
- For the **tag** command: Before setting a tag on the last row, the system checks **databasechangelog** for any existing rows with the same tag. If found, the existing tag value is cleared.
- Updated tagging logic **TagDatabaseGenerator** for default databases: Rows are now selected based on the highest **DATEEXECUTED** and, subsequently, the highest **ORDEREXECUTED**.

## Things to be aware of
If we have to changesets, for example:

    <changeSet id="TAG-1" author="dev1">
        <tagDatabase tag="release/1.0"/>
    </changeSet>
    <changeSet id="TAG-1" author="dev2">
        <tagDatabase tag="release/1.0"/>
    </changeSet>

We will get two records in databasechangelog, but only one will have the tag field filled. In other words, one tagChangeset record will be without a tag

## Things to worry about
I've implemented **ChainableAbstractSqlGenerator** to handle the responsibility of advancing in the SqlGeneratorChain. Maybe it's an overhead.

### DIFF
--- a/liquibase-integration-tests/src/test/groovy/liquibase/command/core/TagIntegrationTest.groovy
+++ b/liquibase-integration-tests/src/test/groovy/liquibase/command/core/TagIntegrationTest.groovy
@@ -1,0 +1,73 @@
+package liquibase.command.core
+
+import liquibase.Scope
+import liquibase.command.CommandScope
+import liquibase.command.core.helpers.DbUrlConnectionArgumentsCommandStep
+import liquibase.command.util.CommandUtil
+import liquibase.extension.testing.testsystem.DatabaseTestSystem
+import liquibase.extension.testing.testsystem.TestSystemFactory
+import liquibase.extension.testing.testsystem.spock.LiquibaseIntegrationTest
+import spock.lang.Shared
+import spock.lang.Specification
+
+@LiquibaseIntegrationTest
+class TagIntegrationTest extends Specification {
+
+    @Shared
+    private DatabaseTestSystem h2 = (DatabaseTestSystem) Scope.getCurrentScope().getSingleton(TestSystemFactory.class).getTestSystem("h2")
+
+    def "validate update command with duplicated tag changes is successfully executed with only one tag applied"() {
+        when:
+        def updateCommand = new CommandScope(UpdateCommandStep.COMMAND_NAME)
+        updateCommand.addArgumentValue(DbUrlConnectionArgumentsCommandStep.URL_ARG, h2.getConnectionUrl())
+        updateCommand.addArgumentValue(DbUrlConnectionArgumentsCommandStep.USERNAME_ARG, h2.getUsername())
+        updateCommand.addArgumentValue(DbUrlConnectionArgumentsCommandStep.PASSWORD_ARG, h2.getPassword())
+        updateCommand.addArgumentValue(UpdateSqlCommandStep.CHANGELOG_FILE_ARG, "liquibase/tag-changelog.xml")
+        updateCommand.execute()
+
+        then:
+        def detailsResultSet1 = h2.getConnection().createStatement().executeQuery("select count(*) from databasechangelog where tag = 'release/1.0'")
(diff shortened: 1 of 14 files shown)


MalloD12:
Hi @manunin,

In the name of the Liquibase team, I want to thank you for contributing to fixing #3763. I've been looking at this PR (no super in detail), and it looks good to me in general. Unfortunately, I don't think we will be able to move forward with it because it could be a breaking change for some people. As we know there could be duplicate tags, and in cases like a `rollback` command execution cases like this could be handled by using:

 `--tag-version=<string>`  |  Tag version to use for multiple occurrences of a tag. Valid values are NEWEST and OLDEST. If set to NEWEST, Liquibase rolls back changes up to the most recent instance of the tag. Any older instances of the tag remain in the DATABASECHANGELOG table. If set to OLDEST, Liquibase rolls back changes up to the first instance of the tag. **Default**: OLDEST

Moving forward with this PR would mean a breaking change for the existing logic, and that could affect many users. Again, I think your proposed changes make total sense to me, but I think we should consider a change like this on a major version (Liquibase 5.x). We have already started building a road map for it, so I think this is one more item we need to include there as something we need to refactor and implement probably in the same or similar way that you have done here.  

Hope you can understand our concern here.

Thanks,
Daniel.

MalloD12:
You can close it yourself, or I can do it if you wish. We love community contributions, but sometimes, in cases like this, when changes are modifying current behavior, it might be worth first discussing the issue we are trying to address to avoid making changes that would break current use of some functionality the wider community might be using.

Thanks,
Daniel.

PR #6536, opened 2024-11-22
Implement copy method in FilesystemProjectCopier (DAT-18890)

abrackx:
## Impact


wwillard7800 (approved):

liquibase-standard/src/main/java/liquibase/command/copy/FileSystemProjectCopier.java
```diff
@@ -59,6 +60,11 @@ public File createWorkingStorage(String projectDir, boolean keepTempFiles) {
      */
     @Override
     public void copy(String source, String target, boolean recursive) {
-        throw new UnexpectedLiquibaseException("The command 'init copy' requires s3:// paths and cannot be used with local file system paths. Learn more at https://docs.liquibase.com/commands/init/copy.html");
+        try {
+            String contents = FileUtil.getContents(new File(source));
+            FileUtil.write(contents, new File(target));
```
StevenMassaro, line 65: This is not going to work for large files. Can you use the apache commons method called copyLarge? https://commons.apache.org/proper/commons-io/apidocs/org/apache/commons/io/IOUtils.html#copyLarge(java.io.InputStream,java.io.OutputStream,byte%5B%5D)

StevenMassaro (changes requested):

liquibase-standard/src/main/java/liquibase/command/copy/FileSystemProjectCopier.java
abrackx: Good catch I'm sure this will work!

MalloD12 (approved):

StevenMassaro (approved):

Issue #6539, opened 2024-11-25
LoadDataChange and SpringResourceAccessor can raise NPE for CLOB values


sapo-di:
### Search first

- [x] I searched and no similar issues were found

### Description

I am receiving the following exception after updating from liquibase 4.26.0 to 4.30.0:

<details>
<summary>Stacktrace</summary>

```
org.springframework.beans.factory.BeanCreationException: Error creating bean with name 'entityManagerFactory' defined in class path resource [org/springframework/boot/autoconfigure/orm/jpa/HibernateJpaConfiguration.class]: Failed to initialize dependency 'liquibase' of LoadTimeWeaverAware bean 'entityManagerFactory': Error creating bean with name 'liquibase' defined in class path resource [org/example/db_provider/configuration/LiquibaseConfig.class]: liquibase.exception.CommandExecutionException: liquibase.exception.LiquibaseException: liquibase.exception.MigrationFailedException: Migration failed for changeset my-changeset:
     Reason: java.lang.RuntimeException: java.lang.NullPointerException: Cannot invoke "String.startsWith(String)" because "relativePath" is null
        at org.springframework.beans.factory.support.AbstractBeanFactory.doGetBean(AbstractBeanFactory.java:326)
        at org.springframework.beans.factory.support.AbstractBeanFactory.getBean(AbstractBeanFactory.java:205)
        at org.springframework.context.support.AbstractApplicationContext.finishBeanFactoryInitialization(AbstractApplicationContext.java:954)
        at org.springframework.context.support.AbstractApplicationContext.refresh(AbstractApplicationContext.java:625)
        at org.springframework.boot.web.servlet.context.ServletWebServerApplicationContext.refresh(ServletWebServerApplicationContext.java:146)
        at org.springframework.boot.SpringApplication.refresh(SpringApplication.java:754)
        at org.springframework.boot.SpringApplication.refreshContext(SpringApplication.java:456)
        at org.springframework.boot.SpringApplication.run(SpringApplication.java:335)
        at org.springframework.boot.SpringApplication.run(SpringApplication.java:1363)
        at org.springframework.boot.SpringApplication.run(SpringApplication.java:1352)
        at org.example.db_provider.DbProviderApp.main(DbProviderApp.java:33)
        at java.base/jdk.internal.reflect.NativeMethodAccessorImpl.invoke0(Native Method)
[...]

PR #6540, opened 2024-11-25
Fix null value handling when loading resources

sapo-di:
## Impact

## Description

Fix for #6539
Fixes NPE when accessing resources via `SpringResourceAccessor`.

### DIFF
--- a/liquibase-standard/src/main/java/liquibase/change/core/LoadDataChange.java
+++ b/liquibase-standard/src/main/java/liquibase/change/core/LoadDataChange.java
@@ -39,7 +39,6 @@ import liquibase.util.StreamUtil;
 import liquibase.util.StringUtil;
 import liquibase.util.csv.CSVReader;
 import lombok.Setter;
-import org.apache.commons.lang3.BooleanUtils;
 import org.apache.commons.lang3.StringUtils;
 
 import java.io.IOException;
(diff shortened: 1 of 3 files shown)


tati-qalified:
Hi @sapo-di, thank you for your contribution!

When manually testing this code with PostgreSQL I found that empty strings in general are handled as `NULL`, except for the case where the last column of the table is empty. There, it's handled as an empty string.

image

Could you look into this? Let us know if you need any guidance.

Thanks again,
Tatiana

sapo-di:
Hi!

I tried several ways to reproduce this. The target column I had the issue with was of type `TEXT`. I could nor reproduce the behaviour with a last column but I was able to reproduce the behaviour for varchat(255) columns. 

If your test data would run through this else-if block: 

```java
                        } else if (columnConfig.getTypeEnum() == LOAD_DATA_TYPE.STRING) {
                            valueConfig.setType(columnConfig.getType());
                            valueConfig.setValue(value == null ? "" : value);
```

that would not insert any `NULL` values. I don't know what the intended behaviour is.

MalloD12 (approved):
**Approved**.

Thank you, @sapo-di for this PR. The null check added logic added looks good to me. Build and test checks have been successfully executed. 

Daniel.

rberezen (approved):

tati-qalified (approved):
Tested manually, it seems to work as expected.
Thank you @sapo-di for your contribution!

PR #6555, opened 2024-12-02
Generate zip file via maven, add zip and tar integration tests (DAT-18855)

abrackx:
## Impact

## Description
Instead of creating the zip in the workflow, create it using the maven assembly. This means we can also test both the tar and the zip using maven. The addition of `failsafe:integration-test failsafe:verify` to the package step in the workflow runs the maven failsafe tests against the tar and zip generated from the package phase.

## Things to worry about
This changes how we generate the zip file.


.github/workflows/run-tests.yml
```diff
@@ -445,7 +445,7 @@ jobs:
 
   run-build-publish-file:
     needs: [ build_tests,integration-test ]
-    uses: liquibase/liquibase/.github/workflows/build.yml@master
+    uses: liquibase/liquibase/.github/workflows/build.yml@DAT-18855
```
abrackx, line 448:
```suggestion
    uses: liquibase/liquibase/.github/workflows/build.yml@master
```
Need to revert this to master before merging.

abrackx:
Successful build can be found [here](https://github.com/liquibase/liquibase/actions/runs/12127657700/job/33813205609).

MalloD12 (approved):
Tests seem alright to me. I think if time allows it I would add two additional tests for each test class for an empty dir case and one other negative case for a non-matching file scenario.

Rest looks good.

liquibase-dist/src/test/java/ZipIT.java
StevenMassaro, line 1: This test class duplicates a lot of code from TarIT

StevenMassaro (approved):
One minor comment

StevenMassaro:
Did you compare the zip from this build and the zip from the master build to make sure the contents are the same?

filipelautert (approved):
Looks, good, nice to get rid of some build lines.

abrackx:
> Did you compare the zip from this build and the zip from the master build to make sure the contents are the same?

Yes! Looks good to me, only diff found was versioning differences.
```Only in branch: .DS_Store
diff --color -bur master/GETTING_STARTED.txt branch/GETTING_STARTED.txt
--- master/GETTING_STARTED.txt	2024-12-03 13:12:44
+++ branch/GETTING_STARTED.txt	2024-12-02 21:50:06
@@ -1,9 +1,9 @@
-Liquibase Version master-SNAPSHOT
+Liquibase Version DAT-18855-SNAPSHOT
 
 
 Getting Started
 --------------------------------------------
-Thanks for downloading and installing Liquibase master-SNAPSHOT from Liquibase.
+Thanks for downloading and installing Liquibase DAT-18855-SNAPSHOT from Liquibase.
 
 By using the installer, ...

```

.github/workflows/build.yml
```diff
@@ -304,7 +272,7 @@ jobs:
         uses: actions/upload-artifact@v3
         with:
           name: liquibase-zip-${{ needs.setup.outputs.thisBranchName }}
-          path: liquibase-dist/target/liquibase-${{ needs.setup.outputs.thisBranchName }}-SNAPSHOT/*
+          path: artifacts/*.zip
```
abrackx, line 275: This is causing the zip to contain another directory, which is breaking functional tests.

abrackx: Fixed by extracting the tarball to a directory and uploading that instead. This is basically what happened before, without all the `diff`-ing.

StevenMassaro: That sucks

.github/workflows/run-tests.yml
abrackx, line 448:
```suggestion
    uses: liquibase/liquibase/.github/workflows/build.yml@master
```
Revert to master once ready to merge.

PR #6556, opened 2024-12-02
feat(DAT-19013): Enhance History Command to filter changesets by tags

filipelautert:
- Added new command arguments `tags` and `tag` to filter changesets by tags
- Refactored `HistoryCommandStep` to use `getOrUpdateReportPrinter` method for better code reuse
- Updated `HistoryCommandStepTest` to include tests for the new tag filtering functionality

## Impact

### DIFF
--- a/liquibase-standard/src/main/java/liquibase/command/core/HistoryCommandStep.java
+++ b/liquibase-standard/src/main/java/liquibase/command/core/HistoryCommandStep.java
@@ -26,8 +26,8 @@ public class HistoryCommandStep extends AbstractCommandStep {
 
     public static final CommandArgumentDefinition<HistoryFormat> FORMAT_ARG;
     public static final CommandArgumentDefinition<DateFormat> DATE_FORMAT_ARG;
-    public static final CommandArgumentDefinition<Boolean> TAGS_ARG;
-    public static final CommandArgumentDefinition<String> TAG_ARG;
+    public static final CommandArgumentDefinition<Boolean> SHOW_TAGS_ARG;
+    public static final CommandArgumentDefinition<String> TAG_FILTER_ARG;
     public static final CommandResultDefinition<DeploymentHistory> DEPLOYMENTS_RESULT;
 
     static {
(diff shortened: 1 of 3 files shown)


liquibase-standard/src/main/java/liquibase/command/core/HistoryCommandStep.java
```diff
@@ -37,6 +40,13 @@ public class HistoryCommandStep extends AbstractCommandStep {
                 .defaultValue(DateFormat.getDateTimeInstance(DateFormat.SHORT, DateFormat.SHORT), "Platform specific 'short' format")
                 .hidden()
                 .build();
+        TAGS_ARG = builder.argument("tags", Boolean.class)
```
MalloD12, line 43: These two command argument names seem pretty similar. I would rename the boolean one to something like "includeTaggedChangesets" or something like that. I would also add to its description that it will include any tagged changeset regardless of whether a tag filter is provided or not.

liquibase-standard/src/main/java/liquibase/command/core/HistoryCommandStep.java
```diff
@@ -37,6 +40,13 @@ public class HistoryCommandStep extends AbstractCommandStep {
...
+                .description("Include only tagged changesets")
+                .defaultValue(false)
+                .build();
+        TAG_ARG = builder.argument("tag", String.class)
```
MalloD12, line 47: One other question, considering this is being used as a filter. Should it be a single value or could it be comma separate value and contain multiple tags that we should filter by?

liquibase-standard/src/main/java/liquibase/command/core/HistoryCommandStep.java
filipelautert: @MalloD12 could we discuss it in the jira ticket https://datical.atlassian.net/browse/DAT-19013 so we keep the specification history there?

MalloD12: Sure

MalloD12 (approved):
LGTM now.

tati-qalified:
@filipelautert when there are multiple tagged changesets, the output shows each one in a different table instead of all of them in one table.

`showTags`:
image


`tagFilter=first,third`:
image

tati-qalified (approved):

PR #6558, opened 2024-12-03
Added stub method in to handle duplicates in sub-commands DAT-19049

wwillard7800:
This method allows commands to mark themselves as "stubs", which can be implemented to handle situations where an extension that provides the command does not exist.  Unfortunately, we have to do this because Pico does not allow duplicate sub-commands.

## Impact


liquibase-standard/src/main/java/liquibase/command/CommandFactory.java
```diff
@@ -184,6 +189,15 @@ public SortedSet<CommandDefinition> getCommands(boolean includeInternal) {
         return Collections.unmodifiableSortedSet(commands);
 
     }
+    private static String findKeyInMapIgnoreCase(String key, Map<String, String[]> map) {
```
filipelautert, line 192: What is the idea behind this `findKeyInMapIgnoreCase` ?

filipelautert (approved):

PR #6569, opened 2024-12-09
add changeset timeout

Ryuzaki1947:
## Impact

## Description
        
The aim is to provide custom changes timeout independent of underlying database. Here is a small code sample approach along with the points to consider:
For function-> ```liquibase.executor.AbstractExecutor.execute(liquibase.change.Change, java.util.List<liquibase.sql.visitor.SqlVisitor>)```
[https://github.com/liquibase/liquibase/blob/38cc3e90c49f99b582de177951bf9052a7ed9863/liquibase-standard/src/main/java/liquibase/executor/AbstractExecutor.java#L132](url)

1.  SQL statements are derived from the change and executed one by one in ```execute(statement, sqlVisitors);```
2.  Here we will maintain a timer and if the time passes , the timeout exception is thrown.
3.  While handling exceptions externally in the calling class, database.rollback() is triggered, hence partial changes of a changeset will be rolled back

Further discussions:
1. We can add an accessible KeyStore in a scope which is cleaned up after exiting scope, just like MDCObject map.
2. We can use this tag feature to add a tag in caller function(where we call this function specifically for triggering change-set while executing changelog).
3. If this function is called internally via different functionalities, we only want a timeout in a specific liquibase command(eg update),then this tag feature can help us.
4. For now , we make an assumption (which is true for now) that this function is called only when changeset has to be executed and nowhere else.
5. Timeout is statically declared, but we will move that in the logic of parameter parsing and set it here while creation of Executor object. The timeout will be approximately applied since the execution happens in main thread only, and if the per-statement function i.e. ```execute(statement, sqlVisitors);``` is running, while timeout passes, then we will wait till this execution completes, and timeout will happen when the time check is visited again.
    This can be mitigated in two ways
    -  Use multithreading, and whenever a statement is executed, we trigger it in one thread and the other thread checks the timer, there is a way to cancel PreparedStatement in this manner, and it can give us better resolution, but complex changes required throughout liquibase repo
    -  Add a statement timeout(query level timeout= Ts), so even if the slowest query is causing the transaction timeout to occur, the timeout will be delayed by atmost Ts time
    -  Close the connection when timeout happens

### DIFF
--- a/liquibase-standard/src/main/java/liquibase/executor/AbstractExecutor.java
+++ b/liquibase-standard/src/main/java/liquibase/executor/AbstractExecutor.java
@@ -22,6 +22,7 @@ import java.util.List;
 public abstract class AbstractExecutor implements Executor {
     protected Database database;
     protected ResourceAccessor resourceAccessor;
+    protected static final long TIMEOUT_MILLIS=2000;//2 sec timeout
 
     /**
      *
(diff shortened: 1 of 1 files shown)


MalloD12:
Hi @Ryuzaki1947,

I was thinking we could define an `update` command  argument for this:

```java
public static final CommandArgumentDefinition<Long> CHANGESET_EXECUTION_TIMEOUT_ARG;

    static {
        CommandBuilder builder = new CommandBuilder(COMMAND_NAME, LEGACY_COMMAND_NAME);
       ...
        CHANGESET_EXECUTION_TIMEOUT_ARG = builder.argument("changeSetExecutionTimeout", Long.class)
                .addAlias("changeSetTimeout")
                .description("....")
                .build();
    }
```

Then probably we can read this command argument, save it in the current scope, and use it during the changeset `execute()` method.

What do you think about it?

Thanks,
Daniel.

mariochampion:
hey @Ryuzaki1947 thanks for the initial work, but are you still interested in help to move this PR along? Let us know, please, and thanks.

filipelautert:
Hi @Ryuzaki1947, thanks for the contribution and the interesting idea around changeset timeouts.

After reviewing, there are several things that would need to be addressed before this could move forward:

- The timeout is hardcoded to 2 seconds with no way to configure it — this would break most real migrations
- The timeout only checks between statements, not during execution, so a single long-running statement would bypass it entirely
- The large comment block with design notes needs to be removed from production code
- No test coverage for the new behavior
- CLA is not signed

Since we haven't heard back in a while, we're closing this for now. If you'd like to revisit, feel free to reopen or submit a new PR with the above addressed. We'd suggest looking into `Statement.setQueryTimeout()` for per-statement timeouts and making the timeout configurable via changeset attributes or global configuration.

Thanks again for the effort!

Issue #6538, opened 2024-11-23
Changelog Lock Occasionally Not Released: Request for a Watchdog Mechanism to Refresh or Force Unlock Expired Locks


LucasInsight:
### Search first

- [x] I searched and no similar issues were found

### Description

Description
In certain scenarios, the Liquibase changelog lock is not released as expected due to issues such as process interruptions, network disconnections, or unexpected errors. While the releaseLocks command can resolve the problem manually, this is not an efficient solution in automated CI/CD pipelines or high-frequency deployment environments.

To address this, I propose implementing a "watchdog" mechanism that periodically checks and refreshes changelog locks. If a lock is found to be expired or inactive for a prolonged period, it should be forcefully released automatically to prevent blocking subsequent operations.

Use Case
Context: In large-scale team collaborations or continuous delivery environments, the Liquibase changelog lock is a critical mechanism to ensure the safety of database migrations. However, when locks are not released properly, they can disrupt workflows and delay deployments.
Problem: Relying on manual intervention with releaseLocks or external scripts is suboptimal, increasing maintenance overhead and reducing user experience.
Improvement Goal: Automate the process of monitoring, refreshing, and force-releasing locks to ensure seamless and reliable database operations.
Suggested Solution
Introduce a built-in watchdog feature that:

Periodically checks the status of changelog locks in the database.
Automatically refreshes active locks to maintain their validity.
Forcefully releases locks that exceed a defined expiration period or show no activity.
Benefits
Eliminates manual intervention for lock issues.
Reduces downtime and improves CI/CD workflow reliability.
Enhances user experience by providing a built-in solution for a common operational challenge.
Additional Information
This feature should be configurable, allowing users to set the lock expiration duration and enable/disable the watchdog functionality based on their needs.
A detailed audit log for lock refresh and release actions would also be valuable for debugging and monitoring.

### Steps To Reproduce

none

### Expected/Desired Behavior

none

### Liquibase Version

4.30.0

### Are you willing to submit a PR?

- [ ] I'm willing to submit a PR (Thank you!)

PR #6574, opened 2024-12-11
Implement Watchdog Mechanism for Changelog Lock Refresh (#6538)

LucasInsight:
Description
This pull request implements a Watchdog Mechanism for refreshing or forcibly unlocking expired changelog locks, addressing the issue described in [#6538](https://github.com/org/repo/issues/6538).

The feature was originally developed based on version 4.9.1 and has been verified to work as intended in that context. However, during the merge with the latest version (4.30.1), a crash was identified in the following file:
liquibase/Scope.java at line 230

Key Changes:
Introduced a Watchdog mechanism to:
Periodically refresh locks.
Handle force unlock for expired locks.
Added new configurations to control the watchdog behavior (e.g., refresh intervals, timeout thresholds).
Ensured backward compatibility with existing lock management.
Known Issue:
The current implementation causes a crash when integrated with version 4.30.1, specifically in the Scope.java file at line 230. Further investigation is required to identify the root cause and adapt the feature for the latest version.

Testing
Environment:
Verified functionality on version 4.9.1.
Test Cases:
Ensured locks are refreshed at regular intervals.
Validated force unlock for expired locks.
Confirmed no regression in existing lock management behavior.
Steps to Reproduce Crash in 4.30.1:
Merge the branch with version 4.30.1.
Execute lock-related operations that trigger the Watchdog mechanism.
Observe the crash at liquibase/Scope.java:230.

### DIFF
--- a/liquibase-core/src/main/java/liquibase/GlobalConfiguration.java
+++ b/liquibase-core/src/main/java/liquibase/GlobalConfiguration.java
@@ -19,6 +19,7 @@ public class GlobalConfiguration implements AutoloadedConfigurations {
     public static final ConfigurationDefinition<String> OUTPUT_LINE_SEPARATOR;
     public static final ConfigurationDefinition<String> OUTPUT_FILE_ENCODING;
     public static final ConfigurationDefinition<Charset> FILE_ENCODING;
+    public static final ConfigurationDefinition<Long> CHANGELOGLOCK_TIMEOUT_WITH_WATCHDOG;
     public static final ConfigurationDefinition<Long> CHANGELOGLOCK_WAIT_TIME;
     public static final ConfigurationDefinition<Long> CHANGELOGLOCK_POLL_RATE;
     public static final ConfigurationDefinition<Boolean> CONVERT_DATA_TYPES;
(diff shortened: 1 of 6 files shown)


lbadam:
@LucasInsight thanks for taking this issue on. This is definitely a pain point for some Liquibase users.

Can you elaborate more on how you maintain backwards compatibility? Once this feature is implemented, what is the user experience like?

LucasInsight:
> @LucasInsight thanks for taking this issue on. This is definitely a pain point for some Liquibase users.
> 
> Can you elaborate more on how you maintain backwards compatibility? Once this feature is implemented, what is the user experience like?

Maintaining Backwards Compatibility
To maintain backwards compatibility, when a program instance starts, Liquibase, after successfully acquiring the changelog lock, uses a watchdog mechanism to refresh the lock every 10 seconds, preventing the lock from being unintentionally released during execution.

User Experience
Once this feature is implemented:

During normal execution: The watchdog will periodically refresh the lock until the task completes and the lock is properly released.
When the lock becomes invalid: If, for any reason, the watchdog fails and the lock duration exceeds the default threshold of 1 minute, the lock will be considered expired and forcibly released to prevent it from remaining in an invalid state.

filipelautert:
Hi @LucasInsight , thanks for the PR!
I~m reviewing it, and would not the watchdog verification fail if Liquibase takes more than 60 seconds during an operation? As we are using the same executor/jdbc connection, the update from the watchdog thread would be waiting for that long running operation to finish, and then another process would believe it finished and would acquire the lock. Or did you see a different behavior?

LucasInsight:
> Hi @LucasInsight , thanks for the PR! I~m reviewing it, and would not the watchdog verification fail if Liquibase takes more than 60 seconds during an operation? As we are using the same executor/jdbc connection, the update from the watchdog thread would be waiting for that long running operation to finish, and then another process would believe it finished and would acquire the lock. Or did you see a different behavior?


You're absolutely correct, the watchdog process requires a separate JDBC connection to function properly. Without it, the watchdog thread would be blocked by the long-running Liquibase operation, leading to potential issues like the one you described.

PR #6593, opened 2024-12-19
expire analytics cache after 60 minutes (DAT-19278)

StevenMassaro:
## Impact

## Description

Long running applications do not ever refresh the cached analytics configuration data. This PR adds a 60 minute expiration policy to the cached data.

### DIFF
--- a/.github/workflows/run-functional-tests.yml
+++ b/.github/workflows/run-functional-tests.yml
@@ -13,7 +13,7 @@ on:
         required: false
         default: ''
         type: string
-  pull_request:
+  pull_request_target:
     types: [ labeled ]
 
 jobs:
--- a/.github/workflows/run-pro-build-and-test.yml
+++ b/.github/workflows/run-pro-build-and-test.yml
@@ -13,7 +13,7 @@ on:
         required: false
         default: ''
         type: string
-  pull_request:
+  pull_request_target:
     types: [labeled]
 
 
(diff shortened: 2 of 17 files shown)


ben-manes:
fwiw, a single element cache might view a caching library as too heavy-weight. If you needed a thread-safe, expiring `Supplier`, then you might consider mirroring the double-checked locking approach used in Guava's [Suppliers.memoizeWithExpiration(supplier)](https://github.com/google/guava/blob/cc2c5d3d6623fe66a969c29fcb422bf02fb57a1f/guava/src/com/google/common/base/Suppliers.java#L286-L346).

StevenMassaro:
> fwiw, a single element cache might view a caching library as too heavy-weight. If you needed a thread-safe, expiring `Supplier`, then you might consider mirroring the double-checked locking approach used in Guava's [Suppliers.memoizeWithExpiration(supplier)](https://github.com/google/guava/blob/cc2c5d3d6623fe66a969c29fcb422bf02fb57a1f/guava/src/com/google/common/base/Suppliers.java#L286-L346).

Thanks @ben-manes ! We had some plans of what we might do with caffeine going forward, which was part of the reason for bringing it in now. Unfortunately we still need to support Java 8, so it looks like that won't be an option. In any case, thanks for pointing out the Guava code, I'll review it.

filipelautert (approved):

ben-manes:
> Unfortunately we still need to support Java 8, so it looks like that won't be an option. 

Caffeine 2.x is Java 8 based if you’d like to use that. In v3 we switched from Unsafe to VarHandles, removed deprecations, etc but it’s mostly equivalent between those major versions thanks to a n exhaustive test suite.

abrackx (approved):
Nice! Can you remove the mention of caffeine in the pr?

PR #6597, opened 2024-12-23
Handle absence or older version of checks JAR DAT-19289

wwillard7800:
## Impact

### DIFF
--- a/liquibase-maven-plugin/src/main/java/org/liquibase/maven/plugins/AbstractLiquibaseChecksMojo.java
+++ b/liquibase-maven-plugin/src/main/java/org/liquibase/maven/plugins/AbstractLiquibaseChecksMojo.java
@@ -19,4 +19,13 @@ public abstract class AbstractLiquibaseChecksMojo extends AbstractLiquibaseMojo{
     public boolean databaseConnectionRequired() {
         return false;
     }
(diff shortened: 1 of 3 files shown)


liquibase-maven-plugin/src/main/java/org/liquibase/maven/plugins/AbstractLiquibaseChecksMojo.java
```diff
@@ -19,4 +19,13 @@ public abstract class AbstractLiquibaseChecksMojo extends AbstractLiquibaseMojo{
     public boolean databaseConnectionRequired() {
         return false;
     }
+    protected boolean doesMarkerClassExist() {
```
abrackx, line 22: Does this method exist somewhere else?

abrackx (approved):
Nice, looks good.

liquibase-maven-plugin/src/main/java/org/liquibase/maven/plugins/AbstractLiquibaseChecksMojo.java
wwillard7800: Something like it in the Checks repo

filipelautert (approved):

Issue #6611, opened 2025-01-04
Apply include & includeAll contextFilter attributes to property tags in included changeLogs


eichingertim:
### Search first

- [x] I searched and no similar issues were found

### Description

**Feature Request**:
include & includeAll context attribute should also be appled on property in the included changeLogs not just changeSets

**As Is**
According to https://docs.liquibase.com/concepts/changelogs/attributes/contexts.html:
> You can specify the context attribute in <[include](https://docs.liquibase.com/change-types/include.html)> or <[includeAll](https://docs.liquibase.com/change-types/includeall.html)> tags. If specified, the given context is added to all changesets in the included file(s).

So currently if a property is specified in the included changelog, the context is not applied and thus the property will still be included / created during execution. This causes unwanted behavior in the scenario when two changelogs are included which specify the same property but with different values. 

See steps to reproduce for an example.

As a workaround you can of course specify the context explicitly on the property, but IMO it is more convenient and closer to user expectation if the context applies to all tags in included changelog.

### Steps To Reproduce

In the below example if liquibase is ran with context set to apac, the property countries will still contain value "DE,FR" as the property is still created from the emea changelog, as the context only applies to changeSets.

Example:

master.xml:
```
<databaseChangeLog
    xmlns="http://www.liquibase.org/xml/ns/dbchangelog"
    xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
    xmlns:ext="http://www.liquibase.org/xml/ns/dbchangelog-ext"
    xmlns:pro="http://www.liquibase.org/xml/ns/pro"
    xsi:schemaLocation="http://www.liquibase.org/xml/ns/dbchangelog
        http://www.liquibase.org/xml/ns/dbchangelog/dbchangelog-latest.xsd
        http://www.liquibase.org/xml/ns/dbchangelog-ext
        http://www.liquibase.org/xml/ns/dbchangelog/dbchangelog-ext.xsd
        http://www.liquibase.org/xml/ns/pro
        http://www.liquibase.org/xml/ns/pro/liquibase-pro-latest.xsd">

	<include file="A.xml" context=emea />
        <include file="B.xml" context=apac />

</databaseChangeLog>

```

A.xml:
```
<databaseChangeLog
    xmlns="http://www.liquibase.org/xml/ns/dbchangelog"
    xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
[...]

PR #6612, opened 2025-01-05
[ISSUE-#6611] Implement contextFilter (context) of include tag should also apply for property tags

eichingertim:
## Impact

## Description

If the <include /> tag has a contextFilter / context set, then before this change it would only apply to changeSets of the included changeLog. 

After this change, the same logic should apply to <property /> tags of the included changeLog.
Important Changes are in lines 487-492 of liquibase-standard/src/main/java/liquibase/changelog/DatabaseChangeLog.java

For more details on the problem description, see #6611.

## Things to be aware of

Minimal invasive: Applying the context already during property load of the changeLog. Existing functionality used for getting the included contexts by querying the `getIncludeContextFilter()` of the changeLog during parsing.

## Things to worry about

I ignore contexts from include when the property already has a context set. 
-> To be discussed, if it would make sense to append it if present.

## Additional Context

n/a

### DIFF
--- a/liquibase-standard/src/main/java/liquibase/changelog/DatabaseChangeLog.java
+++ b/liquibase-standard/src/main/java/liquibase/changelog/DatabaseChangeLog.java
@@ -36,6 +36,7 @@ import liquibase.util.ExceptionUtil;
 import liquibase.util.FileUtil;
 import liquibase.util.StringUtil;
 import lombok.Getter;
+import lombok.Setter;
 import org.apache.commons.io.FilenameUtils;
 import org.apache.commons.lang3.StringUtils;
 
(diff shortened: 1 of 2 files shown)


suryaaki2:
Why is this a breaking change? Could someone explain?

eichingertim:
> Why is this a breaking change? Could someone explain?

Existing implementation would still consider a property in an included changelog, even when a contextFilter on include does not match execution contextFilter.

With this change, the property (changeLogParam) will not be considered anymore as well, if include contextFilter is not met.

At least for me that changes a behavior, but happy to update if more experienced liquibase devs would count it as non breaking.

eichingertim:
@filipelautert recreated the PR as the previous branch name was failing builds: https://github.com/liquibase/liquibase/pull/6677.

PR #6615, opened 2025-01-06
DAT-19258

filipelautert:
## Impact


suryaaki2:
We need this PR only if we are doing a patch release we are waiting to confirm with the custom about a patch release vs snapshot build

PR #6621, opened 2025-01-08
DAT-19082: Fix issue with losing changed datatype attributes when one datatype is an int and the other is not

abrackx:
## Impact

## Description
This was introduced in #6405. The type comparisons should be using their respective db objects.


liquibase-standard/src/main/java/liquibase/diff/compare/core/ColumnComparator.java
```diff
@@ -87,7 +87,7 @@ public ObjectDifferences findDifferences(DatabaseObject databaseObject1, Databas
         ObjectDifferences differences = chain.findDifferences(databaseObject1, databaseObject2, accordingTo, compareControl, exclude);
 
         DataType type1 = ((Column) databaseObject1).getType();
-        DataType type2 = ((Column) databaseObject1).getType();
+        DataType type2 = ((Column) databaseObject2).getType();
```
filipelautert, line 90: 🤦🏻‍♂️

filipelautert (approved):

StevenMassaro (approved):
Nice catch

PR #6626, opened 2025-01-09
DAT-18331 PRO: integrate checks extension into tarball/zip, not mac os dmg installer

jandroav:
This pull request includes several changes to the GitHub Actions workflows for building and releasing Liquibase extensions. The most significant updates involve standardizing string quotes, adding new build steps for `liquibase-checks`, and commenting out the OWASP scanner and file attachment steps in the release workflow.

### Standardization and Formatting:
* [`.github/workflows/build-extension-jars.yml`](diffhunk://#diff-dc83207379251c4eeafe63714db8ce416c176c525edf14a8bd6ad57ca0e9027bL7-R42): Changed single quotes to double quotes for consistency in the description fields and other string values. [[1]](diffhunk://#diff-dc83207379251c4eeafe63714db8ce416c176c525edf14a8bd6ad57ca0e9027bL7-R42) [[2]](diffhunk://#diff-dc83207379251c4eeafe63714db8ce416c176c525edf14a8bd6ad57ca0e9027bL71-R73) [[3]](diffhunk://#diff-dc83207379251c4eeafe63714db8ce416c176c525edf14a8bd6ad57ca0e9027bL86-R97) [[4]](diffhunk://#diff-dc83207379251c4eeafe63714db8ce416c176c525edf14a8bd6ad57ca0e9027bL127-R136)

### New Build Steps:
* [`.github/workflows/build-extension-jars.yml`](diffhunk://#diff-dc83207379251c4eeafe63714db8ce416c176c525edf14a8bd6ad57ca0e9027bL86-R97): Added a new job `build-liquibase-checks` to handle the build process for `liquibase-checks` if it is included in the inputs.
* [`.github/workflows/build-extension-jars.yml`](diffhunk://#diff-dc83207379251c4eeafe63714db8ce416c176c525edf14a8bd6ad57ca0e9027bR245): Updated the extension build steps to skip `liquibase-checks` during the checkout and build process. [[1]](diffhunk://#diff-dc83207379251c4eeafe63714db8ce416c176c525edf14a8bd6ad57ca0e9027bR245) [[2]](diffhunk://#diff-dc83207379251c4eeafe63714db8ce416c176c525edf14a8bd6ad57ca0e9027bR278)

### Workflow Dispatch Changes:
* [`.github/workflows/create-release.yml`](diffhunk://#diff-8ac33fe295df086d3e55df1eb01194819d34cc0f5f54076dba96ceac0e40bd79L60-R60): Added `liquibase-checks` to the list of extensions to be released.

### Commented Out Steps:
* [`.github/workflows/create-release.yml`](diffhunk://#diff-8ac33fe295df086d3e55df1eb01194819d34cc0f5f54076dba96ceac0e40bd79L77-R94): Commented out the OWASP scanner job and the steps for attaching files to draft releases and extracting URLs for dry-run releases. [[1]](diffhunk://#diff-8ac33fe295df086d3e55df1eb01194819d34cc0f5f54076dba96ceac0e40bd79L77-R94) [[2]](diffhunk://#diff-8ac33fe295df086d3e55df1eb01194819d34cc0f5f54076dba96ceac0e40bd79L312-R350)

These changes aim to improve the build and release processes by ensuring consistency in string formatting, adding support for new extensions, and temporarily disabling certain steps to streamline the workflow.

### DIFF
--- a/.github/workflows/create-release.yml
+++ b/.github/workflows/create-release.yml
@@ -57,10 +57,10 @@ on:
 
 env:
   DEPENDENCIES: "liquibase-bigquery"
-  EXTENSIONS:  |
+  EXTENSIONS: |
     [
-      {"name": "liquibase-commercial-bigquery", "combined_jars": false},
-      {"name": "liquibase-checks", "combined_jars": true}
+      {"name": "liquibase-commercial-bigquery", "combined_jars": "false"},
+      {"name": "liquibase-checks", "combined_jars": "true"}
     ]
 
 jobs:
--- a/.github/workflows/create-release.yml
+++ b/.github/workflows/create-release.yml
@@ -95,7 +95,7 @@ jobs:
   
   build-extension-jars:
    needs: [ setup ]
-   uses: liquibase/liquibase/.github/workflows/build-extension-jars.yml@master
+   uses: liquibase/liquibase/.github/workflows/build-extension-jars.yml@DAT-18331
    with:
      liquibase-version: ${{ needs.setup.outputs.version }}
      dependencies: ${{ needs.setup.outputs.dependencies }}
--- a/.github/workflows/create-release.yml
+++ b/.github/workflows/create-release.yml
@@ -101,7 +101,7 @@ jobs:
      dependencies: ${{ needs.setup.outputs.dependencies }}
      extensions: ${{ needs.setup.outputs.extensions }}
      branch: ${{ needs.setup.outputs.branch }}
-  #  secrets: inherit
+     secrets: inherit
 
   # reversion:
   #   needs: [ setup, build-extension-jars ]
--- a/.github/workflows/create-release.yml
+++ b/.github/workflows/create-release.yml
@@ -57,7 +57,8 @@ on:
 
 env:
   DEPENDENCIES: "liquibase-bigquery"
-  EXTENSIONS: '[{"name":"liquibase-commercial-bigquery","combined_jars":"false"},{"name":"liquibase-checks","combined_jars":"true"}]'
+  EXTENSIONS: |
+    [{"name":"liquibase-commercial-bigquery","combined_jars":"false"},{"name":"liquibase-checks","combined_jars":"true"}]
 
 jobs:
   setup:
--- a/.github/workflows/create-release.yml
+++ b/.github/workflows/create-release.yml
@@ -101,7 +101,7 @@ jobs:
      dependencies: ${{ needs.setup.outputs.dependencies }}
      extensions: ${{ needs.setup.outputs.extensions }}
      branch: ${{ needs.setup.outputs.branch }}
-     secrets: inherit
+   secrets: inherit
 
   # reversion:
   #   needs: [ setup, build-extension-jars ]
--- a/.github/workflows/create-release.yml
+++ b/.github/workflows/create-release.yml
@@ -81,27 +81,26 @@ jobs:
   #     branch: ${{ needs.setup.outputs.branch }}
   #   secrets: inherit
 
-  # build-azure-uber-jar:
-  #  needs: [ setup ]
-  #  uses: liquibase/liquibase/.github/workflows/build-azure-uber-jar.yml@master
-  #  with:
-  #    branch: ${{ needs.setup.outputs.branch }}
-  #    liquibase-version: ${{ needs.setup.outputs.version }}
-  #  secrets: inherit
+  build-azure-uber-jar:
+   needs: [ setup ]
+   uses: liquibase/liquibase/.github/workflows/build-azure-uber-jar.yml@master
+   with:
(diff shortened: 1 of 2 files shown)


.github/workflows/create-release.yml
```diff
@@ -74,24 +74,24 @@ jobs:
...
-    uses: liquibase/build-logic/.github/workflows/owasp-scanner.yml@main
-    with:
-      branch: ${{ needs.setup.outputs.branch }}
-    secrets: inherit
+  # owasp-scanner:
+  #   needs: [ setup ]
+  #   uses: liquibase/build-logic/.github/workflows/owasp-scanner.yml@main
+  #   with:
+  #     branch: ${{ needs.setup.outputs.branch }}
+  #   secrets: inherit
```
StevenMassaro, line 82: Is this intended to be commented?

.github/workflows/create-release.yml
```diff
@@ -74,24 +74,24 @@ jobs:
...
    with:
      branch: ${{ needs.setup.outputs.branch }}
      liquibase-version: ${{ needs.setup.outputs.version }}
    secrets: inherit
   
   build-extension-jars:
-   needs: [ setup, owasp-scanner ]
-   uses: liquibase/liquibase/.github/workflows/build-extension-jars.yml@master
+   needs: [ setup ]
+   uses: liquibase/liquibase/.github/workflows/build-extension-jars.yml@DAT-18331
```
StevenMassaro, line 94: Does this need to be changed prior to merging?

.github/workflows/create-release.yml
```diff
@@ -115,7 +115,7 @@ jobs:
           token: ${{ secrets.BOT_TOKEN }}
 
       - name: Download liquibase-artifacts
-        uses: liquibase/action-download-artifact@v2-liquibase
+        uses: dawidd6/action-download-artifact@v7
```
StevenMassaro, line 118: What was the reason for changing this?

.github/workflows/create-release.yml
jandroav: @StevenMassaro we are testing stuff...

.github/workflows/create-release.yml
jandroav: yes

.github/workflows/create-release.yml
jandroav: because of this https://github.com/liquibase/liquibase/actions/runs/12690830290/job/35372586117#step:6:13 and because the reason we forked that action is no longer needed

StevenMassaro: That link you provided has this: `Run actions/download-artifact@v4`. Is that the correct link?

jandroav: oops, no sorry, this one https://github.com/liquibase/liquibase/actions/runs/12690390074/job/35371959175#step:4:1

filipelautert: We upgraded actions/upload to v4 around 1 month ago, and it changed the way files are uploaded. It was causing the error mentioned by Jandro - I even had to fix liquibase-sdk-maven-plugin because of that. So instead of fixing `liquibase/action-download-artifact`, we thought it would be better to test upstream again and see if it works now.

StevenMassaro: Okay. I am only a bit apprehensive about this because it doesn't seem to fit the spirit of this PR. Last time we made this change, which admittedly was years ago, we suffered inconsistent builds for a long time due to the issue with GitHub's API not returning consistent results. If that has been fixed, then I'm okay with trying this, but I'd prefer we do it separately from this PR (if possible) so we can monitor it more closely. We should not just shove this in without telling anyone, especially QA.

StevenMassaro: Additionally, this change introduces new risk with regards to the API rate limit, which was one of the main motivators for the forked action. Now, this might not really be an issue any longer since we're primarily obtaining artifacts using GitHub package manager (which we were not using at the time we forked that action), but it is something to note.

jandroav: Using `liquibase/action-download-artifact@v2-liquibase` blocks the spirit of this PR because the issue prevents us to fully test the liquibase tar balls generation with the `liquibase-checks` extension. If you want I am ok if we create another ticket which will be a blocker for https://datical.atlassian.net/browse/DAT-18331

filipelautert: @StevenMassaro one thing to note is that this action now is only used during release process - in the past it was used during the entire build. In fact I think we can even replace it by actions/download, that is used everywhere in our build. Maybe we should create a new devops ticket to follow it up?

StevenMassaro: If it blocks this work, then I don't think another ticket is necessary.

StevenMassaro (approved):

filipelautert (approved):

Issue #3313, opened 2022-09-27
referencedTableSchemaName is ignored


PotcFdk:
## Environment

**Liquibase Version**: 4.16.1

**Liquibase Integration & Version**: CLI

**Liquibase Extension(s) & Version**: 

**Database Vendor & Version**: Postgres 14.5-1

**Operating System Type & Version**: official docker image

**Infrastructure Type/Provider**: VM

## Description

In an `addColumn` change, `referencedTableSchemaName` for foreign key constraints is ignored.
This used to work in version 4.9.1 if I remember correctly, but does not work in 4.16.1. I am not yet sure when it stopped working.

## Steps To Reproduce

```yaml
databaseChangeLog:
- changeSet:
[...]
    changes:
    - addColumn:
        schemaName: example
        tableName: ex_table
        columns:
        - column:
            name: object_id
            type: UUID
            constraints:
                foreignKeyName: ex_table_object_fk
                referencedTableName: objects
                referencedTableSchemaName: example
                referencedColumnNames: id
```

## Actual Behavior

```sql
ALTER TABLE example.ex_table ADD CONSTRAINT ex_table_object_fk FOREIGN KEY (object_id) REFERENCES objects (id);
```

## Expected/Desired Behavior

```sql
ALTER TABLE example.ex_table ADD CONSTRAINT ex_table_object_fk FOREIGN KEY (object_id) REFERENCES example.objects (id);
```

PotcFdk:
**Update**: The working case I was referring to was in `createTable`, this one is in `addColumn`, I guess this is the difference that actually matters. I hope this helps with finding the bug.

FBurguer:
Thanks for submitting this issue! I could reproduce it in my environment. We will add it to the list of bugs to fix

BearsPunch:
after two years no fix yet?

Xstoudi:
I confirm it's still a problem.

MalloD12:
Hi @Xstoudi,

Could you please help me to reproduce this issue? I cannot reproduce this issue without your fix. I guess I might be not following the right steps. See, I'm using the below changelog:

```xml
databaseChangeLog:
  - changeSet:
      id: add_column_change
      author: mallod
      changes:
        - addForeignKeyConstraint:
            baseColumnNames: person_id
            baseTableCatalogName: cat
            baseTableName: address
            baseTableSchemaName: public
            constraintName: fk_address_person
            referencedColumnNames: id
            referencedTableName: person
[...]

PR #6639, opened 2025-01-13
fix: prefix table name with schema name when adding column constraint

Xstoudi:
Fixes: #3313

## Impact

## Description
As described in #3313, using `referencedTableSchemaName` on `ForeignKeyConstraint` didn't have any effect on the generated sql statement.
Instead, it was possible to pass a schema name prefixing `referencedTableName` with `my_schema.`

The change made on this PR ensure `referencedTableSchemaName` is used as a priority, but fallback on prefix syntax on `referencedTableName` if `null`.

## Things to be aware of

I kept the current behavior (prefix syntax) while fixing the issue : `referencedTableSchemaName` being simply ignored.

## Things to worry about

It's my first contribution, Liquibase is a huge codebase so please ensure it's made correctly and do not hesistate to provide geedback.

### DIFF
--- a/liquibase-standard/src/main/java/liquibase/sqlgenerator/core/AddColumnGenerator.java
+++ b/liquibase-standard/src/main/java/liquibase/sqlgenerator/core/AddColumnGenerator.java
@@ -269,11 +269,12 @@ public class AddColumnGenerator extends AbstractSqlGenerator<AddColumnStatement>
                     refTableName = referencesMatcher.group(1);
                     refColName = referencesMatcher.group(2);
                 } else {
+                    refSchemaName = ((ForeignKeyConstraint) constraint).getReferencedTableSchemaName();
                     refTableName = ((ForeignKeyConstraint) constraint).getReferencedTableName();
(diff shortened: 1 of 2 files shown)


tati-qalified:
Hi @Xstoudi, thank you for your contribution!
I've been testing this manually with MSSQL and it doesn't seem to work with `addColumn`. 
Here's the output:
```
ALTER TABLE example.ex_table_2 ADD CONSTRAINT ex_table_object_fk FOREIGN KEY (object_id) REFERENCES objects (id);
```
It does work with `createTable` and `addForeignKeyConstraint`.
Could you look into it?

Thank you,
Tatiana

MalloD12:
@tati-qalified the scenario you mentioned [here](https://github.com/liquibase/liquibase/pull/6639#issuecomment-2659621452) should be fixed now. If you want to try it again.

@Xstoudi, I think the issue that Tatiana mentioned is fixed, and I also added a generic test to validate that scenario against different DBs, as well as made a little change to the assertion you added in the test.

Thanks,
Daniel.

tati-qalified (approved):
Retested, it works as expected now. 
cc @MalloD12

PR #6642, opened 2025-01-14
DAT-19398 DevOps :: Fix Liquibase Checks versioning in the tarball to match the Liquibase Checks repository

jandroav:
This pull request includes a new job to extract the version of `liquibase-checks` from its `pom.xml` and updates the existing build job to depend on this new job.

Changes to the `.github/workflows/build-extension-jars.yml` file:

* Added a new job `get-liquibase-checks-version` to extract the version of `liquibase-checks` from its `pom.xml` file. This job runs only if `liquibase-checks` is included in the inputs and sets an output variable `version` with the extracted version.
* Updated the `build-liquibase-checks` job to depend on the new `get-liquibase-checks-version` job. This ensures that the version is extracted before the build process begins.

### DIFF
--- a/.github/workflows/build-extension-jars.yml
+++ b/.github/workflows/build-extension-jars.yml
@@ -111,7 +111,7 @@ jobs:
       uses: liquibase/build-logic/.github/workflows/publish-for-liquibase.yml@main
       with:
           repository: liquibase/liquibase-checks
-          version: ${{ inputs.liquibase-version }}
+          version: ${{ needs.get-liquibase-checks-version.outputs.version }}
       secrets: inherit
 
   build-and-deploy-extensions:
--- a/.github/workflows/build-extension-jars.yml
+++ b/.github/workflows/build-extension-jars.yml
@@ -102,8 +102,8 @@ jobs:
             name: Extract version from pom.xml
             shell: bash
             run: |
-              VERSION=$(grep -A1 '<artifactId>liquibase-checks</artifactId>' pom.xml | grep '<version>' | sed 's/.*<version>\(.*\)-SNAPSHOT<\/version>.*/\1/')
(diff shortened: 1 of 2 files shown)


.github/workflows/build-extension-jars.yml
```diff
@@ -87,17 +87,47 @@ jobs:
...
+          - uses: actions/checkout@v4
+            with:
+              repository: liquibase/liquibase-checks
+              token: ${{ secrets.BOT_TOKEN }}
+
+          - id: extract-version
+            name: Extract version from pom.xml
+            shell: bash
+            run: |
+                VERSION=$(grep '<version>' pom.xml | head -n 1 | sed 's/.*<version>\(.*\)-SNAPSHOT<\/version>.*/\1/')
```
filipelautert, line 105: Instead of the grep command we could use maven:
`mvn help:evaluate -Dexpression=project.version -q -DforceStdout`
But I guess grep will always work too.

filipelautert (approved):
Many more ifs in the build script :(

.github/workflows/build-extension-jars.yml
jandroav: Does that command remove the snapshot suffix from the pom's artifact verision?

jandroav:
> Many more ifs in the build script :(

yes...we could modify this structure: https://github.com/liquibase/liquibase/blob/1b6a5b04faa19e129274a06c29966338661ff923/.github/workflows/create-release.yml#L60 to something like

```
EXTENSIONS: |
[{"name": "liquibase-commercial-bigquery", "version":"liquibase"},
{"name": "liquibase-checks", "version":"custom"}]
```

but I will involve changing a lot of stuff and that for today is not possible

PR #6658, opened 2025-01-20
Code review for extensions after 4.31.0 release

filipelautert:
* undeprecate liquibase.exception.PreconditionFailedException#PreconditionFailedException(liquibase.precondition.FailedPrecondition) -> a community contributor added a new method that receives a throwable and deprecated this one, but for neo4j (and maybe other extensions) it still makes sense to have the old method.
* remove parameter validations in RawParameterizedSqlGenerator.java : we removed the validations from the executor class but they still exists in this generator. It`s interesting that they way we use RawParametrized in core don't trigger them, but Neo4j does.


filipelautert:
cc: @fbiville

MalloD12 (approved):
LGTM

fbiville (approved):

PR #6672, opened 2025-01-24
Enhance error message in case of lexical error DAT-18862

wwillard7800:
```Example output:

ERROR: Exception Details
ERROR: Exception Primary Class:  TokenMgrError
ERROR: Exception Primary Reason:  Lexical error at line 2, column 17.  Encountered: '\ufffd' (65533),
ERROR: Exception Primary Source:  PostgreSQL 16.0

Unexpected error running Liquibase: Migration failed for changeset formatted.sql::2::postgres:
     Reason: liquibase.exception.UnexpectedLiquibaseException: formatted.sql::2::postgres (issue at line 14 of lines 13-19)
  - Caused by: Lexical error at line 2, column 17.  Encountered: '\ufffd' (65533),
```

## Impact


filipelautert (approved):

liquibase-standard/src/main/java/liquibase/change/core/RawSQLChange.java
```diff
@@ -23,7 +24,15 @@ public class RawSQLChange extends AbstractSQLChange {
...
     private Boolean rerunnable;
-    
+
+    @Getter
+    @Setter
+    private Integer sqlStartLine;
+
+    @Getter
+    @Setter
+    private Integer sqlEndLine;
```
StevenMassaro, line 34: Will this influence the way checksums are calculated?

liquibase-standard/src/main/java/liquibase/util/StandardSqlParser.java
```diff
@@ -48,17 +53,52 @@ public StringClauses parse(String sqlBlock, boolean preserveWhitespace, boolean
...
+     * @param  e             The thrown exception
+     * @return String        The message to display
+     *
+     */
+    private static String enhanceExceptionMessage(ChangeSet changeSet, Change change, Throwable e) {
+        if (! (change instanceof RawSQLChange)) {
+            return changeSet.toString();
+        }
+        String exceptionMessage = e.getMessage();
+        Pattern p = Pattern.compile("(?i).* line ([\\d]+).*");
```
StevenMassaro, line 84: I thought you weren't going to try to parse out the line number from the exception message? Or was it simple enough that you figured we might as well try this first?

liquibase-standard/src/main/java/liquibase/util/StandardSqlParser.java
```diff
@@ -48,17 +53,52 @@ public StringClauses parse(String sqlBlock, boolean preserveWhitespace, boolean
                 }
                 token = t.getNextToken();
             }
-
         } catch (Throwable e) {
             if (changeSet != null) {
-                throw new UnexpectedLiquibaseException(changeSet.toString(), e);
+                Change change = Scope.getCurrentScope().get(ChangeSet.CHANGE_KEY, Change.class);
+                String message = enhanceExceptionMessage(changeSet, change, e);
```
StevenMassaro, line 59: I think the call to enhanceExceptionMessage should itself be inside a try catch. We don't want to fail on the exception processing while trying to enhance the message. If we can't enhance the message, just use the existing message.

StevenMassaro (changes requested):
This is a pretty neat enhancement.

liquibase-standard/src/main/java/liquibase/util/StandardSqlParser.java
wwillard7800: Yes, I thought it was straight-forward enough to try to match on it.  Worse case is that we can still display the range of lines from the changelog.

liquibase-standard/src/main/java/liquibase/util/StandardSqlParser.java
wwillard7800: I agree.

liquibase-standard/src/main/java/liquibase/util/StandardSqlParser.java
StevenMassaro: This is a good approach. Postgres will do a better job of identifying the line number than we can likely do.

liquibase-standard/src/main/java/liquibase/change/core/RawSQLChange.java
wwillard7800: No, the attributes aren't used in the calculation, just the generated SQL for the change.

StevenMassaro (approved):

filipelautert (approved):

PR #6675, opened 2025-01-27
recursively check for overrides of the override command step (DAT-19485)

StevenMassaro:
## Impact

## Description

This change allows multiple layers of command overrides. For example, there may be a command override in a database extension, which overrides a Pro override, which overrides an OSS command.


abrackx (approved):
Not too complicated, nice!

filipelautert (approved):

PR #6682, opened 2025-01-28
✨ feat: DAT-19341:publish_assets_to_internal_nexus repo.liquibase.net

sayaliM0412:
This pull request includes changes to the `.github/workflows/release-published.yml` file to correct a parameter name and add a new job for publishing assets to an internal Nexus repository.

Parameter name correction:

* Changed `filename` to `fileName` in multiple steps to ensure consistency and correctness. [[1]](diffhunk://#diff-f7ca0660afb7c5d6e7220dee8a6db81db251b3c98818f4d76a3e4441b85c7a6dL132-R132) [[2]](diffhunk://#diff-f7ca0660afb7c5d6e7220dee8a6db81db251b3c98818f4d76a3e4441b85c7a6dL232-R232) [[3]](diffhunk://#diff-f7ca0660afb7c5d6e7220dee8a6db81db251b3c98818f4d76a3e4441b85c7a6dL491-R491)

New job addition:

* Added a new job `publish_assets_to_internal_nexus` to publish OSS released assets to `repo.liquibase.net`, including steps to download released assets and publish them to the internal Nexus repository.

### DIFF
--- a/.github/workflows/release-published.yml
+++ b/.github/workflows/release-published.yml
@@ -129,7 +129,7 @@ jobs:
         with:
           repository: "liquibase/liquibase"
           tag: "${{ needs.setup.outputs.tag }}"
-          filename: "*"
+          fileName: "*"
           out-file-path: "."
 
       - name: Set up Java for publishing to Maven Central Repository
@@ -229,7 +229,7 @@ jobs:
         with:
           repository: "liquibase/liquibase"
           tag: "${{ needs.setup.outputs.tag }}"
-          filename: "liquibase-additional*.zip"
+          fileName: "liquibase-additional*.zip"
           out-file-path: "."
 
       - name: Unpack javadoc files and upload to s3
@@ -488,7 +488,7 @@ jobs:
         with:
           repository: "liquibase/liquibase"
           releaseId: "${{ inputs.dry_run_release_id }}"
-          filename: "*"
+          fileName: "*"
           out-file-path: "."
 
       - name: Set up Java for publishing to Maven Central Repository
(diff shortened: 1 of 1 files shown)


StevenMassaro:
What is the reason for this change?

sayaliM0412:
> What is the reason for this change?

There were times, when we had overwritten the assets to the OSS release job. Also times when there were issues with hotfix and we needed the original assets and they werent available. The QA's and the devs had the assets locally that saved the day. Also, there have been an ask to have them stored somewhere for easy retrieval if need be.

filipelautert (approved):

.github/workflows/release-published.yml
```diff
@@ -538,4 +538,36 @@ jobs:
...
+          for file in ./${{ needs.setup.outputs.version }}-released-assets/*; do
+            mvn deploy:deploy-file \
+              -DrepositoryId=${version}-release-assets \
+              -Durl=https://repo.liquibase.net/repository/oss-release-assets \
+              -Dfile="$file" \
+              -DgroupId=org.liquibase \
+              -DartifactId=$(basename "$file" | rev | cut -d'.' -f2- | rev) \ 
+              -Dversion=${version} \
+              -Dpackaging=$(basename "$file" | cut -d'.' -f2-)
+          done
```
StevenMassaro, line 572: What happens if this were to run a second time, with the same filenames? Would it overwrite the files already in nexus? If so, then that doesn't seem helpful.

sayaliM0412: seems like the check I had for "tag existence" was removed locally during some changes. I'll add it back. Good catch!

sayaliM0412: Changes made. Good catch. 🎸

.github/workflows/release-published.yml
```diff
@@ -538,4 +538,53 @@ jobs:
               -Dclassifiers=,sources,javadoc,
           
           done
-            
\ No newline at end of file
+            
+  publish_assets_to_internal_nexus:
+    if: ${{ inputs.dry_run == false }}
+    name: Publish OSS released assets to repo.liquibase.net
+    needs: [ setup ]
```
StevenMassaro, line 545: Another thought. Should the actual release step depend on this step? If not, won't they run concurrently, and the release step would still overwrite the existing release artifacts?

PR #6688, opened 2025-01-30
BugFix: Issue #6564: added a unique changelog history query in StandardChangeLogHistoryService

### LINKED ISSUE #6564: Recalculating checksums results in exponential performance degradation
### Search first

### Description

I have a database that reruns Liquibase frequently to potentially add table partitions as well as execute a procedure that copies data from one table to another. This results in a higher number of rows in the databasechangelog table than "normal" scenarios. In my test case, 27k rows.

I have always cleared checksums before running Liquibase due to the historical inability of Liquibase to accurately calculate checksums (I suspect due to the use of system properties within the changesets).

In this case, it's obvious that Liquibase is recalculating the checksums for the exact same changeset once for every instance of the changeset in the table. For example, there are changesets that have executed 1577 times. While Liquibase updates all 1577 rows at once:

[2024-12-04 16:16:34] FINE [liquibase.changelog] Updating null or out of date checksum on changeSet JdbcAuditChangeLog.xml::430::SYSTEM to correct value
[2024-12-04 16:16:34] FINE [liquibase.executor] UPDATE public.databasechangelog SET MD5SUM = '9:cc72d6f444baf60e74560b719ecc7687' WHERE ID = '430' AND AUTHOR = 'SYSTEM' AND FILENAME = 'JdbcAuditChangeLog.xml'
[2024-12-04 16:16:34] FINE [liquibase.executor] 1577 row(s) affected

It does it 1577 times. Searching my log file for that statement yields a count of 1577:

$ grep "UPDATE public.databasechangelog SET MD5SUM = '9:cc72d6f444baf60e74560b719ecc7687' WHERE ID = '430' AND AUTHOR = 'SYSTEM' AND FILENAME = 'JdbcAuditChangeLog.xml" db-init-audit-log.txt  | wc -l

 **1577**

### Steps To Reproduce

1. Import the attached databasechangelog.csv into an empty databasechangelog table.
2. Run clearchecksums on this database, with --log-level=debug
3. Note how many update statements are logged for clearing the checksums.

### Expected/Desired Behavior

Each clearchecksums update statement should execute only once per changeset ID.

### Liquibase Version

4.30.0

### Database Vendor & Version

_No response_

### Liquibase Integration

CLI

### Liquibase Extensions

n/a

Ubuntu

### Additional Context

_No response_

dhsmith1001:
Fixes #6564. 

Added the new method, StandardChangeLogHistoryService.getUniqueRanChangeSets(), to retrieve only one row for each unique changeSet from databasechangelog. This reduces redundant calculation of md5sums and updates to databasechangelog.

## Impact

## Description
I tracked the slow performance in recalculating checksums to the AbstractChangeLogHistoryService.upgradeChecksums() method. The outer loop driven by the result set from StandardChangeLogHistoryService.getRanChangeSets() method causes the loop to execute once for every row in databasechangelog, which is unnecessary for this purpose as there could be multiple rows in databasechangelog for the same changeset which would therefore have the same md5sum.

I created a new method, StandardChangeLogHistoryService.getUniqueRanChangeSets() with a corresponding new method in the ChangeLogHistoryService interface, and supporting methods for the where clause generation in SelectFromDatabaseChangeLogStatement.

The changes described above correspond to a SQL statement like:

select id, author, filename, exectype, count(*)
from databasechangelog
where 1 = 1
group by id, author, filename, exectype

This greatly reduces the number of iterations in AbstractChangeLogHistoryService.upgradeChecksums(), and eliminates the redundant calculation and update of checksums.

## Things to be aware of
The additional method in the ChangeLogHistoryService required adding that method to MockChangeLogHistoryService and OfflineChangeLogHistoryService. For those classes I just returned the same result as the getRanChangeSets() method. 

## Things to worry about
All unit tests pass, but I was unable to perform an integration test of the change to OfflineChangeLogHistoryService.

## Additional Context
The motivation for this fix is a specific use-case where I run a Liquibase changeLog repeatedly with several "run always" changeSets in order to create new partitions for two partitioned tables. This causes many rows in databasechangelog for the same changeSet, which caused a major performance problem as Liquibase was updating the md5sums once for every row in databasechangelog rather than just once per unique changeSet, with the problem getting worse over time.

### DIFF
--- a/liquibase-standard/src/main/java/liquibase/changelog/AbstractChangeLogHistoryService.java
+++ b/liquibase-standard/src/main/java/liquibase/changelog/AbstractChangeLogHistoryService.java
@@ -65,7 +65,7 @@ public abstract class AbstractChangeLogHistoryService implements ChangeLogHistor
     @Override
     public void upgradeChecksums(final DatabaseChangeLog databaseChangeLog, final Contexts contexts,
                                  LabelExpression labels) throws DatabaseException {
-        for (RanChangeSet ranChangeSet : this.getRanChangeSets()) {
+        for (RanChangeSet ranChangeSet : this.getUniqueRanChangeSets()) {
             if (ranChangeSet.getLastCheckSum() == null) {
                 List<ChangeSet> changeSets = databaseChangeLog.getChangeSets(ranChangeSet);
                 for (ChangeSet changeSet : changeSets) {
--- a/liquibase-standard/src/main/java/liquibase/changelog/ChangeLogHistoryService.java
+++ b/liquibase-standard/src/main/java/liquibase/changelog/ChangeLogHistoryService.java
@@ -33,6 +33,8 @@ public interface ChangeLogHistoryService extends Plugin {
 
     List<RanChangeSet> getRanChangeSets() throws DatabaseException;
 
+    List<RanChangeSet> getUniqueRanChangeSets() throws DatabaseException;
+
     /**
      * This method was created to clear out MD5sum for upgrade purpose but after some refactoring the logic was moved to Update commands and it should have been removed
      * as everywhere it is called only with boolean false, so for core it is the same as getRanChangeSets().
--- a/liquibase-standard/src/main/java/liquibase/changelog/MockChangeLogHistoryService.java
+++ b/liquibase-standard/src/main/java/liquibase/changelog/MockChangeLogHistoryService.java
@@ -62,6 +62,12 @@ public class MockChangeLogHistoryService implements ChangeLogHistoryService {
     }
 
     @Override
+    public List<RanChangeSet> getUniqueRanChangeSets() throws DatabaseException
+    {
+        return ranChangeSets;
+    }
+
+    @Override
     public List<RanChangeSet> getRanChangeSets(boolean a) throws DatabaseException {
         return ranChangeSets;
     }
--- a/liquibase-standard/src/main/java/liquibase/changelog/OfflineChangeLogHistoryService.java
+++ b/liquibase-standard/src/main/java/liquibase/changelog/OfflineChangeLogHistoryService.java
@@ -144,6 +144,12 @@ public class OfflineChangeLogHistoryService extends AbstractChangeLogHistoryServ
     }
 
     @Override
+    public List<RanChangeSet> getUniqueRanChangeSets() throws DatabaseException
+    {
+        return getRanChangeSets();
+    }
+
+    @Override
     public List<RanChangeSet> getRanChangeSets() throws DatabaseException {
         try (
                 Reader reader = new InputStreamReader(Files.newInputStream(this.changeLogFile.toPath()), GlobalConfiguration.OUTPUT_FILE_ENCODING.getCurrentValue())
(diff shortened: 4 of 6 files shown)


MalloD12:
Hi @dhsmith1001,

Thank you for submitting this PR and contributing with LIquibase. I have been reviewing this PR, and I have a few questions/comment about it:

- There are a few CodeQL reported issues for minor things like parameters not used, variables never read, etc. You can see the entire list above.
- There are some tests currently failing, you can see execution details [here](https://github.com/liquibase/liquibase/actions/runs/13252230945/job/37189678892?pr=6688). Basically, there are some existent integration tests failing because of the method `GroupByIdAuthorFilename.generateSql()` this method produces the below query:

```sql
SELECT ID,AUTHOR,FILENAME,EXECTYPE FROM PUBLIC.DATABASECHANGELOG WHERE MD5SUM is not null GROUP BY ID, AUTHOR, FILENAME, EXECTYPE ORDER BY FILENAME, ID asc
``` 

that makes the below call:

```java
List<Map<String, ?>> results = queryDatabaseChangeLogTableUnique(database);
```

return `none` results. The problem with the below query seems to be that is trying to pull entries where MD5 is `null` which I'm not sure to understand why.

On the other hand, I'm not sure to understand the context of the usage of the `databasechangelog.csv`
 file, how all those `databasechangelog` table entries were generated, because even though changeset is deployed with `runAlways = true` there shouldn't be a repeated entry in the `databasechangelog` table. That's what the [documentation](https://docs.liquibase.com/concepts/tracking-tables/databasechangelog-table.html) states here:

> The table tracks each changeset as a row, identified by a combination of the id, author, and filename columns. There is no primary key on the table. This is to avoid any database-specific restrictions on key lengths. The composite of id, author, and filename is unique across all rows of the table.

I would appreciate it If there is any additional detail you can share about your use case that could help me understand your scenario.

Thanks,
Daniel.

MalloD12:
Hi @dhsmith1001,

Exactly, it seems the  problem here is why/how duplicated rows are created in `DBCL` table. That shouldb't been happening. I tried deploying  a very simple changeset like the one below:

```xml
<changeSet author="SYSTEM" id="1300" runAlways="true">
   <sql>
       SELECT 1 FROM databasechangelog
   </sql>
</changeSet>
``` 
multiple times, but only first time it was executed an entry was created in the `DBCL` table. If duplicate entries are generated for a single changeset, then in my opinion that would be the issue here. I think we can treat that separately (in a new issue) and discuss about it, if there is any additional detail you haven't shared yet with us, please add it there. It will be helpful for troubleshooting and find out what's happening.

Thanks,
Daniel.

MalloD12:
Hi @dhsmith1001,

I want to follow up on this issue with you. Are you planning to create a separate issue to follow up the duplicate entries? If not, would you mind adding some additional details here that help us to reproduce the same issue you do?

Thanks,
Daniel.

dhsmith1001:
Hi Daniel,

Yes, I will add a new issue. I've been out of the office for the past 1 1/2
weeks, so haven't gotten to it yet.

Thanks,
Don

dhsmith1001:
Closed the issue associated with this PR, so closing the PR too. The problem that this PR addressed shouldn't be happening as the databasechangelog table should not have duplicate rows.

New issue: https://github.com/liquibase/liquibase/issues/6821

PR #6692, opened 2025-01-31
Snapshot all configuration values for Oracle sequences

### LINKED ISSUE #6632: Oracle Sequences: Missing changes when working with default values
### Search first

### Description

When creating a change log from two schema versions, default values for sequence properties in the target schema are not applied.

### Steps To Reproduce

Have the following sequence in the current schema: 

> CREATE SEQUENCE  MYSCHEMA.MYSEQUENCE  **MINVALUE 0** MAXVALUE 999999900 INCREMENT BY 30 START WITH 1 CACHE 20 NOORDER  CYCLE  NOKEEP  NOSCALE  GLOBAL ;

And this sequence in the target schema (where MINVALUE 1 is considered the Oracle Default):

> CREATE SEQUENCE  MYSCHEMA.MYSEQUENCE  **MINVALUE 1** MAXVALUE 999999900 INCREMENT BY 30 START WITH 1 CACHE 20 NOORDER  CYCLE  NOKEEP  NOSCALE  GLOBAL ;

This will result in an empty change set, leaving MINVALUE at 0.  

### Expected/Desired Behavior

Detect the actual schema change.

### Liquibase Version

Since 3.6

### Database Vendor & Version

Oracle 19c

### Liquibase Integration

CLI

### Liquibase Extensions

none

Windows and Linux

### Additional Context

**Analysis:**

Since commit 88beffd8a6cbe9725a11eca35ba611ac0444aee9 default values are returned as `null`, which results in the following difference:

> MYSEQUENCE: {minValue=minValue changed from 'null' to '0'}

The method `ChangedSequenceChangeGenerator.fixChanged()` creates an `AlterSequenceChange` where `maxValues`is `null`which is not considered a change.

marchof:
## Impact

## Description

In commit 88beffd sequence configuration values with Oracle defaults have been converted to null to exclude them from the change log. Unfortunatelly this broke the ability to create correct diffs in the situation where the new state should be the default value (which is currently represented as null and therefore not included in the change log).

Restore capturing of all attributes like it is done for other databases and elements.

Fixes #6632

## Things to be aware of

Needs to be built and verified in the environment where we originally detected the issue.

## Things to worry about

No regression tests?

### DIFF
--- a/liquibase-standard/src/main/java/liquibase/snapshot/jvm/SequenceSnapshotGenerator.java
+++ b/liquibase-standard/src/main/java/liquibase/snapshot/jvm/SequenceSnapshotGenerator.java
@@ -190,20 +190,14 @@ public class SequenceSnapshotGenerator extends JdbcSnapshotGenerator {
             if (catalogName == null || catalogName.isEmpty()) {
                 catalogName = database.getDefaultCatalogName();
             }
-            StringBuilder sql = new StringBuilder("SELECT sequence_name, \n")
-                    .append("CASE WHEN increment_by > 0 \n")
-                    .append("     THEN CASE WHEN min_value=1 THEN NULL ELSE min_value END\n")
-                    .append("     ELSE CASE WHEN min_value=(-999999999999999999999999999) THEN NULL else min_value END\n")
-                    .append("END AS min_value, \n")
-                    .append("CASE WHEN increment_by > 0 \n")
-                    .append("     THEN CASE WHEN max_value=999999999999999999999999999 THEN NULL ELSE max_value END\n")
-                    .append("     ELSE CASE WHEN max_value=last_number THEN NULL else max_value END \n")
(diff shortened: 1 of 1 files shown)


marchof:
I successfully tested this fix in our environment. Only when the target state is default (and the actual state has a different value) for example the following valid SQL is generated and the database is adjusted as expected:

    ALTER SEQUENCE MYSCHEMA.MYSEQUENCE MINVALUE 1;
    
Afterwards, when both (target and actual) have the default value of 1 no diff is reported any more.

filipelautert (approved):
This PR snapshots sequence values that were being ignored for being considered default values, but it affected diff command. This behavior was specific for Oracle, and it was different from all the others databases. 
Thanks for the fix @marchof !

tati-qalified (approved):

rberezen (approved):

PR #6694, opened 2025-01-31
Show message and exit if user attempts to use Azure without extension DAT-19430

wwillard7800:
## Impact

### DIFF
--- a/liquibase-cli/src/main/java/liquibase/integration/commandline/LiquibaseCommandLine.java
+++ b/liquibase-cli/src/main/java/liquibase/integration/commandline/LiquibaseCommandLine.java
@@ -99,10 +99,32 @@ public class LiquibaseCommandLine {
             }
         }
 
+        //
+        // Check for arguments which contain azure-storage and get out
+        //
+        if (args.length > 1 && azureArgumentIsPresent(args)) {
+            try {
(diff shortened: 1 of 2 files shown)


filipelautert (approved):

StevenMassaro (changes requested):
We did this in a different way for S3. Can we emulate that here instead of doing it this way? See `S3PathHandlerJarExistenceCheck`

wwillard7800:
> We did this in a different way for S3. Can we emulate that here instead of doing it this way? See `S3PathHandlerJarExistenceCheck`

The code in the corresponding DAT-19430 Pro PR emulates that.  This code handles the storage account ID when it is passed in as a command line argument.  Otherwise, the message is just an "unknown argument".

StevenMassaro (approved):
I see. The code here looks fine but I'm not a huge fan of this living in this repo. I don't have an alternative suggestion though, so if it's a requirement, then I suppose this is what we must do.

wwillard7800:
> I see. The code here looks fine but I'm not a huge fan of this living in this repo. I don't have an alternative suggestion though, so if it's a requirement, then I suppose this is what we must do.

I can't say I'm thrilled with it, either.  We do something similar for `checks` commands.

Issue #6681, opened 2025-01-28
Missing reserved keyword "row" for MySQL 8.0.2


stefanzugal:
### Search first

- [x] I searched and no similar issues were found

### Description

We use `row` as a column name. When updating to MySQL 8.0, this is now a reserved keyword, but Liquibase does not recognize it as such and doesn't quote it in table creation statements.

See also: https://dev.mysql.com/doc/refman/8.0/en/keywords.html#keywords-8-0-detailed-R
> ROW (R); became reserved in 8.0.2

### Steps To Reproduce

Use the `row` keyword as column name in a Liquibase changeset for `createTable`.

### Expected/Desired Behavior

Liquibase should quote the reserved keyword for MySQL >= 8.0.

### Liquibase Version

4.31.0

### Database Vendor & Version

MySQL 8.0

### Liquibase Integration

Spring Boot

### Liquibase Extensions

None

### Additional Context

There is already a pr for fixing the keyword `MANUAL` for 8.4: https://github.com/liquibase/liquibase/pull/6128

I think it should be reasonable easy to build on it so that this will work for this issues as well. Also, I have seen that there are these bug reports, but the seem to be forgotten/abandoned, so I have created this issue with a specific suggestion on how to fix it. See:
- https://liquibase.jira.com/browse/CORE-3235 general issue
- https://liquibase.jira.com/browse/CORE-3154 ROWS

### Are you willing to submit a PR?

- [x] I'm willing to submit a PR (Thank you!)

MalloD12:
Hi @stefanzugal,

If you are keen to submit a PR for this, please feel free to do it. We will assist you if you need any help.

Thanks,
Daniel.

stefanzugal:
Hi @MalloD12,

thanks for the offer! I tried to create a PR here: https://github.com/liquibase/liquibase/pull/6697

Looking forward to your thoughts.

Best wishes,
Stefan

PR #6697, opened 2025-02-04
Add reserved words for MySQL 8.0

stefanzugal:
Fixes: #6681

## Impact

## Description

Marks the following reserved words for MySQL 8.0:
- `FUNCTION`
- `ROW`
- `ROWS`

Side note: `CUBE` was also marked in 8.0, but as the code already defined it as reserved, this change does not add it. See also: https://dev.mysql.com/doc/refman/8.0/en/keywords.html#keywords-8-0-detailed-R

## Things to be aware of

The field `RESERVED_WORDS` from `MySQLDatabase` is still static, so once you connect to a MySQL database >= 8.0, the reserved words are added for all.

I have seen that there were some refactorings in https://github.com/liquibase/liquibase/pull/6128/commits/72f7de91b261a5a48de27e32d3e08c3a8aba3f17 - but it seems that the commit was not merged; only a single commit was merged in https://github.com/liquibase/liquibase/pull/6128 instead. From reading the PR, I could not figure out why this was done.

## Things to worry about

Possibly the same as to be aware of.

## Additional Context

MySQL adds/removes reserved words over time, see: https://dev.mysql.com/doc/refman/8.0/en/keywords.html#keywords-8-0-detailed-R


stefanzugal:
I have seen that the tests failed with the following issue:

> Error:  Failures: 
Error:    MySQLDatabaseTest.getVersionSpecificMySQLReservedWords:119 Expected FUNCTION to be non-reserved in MySQL < 8 ==> expected: false but was: true

The tests run successfully on my computer (classic developer excuse :wink:), so I suspect that may have to do with the issue sketched in "Things to be aware of".

MalloD12 (approved):
**Approved**.

Changes look good to me. Fixed failing test and split the existing test for the current change into two separate tests (one to validate new reserved words are not part of an `5.7` instance and one other to validate they are for a `8.0` instance).

Thank you for your contribution, @stefanzugal!

Daniel.

filipelautert (approved):

rberezen (approved):

tati-qalified (approved):

stefanzugal:
Very nice, thx for the support @MalloD12!

PR #6698, opened 2025-02-04
🔧 DAT-19341: publish oss released assets to s3

sayaliM0412:
This pull request includes changes to the `.github/workflows/release-published.yml` file to update the process for publishing OSS released assets. The main change is switching from publishing to an internal Nexus repository to publishing to an S3 bucket.

Changes to publishing process:

* Renamed the job from `publish_assets_to_internal_nexus` to `publish_assets_to_s3_bucket` and updated the job name to reflect the new destination.
* Removed the environment variables related to Nexus and added AWS credentials for accessing the S3 bucket.
* Removed the step that checks if the version already exists in Nexus, as this is no longer relevant.
* Updated the publishing step to use `aws s3 sync` instead of Maven commands for deploying files to the S3 bucket.

### DIFF
--- a/.github/workflows/release-published.yml
+++ b/.github/workflows/release-published.yml
@@ -539,32 +539,12 @@ jobs:
           
   #         done
             
-  publish_assets_to_internal_nexus:
+  publish_assets_to_s3_bucket:
     if: ${{ inputs.dry_run == false }}
-    name: Publish OSS released assets to repo.liquibase.net
+    name: Publish OSS released assets to s3 bucket liquibaseorg-origin
     needs: [ setup ]
     runs-on: ubuntu-22.04
-    env:
-      MAVEN_USERNAME: ${{ secrets.INTERNAL_NEXUS_USERNAME }}
-      MAVEN_PASSWORD: ${{ secrets.INTERNAL_NEXUS_PASSWORD }}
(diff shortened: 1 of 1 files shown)


.github/workflows/release-published.yml
```diff
@@ -539,32 +539,12 @@ jobs:
...
+    name: Publish OSS released assets to s3 bucket liquibaseorg-origin
     needs: [ setup ]
     runs-on: ubuntu-22.04
-    env:
-      MAVEN_USERNAME: ${{ secrets.INTERNAL_NEXUS_USERNAME }}
-      MAVEN_PASSWORD: ${{ secrets.INTERNAL_NEXUS_PASSWORD }}
-      GPG_PASSWORD: ${{ secrets.GPG_PASSPHRASE }}
     steps:
-      # Check in nexus if the folder with the version already exists, if exists exit with error code
-      - name: Check if the version already exists in nexus
```
filipelautert, line 553: Aren't we missing this check now? Cause in a rerun we would overwrite them in S3.
Or we could add a timestamp to S3 directory and so even if we rerun it we'll have a new directory in S3 that we can remove later on.

filipelautert (changes requested):
A lot better now! thanks . I just added one small detail that seems to have been missing.

filipelautert (approved):

rberezen (approved):

PR #6703, opened 2025-02-05
DAT-19476: generate minimal tar.gz and zip for OSS

sayaliM0412:
This pull request includes significant changes to the build and packaging process, primarily focusing on the introduction of a new "minimal" version of the Liquibase distribution. The changes involve modifications to various GitHub workflow files and scripts to support this new version.

#### Build and Packaging Enhancements:
* Added steps to create and package a minimal version of Liquibase in `.github/util/re-version.sh`. This includes extracting, copying necessary files, and repackaging both the full and minimal versions.
* Updated `.github/workflows/build.yml` to build both the full and minimal versions of Liquibase, including backing up and restoring tarballs to prevent overwriting. [[1]](diffhunk://#diff-5c3fa597431eda03ac3339ae6bf7f05e1a50d6fc7333679ec38e21b337cb6721L211-R240) [[2]](diffhunk://#diff-5c3fa597431eda03ac3339ae6bf7f05e1a50d6fc7333679ec38e21b337cb6721R250-R259) [[3]](diffhunk://#diff-5c3fa597431eda03ac3339ae6bf7f05e1a50d6fc7333679ec38e21b337cb6721R295)
* Modified `.github/workflows/create-release.yml` to include minimal version artifacts in the release process.

#### Workflow Cleanup:
* Removed or commented out several steps and workflows that are no longer needed or are being temporarily disabled, such as functional tests, test harness, and OWASP scanner. [[1]](diffhunk://#diff-5c3fa597431eda03ac3339ae6bf7f05e1a50d6fc7333679ec38e21b337cb6721L305-R362) [[2]](diffhunk://#diff-8ac33fe295df086d3e55df1eb01194819d34cc0f5f54076dba96ceac0e40bd79L77-R95) [[3]](diffhunk://#diff-7314d0ebbd2e9537ae4889316745b4fd2fa43cb86275c9caae18a86ba228b642L441-R489)

#### Configuration Updates:
* Introduced a new profile in `liquibase-dist/pom.xml` for the minimal distribution, including setting the expected distribution contents filename. [[1]](diffhunk://#diff-b76b3ecdecffd67680c34289b12d668c180a0254226ef32428f02f8447a3b0e6R37) [[2]](diffhunk://#diff-b76b3ecdecffd67680c34289b12d668c180a0254226ef32428f02f8447a3b0e6R415-R433)
* Added a new file `liquibase-dist/expected-distribution-contents-minimal.txt` to define the contents of the minimal distribution.

### DIFF
--- a/.github/util/re-version.sh
+++ b/.github/util/re-version.sh
@@ -175,11 +175,27 @@ if [ -z "$extension_name" ]; then
   cp $outdir/liquibase-commercial-$version.jar $workdir/internal/lib/liquibase-commercial.jar ##save versioned jar as unversioned to include in zip/tar
 
   ## Extract tar.gz and rebuild it back into the tar.gz and zip
-  mkdir $workdir/tgz-repackage
+  mkdir $workdir/tgz-repackage $workdir/tgz-repackage-minimal
+
   tar -xzf $workdir/liquibase-$MODIFIED_BRANCH_NAME-SNAPSHOT.tar.gz -C $workdir/tgz-repackage
+  tar -xzf $workdir/liquibase-minimal-$MODIFIED_BRANCH_NAME-SNAPSHOT.tar.gz -C $workdir/tgz-repackage-minimal
+
   cp $workdir/internal/lib/liquibase-core.jar $workdir/tgz-repackage/internal/lib/liquibase-core.jar
   cp $workdir/internal/lib/liquibase-commercial.jar $workdir/tgz-repackage/internal/lib/liquibase-commercial.jar
+
+  # copy core and commercial jars to minimal
+  cp $workdir/internal/lib/liquibase-core.jar $workdir/tgz-repackage-minimal/internal/lib/liquibase-core.jar
+  cp $workdir/internal/lib/liquibase-commercial.jar $workdir/tgz-repackage-minimal/internal/lib/liquibase-commercial.jar
+
+  # replace the versions in all the text files
   find $workdir/tgz-repackage -name "*.txt" -exec sed -i -e "s/\(0\|release\|master\)-SNAPSHOT/$version/" {} \;
+
   (cd $workdir/tgz-repackage && tar -czf $outdir/liquibase-$version.tar.gz *)
   (cd $workdir/tgz-repackage && zip -qr $outdir/liquibase-$version.zip *)
+
+  # replace the versions in all the text files in minimal
+  find $workdir/tgz-repackage-minimal -name "*.txt" -exec sed -i -e "s/\(0\|release\|master\)-SNAPSHOT/$version/" {} \;
+
+  (cd $workdir/tgz-repackage-minimal && tar -czf $outdir/liquibase-minimal-$version.tar.gz *)
+  (cd $workdir/tgz-repackage-minimal && zip -qr $outdir/liquibase-minimal-$version.zip *)
 fi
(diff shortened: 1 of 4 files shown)


.github/workflows/run-tests.yml
```diff
@@ -454,7 +454,7 @@ jobs:
 
   run-build-publish-file:
     needs: [ build_tests,integration-test ]
-    uses: liquibase/liquibase/.github/workflows/build.yml@master
+    uses: ./.github/workflows/build.yml
```
rberezen, line 457: @sayaliM0412 just to confirm if it's desired change?

sayaliM0412: yes, the workflow it is referencing is in the same repo ie `liquibase`, therefore we can refer to it as `./.`
:)

StevenMassaro (approved):
I just realized that we could add the `-minimal` to the targz when we build it in maven, which would probably simplify things for you here. I guess it's too late for that.

.github/util/re-version.sh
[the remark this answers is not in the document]
filipelautert: Soon we may come up with OSS only distribution. Then it will be another set of copy & paste... or functions!

.github/workflows/build.yml
```diff
@@ -208,21 +204,59 @@ jobs:
           echo "Saving windows key"
           echo "$INSTALL4J_WINDOWS_KEY" | base64 -d > liquibase-dist/target/keys/datical_windows.pfx
           # we are packaging and deploying liquibase tar which includes liquibase-commercial
-          ./mvnw -B -pl liquibase-dist -P liquibase-commercial source:jar package failsafe:integration-test failsafe:verify "-Dliquibase.checks=true" "-Dliquibase-pro.version=${{ needs.setup.outputs.proBranchNamePrefix }}-SNAPSHOT" "-Dbuild.repository.owner=liquibase" "-Dbuild.repository.name=liquibase" "-Dbuild.branch=${{ needs.setup.outputs.thisBranchName }}" "-Dbuild.number=${{ github.run_number }}" "-Dbuild.commit=${{ needs.setup.outputs.latestMergeSha }}-SNAPSHOT" "-Dbuild.timestamp=${{ needs.setup.outputs.timeStamp }}"
+          # Build liquibase-commercial version
```
filipelautert, line 207: We can use functions here too

.github/workflows/build.yml
```diff
@@ -256,10 +290,11 @@ jobs:
 
           ##prepare branch-named convenience artifacts directories
           mkdir artifacts-named
-          
+
           cp liquibase-dist/target/liquibase-${{ needs.setup.outputs.thisBranchName }}-SNAPSHOT.tar.gz artifacts-named/liquibase-${{ needs.setup.outputs.thisBranchName }}.tar.gz
+          cp liquibase-dist/target/liquibase-minimal-${{ needs.setup.outputs.thisBranchName }}-SNAPSHOT.tar.gz artifacts-named/liquibase-${{ needs.setup.outputs.thisBranchName }}-minimal.tar.gz
```
filipelautert, line 295: I see that the tgz is copied here, but I don't see where the zip is.

filipelautert (approved):

.github/workflows/build.yml
sayaliM0412: Following the code we had for commercial over here. The final installer does has the `zip` and the `tar.gz`

PR #6710, opened 2025-02-06
Handle changes skipped due to OS mismatch DAT-19334

wwillard7800:
## Impact

### DIFF
--- a/liquibase-standard/src/main/java/liquibase/changelog/visitor/StatusVisitor.java
+++ b/liquibase-standard/src/main/java/liquibase/changelog/visitor/StatusVisitor.java
@@ -1,11 +1,14 @@
 package liquibase.changelog.visitor;
 
 import liquibase.Scope;
+import liquibase.change.Change;
+import liquibase.change.core.ExecuteShellCommandChange;
 import liquibase.changelog.*;
 import liquibase.changelog.filter.ChangeSetFilterResult;
 import liquibase.changelog.filter.NotInChangeLogChangeSetFilter;
 import liquibase.database.Database;
 import liquibase.exception.LiquibaseException;
+import org.apache.tools.ant.taskdefs.Execute;
 
 import java.util.*;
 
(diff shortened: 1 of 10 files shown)


liquibase-standard/src/main/java/liquibase/changelog/visitor/StatusVisitor.java
```diff
@@ -1,11 +1,14 @@
 package liquibase.changelog.visitor;
 
 import liquibase.Scope;
+import liquibase.change.Change;
```
filipelautert, line 4: do we need those new imports?

filipelautert (approved):

liquibase-standard/src/main/java/liquibase/changelog/visitor/StatusVisitor.java
wwillard7800: No, I should have optimized these guys.

liquibase-standard/src/main/java/liquibase/util/ShowSummaryUtil.java
```diff
@@ -181,7 +184,7 @@ private static FilteredChanges showDetailTable(List<ChangeSet> skippedChangeSets
         SortedMap<String, Integer> mdcSkipCounts = new TreeMap<>();
         mdcSkipCounts.put(totalSkippedMdcKey, skippedChangeSets.size() + filterDenied.size());
 
-        List<ChangeSetStatus> finalList = createFinalStatusList(skippedChangeSets, skippedBecauseOfLicenseChangeSets, filterDenied, mdcSkipCounts);
+        List<ChangeSetStatus> finalList = createFinalStatusList(skippedChangeSets, skippedBecauseOfLicenseChangeSets, skippedBecauseOfOsMismatchChangeSets, filterDenied, mdcSkipCounts);
```
filipelautert, line 187: The `mdcSkipCounts` is not being used in the method and I allowed Copilot to remove it from the method signature.
Now the build fails 🤦🏻  
But if it's not used because we don't need we can remove the calculation from here, right? or should it be used in `createFinalStatusList`?

filipelautert (changes requested):

liquibase-standard/src/main/java/liquibase/util/ShowSummaryUtil.java
wwillard7800: It looks like it is used in `showDetailTable`, but not `createFinalStatusList`.

filipelautert (approved):

filipelautert (approved):

filipelautert (approved):

Issue #6660, opened 2025-01-21
Remove SNAPSHOT creation for indexExists, tableExists, viewExists for DB2


Stefan4112:
### Search first

- [x] I searched and no similar issues were found

### Description

The pre conditions indexExists, tableExists, viewExists create always a SNAPSHOT. This results in bad performance. The objects can be found in the system tables with simple SQL:

DB2 LUW
view: `SELECT count(*) FROM syscat.views WHERE viewschema = 'MYSCHEMA' AND viewname = 'MYVIEW';`
table: `SELECT count(*) FROM syscat.tables WHERE tabschema = 'MYSCHEMA' AND tabname = 'MYTABLE';`
index: `SELECT count(*) FROM syscat.indexes WHERE indschema = 'MYSCHEMA' AND indname = 'MYINDEX';`

DB2 IBM
view: `SELECT count(*) FROM qsys2.sysindexes WHERE index_schema = 'MYSCHEMA' and index_name = 'MYVIEW' AND TABLE_TYPE = 'V'`
table (including MQT): `SELECT count(*) FROM qsys2.sysindexes WHERE index_schema = 'MYSCHEMA' and index_name = 'MYTABLE AND (TABLE_TYPE = 'M' OR TABLE_TYPE = 'T')'`
index: `SELECT count(*) FROM qsys2.sysindexes WHERE index_schema = 'MYSCHEMA' and index_name = 'MYINDEX'`

DB2 z/OS
I don't know this database, but should be similar.

### Steps To Reproduce

Example of pre condition:
`<changeSet id="slow-example" author="menten GmbH" dbms="db2" runOnChange="false">
	<preConditions>
		<indexExists tableName="myTable" schemaName="mySchema" />
	</preConditions>
</changeSet>`

### Expected/Desired Behavior

I expect no SNAPSHOTs if indexExists, tableExists, viewExists are used.

### Liquibase Version

4.30

### Database Vendor & Version

DB2 LUW, DB2 IBM and maybe DB2 z/OS

### OS and/or Infrastructure Type/Provider

Ubuntu, IBM i

### Are you willing to submit a PR?

- [ ] I'm willing to submit a PR (Thank you!)

t-beckmann:
Most likely you are affected by https://github.com/liquibase/liquibase/issues/6685 a fix for which I provided already. Notice, I opened another issue as yours requested the removal of snapshots which I'd doubt to happen...

Stefan4112:
[...]

PR #6774, opened 2025-02-27
Use hasIgnoreNested method for all database preconditions

wwillard7800:
Also added some test file cleanup

Github issue #6660

## Impact


filipelautert (approved):
Thanks @wwillard7800 !

rberezen (approved):

tati-qalified (approved):

danielthegray:
This commit broke the `uniqueConstraintExists` precondition on Oracle. The unique constraint is no longer reported as existing, which breaks the subsequent scripts.

I appreciate that taking the larger snapshot with nested objects was slower, but we're basically trading-off correctness for execution speed.

Please revert this commit and find another/more correct way of fixing your execution speed issue.

PR #6787, opened 2025-03-04
For SQL Server, sync up the backing index for PK and UC to the one that is auto-generated  DAT-19267

wwillard7800:
## Impact

### DIFF
--- a/liquibase-standard/src/main/java/liquibase/snapshot/SnapshotGeneratorFactory.java
+++ b/liquibase-standard/src/main/java/liquibase/snapshot/SnapshotGeneratorFactory.java
@@ -301,7 +301,7 @@ public class SnapshotGeneratorFactory {
 
         //
         // For SQL Server, try to set the backing index for primary key and
-        // unique constraints to the index that is created
+        // unique constraints to the index that is created automatically
         //
         if (database instanceof MSSQLDatabase) {
             Set<PrimaryKey> primaryKeys = snapshot.get(PrimaryKey.class);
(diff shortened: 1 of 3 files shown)


liquibase-standard/src/main/java/liquibase/structure/core/PrimaryKey.java
```diff
@@ -149,6 +149,16 @@ public String toString() {
 
     public List<Column> getColumns() {
         return getAttribute("columns", List.class);
+        /*
+        List<Object> cols = getAttribute("columns", List.class);
```
abrackx, line 153: Can we get rid of this?

wwillard7800: Yes, thanks for the catch.

liquibase-standard/src/main/java/liquibase/snapshot/SnapshotGeneratorFactory.java
```diff
@@ -447,4 +469,35 @@ private void getContainerTypes(Class<? extends DatabaseObject> type, Database da
         }
 
     }
+
+    private void syncBackingIndex(PrimaryKey pk, DatabaseSnapshot snapshot) {
```
abrackx, line 473: Can we combine these two method? They look very similar.

wwillard7800: I was able to refactor to 1 method by using the underlying attributes map.

abrackx (approved):

PR #6793, opened 2025-03-07
fix: classpath expression should also work when it is separated by an empty space instead of an =

filipelautert:
## Impact


MalloD12 (approved):
This looks good to me. My only comment here is it would be nice to add a `ParameterUtilTest` class under `liquibase-cli/src/test/groovy/liquibase/integration/commandline` and then add a test that specifies a matrix with the different ways of specifying parameters using CLI. I can help with that if you wish.

Thanks,
Daniel.

tati-qalified (approved):

PR #6794, opened 2025-03-07
Added option to force schema attributes to be included if they are default

wwillard7800:
DAT-19043

## Impact


liquibase-standard/src/main/java/liquibase/GlobalConfiguration.java
```diff
@@ -182,6 +184,16 @@ public class GlobalConfiguration implements AutoloadedConfigurations {
                 .setDefaultValue(false)
                 .build();
 
+        INCLUDE_CATALOG_NAME_FOR_DEFAULT = builder.define("includeCatalogNameForDefault", Boolean.class)
+                .setDescription("If true, the catalog name is included for the default schema")
```
filipelautert, line 188: shouldn't it be `the catalog name is included for the default catalog` ?

filipelautert (changes requested):

liquibase-standard/src/main/java/liquibase/GlobalConfiguration.java
wwillard7800: Thanks for pointing that out

filipelautert (approved):

filipelautert (approved):

abrackx (approved):

PR #6801, opened 2025-03-10
reduce number of API calls for analytics by caching events (DAT-19789)

StevenMassaro:
## Impact

### DIFF
--- a/liquibase-standard/src/main/java/liquibase/analytics/LiquibaseAnalyticsListener.java
+++ b/liquibase-standard/src/main/java/liquibase/analytics/LiquibaseAnalyticsListener.java
@@ -20,6 +20,8 @@ import org.yaml.snakeyaml.nodes.Tag;
 import java.net.HttpURLConnection;
 import java.net.URL;
 import java.nio.charset.StandardCharsets;
+import java.util.ArrayList;
+import java.util.List;
 import java.util.concurrent.atomic.AtomicBoolean;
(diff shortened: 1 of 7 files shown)


liquibase-standard/src/main/java/liquibase/analytics/LiquibaseAnalyticsListener.java
```diff
@@ -56,6 +61,38 @@ public int getPriority() {
...
+                    Scope.getCurrentScope().getLog(getClass()).warning("Failed to send analytics events during shutdown hook.", e);
+                }
+            });
+            Runtime.getRuntime().addShutdownHook(haltedHook);
+        }
+    }
+
+    private int getDefaultMaxAnalyticsCacheSize(Event event) {
+        if (event != null && StringUtils.equals(event.getLiquibaseInterface(), Event.JAVA_API_INTEGRATION_NAME)) {
+            return 10;
```
filipelautert, line 90: Just out of curiosity: 10 is a magic number or it was decided based on the statistics already reported?

filipelautert (approved):

liquibase-standard/src/main/java/liquibase/analytics/LiquibaseAnalyticsListener.java
StevenMassaro: Totally arbitrary number.

PR #6802, opened 2025-03-10
Fix snapshot for MSSQL decimal sequences (DAT-19504)

abrackx:
## Impact

## Description
When snapshotting a sequence created with decimal(19) because we cast everything to a bigint we may hit an arithmetic overflow when casting the min, max, or start value of the sequence. 

This change:
1. Captures the data type and precision of the decimal sequence in our snapshot query
2. Generates the correct sql if we have a decimal sequence. Previously all MSSQL snapshots were assumed to be of type bigint, which is not necessary to include in your create sequence statement. 

## Things to worry about
Are there other data types that I should worry about while I'm here? I very intentionally did not add the data type for all sequences so as to not produce differences in diff from a previously generated changelog.

### DIFF
--- a/liquibase-standard/src/main/java/liquibase/snapshot/jvm/SequenceSnapshotGenerator.java
+++ b/liquibase-standard/src/main/java/liquibase/snapshot/jvm/SequenceSnapshotGenerator.java
@@ -225,10 +225,7 @@ public class SequenceSnapshotGenerator extends JdbcSnapshotGenerator {
                 return new RawParameterizedSqlStatement(sql.toString());
             }
         } else if (database instanceof MSSQLDatabase) {
-            StringBuilder sql = new StringBuilder("SELECT SEQUENCE_NAME, cast(START_VALUE AS BIGINT) AS START_VALUE, cast(MINIMUM_VALUE AS BIGINT) AS MIN_VALUE, ")
-                    .append("cast(MAXIMUM_VALUE AS BIGINT) AS MAX_VALUE, CAST(INCREMENT AS BIGINT) AS INCREMENT_BY, CYCLE_OPTION AS WILL_CYCLE ")
-                    .append("FROM INFORMATION_SCHEMA.SEQUENCES WHERE SEQUENCE_SCHEMA = ?");
-            return new RawParameterizedSqlStatement( sql.toString(), schema.getName());
+            return getMSSQLQuery(schema);
         } else if (database instanceof MariaDBDatabase) {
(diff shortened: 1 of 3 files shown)


liquibase-standard/src/main/java/liquibase/snapshot/jvm/SequenceSnapshotGenerator.java
```diff
@@ -266,4 +263,16 @@ protected SqlStatement getSelectSequenceStatement (Schema schema, Database datab
...
     }
+
+    private static RawParameterizedSqlStatement getMSSQLQuery(Schema schema) {
+        String sql = "SELECT SEQUENCE_NAME, START_VALUE, MINIMUM_VALUE AS MIN_VALUE, " +
+                "MAXIMUM_VALUE AS MAX_VALUE, INCREMENT AS INCREMENT_BY, CYCLE_OPTION AS WILL_CYCLE, " +
+                // If we have a decimal sequence we want to include the precision with the data type.
+                // This is not necessary with types like bigint. Previously we did not include the data type
+                // in our query which would create the sequence with data type bigint. So we normalize all other
+                // non-decimal seq_type values to null.
+                "IIF(DATA_TYPE = 'decimal', DATA_TYPE + '(' + CAST(NUMERIC_PRECISION AS VARCHAR) + ')', NULL) AS SEQ_TYPE " +
```
abrackx, line 274: I can also capture this `NUMERIC_PRECISION` column in the sequence class and do the concat in one of our generators, but this felt less intrusive.

liquibase-standard/src/main/java/liquibase/snapshot/jvm/SequenceSnapshotGenerator.java
```diff
@@ -266,4 +263,16 @@ protected SqlStatement getSelectSequenceStatement (Schema schema, Database datab
             throw new UnexpectedLiquibaseException("Don't know how to query for sequences on " + database);
         }
     }
+
+    private static RawParameterizedSqlStatement getMSSQLQuery(Schema schema) {
+        String sql = "SELECT SEQUENCE_NAME, START_VALUE, MINIMUM_VALUE AS MIN_VALUE, " +
```
MalloD12, line 268: Hey Alex, can we keep the usage of StringBuilder here? Otherwise, I think this will result in a CodeQL scan issue because of the concatenated string.

StevenMassaro (approved):

filipelautert (approved):
It will be nice to run test-harness against this change as it runs on many sqlserver version

PR #6835, opened 2025-03-25
feat: new tagDatabase parameter to keep tag on rollback

### LINKED ISSUE #6728: Rollback removes changeset with target tag
### Search first

### Description

After upgrade from 3.5.x to 4.31.0 rollback to tag removes also changeset which added this tag which is contrary to docs.
This worked in 3.5.x  as stated in docs and rollback was performed only up to changeset which added target tag leaving it untouched.

### Steps To Reproduce

1. Upgrade from 3.5.x to 4.31.0
2. Run rollback to tag
3. Rollback removes also changeset which added tag

### Expected/Desired Behavior

Rollback should leave (not attempt to rollback) changeset with target tag

### Liquibase Version

4.31.0

### Database Vendor & Version

PostgreSQL 16

### Liquibase Integration

_No response_

### Liquibase Extensions

_No response_

Rocky Linux 9

### Additional Context

_No response_

filipelautert:
## Impact

## Description

Implement functionality to preserve tags when performing rollback operations in Liquibase, by adding a new `keepTagOnRollback` parameter to the `tag` change type.
 This ensures tag information is maintained in the database change history during rollback scenarios, providing better traceability and consistency for database version control.

## Things to be aware of

The `keepTagOnRollback` flag is not stored in dbcl table, but is always reloaded from the changelog. It's also not used to calculate the checksum. 
So you can change the value and it will be reflected in the next execution.

Fixes #6728

### DIFF
--- a/liquibase-integration-tests/src/test/groovy/liquibase/command/core/RollbackIntegrationTest.groovy
+++ b/liquibase-integration-tests/src/test/groovy/liquibase/command/core/RollbackIntegrationTest.groovy
@@ -118,4 +118,25 @@ class RollbackIntegrationTest extends Specification {
         cleanup:
         CommandUtil.runDropAll(h2)
     }
+
+    def "Make sure rollback does not remove tagged changeset"() {
+        given:
+        def changelogFile = 'target/test-classes/changelogs/common/test.tag.not.removed.by.rollback.xml'
+        CommandUtil.runUpdate(h2, changelogFile)
+        ConsoleUIService console = Scope.getCurrentScope().getUI() as ConsoleUIService
+        def outputStream = new ByteArrayOutputStream()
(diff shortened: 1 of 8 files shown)


MalloD12:
@AMBERMW13 here is the PD [ticket](https://datical.atlassian.net/browse/PD-5824) created for this change.

MalloD12 (approved):
Code changes look good to me.

tati-qalified (approved):
Code looks good to me! Tested manually with all changelog filetypes, H2 and PostgreSQL.

Found a bug that was not introduced by these changes: using the tagDatabase change type in a SQL changelog with H2, the tags are created correctly but then aren't found when running tag-exists. This means that the changes can't be rolled back, because Liquibase tries to undo a change that it doesn't recognise.

rberezen (approved):

PR #6840, opened 2025-03-26
Comments field from CreateProcedure change renamed to 'comment'

MalloD12:
## Impact

## Description

This change is code enhancement to make a match between `CreateProcedure` change class and `createProcedure` element from XML XSD schema. This change comes from after talking about it with the team, but it is worth mentioning it doesn't fix ##6425 that still to make it worth there it mapping and usage need to be implemented.


liquibase-standard/src/test/java/liquibase/serializer/core/xml/XMLChangeLogSerializerTest.java
```diff
@@ -309,7 +309,7 @@ public void createNode_CreateIndexChange() throws Exception {
     public void createNode_CreateProcedureChange() throws Exception {
         CreateProcedureChange refactoring = new CreateProcedureChange();
         refactoring.setProcedureText("CREATE PROC PROCBODY HERE");
-        refactoring.setComments("Comments go here");
```
filipelautert, line 312: Do you think we can add a new test for it?

MalloD12: I'm not sure I understand what would be the new scenario we want to add.

filipelautert: I mean that we are not really testing the comments here. We set them but we are not confirming that they are generated in `update-sql` command.

MalloD12: Yeap, we removed that part because that comment is not being displayed anywhere. I am still not sure whether this comment should be in a procedure. If it's part of the body then it should be set as part of it, and then if we want to do some check against we should check whether a comment is contained or not as part of the body. If that's not the case, I don't know where it should be in an `update-sql` output.

tati-qalified (approved):

liquibase-standard/src/test/java/liquibase/serializer/core/xml/XMLChangeLogSerializerTest.java
filipelautert: So we could change the description of the fields, as the help says:
```
  comment string 
    Description: Inline comments generated by update-sql. Not applied to the database
```
But checking comment [here](https://docs.liquibase.com/concepts/changelogs/changeset.html), it says: 
```
Specifies the description of the changeset. XML comments provide the same benefit.
```

I think we could use the same description and it would settle this issue for all.

cc @mariochampion

filipelautert (approved):

PR #6858, opened 2025-04-01
SnowflakeOAuthConnection Implementation for Null Username Issue

CharlesQueiroz:
## Impact

## Description

Fixes an issue where connections to Snowflake using OAuth authentication (with token in URL) would display "null@jdbc:snowflake:..." in logs and status messages. This occurred because the JDBC driver returns null for username when using OAuth authentication with the token in the URL.

This PR introduces a specialized SnowflakeOAuthConnection class that provides meaningful username values for OAuth connections. The connection is automatically detected and wrapped by the SnowflakeDatabase class when OAuth authentication is detected in the URL.

## Things to be aware of

- Added a new class SnowflakeOAuthConnection that extends JdbcConnection to handle OAuth authentication
- Modified SnowflakeDatabase.setConnection() to detect OAuth connections and apply the wrapper

## Things to worry about

- The detection of OAuth authentication is based on the URL containing "authenticator=oauth" - this should be reliable but assumes a specific format

## Additional Context

When using an OAuth token in the URL, the connection would appear as "null@jdbc:snowflake:..." in logs because the JDBC driver's getUserName() method returns null in this case.

### DIFF
--- a/liquibase-snowflake/src/main/java/liquibase/database/core/SnowflakeDatabase.java
+++ b/liquibase-snowflake/src/main/java/liquibase/database/core/SnowflakeDatabase.java
@@ -5,6 +5,7 @@ import liquibase.Scope;
 import liquibase.database.AbstractJdbcDatabase;
 import liquibase.database.DatabaseConnection;
 import liquibase.database.jvm.JdbcConnection;
+import liquibase.database.jvm.SnowflakeOAuthConnection;
 import liquibase.exception.DatabaseException;
 import liquibase.executor.ExecutorService;
 import liquibase.statement.core.RawParameterizedSqlStatement;
@@ -171,6 +172,18 @@ public class SnowflakeDatabase extends AbstractJdbcDatabase {
 
     @Override
     public void setConnection(DatabaseConnection conn) {
+        // If using JDBC connection and OAuth authenticator, wrap it with our OAuth-aware connection
+        if (conn instanceof JdbcConnection) {
+            JdbcConnection jdbcConnection = (JdbcConnection) conn;
+            String url = jdbcConnection.getURL();
+            
+            if (url != null && url.contains("authenticator=oauth")) {
+                if (!(jdbcConnection instanceof SnowflakeOAuthConnection)) {
+                    conn = new SnowflakeOAuthConnection(jdbcConnection.getUnderlyingConnection());
+                }
+            }
+        }
+        
         super.setConnection(conn);
 
         configureSession();
--- a/liquibase-snowflake/src/main/java/liquibase/database/jvm/SnowflakeOAuthConnection.java
+++ b/liquibase-snowflake/src/main/java/liquibase/database/jvm/SnowflakeOAuthConnection.java
@@ -1,0 +1,69 @@
+package liquibase.database.jvm;
+
+import java.sql.Connection;
+import java.util.regex.Matcher;
+import java.util.regex.Pattern;
+
+/**
+ * A JDBC connection wrapper for Snowflake that specifically handles OAuth authentication.
+ * 
+ * This class addresses the issue where JDBC connection metadata returns null for username when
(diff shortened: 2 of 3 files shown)


suryaaki2 (approved):

filipelautert (approved):

SvampX:
I see 2 potential problems:

1. We need to test if we are using `liquibase-pro` this connection will trigger.
2. We need to synchronize this changes with https://github.com/liquibase/liquibase-pro/pull/2218 and https://github.com/liquibase/liquibase-pro/pull/2203

suryaaki2 (approved):

liquibase-snowflake/src/main/java/liquibase/database/jvm/SnowflakeConnectionWrapper.java
```diff
@@ -0,0 +1,50 @@
...
+    }
+
+    @Override
+    public boolean supports(String url) {
+        return url != null && url.startsWith("jdbc:snowflake:");
+    }
+
+    @Override
+    public int getPriority() {
+        return PRIORITY_DATABASE + 100;
```
SvampX, line 31: We don't need such a big priority here. It worked only because of priority was same as here https://github.com/liquibase/liquibase-pro/blob/master/pro/src/main/java/com/datical/liquibase/ext/database/jvm/ProJdbcConnection.java#L28
We need a priority to be higher than current repository's top connection priority. This should be enough and will require to add the same check in https://github.com/liquibase/liquibase-pro/blob/master/pro/src/main/java/com/datical/liquibase/ext/database/jvm/SnowflakeProJdbcConnection.java for a generic case, not for OAuth type authentication, as we are managing all the parameters via driver properties.

SvampX (changes requested):

liquibase-snowflake/src/main/java/liquibase/database/jvm/SnowflakeConnectionWrapper.java
SvampX: @CharlesQueiroz  Got even better solution. We could add this check in the https://github.com/liquibase/liquibase/blob/master/liquibase-standard/src/main/java/liquibase/database/jvm/JdbcConnection.java It will inherently come both to `ProJdbcConnection` and to `SnowflakeProJdbcConnection`. We just need to mark it somehow with a comment not to forget to move to Snowflake extension later.

SvampX: In this case we don't need a connection wrapper and changes in 2 repositories.

filipelautert: @SvampX you mean adding this logic to `open` and `getConnectionUserName` in `JdbcConnection.java` ? So it would be something like that:

```
    public void open(String url, Driver driverObject, Properties driverProperties) throws DatabaseException {
    if (url != null && url.startsWith("jdbc:snowflake:")) {
       this.isOAuth = url.contains("authenticator=oauth");
    } 
    ..... # code that is already there
}
```

SvampX: Right, we have all the properties and url in the `open()` method

filipelautert: @CharlesQueiroz we need to move this code to liquibase-pro class [SnowflakeProJdbcConnection](https://github.com/liquibase/liquibase-pro/blob/master/pro/src/main/java/com/datical/liquibase/ext/database/jvm/SnowflakeProJdbcConnection.java) , otherwise it will conflict with recent snowflake pro implementation.

filipelautert (changes requested):

PR #3926, opened 2023-03-08
only set the execType to RERAN, if the changeset was actually executed

AlexGruebel:
## Impact

- [x] Bug fix (non-breaking change which fixes expected existing functionality)
- [ ] Enhancement/New feature (adds functionality without impacting existing logic)
- [ ] Breaking change (fix or feature that would cause existing functionality to change)
 
## Description

Fixes #3913. 

In the UpdateVisitor class the result of the changeset execution "execType" is automatically set to RERAN, if the previous runstatus of the changeset is not equal to NOT_RUN (even if the changeset got skipped or failed). I changed the logic so that the execType is set to RERAN, only if the changeset executed successfully.

PR #6860, opened 2025-04-02
fix: Prevent duplicate entries in databasechangelog with runAlways and onFail="MARK_RAN"

### LINKED ISSUE #6821: Liquibase inserts duplicate rows in databasechangelog
### Search first

### Description

Liquibase documentation states that rows in databasechangelog will be unique based on id, author and filename. From https://docs.liquibase.com/concepts/tracking-tables/databasechangelog-table.html:

_The table tracks each changeset as a row, identified by a combination of the id, author, and filename columns. There is no primary key on the table. This is to avoid any database-specific restrictions on key lengths. The composite of id, author, and filename is unique across all rows of the table._

This is false, as when I added a unique constraint to the databasechangelog table and re-ran a changelog that contains "runAlways" changeSets, it failed with a duplicate key violation:

liquibase.exception.CommandExecutionException: liquibase.exception.LiquibaseException: liquibase.exception.DatabaseException: liquibase.exception.DatabaseException: ERROR: duplicate key value violates unique constraint "databasechangelog_ui1"
  Detail: Key (id, author, filename)=(430, SYSTEM, JdbcAuditChangeLog.xml) already exists. [Failed SQL: (0) INSERT INTO public.databasechangelog (ID, AUTHOR, FILENAME, DATEEXECUTED, ORDEREXECUTED, MD5SUM, DESCRIPTION, COMMENTS, EXECTYPE, CONTEXTS, LABELS, LIQUIBASE, DEPLOYMENT_ID) VALUES ('430', 'SYSTEM', 'JdbcAuditChangeLog.xml', NOW(), 55, '9:cc72d6f444baf60e74560b719ecc7687', 'createTable tableName=iga_cef_event', '', 'MARK_RAN', NULL, NULL, 'DEV', '2243048625')]

### Steps To Reproduce

1. Create a database named "audit", owned by user "audit" with password "audit"
2. Unzip the attached file demo-duplicate.zip into a working directory
3. cd to demo-duplicate
4. Run the script duplicate-row-defect.sh script

The script will do the following:
1. Run a changelog to create two tables in the audit database
2. Add a unique constraint to databasechangelog on columns (id, author, filename)
3. Run the same changelog again. This time it will fail with the duplicate key error.

### Expected/Desired Behavior

Liquibase should function according to its documentation, i.e. not create duplicate rows for an id, author and filename in databasechangelog.

### Liquibase Version

4.31.0

### Database Vendor & Version

PostgreSQL 15.7

### Liquibase Integration

CLI

### Liquibase Extensions

none

Ubuntu

### Additional Context

_No response_

filipelautert:
When a changeset uses both `runAlways` and `preConditions onFail="MARK_RAN"`,
duplicate entries are created in the databasechangelog table. This happens
because:

1. When a precondition fails, we set `execType = ExecType.MARK_RAN`
2. For MARK_RAN, the `ranBefore` attribute is false
3. MarkChangeSetRanGenerator inserts a new row instead of updating existing one

The fix corrects the logic in ChangeSet execution to properly handle this
combination and prevent multiple entries for the same changeset.

Fixes #6821

### DIFF
--- a/liquibase-standard/src/main/java/liquibase/changelog/ChangeSet.java
+++ b/liquibase-standard/src/main/java/liquibase/changelog/ChangeSet.java
@@ -725,6 +725,9 @@ public class ChangeSet implements Conditional, ChangeLogChild {
                     this.getChangeLog().getSkippedBecauseOfPreconditionsChangeSets().add(this);
                 } else if (preconditions.getOnFail().equals(PreconditionContainer.FailOption.MARK_RAN)) {
                     execType = ExecType.MARK_RAN;
+                    if (this.alwaysRun && databaseChangeLog.getChangeSet(this.getStoredFilePath(), this.author, this.id) != null) {
+                        execType = ExecType.SKIPPED;
+                    }
                     skipChange = true;
 
                     log.info("Marking ChangeSet: \"" + this + "\" as ran despite precondition failure due to onFail='MARK_RAN': " + message);
@@ -753,6 +756,7 @@ public class ChangeSet implements Conditional, ChangeLogChild {
                     this.getChangeLog().getSkippedBecauseOfPreconditionsChangeSets().add(this);
                 } else if (preconditions.getOnError().equals(PreconditionContainer.ErrorOption.MARK_RAN)) {
                     execType = ExecType.MARK_RAN;
+
                     skipChange = true;
 
                     log.info("Marking ChangeSet: " + this + " ran despite precondition error: " + message);


liquibase-standard/src/main/java/liquibase/changelog/ChangeSet.java
rpsandiford, line 1: I ran into what is almost certainly the same issue. I tracked it down to the fix for this issue: https://github.com/liquibase/liquibase/pull/3926. It made this change in UpdateVisitor:
original:
       if (!Objects.equals(runStatus, ChangeSet.RunStatus.NOT_RAN)) {
Modified:
        if (!Objects.equals(runStatus, ChangeSet.RunStatus.NOT_RAN) && Objects.equals(execType, ExecType.EXECUTED)) {
            execType = ChangeSet.ExecType.RERAN;
        }

My resolution was to change the test to this:
        if (!Objects.equals(runStatus, RunStatus.NOT_RAN) 
            && (Objects.equals(execType, ExecType.EXECUTED) || Objects.equals(execType, ExecType.MARK_RAN))) {
            execType = ExecType.RERAN;
        }

Which fix is better should be considered - I don't have the background across the implications to know myself which would be preferred.

rpsandiford:
I ran into what is almost certainly the same issue. It was introduced in the fix for https://github.com/liquibase/liquibase/pull/3926
That fix modified UpdateVisitor.java and changed this:
        if (!Objects.equals(runStatus, ChangeSet.RunStatus.NOT_RAN)) {
            execType = ChangeSet.ExecType.RERAN;
        }
to this:
        if (!Objects.equals(runStatus, ChangeSet.RunStatus.NOT_RAN) && Objects.equals(execType, ExecType.EXECUTED)) {
            execType = ChangeSet.ExecType.RERAN;
        }

We have a local copy of liquibase we maintain (have some customizations we're trying to contribute back - but that's a different story :) ). My fix was this:

        if (!Objects.equals(runStatus, RunStatus.NOT_RAN) 
            && (Objects.equals(execType, ExecType.EXECUTED) || Objects.equals(execType, ExecType.MARK_RAN))) {
            execType = ExecType.RERAN;
        }

i.e. consider a changeset that didn't actually execute, but was set to MARK_RAN, as eligible for execType RERAN. This will wind down into MarkChangeSetRanGenerator and perform the Update (as it did before the fix for 3926) rather than an Insert causing duplicate ID values in the databasechangelog table.

filipelautert:
> I ran into what is almost certainly the same issue. It was introduced in the fix for #3926 That fix modified UpdateVisitor.java and changed this: if (!Objects.equals(runStatus, ChangeSet.RunStatus.NOT_RAN)) { execType = ChangeSet.ExecType.RERAN; } to this: if (!Objects.equals(runStatus, ChangeSet.RunStatus.NOT_RAN) && Objects.equals(execType, ExecType.EXECUTED)) { execType = ChangeSet.ExecType.RERAN; }
> 
> We have a local copy of liquibase we maintain (have some customizations we're trying to contribute back - but that's a different story :) ). My fix was this:
> 
> ```
>     if (!Objects.equals(runStatus, RunStatus.NOT_RAN) 
>         && (Objects.equals(execType, ExecType.EXECUTED) || Objects.equals(execType, ExecType.MARK_RAN))) {
>         execType = ExecType.RERAN;
>     }
> ```
> 
> i.e. consider a changeset that didn't actually execute, but was set to MARK_RAN, as eligible for execType RERAN. This will wind down into MarkChangeSetRanGenerator and perform the Update (as it did before the fix for 3926) rather than an Insert causing duplicate ID values in the databasechangelog table.

Interesting @rpsandiford ! Your fix seems better, I don't like the idea of mine because of the drawbacks outlined in the description. I'll give it a try next week and modify this pr if it works for this case too.

Thanks!

filipelautert:
@rpsandiford it works, thanks!
As you said you have more fixes, please share it back :) You can even put them in a issue body, so we can create the PR and get them merged.

rpsandiford:
The additional things we're trying to contribute back aren't related to this...  So, nothing else to add to this particular issue code-wise.

rpsandiford (approved):

rberezen (approved):

tati-qalified (approved):
Passes manual testing as expected. Note: every consecutive run will update the `orderExecuted` column.

MalloD12 (approved):
**Approved**.

Code change looks good to me. I've added an happy path integration test to verify this scenario works correctly now.

PR #6866, opened 2025-04-03
feat: add support for default value sequence next to hsqldb

filipelautert:
## Impact

## Description

Before this PR, `defaultValueSequenceNext` was documented as unsupported by liquibase for Hsql. But in fact it was _kind of_  supported:

* when used in a `<insert.. <column name="id"  defaultValueSequenceNext="seq"/>` it **worked**
* when used in a `<createTable...  <column name="intA" type="int" defaultValueSequenceNext="tableDefault_seq">` it **generated invalid sql**
* when used in a `<addDefaultValue  ` it throws **an error** saying that it was unsupported.

This PR makes sure that it works for all of those cases now. 

## Additional Context

DAT-19776


liquibase-standard/src/main/java/liquibase/sqlgenerator/core/CreateTableGenerator.java
```diff
@@ -128,44 +119,7 @@ public Sql[] generateSql(CreateTableStatement statement, Database database, SqlG
...
-                    if (statement.getDefaultValue(column).toString().contains("CURRENT USER")) {
-                        buffer.append("SESSION_USER ");
-                    }
-                    if (statement.getDefaultValue(column).toString().contains("CURRENT SQLID")) {
-                        buffer.append("CURRENT SQLID ");
-                    }
-                } else {
-                    buffer.append(statement.getColumnTypes().get(column).objectToSql(defaultValue, database));
-                }
+                handleDefaultValue(statement, database, column, buffer);
```
filipelautert, line 122: Extracting this piece of if code to a new method.

liquibase-standard/src/main/java/liquibase/sqlgenerator/core/CreateTableGenerator.java
```diff
@@ -413,6 +367,54 @@ public Sql[] generateSql(CreateTableStatement statement, Database database, SqlG
...
 
+    private void handleDefaultValue(CreateTableStatement statement, Database database, String column, StringBuilder buffer) {
+        Object defaultValue = statement.getDefaultValue(column);
+        if (database instanceof MSSQLDatabase) {
+            handleMsSqlConstraintForDefaultValue(statement, database, column, buffer);
+        }
+        if (((database instanceof OracleDatabase) || (database instanceof PostgresDatabase) || database.getShortName().equalsIgnoreCase("databricks"))
+                && statement.getDefaultValue(column).toString().startsWith("GENERATED ALWAYS ")) {
+            buffer.append(" ");
+        } else if (database instanceof HsqlDatabase && statement.getDefaultValue(column) instanceof SequenceNextValueFunction) {
```
filipelautert, line 378: Actual changes are in this `if`

liquibase-standard/src/main/java/liquibase/sqlgenerator/core/InsertGenerator.java
```diff
@@ -94,7 +96,11 @@ public void appendValue(StringBuilder sql, Database database, Object newValue) {
                 sql.append(DataTypeFactory.getInstance().getFalseBooleanValue(database));
             }
         } else if (newValue instanceof DatabaseFunction) {
-            sql.append(database.generateDatabaseFunctionValue((DatabaseFunction) newValue));
+            if (newValue instanceof SequenceNextValueFunction && database instanceof HsqlDatabase) {
+                sql.append("NEXT VALUE FOR ").append(((SequenceNextValueFunction) newValue).getValue());
+            } else {
+                sql.append(database.generateDatabaseFunctionValue((DatabaseFunction) newValue));
+            }
```
StevenMassaro, line 103: Would it be better to make this an abstract/default method on either the database or whatever class `SequenceNextValueFunction` implements/extends?

StevenMassaro (approved):

liquibase-standard/src/main/java/liquibase/sqlgenerator/core/InsertGenerator.java
filipelautert: We have `database.sequenceNextValueFunction` that is used by `database.generateDatabaseFunctionValue` in a generic way, but sadly seems HSQL database it's the only db that uses different expressions to do the samething when used in different contexts. So I would go with this if now, but _if_ another db needs it we can create another  abstract/default method to handle it.

rberezen (approved):

liquibase-integration-tests/src/test/resources/changelogs/common/common.tests.changelog.xml
```diff
@@ -1007,7 +1007,7 @@
   </changeSet>
 
   <!-- should work for all databases but only tested on h2 and psotgres for now, oracle actually does not support sequences as default values -->
```
tati-qalified, line 1009: a bit nitpicky, but this comment should be changed to include hsqldb as well

tati-qalified (approved):
Tested manually, works as expected.

Issue #6850, opened 2025-03-28
The commandDefinitions cache in the CommandFactory cannot be used.


FatovDI:
### Search first

- [x] I searched and no similar issues were found

### Description

The CommandFactory has the commandDefinitions cache:

`private final Map<String[], CommandDefinition> commandDefinitions = new ConcurrentHashMap<>();`

This cache uses in getCommandDefinition method:

`    public CommandDefinition getCommandDefinition(String... commandName) throws IllegalArgumentException{
        CommandDefinition commandDefinition = commandDefinitions.get(commandName);
        if (commandDefinition == null) { //Check if we have already computed arguments, dependencies, pipeline and adjusted definition
            commandDefinition = new CommandDefinition(commandName);
            computePipelineForCommandDefinition(commandDefinition);
            consolidateCommandArgumentsForCommand(commandDefinition);
            adjustCommandDefinitionForSteps(commandDefinition);
            commandDefinitions.put(commandName, commandDefinition);
        }
        return commandDefinition;
    }`

But, this cache cannot be used because this one uses the key as the array of String. Method `getCommandDefinition` has the vararg type, thats why condition  `if (commandDefinition == null)` is always true. In this case, each called method creates a new entry in the commandDefinitions and renders the cache ineffective.

Consequences:

*   Cache Misses on Every Call: The if (commandDefinition == null) condition in getCommandDefinition is *always* true.
*   Unnecessary Computations:  The expensive operations within the if block (creating a new CommandDefinition, calling computePipelineForCommandDefinition, consolidateCommandArgumentsForCommand, and adjustCommandDefinitionForSteps) are performed repeatedly for the same command definitions.
*   Memory Overhead: The commandDefinitions cache grows unbounded, storing duplicate entries for the same commands. This leads to increased memory consumption.

### Steps To Reproduce

'CommandFactory.getInstance().getCommandDefinition(*UpdateCommandStep.COMMAND_NAME)'
Each call this method returns a new instance CommandDefinition.

Due to the nature of varargs, *each call* to getCommandDefinition with the same command name arguments *creates a new String[] instance*.  Even if the content of the String[] is identical, the get() method on the ConcurrentHashMap will return null because it compares array *instances*, not array *content*.

### Expected/Desired Behavior

CommandFactory.getInstance().getCommandDefinition(*UpdateCommandStep.COMMAND_NAME)'
Each call this method returns existed CommandDefinition from the cache by name `UpdateCommandStep.COMMAND_NAME`.

[...]

PR #6872, opened 2025-04-04
fix: change String[] to List<String> in COMMAND_DEFINITIONS cache

FatovDI:
## Impact

## Description
Fixes https://github.com/liquibase/liquibase/issues/6850


FatovDI:
Fixes #6850

filipelautert (approved):
Thanks for the pr @FatovDI ! I just changed the test as I tried it with the old code and it was still working.

tati-qalified (approved):

filipelautert:
@FatovDI I had to remove 2 already existing tests that started to fail after the changes in this PR. I'm still trying to understand why they failed, any ideas?

FatovDI:
> I had to remove 2 already existing tests that started to fail after the changes in this PR. I'm still trying to understand why they failed, any ideas?

Yes, of course. I have an idea that in the regester and the unregester methods we schould work with COMMAND_DEFINITIONS cache too. Add a value to the COMMAND_DEFINITIONS cache in the regester method and remove a value from the COMMAND_DEFINITIONS cache. I have added new commits, could you check it?

rberezen (approved):

PR #6882, opened 2025-04-07
Try to implement fix for #6805

### LINKED ISSUE #6805: loadData creates invalid JDBC requests
### Search first

### Description

The way how syntax errors and computed values in CSV input for `loadData` are handled and how prepared statements are created leads to JDBC errors, which are hard to understand for users.

### Steps To Reproduce

Update a database schema with the following change set:

    <databaseChangeLog
        xsi:schemaLocation="http://www.liquibase.org/xml/ns/dbchangelog http://www.liquibase.org/xml/ns/dbchangelog/dbchangelog-3.5.xsd
        http://www.liquibase.org/xml/ns/dbchangelog-ext http://www.liquibase.org/xml/ns/dbchangelog/dbchangelog-ext.xsd">
    
    	<changeSet id="reproducer" author="test" runAlways="true">
    	
    		<createTable tableName="REPRODUCER">
    			<column name="DATE" type="DATE"/>
    			<column name="NAME" type="VARCHAR2(32)"/>
    		</createTable>
    
    		<loadData tableName="REPRODUCER" file="reproducer.csv" relativeToChangelogFile="true">
    			<column name="DATE"/>
    			<column name="NAME"/>
    		</loadData>
    
    	</changeSet>
        
    </databaseChangeLog>

With this CSV file:

    DATE,NAME
    xxx,Alice
    2025-03-12,Bob

Note the syntax error in the first column of the first row.

### Actual Behaviour

The outcome is the following JDBC exception, when the second(!) row is processed:

    liquibase.exception.DatabaseException: java.sql.SQLException: ORA-17003: Invalid column index https://docs.oracle.com/error-help/db/ora-17003/
            at liquibase.statement.ExecutablePreparedStatementBase.execute(ExecutablePreparedStatementBase.java:123)
            at liquibase.executor.jvm.JdbcExecutor.execute(JdbcExecutor.java:175)
            at liquibase.executor.AbstractExecutor.execute(AbstractExecutor.java:141)
            at liquibase.database.AbstractJdbcDatabase.executeStatements(AbstractJdbcDatabase.java:1189)
            at liquibase.changelog.ChangeSet.execute(ChangeSet.java:790)

### Expected/Desired Behavior

An exception which is points to the invalid input ('xxx') is created.  

### Liquibase Version

4.31.0

### Database Vendor & Version

Oracle 19c

### Liquibase Integration

CLI

### Liquibase Extensions

_No response_

Windows 11

### Analysis

From my understanding here is what happens:

1. The `ColumnConfig` class sets a value for `valueComputed`, if a date cannot be parsed (comment: "probably a function").
2. Those computed values are not considered as statement parameters when
  a. generating the SQL statement in `InsertExecutablePreparedStatement.generateSql()`
  b. populating parameters for each row in `ExecutablePreparedStatementBase.attachParams()`
4. The method `BatchDmlExecutablePreparedStatement.generateSql()` uses the SQL for the statement generated for the first data row which then re-used with the parameters for each row. The comment in this method says "By convention, all of the statements are the same except the bind values." -- which does not hold true due to the previous point.
5. Due to the behavior described in 2. each row may have a different number of parameters and may not be consistent with the prepared statement from the first row leading to different JDBC errors.

Note: There are more situations where this breaks, also for example if a syntactically correct value computation is given for some columns in the CSV file. See assignments to `valueComputed` in `ColumnConfig`

marchof:
## Impact

## Description

Trying to fix #6805.


marchof:
Hi @filipelautert, can you please enable the GitHub action "Unit and Integration Tests" so that I can see whether I can reproduce the issue with the integration test added by me ("test first")?

Unfortunatelly I cannot [run this in my GitHub context](https://github.com/marchof/liquibase/actions/runs/14312098583) as the tests seems to depend on a proprietary service I don't have credentials for:

> The workflow is not valid. .github/workflows/run-tests.yml (Line: 286, Col: 3): Error calling workflow 'liquibase/build-logic/.github/workflows/fossa_ai.yml@main'. The nested job 'fossa-scan' is requesting 'id-token: write', but is only allowed 'id-token: none'. .github/workflows/run-tests.yml (Line: 286, Col: 3): Error calling workflow 'liquibase/build-logic/.github/workflows/fossa_ai.yml@main'. The nested job 'generate-oss-pro-sbom-reports' is requesting 'id-token: write', but is only allowed 'id-token: none'.

filipelautert:
> Hi @filipelautert, can you please aprove the GitHub action "Unit and Integration Tests" so that I can see whether I can reproduce the issue with the integration test added by me ("test first")?

np, running!

marchof:
The reproducer now shows the same problem for Oracle when an invalid value `xxx` is provided for a date column:

    Error:    OracleIntegrationTest>AbstractIntegrationTest.testUpdateTwice:547 » CommandExecution liquibase.exception.LiquibaseException: liquibase.exception.MigrationFailedException: Migration failed for changeset changelogs/common/common.tests.changelog.xml::shouldHandleComputedAndNonComputedColumnValues::marchof:
         Reason: liquibase.exception.DatabaseException: java.sql.SQLException: Invalid column index

filipelautert:
> The reproducer now shows the same problem for Oracle when an invalid value `xxx` is provided for a date column:
> 
> ```
> Error:    OracleIntegrationTest>AbstractIntegrationTest.testUpdateTwice:547 » CommandExecution liquibase.exception.LiquibaseException: liquibase.exception.MigrationFailedException: Migration failed for changeset changelogs/common/common.tests.changelog.xml::shouldHandleComputedAndNonComputedColumnValues::marchof:
>      Reason: liquibase.exception.DatabaseException: java.sql.SQLException: Invalid column index
> ```

Looks like the xxx is being interpreted as a column instead of throwing a date parsing error.

marchof:
What I think what is happening is:

* "xxx" fails to parse as date, therefore the column is [considered computed](https://github.com/liquibase/liquibase/blob/master/liquibase-standard/src/main/java/liquibase/change/ColumnConfig.java#L402)
* For computed values [no parameter is inserted](https://github.com/liquibase/liquibase/blob/master/liquibase-standard/src/main/java/liquibase/statement/InsertExecutablePreparedStatement.java#L41)
* Only [the first SQL is used for a prepared batch statement](https://github.com/liquibase/liquibase/blob/master/liquibase-standard/src/main/java/liquibase/statement/BatchDmlExecutablePreparedStatement.java#L59), assuming you can use the same prepared statement for all rows.
* Now when the parameters for the second row are populated you get the index error (two instead of one parameter).

Since computed values have been introduced the statement "By convention, all of the statements are the same except the bind values" does now hold true any more. Depending on whether computed values are included in a row or not the SQL statement and the number of parameters my vary.

marchof:
Probably the fix is to re-work BatchDmlExecutablePreparedStatement that it stops the batch as soon as the generated SQL would be different.

MalloD12:
Hi @marchof,

Are you planning to refactor  `BatchDmlExecutablePreparedStatement`, please let us know if there is anything we can help you with.

Thanks,
Daniel.

marchof:
@MalloD12 Thanks for offering your help! So far I just created a regression test. I need to find time to understand the `SqlStatement` class hierarchy and see how things interact. Is there any additional documentation that could help here?

MalloD12:
Hi @marchof,

As far as I can see on our internal docs, I think we don't have something like that, but I would encourage you to post your questions here, and if I cannot answer your questions, I'll share them with the team, and we will provide you with an answer for them.

Thanks,
Daniel.

MalloD12:
Hey @marchof,

I could see your test changeset added to `common.tests.changelog.xml` is successfully reproducing the issue.

Thanks,
Daniel.

PR #6901, opened 2025-04-15
feat: Add PostgreSQL Index Function Support via USING Clause

filipelautert:
## Impact

## Description

This enhancement allows specifying index types like btree, gin, gist, etc. when creating/snapshoting indexes in PostgreSQL, providing better control over index behavior and performance.

1. Adds a `using` property to `CreateIndexChange` and `Index` classes
2. Enhances JDBC snapshot capabilities to capture index types from PostgreSQL
3. Updates `CreateIndexGeneratorPostgres` to include the USING clause in SQL generation
4. Extends schema definitions by adding the `using` attribute to createIndex in XML
5. Updates related generators and statement classes to support the new property

### DIFF
--- a/liquibase-standard/src/main/java/liquibase/change/core/CreateIndexChange.java
+++ b/liquibase-standard/src/main/java/liquibase/change/core/CreateIndexChange.java
@@ -230,7 +230,8 @@ public class CreateIndexChange extends AbstractChange implements ChangeWithColum
     @Override
     public String[] getExcludedFieldFilters(ChecksumVersion version) {
         return new String[]{
-                "associatedWith"
+                "associatedWith",
+                "using"
         };
     }
 }
--- a/liquibase-standard/src/main/java/liquibase/change/core/CreateIndexChange.java
+++ b/liquibase-standard/src/main/java/liquibase/change/core/CreateIndexChange.java
@@ -44,6 +44,8 @@ public class CreateIndexChange extends AbstractChange implements ChangeWithColum
     private String associatedWith;
     @Setter
     private Boolean clustered;
+    @Setter
+    private String using;
 
 
     public CreateIndexChange() {
@@ -107,6 +109,7 @@ public class CreateIndexChange extends AbstractChange implements ChangeWithColum
                         getTableName(),
                         this.isUnique(),
                         getAssociatedWith(),
+                        getUsing(),
                         getColumns().toArray(new AddColumnConfig[0]))
                         .setTablespace(getTablespace())
                         .setClustered(getClustered())
@@ -177,6 +180,11 @@ public class CreateIndexChange extends AbstractChange implements ChangeWithColum
     @DatabaseChangeProperty(description = "Whether to create a clustered index")
     public Boolean getClustered() {
         return clustered;
+    }
+
+    @DatabaseChangeProperty(description = "The index type to use. For example, 'btree' or 'gin'")
+    public String getUsing() {
+        return using;
     }
 
     @Override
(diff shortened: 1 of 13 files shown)


filipelautert:
TH execution: https://github.com/liquibase/liquibase-test-harness/actions/runs/14477576248/job/40606890121

liquibase-standard/src/main/java/liquibase/change/core/CreateIndexChange.java
```diff
@@ -179,6 +182,11 @@ public Boolean getClustered() {
         return clustered;
     }
 
+    @DatabaseChangeProperty(description = "The index type to use. For example, 'btree' or 'gin'")
```
StevenMassaro, line 185: Is this a breaking change for checksums?

liquibase-standard/src/main/java/liquibase/snapshot/JdbcDatabaseSnapshot.java
```diff
@@ -276,9 +276,46 @@ public List<CachedRow> fastFetch() throws SQLException, DatabaseException {
...
+                        }
+
+                        List<CachedRow> usingTypeInformation = executeAndExtract(sql.toString(), database);
+
+                        for (CachedRow row : usingTypeInformation) {
+                            // Postgres returns the index type for the using field, so we need to set it to "INDEX_TYPE" in the matching entries from the returnList
+                            for (CachedRow returnRow : returnList) {
+                                if (returnRow.getString("INDEX_NAME").equalsIgnoreCase(row.getString("INDEX_NAME"))
+                                    && returnRow.getString("TABLE_NAME").equalsIgnoreCase(row.getString("TABLE_NAME"))
+                                    && returnRow.getString("TABLE_SCHEM").equalsIgnoreCase(row.getString("TABLE_SCHEMA"))) {
```
StevenMassaro, line 310: Can't we do this filtering in the SQL statement, instead of in memory in Java? This could have performance implications for large result sets.

liquibase-standard/src/main/java/liquibase/change/core/CreateIndexChange.java
filipelautert: I added this new parameter to the excluded fields to prevent this down below in this same file -> https://github.com/liquibase/liquibase/pull/6901/files#diff-f1647dc293a46d72c22151fa77240ea144cc38788c2af645e71b1e74987e30a2R234

liquibase-standard/src/main/java/liquibase/snapshot/JdbcDatabaseSnapshot.java
filipelautert: The filtering is done in the SQL when we are not in `isBulkFetchMode` .  In this mode the only other option would be to run this query for every single `index`... we would save memory but perform many select requests instead.

StevenMassaro (approved):

filipelautert:
> @filipelautert this looks ok to me. I think we could add at least a basic test coverage using a JSON changelog format and specifying a GIN index. I can help with that if you wish.

@MalloD12 I'm not sure if it's required, as there is an integration tests for snapshot and a SQL generator for generation... both of them should be enough as we are covering both ways.

MalloD12:
@filipelautert Ah, ok, I thought it could be nice to have an additional integration test scenario (for Postgres) that makes use of the using clause with `GIN` value, and maybe could have a negative test as well to validate warning message displays when trying to use it for any other DB.

Issue #6820, opened 2025-03-17
Liquibase fails on second run attempting to create databasechangelog table.


sresnik-k-care:
### Search first

- [x] I searched and no similar issues were found

### Description

I have been using Liquibase for quite a while on MS SQL Server but my experience with Azure SQL Server Managed Instances not so much.  While experimenting with moving to Azure, I noticed Liquibase would fail on the second run because it would fail while attempting to create the databasechangelog table when one already exists.  This only appears to happen on the `master` database.

The failure is very much similar to #6666 .  However the root cause seems to be the result of [this query](https://github.com/liquibase/liquibase/blame/12ab20b176e5659622c240b4ba4f056ccf17c27d/liquibase-standard/src/main/java/liquibase/snapshot/JdbcDatabaseSnapshot.java#L1250).  The query never returns anything for the databasechangelog table.  After some experimentation it appears that including the `db_id('database_name')` as parameters to the `object_schema_name` method calls within the query are causing the problem.  Removing those allow the query to return appropriately.

### Steps To Reproduce

Using the project here: https://github.com/sresnik-k-care/simple-liquibase
1. Starting with a fresh Azure SQL Managed Instance.
2. Update necessary properties in the project.
3. Run the linked project's "update" Gradle task.
4. Note that it runs successfully.
5. Run the "update" task again.
6. Note that it fails attempting to create the `DATABASECHANGELOG` again as Liquibase failed to find it.

### Expected/Desired Behavior

The expectation was that the databasechangelog would be found, and read on the second run.  This failure does not happen with local Docker based MS SQL Server instances.

### Liquibase Version

4.31.1

### Database Vendor & Version

Microsoft SQL Server (ver. 12.00.2000) Case sensitivity: plain=mixed, delimited=mixed Driver: Microsoft JDBC Driver 12.8 for SQL Server (ver. 12.8.1.0, JDBC4.3) Effective version: Azure SQL Database (ver. 12.0.2000)

### Liquibase Integration

Gradle 8.13

### OS and/or Infrastructure Type/Provider

Ubuntu and MacOS

### Are you willing to submit a PR?

- [x] I'm willing to submit a PR (Thank you!)

tati-qalified:
[...]

PR #6902, opened 2025-04-15
gh-6820 Fix metadata queries returning null in Azure SQL Managed Instances.

sresnik-radicle:
## Impact

## Description

This is a fix for #6820 .  In the metadata queries for the `DATABASECHANGELOG` table, the db_id is passed in as a parameter to the `object_name` and `object_schema_name` functions in the query for table meta data, and column meta data.  This results in an empty result set, and a failure when executing Liquibase.

## Things to be aware of
N/A

## Things to worry about
N/A

## Additional Context

This issue was discovered by running against the `master` database on an Azure SQL Managed Instance more than once.  The first time, the `DATABASECHANGELOG` table is created correctly, subsequent runs fail when Liquibase tries to create the table again.

### DIFF
--- a/liquibase-standard/src/main/java/liquibase/snapshot/JdbcDatabaseSnapshot.java
+++ b/liquibase-standard/src/main/java/liquibase/snapshot/JdbcDatabaseSnapshot.java
@@ -697,10 +697,12 @@ public class JdbcDatabaseSnapshot extends DatabaseSnapshot {
                     databasePrefix = "[" + databaseName + "].";
                 }
 
+                String jdbcSchemaName =
+                    ((AbstractJdbcDatabase) database).getJdbcSchemaName(catalogAndSchema);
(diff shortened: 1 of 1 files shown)


sresnik-k-care:
I would appreciate a little guidance on any tests that should be added for this change.

filipelautert (commented):
Hello @sresnik-radicle !
Sorry for the long silence here. Two questions before this can move:
object_name(c.object_id) lost its db_id parameter in the select, filter and order by, so cross database snapshots would resolve names in the wrong database. Was that intentional?
Does object_schema_name return null on Managed Instances for objects in other schemas too? If so the isnull fallback in the where clause would pull other schemas' objects into the snapshot, maybe the fallback should only apply when we detect a Managed Instance.
For tests, a unit test asserting the generated SQL for the MI case is enough to start.

PR #6903, opened 2025-04-15
DAT-19917: Refactor analytics setup to prepare for pro analytics

abrackx:
## Impact

## Description
Functionally this should be no different than before.

Sets up mechanism to override enabling analytics behavior.

### DIFF
--- a/liquibase-standard/src/main/java/liquibase/analytics/AnalyticsListener.java
+++ b/liquibase-standard/src/main/java/liquibase/analytics/AnalyticsListener.java
@@ -19,7 +19,5 @@ public interface AnalyticsListener extends Plugin {
      */
     void handleEvent(Event event) throws Exception;
 
-    default boolean isEnabled() {
-        return true;
-    }
+    boolean isEnabled();
 }
--- a/liquibase-standard/src/main/java/liquibase/analytics/AnalyticsListener.java
+++ b/liquibase-standard/src/main/java/liquibase/analytics/AnalyticsListener.java
@@ -18,4 +18,8 @@ public interface AnalyticsListener extends Plugin {
      * @param event the event to be handled.
      */
     void handleEvent(Event event) throws Exception;
+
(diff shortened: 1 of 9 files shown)


liquibase-standard/src/main/java/liquibase/analytics/AnalyticsListener.java
```diff
@@ -18,4 +18,8 @@ public interface AnalyticsListener extends Plugin {
      * @param event the event to be handled.
      */
     void handleEvent(Event event) throws Exception;
+
+    default boolean isEnabled() {
+        return true;
+    }
```
StevenMassaro, line 24: I would not make this a default method, but instead make implementing classes provide implementations.

liquibase-standard/src/main/java/liquibase/analytics/configuration/AnalyticsArgs.java
```diff
@@ -29,7 +26,7 @@ public class AnalyticsArgs implements AutoloadedConfigurations {
         ConfigurationDefinition.Builder builder = new ConfigurationDefinition.Builder("liquibase.analytics");
 
         ENABLED = builder.define("enabled", Boolean.class)
-                .setDescription("Enable or disable sending product usage data and analytics to Liquibase. Learn more at https://docs.liquibase.com/analytics. DEFAULT: true for OSS users | false for PRO users")
+                .setDescription("Enable or disable sending product usage data and analytics to Liquibase. Learn more at https://docs.liquibase.com/analytics.")
```
StevenMassaro, line 29: It might be useful to provide a default value here, even if it is just symbolic. This way users won't be surprised to learn that on is the default.

abrackx: Resolved that I'm going to keep this as is over introducing drift in the pro repository

filipelautert (approved):

PR #6906, opened 2025-04-18
feat: Override isInternal method to return false in InitCopyCommandStep

filipelautert:
## Impact

## Description

Caused by https://github.com/liquibase/liquibase/pull/6899 .


liquibase-standard/src/main/java/liquibase/command/core/init/InitCopyCommandStep.java
```diff
@@ -81,4 +81,9 @@ public static ProjectCopier determineProjectCopier(CommandScope commandScope, St
         }
         return copier;
     }
+
+    @Override
+    public boolean isInternal() {
+        return false;
```
abrackx, line 87: I think this should be marked as internal.

OliaLub (approved):

Issue #6788, opened 2025-03-04
Liquibase does not handle alternative syntax options for TIMESTAMP WITH TIMEZONE for all database types


deg-hrisser:
### Search first

- [x] I searched and no similar issues were found

### Description

I was investigating differences in behaviour between two projects persisting `OffsetDateTime`s as `TIMESTAMP WITH TIME ZONE` in an Oracle Database, when I discovered that the project using Liquibase actually created `TIMESTAMP`-columns instead.

-----

In my testing, I seem to be able to confirm that the following column definitions create simple `TIMESTAMP`-columns with Oracle: ⚠

```xml
<column name="offset_date_time" type="timestamptz"/>
<column name="offset_date_time" type="java.sql.types.timestamp_with_timezone"/>
```

-----

and this definition causes an error, preventing table creation completely (`ORA-02000: missing TIME keyword`): ❌

```xml
<column name="offset_date_time" type="java.sql.Types.TIMESTAMP_WITH_TIMEZONE"/>
```

-----

On the contrary, these similar definitions correctly create a `TIMESTAMP WITH TIME ZONE`: ✔

```xml
<column name="offset_date_time" type="timestamp with time zone"/>
<column name="offset_date_time" type="TIMESTAMP WITH TIME ZONE"/>
```

### Overview

| type definition                        | result Oracle      | result PostgreSQL  | result H2 database |
| -------------------------------------- | ------------------ | ------------------ | ------------------ |
| timestamp with time zone               | ✔ with timezone    | ✔ with timezone    | ✔ with timezone    |
| TIMESTAMP WITH TIME ZONE               | ✔ with timezone    | ✔ with timezone    | ✔ with timezone    |
| java.sql.Types.TIMESTAMP_WITH_TIMEZONE | ❌ ORA-02000        | ✔ with timezone    | ✔ with timezone    |
| java.sql.types.timestamp_with_timezone | ⚠ without timezone | ⚠ without timezone | ⚠ without timezone |
| timestamptz                            | ⚠ without timezone | ✔ with timezone    | ⚠ without timezone |

### Steps To Reproduce

See above.

- Create a minimal changeset creating a table with a column.
- Use the column type according to the table above
- Let liquibase create the database tables
- Check the resulting database column types or check the executed SQL via logging on `liquibase.executor`

### Expected/Desired Behavior

I believe that all type definitions in the table above should be mapped to `TIMESTAMP WITH TIME ZONE` in all databases that support it.

[...]

PR #6908, opened 2025-04-20
fix: ensure time zone information of timestamptz on creating TimestampType

### LINKED ISSUE #6551: Diff generates a timestamp without considering the time zone
### Search first

### Description

On creating a `diff` between two databases, when the reference database has `timestamp with timezone`, and the target one has `string`, the generated commands has `timestamp` without the time zone.

### Steps To Reproduce

- In the target database, create a table with a column of type `timestamp with time zone`, e.g. `CREATE TABLE telegram (message_time timestamp with time zone);`
- In the reference database, create a table with a column of type `varchar`, e.g. `CREATE TABLE telegram (message_time varchar);`
- Run the `diff change` command

### Expected/Desired Behavior

The output should have a newDataType of `timestamp with time zone` instead of `timestamp`

### Liquibase Version

4.30.0

### Database Vendor & Version

PostgreSQL 16

### Liquibase Integration

API (DiffChangelogCommandStep)

### Liquibase Extensions

_No response_

Windows 11

### Additional Context

_No response_

asashour:
Fixes #6551

## Impact

## Description

On creating a `TimestampType` from `timestamptz` (in `DataTypeFactory`), the time zone is lost from the time stamp.

This PR ensures that the time zone additional information is added.

## Things to be aware of

There was already a similar code of deciding the additional information in `TimestampType`, which is extracted into a public static method, which is used by `DataTypeFactory`.

## Things to worry about

The additional verification in `DataTypeFactoryTest` shows that when the description contains "with timezone" for PostgreSQL/H2, the output also includes "with timezone", however, there should be space between `time` and `zone`. This is probably an incorrect input, but that's not the scope of this PR.


liquibase-standard/src/main/java/liquibase/datatype/DataTypeFactory.java
```diff
@@ -179,11 +176,14 @@ public LiquibaseDataType fromDescription(String dataTypeDefinition, Database dat
                 additionalInfo = splitTypeName[1];
             }
         }
+        if (dataTypeName.toLowerCase(Locale.US).equals("timestamptz")) {
```
MalloD12, line 179: Hi @asashour,

Overall, changes look good to me, but I was wondering whether we should be modifying this `IF` block, because if I'm not wrong `timestamptz` column type is `Postgres` specific, but I guess if we do that it would imply a little refactoring on the `getTimeZoneAdditionInformation()`. What do you think?

cc: @filipelautert

asashour: Thanks for the feedback.

It seems that `timestamptz` is relevant to PostgreSQL (not to Oracle for instance), so yes, let's do it for only it for now.

However, it may be worth mentioning that the original issue happens with also Oracle, but it is due to the fact that [this line](https://github.com/liquibase/liquibase/blob/master/liquibase-standard/src/main/java/liquibase/datatype/DataTypeFactory.java#L124) changes `TIMESTAMP(6) WITH TIME ZONE(13)` to `TIMESTAMP`, thus removing the `time zone`. This should be also fixed, possibly in another PR.

And this also adds the fact that possibly other databases needs to be tested as well.

MalloD12: @asashour, one other option is if you want to merge the code changes made by you here with the ones I included in #6929, or I can do that and include both here.

Thanks,
Daniel.

asashour: Thanks, I think we shouldn't handle both PRs at the same time. Once the other PR is merged, I would resolve the conflict, let's keep this PR for PostgreSQL, and the other one for Oracle.

MalloD12 (approved):
**Approved**.

Code changes here looks good to me. As discussed previously we have made an agreement to deel with this PR and #6929 separately, as they are related and touching a related logic we will have some conflicts to resolved when merging them. We will first merged #6929 and then this one.

Thanks @asashour for your contribution!

Daniel.

MalloD12:
This is related to #6929 (which should be merged first).

filipelautert (approved):

tati-qalified (changes requested):
Hi @asashour, thank you for your contribution! I've been testing this manually with H2, Oracle, Postgres and MySQL. 

Used the changeset
```
    <changeSet id="create_table_with_timestamp" author="tati">
        <createTable tableName="testTimeZone">
            <column name="timestamptezeta" type="timestamptz" />
            <column name="java_lowercase" type="java.sql.types.timestamp_with_timezone" />
            <column name="java_uppercase" type="java.sql.Types.TIMESTAMP_WITH_TIMEZONE" />
            <column name="timestamp_w_t_z_lower" type="timestamp with time zone" />
            <column name="timestamp_w_t_z_upper" type="TIMESTAMP WITH TIME ZONE" />
        </createTable>
    </changeSet>
```
Ran `liquibase update-sql` to see the generated SQL script, then ran `liquibase update` to verify correctness.

Here are some of my findings:

1. This PR reverts some changes made in #6929: the datatype for Oracle had been changed to `with time zone`, but with these changes it goes back to `with timezone`, which is incorrect.
Generated script for Oracle:
```
CREATE TABLE DBUSER.testTimeZone (
    timestamptezeta TIMESTAMP WITH TIMEZONE,
    java_lowercase TIMESTAMP,
    java_uppercase TIMESTAMP WITH TIMEZONE,
    timestamp_w_t_z_lower TIMESTAMP with time zone,
    timestamp_w_t_z_upper TIMESTAMP WITH TIME ZONE
);
```

2. For Postgres and Oracle, the data type `java.sql.types.timestamp_with_timezone` in lowercase will give the wrong output, but using uppercase (`java.sql.Types.TIMESTAMP_WITH_TIMEZONE`) generates the correct output.
With lowercase, Oracle generates `TIMESTAMP` and Postgres generates `TIMESTAMP WITHOUT TIME ZONE` - this exists, but it makes no sense that `with_timezone` is creating `WITHOUT TIME ZONE`.
With uppercase, both DBs generate `TIMESTAMP WITH TIME ZONE` (but with the error in point 1 for Oracle).

3. Lastly, for MySQL, using the data types `timestamp with time zone` or `TIMESTAMP WITH TIME ZONE` outputs columns with those exact data types, which do not exist in MySQL syntax. The correct output would be just `TIMESTAMP`.
Generated output:
```
CREATE TABLE lbcat.testTimeZone (
    timestamptezeta timestamp NULL,
    java_lowercase timestamp NULL,
    java_uppercase timestamp NULL,
    timestamp_w_t_z_lower timestamp with time zone NULL,
    timestamp_w_t_z_upper TIMESTAMP WITH TIME ZONE NULL
);
```

Could you please look into these issues?
Let us know if you need any guidance.
Thank you!
Tatiana

asashour:
Thanks for your feedback.

This PR changes `main` and touches only PostgreSQL, this PR would have conflict with #6929 (which should be merged first).

I don't consider this as a revert, because the scope of the PR is to retain all other DBs, and change for only PostgreSQL, so once the other ticket is merged, this PR would be changed to consider those changes as well.

MalloD12:
Hey @asashour,

Last week I worked on #6929. Could you please have a look at that PR and make any additional changes if required? If there is nothing and everything is covered, would you mind closing this one?

Thanks,
Daniel.

asashour:
As of the current `master`, the issue #6551 seems to still happen with both Oracle and PostgreSQL. I believe that is a different scenario than #6788.

This PR is still valid for PostgreSQL.

MalloD12:
I think overall of the changes made here are already covered by #6929. I think this PR should be close because of the following reasons:

1. Redundant logic: The functionality is already fully implemented in `TimestampType.toDatabaseDataType()` that is already in master
2. Adds complexity Without Benefit: Pre-setting additionalInfo in DataTypeFactory adds an extra step that gets overwritten anyway.
3. The helper method has a bug: Because for `Oracle` it sets additioanl info `"WITH TIMEZONE"`, when it should set `WITH TIME ZONE`. 
 4. Code Maintenance: Having timezone logic in two places (DataTypeFactory AND TimestampType) makes the code harder to maintain

I think we should close this one, but before doing that could you @asashour provide the steps to reproduce the issue you are still seeing? cc: @filipelautert 

Thanks,
Daniel.

asashour:
Closing, but it seems that the issue still exists. Probably another ticket would be opened.

PR #6916, opened 2025-04-23
Added new argument to AbstractFormattedChangeLogParser.handleAdditionalLines method

wwillard7800:
Also added a field to DTOs to allow type of content detection

INT-1217

## Impact


filipelautert (approved):
There is a change to a protected method signature, but i guess we can't call it breaking change. Or could we? could this be used in other extensions like... mongo?

wwillard7800:
> There is a change to a protected method signature, but i guess we can't call it breaking change. Or could we? could this be used in other extensions like... mongo?

Yes, and I have communicated with @KushnirykOleh about this change.

liquibase-standard/src/main/java/liquibase/parser/AbstractFormattedChangeLogParser.java
```diff
@@ -761,7 +761,7 @@ protected static void resetSequences(StringBuilder currentSequence, StringBuilde
         currentRollbackSequence.setLength(0);
     }
 
-    protected boolean handleAdditionalLines(DatabaseChangeLog changeLog, ResourceAccessor resourceAccessor, String line)
+    protected boolean handleAdditionalLines(DatabaseChangeLog changeLog, ResourceAccessor resourceAccessor, String line, StringBuilder currentSequence)
```
filipelautert, line 764: Would be good to keep and deprecate the old method signature as this may be used by some extensions.

filipelautert (changes requested):

filipelautert (approved):

PR #6920, opened 2025-04-29
fix: Change the rename column query for mysql v8 and later using rename column to

### LINKED ISSUE #6876: RenameColumn silently drops default value in MySQL
### Search first

### Description

If a column with a default value is renamed via the change type `renameColumn`, the default value is silently dropped when using MySQL. I assume this is also happening for MariaDB but I have not tested it.

The problem is that a changeset like this:
```  
- changeSet:
    id: mychangeset-001
    author: dilbert
    changes:
      - renameColumn:
          tableName: mytable
          columnDataType: tinyint
          oldColumnName: old_name
          newColumnName: new_name
```
is translated to this for MySQL, thereby dropping the default value:
```
ALTER TABLE mytable CHANGE old_name new_name TINYINT;
```

**Solution**

Since MySQL 8.X and MariaDB 10.5.2 there is a new command ALTER TABLE ... RENAME which should be used when the database target is one of those version (or newer).

Here are the docs for RENAME COLUMN: [MariaDB](https://mariadb.com/kb/en/alter-table/#rename-column), [MySQL](https://dev.mysql.com/doc/refman/8.0/en/alter-table.html)

### Steps To Reproduce

Create table with a column that has a default value

```
create table mytable
(
    old_name tinyint default 0 null,
);
```

Run this changeset
```  
- changeSet:
    id: mychangeset-001
    author: dilbert
    changes:
      - renameColumn:
          tableName: mytable
          columnDataType: tinyint
          oldColumnName: old_name
          newColumnName: new_name
```

After running the changeset, the table defintion looks like this:
```
create table mytable
(
    new_name tinyint null,
);
```

### Expected/Desired Behavior

The default value should not be changed when using the change type `renameColumn`

### Liquibase Version

4.31.1 #6739

### Database Vendor & Version

MySQL 8.0.36

### Liquibase Integration

maven

### Liquibase Extensions

_No response_

AWS RDS

### Additional Context

_No response_

quangdutran:
## Impact

Fixes #6876

- [x ] Bug fix (non-breaking change which fixes expected existing functionality)
 

## Description

From MySQL v8, rename column is introduced to rename column without altering the column definition, which could cause default value to be reset

## Things to worry about

RenameColumnGeneratorTest unit test is commented out due to past issue so UT is not implemented for this change

### DIFF
--- a/liquibase-standard/src/main/java/liquibase/sqlgenerator/core/RenameColumnGenerator.java
+++ b/liquibase-standard/src/main/java/liquibase/sqlgenerator/core/RenameColumnGenerator.java
@@ -37,7 +37,10 @@ public class RenameColumnGenerator extends AbstractSqlGenerator<RenameColumnStat
         validationErrors.checkRequiredField("newColumnName", renameColumnStatement.getNewColumnName());
 
         if (database instanceof MySQLDatabase) {
-            validationErrors.checkRequiredField("columnDataType", StringUtil.trimToNull(renameColumnStatement.getColumnDataType()));
+            MySQLDatabase mySQLDatabase = (MySQLDatabase) database;
+            if (!isRenameKeywordSupported(mySQLDatabase)) {
+                validationErrors.checkRequiredField("columnDataType", StringUtil.trimToNull(renameColumnStatement.getColumnDataType()));
+            }
         }
 
         return validationErrors;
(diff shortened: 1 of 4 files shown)


liquibase-standard/src/main/java/liquibase/sqlgenerator/core/RenameColumnGenerator.java
```diff
@@ -51,7 +50,15 @@ public Sql[] generateSql(RenameColumnStatement statement, Database database, Sql
         	// do no escape the new column name. Otherwise it produce "exec sp_rename '[dbo].[person].[usernae]', '[username]'"
             sql = "exec sp_rename '" + database.escapeTableName(statement.getCatalogName(), statement.getSchemaName(), statement.getTableName()) + "." + database.escapeColumnName(statement.getCatalogName(), statement.getSchemaName(), statement.getTableName(), statement.getOldColumnName()) + "', '" + statement.getNewColumnName() + "', 'COLUMN'";
         } else if (database instanceof MySQLDatabase) {
-            sql ="ALTER TABLE " + database.escapeTableName(statement.getCatalogName(), statement.getSchemaName(), statement.getTableName()) + " CHANGE " + database.escapeColumnName(statement.getCatalogName(), statement.getSchemaName(), statement.getTableName(), statement.getOldColumnName()) + " " + database.escapeColumnName(statement.getCatalogName(), statement.getSchemaName(), statement.getTableName(), statement.getNewColumnName()) + " " + DataTypeFactory.getInstance().fromDescription(statement.getColumnDataType(), database).toDatabaseDataType(database);
+            try {
```
filipelautert, line 53: As we need this Exception handling, maybe you could extract it to a `generateMysqlStatement` method ?

quangdutran:
Tested with MySQL 8 and Springboot integration:
image

Tested with MySQL 7
image

liquibase-standard/src/main/java/liquibase/sqlgenerator/core/RenameColumnGenerator.java
```diff
@@ -80,6 +79,21 @@ public Sql[] generateSql(RenameColumnStatement statement, Database database, Sql
         };
     }
 
+    private String generateMysqlStatement(MySQLDatabase database, RenameColumnStatement statement) {
+        String sql;
+        try {
+            // https://dev.mysql.com/worklog/task/?id=10761 RENAME COLUMN introduced in 8.0
+            if (database.getDatabaseMajorVersion() >= 8) {
```
filipelautert, line 86: @quangdutran I just remembered that `MariaDBDatabase` extends `MysqlDatabase`, so it will also get here and changes will be applied to `MariaDB >= 8` . We will need to change this validation to account for `MariaDB >= 10.5.2` to take advantage of this change too. I think you could remove the cast to `MysqlDatabase` when invoking the method and change the validation to something like
```
boolean isMariaDB = database instanceof MariaDBDatabase;
boolean isNewerVersion = isMariaDB
    ? (database.getDatabaseMajorVersion() == 10 && database.getDatabaseMinorVersion() >= 5)
        || database.getDatabaseMajorVersion() >= 11
    : database.getDatabaseMajorVersion() >= 8;

if (isNewerVersion) {

```

filipelautert (changes requested):

filipelautert (approved):

rberezen:
@quangdutran @filipelautert The test needs to be updated following the proposed change. Thank you!
```
Error:  Failures: 
Error:    VerifyChangeClassesTest.compareGeneratedSqlWithExpectedSqlForMinimalChangesets:141 Unexpected difference in /home/runner/work/liquibase/liquibase/liquibase-standard/src/test/resources/verify/saved_state/compareGeneratedSqlWithExpectedSqlForMinimalChangesets/renameColumn/mariadb.sql
Original:
[-- Database: mariadb
-- Change Parameter: columnDataType=int
-- Change Parameter: newColumnName=full_name
-- Change Parameter: oldColumnName=name
-- Change Parameter: tableName=person
ALTER TABLE person CHANGE name full_name INT;]
New state:
[-- Database: mariadb
-- Change Parameter: columnDataType=int
-- Change Parameter: newColumnName=full_name
-- Change Parameter: oldColumnName=name
-- Change Parameter: tableName=person
ALTER TABLE person RENAME COLUMN name TO full_name;]
```

tati-qalified:
@filipelautert @rberezen the renameColumn changeset validation still requires a datatype to be specified, even though the sql syntax doesn't call for it:
```
ALTER TABLE table_name 
RENAME COLUMN old_column_name TO new_column_name;
```

```
     5 changes have validation failures
          columnDataType is required for renameColumn on mysql, changelog.xml::1::tati
```

In my opinion this requirement should be removed. What do you think?

filipelautert:
> @filipelautert @rberezen the renameColumn changeset validation still requires a datatype to be specified, even though the sql syntax doesn't call for it:
> 
> ```
> ALTER TABLE table_name 
> RENAME COLUMN old_column_name TO new_column_name;
> ```
> 
> ```
> Unexpected error running Liquibase: Validation Failed:
>      5 changes have validation failures
>           columnDataType is required for renameColumn on mysql, changelog.xml::1::tati
> ```
> 
> In my opinion this requirement should be removed. What do you think?

@tati-qalified good catch!

@quangdutran there is validation around line 40 for `columnDataType` parameter -> https://github.com/liquibase/liquibase/pull/6920/files#diff-73a753088213f741b69e4acf8555e855f8d1b1c27f0633ccaaaf872ca1bdfb2eR40 . Could you fix it using the same version logic from `generateMysqlStatement` (maybe it can be extracted to another method and used there too) ?

quangdutran:
Hi @tati-qalified how should I test this?

tati-qalified:
@quangdutran Ideally having a changeset of renameColumn change type, without specifying data type for a MySQL version where `rename` is supported, and specifying data type where `rename` is not supported.
[Here](https://github.com/liquibase/liquibase/blob/278f62b1f3334537b822ce54718a82972d0e70f7/liquibase-integration-tests/src/test/java/liquibase/dbtest/pgsql/PostgreSQLIntegrationTest.java#L261) is an example of checking version numbers, as a reference.

filipelautert:
Hello @quangdutran , circling back here:  can we test this pr again?

quangdutran:
Hi @filipelautert,

Sorry this was carried away. I just checked the pipeline and `VerifyChangeClassesTest.compareGeneratedSqlWithExpectedSqlForMinimalChangesets` fails cause `mysql.sql` and `mariadb.sql` still has the line ```-- Change Parameter: columnDataType=int```

I remove this line and another test fails `ChangeDefinitionTest` with message: Change name renameColumn does not match

I try to remove the `columnDataType` under `renameColumn` in `ChangeDefinitionTest.yml` but I don't think it works.

Not sure why there are so many test failures now comparing to last time. How should I resolve these failures pls?

filipelautert:
> Hi @filipelautert,
> 
> Sorry this was carried away. I just checked the pipeline and `VerifyChangeClassesTest.compareGeneratedSqlWithExpectedSqlForMinimalChangesets` fails cause `mysql.sql` and `mariadb.sql` still has the line `-- Change Parameter: columnDataType=int`
> 
> I remove this line and another test fails `ChangeDefinitionTest` with message: Change name renameColumn does not match
> 
> I try to remove the `columnDataType` under `renameColumn` in `ChangeDefinitionTest.yml` but I don't think it works.
> 
> Not sure why there are so many test failures now comparing to last time. How should I resolve these failures pls?

Hey, thanks for the update. I just fixed the tests, they happened because of the required change for the datatype.

quangdutran:
@filipelautert Manual test is positive:

**MySQL 5.7:**
SQL generated with data type
Screenshot 2025-07-08 230830
Validation fails without data type
Screenshot 2025-07-08 231108

**MySQL 8:**
Screenshot 2025-07-08 231527


**MariaDB 10.4:**
Screenshot 2025-07-08 232027
Screenshot 2025-07-08 232114

**MariaDB 10.5.4:**
maria new

rberezen (approved):

rossdanderson:
I believe this has resulted in a regression as `COMMENT` is not a valid parameter for `RENAME COLUMN` and it generates the following

```
		<renameColumn tableName="dealside" oldColumnName="spotPrice " newColumnName="spotPrice2" columnDataType="VARCHAR(32)" remarks="somecomment"/>
```
```
ALTER TABLE nomad.dealside RENAME COLUMN spotPrice TO spotPrice2 COMMENT 'somecomment'
```

https://mariadb.com/docs/server/reference/sql-statements/data-definition/alter/alter-table#rename-column

PR #6929, opened 2025-05-05
- TIME ZONE keyword updated for Oracle.

MalloD12:
- Test updated.

## Impact

### DIFF
--- a/liquibase-standard/src/main/java/liquibase/datatype/core/TimestampType.java
+++ b/liquibase-standard/src/main/java/liquibase/datatype/core/TimestampType.java
@@ -126,19 +126,17 @@ public class TimestampType extends DateTimeType {
             type = new DatabaseDataType("TIMESTAMP");
         }
 
-        if (originalDefinition.startsWith("java.sql.Types.TIMESTAMP_WITH_TIMEZONE")
+        if (originalDefinition.toUpperCase().startsWith("JAVA.SQL.TYPES.TIMESTAMP_WITH_TIMEZONE")
             && (database instanceof PostgresDatabase
             || database instanceof OracleDatabase
             || database instanceof H2Database
             || database instanceof HsqlDatabase
             || database instanceof SybaseASADatabase)) {
 
-            if (database instanceof PostgresDatabase
-            || database instanceof H2Database
-            || database instanceof SybaseASADatabase) {
-                type.addAdditionalInformation("WITH TIME ZONE");
-            } else {
+            if (database instanceof HsqlDatabase) {
                 type.addAdditionalInformation("WITH TIMEZONE");
+            } else {
+                type.addAdditionalInformation("WITH TIME ZONE");
             }
 
             return type;
--- a/liquibase-standard/src/test/groovy/liquibase/datatype/DataTypeFactoryTest.groovy
+++ b/liquibase-standard/src/test/groovy/liquibase/datatype/DataTypeFactoryTest.groovy
@@ -227,8 +227,8 @@ class DataTypeFactoryTest extends Specification {
         "java.sql.Timestamp(6)"                        | new OracleDatabase()   | "TIMESTAMP(6)"                                 | TimestampType | false
         "java.sql.Types.TIMESTAMP"                     | new OracleDatabase()   | "TIMESTAMP"                                    | TimestampType | false
         "java.sql.Types.TIMESTAMP(6)"                  | new OracleDatabase()   | "TIMESTAMP(6)"                                 | TimestampType | false
-        "java.sql.Types.TIMESTAMP_WITH_TIMEZONE"       | new OracleDatabase()   | "TIMESTAMP WITH TIMEZONE"                      | TimestampType | false
-        "java.sql.Types.TIMESTAMP_WITH_TIMEZONE(6)"    | new OracleDatabase()   | "TIMESTAMP(6) WITH TIMEZONE"                   | TimestampType | false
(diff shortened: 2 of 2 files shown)


liquibase-standard/src/main/java/liquibase/datatype/core/TimestampType.java
```diff
@@ -126,19 +126,17 @@ public DatabaseDataType toDatabaseDataType(Database database) {
...
             || database instanceof H2Database
             || database instanceof HsqlDatabase
             || database instanceof SybaseASADatabase)) {
 
-            if (database instanceof PostgresDatabase
-            || database instanceof H2Database
-            || database instanceof SybaseASADatabase) {
-                type.addAdditionalInformation("WITH TIME ZONE");
-            } else {
+            if (database instanceof HsqlDatabase) {
```
filipelautert, line 136: PR title says Oracle, but here we are changing other databases behavior. Is it expected?

filipelautert (changes requested):

liquibase-standard/src/main/java/liquibase/datatype/core/TimestampType.java
MalloD12: I changed the order of the logic. Previously, `Postgres`, `H2`, and `SybaseASADatabase` were already processing as `WITH TIME ZONE`, so I thought I could revert the logic of this IF, instead of adding `Oracle` to the existing IF. 

I can revert this and just add `Oracle` to the previous `IF` logic.

rberezen:
@MalloD12 this will require some test changes in pro-tests and test-harness, I'm working on that.
https://github.com/liquibase/liquibase-pro-tests/actions/runs/15451576374
https://github.com/liquibase/liquibase-test-harness/actions/runs/15451612199

MalloD12:
> fix: ensure time zone information of timestamptz on creating TimestampType #6908

Hey @rberezen,

Have in mind changes from #6908, which are also related to these. We can talk about it if you have any questions.

Thanks,
Daniel.

tati-qalified (changes requested):
This is working as expected for most databases, but not for MySQL. I've tested:
- Oracle
- PostgreSQL
- H2
- MSSQL
- MySQL

With the following changeset:
```
    <changeSet id="create_table_with_timestamp" author="tati">
        <createTable tableName="testTimeZone">
            <column name="timestamptz" type="timestamptz" />
            <column name="java_lowercase" type="java.sql.types.timestamp_with_timezone" />
            <column name="java_uppercase" type="java.sql.Types.TIMESTAMP_WITH_TIMEZONE" />
            <column name="timestamp_w_t_z_lower" type="timestamp with time zone" />
            <column name="timestamp_w_t_z_upper" type="TIMESTAMP WITH TIME ZONE" />
        </createTable>
    </changeSet>
```

The following SQL is generated for Oracle (which works correctly):
```
CREATE TABLE
    DBUSER.testTimeZone (
        timestamptz TIMESTAMP WITH TIME ZONE,
        java_lowercase TIMESTAMP WITH TIME ZONE,
        java_uppercase TIMESTAMP WITH TIME ZONE,
        timestamp_w_t_z_lower TIMESTAMP with time zone,
        timestamp_w_t_z_upper TIMESTAMP WITH TIME ZONE
    );
```

And for MySQL:
```
CREATE TABLE
    lbcat.testTimeZone (
        timestamptz timestamp NULL,
        java_lowercase timestamp NULL,
        java_uppercase timestamp NULL,
        timestamp_w_t_z_lower timestamp with time zone NULL,
        timestamp_w_t_z_upper TIMESTAMP WITH TIME ZONE NULL
    );
```
but `timestamp with time zone` does not exist in MySQL. There's only `timestamp` and `datetime`.

As an example of another dbms that does not have the `timestamp with time zone` datatype, the generated SQL for MSSQL is as follows (and works correctly):
```
CREATE TABLE
    testTimeZone (
        timestamptz datetime2,
        java_lowercase datetime2,
        java_uppercase datetime2,
        timestamp_w_t_z_lower datetime2,
        timestamp_w_t_z_upper datetime2
    );
```

MalloD12:
@filipelautert could you please review these changes one more time? I made some additional changes to fix the MySQL case @tati-qalified has reported. If it's ok, then she can verify it, and then we can move forward with this PR that we need to merge before #6908.

Thanks,
Daniel.

liquibase-standard/src/test/groovy/liquibase/datatype/DataTypeFactoryTest.groovy
```diff
@@ -221,14 +234,14 @@ class DataTypeFactoryTest extends Specification {
         "TIMESTAMP(6) WITH TIMEZONE"                   | new OracleDatabase()   | "TIMESTAMP(6) WITH TIMEZONE"                   | TimestampType | false
```
filipelautert: Testing "TIMESTAMP WITH TIMEZONE" for Oracle, but shouldn't this be normalized to "TIMESTAMP WITH TIME ZONE" based on the code logic? Unless the code is supposed to preserve the exact input when it's already in proper Oracle syntax.

MalloD12: @filipelautert this has been fixed in case you want to have an additional look at it.

Thanks,
Daniel.

MalloD12:
@tati-qalified, when you have a chance, could you please test this one? If everything is ok, we can merge and check if we still need #6908.

tati-qalified:
@MalloD12 it's working as expected now. Tested generating sql script and also running update

PR #6936, opened 2025-05-07
fix: Change column type from tinyint to bool in mysql

### LINKED ISSUE #6763: MySQL and BOOLEAN column types - needs a bit more work.
### Search first

### Description

A previous issue was created for "Fixes MySQL BIT / TINYINT(1) Boolean Behaviours":  https://github.com/liquibase/liquibase/pull/4090#issuecomment-1764858220.  This references some MySQL documentation about how BOOL and BOOLEAN from other vendors is mapped to TINYINT.   However, MySQL engineers themselves, in Connector/J, further refine this (see below).

The resolution at that time was to change the MySQL BOOLEAN handling to be simply a "TINYINT".  We're running MySQL 5.7 and MySQL Connector/J 8.3.0 and see that a MySQL "TINYINT" column type results in a tinyint(4) column. MySQL Connector/J treats this as an integer rather than a boolean value.  It will treat bit(1) or (by default) tinyint(1) as boolean, and any other tinyint type as integer.

Here's a link to MySQL Connector/J on how it handles bit(1) and tinyint(1) for booleans, and any other tinyint type as integer: https://dev.mysql.com/doc/connector-j/en/connector-j-reference-type-conversions.html

IMHO Liquibase should treat a Liquibase BOOLEAN column type as either a bit(1) (recommended) or a tinyint(1) (would work - unless the tinyInt1isBit property is set to false.  bit(1) should ALWAYS be treated as boolean.)

I think the fix is in BooleanType, in the case for MySQLDatabase, where the original liquibase type does not start with 'bit', and where it isn't also a MariaDBDatabase type.  Change the TINYINT to be (recommended) BIT(1) as it was before the change from the referenced issue.

### Steps To Reproduce

Either:
Create a <createTable> refactoring where one of the columns is defined with type "BOOLEAN", or an <addColumn> refactoring where the column type is "BOOLEAN".
Execute the change set against a MySQL DB.

Or:
In a MySQL DB, create a table with a column of type "tinyint" (as Liquibase does).

Using a MySQL tool (e.g. MySQL Workbench), query the table and look at the column type. It is showing as "tinyint(4)".
Add a row to that table.
Using MySQL Connector/J, run a query against that table, and see that the result set, the underlying column type for the tinyint(4) column is 'Integer'.

Do the same, but create the column as type tinyint(1) or as bit(1).  Now MySQL Connector/J should show the column type as 'Boolean'.

Note that one can directly use tinyint(1) as a column type in a MySQL alter table / create table, and it will create the column with type tinyint(1).  As will using 'bool' as the MySQL column type.   At least with MySQL 5.7 - I don't have an 8.x or higher available to me.

### Expected/Desired Behavior

In MySQL, creating a column with Liquibase type BOOLEAN results in a column that MySQL Connector/J recognizes as a boolean rather than an integer type.

### Liquibase Version

4.27

### Database Vendor & Version

MySQL 5.7.43

### Liquibase Integration

Spring Boot / Java

### Liquibase Extensions

_No response_

N/A.

### Additional Context

I'm only available until the end of March, so while I'm willing to submit a PR for this, there isn't much time left.

quangdutran:
## Impact

## Description
Change column type from tinyint to bool in mysql
Fix #6763

### DIFF
--- a/liquibase-standard/src/main/java/liquibase/datatype/core/BooleanType.java
+++ b/liquibase-standard/src/main/java/liquibase/datatype/core/BooleanType.java
@@ -40,7 +40,7 @@ public class BooleanType extends LiquibaseDataType {
             if (originalDefinition.toLowerCase(Locale.US).startsWith("bit")) {
                 return new DatabaseDataType("BIT", getParameters());
             }
-            return database instanceof MariaDBDatabase ? new DatabaseDataType("TINYINT(1)") : new DatabaseDataType("TINYINT");
+            return database instanceof MariaDBDatabase ? new DatabaseDataType("TINYINT(1)") : new DatabaseDataType("BOOL");
         } else if (database instanceof OracleDatabase) {
             try {
                 if (database.getDatabaseMajorVersion() >= OracleDatabase.ORACLE_23C_MAJOR_VERSION) {
--- a/liquibase-standard/src/test/java/liquibase/sqlgenerator/core/AddColumnGeneratorTest.java
+++ b/liquibase-standard/src/test/java/liquibase/sqlgenerator/core/AddColumnGeneratorTest.java
@@ -170,11 +170,11 @@ public class AddColumnGeneratorTest extends AbstractSqlGeneratorTest<AddColumnSt
         Sql[] sql = instance.generateSql(statements, new MySQLDatabase());
 
         assertEquals(5, sql.length);
-        assertEquals("ALTER TABLE schema_name.table_name ADD column1 BIGINT NULL, ADD column2 TINYINT NULL", sql[0].toSql());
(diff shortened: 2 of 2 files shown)


quangdutran:
### Test

Test with MySQL 5, MySQL 8, MySQL Connector/J 8.3.0, 9.3.0

Changeset:
```
databaseChangeLog:
  - changeSet:
      id: create-table-users
      author: YourName
      changes:
        - createTable:
            tableName: users
            columns:
              - column:
                  name: id
                  type: INT
                  constraints:
                    primaryKey: true
                    nullable: false
              - column:
                  name: name
                  type: VARCHAR(255)
              - column:
                  name: dob
                  type: timestamp
              - column:
                  name: married
                  type: boolean
```

Through Connector/J, `married` column has `name=married, type=BIT (-7), precision=1`

PR #6968, opened 2025-05-21
♻️ DAT-20252: add GitHub App token retrieval for access token

sayaliM0412:
This pull request updates the `.github/workflows/release-published.yml` file to enhance security by using a GitHub App token for repository access instead of the default `GITHUB_TOKEN`. The most important changes are outlined below:

### Security Enhancements:

* **GitHub App Token Integration**: Added a new step, `Get GitHub App token`, which uses the `actions/create-github-app-token@v2` action to generate a GitHub App token. This step retrieves the token using the app's ID and private key, ensuring more secure access to the `liquibase-pro` repository.

* **Token Replacement**: Updated the `Download liquibase-pro xsd` step to use the newly generated GitHub App token (`steps.get-token.outputs.token`) instead of the default `GITHUB_TOKEN`. This change aligns with the enhanced security approach.

### DIFF
--- a/.github/workflows/release-published.yml
+++ b/.github/workflows/release-published.yml
@@ -310,11 +310,21 @@ jobs:
           path: liquibase-core-repo
           repository: "liquibase/liquibase"
 
+      - name: Get GitHub App token
+        id: get-token
+        uses: actions/create-github-app-token@v2
+        with:
+          app-id: ${{ secrets.LIQUIBASE_GITHUB_APP_ID }}
+          private-key: ${{ secrets.LIQUIBASE_GITHUB_APP_PRIVATE_KEY }}
+          owner: ${{ github.repository_owner }}
+          repositories: liquibase-pro
+          permission-contents: read
+
       - name: Download liquibase-pro xsd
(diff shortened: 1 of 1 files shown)


.github/workflows/release-published.yml
```diff
@@ -310,11 +310,21 @@ jobs:
...
 
+      - name: Get GitHub App token
+        id: get-token
+        uses: actions/create-github-app-token@v2
+        with:
+          app-id: ${{ secrets.LIQUIBASE_GITHUB_APP_ID }}
+          private-key: ${{ secrets.LIQUIBASE_GITHUB_APP_PRIVATE_KEY }}
+          owner: ${{ github.repository_owner }}
+          repositories: liquibase-pro
+          permission-contents: read
```
rberezen, line 321: read or write? just to confirm

jnewton03: @rberezen it depends on what the GitHub app token needs to do.  If it's committing files, creating PR, interacting with issues, etc it needs `write`.

jnewton03 (approved):

PR #6986, opened 2025-05-27
only destroy lock service if drop-all succeeds (DAT-13088)

StevenMassaro:
## Impact

## Description

In the event that the `drop-all` command execution fails, like in the event that a procedure cannot be dropped because it is overloaded, the DBCLL table is erroneously dropped. This PR changes the behavior to only destroy the lock service if `drop-all` is successful.


filipelautert (approved):

StevenMassaro:
> @StevenMassaro Approved, but would not that imply that if `dropall` fails for some other reason (like permissions) the lock will still be held and the user will need to unlock it manually?

No, the lock is always released. See line 115: `lockService.releaseLock();`. The destruction of the lock service also drops the DBCLL table.

PR #6998, opened 2025-05-30
Validate command update to restrict usage of non-boolean properties with an empty value while on STRICT mode

MalloD12:
The validate command can now be used to ensure **non-boolean** properties are not set with an empty value when STRICT mode is enabled.

## Impact


liquibase-standard/src/main/java/liquibase/command/core/ValidateCommandStep.java
```diff
@@ -36,8 +30,42 @@ public List<Class<?>> requiredDependencies() {
 
     @Override
     public void run(CommandResultsBuilder resultsBuilder) throws Exception {
-        // Do nothing. This is all handled in DatabaseChangelogCommandStep.
-        Scope.getCurrentScope().getUI().sendMessage(coreBundle.getString("no.validation.errors.found"));
```
filipelautert, line 40: We are missing this message now, could you re-add it?

liquibase-standard/src/test/groovy/liquibase/filters/DbmsChangeSetFilterTest.groovy
```diff
@@ -0,0 +1,39 @@
...
+        then:
+        changeSetFilterResult !=null
+        changeSetFilterResult.isAccepted() == false
+        changeSetFilterResult.getMessage() == "dbms value cannot be empty while on Strict mode"
+
+
+        where:
+        dbms  | _
+        ""    | _
+        null  | _
```
filipelautert: When it's null it should be accepted, right? As null means that the tag wasn't even provided. In fact null should be accepted for all...

MalloD12: I understood it shouldn't if we are using `strict` flag. Because specifying something like this `dbms=` should be considered a validation error. @mariochampion: Could you please confirm?

liquibase-standard/src/main/java/liquibase/command/core/ValidateCommandStep.java
MalloD12: True, I'll add it back if no validation errors were found.

MalloD12: Done

liquibase-standard/src/test/groovy/liquibase/filters/DbmsChangeSetFilterTest.groovy
filipelautert: But null also means that  `dbms=` property is not added. So if you check against null you'll be forcing the users to always add the dbms property.

MalloD12: Done

liquibase-standard/src/main/java/liquibase/executor/ExecutorService.java
```diff
@@ -134,7 +134,9 @@ private static class Key {
         }
 
         private String normalizeExecutorName(String executorName) {
-            return executorName.toLowerCase();
+            if (executorName == null) { return ""; } else {
```
filipelautert: I  believe with the other fixes , this change is not required anymore. Could you rollback it?

filipelautert (approved):
Just 1 comment

liquibase-standard/src/main/java/liquibase/executor/ExecutorService.java
MalloD12: done

liquibase-standard/src/main/java/liquibase/changelog/ChangeSet.java
```diff
@@ -376,7 +374,8 @@ public void setStoredFilePath(String storedFilePath) {
      * @return the runWith value. If the runWith value is empty or not set this method will return null.
      */
     public String getRunWith() {
-        return runWith == null || runWith.isEmpty() ? null : runWith;
+        //return runWith == null || runWith.isEmpty() ? null : runWith;
```
rberezen, line 377: @MalloD12 I believe this commented line should be removed

MalloD12:
@rberezen @filipelautert: guys, I'll close this PR since changes made here were also refactored for DEVX-30 (for which PR is  #7096). As mentioned in DEVX-1072, the testing for that ticket should be using the DEVX-30 branch. If you have any questions, please let me know.

PR #6999, opened 2025-05-30
🔧 DAT-20296 

sayaliM0412:
This pull request updates the GitHub Actions workflow in `.github/workflows/build.yml` to enhance authentication and repository checkout processes. The most significant changes involve replacing the default GitHub token with a GitHub App token for improved security and modifying the repository checkout logic to target the `liquibase-pro` repository.

### Enhancements to authentication:

* Added a new step, "Get GitHub App token," to generate a GitHub App token using the `actions/create-github-app-token@v2` action. This step uses secrets for the app ID and private key to securely obtain the token.

### Updates to repository checkout and token usage:

* Changed the repository checkout step to use the `liquibase-pro` repository instead of the `liquibase` repository. The step now uses the GitHub App token generated earlier for authentication.
* Updated the "Check Branch" step to use the GitHub App token instead of the default GitHub token for executing scripts via the `actions/github-script@v7` action.

### DIFF
--- a/.github/workflows/run-tests.yml
+++ b/.github/workflows/run-tests.yml
@@ -1,3 +1,4 @@
+
 name: Unit and Integration Tests
 
 permissions:


wwillard7800 (approved):

filipelautert (approved):

sayaliM0412:
NOTE: To check out the PR code during a pull_request_target, we must explicitly override it. The above PR change only gets the code — the workflow file will still come from `master` .
cc: @rberezen

PR #7016, opened 2025-06-03
Demo of using picocli to provide some color options for console logging

abrackx:
## Impact

## Description
This will not work in its current form but it proves a point!

I think a better idea may just be to modify the UIService, adding a new `printStyled` method and defaulting it to sending a "normal" message. Then we can continue to use ConsoleUIService.


abrackx:
The new `CommandLine` class is just a picocli dependency in liquibase-standard.

PR #7027, opened 2025-06-06
fix: set logical file path for changelog

### LINKED ISSUE #7019: LogicalFilePath with parameter is not replaced
### Search first

### Description

After updating to v4.31.0, an error appears at spring boot start and test that the checksum validation are not successful.

I discovered that the parameters in LogicalFilePath are no longer replaced. It still works in labels and other sections.

Example

```
--liquibase formatted sql logicalFilePath:db/changelogs/${schemaName}/1_changelog.sql

--changeset melle:1 labels:tenant-${schemaName}
CREATE SCHEMA IF NOT EXISTS "${schemaName}";

--changeset melle:2 labels:tenant-${schemaName}
CREATE TABLE IF NOT EXISTS "${schemaName}".testTable
(
    id
    BIGSERIAL
    PRIMARY
    KEY,
    test_name
    VARCHAR
(
    255
) NOT NULL
    )
```

schemaName = foo

I see in databasechanglog table this output

filename: db/changelogs/${schemaName}/1_changelog.sql --> before it was db/changelogs/foo/1_changelog.sql
labels: tenant-foo
author: melle

Schema foo is created with testTable

### Steps To Reproduce

Execute this changeset with parameter schemaName=foo

```

--liquibase formatted sql logicalFilePath:db/changelogs/${schemaName}/1_changelog.sql

--changeset melle:1 labels:tenant-${schemaName}
CREATE SCHEMA IF NOT EXISTS "${schemaName}";

--changeset melle:2 labels:tenant-${schemaName}
CREATE TABLE IF NOT EXISTS "${schemaName}".testTable
(
    id
    BIGSERIAL
    PRIMARY
    KEY,
    test_name
    VARCHAR
(
    255
) NOT NULL
    )
```

### Expected/Desired Behavior

The filename should be correct like before :)

### Liquibase Version

4.31.0

### Database Vendor & Version

_No response_

### Liquibase Integration

spring boot

### Liquibase Extensions

_No response_

_No response_

### Additional Context

_No response_

MelleD:
## Impact

## Description
This pull request introduces a change to the `parse` method in `AbstractFormattedChangeLogParser.java` to ensure the `logicalFilePath` is explicitly set on the `changeLog` object after being expanded like in the YamlChangeLogParser. 
This fix #7019  

Key change:

* [`liquibase-standard/src/main/java/liquibase/parser/AbstractFormattedChangeLogParser.java`](diffhunk://#diff-72a90d1968d64a41245b1069d04619bd52543261610281ecf7402ecd8cb63a79R445-R446): Added a call to `changeLog.setLogicalFilePath(logicalFilePath)` after expanding the `logicalFilePath` expressions to ensure it is properly set in the `changeLog` object.

Fix #7019

### DIFF
--- a/liquibase-standard/src/main/java/liquibase/parser/AbstractFormattedChangeLogParser.java
+++ b/liquibase-standard/src/main/java/liquibase/parser/AbstractFormattedChangeLogParser.java
@@ -451,8 +451,8 @@ public abstract class AbstractFormattedChangeLogParser implements ChangeLogParse
                     }
                     if (logicalFilePath != null) {
                         logicalFilePath = changeLogParameters.expandExpressions(logicalFilePath, changeLog);
+                        changeLog.setLogicalFilePath(logicalFilePath);
                     }
-                    changeLog.setLogicalFilePath( logicalFilePath );
                     
                     String dbms = handleDbms(changeLogParameters, line, changeLog);
 


liquibase-standard/src/main/java/liquibase/parser/AbstractFormattedChangeLogParser.java
```diff
@@ -442,6 +442,8 @@ public DatabaseChangeLog parse(String physicalChangeLogLocation, ChangeLogParame
                     if (logicalFilePath != null) {
                         logicalFilePath = changeLogParameters.expandExpressions(logicalFilePath, changeLog);
                     }
+                    changeLog.setLogicalFilePath( logicalFilePath );
```
tati-qalified: Hi @MelleD, thank you for your contribution!
This line of code is outside of the `if (logicalFilePath != null)` clause, so I believe the problem isn't coming from here. It's most likely that the `logicalFilePath` attribute is reaching line #422 as null instead of finding the value from the manually-set property.

Could you look into this?
Thank you!
Tatiana

tati-qalified (changes requested):

liquibase-standard/src/main/java/liquibase/parser/AbstractFormattedChangeLogParser.java
MelleD: Hey @tati-qalified,

thanks for your feedback.
I'm not sure what https://github.com/liquibase/liquibase/pull/422 has to do with the issue? Can you please explain?

I noticed that from version 4.29 to > 4.30.0, setting the logicalFilePath no longer works. I couldn't figure out the exact change because there were quite a few changes during that time.

For this reason, I looked into why it works in YAML but not in SQL.
When I debug, I see that the logicalFilePath is always expanded correctly. However, it is never set in the changelog; it always remains null. It is set in the changeset itself, but not in the changelog.

In YAML, it is also set in the changelog. That's why I added the line.
I could also move the line or make a null check.

tati-qalified: @MelleD what I meant is that line 422 checks whether the `logicalFilePath` is null or not, but your addition is outside of that `if` clause. So you may be setting a null `logicalFilePath`, which is unnecessary as it's not a required parameter.

MelleD: @tati-qalified 

Okay, I was confused because you directly linked to a PR.

I thought it wouldn't make a difference if the path was `null` and `null` was set again in the setter.
I'll move the line.

tati-qalified (changes requested):
Having tested this manually, here are my findings:
| | Changelog level  | Changeset level |
| ------------ | ------------- | ------------- |
| xml | ❌ | ✅ |
| sql | ❌  | ✅ |
| json | ✅ | ✅ |
| yml | ✅ | ✅ |

For xml, adding `logicalFilePath` to the `databaseChangeLog` tag results in the property not being substituted:
```
<databaseChangeLog xmlns="http://www.liquibase.org/xml/ns/dbchangelog" (...)
    logicalFilePath="inTag/${myProp}">

    <changeSet id="123" author="tati">
        <createTable tableName="withProp">
            <column name="col1" type="int"></column>
        </createTable>
    </changeSet>
```
generates the following sql script:
```
-- Changeset inTag/${myProp}::123::tati
CREATE TABLE PUBLIC.withProp (col1 INT);
```

-------

For sql the issue is different and a bit confusing - I set up a changelog with a general logicalFilePath, one changeset without its own logicalFilePath and one with.
```
--liquibase formatted sql logicalFilePath:db/changelogs/${myProp}/1_changelog.sql

--changeset melle3:2
CREATE TABLE IF NOT EXISTS testTab
(
    id BIGSERIAL PRIMARY KEY,
    test_name VARCHAR(255) NOT NULL
)

--changeset mel3:2 logicalFilePath:changeset/sql/${myProp}
CREATE TABLE IF NOT EXISTS testTable
(
    id BIGSERIAL PRIMARY KEY,
    test_name VARCHAR(255) NOT NULL
)
```
The first changeset should get the changelog-level logicalFilePath, and the second changeset should get its own logicalFilePath. 
However, both changesets are getting the path from the second one:
```
-- Changeset changeset/sql/abc::2::melle3
CREATE TABLE IF NOT EXISTS testTab
(
    id BIGSERIAL PRIMARY KEY,
    test_name VARCHAR(255) NOT NULL
);

-- Changeset changeset/sql/abc::2::mel3
CREATE TABLE IF NOT EXISTS testTable
(
    id BIGSERIAL PRIMARY KEY,
    test_name VARCHAR(255) NOT NULL
);
```



Could you look into these cases?
Thanks!

MalloD12:
Hi Dennis (@MelleD), 

Could you please look after this PR when you have a chance? As @tati-qalified mentioned it seems there are some changes required to move forward with it. It seems we probably need to update the below method:
```java
 protected void setLogicalFilePath(DatabaseChangeLog changeLog, String line, Matcher changeLogPatternMatcher) {
        if (changeLog.getLogicalFilePath() == null && changeLogPatternMatcher.matches()) {
            Matcher logicalFilePathMatcher = LOGICAL_FILE_PATH_PATTERN.matcher(line);
            changeLog.setLogicalFilePath(parseString(logicalFilePathMatcher, LOGICAL_FILE_PATH));
        }
    }
```
in order to fix this issue for both changeset and changelog levels. It would be nice if we can add some unit tests for the changes made. Please let me know if you need any asistence for this.

Thanks,
Daniel.

MalloD12:
Hi Dennis (@MelleD),

FYI, we have created this [PR](https://github.com/liquibase/liquibase/pull/7451) to fix the `logicalFilePath` issue. Please have a look at it and please let us know if you have any feedback.

Thanks
Daniel.

MelleD:
Hey Daniel (@MalloD12),

thanks for the effort.

Unfortunately, the whole process took too long for me, with constantly new requirements. I needed a quick fix for my case first, after which all cases could have been fixed. Since we needed a quick solution, we have now expanded the logical path and rebuilt all liquibase scripts, which was faster and better in our case.

MalloD12:
Dennis (@MelleD), I understand and let me apologize from the team for taking long to resolve this issue.  If at some point, you want to test these changes that would be awesome too, or just try when a new version is released. 

Thanks,
Daniel.

PR #7032, opened 2025-06-06
feat: last-wins local parameters search in changeLog hierarchy

boolivar:
## Impact

## Description

This PR introduces back feature of searching local parameters in changeLog hierarchy (lost in #2656), making it possible to inherit local properties in included changeLogs.
Property search starts from system then global properties and continues for local properties from the last property in current changeLog up to the root.

## Things to be aware of

Although [documentation](https://docs.liquibase.com/concepts/changelogs/property-substitution.html) states
> Once a property has been set, it cannot be changed. Also, only the first definition is used, others are skipped.

current liquibase behavior is opposite and this PR is not going to change it in order to prevent breaking changes.

## Additional Context

Original PR with (breaking) "first wins" behavior #6397

Motivation: https://github.com/liquibase/liquibase/pull/4267#issuecomment-2380604946


MalloD12 (approved):
Thanks, @boolivar! This PR looks good to me. 

Quick question, are we planning to get rid of (close) #6397?

PR #7056, opened 2025-06-19
fix: update snapshot logic to handle null snapshot returns

filipelautert:
## Impact

## Description

- Initialize objectToSnapshot as null , so if nothing is found we return null
- Updated `SnapshotGeneratorChain` to pass `example` only if the previous generator could not find it.
- Don't overwrite the existing snapshot if the current generator returns null.


filipelautert:
@fbiville Another change to the snapshot logic. Could you confirm if this won't break Neo4j extension?

wwillard7800 (approved):
Checks out against snapshot with Databricks.

fbiville:
@filipelautert will try it out this week and report back

fbiville:
@filipelautert I did a quick run and this seems to work fine

PR #7069, opened 2025-06-25
Handle situation where DBCL presence was not detected due to permissions DAT-20314

wwillard7800:
## Impact

### DIFF
--- a/liquibase-standard/src/main/java/liquibase/changelog/StandardChangeLogHistoryService.java
+++ b/liquibase-standard/src/main/java/liquibase/changelog/StandardChangeLogHistoryService.java
@@ -273,11 +273,18 @@ public class StandardChangeLogHistoryService extends AbstractChangeLogHistorySer
 
         for (SqlStatement sql : statementsToExecute) {
             if (SqlGeneratorFactory.getInstance().supports(sql, database)) {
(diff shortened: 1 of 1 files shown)


liquibase-standard/src/main/java/liquibase/changelog/StandardChangeLogHistoryService.java
```diff
@@ -273,11 +273,22 @@ public void init() throws DatabaseException {
...
+                try {
+                    ChangelogJdbcMdcListener.execute(getDatabase(), ex -> ex.execute(sql));
+                    getDatabase().commit();
+                } catch (Exception e) {
+                    //
+                    // Trap for an exception here so we can add a message with more information.
+                    // If the user does not have read permission, we can end up here because the
+                    // presence of the DBCL was not detected earlier.
+                    //
+                    String message = "An error occurred while attempting to create the database history table.  Please make sure that you " +
```
filipelautert: Maybe here we could use the current dbcl table name?

wwillard7800: Yes, I should use that.  It is used in the other log messages.

StevenMassaro (approved):

PR #7079, opened 2025-07-02
feat: add logic to exclude resources from liquibaseHomeUri/internal in resource processing for includeAll

filipelautert:
This pull request introduces a mechanism to exclude resources located in liquibase `internal `directory when searching for changelog resources to be included. It also adds a utility method to retrieve the URI of this internal directory. Below is a summary of the most important changes:

### Enhancements to Resource Filtering:
* Updated the `findResources` method in `DatabaseChangeLog` to exclude resources located in the `internal` subdirectory of the Liquibase home directory. This is determined by checking if the resource URI starts with `jar:<LiquibaseHome>/internal`.

How to reproduce it: 

Create a changelog.yml like this one: 

```
databaseChangeLog:
- includeAll:
    path: .
    endsWithFilter: .sql
```

Notice that we are not using `relativeToChangelogFile: true` neither `searchPath` property. So you need to be inside the directory with your sql files to be included when trying to reproduce this issue.

### DIFF
--- a/liquibase-standard/src/main/java/liquibase/changelog/DatabaseChangeLog.java
+++ b/liquibase-standard/src/main/java/liquibase/changelog/DatabaseChangeLog.java
@@ -34,13 +34,17 @@ import liquibase.resource.ResourceAccessor;
 import liquibase.servicelocator.LiquibaseService;
 import liquibase.util.ExceptionUtil;
 import liquibase.util.FileUtil;
+import liquibase.util.LiquibaseLauncherSettings;
 import liquibase.util.StringUtil;
 import lombok.Getter;
 import org.apache.commons.io.FilenameUtils;
 import org.apache.commons.lang3.StringUtils;
 
+import java.io.File;
 import java.io.IOException;
 import java.io.InputStream;
+import java.net.URI;
+import java.nio.file.InvalidPathException;
 import java.nio.file.Path;
 import java.nio.file.Paths;
 import java.util.*;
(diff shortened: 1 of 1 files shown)


liquibase-standard/src/main/java/liquibase/changelog/DatabaseChangeLog.java
```diff
@@ -899,7 +903,12 @@ public SortedSet<Resource> findResources(
             List<Resource> unsortedResources = getUnsortedResources(pathName, errorIfMissingOrEmpty, resourceAccessor, relativeTo, seenChangelogPaths, searchOptions);
             SortedSet<Resource> resources = new TreeSet<>((o1, o2) -> resourceComparator.compare(o1.getPath(), o2.getPath()));
             if (unsortedResources != null) {
+                URI liquibaseHomeInternalUri = getLiquibaseHomeInternalUri();
                 for (Resource resourcePath : unsortedResources) {
+                    // if the resource is inside a jar located in liquibaseHomeUri/internal, we don't want to include it
+                    if (liquibaseHomeInternalUri != null && resourcePath.getUri().toString().startsWith("jar:" + liquibaseHomeInternalUri.toString())) {
+                        continue;
```
StevenMassaro, line 911: Can you add logging here? Otherwise this could be a very frustrating edge case for someone to experience.

filipelautert: Sure, done!

wwillard7800 (approved):

PR #7096, opened 2025-07-09
Validate command changes to detect empty and invalid properties values

MalloD12:
## Impact

## Description

This PR addresses DEVX-30 and DEVX-1072.


liquibase-standard/src/main/java/liquibase/changelog/RawChangeSet.java
```diff
@@ -0,0 +1,84 @@
+package liquibase.changelog;
```
filipelautert: Could you move all those new classes inside a `validator` package ?

liquibase-standard/src/main/java/liquibase/changelog/filter/propertyvalidator/DbmsValidatorFilter.java
```diff
@@ -0,0 +1,64 @@
...
+        StringBuilder errors = new StringBuilder();
+
+        if(strict) {
+            if (changeSet.getDbms() == null) {
+                return new ChangeSetFilterResult(true, "", ChangeSetFilter.class ,"validDBMS", "dbms");
+            }
+
+            String[] dbmsList = changeSet.getDbms() != null ? changeSet.getDbms().split(",") : null;
+
+            for (String dbms : dbmsList) {
```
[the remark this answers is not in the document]
filipelautert: Please handle this warning

liquibase-standard/src/main/java/liquibase/command/core/ValidateCommandStep.java
```diff
@@ -31,13 +34,48 @@ public void adjustCommandDefinition(CommandDefinition commandDefinition) {
 
     @Override
     public List<Class<?>> requiredDependencies() {
-        return Arrays.asList(Database.class, DatabaseChangeLog.class);
+        return Arrays.asList(Database.class);
```
filipelautert, line 42: By removing `DatabaseChangeLog` dependency we are missing XML validation. Would make sense to manually  add it only for XML files?

filipelautert (changes requested):
For all the new classes: could you add some javadoc?

liquibase-standard/src/main/java/liquibase/command/core/ValidateCommandStep.java
MalloD12: I'm not sure to be getting this one. There is a content extractor for XML as well, so RawChangeSet will be created as well for changesets found, and then property values will be validated. Compared with other changelog formats, the only property not being validated will be preconditions, because for this format we rely on XSD, we won't need to validate that an invalid preconditions tag was entered.

liquibase-standard/src/main/java/liquibase/changelog/RawChangeSet.java
MalloD12: I moved everything to a `validator` package, the only one I'm not moving there is `ValidateChangeLogIterator`, which remains in `changelog` package. I think we can leave that one there, since other iterators are there as well. What do you think?

filipelautert (approved):

liquibase-standard/src/main/java/liquibase/changelog/ValidateChangeLogIterator.java
```diff
@@ -0,0 +1,91 @@
...
+        } else if (changelogPath.toLowerCase().endsWith(".yaml") || changelogPath.toLowerCase().endsWith(".yml")) {
+            return "yaml";
+        } else if (changelogPath.toLowerCase().endsWith(".sql")) {
+            return "sql";
+        } else {
+            return "unknown";
+        }
+    }
+
+    private List<RawChangeSet> extractRawChangeSets(String content, String format) {
```
tati-qalified, line 81: Is this command not available for mongodb changelogs?

MalloD12: I don't think so, at least not at this stage. @mariochampion could you please confirm when you have a chance?

Thanks,
Daniel.

tati-qalified (changes requested):
Manually tested. I've found a few bugs.

**Bugs:**
1. Included files aren't validated. Ticket [DEVX-1138](https://datical.atlassian.net/browse/DEVX-1138)
2. Related to num. 1, the paths for the included files aren't checked, so invalid filepaths are passing validation. Ticket [DEVX-1139](https://datical.atlassian.net/browse/DEVX-1139)
3. SQL changelogs aren't being validated at all. Ticket [DEVX-1140](https://datical.atlassian.net/browse/DEVX-1140)
4. Yaml changelogs are causing the error `Caused by: java.lang.StringIndexOutOfBoundsException: begin 0, end -1, length 0`. This does not happen when running `liquibase update-sql`. Ticket [DEVX-1141](https://datical.atlassian.net/browse/DEVX-1141)


**Enhancements:**
Ticket: [DEVX-1142](https://datical.atlassian.net/browse/DEVX-1142)
1. Add information about the changeset that includes the error
   - If the same error appears in multiple changesets, group the info by error type or by changeset. 
2. Show the output only once (currently printed twice to the console, even with log-level set to severe)
3. `{..} value cannot be empty while on Strict mode` should be `strict` in lowercase.

PR #7138, opened 2025-07-21
Remove pro code from Liquibase Maven plugin

filipelautert:
## Impact

## Description

Build artifacts in https://github.com/liquibase/liquibase/actions/runs/16528045714

This commit removes all references to pro code from the Liquibase Maven plugin. This introduces changes that may impact projects relying on previously included pro features. Users relying on this functionality will need to adjust accordingly.

## Things to be aware of

- Complete removal of pro code may impact users who were dependent on pro-specific features in the Maven plugin.

## Things to worry about

- Ensure that documentation reflects removal of pro code and the associated impact on existing users.
- Confirm compatibility and proper functionality for Maven plugin users post-removal.

## Additional Context

Removing pro code aligns with overarching goals to streamline the Maven plugin and its dependencies.

### DIFF
--- a/liquibase-maven-plugin/src/main/java/org/liquibase/maven/plugins/AbstractLiquibaseMojo.java
+++ b/liquibase-maven-plugin/src/main/java/org/liquibase/maven/plugins/AbstractLiquibaseMojo.java
@@ -368,16 +368,6 @@ public abstract class AbstractLiquibaseMojo extends AbstractMojo {
     protected boolean showBanner = true;
 
     /**
-     *
-     * Enable or disable reports
-     *
-     * @parameter property="liquibase.reportsEnabled"
-     *
-     */
-    @PropertyElement
-    protected boolean reportsEnabled = true;
-
-    /**
      * Specifies the server ID in the Maven <i>settings.xml</i> to use when authenticating.
      *
      * @parameter property="liquibase.server"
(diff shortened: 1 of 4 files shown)


liquibase-maven-plugin/src/main/java/org/liquibase/maven/plugins/LiquibaseDropAll.java
```diff
@@ -43,9 +36,9 @@ protected void performLiquibaseTask(Liquibase liquibase) throws LiquibaseExcepti
             for (String name : schemas.split(",")) {
                 schemaObjs.add(new CatalogAndSchema(catalog, name));
             }
-            liquibase.dropAll(dropDbclhistory, schemaObjs.toArray(new CatalogAndSchema[0]));
```
MalloD12, line 46: Quick question, I know `dropDbclhistory` is PRO mvn parameter, but considering we are updating this logic instead of removing it, shouldn't we create a new one with the updates you made and deprecate the one we want to get rid of?

filipelautert: The pro-logic still exists in the new commercial-maven-plugin: https://github.com/liquibase/liquibase-pro/tree/master/liquibase-commercial-maven-plugin  . So the ones using pro license need to migrate to it (that in fact is a copy of this one, just changing `org.liquibase` by `com.liquibase`) .

MalloD12 (approved):

PR #7140, opened 2025-07-22
NullPointerException in SpringLiquibase.toString()

tweimer:
Fixed NPE in toString()

## Impact

## Description

I fixed a NullPointerException in {{toString()}} Method. I observed it while debugging and it propably doesn't happen in any production code, but anyways....

There are no other changes, as {{toString()}} is implicitely invoked for string concat. The code is shorter and the NPE is removed.


filipelautert (approved):

filipelautert:
Thanks @tweimer ! Makes sense , we don't need to call `toString` here (and in a lot of other places too....)

PR #7192, opened 2025-08-12
Fixes for regressions in DAT-20619, DAT-20620, and DAT-20621

wwillard7800:
## Impact

### DIFF
--- a/liquibase-standard/src/main/java/liquibase/serializer/core/formattedsql/FormattedSqlChangeLogSerializer.java
+++ b/liquibase-standard/src/main/java/liquibase/serializer/core/formattedsql/FormattedSqlChangeLogSerializer.java
@@ -48,7 +48,7 @@ public class FormattedSqlChangeLogSerializer  implements ChangeLogSerializer {
             ChangeSet changeSet = (ChangeSet) object;
             Database database = Scope.getCurrentScope().get(DiffToChangeLog.DIFF_SNAPSHOT_DATABASE, Database.class);
             Database targetDatabase = getTargetDatabase(changeSet);
-            if (database == null || ! database.getShortName().equals(targetDatabase.getShortName())) {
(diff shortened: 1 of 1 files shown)


liquibase-standard/src/main/java/liquibase/serializer/core/formattedsql/FormattedSqlChangeLogSerializer.java
```diff
@@ -45,8 +45,9 @@ public String serialize(LiquibaseSerializable object, boolean pretty) {
             //
             ChangeSet changeSet = (ChangeSet) object;
             Database database = Scope.getCurrentScope().get(DiffToChangeLog.DIFF_SNAPSHOT_DATABASE, Database.class);
-            if (database == null) {
-                database = getTargetDatabase(changeSet);
+            Database targetDatabase = getTargetDatabase(changeSet);
+            if (database == null || ! database.getShortName().equals(targetDatabase.getShortName())) {
```
filipelautert: What's the idea behind this new check here?

wwillard7800: The use case here is where the database from the URL is of one type (mssql), but the requested output database from the file name (output.postgresql.sql) is another.

wwillard7800:
> Tests are failing ...

Yes, I found the issue.

liquibase-standard/src/main/java/liquibase/serializer/core/formattedsql/FormattedSqlChangeLogSerializer.java
wwillard7800: I ended up refactoring the code in another place, and I was able to remove the short name check.

PR #7269, opened 2025-09-12
INT-1392: add get/setArguments and get/setProcedureName methods into …

HorbatenkoYehor:


liquibase-standard/src/main/java/liquibase/structure/core/StoredProcedure.java
```diff
@@ -10,4 +13,34 @@ public StoredProcedure(String catalogName, String schemaName, String procedureNa
...
+        setAttribute("procedureName", procedureName);
+        return this;
+    }
+
+    /**
+     * For databases (like Snowflake) that support procedure overloading {@link #getName()} will contain id (name with input arguments)
+     * so actual name should be fetched from {@link #getProcedureName()}
+     */
+    public static String getProcedureName(DatabaseObject dbObject) {
+        if (dbObject instanceof StoredProcedure) {
```
wwillard7800: A very minor note, but since we have moved to Java 17+, you can use the `instanceof <type> <variable>` syntax now.

So you can say `if (dbobject instanceof StoredProcedure dbObject)` and avoid the cast.

HorbatenkoYehor: Thanks for the advice! I updated

PR #6627, opened 2025-01-09
fix: Allow skipping a missing custom change

Mahoney:
## Impact

- [x] Bug fix (non-breaking change which fixes expected existing functionality)
- [ ] Enhancement/New feature (adds functionality without impacting existing logic)
- [ ] Breaking change (fix or feature that would cause existing functionality to change)
 

## Description

Prior to this a custom change with a class name that is not on the classpath would always fail the entire changelog.

Now `changeSet.failOnError: false` will be honoured if the cause of the failure is the failure to instantiate the change class.

In addition, a failing precondition for the changeset with onFail="MARK_RAN" will be honoured if the class is not on the classpath.

Fixes #6520

## Things to be aware of

N/A

## Things to worry about

N/A

## Additional Context

N/A

PR #7390, opened 2025-10-21
Allow skipping a missing custom change (Copy of 6627)

MalloD12:
## Impact

### DIFF
--- a/liquibase-standard/src/main/java/liquibase/change/custom/CustomChangeWrapper.java
+++ b/liquibase-standard/src/main/java/liquibase/change/custom/CustomChangeWrapper.java
@@ -64,7 +64,7 @@ public class CustomChangeWrapper extends AbstractChange {
                 this.customChange = loadCustomChange(className);
             } catch (CustomChangeException e) {
                 final Logger log = Scope.getCurrentScope().getLog(getClass());
-                log.warning("Exception thrown loading " + getClassName(), e);
+                log.warning(String.format("Custom change class %s cannot be loaded. Please verify the class exists and is on the classpath.", getClassName()), e);
                 return null;
             }
         }


MalloD12:
To be closed when #6627 is merged.

liquibase-standard/src/main/java/liquibase/change/custom/CustomChangeWrapper.java
```diff
@@ -54,9 +55,19 @@ public boolean generateStatementsVolatile(Database database) {
...
+     * Returns null if the custom change class cannot be loaded.
      */
     @DatabaseChangeProperty(isChangeProperty = false)
     public CustomChange getCustomChange() {
+        if (this.customChange == null) {
+            try {
+                this.customChange = loadCustomChange(className);
+            } catch (CustomChangeException e) {
+                final Logger log = Scope.getCurrentScope().getLog(getClass());
+                log.warning("Exception thrown loading " + getClassName(), e);
```
wwillard7800: The log message here shouldn't talk about the exception, but should instead state that we are skipping a missing custom change.

filipelautert: Right now this only logs the exception. It would be good to state that we are skipping the missing change too.

filipelautert (changes requested):

liquibase-standard/src/main/java/liquibase/change/custom/CustomChangeWrapper.java
MalloD12: done

MalloD12: cc: @filipelautert

Issue #7320, opened 2025-09-30
Leaving mandatory file property of include empty gives strange error


b-gyula:
### Search first

- [x] I searched and no similar issues were found

### Description

When leaving the `file` property of the `include` element empty a strange error reported:
```
ERROR: Exception Primary Reason:  Found 11 files with the path '':
    - file:///D:/dev/liquibase/groovy-dsl/
    - file:///D:/prg/sdkman/candidates/liquibase/current/lib/
    - jar:file:/D:/prg/sdkman/candidates/liquibase/current/internal/lib/jaxb-api.jar!/META-INF/versions/9/
    - jar:file:/D:/prg/sdkman/candidates/liquibase/current/internal/lib/mariadb-java-client.jar!/META-INF/versions/11/
    - jar:file:/D:/prg/sdkman/candidates/liquibase/current/internal/lib/picocli.jar!/META-INF/versions/9/
    - jar:file:/D:/prg/sdkman/candidates/liquibase/current/internal/lib/slf4j-api.jar!/META-INF/versions/9/
    - jar:file:/D:/prg/sdkman/candidates/liquibase/current/internal/lib/slf4j-nop.jar!/META-INF/versions/9/
    - jar:file:/D:/prg/sdkman/candidates/liquibase/current/internal/lib/snakeyaml.jar!/META-INF/versions/9/
    - jar:file:/D:/prg/sdkman/candidates/liquibase/current/internal/lib/sqlite-jdbc.jar!/META-INF/versions/9/
    - file:///D:/prg/sdkman/candidates/liquibase/current/internal/lib/
    - file:///D:/prg/sdkman/candidates/liquibase/current/internal/extensions/
  Search Path: 
    - D:\dev\liquibase\groovy-dsl
    - D:\prg\sdkman\candidates\liquibase\current\internal\lib\liquibase-core.jar
    - D:\prg\sdkman\candidates\liquibase\current\lib
    - D:\prg\sdkman\candidates\liquibase\current\internal\lib\commons-collections4.jar
    - D:\prg\sdkman\candidates\liquibase\current\internal\lib\commons-io.jar
    - D:\prg\sdkman\candidates\liquibase\current\internal\lib\commons-lang3.jar
    - D:\prg\sdkman\candidates\liquibase\current\internal\lib\commons-text.jar
    - D:\prg\sdkman\candidates\liquibase\current\internal\lib\h2.jar
[...]

PR #7395, opened 2025-10-25
Liquibase throws exception when the filePath property is empty #7320

Parthiee:
## Summary by CodeRabbit

* **Bug Fixes**
  * Improved changelog file path validation during serialization to ensure paths are properly defined and non-empty, preventing invalid or incomplete changelog configurations.

* **Tests**
  * Added comprehensive test coverage to verify file path validation is correctly enforced across supported YAML and JSON changelog serialization formats.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -1,11 +1,16 @@
 
 name: Build and Publish
 
+# Security: Minimal permissions for build workflow
+# - contents: read - checkout code (write not needed for builds)
+# - packages: write - publish snapshots to GitHub Packages
+# - actions: read - read workflow artifacts
+# - id-token: write - AWS OIDC authentication
 permissions:
-  contents: write
+  contents: read
   packages: write
   actions: read
-  id-token: write # Required for OIDC authentication with AWS
+  id-token: write
 
 # The concurrency check will cancel in-progress runs on non-master branches and pull requests on forked repositories regardless of the target branch.
 concurrency:
--- a/.github/workflows/claude-code-review.yml
+++ b/.github/workflows/claude-code-review.yml
@@ -4,10 +4,14 @@ on:
   pull_request:
     types: [opened, synchronize]
 
+# Security: Minimal permissions for code review
+# - contents: read - read code for review
+# - pull-requests: write - post review comments
+# - id-token: write - authentication
+# Note: issues: write removed as code review doesn't create issues
 permissions:
   contents: read
   pull-requests: write
-  issues: write
   id-token: write
   actions: read
   checks: read
@@ -16,4 +20,12 @@ permissions:
 jobs:
   claude-review:
     uses: liquibase/build-logic/.github/workflows/claude-code-review.yml@main
+    permissions:
+      contents: read
+      pull-requests: write
+      issues: read
+      id-token: write
+      actions: read
+      checks: read
+      statuses: read
     secrets: inherit
--- a/.github/workflows/claude.yml
+++ b/.github/workflows/claude.yml
@@ -10,6 +10,11 @@ on:
   pull_request_review:
     types: [submitted]
 
+# Security: Permissions for Claude AI assistant
+# - contents: read - read code
+# - pull-requests: write - respond to PR comments
+# - issues: write - respond to issue mentions (required for @claude in issues)
+# - id-token: write - authentication
 permissions:
   contents: read
   pull-requests: write
--- a/.github/workflows/create-release.yml
+++ b/.github/workflows/create-release.yml
@@ -1,9 +1,12 @@
 name: Attach Artifacts To Draft Release
 
+# Security: These write permissions are required for release creation
+# - contents: write - create release tags and upload assets
+# - actions: write - trigger downstream workflows
+# - id-token: write - AWS OIDC authentication
 permissions:
   contents: write
   actions: write
-  packages: write
   id-token: write
 
 on:
@@ -105,7 +108,7 @@ jobs:
           parse-json-secrets: true
 
       - name: Download liquibase-artifacts
-        uses: dawidd6/action-download-artifact@v11
+        uses: dawidd6/action-download-artifact@v12
         with:
           workflow: run-tests.yml
           run_id: ${{ needs.setup.outputs.runId }}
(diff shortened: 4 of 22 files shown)


Parthiee:
Hi @filipelautert / @wwillard7800 , I have made few changes that fixes the issue addressed in #7320. Kindly look into this and let me know this needs any change

filipelautert:
Thanks @Parthiee ! We will take a look soon, but at first sight it makes sense.

Parthiee:
I have addressed the copilot review in the latest commit

Parthiee:
Hi @filipelautert , Any updates on the PR?

filipelautert:
> Hi @filipelautert , Any updates on the PR?

IT's on our review queue, will get there soon 👍🏻 
CC @MalloD12

MalloD12:
Hey @Parthiee would you mind having a look at the coderabbit review comments? I think the two comment are valid points I would fix them before moving on with this PR.

Thanks,
Daniel.

Parthiee:
Hi @MalloD12 ! Addressed the changes in the latest commit.

MalloD12:
Hi @Parthiee,

Thank you for submitting this PR and addressing some of the review comments the IA tools have provided. It seems it's well oriented, but I'm still seeing some inconsistencies:

- If I'm not wrong, you mentioned this PR aims to fix #7320, but looking at that issue, that one is related to `file` attribute of an `include` object. I think maybe instead of adding the below method you added to `ChangeLogSerializer` interface, we can replace that one with one that also validates include/includeAll attributes:

For example, something like:

```java
 default void validatePathAttribute(ChangeLogChild child) {
      String fieldName = null;
      Object pathObject = null;

      if (child instanceof ChangeSet) {
          fieldName = "filePath";
          pathObject = ((ChangeSet) child).getSerializableFieldValue("filePath");
      } else if (child instanceof ChangeLogInclude) {
          fieldName = "file";
          pathObject = ((ChangeLogInclude) child).getSerializableFieldValue("file");
      } else if (child instanceof ChangeLogIncludeAll) {
          fieldName = "path";
          pathObject = ((ChangeLogIncludeAll) child).getSerializableFieldValue("path");
      } else {
          return; 
      }

      String pathValue = StringUtils.trimToNull(
          Optional.ofNullable(pathObject).map(Object::toString).orElse(null)
      );

      if (StringUtils.isEmpty(pathValue)) {
          throw new IllegalArgumentException(
              String.format("The '%s' attribute cannot be empty or null", fieldName)
          );
      }
  }
```

- Something else I noticed is that this solution is considering YAML/JSON, but not considering XML. And for SQL format, currently we are only validating `filePath` is not `null`, so it would be good to extend this implementation to all the formats. 

I think that's it I have at the moment. Please let me know if there is anything else I can help with.

Thanks,
Daniel.

Parthiee:
Hi @MalloD12 ! Your comments make sense to me. I will look into this ASAP. Additionally, I also noted the inconsistent naming of the fields like "file", "filePath" and "path". I think  it would make more sense if we have a consistent convention. What is your opinion on this?

MalloD12:
@Parthiee, I personally think unifying a name would make sense for me for `filePath` and `file`, considering both make a reference to a single file, but `path` references a directory (maybe this one could have a different name). With that said, we cannot change these names, because it would do we would be introducing a big breaking change that would affect a lot of people. 

Thanks,
Daniel.

Parthiee:
Hi @MalloD12 !
I was looking at your suggestions and I realized there were no usages of `ChangeLogInclude` and `ChangeLogIncludeAll` in the code. Can you help me understand why we should consider multifurcation for different types when they have no usage in the code?

 Also, about this:

>  And for SQL format, currently we are only validating filePath is not null, so it would be good to extend this implementation to all the formats.

Do you mean we should be extending this validation for other attributes?

MalloD12:
Hi @Parthiee,

I've been reviewing my previous suggestion and yes, you are right  `ChangeLogInclude` / `ChangeLogIncludeAll` should be use, but currently they are not used (there is an open PR) where somebody was refactoring DatabaseChangeLog class and moving the related logic to these two classes, but I don't know when that PR will be merged because of some other reasons. So, I think the proposed solution shouldn't take them into account. Something else I noticed from that method and also from your proposed solution is that here we are considering `filePath` validation from a changeset, but that attribute is not specified for a changeSet when defining one in a changelog. In my opinion, a fix like this would make more sense for `include` (as reported in #7320) / `includeAll`, and in that case, I would propose something like this:

```java
 private void validatePathAttribute(ParsedNode node, String nodeName) {
      String fieldName;
      String pathValue;

      if (INCLUDE_CHANGELOG.equals(nodeName)) {
          fieldName = FILE;
          pathValue = node.getChildValue(null, FILE, String.class);
      } else if (INCLUDE_ALL_CHANGELOGS.equals(nodeName)) {
          fieldName = PATH;
          pathValue = node.getChildValue(null, PATH, String.class);
      } else {
          // Not a node type that requires path validation, nothing to do
          return;
      }

      pathValue = StringUtils.trimToNull(pathValue);

      if (StringUtils.isEmpty(pathValue)) {
          throw new IllegalArgumentException(
              String.format("The '%s' attribute cannot be empty or null in <%s>", fieldName, nodeName)
          );
      }
  }

```

And then call this function before each of the following calls in `DatabaseChangeLog` class (L484) :
```java
 case INCLUDE_CHANGELOG: {
                handleInclude(node, resourceAccessor, nodeScratch);
                break;
            }
            case INCLUDE_ALL_CHANGELOGS: {
                handleIncludeAll(node, resourceAccessor, nodeScratch);
                break;
            }
```

Regarding the part I was referring to others format, please ignore that too, because with the proposed changes, since the path validation would be in the `DatabaseChangeLog` class, there wouldn't be a need to update any of the serializers, so the change would be applicable for any changelog format.

Please let me know if you have any additional questions/concerns. 

Thanks,
Daniel.

Parthiee:
Hi @MalloD12 !

I have refactored the code to handle filePath in `databaseChangeLog` class. I made sure it was generic enough to validate any attribute type. Please review it.

Thanks
Parthiban

Parthiee:
Hi @MalloD12 ! Any updates on this PR?

MalloD12:
Hi @Parthiee,

I'm starting to review your last changes now.

MalloD12:
@tati-qalified would you mind testing these changes when you have a chance?

Thanks,
Daniel.

PR #7410, opened 2025-11-03
include(All), property file/path attribute starting with '.' sets relativeToChangelogFile` true by default

b-gyula:
## Impact

## Description

Starting `include@file`, `includeAll@path`, `property@file`, `load(Update)Data@file`, `createView/Procedure@path` and `sqlFile@path` with `.` sets `relativeToChangelogFile` true by default

Improved error handling for these attributes:
- tag name and changeset id added to error messages related to file opening like in any other change property validation error. E.g
`'/lb/dir/' is a directory set as createProcedure:path in changeset inc/createProc.yml::1::a`
instead of `/lb/dir is a directory`

- 'set as include:file' or 'set as property:file' added to the error message when the file set for `include@file` or `property@file` not found instead of just `The file x was not found in the configured search path:...`. e.g.  File '/root/lb/a' not found set as property:file in 'prop.yml'

- The full path added in the error message for missing file declared relative to help identify the file not found: `File '/root/lb/a.prop' not found set as property:file in 'test.yml'` instead of `The file a.prop was not found in the configured search path: ...` since there is no search done in this case and helps the user to see if the `relativeToChangelogFile` was not set correctly

`property` attributes related error improved:
- missing `file`, `value` or `name` for 'property' resulted in `NullPointerException:  Cannot invoke "String.equalsIgnoreCase(String)" because the return value of "liquibase.changelog.ChangeLogParameters$ChangeLogParameter.getKey()" is null`. or `Cannot invoke "java.util.Map.get(Object)" because "property" is null`

  new error message: `Either 'file' or 'name' + 'value' must be set for 'property'`

- Warning if 'file' is set for 'property', then 'value' and 'name' are ignored
```xml
<databaseChangeLog>
  <property file='a' name='n'/>
```

- Empty `name` attribute for `property` created a property with no name silently, which of course could not be used. Now results error message: `'value' is required for 'property' when 'name' is set`
```xml
<databaseChangeLog>
  <property name=''/>
```

- empty `file` attribute for `include` and `property` results niceer error message `'file' is empty for 'include'` instead of cryptic error listing all folders and files on the search path
```xml
<databaseChangeLog>
  <property file=''/>
```

- Added character ' around both the field and the tag names in validation error messages like `'file' is empty for 'include'`

- database type removed from required field validation error messages, if the property is required for ALL databases, since it has no added value. e.g. `file is required for loadData on h2` -> 'file' is required for 'loadData'

Code related changes:
- Duplicated property handling removed from `YamlChangeLogParser`

- Generic file handling related properties: `file`, `relativeToChangelogFile`, `encoding` extracted into interface `HasFileProperty`

- the default encoding used for `createView/Procedure` from `GlobalConfiguration.FILE_ENCODING`

- `Resource` got new method `isFile`

- `ResourceAccessor` got new method `getExistingFile` checking if the path points to a file beyond the existence check

- The more than 50 % exact same code in `createView/Procedure` extracted into common parent class `AbstractSQLAndFileChange`

## Things to be aware of
Empty `path` for `includeAll` gives error message `'path' is empty for 'includeAll'` instead of processing the complete search path. `path: '.'` reads the actual folder.

## Things to worry about
- `sqlFile` validates file existence in the validation stage in the `validate` method, whereas `load(Update)Data@file`, `createView/Procedure` does not. *This shouldn't be consistent?*
With the new `ValidationErrors.fileExisting` it should be easy to make.
- Why is file name field named `path` for change `createView/Procedure` and `sqlFile` instead of `file` as for `load(Update)Data` and `include` used for the same thing? It is very confusing to the user and results lot of unnecessary code duplication.


jnewton03:
@coderabbitai review

PR #7426, opened 2025-11-11
Fixed issue #7415: MSSQL, SQL Anywhere: Does not quote tablespace name

### LINKED ISSUE #7415: MSSQL, SQL Anywhere: Does not quote tablespace name
### Search first

### Description

Liquibase stops updating a MSSQL database with the following error message:

```
Unexpected error running Liquibase: Migration failed for changeset modules/pb-dev/pb-dev.xml::create-index-pbcatc_x::karg@quipsy.de:
     Reason: liquibase.exception.DatabaseException: Falsche Syntax in der Nähe von "Index". Falls dies Teil eines Tabellenhinweises sein soll, sind ein WITH-Schlüsselwort und Klammern erforderlich. Informationen zur richtigen Syntax finden Sie der SQL Server-Onlinedokumentation. [Failed SQL: (1018) CREATE UNIQUE NONCLUSTERED INDEX [pbcatc_x] ON [QADMIN].[pbcatcol]([pbc_tnam], [pbc_ownr], [pbc_cnam]) ON Index;]
```

Apparently the *name* `Index` of the tablespace is **not** getting quoted!

This is weird, as `Index` is contained in the reserved words collection of `MSSQLDatabase`! 🤔 

The same happens with SQL Anywhere (but certainly with different error message).

### Steps To Reproduce

```xml
<createIndex indexName="pbcatc_x" schemaName="QADMIN" tableName="pbcatcol" unique="true" tablespace="Index">
    <column name="pbc_tnam"/>
    <column name="pbc_ownr"/>
    <column name="pbc_cnam"/>
</createIndex>
```

### Expected/Desired Behavior

```sql
CREATE UNIQUE NONCLUSTERED INDEX [pbcatc_x] ON [QADMIN].[pbcatcol]([pbc_tnam], [pbc_ownr], [pbc_cnam]) ON [Index];
```
(The last word `Index` must be written as `[Index]`)

### Liquibase Version

5.0.1

### Database Vendor & Version

MSSQL 2022 Express

### Liquibase Integration

CLI

### Liquibase Extensions

none

Windows 11

### Additional Context

_No response_

mkarg:
## Impact

## Description

Fixes #7415 

## Things to be aware of

As the bug is only reported against MSSQL and SQL Anywhere, this PR only fixes MSSQL and SQL Anywhere.

For sake of this fix solely, a pure *marker* class named `Tablespace` is added, as it is needed for the pre-existing `Database::encodeObjectName` method. No instances are actually created, and aren't intended to get created. A new method `Database::encodeTablespaceName` is added in analogy to the pre-existing `Database::encodeTableName` method, and intentionally *only* used in those very places where it was *essential* to call it to fix this bug.

## Things to worry about

If anybody wants *other* RDBMS to get fixed, he/she needs to first check if that RDBMS *does* support keywords as tablespace names, and under what conditions; then, possibly needs to adapt the DBMS-specific `Database::encodeTablespaceName` method, followed by calling it in the very places indicated by this PR.

## Additional Context

Change has proven correctness and stability in production already using a custom build ontop Liquibase 5.0.1.

### DIFF
--- a/liquibase-standard/src/test/java/liquibase/database/core/MSSQLDatabaseTest.java
+++ b/liquibase-standard/src/test/java/liquibase/database/core/MSSQLDatabaseTest.java
@@ -161,4 +161,100 @@ public class MSSQLDatabaseTest extends AbstractJdbcDatabaseTest {
         }
         return super.getExpectedSqlVisitors(database);
     }
+
+    @Test
+    public void testEscapeTablespaceName_withoutSpaces() throws DatabaseException {
+        // Given: A tablespace name without spaces
+        try (Database database = new MSSQLDatabase()) {
+            String tablespaceName = "MyTablespace";
+
+            // When: Escaping the tablespace name
+            String result = database.escapeTablespaceName(tablespaceName);
+
+            // Then: The tablespace name should be properly escaped without additional quotes
+            assertEquals("MyTablespace", result);
+        } catch (final DatabaseException e) {
+            throw e;
+        }
+    }
+
+    @Test
+    public void testEscapeTablespaceName_withSpaces() throws DatabaseException {
+        // Given: A tablespace name with spaces
+        try (Database database = new MSSQLDatabase()) {
+            String tablespaceName = "My Tablespace";
+
+            // When: Escaping the tablespace name
+            String result = database.escapeTablespaceName(tablespaceName);
+
+            // Then: The tablespace name should be bracket-escaped (spaces require brackets) and wrapped in double quotes
+            assertEquals("\"[My Tablespace]\"", result);
+        } catch (final DatabaseException e) {
+            throw e;
+        }
+    }
+
+    @Test
+    public void testEscapeTablespaceName_null() throws DatabaseException {
+        // Given: A null tablespace name
+        try (Database database = new MSSQLDatabase()) {
+            String tablespaceName = null;
+
+            // When: Escaping the tablespace name
+            String result = database.escapeTablespaceName(tablespaceName);
+
+            // Then: The result should be null
+            assertNull(result);
+        } catch (final DatabaseException e) {
+            throw e;
+        }
+    }
+
+    @Test
+    public void testEscapeTablespaceName_withParenthesis() throws DatabaseException {
+        // Given: A tablespace name with parenthesis
+        try (Database database = new MSSQLDatabase()) {
+            String tablespaceName = "Tablespace(1)";
+
+            // When: Escaping the tablespace name
+            String result = database.escapeTablespaceName(tablespaceName);
+
+            // Then: The tablespace name should not be escaped (treated as function)
+            assertEquals("Tablespace(1)", result);
+        } catch (final DatabaseException e) {
+            throw e;
+        }
+    }
+
+    @Test
+    public void testEscapeTablespaceName_withSpecialCharacters() throws DatabaseException {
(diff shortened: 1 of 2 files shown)


mkarg:
@filipelautert Kindly asking you to consider a review / merge of my bug fix. Thank you! 🙂

liquibase-standard/src/main/java/liquibase/database/core/MSSQLDatabase.java
```diff
@@ -315,6 +315,20 @@ public String escapeTableName(String catalogName, String schemaName, String tabl
...
 
+    @Override
+    public String escapeTablespaceName(String tablespaceName) {
+        //
+        // If the tablespace name has a parenthesis in it, the escape logic might mistake it for stored logic
+        // and not add quotes. We can check for a space and the lack of a quote at the beginning and quote
+        // it anyway.
+        //
+        tablespaceName = escapeObjectName(tablespaceName, Tablespace.class);
+        if (tablespaceName != null && tablespaceName.contains(" ") && ! tablespaceName.startsWith("\"")) {
```
MalloD12, line 326: Hey @mkarg,

Should not be this line also checking  by "[", something like this: 

```java
 if (tablespaceName != null && tablespaceName.contains(" ") &&
      !tablespaceName.startsWith("\"") && !tablespaceName.startsWith("[")) {
```

or maybe a bit more generic:
```java
if (tablespaceName != null && tablespaceName.contains(" ") &&
      !tablespaceName.startsWith(getQuotingStartCharacter())) {
      tablespaceName = quoteObject(tablespaceName, Tablespace.class);
  }
```
? What do you think?

mkarg: The current PR copies the existing method `escapeTableName` and works very well in production. If you change it the way you propose, `escapeTableSpaceName` would *deviate* from `escapeTableName`, which is not a good things, as it would confuse users. Moreover, reducing to `getQuotingStartCharacter` even makes the code *fail*, as you would *double-quote* in the now-missing `"` case. So from the view of a user my answer is: Please keep the PR *as-is*.

From a developer's view I understand your reasoning, but would propose to separate concerns. First merge this PR *as-is* (as it is correct and perfectly copies exactly the existing behavior of `escapeTableName`, and it solves the target issue #7415), but then put *another* PR ontop which adds the `[` case (to solve a not-yet-filed issue). In **no** case the code should reduce to `getQuotingStartCharacter` as that will fail (see above).

MalloD12: Hey Markus,

Ok, sounds good let's move forward with the approach on splitting this up into two different PRs. To move ahead with this one, could you please add some unit tests for MSSQL for the changes made? If you are not able to do it I can help with it.

Thanks,
Daniel.

mkarg: Can you please provide the needed tests? I am booked out this week. Thanks.

mkarg:
> **Approved**.
> 
> Code looks good to me. As discussed in comment there will be a second part of these changes coming up later. Also, as discussed here, I have added a few tests.
> 
> Thanks, Daniel.

@MalloD12 Here is the second part as you suggested: https://github.com/liquibase/liquibase/pull/7448.

PR #7436, opened 2025-11-16
Add support for H2 lowercase options (from v1.4.199) 

b-gyula:
## Impact

## Description
1. Added support to new `DECFLOAT` type for option `MODE=PostgreSQL`. Without it `decimal` columns created without scale and precision defined, `load(Update)Data` is not able to load values into these columns, because it can not recognize these columns as NUMERIC type.

2. Added support for options: `DATABASE_TO_UPPER=false` and `DATABASE_TO_LOWER=true`. When any of these options used running an `update` the second time error : `Unexpected error running Liquibase: Table "DATABASECHANGELOG" already exists;` happens, because the incorrect character casing won't find the table name in the snapshot

## Things to be aware of
1. Instead of multiplication the same code the 6th time, a new abstract parent `NumrericType` created and used for all numeric types.
2.


MalloD12 (approved):
**Approved**.

Thank you for this PR, @b-gyula! Overall it looks good to me. I made some pretty minor changes, removing commented out code, null-check added, etc.

filipelautert (approved):

tati-qalified:
Hi @b-gyula, thank you for your contribution!
About point 2: does `preserveSchemaCase` not solve the issue for you? Here is what the [documentation](https://contribute.liquibase.com/extensions-integrations/directory/database-tutorials/h2/) says:
> If you need to ensure casing is preserved when using H2 with Liquibase, set `liquibase.preserveSchemaCase` to `true` in your `liquibase.properties` file.

If it doesn't work, could you provide a step-by-step of what fails?
Thank you!

rberezen (commented):
## Code Review Notes

**1. Commented-out code in `SqlUtilTest.groovy`**

```groovy
//  @Unroll("SqlUtil.parseValue('#value' on #db as '#dataType)")
    def "ParseValue '#value' on #db as '#dataType"() {
```
Looks like a debugging leftover — the old `@Unroll` annotation is commented out instead of removed.

**2. Unused imports** — `LoadDataChange` is still imported but no longer used in these files (since `getLoadTypeName()` moved to `NumericType`):
- `DecimalType.java`
- `SmallIntType.java`
- `IntType.java`
- `TinyIntType.java`

rberezen:
@b-gyula , could you please resolve the merge conflicts and address the code review findings?

filipelautert (commented):
Hey @b-gyula — circling back on this one. The approach looks good and it carries two approvals, but it can't move forward in its current state. To unblock it:

1. **Merge conflicts** — the branch is now conflicting with `main`; needs a rebase/merge to go green.
2. **Open review findings from @rberezen** (Apr 6) — the commented-out `@Unroll` in `SqlUtilTest.groovy`, and the now-unused `LoadDataChange` imports in `DecimalType`, `SmallIntType`, `IntType`, and `TinyIntType` (dead since `getLoadTypeName()` moved up to `NumericType`).
3. **Open QA question from @tati-qalified** (Jan 2) — does setting `liquibase.preserveSchemaCase=true` already cover your case? A quick step-by-step of what fails with that flag set would confirm this PR is the right fix vs. a config solution.

Once the conflicts are resolved and those are addressed, happy to take another pass.

PR #7439, opened 2025-11-17
Fix performance degradation since implementing analytics

### LINKED ISSUE #7213: Liquibase performance degradation 4.3x.x
### Search first

### Description

There was similar issue reported with liquibase 4.30.0 and claimed to be fixed with 4.32.0, but unfortunately it does not fix my usecase.
Similar issues:
* https://github.com/liquibase/liquibase/issues/6685
* https://github.com/liquibase/liquibase/pull/6686

My liquibase uscase is following:
* we customized liquibase to be multitenant - each tenant creates new `SpringLiquibase` and invokes `afterPropertiesSet()` method which triggers migrations
* each tenant is on separate `DataSource`
* everything is multithreaded - we run up to 16 threads at the same time
* we had to customize liquibase to support multithreading and multitenancy properly - at this point i do not reject that we might use liquibase framework in 'not expected way', however the implementation works since liquibase 1.9

On top of that, we have many types of migrations:
* from library, per tenant
* from library, for application
* from application, per tenant
* from application, for application

There is one library that defines empty changelog files, like this:
```
<databaseChangeLog
        xsi:schemaLocation="http://www.liquibase.org/xml/ns/dbchangelog https://www.liquibase.org/xml/ns/dbchangelog/dbchangelog-latest.xsd">
</databaseChangeLog> 
```

After upgrading liquibase to version higher than 4.29.2 we see significant slowness in running 'empty changeset' - it increased from 200ms-500ms to around 5s-9s per single tenant! In total, migration execution time increased from 1 minute to 30minutes, which slows down all deployments.

I tested all versions:
* 4.29.2 - works fine
* 4.30.0 - does not work
* 4.31.1 - does not work
* 4.32.0 - does not work (disabled analytics)
* 4.33.0 - does not work (disabled analytics)

And that makes me thinking there must be another issue i am facing here. 

### Steps To Reproduce

Unfortunately i am not sure we can reproduce it without having our custom code, which obviously i cannot share :(

### Expected/Desired Behavior

The expected behaviour is that empty migration will be super fast - usually should be less than 300ms.

### Liquibase Version

4.33.0

### Database Vendor & Version

PGSQL 15

### Liquibase Integration

maven, spring-boot 

### Liquibase Extensions

_No response_

_No response_

### Additional Context

_No response_

tati-qalified:
## Impact

## Description
Fixes #7213 
Modified execute() command to create an `Event` object only if analytics are enabled. This will avoid creating unnecessary Event objects that automatically run resource-heavy methods and impact performance.

### DIFF
--- a/liquibase-standard/src/main/java/liquibase/sqlgenerator/core/AddPrimaryKeyGenerator.java
+++ b/liquibase-standard/src/main/java/liquibase/sqlgenerator/core/AddPrimaryKeyGenerator.java
@@ -47,12 +47,11 @@ public class AddPrimaryKeyGenerator extends AbstractSqlGenerator<AddPrimaryKeySt
     @Override
     public Sql[] generateSql(AddPrimaryKeyStatement statement, Database database, SqlGeneratorChain sqlGeneratorChain) {
         String sql;
-        if ((statement.getConstraintName() == null) || (database instanceof MySQLDatabase) || (database instanceof
-            SybaseASADatabase)) {
+        if ((statement.getConstraintName() == null) || (database instanceof MySQLDatabase)) {
             sql = "ALTER TABLE " + database.escapeTableName(statement.getCatalogName(), statement.getSchemaName(), statement.getTableName()) + " ADD PRIMARY KEY (" + database.escapeColumnNameList(statement.getColumnNames()) + ")";
         } else {
             sql = "ALTER TABLE " + database.escapeTableName(statement.getCatalogName(), statement.getSchemaName(), statement.getTableName()) + " ADD CONSTRAINT " + database.escapeConstraintName(statement.getConstraintName())+" PRIMARY KEY";
-            if ((database instanceof MSSQLDatabase) && (statement.isClustered() != null)) {
+            if ((database instanceof MSSQLDatabase || database instanceof SybaseASADatabase) && (statement.isClustered() != null)) {
                 if (statement.isClustered()) {
                     sql += " CLUSTERED";
                 } else {


liquibase-standard/src/main/java/liquibase/analytics/Event.java
```diff
@@ -205,4 +199,46 @@ private static String getExtensionVersion(String extensionName) {
...
+
+            cachedIsDocker = Boolean.TRUE.equals(ExceptionUtil.doSilently(() -> {
+                if (BooleanUtils.toBoolean(System.getenv("DOCKER_LIQUIBASE")) ||
+                        BooleanUtils.toBoolean(System.getenv("DOCKER_AWS_LIQUIBASE"))) {
+                    return true;
+                }
+
+                // Only do file I/O on Linux/Unix systems
+                String osName = System.getProperty("os.name", "").toLowerCase();
+                if (!osName.contains("linux") && !osName.contains("unix")) {
```
filipelautert, line 222: Besides the synchronized, the is is the only change to previous logic right?

liquibase-standard/src/main/java/liquibase/command/CommandScope.java
```diff
@@ -419,6 +436,28 @@ private String getCommandStepName(CommandStep commandStep) {
         return commandStepName.toString();
     }
 
+    /**
+     * Creates an analytics event if analytics is enabled, or returns null if disabled.
+     * This method handles all potential failures gracefully to ensure analytics issues
+     * never break command execution.
+     */
+    private Event createAnalyticsEventIfEnabled(String commandName) {
```
filipelautert, line 444: Nice, you switched from `exception based` code to a simple check .

filipelautert (approved):
That's nice!

filipelautert:
@rberezen as this fix was developed by @tati-qalified , could you get someone else to test it?

rberezen:
> @rberezen as this fix was developed by @tati-qalified , could you get someone else to test it?

I'll take a look, thanks

liquibase-standard/src/main/java/liquibase/analytics/Event.java
tati-qalified: Yes, all other changes are about the order of execution

sta-szek:
hi, when can we expect release with this fix and what will be the version? i hope you can do it in line 4.x.x

tati-qalified:
Hi @sta-szek, this will be released with Liquibase Community 5.0.2 and Liquibase Secure 5.1.0. As of now there are not plans of releasing new 4.x versions.

PR #7452, opened 2025-12-01
fix: detect comment changes on columns and tables

peteraisher:
## Impact

## Description

Fixes #https://github.com/liquibase/liquibase-hibernate/issues/824

Comments on a column or table, e.g. using Hibernate's `@Comment` annotation, were successfully detected by the database implementation, but no diff was generated when a comment was changed. This change adds a check for changed comments on a column or table, allowing the existing functionality for creating diff commands to work as expected.

### Steps to reproduce:
1. Create a JPA entity definition with Hibernate's @Comment annotation on the table or a column
2. run the diffChangelog command
3. observe that the comment annotation has been picked up in the generated changes
4. change or remove the comment, or add a comment to another column or table
5. run the diffChangelog command again

### Expected behaviour:
The generated changes include commands to change or remove the comment, or to add a comment to an existing column or table.

### Actual behaviour:
The generated changes don't include commands to update, add or remove comments for existing tables or columns.

Users who want to document the semantics of their columns or tables within java code currently cannot use e.g. Hibernate's features to do so.

With this change, comments specified in code will be persisted in the database schema, making life easier for data analysts and other stakeholders without source code access who want to benefit from column and table documentation.

I've included tests which cover the changes.

## Things to be aware of

- I added similar functionality in two classes to support this change for tables and columns.
- I also cleaned up some unused imports in test classes.

## Things to worry about

- It might make sense to make this an opt-in feature (or opt-out) controlled by a setting.

## Additional Context

Comments are successfully detected on _new_ columns and tables. In other contexts, a more descriptive name might be preferable to a comment, but e.g. in Postgres, identifiers are limited to max. 63 bytes.

### DIFF
--- a/liquibase-standard/src/test/groovy/liquibase/diff/compare/core/ColumnComparatorTest.groovy
+++ b/liquibase-standard/src/test/groovy/liquibase/diff/compare/core/ColumnComparatorTest.groovy
@@ -53,7 +53,7 @@ class ColumnComparatorTest extends Specification {
         then:
         def differences = comparator.findDifferences(column1, column2, new PostgresDatabase(), CompareControl.STANDARD, new DatabaseObjectComparatorChain(Collections.emptyList(), null), new HashSet<String>())
         !differences.differences.isEmpty()
-        differences.differences.containsKey("remarks")
+        differences.isDifferent("remarks")
 
         where:
         before | after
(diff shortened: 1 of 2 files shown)


MalloD12 (approved):
**Approved**.

Thank you, @peteraisher! This PR looks good to me, I have made a little change to one of the tests added to make sure `remarks` is part of the `differences` found.  

PR is showing some test failures, but those are not related to this changes and we are already working on it to get that fix.

tati-qalified (approved):

PR #7464, opened 2025-12-11
fix(mysql/mariadb): respect columnDataType in renameColumn for modern versions

### LINKED ISSUE #7339: MariaDB renameColumn regression bug: columnDataType is ignored
### Search first

### Description

I found that using MariaDB 10.11 and liquibase-maven-plugin version 5.0.0, the renameColumn change type stops working as expected.

I drilled down to this bugfix: #6876 which introduces the `RENAME COLUMN` syntax for MySQL/MariaDb.
However, this bugfix introduced regression when changing the column's datatype together with its name.

### Steps To Reproduce

Initial ChangeSet:
```
<createTable tableName="PERSON">
      ...
      <column name="NAME" type="VARCHAR(32)">
      </column>
      ...
 </createTable>
```

Followup ChangeSet:
```
<renameColumn tableName="PERSON" oldColumnName="NAME"
      newColumnName="NEW_NAME" columnDataType="VARCHAR(64)"/>
```

This effectively ignores the columnDataType set on the renameColumn and in my opinion is an unwanted change in existing behavior.

The generated SQL changes from:

`ALTER TABLE PERSON CHANGE NAME NEW_NAME VARCHAR(64);`
to
`ALTER TABLE PERSON RENAME COLUMN NAME TO NEW_NAME;`

### Expected/Desired Behavior

One could argue that the renameColumn change type is not intended to change the columnDataType in one go, but it causes regression as the columnDataType field is just ignored instead of applied. The desired behavior is in my opinion, that if the columnDataType field is supplied to the changeSet, it is taken into account in the alter table statement and not just forgotten.

### Liquibase Version

5.0.0

### Database Vendor & Version

MariaDB 10.11

### Liquibase Integration

Maven plugin

### Liquibase Extensions

mariadb-java-client

_No response_

### Additional Context

_No response_

filipelautert:
## Summary
- Uses `CHANGE` syntax when `columnDataType` is provided on MariaDB 10.5+ and MySQL 8.0+
- Uses `RENAME COLUMN` syntax when only renaming (no type change) is needed
- Preserves backward compatibility with older database versions

## Problem
In Liquibase 5.0.0, the `renameColumn` change type for MariaDB/MySQL ignores the `columnDataType` parameter when the database supports `RENAME COLUMN` syntax. This causes the type change to be silently discarded.

**Example:**
```xml
<renameColumn tableName="PERSON" oldColumnName="NAME"
  newColumnName="NEW_NAME" columnDataType="VARCHAR(64)"/>
```

**Expected:** `ALTER TABLE PERSON CHANGE NAME NEW_NAME VARCHAR(64);`
**Actual:** `ALTER TABLE PERSON RENAME COLUMN NAME TO NEW_NAME;` (type change ignored!)

## Solution
Modified `RenameColumnGenerator.generateMysqlStatement()` to check if `columnDataType` is provided:
- If `columnDataType` is null/empty AND `RENAME COLUMN` is supported → use `RENAME COLUMN` syntax
- If `columnDataType` is provided → use `CHANGE` syntax (regardless of version)

## Test plan
- [x] Added 37 Spock tests covering all MySQL/MariaDB version combinations
- [x] Verified existing RenameColumn tests still pass (41 total tests)
- [ ] Manual testing with MariaDB 10.11 to verify type change works

Fixes #7339

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -95,13 +95,13 @@ jobs:
           ref: ${{ github.event.pull_request.head.sha || github.event.after}}
       # this includes all the tar files included in the previous runs. So in the next step we deploy what was previously build
       - name: Download Artifacts for build.yml
-        uses: actions/download-artifact@v6
+        uses: actions/download-artifact@v7
         with:
           name: temp-artifact
 
       ##Cache based on install4j file, since changes to JVM is seen in there. If install4j version changes without changing the file, change the prefix letter before hashFiles to force a new cache
       - name: Install4j Cache
-        uses: actions/cache@v4.3.0
+        uses: actions/cache@v5.0.1
         with:
           key: install4j-A${{ hashFiles('liquibase-dist/src/main/install4j/liquibase.install4j') }}
           path: ~/.install4j8/**
@@ -235,19 +235,19 @@ jobs:
           done
 
       - name: Archive Packages
-        uses: actions/upload-artifact@v5
+        uses: actions/upload-artifact@v6
         with:
           name: liquibase-artifacts
           path: artifacts/*
 
       - name: Archive Convenience Zip
-        uses: actions/upload-artifact@v5
+        uses: actions/upload-artifact@v6
         with:
           name: liquibase-zip-${{ needs.setup.outputs.thisBranchName }}
           path: branch-artifacts/*
 
       - name: Archive Convenience Artifacts
-        uses: actions/upload-artifact@v5
+        uses: actions/upload-artifact@v6
         with:
           name: liquibase-artifacts-${{ needs.setup.outputs.thisBranchName }}
           path: artifacts-named/*
--- a/.github/workflows/codeql.yml
+++ b/.github/workflows/codeql.yml
@@ -83,7 +83,7 @@ jobs:
         outputDir: ./reports/
 
     - name: Upload Security Report
-      uses: actions/upload-artifact@v5
+      uses: actions/upload-artifact@v6
       with:
         name: security-report
         path: ./reports/summary.pdf
--- a/.github/workflows/create-release.yml
+++ b/.github/workflows/create-release.yml
@@ -195,7 +195,7 @@ jobs:
           mv re-version/out/liquibase-additional-${{ needs.setup.outputs.version }}.zip re-version/final
 
       - name: Cache Completed Artifacts
-        uses: actions/cache@v4.3.0
+        uses: actions/cache@v5.0.1
         with:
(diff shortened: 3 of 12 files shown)


liquibase-standard/src/main/java/liquibase/sqlgenerator/core/RenameColumnGenerator.java
```diff
@@ -83,10 +83,35 @@ public Sql[] generateSql(RenameColumnStatement statement, Database database, Sql
...
+                    + " RENAME COLUMN " + database.escapeColumnName(statement.getCatalogName(), statement.getSchemaName(), statement.getTableName(), statement.getOldColumnName())
+                    + " TO " + database.escapeColumnName(statement.getCatalogName(), statement.getSchemaName(), statement.getTableName(), statement.getNewColumnName());
+        }
+
+        // Use CHANGE syntax when columnDataType is specified (for type changes) or when RENAME COLUMN is not supported
+        return "ALTER TABLE " + database.escapeTableName(statement.getCatalogName(), statement.getSchemaName(), statement.getTableName())
+                + " CHANGE " + database.escapeColumnName(statement.getCatalogName(), statement.getSchemaName(), statement.getTableName(), statement.getOldColumnName())
+                + " " + database.escapeColumnName(statement.getCatalogName(), statement.getSchemaName(), statement.getTableName(), statement.getNewColumnName())
+                + " " + DataTypeFactory.getInstance().fromDescription(columnDataType, database).toDatabaseDataType(database);
```
MalloD12, line 114: This is a question not from this current PR changes, but I was just looking at the `isRenameKeywordSupported()` function, and I was wondering whether the `else` block should not be checking if database is a `mysql` one to avoid this getting applied to any other DB that a major version greater or equals than `8`.

filipelautert: Good question! The concern isn't actually an issue here because the method signature isRenameKeywordSupported(MySQLDatabase database) already constrains the input to only MySQL-family databases.

  The else block can only be reached by MySQLDatabase instances (since MariaDBDatabase is caught by the if). No other database type can get here - Java's type system guarantees it.

  So the version check >= 8 is safe and will only ever apply to actual MySQL databases. No additional instanceof check needed! 👍

MalloD12: Good point, didn't realize the database parameter was already a MySQLDtabase instance. 

Thanks!

PR #7479, opened 2025-12-21
Extract checksum compatibility logic in StandardChangeLogHistoryService

maximevw:
## Impact
Enhancement
 

## Description
This is a simple refactoring extracting the query retrieving the incompatible changelogs in an overridable method.
Doing this is a prerequisite to solve this issue in `liquibase-cassandra` module: https://github.com/liquibase/liquibase-cassandra/issues/375.

Indeed, the query executed here is not compatible with Cassandra-CQL (not supporting neither `NOT NULL` nor `!=` operators). So, making this query easily overridable in the `liquibase-cassandra`, we will be able to write a method compatible with Cassandra databases providing the expected results.

## Things to be aware of
N/A

## Things to worry about
N∕A

## Additional Context
See the analysis provided here: https://github.com/liquibase/liquibase-cassandra/issues/375#issuecomment-3477924353

### DIFF
--- a/liquibase-standard/src/main/java/liquibase/changelog/StandardChangeLogHistoryService.java
+++ b/liquibase-standard/src/main/java/liquibase/changelog/StandardChangeLogHistoryService.java
@@ -250,7 +250,7 @@ public class StandardChangeLogHistoryService extends AbstractChangeLogHistorySer
             }
 
             //check if any checksum is not using the current version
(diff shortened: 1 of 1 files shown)


filipelautert (approved):

filipelautert:
Thanks for the improvement @maximevw !

maximevw:
You're welcome @filipelautert.
As soon as the change will be released and integrated in liquibase-cassandra, I'll suggest a fix in the cassandra extension for the original issue.

PR #7488, opened 2025-12-30
SEARCH_PATH setup as LOCAL for Postgres

### LINKED ISSUE #7092: Set of search_path for connection (PostgreSQL) should not be persistent
### Search first

### Description

I created a topic regarding the set of the `search_path` for PostgreSQL in the Liquibase forum: https://forum.liquibase.org/t/why-do-you-set-the-session-search-path/10542

As mentioned in multiple other issues (like for example https://github.com/liquibase/liquibase/issues/6227), Liquibase sets the `search_path` for PostgreSQL.

In our case we use the Java API (`org.liquibase:liquibase-core:4.29.1`) to define the schema to use via `liquibase.database.Database#setDefaultSchemaName`.

In the database log I can see then the following (as an example):

```
LOG:  execute <unnamed>: BEGIN
LOG:  execute <unnamed>: SHOW SEARCH_PATH
LOG:  execute <unnamed>: SET SEARCH_PATH TO customer, "$user","public"
LOG:  execute <unnamed>: UPDATE customer.databasechangeloglock SET LOCKED = FALSE, LOCKEDBY = NULL, LOCKGRANTED = NULL WHERE ID = 1
LOG:  execute PGBOUNCER_1: COMMIT
```

As we can see, that `search_path` is set to customer in this case. Now I wonder why this is done, as the following statement is fully qualified - the the schema is used here already (`UPDATE customer.databasechangeloglock`). I assume that this is possibly needed to prevent the need for defining the schema within native SQL queries. For every changeset which uses the descriptive language (like XML), Liquibase could use the full qualified name automatically - and I think it does already. Respectively I can see this also within the log:

```
LOG:  execute S_9: BEGIN
LOG:  execute S_10: SHOW SEARCH_PATH
LOG:  execute <unnamed>/C_24: CREATE TABLE customer.t_msglog_property (...)
LOG:  execute <unnamed>/C_25: CREATE INDEX string_value_index ON customer.t_msglog_property(c_name, c_string_value)
[...]
LOG:  execute S_2: COMMIT
```

Now the problem with this: We use PgBouncer with transaction pool mode and here the use of session based functionality/information is not allowed or supported. See also: [PgBouncer features](https://www.pgbouncer.org/features.html).

> The following table list various PostgreSQL features and whether they are compatible with PgBouncer pooling modes. **Note that “transaction” pooling breaks client expectations of the server by design and can be used only if the application cooperates by not using non-working features**.

| Feature            | Session pooling | Transaction pooling |
|--------------------|-----------------|---------------------|
| Startup parameters | Yes             | Yes                 |
| **SET/RESET**      | Yes             | **Never**           |
| LISTEN             | Yes             | Never               |
| NOTIFY             | Yes             | Yes                 |
| ...                | ...             | ...                 |

When you set the `search_path` (in the way you do it), the physical connection between PgBouncer and the database holds this information/configuration. And this connection is reused also for other services - so the services influence each other. And so the set/modification of the `search_path` by Liquibase is not a good idea in my opinion. Maybe using `SET LOCAL SEARCH_PATH TO ...` would be better here!?

So in the end the configuration of the `search_path` should not be persistent.

Hint: In the documentation of PgBouncer I looks like that a `SET/RESET` is never allowed/supported, but for when using `LOCAL` for setting such a property I would not expect any problems.

### Steps To Reproduce

* Startup PostgreSQL instance with statement logging (log_statement=all)
* Execute a Liquibase script
* Check the log

### Expected/Desired Behavior

The search_path should not be persistent in order to prevent it from influencing other services.

### Liquibase Version

4.29.1

### Database Vendor & Version

PostgreSQL 15.5

### Liquibase Integration

Maven

### Liquibase Extensions

_No response_

Ubuntu 22.04.5 LTS

### Additional Context

_No response_

MalloD12:
- Update initializeDatabase method to make SEARCH_PATH be set as Local (transaction scope) for Postgres.
- Test added.

## Impact

## Description

Liquibase uses persistent `SET SEARCH_PATH` commands for `PostgreSQL`, which sets the search_path for the entire database session. When using PgBouncer in transaction pooling mode, this causes cross-service interference because:

  1. PgBouncer reuses the same database connection across multiple client applications
  2. The search_path setting from one application persists on the connection
  3. The next application using that pooled connection inherits the wrong search_path

That's fixed by using  `SET LOCAL SEARCH_PATH` instead, which scopes the setting to the current transaction only. This ensures the search_path automatically reverts after each transaction commits, making Liquibase compatible with PgBouncer's transaction pooling while maintaining all existing functionality.

Fixes #7092

### DIFF
--- a/liquibase-standard/src/main/java/liquibase/database/core/DatabaseUtils.java
+++ b/liquibase-standard/src/main/java/liquibase/database/core/DatabaseUtils.java
@@ -60,7 +60,7 @@ public class DatabaseUtils {
                         });
                     }
 
-                    executor.execute(new RawParameterizedSqlStatement(String.format("SET SEARCH_PATH TO %s", finalSearchPath)));
+                    executor.execute(new RawParameterizedSqlStatement(String.format("SET LOCAL SEARCH_PATH TO %s", finalSearchPath)));
                 }
 
             } else if (database instanceof AbstractDb2Database) {
--- a/liquibase-standard/src/main/java/liquibase/sqlgenerator/core/CreateProcedureGenerator.java
+++ b/liquibase-standard/src/main/java/liquibase/sqlgenerator/core/CreateProcedureGenerator.java
@@ -161,11 +161,11 @@ public class CreateProcedureGenerator extends AbstractSqlGenerator<CreateProcedu
 
                 if (!originalSearchPath.equals(schemaName) && !originalSearchPath.startsWith(schemaName + ",") && !originalSearchPath.startsWith("\"" + schemaName + "\",")) {
                     if (database instanceof EnterpriseDBDatabase){
-                        sql.add(0, new UnparsedSql("ALTER SESSION SET SEARCH_PATH TO " + database.escapeObjectName(defaultSchema, Schema.class) + ", " + originalSearchPath));
-                        sql.add(new UnparsedSql("ALTER SESSION SET CURRENT SCHEMA " + originalSearchPath));
+                        sql.add(0, new UnparsedSql("SET LOCAL SEARCH_PATH TO " + database.escapeObjectName(defaultSchema, Schema.class) + ", " + originalSearchPath));
+                        sql.add(new UnparsedSql("SET LOCAL CURRENT SCHEMA " + originalSearchPath));
                     } else {
-                        sql.add(0, new UnparsedSql("SET SEARCH_PATH TO " + database.escapeObjectName(schemaName, Schema.class) + ", " + originalSearchPath));
-                        sql.add(new UnparsedSql("SET CURRENT SCHEMA " + originalSearchPath));
+                        sql.add(0, new UnparsedSql("SET LOCAL SEARCH_PATH TO " + database.escapeObjectName(schemaName, Schema.class) + ", " + originalSearchPath));
+                        sql.add(new UnparsedSql("SET LOCAL CURRENT SCHEMA " + originalSearchPath));
                     }
                 }
             }
--- a/liquibase-standard/src/main/java/liquibase/sqlgenerator/core/CreateProcedureGenerator.java
+++ b/liquibase-standard/src/main/java/liquibase/sqlgenerator/core/CreateProcedureGenerator.java
@@ -160,12 +160,7 @@ public class CreateProcedureGenerator extends AbstractSqlGenerator<CreateProcedu
                 }
 
                 if (!originalSearchPath.equals(schemaName) && !originalSearchPath.startsWith(schemaName + ",") && !originalSearchPath.startsWith("\"" + schemaName + "\",")) {
-                    if (database instanceof EnterpriseDBDatabase){
-                        sql.add(0, new UnparsedSql("SET LOCAL SEARCH_PATH TO " + database.escapeObjectName(defaultSchema, Schema.class) + ", " + originalSearchPath));
-                        sql.add(new UnparsedSql("ALTER SESSION SET CURRENT SCHEMA " + originalSearchPath));
-                    } else {
-                        sql.add(0, new UnparsedSql("SET LOCAL SEARCH_PATH TO " + database.escapeObjectName(schemaName, Schema.class) + ", " + originalSearchPath));
-                    }
+                    sql.add(0, new UnparsedSql("SET LOCAL SEARCH_PATH TO " + database.escapeObjectName(schemaName, Schema.class) + ", " + originalSearchPath));
                 }
             }
         }
--- a/liquibase-standard/src/main/java/liquibase/sqlgenerator/core/CreateProcedureGenerator.java
+++ b/liquibase-standard/src/main/java/liquibase/sqlgenerator/core/CreateProcedureGenerator.java
@@ -162,10 +162,9 @@ public class CreateProcedureGenerator extends AbstractSqlGenerator<CreateProcedu
                 if (!originalSearchPath.equals(schemaName) && !originalSearchPath.startsWith(schemaName + ",") && !originalSearchPath.startsWith("\"" + schemaName + "\",")) {
                     if (database instanceof EnterpriseDBDatabase){
                         sql.add(0, new UnparsedSql("SET LOCAL SEARCH_PATH TO " + database.escapeObjectName(defaultSchema, Schema.class) + ", " + originalSearchPath));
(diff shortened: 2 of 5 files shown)


liquibase-standard/src/main/java/liquibase/database/core/DatabaseUtils.java
```diff
@@ -60,7 +60,7 @@ public static void initializeDatabase(String defaultCatalogName, String defaultS
                         });
                     }
 
-                    executor.execute(new RawParameterizedSqlStatement(String.format("SET SEARCH_PATH TO %s", finalSearchPath)));
+                    executor.execute(new RawParameterizedSqlStatement(String.format("SET LOCAL SEARCH_PATH TO %s", finalSearchPath)));
```
filipelautert, line 63: Did you check if `LOCAL SEARCH_PATH` is compatible with all Postgresql versions?

filipelautert (changes requested):

liquibase-standard/src/main/java/liquibase/database/core/DatabaseUtils.java
MalloD12: From what I could see this is safe, since `SET LOCAL` has been introduced in a `Postgres 7.3`, and we have  support starting from version 12+ (see [here](https://docs.liquibase.com/oss/integration-guide-4-33/what-support-does-liquibase-have-for-postgresql)). So, it seems this work in all postgres version and there is no need to add any additional treatment for a specific version.

liquibase-standard/src/main/java/liquibase/sqlgenerator/core/CreateProcedureGenerator.java
```diff
@@ -160,13 +160,7 @@ public static void surroundWithSchemaSets(List<Sql> sql, String schemaName, Data
...
 
                 if (!originalSearchPath.equals(schemaName) && !originalSearchPath.startsWith(schemaName + ",") && !originalSearchPath.startsWith("\"" + schemaName + "\",")) {
-                    if (database instanceof EnterpriseDBDatabase){
-                        sql.add(0, new UnparsedSql("ALTER SESSION SET SEARCH_PATH TO " + database.escapeObjectName(defaultSchema, Schema.class) + ", " + originalSearchPath));
-                        sql.add(new UnparsedSql("ALTER SESSION SET CURRENT SCHEMA " + originalSearchPath));
-                    } else {
-                        sql.add(0, new UnparsedSql("SET SEARCH_PATH TO " + database.escapeObjectName(schemaName, Schema.class) + ", " + originalSearchPath));
-                        sql.add(new UnparsedSql("SET CURRENT SCHEMA " + originalSearchPath));
-                    }
+                    sql.add(0, new UnparsedSql("SET LOCAL SEARCH_PATH TO " + database.escapeObjectName(schemaName, Schema.class) + ", " + originalSearchPath));
```
filipelautert, line 163: Aren't we over simplifying here? We had a special check for enterprisedb that now is gone.

filipelautert (changes requested):

liquibase-standard/src/main/java/liquibase/sqlgenerator/core/CreateProcedureGenerator.java
MalloD12: If I think the changes are accurate. Some reasons are:

- First, EnterpriseDB is an extension of Postgres and supports all standard PostgreSQL syntax, so these changes will now apply to both.
- EnterpriseDB  block also contained a bug (used defaultSchema instead of schemaName)                                                                                                  
- The new approach uses SET LOCAL, which is transaction-scoped and cleaner.                                                                                                               

In my opinion, we are good with this PR as it is, but to be honest, I haven't manually tested it against an Enterprise DB.

filipelautert (approved):

MalloD12:
@tati-qalified, when testing this issue, could you please also test against an EnterpriseDB? We have removed some code that, according to what we have seen, EnterpriseDB should behave fine with the new implementation, but since we haven't tested it against we would like to make sure we are not breaking it.

cc: @filipelautert @rberezen 

Thanks
Daniel.

liquibase-standard/src/main/java/liquibase/sqlgenerator/core/CreateProcedureGenerator.java
tati-qalified: @MalloD12 @filipelautert did you reach an agreement about this? Are we keeping it as is?

MalloD12: Yeap, the agreement we made is to leave it as it is. Then, if while testing, we see EnterpriseDB does not behave as before, we can go back and revert part of these changes (and return the specific EnterpriseDB logic we removed in this PR).

tati-qalified (approved):

MalloD12:
Moving this to RTM.

cc: @filipelautert

rberezen (commented):
Tested the fix on PostgreSQL 17 with statement logging enabled.

**Before master:** `SET SEARCH_PATH TO customer, ...` — persists on connection after transaction
**After PR:** `SET LOCAL SEARCH_PATH TO customer, ...` — reverts to default after COMMIT

Verified with `SHOW SEARCH_PATH` on a fresh query after the update completes: returns `"$user", public` (default), not `customer, ...`. This is exactly what PgBouncer transaction pooling requires.

The `CreateProcedureGenerator` simplification is cleaner — removes the post-procedure restore since `SET LOCAL` auto-reverts. One note: the `EnterpriseDBDatabase` code path changed — it previously used `ALTER SESSION SET SEARCH_PATH` and now uses `SET LOCAL SEARCH_PATH` like regular Postgres. If EDB requires the old syntax, this could regress there — but CI tests pass.

LGTM. Nice fix for #7092.

rberezen (approved):

rberezen (commented):
Follow-up on my earlier EDB concern — I verified on EDB 14.4 (EnterpriseDB Advanced Server) that `SET LOCAL SEARCH_PATH` works identically to vanilla PostgreSQL.

Tested `liquibase update --default-schema-name=test_schema` against EDB 14.4:
- Update succeeds, table created in `test_schema.test_tbl`
- After COMMIT, `SHOW SEARCH_PATH` returns `"$user", public` (default — reverted correctly)

No regression for EDB. The simplification is safe.

PR #7542, opened 2026-01-29
fix: Capture float precision again on Oracle DBs

marchof:
## Impact

## Description

With 8f76ffb the capturing of the float precision setting was missed in the new class ColumnSnapshotGeneratorOracle. Therefore precision is now ignored when creating diffs.

This change adds an integration test demonstrating the problem and re-adds reading the precision of float columns for Oracle.

### DIFF
--- a/liquibase-standard/src/test/java/liquibase/test/SnapshotAssert.java
+++ b/liquibase-standard/src/test/java/liquibase/test/SnapshotAssert.java
@@ -1,5 +1,9 @@
 package liquibase.test;
 
+import static org.junit.Assert.fail;
+
+import java.util.Optional;
+import java.util.function.Consumer;
 import java.util.function.Predicate;
 import liquibase.snapshot.DatabaseSnapshot;
 import liquibase.structure.DatabaseObject;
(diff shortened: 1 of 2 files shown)


marchof:
Hi @filipelautert, can you please approve the GitHub workflow "Unit and Integration Tests" so that I can see whether the tests run in your CI. Execution on my fork was green (beside some failures due to missing fossa-scan credentials - I wonder whether this task might be better a separate workflow, so that contributors can run the integration tests in their accounts, but this is a different topic).

filipelautert:
> Hi @filipelautert, can you please approve the GitHub workflow "Unit and Integration Tests" so that I can see whether the tests run in your CI. Execution on my fork was green (beside some failures due to missing fossa-scan credentials - I wonder whether this task might be better a separate workflow, so that contributors can run the integration tests in their accounts, but this is a different topic).

@marchof sure, sorry for the delay!

marchof:
@filipelautert Looks like integration tests are green. Also we tested a patched Liquibase version on our systems and precision is applied correctly with this patch.

filipelautert (approved):
Clean fix for a real regression. Thanks for the PR @marchof !

marchof:
Can you please give @quangkhainguyen credits for this? This is my colleague who actually found and debugged the issue. Thx!!!

liquibase-standard/src/test/java/liquibase/test/SnapshotAssert.java
tati-qalified, line 1: Are these changes necessary? SnapshotAssert is only for tests, so you could modify your tests to fit the current behaviour.

marchof: `SnapshotAssert` is a useful tool to check for the existence for a specific database object in the snapshot. For the new tests we also need the selected object to assert properties on them. I thought we can use the existing `SnapshotAssert` methods to return the selected objects instead of the `SnapshotAssert` instance (especially as the current return value is not used so far anyways).

I have no strong opinion here: If you prefer we can leave `SnapshotAssert` untouched and replicate the logic in the new test. Another option would be to add a new `SnapshotAssert` method which accepts a `Consumer` so the caller can inject its own verification code.

Just let me know what you prefer.

marchof: Hi @tati-qalified, please let me know how I should proceed. I'm happy to adjust the test to your requirements.

tati-qalified: @marchof adding a new `SnapshotAssert` method would be the best way to move forward. Thank you!

marchof: Hi @tati-qalified, I pushed a new commit with does not change the existing API of `SnapshotAssert` anymore. Anyhow I changed the signature of the existing private method, otherwise I would need to copy & paste most of the code to an new method. I hope this is acceptable.

marchof:
@filipelautert May I ask you to approve the GitHub workflow "Unit and Integration Tests" again? Sorry.

liquibase-standard/src/test/java/liquibase/test/SnapshotAssert.java
```diff
@@ -19,16 +20,21 @@
 	}
 
 	public SnapshotAssert containsObject(DatabaseObject object) {
-		return checkContainsObject(object.getClass(), object::equals, object + " not found");
+		return checkContainsObject(object.getClass(), object::equals, ignore -> {}, object + " not found");
```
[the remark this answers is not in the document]
marchof: Correct, because in this case we have a no-op consumer. But we still need to give a parameter name to fulfill the functional interface. That's why I called it "ignore". Since Java 22 we can use an underscore _ here.

@filipelautert How do I address this code scanning issue?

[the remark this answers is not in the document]
filipelautert: I'm dismissing it, thanks.

PR #7560, opened 2026-02-13
fix: Update list of reserved keywords for MySQL and MariaDB

### LINKED ISSUE #7559: Reserved keywords for MySQL and MariaDB are not up-to-date
### Search first

### Description

A few keywords are not quoted correctly. When e.g. a table should be created with such a keyword as name, this results in a SQL syntax error.

This results in a failed run:
```
2026-02-13T18:36:46.440+01:00 ERROR 19968 --- [demo] [           main] o.s.boot.SpringApplication               : Application run failed

org.springframework.beans.factory.BeanCreationException: Error creating bean with name 'liquibase' defined in class path resource [org/springframework/boot/liquibase/autoconfigure/LiquibaseAutoConfiguration$LiquibaseConfiguration.class]: liquibase.exception.CommandExecutionException: liquibase.exception.LiquibaseException: liquibase.exception.MigrationFailedException: Migration failed for changeset db/changelog/db.changelog-master.xml::1::my_name:
     Reason: liquibase.exception.DatabaseException: You have an error in your SQL syntax; check the manual that corresponds to your MySQL server version for the right syntax to use near 'parallel INT NULL, CONSTRAINT PK_TEST_TABLE PRIMARY KEY (test_id))' at line 1 [Failed SQL: (1064) CREATE TABLE mydatabase.test_table (test_id INT NOT NULL, parallel INT NULL, CONSTRAINT PK_TEST_TABLE PRIMARY KEY (test_id))]
```

### Steps To Reproduce

1. Create a new Spring Boot application with Liquibase and MySQL. You can use the [preset here](https://start.spring.io/#!type=gradle-project&language=java&platformVersion=4.0.2&packaging=jar&configurationFileFormat=properties&jvmVersion=17&groupId=com.example&artifactId=demo&name=demo&description=Demo%20project%20for%20Spring%20Boot&packageName=com.example.demo&dependencies=liquibase,docker-compose,mysql).
2. Configure in application.properties (because it searches for yaml as default now apparently, different to [what the docs say](https://contribute.liquibase.com/extensions-integrations/directory/integration-docs/springboot/#add-a-changelog-file)): `spring.liquibase.change-log=classpath:db/changelog/db.changelog-master.xml`
3. Write the following in `src/main/resources/db/changelog/db.changelog-master.xml`:

```xml
<databaseChangeLog
        xsi:schemaLocation="http://www.liquibase.org/xml/ns/dbchangelog http://www.liquibase.org/xml/ns/dbchangelog/dbchangelog-latest.xsd">

    <changeSet id="1" author="my_name">
        <createTable tableName="test_table">
            <column name="test_id" type="int">
                <constraints primaryKey="true"/>
            </column>
            <column name="parallel" type="INT"/>
        </createTable>
    </changeSet>

</databaseChangeLog>
```
4. Start the project with `./gradlew bootRun`

### Expected/Desired Behavior

Liquibase should run the `CREATE` statement with quotes around `parallel` because it is a keyword. The table should be created without issues.

### Liquibase Version

liquibase-core-5.0.1.jar

### Database Vendor & Version

MySQL 9.6.0

### Liquibase Integration

Spring Boot

### Liquibase Extensions

_No response_

_No response_

### Additional Context

Current reserved MySQL 9.6 keywords: https://dev.mysql.com/doc/refman/9.6/en/keywords.html

MariaDB reserved words: https://mariadb.com/docs/server/reference/sql-structure/sql-language-structure/reserved-words

profiluefter:
## Impact

## Description

This MR updates the list of reserved keywords for both MySQL and MariaDB. There were quite a few less-common used words missing but also common words like `parallel` or `offset` (MariaDB only).

I have included a short JS snippet for both pages that can be executed to extract the list in a format that can be pasted into the code. For the different MySQL versions I have diffed the list of keywords between them and modified the list based on the version which was detected. 

## Things to be aware of

This list is important for proper quoting of names in SQL statements. It is not otherwise used at the moment, so having too many words in there is better than too little I think.

## Things to worry about

Quite a few keywords from MySQL are not reserved in MariaDB syntax but MariaDB extends MySQL here so they will be included. Also the versioning from MariaDB and MySQL do not match so the function `addMySQLVersionedReservedWords`, which will also be executed for MariaDB, will work with bogus version numbers. This is not a big concern at the moment because MariaDB versions should be bigger than the biggest MySQL version (MariaDB 10 was released in 2014).

For MySQL there were many words in the list which did not have the `(R)` appended in the docs page, so they are keywords but not reserved names so they do not need to be quoted. I tested creating a table with a column named like all keywords that I removed.

There are not yet tests for this change because I was unsure on how to test this. I think a test which creates a table with columns with all these names would be one way to do it. Other similar PRs (linked below) did not test such a change.

## Additional Context

Current reserved MySQL 9.6 keywords: https://dev.mysql.com/doc/refman/9.6/en/keywords.html

MariaDB reserved words: https://mariadb.com/docs/server/reference/sql-structure/sql-language-structure/reserved-words

Fixes #7559

Other PRs that touched this (that I could find) are: #6128 and #6697


liquibase-standard/src/main/java/liquibase/database/core/MariaDBDatabase.java
```diff
@@ -19,7 +19,33 @@ public class MariaDBDatabase extends MySQLDatabase {
...
     public MariaDBDatabase() {
-        addReservedWords(Collections.singletonList("PERIOD"));
+        /*
+         * From https://mariadb.com/docs/server/reference/sql-structure/sql-language-structure/reserved-words
+         * A list can be extracted with the following JS snippet:
+         * console.log($x("//h2[@id='reserved-words']/following-sibling::div[1]//div[@role='row']/div[@role='cell'][1]//p")
+         *     .map(x => `"${x.textContent.trim()}"`).join(",\n"))
+         * Most of them are already covered by MySQLDatabase.
+         */
+        addReservedWords(List.of("CURRENT_ROLE",
```
filipelautert, line 29: This new list drops `PERIOD`, which the previous code (`Collections.singletonList("PERIOD")`) added explicitly. It was introduced deliberately in commit *add PERIOD to reserved words for MariaDB (fixes CORE-3437)* — your extraction snippet won't surface it because the current docs page doesn't list it as reserved, but it caused a real quoting bug. Removing it re-opens CORE-3437.

Given your own "too many is better than too few" principle, please add `"PERIOD"` back to the list.

liquibase-standard/src/main/java/liquibase/database/core/MySQLDatabase.java
```diff
@@ -713,19 +655,71 @@ public void addReservedWords(Collection<String> words) {
...
+                        "SYSTEM",
+                        "WINDOW"
+                ));
+            }
+
             // words that became reserved in 8.4
-            if(getDatabaseMajorVersion() >= 9 || (getDatabaseMajorVersion() == 8 && getDatabaseMinorVersion() >= 4)) {
+            if (major >= 9 || (major == 8 && minor >= 4)) {
+                reservedWords.remove("MASTER_BIND");
+                reservedWords.remove("MASTER_SSL_VERIFY_SERVER_CERT");
```
filipelautert, line 704: Heads up on an interaction with MariaDB: `MariaDBDatabase` declares `MASTER_SSL_VERIFY_SERVER_CERT` as reserved, but this `remove(...)` also runs for MariaDB (its major version is always >= 9), so the word gets un-reserved once a real connection's version is read (the `JdbcConnection` re-call after the constructor). Net effect: it won't be quoted for MariaDB, contradicting the intent in that list.

Low impact since it works unquoted there anyway, but worth a guard (skip these removals for MariaDB) or at least a comment so it isn't silently undone.

filipelautert (changes requested):
Hey, reviewed the changes. Solid research and the version-aware approach is a real improvement. One regression to fix before merge (PERIOD), plus a couple of smaller notes left inline.

**Tests** — you asked how to test this. A small unit test that sets a mock connection version and asserts `isReservedWord(...)` per branch would lock in the new logic and guard against the PERIOD regression: e.g. `PARALLEL` reserved on 8.4+ but not 8.0, and `PERIOD` always reserved on MariaDB. Happy to point you at an existing `*DatabaseTest` to mirror if useful.

The documented MariaDB/MySQL version-number conflation (`LIBRARY`/`EXTERNAL`/`SETS` landing on MariaDB via `major >= 10`) is fine as an accepted over-quoting tradeoff — no need to change that. Nice work on the per-version diffing and the extraction snippets.

liquibase-standard/src/main/java/liquibase/database/core/MariaDBDatabase.java
profiluefter: I have added it back 👍 didn't notice the original ticket; interesting edge case! I would not have expected there to be library specific quote requirements 😅

liquibase-standard/src/main/java/liquibase/database/core/MySQLDatabase.java
profiluefter: I didn't notice this initially, however while diffing the latest MySQL against MariaDB i have added it back in the MariaDB keywords already and those get added after this remove is run, so it is in the list again. The tests validate this now and I have added a comment to the line that removes it.

profiluefter:
@filipelautert thanks for the good input! I have now added tests using JUnits ParameterizedTest so that it is easy to expand and covers a lot without a huge amount of code per version. It increases the complexity of the tests a bit but I hope it's okay because it is very self contained.

liquibase-standard/src/test/java/liquibase/database/core/MySQLDatabaseTest.java
```diff
@@ -98,42 +100,39 @@ public void escapeStringForDatabase_withBackslashes() {
...
+                forVersion(5, 7, List.of("MASTER_SSL_VERIFY_SERVER_CERT"), List.of("FUNCTION", "ROW", "ROWS", "PERIOD", "CURRENT_ROLE")),
+                forVersion(8, 0, List.of("FUNCTION", "ROW", "ROWS", "MASTER_BIND"), List.of()),
+                forVersion(8, 4, List.of("FUNCTION", "ROW", "ROWS", "MANUAL"), List.of("MASTER_SSL_VERIFY_SERVER_CERT")),
+                // no changes in 9.0
+                forVersion(9, 0, List.of("FUNCTION", "ROW", "ROWS", "MANUAL"), List.of("MASTER_SSL_VERIFY_SERVER_CERT", "LIBRARY")),
+                forVersion(9, 3, List.of("LIBRARY", "MANUAL", "SYSTEM"), List.of("MASTER_BIND", "EXTERNAL")),
+                forVersion(9, 4, List.of("EXTERNAL", "LIBRARY", "SYSTEM"), List.of("SETS", "MASTER_BIND", "PERIOD")),
+                forVersion(9, 6, List.of("SETS", "EXTERNAL", "LIBRARY", "CUBE", "ACCESSIBLE"), List.of("SOMETHING_ELSE", "MASTER_BIND")),
+                forVersion(10, 0, List.of("SETS", "EXTERNAL", "LIBRARY", "CUBE", "ACCESSIBLE"), List.of("SOMETHING_ELSE", "MASTER_BIND"))
+        ).flatMap(s -> s);
```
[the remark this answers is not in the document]
profiluefter: MySQL 9.7 removes this keyword again, which is a bit funny. I have added it to the tests.

filipelautert (approved):
PERIOD is back and the version matrix tests are exactly what I asked for, thanks! Just a typo at the 9.7 comment in MySQLDatabase.java, 'sopped' -> 'stopped'.

liquibase-standard/src/main/java/liquibase/database/core/MariaDBDatabase.java
filipelautert: thanks!

profiluefter:
typo is fixed and i rebased the PR on main :+1:

filipelautert (approved):
Thanks!

kesleyfort:
**Verdict: PASS**

## Manual repro (CLI) — before/after evidence

Pre-existing state was deployed first (changeset `0`, a plain table) so the keyword changeset ran against a non-empty target; the failed baseline deploy was then re-run with the fixed build — the customer's upgrade path. Changelog changeset `1` mirrors the issue verbatim:

```xml
<createTable tableName="test_table">
    <column name="test_id" type="int"><constraints primaryKey="true"/></column>
    <column name="parallel" type="INT"/>   <!-- "offset" on MariaDB -->
</createTable>
```

### MySQL 9.6 (`mysql:9.6`, column `parallel`)

```
# BUGGY (baseline 9a9ea95) — update-sql emits the identifier unquoted, update fails
CREATE TABLE intuser_db.test_table (test_id INT NOT NULL, parallel INT NULL, CONSTRAINT PK_TEST_TABLE PRIMARY KEY (test_id));
Running Changeset: changelog-mysql.xml::0::qa        # pre-existing state deploys fine
Running Changeset: changelog-mysql.xml::1::my_name
Unexpected error running Liquibase: Migration failed for changeset changelog-mysql.xml::1::my_name:
     Reason: liquibase.exception.DatabaseException: You have an error in your SQL syntax; ... near
     'parallel INT NULL, CONSTRAINT PK_TEST_TABLE PRIMARY KEY (test_id))' at line 1
     [Failed SQL: (1064) CREATE TABLE intuser_db.test_table (test_id INT NOT NULL, parallel INT NULL, ...)]
```

```
# FIXED (PR 8fa6570) — identifier quoted, update succeeds on the same target
CREATE TABLE intuser_db.test_table (test_id INT NOT NULL, `parallel` INT NULL, CONSTRAINT PK_TEST_TABLE PRIMARY KEY (test_id));
Running Changeset: changelog-mysql.xml::1::my_name
Run:                          1
-- SHOW CREATE TABLE: `parallel` int DEFAULT NULL   ✓ table created
```

### MariaDB 11.4 (harness `docker-compose-11.4.yml`, column `offset`)

```
# BUGGY (baseline 9a9ea95)
CREATE TABLE intuser_db.test_table (test_id INT NOT NULL, offset INT NULL, CONSTRAINT PK_TEST_TABLE PRIMARY KEY (test_id));
Running Changeset: changelog-mariadb.xml::1::my_name
     Reason: liquibase.exception.DatabaseException: (conn=5) You have an error in your SQL syntax; ... near
     'offset INT NULL, CONSTRAINT PK_TEST_TABLE PRIMARY KEY (test_id))' at line 1 [Failed SQL: (1064) ...]
```

```
# FIXED (PR 8fa6570)
CREATE TABLE intuser_db.test_table (test_id INT NOT NULL, `offset` INT NULL, CONSTRAINT PK_TEST_TABLE PRIMARY KEY (test_id));
Running Changeset: changelog-mariadb.xml::1::my_name
Run:                          1
-- SHOW CREATE TABLE: `offset` int(11) DEFAULT NULL   ✓ table created
```

kesleyfort (approved):

PR #7565, opened 2026-02-18
Fix global logicalFilePath not applied in formatted SQL changelogs

### LINKED ISSUE #7388: logicalFilePath inheritance does not work as expected in formatted SQL changelogs
### Search first

### Description

According to the examples in the [documentation](https://docs.liquibase.com/reference-guide/changelog-attributes/logicalfilepath), formatted SQL changelogs should support setting a global logicalFilePath using:
```sql
--liquibase formatted sql logicalFilePath:xyz
```
However, in practice, this value is not applied to the individual changeSets as expected. I’ve tested multiple combinations involving parent changelogs (YAML) and included formatted SQL changelogs, and in many cases the resulting `ChangeSet.storedFilePath` values do not reflect the expected inheritance behavior.

### Steps To Reproduce

Use combination of following change logs to cover all scenarios

SQL change log wit per changeSet setting
```sql
--liquibase formatted sql

--changeset 1.0.0:1 logicalFilePath:create-table-application
create table if not exists application
(
    id uuid  primary key,
    name varchar(100) not null,
);
--rollback DROP TABLE application;
```
SQL change log with global setting
```sql
--liquibase formatted sql logicalFilePath:create-table-application

--changeset 1.0.0:1
create table if not exists application
(
    id uuid  primary key,
    name varchar(100) not null,
);
--rollback DROP TABLE application;
```
SQL change log without setting
```sql
--liquibase formatted sql 

--changeset 1.0.0:1
create table if not exists application
(
    id uuid  primary key,
    name varchar(100) not null,
);
--rollback DROP TABLE application;
```
YAML parent change log with setting
```yaml
databaseChangeLog:
   - logicalFilePath: db.changelog-master
  - include:
      file: create-table-application.sql
      relativeToChangelogFile: true
```
YAML parent change log without setting
```yaml
databaseChangeLog:
  - include:
      file: create-table-application.sql
      relativeToChangelogFile: true
```
### Expected/Desired Behavior

I would expect Liquibase to apply the nearest defined logicalFilePath to each changeSet, following this hierarchy:

1. If defined on the changeSet, use that.
2. If not defined on the changeSet, use the value from the changelog file it belongs to.
3. If not defined in the changelog file, inherit from the nearest parent changelog that defines it.
4. If not defined anywhere, fall back to the physical file path of the changelog.

This should work consistently across changelog formats (YAML, SQL, etc.), and the values of changeSet.logicalFilePath, changeSet.filePath, and changeSet.storedFilePath should reflect their names:

`logicalFilePath`: the logicalFilePath (explicit or inherited)
`filePath`: the actual file path of the changelog
`storedFilePath`: effective logicalFilePath or filePath when logicalFilePath is `null`

### Liquibase Version

5.0.1

### Database Vendor & Version

PostgreSQL 13.21

### Liquibase Integration

_No response_

### Liquibase Extensions

_No response_

Ubuntu 25.04

### Additional Context

| Scenario | Expected `ChangeSet.storedPath` | Actual `ChangeSet.storedFilePath` |
|-|-|-|
| Parent changelog with logicalFilePath, SQL changelog with global logicalFilePath changelog value | **SQL Changelog global value** | Parent changelog value |
| Parent changelog with logicalFilePath, SQL changelog with per-changeSet logicalFilePath  |	**per-changeSet value** | Parent changelog value |
| Parent changelog with logicalFilePath, SQL changelog without logicalFilePath | Parent changelog value | Parent changelog value |
| No parent changelog, SQL changelog with global logicalFilePath	changelog value | **SQL Changelog global value** | full path of SQL file |
| No parent changelog, SQL changelog with per-changeSet logicalFilePath | per-changeSet value | per-changeSet value |
| No parent changelog, SQL changelog without logicalFilePath | full path of SQL file | full path of SQL file |
| Parent changelog without logicalFilePath, SQL changelog with per-changeSet logicalFilePath | per-changeSet value | per-changceSet value |
| Parent changelog without logicalFilePath, SQL changelog with global logicalFilePath | **SQL changelog global value** | full path of SQL file |
| Parent changelog without logicalFilePath, SQL changelog without global logicalFilePath | full path of SQL file | full path of SQL file |

Lumenol:
Use `getRawLogicalFilePath()` instead of `getLogicalFilePath()` in `AbstractFormattedChangeLogParser.setLogicalFilePath()` so the global `logicalFilePath` declared in formatted SQL headers is correctly stored and applied to changesets.

## Impact

## Description

Fixes #7388

In `AbstractFormattedChangeLogParser.setLogicalFilePath()`, the condition `changeLog.getLogicalFilePath() == null` was always `false` because `getLogicalFilePath()` falls back to `physicalFilePath` when no explicit logicalFilePath is set. This made the global `logicalFilePath` from `--liquibase formatted sql logicalFilePath:xyz` headers silently ignored.

The regression was introduced in commit c2df44ed9 (#7200) when a null guard was added to prevent duplicate header overwrite. The guard used `getLogicalFilePath()` instead of `getRawLogicalFilePath()`.

This complements PR #7451 which fixed logicalFilePath inheritance for YAML/XML changelogs but did not address the formatted SQL parsing case.

## Things to be aware of

- One-line fix: `getLogicalFilePath()` → `getRawLogicalFilePath()` in `AbstractFormattedChangeLogParser.setLogicalFilePath()`
- New unit tests in `FormattedSqlChangeLogParserTest.groovy` (no existing logicalFilePath tests existed)
- New integration tests in `ChangeLogIncludeTest.groovy` with SQL test fixtures, mirroring existing YAML tests from #7451

## Things to worry about

- The fix preserves the original intent of c2df44ed9: preventing duplicate header overwrite. With `getRawLogicalFilePath()`, once set, the guard correctly prevents re-setting.

## Additional Context

- Related PR: #7451 (Fixed undesired logicalFilePath inheritance of included changesets)
- Regression commit: c2df44ed9 ("Remove duplicate header check for formatted SQL", #7200)
- Discussion and validation: https://github.com/liquibase/liquibase/issues/7388#issuecomment-3883252915


MalloD12 (approved):
LGTM, non-related failing tests to this PR have already been fixed.

Thank you, @Lumenol!

filipelautert (approved):
Hey, reviewed this. Clean one-liner that nails the root cause — getLogicalFilePath() falls back to physicalFilePath so the guard was always false. Swapping to getRawLogicalFilePath() is exactly right.

Test coverage is thorough — unit tests cover global, per-changeset override, fallback, and mixed scenarios, plus integration tests on top. Nothing to flag here.

LGTM, ready to merge.

rberezen (commented):
@Lumenol 

## Test Review

**Unit tests (FormattedSqlChangeLogParserTest)**: All 80 pass, including the 4 new logicalFilePath tests. The core fix looks correct.

**Integration test failure**: The test `"formatted SQL child changelog's own logicalFilePath takes highest priority over include statement and parent"` fails locally.

The test expects the child SQL file's own `logicalFilePath` (`child-own-logical-path`) to override the include statement's `logicalFilePath` (`include-statement-path`), but the include statement wins. The log shows:

```
ChangeSet include-statement-path::sql_child_own_test::testauthor ran successfully
```

This happens because the include statement sets the logicalFilePath on the changelog *before* the parser reads the file header, so when `setLogicalFilePath()` checks `getRawLogicalFilePath() == null`, it's already set and the child's own header value is skipped.

**CI blind spot**: `ChangeLogIncludeTest` is **never executed in CI** because:
- The unit test step (`-P skip-integration-tests`) only runs `*ExtensionTests` from `liquibase-integration-tests`
- The integration test step uses `-Dtest=*IntegrationTest,*ExecutorTest`
- `ChangeLogIncludeTest` matches neither pattern

This affects all 10 tests in the class, not just the new ones. The class should probably be renamed to `ChangeLogIncludeIntegrationTest` to be picked up by the integration test filter.

@filipelautert @MalloD12

wwillard7800 (approved):
This all looks good.

freddiN:
Is this waiting for something? I ran into this today and came here through https://github.com/liquibase/liquibase/issues/7388

freddiN:
One month later without visible progress, I'll try it again: Is this good to go?

filipelautert:
@rberezen please re-review.
@Lumenol I took the liberty of fixing the tests as described.

filipelautert (approved):
Hey, re-reviewed. The production change is still the single targeted line — `getRawLogicalFilePath()` instead of `getLogicalFilePath()` in `AbstractFormattedChangeLogParser.setLogicalFilePath()` — which is exactly right: `getLogicalFilePath()` falls back to the physical path, so the global header value was never seen as "unset" and never applied. Everything since my March approval is just `main` merges plus the `ChangeLogIncludeTest` → `…IntegrationTest` rename, and all real CI (every Java/OS + integration suites) is green. Re-approving.

rberezen (commented):
Re-tested the updated PR. The test class rename is good — CI now picks it up. The core fix (`getLogicalFilePath()` → `getRawLogicalFilePath()` in `setLogicalFilePath()`) is the same and the unit tests pass.

**However**, running `ChangeLogIncludeIntegrationTest` locally produces 1 failure out of 10: the `"formatted SQL child changelog's own logicalFilePath takes highest priority over include statement and parent"` test still fails for me with:
```
Expected: ChangeSet child-own-logical-path::sql_child_own_test::testauthor
Actual:   ChangeSet include-statement-path::sql_child_own_test::testauthor
```

CI shows 10/10 passing (2.0s runtime) but my local consistently fails (7.8s runtime) — same code, same flags. There may be test state pollution from earlier tests in the class affecting this one (it's the 10th and last test). Worth investigating before merge — if the test passes only when run alongside the others, the assertion might be relying on side effects rather than verifying the intended behavior.

PR #7566, opened 2026-02-19
fix: reduce misleading WARN log when CLOB value contains spaces

### LINKED ISSUE #7386: Misleading warning when TEXT column (mapped as CLOB) contains spaces in value during loadUpdateData
### Search first

### Description

When inserting data using loadUpdateData into a PostgreSQL table that has a TEXT column, Liquibase treats the column as a CLOB.
As a result, it first tries to interpret the column value as a file name and load its content.

If the value is not a valid URI — for example, if it contains spaces — Liquibase logs a warning with a full stack trace:

```
WARN liquibase.resource - Error handling URI syntax for file inside jar. Defaulting to previous behavior.
java.net.URISyntaxException: Illegal character in path at index 27: /liquibase/changelog/dev/my value with spaces
    at java.base/java.net.URI$Parser.fail(URI.java:2995)
    ...
    at liquibase.resource.AbstractResource.handlePathForJarfile(AbstractResource.java:45)
```

The warning is harmless (Liquibase falls back correctly and inserts the value),
but the message and stack trace are misleading and noisy, especially in automated test logs.

### Steps To Reproduce

PostgreSQL table:

```
CREATE TABLE rec_characteristic_spec (
    id UUID PRIMARY KEY,
    name TEXT,
    value_type TEXT,
    external_system TEXT,
    description TEXT
);
```

Changelog (db/changelog/dev/my-changelog.yaml):
```
changes:
  - loadUpdateData:
      relativeToChangelogFile: true
      file: csv/202510101032_my_changelog/data.csv
      tableName: rec_characteristic_spec
      primaryKey: id
      separator: ","
      commentLineStartsWith: "#"
```

CSV file (csv/202510101032_my_changelog/data.csv):

```
id,name,value_type,external_system,description
d8b8caad-a950-4353-a816-2ab3a3caae2d,CAPACITY_UNIT,STRING,null,my value with spaces
```

## Actual behavior

Liquibase logs the following:

```
WARN liquibase.resource - Error handling URI syntax for file inside jar. Defaulting to previous behavior.
java.net.URISyntaxException: Illegal character in path at index 27: /liquibase/changelog/dev/my value with spaces
```

The stack trace originates from:
[AbstractResource.handlePathForJarfile()](https://github.com/liquibase/liquibase/blob/af3191aa72370e4fa9e27198247f6e0eb3404c22/liquibase-standard/src/main/java/liquibase/resource/AbstractResource.java#L47)

### Expected/Desired Behavior

Since this situation is harmless and expected when TEXT/CLOB values contain spaces:

- The log level should be lowered (e.g., DEBUG instead of WARN),
and/or
- The stack trace should be omitted to avoid misleading error noise.
- Optionally, the path handling could be URI-escaped before constructing the URI to prevent the exception.

### Liquibase Version

5.0.1

### Database Vendor & Version

PostgreSQL 17.5

### Liquibase Integration

Spring executed with JAR

### Liquibase Extensions

_No response_

Ubuntu

## Proposed solutions

1. Lower the log level of this message to DEBUG or FINE and uppress the stack trace when the URISyntaxException is harmless.
3. Optionally, escape the path before creating the URI

jstastny:
## Impact

## Description

Fixes #7386

When using `loadUpdateData` with a PostgreSQL `TEXT` column (mapped as CLOB), Liquibase tries to treat the column value as a file path first (for backwards compatibility). If the value contains spaces or other URI-illegal characters and Liquibase runs from a JAR, the `handlePathForJarfile()` method logs a `WARN`-level message with a full `URISyntaxException` stack trace. This is misleading and noisy because the fallback behavior works correctly — the value is inserted as a plain string.

Additionally, values containing `{`, `}`, or other URI-illegal characters cause an uncaught `IllegalArgumentException` in `NotFoundResource`'s constructor.

This PR:
- Lowers the log level in `AbstractResource.handlePathForJarfile()` from `warning()` to `fine()` (DEBUG), since the catch block already handles this gracefully
- Makes `NotFoundResource` URI construction robust by catching `IllegalArgumentException` from `URI.create()` and falling back to percent-encoding illegal characters
- Adds unit tests for both changes

## Things to be aware of

- The log level change from `warning()` to `fine()` means the `URISyntaxException` message will only appear at DEBUG level. This is intentional — the fallback path works correctly and the warning was confusing users into thinking something was wrong.
- The `createSafeUri()` method in `NotFoundResource` preserves the original behavior for valid paths and only applies percent-encoding when `URI.create()` fails.

## Things to worry about

- None. Both changes are defensive — they handle error cases that were already being caught, just more gracefully.

## Additional Context

See issue #7386 for the original report and stack trace examples.

### DIFF
--- a/liquibase-standard/src/main/java/liquibase/resource/ResourceAccessor.java
+++ b/liquibase-standard/src/main/java/liquibase/resource/ResourceAccessor.java
@@ -302,8 +302,27 @@ public interface ResourceAccessor extends AutoCloseable {
         private final ResourceAccessor resourceAccessor;
 
         public NotFoundResource(String path, ResourceAccessor resourceAccessor) {
-            super(path, URI.create("resourceaccessor:"+path.replace(" ", "%20").replace('\\', '/')));
+            super(path, createSafeUri(path));
             this.resourceAccessor = resourceAccessor;
+        }
+
+        private static URI createSafeUri(String path) {
+            String sanitized = path.replace(" ", "%20").replace('\\', '/');
+            try {
+                return URI.create("resourceaccessor:" + sanitized);
+            } catch (IllegalArgumentException e) {
+                // Percent-encode any remaining URI-illegal characters (e.g. {, })
+                StringBuilder sb = new StringBuilder();
+                for (char c : sanitized.toCharArray()) {
+                    if ((c >= 'a' && c <= 'z') || (c >= 'A' && c <= 'Z') ||
+                            (c >= '0' && c <= '9') || "/-._~:@!$&'()*+,;=%".indexOf(c) >= 0) {
+                        sb.append(c);
+                    } else {
+                        sb.append(String.format("%%%02X", (int) c));
+                    }
+                }
+                return URI.create("resourceaccessor:" + sb);
+            }
         }
 
         @Override
--- a/liquibase-standard/src/main/java/liquibase/resource/ResourceAccessor.java
+++ b/liquibase-standard/src/main/java/liquibase/resource/ResourceAccessor.java
@@ -12,6 +12,7 @@ import java.io.IOException;
 import java.io.InputStream;
 import java.io.OutputStream;
 import java.net.URI;
+import java.net.URISyntaxException;
 import java.util.*;
 
 /**
@@ -307,21 +308,10 @@ public interface ResourceAccessor extends AutoCloseable {
         }
 
         private static URI createSafeUri(String path) {
-            String sanitized = path.replace(" ", "%20").replace('\\', '/');
             try {
-                return URI.create("resourceaccessor:" + sanitized);
-            } catch (IllegalArgumentException e) {
-                // Percent-encode any remaining URI-illegal characters (e.g. {, })
-                StringBuilder sb = new StringBuilder();
-                for (char c : sanitized.toCharArray()) {
-                    if ((c >= 'a' && c <= 'z') || (c >= 'A' && c <= 'Z') ||
-                            (c >= '0' && c <= '9') || "/-._~:@!$&'()*+,;=%".indexOf(c) >= 0) {
-                        sb.append(c);
-                    } else {
-                        sb.append(String.format("%%%02X", (int) c));
-                    }
-                }
-                return URI.create("resourceaccessor:" + sb);
+                return new URI("resourceaccessor", path.replace('\\', '/'), null);
+            } catch (URISyntaxException e) {
+                return URI.create("resourceaccessor:unknown");
             }
         }
 
--- a/liquibase-standard/src/main/java/liquibase/resource/ResourceAccessor.java
+++ b/liquibase-standard/src/main/java/liquibase/resource/ResourceAccessor.java
@@ -309,6 +309,9 @@ public interface ResourceAccessor extends AutoCloseable {
 
         private static URI createSafeUri(String path) {
             try {
+                // The 3-arg URI constructor percent-encodes any illegal characters in the
+                // scheme-specific part, so it does not throw for the inputs seen here. The
+                // catch below is purely defensive and is not expected to be reached.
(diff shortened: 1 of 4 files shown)


liquibase-standard/src/main/java/liquibase/resource/ResourceAccessor.java
```diff
@@ -302,10 +303,18 @@ public class NotFoundResource extends AbstractResource {
...
-            super(path, URI.create("resourceaccessor:"+path.replace(" ", "%20").replace('\\', '/')));
+            super(path, createSafeUri(path));
             this.resourceAccessor = resourceAccessor;
         }
 
+        private static URI createSafeUri(String path) {
+            try {
+                return new URI("resourceaccessor", path.replace('\\', '/'), null);
+            } catch (URISyntaxException e) {
+                return URI.create("resourceaccessor:unknown");
```
filipelautert, line 317: Heads up: this fallback is effectively unreachable. I ran the 3-arg `URI(scheme, ssp, fragment)` constructor against every test input (spaces, `{}`, `[]`, bare `%`, non-ASCII, backslashes) — it quotes all illegal characters and never throws for this usage, so `resourceaccessor:unknown` won't trigger in practice. Fine to keep as defensive code; just noting it'll never fire (and if it somehow did, it silently drops the path). No change needed.

liquibase-standard/src/test/groovy/liquibase/resource/ResourceAccessorTest.groovy
```diff
@@ -101,6 +101,29 @@ class ResourceAccessorTest extends Specification {
...
+        where:
+        path                                  | expectedPath                          | description
+        "value with spaces"                   | "value with spaces"                   | "spaces"
+        "path/with{braces}"                   | "path/with{braces}"                   | "curly braces"
+        "some [bracketed] path"               | "some [bracketed] path"               | "square brackets"
+        "path with {braces} and spaces"       | "path with {braces} and spaces"       | "braces and spaces"
+        "simple/path.txt"                     | "simple/path.txt"                     | "simple path (no special chars)"
+        "path\\with\\backslashes"             | "path/with/backslashes"               | "backslashes"
+        "100% cotton"                         | "100% cotton"                         | "bare percent sign"
+        "path/with%20encoded"                 | "path/with%20encoded"                 | "already encoded percent"
```
filipelautert: Minor mislabel: this row isn't actually preserved as-is in the URI. I verified the 3-arg constructor double-encodes a bare `%`, so `path/with%20encoded` becomes raw URI `resourceaccessor:path/with%2520encoded`. The test passes only because it asserts `getPath()` (stored separately, correct) — not the URI string. Harmless here since a `NotFoundResource` URI is never dereferenced, but "already encoded percent" implies a guarantee that doesn't hold. Consider renaming (e.g. "literal percent sequence") or adding a `getUri()` assertion that documents the real behavior. Same double-encoding CodeRabbit flagged.

filipelautert (approved):
Hey, reviewed the changes. Clean, well-tested fix — the WARN→FINE change directly addresses #7386, and the `NotFoundResource` hardening is a sensible bonus. Verified the URI behavior by running the 3-arg constructor against every test input. Two low-priority notes left inline, nothing blocking. Tests are solid and the `getPath()` round-trip assertions lock down the right contract. Nice work.

jstastny:
@filipelautert -- thanks for the review and the approval. I just added a2afd5b55 which addresses the two minor issues.

rberezen (commented):
Tested locally with `loadUpdateData` and TEXT columns containing edge-case values (spaces, `{braces}`, `100% cotton`, `café résumé`). All values insert correctly, no WARN or stack trace produced, update succeeds.

- 25/25 unit tests pass
- Code change is clean: `warning()` → `fine()` and 3-arg URI constructor for safe percent-encoding

LGTM.

rberezen (approved):

PR #7594, opened 2026-03-05
DAT-22398: Update condition for release-docker step

sayaliM0412:
This pull request updates the release workflow to control the execution of the Docker image release step based on a `dry_run` input parameter, instead of always skipping it.

Workflow logic improvement:

* [`.github/workflows/release-published.yml`](diffhunk://#diff-f7ca0660afb7c5d6e7220dee8a6db81db251b3c98818f4d76a3e4441b85c7a6dL444-R444): Changed the condition for the `release-docker` job to run only when `inputs.dry_run` is false, replacing the previous hardcoded skip (`if: false`). This allows for more flexible control during releases.

### DIFF
--- a/.github/workflows/release-published.yml
+++ b/.github/workflows/release-published.yml
@@ -441,7 +441,7 @@ jobs:
           remoteDir: "/xml/ns/dbchangelog/"
 
   release-docker:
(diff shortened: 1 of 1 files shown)


sayaliM0412:
workflow not in use anymore. Closing the PR

PR #7633, opened 2026-04-01
[DAT-22458] Fix SHA-SNAPSHOT package cleanup — delete step was always skipped

sayaliM0412:
## Problem

The `delete-sha-package` job in `cleanup-branch-builds.yml` has never actually deleted any SHA-SNAPSHOT packages.

**Root cause:** The job runs on branch `delete` events and tries to compare the deleted branch against `master`/`release` using the GitHub compare API. But by the time the workflow runs, the branch is already gone — the API returns 404, the commit list is empty, and the delete step is skipped.

Confirmed by checking all 10 recent workflow runs — the "Delete package versions" step was **skipped every time**.

This caused 53 stale SHA-SNAPSHOT packages to accumulate (~2 GB), which were manually deleted as part of DAT-22458.

## Fix

Replaced the broken branch-comparison logic with orphan detection:

1. List all package versions matching the SHA-SNAPSHOT pattern (`^[0-9a-f]{40}-SNAPSHOT$`)
2. For each SHA, check if the commit is reachable from `master` or `release`
3. If not reachable, check if any active branch points to it
4. Only delete versions where the commit is fully orphaned

## What it does NOT delete

- `master-SNAPSHOT` — does not match the 40-char hex SHA pattern
- `4.30.0-SNAPSHOT` or any semver — does not match the pattern
- `DAT-12345-SNAPSHOT` — does not match the pattern
- Any SHA-SNAPSHOT where the commit is still reachable from `master`/`release` or is a branch head

Screenshot 2026-04-01 at 2 34 09 PM

## Test plan

- [ ] Trigger the workflow manually via `workflow_dispatch` and verify orphan detection logs
- [ ] Confirm orphaned SHA versions are deleted
- [ ] Confirm active SHA versions (on existing branches) are preserved
- [ ] Verify DAT-SNAPSHOT cleanup (first job) still works as before — no changes to that job

### DIFF
--- a/.github/workflows/cleanup-branch-builds.yml
+++ b/.github/workflows/cleanup-branch-builds.yml
@@ -122,7 +122,11 @@ jobs:
         # Cache all branch HEAD SHAs once to avoid per-version API calls
         # and handle pagination (repo has ~190 branches, default page is 30)
         echo "Fetching all branch heads..."
-        BRANCH_HEADS=$(gh api --paginate "repos/$REPO/branches?per_page=100" --jq '.[].commit.sha' 2>/dev/null || echo "")
+        BRANCH_HEADS=$(gh api --paginate "repos/$REPO/branches?per_page=100" --jq '.[].commit.sha')
+        if [ $? -ne 0 ] || [ -z "$BRANCH_HEADS" ]; then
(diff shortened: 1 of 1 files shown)


sayaliM0412:
Recreating with correct branch name DAT-22701

PR #7639, opened 2026-04-06
DAT-22653: update GHA workflows and functionality

sayaliM0412:
1. url-check.yml — lychee-action replacement                                                                                                          
                                                                                                                                                        
  - SHA-pinned correctly                                                                                                                                
  - args: --verbose --timeout 5 --max-retries 2 url-test.md — correctly limits scan to just url-test.md                                                 
  - fail: true — matches original force_pass: false behavior                                                                                            
  - No extra permissions needed (public repo)                                                                                                           
                  
  2. create-release.yml — download-artifact                                                                                                             
                  
  - SHA-pinned correctly
  - run-id comes from workflow input runId — correct
  - Permissions: actions: write already declared — sufficient                                                                                           
  - Upload source is upload-artifact@v7 — compatible with download-artifact@v4                                                                          
                                                                                                                                                        
  3. nightly-release.yml — download-artifact                                                                                                            
   
  - SHA-pinned correctly                                                                                                                                
  - run-id comes from get-run-id step — correct
  - Permissions: actions: read already declared — sufficient
                                                                                                                                                        
  4. installer-build-check.yml — download-artifact + run ID lookup
                                                                                                                                                        
  - SHA-pinned correctly
  - Added actions: read permission — needed for both gh api and artifact download
  - gh api call properly finds latest successful run-tests.yml run on the branch                                                                        
  - Error handling if no successful run found                                                                                                           
                                                                                                                                                        
  5. release-deploy-xsd.yml — lftp replacement                                                                                                          
   
  - Comment explains reasoning clearly                                                                                                                  
  - mirror --reverse --overwrite matches original `forceUpload: true` behaviour
  - `compress: true` on the XSD upload (enables SSH compression for faster transfer) like before


.github/workflows/release-deploy-xsd.yml
```diff
@@ -82,26 +82,31 @@ jobs:
...
-          localDir: "index-file/"
-          remoteDir: "/xml/ns/dbchangelog/"
+        env:
+          SSHPASS: ${{ env.WPENGINE_SFTP_PASSWORD }}
+          SFTP_USER: ${{ env.WPENGINE_SFTP_USER }}
+          SFTP_HOST: ${{ env.WPENGINE_SFTP_HOST }}
+          SFTP_PORT: ${{ env.WPENGINE_SFTP_PORT }}
+        run: |
+          sshpass -e sftp -P "$SFTP_PORT" -oStrictHostKeyChecking=accept-new "$SFTP_USER@$SFTP_HOST" <<'SFTP_EOF'
+          put index-file/index.htm /xml/ns/dbchangelog/index.htm
```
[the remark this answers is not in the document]
sayaliM0412: Fixed in commit 226666820 — changed to `put index-file/index.htm /xml/ns/dbchangelog/index.htm` to avoid uploading the `sed -ie` backup file.

.github/workflows/archive/url-check.yml
```diff
@@ -14,20 +14,10 @@ jobs:
...
-          # Choose whether to print a more verbose end summary with files and broken URLs.
-          verbose: true
-          # The timeout seconds to provide to requests, defaults to 5 seconds
-          timeout: 5
-          # How many times to retry a failed request (each is logged, defaults to 1)
-          retry_count: 2
-          # choose if the force pass or not
-          force_pass : false
+          args: --verbose --timeout 5 --max-retries 2 .github/workflows/url-test.md
+          fail: true
```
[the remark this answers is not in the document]
sayaliM0412: Not applicable — this file is archived in `.github/workflows/archive/` and GitHub Actions won't run workflows from subdirectories. This workflow is effectively disabled.

.github/workflows/release-publish-github-packages.yml
```diff
@@ -67,9 +67,9 @@ jobs:
           server-id: liquibase
 
       - name: Set up Maven
-        uses: stCarolas/setup-maven@d6af6abeda15e98926a57b5aa970a96bb37f97d1 # v5
-        with:
-          maven-version: ${{ env.MAVEN_VERSION }}
+        run: |
+          curl -fsSL "https://archive.apache.org/dist/maven/maven-3/${{ env.MAVEN_VERSION }}/binaries/apache-maven-${{ env.MAVEN_VERSION }}-bin.tar.gz" | tar xz -C /opt
+          echo "/opt/apache-maven-${{ env.MAVEN_VERSION }}/bin" >> $GITHUB_PATH
```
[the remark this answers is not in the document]
sayaliM0412: Acknowledged — the download uses HTTPS from archive.apache.org which provides transport-level integrity. Adding checksum verification is a nice-to-have but low risk for this PR.

.github/workflows/fossa.yml
```diff
@@ -32,31 +32,35 @@ jobs:
...
+                  <snapshots><enabled>true</enabled><updatePolicy>always</updatePolicy></snapshots>
+                </repository>
+              </repositories>
+            </profile>
+          </profiles>
+          <activeProfiles>
+            <activeProfile>liquibase</activeProfile>
+          </activeProfiles>
+        </settings>
+        EOF
```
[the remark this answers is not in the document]
sayaliM0412: Fixed in commit cb4a41e2d — added `GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}` to the env block for Maven settings generation in fossa.yml, run-tests.yml, and weekly-integration-tests.yml.

.github/workflows/release-publish-assets-s3.yml
```diff
@@ -38,12 +38,11 @@ jobs:
...
-        with:
-          repository: "liquibase/liquibase"
-          latest: true
-          fileName: "*"
-          out-file-path: "./${{ inputs.version }}-released-assets"
+        env:
+          GH_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+        run: |
+          mkdir -p "./${{ inputs.version }}-released-assets"
+          gh release download --repo liquibase/liquibase --pattern "*" --dir "./${{ inputs.version }}-released-assets"
```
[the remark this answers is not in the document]
sayaliM0412: Fixed in commit cb4a41e2d — added explicit `v${{ inputs.version }}` tag to `gh release download` in release-publish-assets-s3.yml.

.github/workflows/release-deploy-xsd.yml
```diff
@@ -82,26 +82,31 @@ jobs:
...
-          remoteDir: "/xml/ns/dbchangelog/"
+        env:
+          SFTP_PASS: ${{ env.WPENGINE_SFTP_PASSWORD }}
+          SFTP_USER: ${{ env.WPENGINE_SFTP_USER }}
+          SFTP_HOST: ${{ env.WPENGINE_SFTP_HOST }}
+          SFTP_PORT: ${{ env.WPENGINE_SFTP_PORT }}
+        run: |
+          sshpass -p "$SFTP_PASS" sftp -P "$SFTP_PORT" -oCompression=yes -oStrictHostKeyChecking=no "$SFTP_USER@$SFTP_HOST" <<'SFTP_EOF'
+          put -r liquibase-core-repo/liquibase-standard/src/main/resources/www.liquibase.org/xml/ns/dbchangelog/* /xml/ns/dbchangelog/
+          bye
```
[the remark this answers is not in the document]
sayaliM0412: Fixed in commit 19e3b02da — switched to `sshpass -e` (reads from SSHPASS env var) and `StrictHostKeyChecking=accept-new`.

.github/workflows/release-deploy-xsd.yml
```diff
@@ -82,26 +82,31 @@ jobs:
...
-          localDir: "index-file/"
-          remoteDir: "/xml/ns/dbchangelog/"
+        env:
+          SFTP_PASS: ${{ env.WPENGINE_SFTP_PASSWORD }}
+          SFTP_USER: ${{ env.WPENGINE_SFTP_USER }}
+          SFTP_HOST: ${{ env.WPENGINE_SFTP_HOST }}
+          SFTP_PORT: ${{ env.WPENGINE_SFTP_PORT }}
+        run: |
+          sshpass -p "$SFTP_PASS" sftp -P "$SFTP_PORT" -oStrictHostKeyChecking=no "$SFTP_USER@$SFTP_HOST" <<'SFTP_EOF'
+          put -r index-file/* /xml/ns/dbchangelog/
```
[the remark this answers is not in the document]
sayaliM0412: Fixed in commit 226666820 — changed to `put index-file/index.htm /xml/ns/dbchangelog/index.htm`.

.github/workflows/url-check.yml
```diff
@@ -14,20 +14,10 @@ jobs:
...
-          print_all: false
-          # Choose whether to print a more verbose end summary with files and broken URLs.
-          verbose: true
-          # The timeout seconds to provide to requests, defaults to 5 seconds
-          timeout: 5
-          # How many times to retry a failed request (each is logged, defaults to 1)
-          retry_count: 2
-          # choose if the force pass or not
-          force_pass : false
+          args: --verbose --timeout 5 --max-retries 2 url-test.md
```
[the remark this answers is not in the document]
sayaliM0412: Fixed and file has been archived to `.github/workflows/archive/`.

sayaliM0412:
Both items have been addressed:

1. **url-check.yml path mismatch** — This file has been moved to `.github/workflows/archive/`. GitHub Actions only runs workflows directly under `.github/workflows/`, not subdirectories. The workflow is effectively disabled, so the path inside it is irrelevant.

2. **Maven checksum verification** — Fixed in commit 3217cfcd0. Maven archive is now downloaded to `$RUNNER_TEMP`, verified against Apache's published SHA-512 checksum via `sha512sum --check`, then extracted.

.github/workflows/fossa.yml
```diff
@@ -32,31 +32,37 @@ jobs:
...
-          [
-            {
-              "id": "liquibase",
-              "username": "liquibot",
-              "password": "${{ secrets.GITHUB_TOKEN }}"
-            }
-          ]
+    - name: Generate Maven settings.xml
+      env:
+        GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
```
jandroav: **CRITICAL: `${env.GITHUB_TOKEN}` in settings.xml won't resolve when Maven runs**

This step-level `env:` sets `GITHUB_TOKEN` only for the `cat` command that generates settings.xml. The settings.xml contains `<password>${env.GITHUB_TOKEN}</password>` which Maven resolves at **runtime** by looking up the `GITHUB_TOKEN` env var.

But when Maven runs in the **next** step (`mvn clean install -DskipTests`), `GITHUB_TOKEN` is NOT an env var — step-level `env:` only applies to the step it's defined on.

From the GitHub Actions docs:
> *"Environment variables defined in a step override job and workflow variables with the same name **while the step executes**."*

`secrets.GITHUB_TOKEN` is a secret, not an automatically-available env var. It must be explicitly mapped.

**Fix:** persist it across steps:
```yaml
- name: Generate Maven settings.xml
  run: |
    echo "GITHUB_TOKEN=${{ secrets.GITHUB_TOKEN }}" >> $GITHUB_ENV
    cat > ~/.m2/settings.xml << 'EOF'
    ...
    EOF
```

Or set it at the **job level** `env:` block.

Same issue in `run-tests.yml` and `weekly-integration-tests.yml`. Note: `build.yml` and `create-release.yml` are fine — they use `${env.LIQUIBOT_PAT_GPM_ACCESS}` which is set via vault secrets to `$GITHUB_ENV` and persists across steps.

.github/workflows/archive/url-check.yml
```diff
@@ -14,20 +14,10 @@ jobs:
     steps:
       - uses: actions/checkout@de0fac2e4500dabe0009e67214ff5f5447ce83dd # v6.0.2
       - name: Check URLs
-        uses: urlstechie/urlchecker-action@b643b43e2ac605e1475331c7b67247d242b7dce4 # 0.0.34
+        uses: lycheeverse/lychee-action@8646ba30535128ac92d33dfc9133794bfdd9b411 # v2.8.0
```
jandroav: **Workflow silently disabled, not replaced**

Moving this file to `.github/workflows/archive/` effectively **disables the workflow** — GitHub Actions only runs workflows from `.github/workflows/` (flat, not recursive). The PR description says "url-check.yml — lychee-action replacement" without clarifying the workflow is being archived.

If the intent is to disable URL checking, that's fine — but the PR description should say so explicitly.

Also: the `args` on line 19 still references `.github/workflows/url-test.md`, but `url-test.md` was also moved to `.github/workflows/archive/url-test.md`. If the workflow is ever un-archived, the path would be wrong.

.github/workflows/release-deploy-javadocs.yml
```diff
@@ -38,12 +38,9 @@ jobs:
...
       - name: Download release javadocs
-        uses: robinraju/release-downloader@daf26c55d821e836577a15f77d86ddc078948b05 # v1.12
-        with:
-          repository: "liquibase/liquibase"
-          tag: "${{ inputs.tag }}"
-          fileName: "liquibase-additional*.zip"
-          out-file-path: "."
+        env:
+          GH_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+        run: gh release download "${{ inputs.tag }}" --repo liquibase/liquibase --pattern "liquibase-additional*.zip" --dir "."
```
jandroav: **`${{ inputs.tag }}` used directly in `run:` — script injection vector**

`inputs.tag` is user input from `workflow_dispatch`. While only write-access users can trigger this, if the value contains shell metacharacters it would be interpreted. For a supply-chain-security-focused PR, defense in depth says pass via `env:`:

```yaml
env:
  GH_TOKEN: ${{ secrets.GITHUB_TOKEN }}
  TAG: ${{ inputs.tag }}
run: gh release download "$TAG" --repo liquibase/liquibase --pattern "liquibase-additional*.zip" --dir "."
```

Same pattern applies to `release-deploy-maven.yml` (line 43) and `release-publish-assets-s3.yml` (line 44) which also interpolate `${{ inputs.tag }}` / `${{ inputs.version }}` directly.

.github/workflows/release-deploy-xsd.yml
```diff
@@ -82,26 +82,31 @@ jobs:
...
-          forceUpload: true
-          localDir: "liquibase-core-repo/liquibase-standard/src/main/resources/www.liquibase.org/xml/ns/dbchangelog/"
-          remoteDir: "/xml/ns/dbchangelog/"
+        env:
+          SSHPASS: ${{ env.WPENGINE_SFTP_PASSWORD }}
+          SFTP_USER: ${{ env.WPENGINE_SFTP_USER }}
+          SFTP_HOST: ${{ env.WPENGINE_SFTP_HOST }}
+          SFTP_PORT: ${{ env.WPENGINE_SFTP_PORT }}
+        run: |
+          sshpass -e sftp -P "$SFTP_PORT" -oCompression=yes -oStrictHostKeyChecking=accept-new "$SFTP_USER@$SFTP_HOST" <<'SFTP_EOF'
```
jandroav: **`StrictHostKeyChecking=accept-new` on ephemeral runners is effectively `no`**

`accept-new` means: trust on first use, reject if key changes later. But GitHub Actions runners are ephemeral — each run starts fresh with no `known_hosts`. So every connection is a "first use" and the key is always accepted. No MITM protection on any run.

Consider pinning the WPEngine host key (fetched from vault or hardcoded):
```yaml
- name: Configure known hosts
  run: echo "$WPENGINE_HOST_KEY" >> ~/.ssh/known_hosts
```
Then use `-oStrictHostKeyChecking=yes` instead.

That said, the `env:` handling here is done well — all credentials passed via env vars, not `${{ }}` interpolation in the shell. Better pattern than the `lftp` approach in the liquibase-pro PR.

.github/workflows/dry-run-release.yml
```diff
@@ -257,13 +257,17 @@ jobs:
...
-          SLACK_WEBHOOK: ${{ env.DRY_RUN_RELEASE_SLACK_WEBHOOK }}
-          SLACK_ICON_EMOJI: ":robot_face:"
-          SLACK_FOOTER: "${{ github.repository }}"
-          SLACK_LINK_NAMES: true
+        uses: slackapi/slack-github-action@af78098f536edbc4de71162a307590698245be95 # v3.0.1
+        with:
+          method: chat.postMessage
+          token: ${{ env.DRY_RUN_SLACK_BOT_TOKEN }}
+          payload: |
+            channel: "dryrun-releases"
```
jandroav: **Channel name instead of channel ID — fragile**

The official `slackapi/slack-github-action` docs use `channel: ${{ secrets.SLACK_CHANNEL_ID }}` (channel ID). Using a channel name (`"dryrun-releases"`) works but is fragile — if the channel is ever renamed, this notification silently fails with no error in the workflow.

Consider storing the channel ID in vault alongside `DRY_RUN_SLACK_BOT_TOKEN`.

.github/workflows/release-publish-github-packages.yml
```diff
@@ -67,9 +67,13 @@ jobs:
           server-id: liquibase
 
       - name: Set up Maven
-        uses: stCarolas/setup-maven@d6af6abeda15e98926a57b5aa970a96bb37f97d1 # v5
-        with:
-          maven-version: ${{ env.MAVEN_VERSION }}
+        # Download Maven with SHA-512 checksum verification for supply chain integrity
+        run: |
+          curl -fsSL "https://archive.apache.org/dist/maven/maven-3/${{ env.MAVEN_VERSION }}/binaries/apache-maven-${{ env.MAVEN_VERSION }}-bin.tar.gz" -o "$RUNNER_TEMP/maven.tar.gz"
+          curl -fsSL "https://archive.apache.org/dist/maven/maven-3/${{ env.MAVEN_VERSION }}/binaries/apache-maven-${{ env.MAVEN_VERSION }}-bin.tar.gz.sha512" -o "$RUNNER_TEMP/maven.tar.gz.sha512"
```
jandroav: **SHA-512 checksum from same server provides limited supply chain protection**

Both the binary and its checksum are downloaded from `archive.apache.org`. If that server is compromised, an attacker can replace both files simultaneously.

For stronger supply chain integrity, hardcode the expected SHA-512 for the specific Maven version in the workflow:
```yaml
env:
  MAVEN_SHA512: "abc123..."  # from https://downloads.apache.org/maven/
run: |
  echo "$MAVEN_SHA512  $RUNNER_TEMP/maven.tar.gz" | sha512sum --check
```

Still better than no verification, and the comment on line 70 accurately describes the intent.

jandroav (commented):
**Maven settings inlined in 5 workflows — consider shared templates**

The `liquibase-pro` PR (DAT-22650) took the shared template approach: two XML files in `.github/maven/`, referenced by all workflows via `cp`. This PR inlines the full settings.xml heredoc in each of 5 workflows (`build.yml`, `create-release.yml`, `fossa.yml`, `run-tests.yml`, `weekly-integration-tests.yml`). Any future change (new repo, new server, credential rotation) requires editing all 5 files in lockstep. Consider the shared template approach for consistency across repos.

.github/workflows/archive/url-check.yml
sayaliM0412: removing this workflow. Confirmed from the team.

.github/workflows/release-deploy-xsd.yml
sayaliM0412: Implemented — added a `Configure WPEngine known hosts` step that writes the pinned host key from vault (`WPENGINE_SFTP_HOST_KEY` in `/vault/liquibase`) to `~/.ssh/known_hosts`, and changed `StrictHostKeyChecking` from `accept-new` to `yes`.

.github/workflows/dry-run-release.yml
sayaliM0412: secret added in vault under DRYRUN_RELEASES_SLACK_ID

.github/workflows/release-publish-github-packages.yml
sayaliM0412: Implemented — SHA-512 checksum is now hardcoded as `MAVEN_SHA512` env var in the workflow. The checksum download from `archive.apache.org` has been removed. Verification is now done against the pinned hash, so a compromised server can't supply a matching checksum.

.github/workflows/release-deploy-javadocs.yml
sayaliM0412: done

.github/workflows/fossa.yml
sayaliM0412: Fixed — removed the step-level `env:` block and instead persist `GITHUB_TOKEN` to `$GITHUB_ENV` via `echo "GITHUB_TOKEN=${{ secrets.GITHUB_TOKEN }}" >> $GITHUB_ENV`. This makes it available as an OS env var for all subsequent steps, so Maven can resolve `${env.GITHUB_TOKEN}` at runtime. Applied to fossa.yml, run-tests.yml, and weekly-integration-tests.yml.

sayaliM0412:
Re: [shared Maven templates suggestion](https://github.com/liquibase/liquibase/pull/7639#pullrequestreview-4066211901)

Implemented — replaced all 5 inline heredocs with shared XML templates in `.github/maven/`, matching the approach in the liquibase-pro PR (DAT-22650):

- `.github/maven/settings-snapshots.xml` — used by `build.yml`, `fossa.yml`, `run-tests.yml`, `weekly-integration-tests.yml`
- `.github/maven/settings-releases.xml` — used by `create-release.yml`

All workflows now use `cp .github/maven/settings-*.xml ~/.m2/settings.xml` and set `MAVEN_GPM_PASSWORD` via `$GITHUB_ENV`. Future changes only require editing one XML file.

.github/workflows/archive/url-check.yml
sayaliM0412: Good point — updated the PR description to clarify that url-check.yml is intentionally archived/disabled, not just moved.

.github/workflows/release-deploy-javadocs.yml
sayaliM0412: Implemented — all three files now pass user inputs via `env:` block instead of direct `${{ }}` interpolation in `run:`. release-deploy-javadocs.yml and release-deploy-maven.yml use `TAG`, release-publish-assets-s3.yml uses `VERSION`.

.github/workflows/dry-run-release.yml
sayaliM0412: Implemented — channel name replaced with channel ID stored in vault as `DRYRUN_RELEASES_SLACK_ID` in `/vault/liquibase`. Now references `${{ env.DRYRUN_RELEASES_SLACK_ID }}` so channel renames won't break notifications.

.github/workflows/fossa.yml
sayaliM0412: Fixed — removed step-level `env:` and instead persist `GITHUB_TOKEN` to `$GITHUB_ENV` so Maven can resolve `${env.MAVEN_GPM_PASSWORD}` at runtime. Now using shared Maven templates with `MAVEN_GPM_PASSWORD` set via `$GITHUB_ENV`. Applied to fossa.yml, run-tests.yml, and weekly-integration-tests.yml.

.github/workflows/release-deploy-javadocs.yml
sayaliM0412: Implemented — all three files now pass user inputs via `env:` block instead of direct `${{ }}` interpolation in `run:`. release-deploy-javadocs.yml and release-deploy-maven.yml use `TAG`, release-publish-assets-s3.yml uses `VERSION`.

.github/workflows/release-deploy-xsd.yml
sayaliM0412: Implemented — added a `Configure WPEngine known hosts` step that writes the pinned host key from vault (`WPENGINE_SFTP_HOST_KEY` in `/vault/liquibase`) to `~/.ssh/known_hosts`, and changed `StrictHostKeyChecking` from `accept-new` to `yes`.

.github/workflows/dry-run-release.yml
sayaliM0412: Implemented — channel name replaced with channel ID stored in vault as `DRYRUN_RELEASES_SLACK_ID` in `/vault/liquibase`. Now references `${{ env.DRYRUN_RELEASES_SLACK_ID }}` so channel renames won't break notifications.

.github/workflows/release-publish-github-packages.yml
sayaliM0412: Implemented — SHA-512 checksum is now hardcoded as `MAVEN_SHA512` env var in the workflow. The checksum download from `archive.apache.org` has been removed. Verification is now done against the pinned hash.

jandroav (approved):

PR #7645, opened 2026-04-07
Fix primaryKeyExists precondition ignoring PK name on Oracle

### LINKED ISSUE #1731: primaryKeyExists pre condition with table name and primary key name is malfunctioning on Oracle
## Environment

**Liquibase Version**: 4.3.1

**Liquibase Integration & Version**: n/a

**Database Vendor & Version**: Oracle broken, MSSQL works

**Operating System Type & Version**: n/a

## Description

Given these change sets:
```xml
   <changeSet id="xxx-1" author="n/a">
         <not>
            <primaryKeyExists primaryKeyName="PK_ON_TABLE_X"/>
         </not>
      </preConditions>
      <stop/>
   </changeSet>

   <changeSet id="xxx-2" author="n/a">
         <primaryKeyExists primaryKeyName="NON_EXISTING_PK"/>
      </preConditions>
      <stop/>
   </changeSet>

   <changeSet id="xxx-3" author="n/a">
         <not>
            <primaryKeyExists tableName="TABLE_X_WITH_PK"/>
         </not>
      </preConditions>
      <stop/>
   </changeSet>

   <changeSet id="xxx-4" author="n/a">
         <primaryKeyExists tableName="TABLE_WITHOUT_PK"/>
      </preConditions>
      <stop/>
   </changeSet>

   <changeSet id="xxx-5" author="n/a">
         <primaryKeyExists tableName="NON_EXISTING_TABLE"/>
      </preConditions>
      <stop/>
   </changeSet>

   <changeSet id="xxx-6" author="n/a">
         <not>
            <primaryKeyExists tableName="TABLE_X_WITH_PK" primaryKeyName="PK_ON_TABLE_X"/>
         </not>
      </preConditions>
      <stop/>
   </changeSet>

   <changeSet id="xxx-7" author="n/a">
         <primaryKeyExists tableName="TABLE_X_WITH_PK" primaryKeyName="NON_EXISTING_PK"/>
      </preConditions>
      <stop/>
   </changeSet>

   <changeSet id="xxx-8" author="n/a">
         <primaryKeyExists tableName="TABLE_WITHOUT_PK" primaryKeyName="PK_ON_TABLE_X"/>
      </preConditions>
      <stop/>
   </changeSet>

   <changeSet id="xxx-9" author="n/a">
         <primaryKeyExists tableName="TABLE_WITHOUT_PK" primaryKeyName="NON_EXISTING_PK"/>
      </preConditions>
      <stop/>
   </changeSet>

   <changeSet id="xxx-10" author="n/a">
         <primaryKeyExists tableName="NON_EXISTING_TABLE" primaryKeyName="PK_ON_TABLE_X"/>
      </preConditions>
      <stop/>
   </changeSet>

   <changeSet id="xxx-11" author="n/a">
         <primaryKeyExists tableName="NON_EXISTING_TABLE" primaryKeyName="NON_EXISTING_PK"/>
      </preConditions>
      <stop/>
   </changeSet>

   <changeSet id="xxx-12" author="n/a">
         <primaryKeyExists tableName="TABLE_Y_WITH_PK_BUT_NOT_THE_ONE_IN_THE_OTHER_ATTRIBUTE" primaryKeyName="PK_ON_TABLE_X"/>
      </preConditions>
      <stop/>
   </changeSet>
```

I would expect all to be marked as ran an none to fail due to the `stop` change.
This works properly on MSSQL, but it fails on Oracle.
On Oracle `xxx-7` and `xxx-12` reache the stop change.
It seems on Oracle any PK satisfies the pre-condition despite a name being specified.

filipelautert:
Fixes #1731

## Summary
- `PrimaryKeyComparator.isSameObject()` was only comparing table names when both primary keys had a table set, completely ignoring the PK constraint name
- This caused `primaryKeyExists` preconditions on Oracle (and potentially other databases) to incorrectly report a PK as existing when the table had a different PK than the one specified
- The fix adds PK name comparison when both objects have non-null names, while preserving the wildcard behavior when one side has a null name

## Root Cause
In `SnapshotGeneratorFactory.createAndCheckSnapshot()`, when the first snapshot attempt returns null (correctly), a fallback path iterates all PKs in the schema and uses `PrimaryKeyComparator.isSameObject()` to find a match. The comparator was treating any PK on the same table as a match, regardless of the specified PK name.

## Test plan
- [x] Added `PrimaryKeyComparatorTest` with 9 test cases covering:
  - Same/different PK names without table names
  - Same/different PK names with same/different table names  
  - Null PK name wildcard behavior (matching any PK on a table)
- [ ] Verify against Oracle database with the reproduction scenario from the issue

### DIFF
--- a/liquibase-standard/src/main/java/liquibase/diff/compare/core/PrimaryKeyComparator.java
+++ b/liquibase-standard/src/main/java/liquibase/diff/compare/core/PrimaryKeyComparator.java
@@ -49,7 +49,14 @@ public class PrimaryKeyComparator implements DatabaseObjectComparator {
 
         if ((thisPrimaryKey.getTable() != null) && (thisPrimaryKey.getTable().getName() != null) && (otherPrimaryKey
             .getTable() != null) && (otherPrimaryKey.getTable().getName() != null)) {
-            return DatabaseObjectComparatorFactory.getInstance().isSameObject(thisPrimaryKey.getTable(), otherPrimaryKey.getTable(), chain.getSchemaComparisons(), accordingTo);
+            if (!DatabaseObjectComparatorFactory.getInstance().isSameObject(thisPrimaryKey.getTable(), otherPrimaryKey.getTable(), chain.getSchemaComparisons(), accordingTo)) {
+                return false;
+            }
+            // If both PKs have a name, they must match
+            if (thisPrimaryKey.getName() != null && otherPrimaryKey.getName() != null) {
+                return thisPrimaryKey.getName().equalsIgnoreCase(otherPrimaryKey.getName());
+            }
+            return true;
         } else {
             return StringUtil.trimToEmpty(thisPrimaryKey.getName()).equalsIgnoreCase(otherPrimaryKey.getName());
         }
--- a/liquibase-standard/src/main/java/liquibase/diff/compare/core/PrimaryKeyComparator.java
+++ b/liquibase-standard/src/main/java/liquibase/diff/compare/core/PrimaryKeyComparator.java
@@ -38,6 +38,13 @@ public class PrimaryKeyComparator implements DatabaseObjectComparator {
         }
     }
 
+    /**
+     * Two primary keys are the same object when they belong to the same table and their names match,
+     * where a null name matches any PK on the table. Differing names still match when both sides carry
+     * column definitions (snapshotted objects), since a table has a single PK and auto-generated names
+     * vary across environments. A side without columns is a name-based lookup example
+     * (e.g. the primaryKeyExists precondition) and must match by name.
+     */
     @Override
     public boolean isSameObject(DatabaseObject databaseObject1, DatabaseObject databaseObject2, Database accordingTo, DatabaseObjectComparatorChain chain) {
         if (!((databaseObject1 instanceof PrimaryKey) && (databaseObject2 instanceof PrimaryKey))) {
@@ -52,14 +59,18 @@ public class PrimaryKeyComparator implements DatabaseObjectComparator {
             if (!DatabaseObjectComparatorFactory.getInstance().isSameObject(thisPrimaryKey.getTable(), otherPrimaryKey.getTable(), chain.getSchemaComparisons(), accordingTo)) {
                 return false;
             }
-            // If both PKs have a name, they must match
-            if (thisPrimaryKey.getName() != null && otherPrimaryKey.getName() != null) {
-                return thisPrimaryKey.getName().equalsIgnoreCase(otherPrimaryKey.getName());
+            if ((thisPrimaryKey.getName() == null) || (otherPrimaryKey.getName() == null)
+                || thisPrimaryKey.getName().equalsIgnoreCase(otherPrimaryKey.getName())) {
+                return true;
             }
-            return true;
+            return hasColumns(thisPrimaryKey) && hasColumns(otherPrimaryKey);
         } else {
             return StringUtil.trimToEmpty(thisPrimaryKey.getName()).equalsIgnoreCase(otherPrimaryKey.getName());
         }
+    }
+
+    private static boolean hasColumns(PrimaryKey primaryKey) {
+        return (primaryKey.getColumns() != null) && !primaryKey.getColumns().isEmpty();
     }
 
 
--- a/liquibase-standard/src/test/groovy/liquibase/diff/compare/core/PrimaryKeyComparatorTest.groovy
+++ b/liquibase-standard/src/test/groovy/liquibase/diff/compare/core/PrimaryKeyComparatorTest.groovy
@@ -1,0 +1,63 @@
+package liquibase.diff.compare.core
+
+import liquibase.database.Database
+import liquibase.diff.compare.CompareControl
+import liquibase.diff.compare.DatabaseObjectComparatorChain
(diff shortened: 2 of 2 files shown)


v-petrovych:
@claude please double check potential issues below:
 1. Medium: Potential regression with database-generated PK names in non-hash paths

  If any code path calls isSameObject without hash pre-filtering (as the
  SnapshotGeneratorFactory.createAndCheckSnapshot() fallback does), PKs with auto-generated names (e.g., Oracle's
  SYS_C00123 vs SYS_C00456) on the same table would now be considered different. Previously they were considered "the
  same" (same table).

  This is unlikely to cause real issues because:
  - The precondition fallback path is looking for a specific PK name (the one the user specified)
  - Auto-generated names would only conflict in edge cases where one side is queried and the other is constructed

  But it's worth confirming that no other callers rely on the old "same table = same PK" behavior.

  2. Low: Missing reverse wildcard test case

  The test covers null | "TABLE_A" | "PK_FOO" | "TABLE_A" | true (null first, named second) but not the reverse:
  "PK_FOO" | "TABLE_A" | null | "TABLE_A". The code handles both symmetrically (checks both non-null before comparing
  names), so this is covered by logic but not by test. Adding it would strengthen confidence.

  3. Low: Oracle live verification not completed

  The test plan checklist shows:
  - [ ] Verify against Oracle database with the reproduction scenario from the issue
  This is unchecked. Given the bug was specifically reported on Oracle, live verification would be valuable before
  merge.

v-petrovych (approved):
LGTM!

rberezen (commented):
Tested on Oracle 23c with the scenario from #1731 — PK named `NON_EXISTING_PK` on table with actual PK `PK_ON_TABLE_X`. With the PR fix, the comparator correctly returns false (different names), and `<not><primaryKeyExists/></not>` evaluates correctly.

**Results:**
- 9/9 new unit tests pass, including the bug case ("Different PK name, same table — should NOT match")
- Sanity tests (matching PK name, table-only PK lookup, etc.) still work correctly
- CI green on all 10 database integration tests

Code change is minimal and surgical (4 added lines). LGTM.

rberezen (approved):

filipelautert:
Heads up @v-petrovych @rberezen — I pushed a follow-up commit (43de1e06f) that changes the comparator behavior you previously approved, after spotting a regression risk in the diff path.

**The problem with the previous approach:** the hash pre-filter does *not* protect `diff`/`diffChangeLog` from the new name check. `PrimaryKeyComparator.hash()` indexes a named PK under two buckets (name *and* table name), so `DatabaseObjectCollection.get()` still reaches `isSameObject()` for same-table PKs with different names. With an unconditional name mismatch → `false`, a `diffChangeLog` between two databases whose PK names are auto-generated (Oracle `SYS_C...`, SQL Server `PK__...`) would report every such PK as missing + unexpected, generating drop/add churn where before there was none. The existing `findDifferences()` even does `exclude.add("name")` — PK names were deliberately ignored in diff.

**The fix** mirrors `UniqueConstraintComparator`: a name mismatch only rules objects out when one side carries no column definitions (a name-based lookup example, which is exactly what the `primaryKeyExists` precondition builds — so the #1731 fix is preserved). When both sides have columns (real snapshotted objects), they keep matching by table, since a table has a single PK.

New test cases cover both directions: the two diff scenarios (auto-generated names, same/different columns) fail without this commit, and the precondition scenarios still pass. Full `liquibase-standard` suite: 6204 tests, 0 failures.

rberezen:
Re-reviewed after the follow-up commit `43de1e06f`. Good catch on the regression risk — I re-tested the whole thing end-to-end on Oracle 23c and the new approach is correct.

**Confirmed the regression was real** (on the intermediate commit `f6618c707` that we'd approved): a `diff-changelog` between two identical schemas whose PKs only differ in auto-generated name (`SYS_C008711` vs `SYS_C008712`) produced spurious churn:
```xml
<addPrimaryKey  constraintName="SYS_C008711" tableName="ORDERS" .../>
<dropPrimaryKey constraintName="SYS_C008712" tableName="ORDERS"/>
```

**Fixed on HEAD** — same diff now reports:
```
Missing Primary Key(s): NONE
Unexpected Primary Key(s): NONE
Changed Primary Key(s): NONE
```

**#1731 still preserved** — `primaryKeyExists` precondition on Oracle (table `ORDERS`, real PK `SYS_C008711`):

| precondition | result |
|---|---|
| `primaryKeyName="SYS_C008711"` (correct) | passes → `EXECUTED` |
| `primaryKeyName="WRONG_PK_NAME"` | fails → `MARK_RAN` ✅ (the #1731 fix) |
| no name (table only) | passes → `EXECUTED` |

**No over-suppression** — a genuine PK difference isn't masked: with `lbsrc.ORDERS` keyed on `(ID)` vs a target keyed on `(ID, AMOUNT)`, diff correctly reports `Changed Primary Key(s): … columns changed from '[ORDERS.ID]' to '[ORDERS.ID, ORDERS.AMOUNT]'`. The `hasColumns()` gate does exactly the right thing: name-only lookup examples (preconditions) are excluded on name mismatch, while real snapshotted PKs match by table so their column diffs still surface.

The approach correctly mirrors `UniqueConstraintComparator`, and 181 comparator/PK unit tests pass locally. Re-approving. Thanks @filipelautert for spotting this before it shipped.

PR #7659, opened 2026-04-14
fix: OracleDatabase.correctObjectName() no longer mangles column names starting with "int"

### LINKED ISSUE #7018: OracleDatabase.correctObjectName() converts column names that starts from int to "NUMBER(*, 0)"
### Search first

### Description

In the pull request https://github.com/liquibase/liquibase/pull/6094

The new implementation of `OracleDatabase.correctObjectName()` was added for Oracle database:
```
@Override
public String correctObjectName(String objectName, Class<? extends DatabaseObject> objectType) {
    // Internally Oracle converts int/integer to number, so when we need to extract and compare it we need number too
    if (objectType.equals(Column.class) && StringUtils.startsWithIgnoreCase(objectName, "int")) {
        return "NUMBER(*, 0)";
    }
    return super.correctObjectName(objectName, objectType);
}
```
This method automatically converts all column names that starts from int*, like "internalPhoneNumber" to "NUMBER(*, 0)" which brakes Metadata comparison and diff generation is incorrect.

Exact thing is `StringUtils.startsWithIgnoreCase(objectName, "int")` - looks like `objectName` is not only limited to column type name and might be also the column name which should not be corrected as type should.

I might be wrong, but this looks like bug to me. Could someone from the team help me to clarify that, please?

### Steps To Reproduce

1. We have `User` table with the column named `internalPhoneNumber` of type `varchar(256)`.

2. In our code we have custom `MetaDatabaseObjectFactory` implementation that uses statement:
```
for (config in (change as CreateTableChange).columns) {
    val column = Column()
    column.setName(database.correctObjectName(config.name, Column::class.java))
    ***
    val dataType = DataTypeFactory.getInstance().fromDescription(config.type, database)
    val databaseDataType = dataType.toDatabaseDataType(database)
    val typeName = database.correctObjectName(databaseDataType.type, Column::class.java)
    column.setType(DataType(typeName))
    ***
}
```
where `database` is `liquibase.database.core.OracleDatabase` instance.

This code worked fine with several our supported databases, but it doesn't work for Oracle anymore.

### Expected/Desired Behavior

We call `database.correctObjectName()` twice - for column name and for type name, and for Oracle for column name we expect it returns `internalPhoneNumber`, but actually receive `NUMBER(*, 0)` as value.

### Liquibase Version

4.29.2

### Database Vendor & Version

Oracle

### Liquibase Integration

Gradle, Spring

### Liquibase Extensions

_No response_

Ubuntu 22.04

### Additional Context

_No response_

filipelautert:
## Summary
- **Root cause**: PR #6094 introduced `startsWithIgnoreCase(objectName, "int")` in `OracleDatabase.correctObjectName()`, which matched column **names** like `internalPhoneNumber`, `internal_payload_id`, `integration_type` — not just the `int`/`integer` data types it was intended for — and incorrectly converted them to `NUMBER(*, 0)`.
- **Fix**: Replaced prefix matching with exact case-insensitive matching for `"int"` and `"integer"` only.
- **Tests**: Added 7 data-driven test cases covering both the type mapping (preserved) and column name handling (fixed).

Fixes #7018

## Test plan
- [x] Existing `OracleDatabaseSpec` tests pass (12 original cases)
- [x] New test cases verify `int`/`integer`/`INT`/`INTEGER` still map to `NUMBER(*, 0)`
- [x] New test cases verify `internalPhoneNumber`, `internal_payload_id`, `integration_type` are uppercased normally
- [x] `OracleDatabaseTest` (33 tests) passes with no regressions

### DIFF
--- a/liquibase-standard/src/main/java/liquibase/database/core/OracleDatabase.java
+++ b/liquibase-standard/src/main/java/liquibase/database/core/OracleDatabase.java
@@ -685,7 +685,8 @@ public class OracleDatabase extends AbstractJdbcDatabase {
     @Override
     public String correctObjectName(String objectName, Class<? extends DatabaseObject> objectType) {
         // Internally Oracle converts int/integer to number, so when we need to extract and compare it we need number too
-        if (objectType.equals(Column.class) && StringUtils.startsWithIgnoreCase(objectName, "int")) {
+        if (objectType.equals(Column.class)
+                && ("int".equalsIgnoreCase(objectName) || "integer".equalsIgnoreCase(objectName))) {
             return "NUMBER(*, 0)";
         }
         return super.correctObjectName(objectName, objectType);
--- a/liquibase-standard/src/test/groovy/liquibase/database/core/OracleDatabaseSpec.groovy
+++ b/liquibase-standard/src/test/groovy/liquibase/database/core/OracleDatabaseSpec.groovy
@@ -25,18 +25,27 @@ class OracleDatabaseSpec extends Specification {
         }
 
         where:
-        quotingStrategy           || objectName       | objectType || expectedCorrect  | expectedEscape
-        null                      || 'col1'           | Column     || 'COL1'           | 'col1'
-        null                      || 'COL1'           | Column     || 'COL1'           | 'COL1'
-        null                      || 'Col1'           | Column     || 'COL1'           | 'Col1'
-        null                      || 'col with space' | Column     || 'COL WITH SPACE' | '"col with space"'
-        QUOTE_ONLY_RESERVED_WORDS || 'col1'           | Column     || 'COL1'           | 'col1'
-        QUOTE_ONLY_RESERVED_WORDS || 'COL1'           | Column     || 'COL1'           | 'COL1'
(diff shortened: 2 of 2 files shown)


wwillard7800 (approved):

rberezen (commented):
Tested on Oracle 23c with the exact scenario from #7018.

**Before:** column `internal_payload_id` was mangled to `NUMBER(*, 0)` by `correctObjectName`, breaking `indexExists` precondition checks.
**After:** column names starting with `int` are correctly uppercased (`INTERNAL_PAYLOAD_ID`, `INTEGRATION_TYPE`), while the `int`/`integer` type→`NUMBER(*, 0)` mapping is preserved.

Verified end-to-end:
- `int` type still maps to `NUMBER`
- `internal_payload_id` is stored as `VARCHAR2` (not `NUMBER`)
- `indexExists` precondition correctly finds the index on the second run → `MARK_RAN` works

The test coverage nicely differentiates both cases (7 new test cases). LGTM.

rberezen (approved):

Issue #7544, opened 2026-01-29
Performance issue in liq 5.0.1


mhockett:
### Search first

- [x] I searched and no similar issues were found

### Description

I recently upgraded from 4.29.2 to 5.0.1
We are seeing increase in time spent processing changesets and even in some cases executing the sql,  in both oracle and postgres.
I noticed that stripComments is not working, unless the log is just printing out the whole changeset and not what is parsed. I also noticed more time ( up to 3 milliseconds more) between ChangelogParserFactory and the updateNullChecksums , specifically in the FormattedSqlChangelogParser. That may not sound like much but for large enterprise builds it adds up. Our nightly build went from 40 minutes to over 2 hours. I have had to revert back to 4.29.2.
 

### Steps To Reproduce

I created a demo in my liquibase java example repo 
https://github.com/mhockett/liquibase_java_example

There are instructions for running in the readme.  You may need to run the two scripts to set up the schema and initial databasechangelog table. 
My demo shows there is time difference between the 4.29.2 and 5.0.1 version, although my test is very small it can still show the time difference. 

Look at the logs oracleLog501.log and oracleLog4292.log and compare 
also at logs postgresLog501.log and postgresLog4292.log 

I created two changesets. One with strip comments false and the other true.  The 501 version does not strip the comments, or at least it does not in the log.  The other is the time difference which is reported at the end of the log.  

### Expected/Desired Behavior

same or near same execution time.

### Liquibase Version

5.0.1

### Database Vendor & Version

Postgre 17.6 and Oracle 19C

### Liquibase Integration

maven 

### Are you willing to submit a PR?

- [ ] I'm willing to submit a PR (Thank you!)

rberezen:
Hi @mhockett, thank you for the detailed report and the demo repo with logs.

We compared the 4.29.2 vs 5.0.1 logs and identified two contributors to the slowdown:

**1. Analytics module initialization overhead**
Both 5.0.1 logs show analytics configuration being loaded and then immediately disabled (`Analytics is disabled, because it was turned off by Liquibase`). This initialization takes ~0.4–1.0 seconds per run. In a small demo this is barely noticeable, but for large enterprise builds with many changesets, this overhead could compound significantly.

**2. Enhanced FormattedSqlChangeLogParser**
5.0.1 shows additional attribute parsing (`Matched attribute 'stripComments' = ...`) not present in 4.29.2. This adds per-changeset parser overhead.

[...]

PR #7674, opened 2026-04-23
perf: cache ranChangeSets lookup in SqlChangeLogParser.generateId

andrewcedgar:
### Problem
`SqlChangeLogParser.generateId` is called once per SQL changelog file and performs a defensive `ArrayList` copy + linear stream scan over `ChangeLogHistoryService.getRanChangeSets()`. With M SQL files and N rows in `DATABASECHANGELOG` this is O(M*N) work per update plus O(M*N) short-lived `RanChangeSet` references allocated.

On a real workload (~5,000 SQL changelogs, thousands of historical rows), parsing went from ~3 min on 4.31.1 to >15 min on 4.32.0+. Throughput collapses from ~625 file-events/sec to ~35 once the parser has warmed up (the cliff appears around 1,700 files parsed). Both 4.32.0 and 5.0.2 are affected — the code path is unchanged across those versions.

### Fix
Build the lookup once per `Database` as a `Map<changeLog, interimId>` and reuse it for every subsequent file. Total work becomes O(M+N) and the per-file cost returns to ~O(1).

Behavior is preserved: an index entry is added only when the `RanChangeSet`'s `id` equals the interim id the original filter would have computed, so `getOrDefault(path, "raw")` returns the same value as the previous `Optional.isPresent()` check.

### Tests
Added `SqlChangeLogParserTest` covering:
- default "raw" when nothing matches
- interim id returned when a matching ran changeset exists
- id mismatch is ignored (preserves original filter predicate)
- author mismatch is ignored
- `getRanChangeSets()` is invoked at most once across repeated `generateId` calls on the same `Database` (the performance invariant)

### Related
Likely the underlying cause of #7544.

### DIFF
--- a/liquibase-standard/src/main/java/liquibase/parser/core/sql/SqlChangeLogParser.java
+++ b/liquibase-standard/src/main/java/liquibase/parser/core/sql/SqlChangeLogParser.java
@@ -27,12 +27,19 @@ import liquibase.util.StringUtil;
 import org.apache.commons.lang3.StringUtils;
 
 import java.io.IOException;
-import java.util.ArrayList;
-import java.util.List;
-import java.util.Optional;
+import java.util.Collections;
+import java.util.HashMap;
+import java.util.Map;
+import java.util.concurrent.ConcurrentHashMap;
 
 @SuppressWarnings("java:S2583")
 public class SqlChangeLogParser implements ChangeLogParser {
+
+    // Per-Database index of physicalChangeLogLocation -> interim changeset id, built lazily on
+    // the first generateId call and reused thereafter. Avoids an O(N) scan of getRanChangeSets()
+    // per SQL changelog file, which on workloads with many SQL changelogs and large
+    // DATABASECHANGELOG histories was the dominant cost of changelog parsing.
+    private final Map<Database, Map<String, String>> interimIdIndexByDatabase = new ConcurrentHashMap<>();
 
     @Override
     public boolean supports(String changeLogFile, ResourceAccessor resourceAccessor) {
@@ -107,26 +114,30 @@ public class SqlChangeLogParser implements ChangeLogParser {
         if (database == null || isOldFormat(database)) {
             return "raw";
         }
+        Map<String, String> index = interimIdIndexByDatabase.computeIfAbsent(database, this::buildInterimIdIndex);
+        return index.getOrDefault(physicalChangeLogLocation, "raw");
+    }
 
-        List<RanChangeSet> ranChangeSets = new ArrayList<>();
+    private Map<String, String> buildInterimIdIndex(Database database) {
         try {
-            ranChangeSets = new ArrayList<>(
-               Scope.getCurrentScope().getSingleton(ChangeLogHistoryServiceFactory.class).getChangeLogService(database).getRanChangeSets());
+            Map<String, String> result = new HashMap<>();
+            for (RanChangeSet rc : Scope.getCurrentScope()
+                    .getSingleton(ChangeLogHistoryServiceFactory.class)
+                    .getChangeLogService(database).getRanChangeSets()) {
+                String id = rc.getId();
+                String changeLog = rc.getChangeLog();
+                if (id == null || changeLog == null || !"includeAll".equals(rc.getAuthor())) {
+                    continue;
+                }
+                String expectedInterimId = "raw_" + DatabaseChangeLog.normalizePath(changeLog).replace("/", "_");
+                if (id.equals(expectedInterimId)) {
+                    result.put(changeLog, id);
+                }
+            }
+            return result;
         } catch (Exception dbe) {
-            return "raw";
+            return Collections.emptyMap();
         }
-
-        String interimId = "raw_" + DatabaseChangeLog.normalizePath(physicalChangeLogLocation).replace("/", "_");
-        Optional<RanChangeSet> ranChangeSet =
-            ranChangeSets.stream().filter(rc -> {
-                return rc.getId().equals(interimId) &&
-                       rc.getAuthor().equals("includeAll") &&
-                       rc.getChangeLog().equals(physicalChangeLogLocation);
-            }).findFirst();
-        if (ranChangeSet.isPresent()) {
-            return interimId;
-        }
-        return "raw";
     }
 
     /**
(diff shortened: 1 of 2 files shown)


liquibase-standard/src/main/java/liquibase/parser/core/sql/SqlChangeLogParser.java
```diff
@@ -107,26 +114,30 @@ private String generateId(String physicalChangeLogLocation, Database database) {
         if (database == null || isOldFormat(database)) {
             return "raw";
         }
+        Map<String, String> index = interimIdIndexByDatabase.computeIfAbsent(database, this::buildInterimIdIndex);
```
filipelautert: **Cache never invalidates, so staleness across updates.** If the same `Database` instance is re-used to parse after an `update` has inserted new rows into `DATABASECHANGELOG`, the second parse sees the stale snapshot. Old code re-read `getRanChangeSets()` every call. Real scenarios: rollback-then-update on the same Database, integration tests that parse → update → parse, Spring apps triggering multiple updates. A returned `"raw"` for what should be `"raw_…"` means Liquibase will try to re-run an already-run changeset.

Suggested fixes, in order of preference:
- **Tie the cache to `Scope` instead of the parser singleton** — `Scope.getCurrentScope().getSingleton(...)` or a scope attribute, so the cache lives for the current command invocation only. Liquibase already creates a fresh Scope per command, so staleness disappears naturally.
- **Push caching into `ChangeLogHistoryService`** so writes invalidate it at the source. Higher payoff (any caller benefits) but more invasive.

liquibase-standard/src/test/groovy/liquibase/parser/core/sql/SqlChangeLogParserTest.groovy
```diff
@@ -0,0 +1,128 @@
...
+
+        then:
+        r1 == "raw"
+        r2 == "raw"
+        r3 == "raw"
+        1 * changeLogHistoryService.getRanChangeSets() >> []
+    }
+
+    private Object invokeGenerateId(SqlChangeLogParser parser, String path) {
+        Method m = SqlChangeLogParser.class.getDeclaredMethod("generateId", String.class, Database.class)
```
filipelautert: **Drop the reflection.** Flip `generateId` to package-private and call it directly. Reflection on a private method is brittle against rename and hides the contract the test is actually pinning. Must-fix.

filipelautert (changes requested):
Hey, reviewed the changes. Clean perf fix and behavior is preserved properly. Two things to address — staleness of the cache across updates, and dropping reflection from the test. Details inline.

Nice catch on the null-safety improvement in `buildInterimIdIndex`. Perf writeup in the PR description is excellent — numbers, call-site reasoning, and behavior-equivalence argument all in one place.

filipelautert (approved):
Hey, took another look — both blockers from the prior round are addressed cleanly, and the bonus transient-failure test is a nice touch.

What landed:

1. **Staleness fix** ✅ — `INTERIM_ID_INDEX_BY_SCOPE` keyed by `Scope` with `WeakHashMap`. The cache lives only as long as the current command's Scope is reachable, so a subsequent update→parse cycle starts fresh. This is exactly the shape I had in mind.
2. **Reflection dropped** ✅ — `generateId` flipped to package-private, test calls it directly. Test reads much cleaner now.
3. **Bonus**: the "retries the index build after a transient failure" test pins down the `computeIfAbsent` contract (lambda exceptions don't cache the entry). That's an invariant that's easy to break by accident later — good catch.

A couple of small observations, none blocking:

- **Subtle behavior relaxation in `buildInterimIdIndex`**: the old code would NPE on a `RanChangeSet` with a null `id`, `author`, or `changeLog` and bail out to `"raw"` for the entire call (via the `catch (Exception)`). The new code skips null/malformed rows individually but keeps processing the rest. In practice this is an improvement — one bad row no longer poisons every subsequent file in the same parse — but it is a slight loosening of the prior contract. Worth a line in the commit message or PR description if you want the audit trail.
- The `Map<Scope, Map<Database, Map<String, String>>>` works fine. If you ever want to drop a level of nesting later, `Scope` already supports per-scope attributes (`Scope.getCurrentScope().get(key, ...)` / `addAttribute`), but what's there is straightforward and self-contained — leave it.
- Concurrency on the outer map looks right: `Collections.synchronizedMap(new WeakHashMap<>())` plus the explicit `synchronized` block for the compound `computeIfAbsent`, and `ConcurrentHashMap` at the inner level. No issues.

Once CI's green, this is good to go.

rberezen (commented):
Benchmarked the fix on Oracle 23c and H2 with realistic workloads.

**Oracle (500 SQL files via `includeAll`):**
| Scenario | Master | PR | Improvement |
|---|---|---|---|
| Fresh run (DBCL grows 0→500) | 48.2s | 41.1s | 15% faster |
| Skip path (500 DBCL rows) | 29.7s | 27.2s | 8% faster |

**H2 file-based (1000 SQL files):**
| Scenario | Master | PR | Improvement |
|---|---|---|---|
| Fresh run | 34.0s | 19.0s | 44% faster |
| Skip path (1000 rows) | 11.7s | 7.3s | 38% faster |

The improvement scales as expected with O(M × N) — for the original reporter's 5000-file workload, the savings would be substantially larger. Note that `ChangeLogHistoryService.getRanChangeSets()` already caches the list internally, so the PR's win comes from eliminating M iterations of `new ArrayList<>(cached_list)` + linear stream scan (plus the temporary `RanChangeSet` references that pressure GC).

All 6 unit tests pass and CI integration tests are green. The fix is correct and beneficial. LGTM.

rberezen (approved):

rberezen (commented):
Adding code quality notes:

- **`WeakHashMap<Scope, ...>`** — clean way to scope the cache to a single command without manual eviction. Once the command's Scope is GC'd, the entry naturally drops, so no stale data across update→rollback→update cycles.
- **Transient-failure handling** — the `computeIfAbsent` lambda re-throws, so failed lookups aren't cached. Verified by the "retry after transient failure" test.
- **Performance invariant test** — `1 * changeLogHistoryService.getRanChangeSets()` across 3 calls is a nice way to lock in the optimization.
- **`generateId` made package-private** so tests can stop using reflection — good tradeoff.

cc @mhockett — this is likely the fix that addresses #7544.

PR #7696, opened 2026-05-07
Fix #7602: Set database filter when user-provided ChangeLogParameters is passed

### LINKED ISSUE #7602: DB2-specific property is being set within changelog even though PostgreSQL is being used
### Search first

### Description

When defining a property within a changelog, you can specify that it should only be applied when a specific DBMS is being used by using the *dbms* option. However, this mechanism will fail if all of the following conditions are met:

* Liquibase is directly invoked using the CommandScope class of liquibase-core-5.0.2.jar
* Additional properties are being set using the ChangeLogParameters class.
* PostgreSQL is being used as DBMS.

As a result, this will cause the wrong properties to be applied.

### Steps To Reproduce

1. Create a Java project with the following files:

Main.java:
```Java
package com.testproject.liquibasepropertybug;

import liquibase.changelog.ChangeLogParameters;
import liquibase.command.CommandScope;

import java.util.Arrays;
import java.util.List;

public class Main {

    public static final String URL = "jdbc:postgresql://localhost:5432/postgres?currentSchema=public";
    public static final String USERNAME = "postgres";
    public static final String PASSWORD = "postgres";

    public static final String CHANGELOG = "db.changelog.yaml";
    public static final String CHANGELOG_ARGUMENT_KEY = "changelogFile";
    public static final String URL_ARGUMENT_KEY = "url";
    public static final String USERNAME_ARGUMENT_KEY = "username";
    public static final String PASSWORD_ARGUMENT_KEY = "password";
    public static final String CHANGELOG_PARAMETERS_ARGUMENT_KEY = "changelogParameters";
    public static final String DESCRIPTION_PARAMETER_KEY = "description";
    public static final String DESCRIPTION_PARAMETER_VALUE = "An DBMS specialized in storing and retrieving data.";

    public static void main(String[] args) throws Exception {
        List<String> argsAsList = Arrays.asList(args);

        if (argsAsList.contains("update")) {
            runUpdate();
        } else if (argsAsList.contains("updateWithParameter")) {
            runUpdateWithParameter();
        }
    }

    //Property nameOfDbms will be resolved as 'Some other DBMS'.
    //This is what I expect when using PostgreSQL as my DBMS.
    public static void runUpdate() throws Exception {
        CommandScope command = new CommandScope("update");
        command.addArgumentValue(CHANGELOG_ARGUMENT_KEY, CHANGELOG);
        command.addArgumentValue(URL_ARGUMENT_KEY, URL);
        command.addArgumentValue(USERNAME_ARGUMENT_KEY, USERNAME);
        command.addArgumentValue(PASSWORD_ARGUMENT_KEY, PASSWORD);

        command.execute();
    }

    //Property nameOfDbms will be resolved as 'Db2 for IBM i'.
    //This is not what I would expect, since I am using PostgreSQL as DBMS.
    public static void runUpdateWithParameter() throws Exception {
        ChangeLogParameters parameters = new ChangeLogParameters();
        parameters.set(DESCRIPTION_PARAMETER_KEY, DESCRIPTION_PARAMETER_VALUE);

        CommandScope command = new CommandScope("update");
        command.addArgumentValue(CHANGELOG_ARGUMENT_KEY, CHANGELOG);
        command.addArgumentValue(URL_ARGUMENT_KEY, URL);
        command.addArgumentValue(USERNAME_ARGUMENT_KEY, USERNAME);
        command.addArgumentValue(PASSWORD_ARGUMENT_KEY, PASSWORD);
        command.addArgumentValue(CHANGELOG_PARAMETERS_ARGUMENT_KEY, parameters);

        command.execute();
    }

}
```

pom.xml:
```xml
<project xmlns="http://maven.apache.org/POM/4.0.0"
         xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 http://maven.apache.org/xsd/maven-4.0.0.xsd">
    <modelVersion>4.0.0</modelVersion>

    <groupId>com.testproject</groupId>
    <artifactId>liquibase-property-bug</artifactId>
    <version>1.0-SNAPSHOT</version>

    <properties>
        <maven.compiler.source>21</maven.compiler.source>
        <maven.compiler.target>21</maven.compiler.target>
        <project.build.sourceEncoding>UTF-8</project.build.sourceEncoding>
    </properties>

    <dependencies>
        <dependency>
            <artifactId>liquibase-core</artifactId>
            <version>5.0.2</version>
        </dependency>
        <dependency>
            <groupId>org.postgresql</groupId>
            <artifactId>postgresql</artifactId>
            <version>42.7.10</version>
        </dependency>
    </dependencies>

</project>
```

db.changelog.yaml:
```yaml
databaseChangeLog:
  - property:
      dbms: "db2i"
      name: "nameOfDbms"
      value: "Db2 for IBM i"
  - property:
      name: "nameOfDbms"
      value: "Some other DBMS"
  - changeSet:
      id: "create-table"
      author: "Bug Hunter"
      changes:
        - createTable:
            tableName: "what_dbms_am_i"
            columns:
              - column:
                  name: "name"
                  type: "clob"
              - column:
                  name: "description"
                  type: "clob"
        - sql:
            sql: "INSERT INTO what_dbms_am_i (name, description) VALUES ('${nameOfDbms}', '${description}');"
```

2. Make sure the *db.changelog.yaml* is placed within the source/main/resources folder of your Java project. 
3. Change the values of the USERNAME, PASSWORD, and URL constants so that the Java code can connect to your PostgreSQL instance. You can use the currentSchema parameter within the URL to specify which schema should be used. The schema that you specify should be completely empty.
4. Compile and run the Main class while using the *updateWithParameter* argument.
5. Observe that the *what_dbms_am_i* table is filled with 1 row, of which the value of the *name* column equals 'Db2 for IBM i', even though we are using PostgreSQL as our DBMS.

### Expected/Desired Behavior

The value of property *nameOfDbms* should be resolved as 'Some other DBMS'.

### Liquibase Version

5.0.2

### Database Vendor & Version

PostgreSQL 16.13.1

### Liquibase Integration

The CommandScope class within liquibase-core-5.0.2.jar

### Liquibase Extensions

None

Windows 11

### Additional Context

I used the following Java and Maven versions while testing:

- Java 21, using *Temurin-21.0.7+6* as my JDK.
- Maven version *3.9.10*.

mgustimz:
## Description

When a user creates their own `ChangeLogParameters` instance and passes it via `CommandScope` (e.g., to add custom properties), the database filter was not being set. This caused dbms-specific properties in changelogs to be incorrectly resolved.

## Root Cause

In `DatabaseChangelogCommandStep.getChangeLogParameters()`, when a user-provided `ChangeLogParameters` is passed, the database filter (`filterDatabase`) was never set. The `DatabaseList.definitionMatches()` method with a null database and `returnValueIfEmpty=true` would incorrectly match all dbms filters.

## Example

With PostgreSQL:
- property with `dbms: "db2i"` was incorrectly matching because `DatabaseList.definitionMatches(["db2i"], null, true)` returns `true` when the database is null
- property with no dbms filter (fallback) was being skipped

## Fix

When a user-provided `ChangeLogParameters` is passed, we now set the database filter to the actual database's short name, ensuring dbms filtering works correctly.

## Testing

Added `DatabaseChangelogCommandStepTest` to verify the fix.

Fixes: https://github.com/liquibase/liquibase/issues/7602

### DIFF
--- a/liquibase-standard/src/main/java/liquibase/command/core/helpers/DatabaseChangelogCommandStep.java
+++ b/liquibase-standard/src/main/java/liquibase/command/core/helpers/DatabaseChangelogCommandStep.java
@@ -60,16 +60,29 @@ public class DatabaseChangelogCommandStep extends AbstractHelperCommandStep impl
                 .build();
     }
 
+    /**
+     * Returns the classes provided by this command step.
+     * @return list of dependency classes
+     */
     @Override
     public List<Class<?>> providedDependencies() {
         return Arrays.asList(DatabaseChangeLog.class, ChangeLogParameters.class);
     }
 
+    /**
+     * Returns the classes required by this command step.
+     * @return list of required dependency classes
+     */
     @Override
     public List<Class<?>> requiredDependencies() {
         return Arrays.asList(Database.class);
     }
 
+    /**
+     * Executes the changelog command step.
+     * @param resultsBuilder the command results builder
+     * @throws Exception if an error occurs during execution
+     */
     @Override
     public void run(CommandResultsBuilder resultsBuilder) throws Exception {
         CommandScope commandScope = resultsBuilder.getCommandScope();
@@ -92,6 +105,18 @@ public class DatabaseChangelogCommandStep extends AbstractHelperCommandStep impl
         commandScope.provideDependency(ChangeLogParameters.class, changeLogParameters);
     }
 
+    /**
+     * Retrieves the ChangeLogParameters from the command scope, or creates a new instance if none was provided.
+     * <p>
+     * When a user-provided ChangeLogParameters is passed via CommandScope, the database filter
+     * ({@code filterDatabase}) is set from the actual database's short name to ensure that
+     * dbms-specific properties in changelogs are correctly resolved.
+     *
+     * @param commandScope the command scope containing optional CHANGELOG_PARAMETERS argument
+     * @param database the actual database being used
+     * @return the ChangeLogParameters with database filter set appropriately
+     * @see ChangeLogParameters#setDatabase(String)
+     */
     private ChangeLogParameters getChangeLogParameters(CommandScope commandScope, Database database) {
         ChangeLogParameters changeLogParameters = commandScope.getArgumentValue(CHANGELOG_PARAMETERS);
         if (changeLogParameters == null) {
@@ -129,6 +154,11 @@ public class DatabaseChangelogCommandStep extends AbstractHelperCommandStep impl
         addCommandFiltersMdc(labels, contexts);
     }
 
+    /**
+     * Adds label and context filter values to the MDC for logging purposes.
+     * @param labelExpression the label expression to log
+     * @param contexts the contexts to log
+     */
     public static void addCommandFiltersMdc(LabelExpression labelExpression, Contexts contexts) {
         String labelFilterMdc = labelExpression != null && labelExpression.getOriginalString() != null ? labelExpression.getOriginalString() : "";
         String contextFilterMdc = contexts != null ? contexts.toString() : "";
@@ -136,6 +166,14 @@ public class DatabaseChangelogCommandStep extends AbstractHelperCommandStep impl
         Scope.getCurrentScope().addMdcValue(MdcKey.COMMAND_CONTEXT_FILTER, contextFilterMdc);
     }
 
+    /**
+     * Parses and returns the database change log from the specified file.
+     * @param changeLogFile the path to the changelog file
+     * @param changeLogParameters the parameters for parsing
+     * @param database the database context
+     * @return the parsed DatabaseChangeLog
+     * @throws Exception if parsing fails
+     */
     public static DatabaseChangeLog getDatabaseChangeLog(String changeLogFile, ChangeLogParameters changeLogParameters, Database database) throws Exception {
         ResourceAccessor resourceAccessor = Scope.getCurrentScope().getResourceAccessor();
         AtomicReference<DatabaseChangeLog> changelog = new AtomicReference<>();
@@ -154,6 +192,15 @@ public class DatabaseChangelogCommandStep extends AbstractHelperCommandStep impl
         return changelog.get();
     }
 
+    /**
+     * Checks and initializes the liquibase tables if needed.
+     * @param updateExistingNullChecksums whether to update null checksums
+     * @param databaseChangeLog the database change log
+     * @param contexts the context filter
+     * @param labelExpression the label filter
+     * @param database the database
+     * @throws liquibase.exception.LiquibaseException if an error occurs
+     */
     private void checkLiquibaseTables(boolean updateExistingNullChecksums, DatabaseChangeLog databaseChangeLog,
                                       Contexts contexts, LabelExpression labelExpression, Database database) throws LiquibaseException {
         ChangeLogHistoryService changeLogHistoryService = Scope.getCurrentScope().getSingleton(ChangeLogHistoryServiceFactory.class).getChangeLogService(database);
@@ -169,11 +216,19 @@ public class DatabaseChangelogCommandStep extends AbstractHelperCommandStep impl
         LockServiceFactory.getInstance().getLockService(database).init();
     }
 
+    /**
+     * Returns the command names defined by this step.
+     * @return array of command name arrays
+     */
     @Override
     public String[][] defineCommandNames() {
         return new String[][]{COMMAND_NAME};
     }
 
+    /**
+     * Cleans up resources held by this command step.
+     * @param resultsBuilder the command results builder
+     */
     @Override
     public void cleanUp(CommandResultsBuilder resultsBuilder) {
         Scope.getCurrentScope().getSingleton(ChangeLogHistoryServiceFactory.class).resetAll();
--- a/liquibase-standard/src/main/java/liquibase/command/core/helpers/DatabaseChangelogCommandStep.java
+++ b/liquibase-standard/src/main/java/liquibase/command/core/helpers/DatabaseChangelogCommandStep.java
@@ -98,6 +98,10 @@ public class DatabaseChangelogCommandStep extends AbstractHelperCommandStep impl
             changeLogParameters = new ChangeLogParameters(database);
             changeLogParameters.addJavaProperties();
             changeLogParameters.addDefaultFileProperties();
+        } else {
+            // When a user-provided ChangeLogParameters is passed, we still need to set the database
+            // filter so that dbms-specific properties are correctly resolved.
+            changeLogParameters.setDatabase(database.getShortName());
         }
         extractContextAndLabels(commandScope, changeLogParameters);
         return changeLogParameters;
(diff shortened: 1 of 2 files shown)


filipelautert (commented):
Hey, took a look at the fix.

The production change is correct — `ChangeLogParameters.setDatabase(...)` does exactly what's described, and propagating the database short name into a user-supplied `ChangeLogParameters` is a clean fix for #7602.

Two blocking issues in the test, both in `liquibase-standard/src/test/groovy/liquibase/command/core/helpers/DatabaseChangelogCommandStepTest.groovy`:

1. **`Scope.instance(...)` doesn't exist on `liquibase.Scope`.** That class only exposes `getCurrentScope()` and various `child(...)` overloads (all of which require at least a `Map` or key/value pair plus a `ScopedRunner[WithReturn]`). The test won't compile as written. CI will confirm — what you almost certainly want here is just to drop the `Scope.instance(...)` wrapper entirely; the method under test doesn't read anything from the current scope, so calling it directly is enough:
   ```groovy
   def step = new DatabaseChangelogCommandStep()
   def changeLogParams = (ChangeLogParameters) method.invoke(step, commandScope, postgresDb)
   ```

2. **`method.invoke(null, commandScope, postgresDb)` will NPE at runtime even after (1) is fixed**, because `getChangeLogParameters` is a private *instance* method (`liquibase-standard/src/main/java/liquibase/command/core/helpers/DatabaseChangelogCommandStep.java:95`). The first arg to `Method.invoke` for an instance method has to be the instance — pass `new DatabaseChangelogCommandStep()` (see snippet above).

A couple of smaller things, take or leave:

- The Javadoc additions on `providedDependencies`, `requiredDependencies`, `run`, `addCommandFiltersMdc`, `getDatabaseChangeLog`, `checkLiquibaseTables`, `defineCommandNames`, and `cleanUp` are unrelated to #7602 and mostly restate the method name in prose ("Returns the classes provided by this command step."). I'd scope this PR down to just the fix + a Javadoc on the new `getChangeLogParameters` behavior. The bulk Javadoc cleanup is fine but belongs in its own PR so this one stays focused.
- Reflection-based tests are brittle. Once the assertion is just "filter is set to the database short name," you can probably test that more directly by extracting the body into a package-private helper (or by exercising the public `run(...)` path with a fake `CommandResultsBuilder`). Not blocking; if you'd rather keep the reflection test, that's fine — just fix (1) and (2).
- `database.getShortName()` will NPE if `database` is null, but `run()` looks it up via `getDependency(Database.class)` which throws if missing, so this is fine in practice.

Once the test compiles and runs green, this is good.

mgustimz:
> Hey, took a look at the fix.
> 
> The production change is correct — `ChangeLogParameters.setDatabase(...)` does exactly what's described, and propagating the database short name into a user-supplied `ChangeLogParameters` is a clean fix for #7602.
> 
> Two blocking issues in the test, both in `liquibase-standard/src/test/groovy/liquibase/command/core/helpers/DatabaseChangelogCommandStepTest.groovy`:
> 
> 1. **`Scope.instance(...)` doesn't exist on `liquibase.Scope`.** That class only exposes `getCurrentScope()` and various `child(...)` overloads (all of which require at least a `Map` or key/value pair plus a `ScopedRunner[WithReturn]`). The test won't compile as written. CI will confirm — what you almost certainly want here is just to drop the `Scope.instance(...)` wrapper entirely; the method under test doesn't read anything from the current scope, so calling it directly is enough:
>    ```groovy
>    def step = new DatabaseChangelogCommandStep()
>    def changeLogParams = (ChangeLogParameters) method.invoke(step, commandScope, postgresDb)
>    ```
> 2. **`method.invoke(null, commandScope, postgresDb)` will NPE at runtime even after (1) is fixed**, because `getChangeLogParameters` is a private _instance_ method (`liquibase-standard/src/main/java/liquibase/command/core/helpers/DatabaseChangelogCommandStep.java:95`). The first arg to `Method.invoke` for an instance method has to be the instance — pass `new DatabaseChangelogCommandStep()` (see snippet above).
> 
> A couple of smaller things, take or leave:
> 
> * The Javadoc additions on `providedDependencies`, `requiredDependencies`, `run`, `addCommandFiltersMdc`, `getDatabaseChangeLog`, `checkLiquibaseTables`, `defineCommandNames`, and `cleanUp` are unrelated to #7602 and mostly restate the method name in prose ("Returns the classes provided by this command step."). I'd scope this PR down to just the fix + a Javadoc on the new `getChangeLogParameters` behavior. The bulk Javadoc cleanup is fine but belongs in its own PR so this one stays focused.
> * Reflection-based tests are brittle. Once the assertion is just "filter is set to the database short name," you can probably test that more directly by extracting the body into a package-private helper (or by exercising the public `run(...)` path with a fake `CommandResultsBuilder`). Not blocking; if you'd rather keep the reflection test, that's fine — just fix (1) and (2).
> * `database.getShortName()` will NPE if `database` is null, but `run()` looks it up via `getDependency(Database.class)` which throws if missing, so this is fine in practice.
> 
> Once the test compiles and runs green, this is good.

@filipelautert Thank you for the review.

I already refactor it. Please re-review.

filipelautert (approved):
Hey, took another look — both blockers from last round are fixed: the bogus `Scope.instance(...)` wrapper is gone, and `method.invoke` now passes a real `new DatabaseChangelogCommandStep()` instance (`liquibase-standard/src/test/groovy/liquibase/command/core/helpers/DatabaseChangelogCommandStepTest.groovy:52`). The unused `Scope` import was also cleaned up.

I sanity-checked the test against `ChangeLogParameters`: no-arg constructor exists (`ChangeLogParameters.java:119`), `getDatabase()` returns the `filterDatabase` field (`:327`), so the `changeLogParams.getDatabase() == "postgresql"` assertion will genuinely exercise the fix at `DatabaseChangelogCommandStep.java:128`. Compiles and asserts the right thing.

One small leftover from last review: the drive-by Javadocs on `providedDependencies`, `requiredDependencies`, `run`, `addCommandFiltersMdc`, `getDatabaseChangeLog`, `checkLiquibaseTables`, `defineCommandNames`, and `cleanUp` are still in the diff. Still not blocking — just would've been cleaner as a separate PR.

LGTM, approval-worthy once CI confirms green.

KeyboardToCode:
Hey @mgustimz,

Can you give any indication when this fix will be merged to main? I would like to use a nightly build to see if the fix works for us as well.

Thank you in advance!

yodzhubeiskyi (approved):
Verified fix on against H2 and PostgreSQL.
  
  The root cause was confirmed on main: when a user-provided hangeLogParameters is passed via CommandScope, filterDatabase is never set (remains null), causing Filter.matches() to skip DBMS evaluation entirely - so a  property tagged dbms="db2i" would resolve even when connected to PostgreSQL.

  After the fix, getChangeLogParameters() calls hangeLogParameters.setDatabase(database.getShortName()) in the else branch, which correctly wires DBMS filtering for user-provided params.

  Tested with a changelog defining three overlapping <property> entries (dbms="db2i", dbms="postgresql", no dbms). With a user-provided ChangeLogParameters (no-arg constructor) passed via CommandScope:
  - PostgreSQL -> resolved postgresql-specific value 
  - H2 -> resolved fallback value 
  
  Before the fix both cases resolved to db2i-specific value regardless of the connected database.

  DatabaseChangelogCommandStepTest ? 1 test, 0 failures. ChangeLogParametersTest ? 31 tests, 0 failures.

  LGTM. Thanks for the fix, @mgustimz!

PR #7697, opened 2026-05-07
Fix #7623: Handle null URL in isCorrectDatabaseImplementation

### LINKED ISSUE #7623: NPE in CockroachDatabase.isCorrectDatabaseImplementation when connection.getURL() returns null
## Description

`CockroachDatabase.isCorrectDatabaseImplementation()` calls `conn.getURL().startsWith(...)` without null-checking the URL. When a JDBC driver returns `null` from `DatabaseMetaData.getURL()` (as IBM Informix JDBC 15.0.1.0 does), this throws an NPE.

The confusing part: the NPE stack trace points at `CockroachDatabase` even when the user is connecting to a completely unrelated database (e.g., Informix), because `DatabaseFactory.findCorrectDatabaseImplementation()` iterates over **all** registered database implementations calling `isCorrectDatabaseImplementation()` on each.

## Stack Trace

```
java.lang.NullPointerException: Cannot invoke "String.startsWith(String)" because "url" is null
    at liquibase.database.core.CockroachDatabase.isCorrectDatabaseImplementation(CockroachDatabase.java:65)
    at liquibase.database.DatabaseFactory.findCorrectDatabaseImplementation(DatabaseFactory.java:113)
    at liquibase.database.DatabaseFactory.openDatabase(DatabaseFactory.java:153)
```

## Root Cause

`DatabaseFactory.findCorrectDatabaseImplementation()` iterates all implementations. `CockroachDatabase` happens to be checked early and does:

```java
final String url = conn.getURL();
if (url.startsWith("jdbc:postgres") || url.startsWith("postgres")) {  // NPE when url is null
```

## Affected Classes

The same null-unsafe pattern exists in at least:

| Class | Line | Unsafe Code |
|-------|------|-------------|
| `CockroachDatabase` | 64-65 | `conn.getURL().startsWith("jdbc:postgres")` |
| `EnterpriseDBDatabase` | 18-19 | `conn.getURL().startsWith("jdbc:edb:")` |
| `H2Database` | 325-328 | `getConnection().getURL().startsWith("jdbc:h2:tcp:")` (in `isSafeToRunUpdate`) |
| `MySQLDatabase` | 703 | `getConnection().getURL().contains("useAffectedRows=true")` |

`HsqlDatabase` already handles this correctly with a null guard.

## Suggested Fix

Return `false` when `conn.getURL()` is null in `isCorrectDatabaseImplementation()` — a null URL can never match any database type. For other usages, add appropriate null guards.

## Trigger Scenario

IBM/HCL Informix JDBC driver 15.0.1.0 returns `null` from `DatabaseMetaData.getURL()`. While this is arguably a driver bug (violates JDBC spec), Liquibase should handle it gracefully since it iterates all database implementations during connection setup.

**Liquibase version**: 5.0.2

mgustimz:
## Description

When `DatabaseMetaData.getURL()` returns `null` (e.g., IBM Informix JDBC driver 15.0.1.0), calling `url.startsWith()` throws an NPE.

## Root Cause

`CockroachDatabase.isCorrectDatabaseImplementation()` and `EnterpriseDBDatabase.isCorrectDatabaseImplementation()` were calling `url.startsWith()` without null-checking the URL first. The `DatabaseFactory.findCorrectDatabaseImplementation()` iterates over all registered database implementations, so this NPE would occur even when connecting to unrelated databases.

## Stack Trace

```
java.lang.NullPointerException: Cannot invoke "String.startsWith(String)" because "url" is null
    at liquibase.database.core.CockroachDatabase.isCorrectDatabaseImplementation(CockroachDatabase.java:65)
    at liquibase.database.DatabaseFactory.findCorrectDatabaseImplementation(DatabaseFactory.java:113)
    at liquibase.database.DatabaseFactory.openDatabase(DatabaseFactory.java:153)
```

## Fix

- **CockroachDatabase**: Return `false` early when URL is null. Also check for non-matching patterns before attempting database query.
- **EnterpriseDBDatabase**: Return `false` early when URL is null or doesn't match expected patterns.

A null URL can never match any database type, so returning `false` is the correct behavior.

## Testing

- `CockroachDatabaseTest`: Added test cases for null URL and non-postgres URL
- `EnterpriseDBDatabaseTest`: Added test cases for null URL and oracle URL

Fixes: https://github.com/liquibase/liquibase/issues/7623

### DIFF
--- a/liquibase-standard/src/main/java/liquibase/database/core/CockroachDatabase.java
+++ b/liquibase-standard/src/main/java/liquibase/database/core/CockroachDatabase.java
@@ -59,22 +59,33 @@ public class CockroachDatabase extends PostgresDatabase {
         return this.databaseMinorVersion;
     }
 
+    /**
+     * Checks if the given connection points to a CockroachDB database.
+     * <p>
+     * Returns false early if the URL is null (which can happen with some JDBC drivers
+     * like IBM Informix), since a null URL cannot match any database type.
+     *
+     * @param conn the database connection to check
+     * @return true if this is a CockroachDB connection, false otherwise
+     * @throws DatabaseException if there is an error querying the database
+     */
     @Override
     public boolean isCorrectDatabaseImplementation(DatabaseConnection conn) throws DatabaseException {
         final String url = conn.getURL();
-        if (url.startsWith("jdbc:postgres") || url.startsWith("postgres")) {
-            if (conn instanceof JdbcConnection) {
-                try (Statement stmt = ((JdbcConnection) conn).createStatement()) {
-                    if (stmt != null) {
-                        try (ResultSet rs = stmt.executeQuery("select version()")) {
-                            if (rs.next()) {
-                                return ((String) JdbcUtil.getResultSetValue(rs, 1)).startsWith("CockroachDB");
-                            }
+        if (url == null || (!url.startsWith("jdbc:postgres") && !url.startsWith("postgres"))) {
+            return false;
+        }
+        if (conn instanceof JdbcConnection) {
+            try (Statement stmt = ((JdbcConnection) conn).createStatement()) {
+                if (stmt != null) {
+                    try (ResultSet rs = stmt.executeQuery("select version()")) {
+                        if (rs.next()) {
+                            return ((String) JdbcUtil.getResultSetValue(rs, 1)).startsWith("CockroachDB");
                         }
                     }
-                } catch (SQLException throwables) {
-                    return false;
                 }
+            } catch (SQLException throwables) {
+                return false;
             }
         }
 
--- a/liquibase-standard/src/main/java/liquibase/database/core/EnterpriseDBDatabase.java
+++ b/liquibase-standard/src/main/java/liquibase/database/core/EnterpriseDBDatabase.java
@@ -13,23 +13,34 @@ import java.sql.Statement;
 
 public class EnterpriseDBDatabase extends PostgresDatabase {
 
+    /**
+     * Checks if the given connection points to an EnterpriseDB database.
+     * <p>
+     * Returns false early if the URL is null (which can happen with some JDBC drivers
+     * like IBM Informix), since a null URL cannot match any database type.
+     *
+     * @param conn the database connection to check
+     * @return true if this is an EnterpriseDB connection, false otherwise
+     * @throws DatabaseException if there is an error querying the database
+     */
     @Override
     public boolean isCorrectDatabaseImplementation(DatabaseConnection conn) throws DatabaseException {
         final String url = conn.getURL();
-        if (url.startsWith("jdbc:edb:") || url.startsWith("jdbc:postgres")) {
-            if (conn instanceof JdbcConnection) {
-                try (Statement stmt = ((JdbcConnection) conn).createStatement()) {
-                    if (stmt != null) {
-                        try (ResultSet rs = stmt.executeQuery("select version()")) {
-                            if (rs.next()) {
-                                return ((String) JdbcUtil.getResultSetValue(rs, 1)).contains("EnterpriseDB");
-                            }
+        if (url == null || (!url.startsWith("jdbc:edb:") && !url.startsWith("jdbc:postgres"))) {
+            return false;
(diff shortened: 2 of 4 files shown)


filipelautert (commented):
Hey, took a look at the fix.

The production change is correct — the NPE on null URL is a real bug (some JDBC drivers like IBM Informix return `null` from `getURL()`), and the early-return refactor reads cleaner than the old nested `if`. Behavior is preserved for the non-null path.

One blocking issue:

- **`liquibase-standard/src/test/groovy/liquibase/database/core/CockroachDatabaseTest.groovy`** — the two new test methods (`...returns false when connection URL is null` and `...does not match postgres patterns`) are placed **after** the class's closing brace. They're outside the `CockroachDatabaseTest` class entirely, so the file won't compile (CI is now confirming this). Move both methods inside the class body, before the final `}`.

Smaller things, take or leave:

- For symmetry with `EnterpriseDBDatabaseTest`, you could fold the two new CockroachDB cases into a single `@Unroll`-driven `where:` table instead of two separate methods. Not blocking.
- `CockroachDatabase.isCorrectDatabaseImplementation` and `EnterpriseDBDatabase.isCorrectDatabaseImplementation` now share an almost identical URL-guard prologue. Out of scope here, but a `PostgresDatabase`-level helper would be a nice cleanup later.

Once the test file compiles and CI runs green, this is good to go.

mgustimz:
> Hey, took a look at the fix.
> 
> The production change is correct — the NPE on null URL is a real bug (some JDBC drivers like IBM Informix return `null` from `getURL()`), and the early-return refactor reads cleaner than the old nested `if`. Behavior is preserved for the non-null path.
> 
> One blocking issue:
> 
> * **`liquibase-standard/src/test/groovy/liquibase/database/core/CockroachDatabaseTest.groovy`** — the two new test methods (`...returns false when connection URL is null` and `...does not match postgres patterns`) are placed **after** the class's closing brace. They're outside the `CockroachDatabaseTest` class entirely, so the file won't compile (CI is now confirming this). Move both methods inside the class body, before the final `}`.
> 
> Smaller things, take or leave:
> 
> * For symmetry with `EnterpriseDBDatabaseTest`, you could fold the two new CockroachDB cases into a single `@Unroll`-driven `where:` table instead of two separate methods. Not blocking.
> * `CockroachDatabase.isCorrectDatabaseImplementation` and `EnterpriseDBDatabase.isCorrectDatabaseImplementation` now share an almost identical URL-guard prologue. Out of scope here, but a `PostgresDatabase`-level helper would be a nice cleanup later.
> 
> Once the test file compiles and CI runs green, this is good to go.

@filipelautert Thank you for the review.

I already refactor it. Please re-review.

filipelautert (approved):
Hey, took another look — the prior blocker is fixed, both new tests sit inside the `CockroachDatabaseTest` class body and the file compiles.

`liquibase-standard/src/test/groovy/liquibase/database/core/CockroachDatabaseTest.groovy:65-88` — methods are properly nested before the closing brace, `DatabaseConnection` import added cleanly. Production changes in `CockroachDatabase.java:75-77` and `EnterpriseDBDatabase.java:25-27` correctly short-circuit on null/non-matching URLs.

The two earlier non-blocking notes (folding the Cockroach cases into an `@Unroll where:` table to mirror `EnterpriseDBDatabaseTest`, and deduping the URL-guard prologue between the two `isCorrectDatabaseImplementation` impls) weren't tackled — fine to leave, but they're still on the table if you want to revisit later.

LGTM, approval-worthy.

rberezen (commented):
Tested locally — verified the bug reproduces on master with the PR's null URL test:
```
java.lang.NullPointerException: Cannot invoke "String.startsWith(String)" because "url" is null
    at liquibase.database.core.CockroachDatabase.isCorrectDatabaseImplementation(CockroachDatabase.java:65)
```
Matches the stack trace from the issue. On the PR branch, the same test passes.

- 42/42 unit tests pass
- All 10 CI integration tests pass
- Minimal fix: early `return false` when URL is null or doesn't match the expected pattern

LGTM.

rberezen (approved):

Issue #7385, opened 2025-10-20
DATABASECHANGELOGLOCK table creation fails in PostgreSQL when schema name requires quoting


mkonzal:
### Search first

- [x] I searched and no similar issues were found

### Description

When using Liquibase with PostgreSQL and setting database.objectQuotingStrategy = ObjectQuotingStrategy.QUOTE_ALL_OBJECTS, the DATABASECHANGELOG table is correctly created in the quoted schema. However, the creation of the DATABASECHANGELOGLOCK table fails with a syntax error if the schema name begins with a digit and requires quoting.

Logs

```
25-10-20 21:31:25.710 DEBUG liquibase.logging.core.JavaLogger:37          : Failed to create or initialize the lock table, trying again, iteration 1 of 10

liquibase.exception.UnexpectedLiquibaseException: liquibase.exception.DatabaseException: org.postgresql.util.PSQLException: ERROR: syntax error at or near "b5d63f449304c05a291c4c27136ebb5"
  Position: 21
	at liquibase.lockservice.StandardLockService.isDatabaseChangeLogLockTableCreated(StandardLockService.java:246)
	at liquibase.lockservice.StandardLockService.init(StandardLockService.java:119)
	at liquibase.command.core.helpers.DatabaseChangelogCommandStep.checkLiquibaseTables(DatabaseChangelogCommandStep.java:165)
	at liquibase.command.core.helpers.DatabaseChangelogCommandStep.run(DatabaseChangelogCommandStep.java:88)
	at liquibase.command.CommandScope.lambda$execute$6(CommandScope.java:263)
	at liquibase.Scope.child(Scope.java:223)
	at liquibase.Scope.child(Scope.java:199)
	at liquibase.command.CommandScope.execute(CommandScope.java:251)
	at liquibase.Liquibase.lambda$update$0(Liquibase.java:216)
	at liquibase.Scope.lambda$child$0(Scope.java:214)
	at liquibase.Scope.child(Scope.java:223)
	at liquibase.Scope.child(Scope.java:213)
	at liquibase.Scope.child(Scope.java:192)
	at liquibase.Liquibase.runInScope(Liquibase.java:1366)
	at liquibase.Liquibase.update(Liquibase.java:205)
	at liquibase.Liquibase.update(Liquibase.java:188)
	at liquibase.Liquibase.update(Liquibase.java:175)
[...]

PR #7703, opened 2026-05-11
fix(postgres): avoid double-quoting default schema on rollback

### LINKED ISSUE #7624: Schema is quoted twice in PostgreSQL when using preserveSchemaCase
### Search first

### Description

With Liquibase 5.0.2 and PostgreSQL database. When the `preserveSchemaCase` option is set to `true`, the `defaultSchemaName` will be quoted twice in `PostgresDatabase.rollback()`. For example, with the schema `Test`, the rollback will set `""Test""` in the search path instead of `"Test"`. As this schema won't exist in the database, changes will be applied on an incorrect schema, generally public as it follow in the search path.

This seems to be a regression from changes in #7468 (this was working correctly in Liquibase 5.0.1). The `PostgresDatabase.rollback()` method was changed to escape the schema name before sending it to `DatabaseUtils.initializeDatabase()`. Unfortunately, the `DatabaseUtils.initializeDatabase()` method already have a special handling for PostgreSQL that quote the schema name when `preserveSchemaCase` is set, so the schema ends up quoted twice.

Not sure of the best way to fix this. Maybe `DatabaseUtils.initializeDatabase()` shouldn't quote the schema anymore, but it seems to be used in others contexts where the schema isn't pre-quoted. Or maybe the change in `PostgresDatabase.rollback()` should be reverted to send the raw schema to `DatabaseUtils.initializeDatabase()` like before, and have `DatabaseUtils.initializeDatabase()` quote the schema using `database.escapeObjectName()` to handle special characters instead of just checking `preserveSchemaCase`.

### Steps To Reproduce

Create a database schema:
`CREATE SCHEMA "Test";`

Then try to run some changes with `preserveSchemaCase`:
```
Scope.child(GlobalConfiguration.PRESERVE_SCHEMA_CASE.getKey(), "true", () -> {
  var database = DatabaseFactory.getInstance().findCorrectDatabaseImplementation(new JdbcConnection(connection));
  database.setDefaultSchemaName("Test");

  var liquibase = new Liquibase("db/changelog/db.changelog-master.yaml", new ClassLoaderResourceAccessor(), database);
  liquibase.update();
});
```

### Expected/Desired Behavior

Default schema shouldn't be quoted twice when added to the search path in `PostgresDatabase.rollback()` when the `preserveSchemaCase` option is set.

### Liquibase Version

5.0.2

### Database Vendor & Version

PostgreSQL 16

### Liquibase Integration

Spring Boot

### Liquibase Extensions

_No response_

_No response_

### Additional Context

_No response_

officialasishkumar:
## Impact
- [ ] Enhancement/New feature
- [ ] Breaking Change

## Description
Fixes #7624.

`PostgresDatabase.rollback()` now passes the raw default catalog and schema names to `DatabaseUtils.initializeDatabase(...)`. `DatabaseUtils` remains responsible for PostgreSQL search path escaping, which prevents preserved-case schema names from being quoted twice after rollback.

## Things to be aware of
The existing search path merge behavior is unchanged; this only removes pre-escaping before initialization.

## Things to worry about
None.

## Additional Context
Validation:
- `./mvnw -pl liquibase-standard -Dtest=liquibase.database.core.PostgresDatabaseTest test -DtrimStackTrace=false`

### DIFF
--- a/liquibase-standard/src/main/java/liquibase/database/core/PostgresDatabase.java
+++ b/liquibase-standard/src/main/java/liquibase/database/core/PostgresDatabase.java
@@ -416,7 +416,7 @@ public class PostgresDatabase extends AbstractJdbcDatabase {
         // SET SEARCH_PATH SQL statements
         final Executor executor = Scope.getCurrentScope().getSingleton(ExecutorService.class).getExecutor("jdbc", this);
         if(executor.updatesDatabase()) {
-            DatabaseUtils.initializeDatabase(this.escapeObjectName(getDefaultCatalogName(), Catalog.class), this.escapeObjectName(getDefaultSchemaName(), Schema.class), this);
+            DatabaseUtils.initializeDatabase(getDefaultCatalogName(), getDefaultSchemaName(), this);
         }
     }
 
--- a/liquibase-standard/src/test/java/liquibase/database/core/PostgresDatabaseTest.java
+++ b/liquibase-standard/src/test/java/liquibase/database/core/PostgresDatabaseTest.java
@@ -1,24 +1,36 @@
 package liquibase.database.core;
 
 import liquibase.GlobalConfiguration;
+import liquibase.Scope;
 import liquibase.changelog.column.LiquibaseColumn;
 import liquibase.database.AbstractJdbcDatabaseTest;
 import liquibase.database.Database;
+import liquibase.database.DatabaseConnection;
 import liquibase.database.ObjectQuotingStrategy;
 import liquibase.exception.DatabaseException;
+import liquibase.executor.Executor;
+import liquibase.executor.ExecutorService;
+import liquibase.statement.core.RawParameterizedSqlStatement;
 import liquibase.structure.DatabaseObject;
 import liquibase.structure.core.Schema;
 import liquibase.structure.core.Table;
 import liquibase.structure.core.View;
 import liquibase.util.StringUtil;
+import org.junit.jupiter.api.AfterEach;
 import org.junit.jupiter.api.Test;
 import org.junit.jupiter.params.ParameterizedTest;
 import org.junit.jupiter.params.provider.Arguments;
 import org.junit.jupiter.params.provider.MethodSource;
+import org.mockito.ArgumentCaptor;
 
 import static org.junit.Assert.assertNull;
 import static org.junit.Assert.assertTrue;
 import static org.junit.jupiter.api.Assertions.assertEquals;
+import static org.mockito.ArgumentMatchers.any;
+import static org.mockito.ArgumentMatchers.eq;
+import static org.mockito.Mockito.mock;
+import static org.mockito.Mockito.verify;
+import static org.mockito.Mockito.when;
 
 import java.util.stream.Stream;
 
@@ -29,6 +41,11 @@ public class PostgresDatabaseTest extends AbstractJdbcDatabaseTest {
 
     public PostgresDatabaseTest() throws Exception {
         super(new PostgresDatabase());
+    }
+
+    @AfterEach
+    void resetExecutorService() {
+        Scope.getCurrentScope().getSingleton(ExecutorService.class).reset();
     }
 
     @Override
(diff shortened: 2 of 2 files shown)


filipelautert (approved):
Hey, reviewed. The fix is correct and nicely diagnosed. `rollback()` was the *only* one of the three `DatabaseUtils.initializeDatabase` call sites that pre-escaped its catalog/schema via `escapeObjectName` — `CommandLineUtils` and `AbstractDatabaseConnectionCommandStep` both pass raw names and let `DatabaseUtils` own the Postgres search-path quoting. So passing raw names here just restores that consistency, and `initializeDatabase` quoting once (under `PRESERVE_SCHEMA_CASE` / `@`) eliminates the `"…"` → `"""…"""` double-quote on rollback.

The new test is solid — it captures the executed statement and asserts the single-quoted `SET LOCAL SEARCH_PATH TO "Test", "$user","public"`, and the `@AfterEach` executor reset keeps the mock from leaking into sibling tests. Good hygiene.

Tiny aside (not blocking, pre-existing): `rollbackUsesEscapedSchemaName` doesn't actually call `rollback()` — it only checks `escapeObjectName` — so its name is a bit misleading now that rollback no longer escapes. Worth a rename someday.

Approving.

rberezen:
Thanks for tackling this, @officialasishkumar. I verified the fix against PostgreSQL and it does resolve #7624 — but unfortunately **it reintroduces #7385** (the bug that PR #7468 originally fixed). Requesting changes.

**What I tested** (applied this PR's one-line change onto current `master`, captured the emitted `SET SEARCH_PATH`):

| Scenario | This PR | Correct |
|---|---|---|
| #7624 — `preserveSchemaCase=true`, schema `Test` | `SET SEARCH_PATH TO "Test", …` ✅ | `"Test"` |
| #7385 — `QUOTE_ALL_OBJECTS`, schema `4b5d63f4…` (numeric-leading) | `SET SEARCH_PATH TO 4b5d63f4…, …` ❌ | `"4b5d63f4…"` |

The second row is a real regression. Confirmed on PostgreSQL 16:
```
postgres=> SET SEARCH_PATH TO 4b5d63f449304c05a291c4c27136ebb5, "$user","public";
ERROR:  trailing junk after numeric literal at or near "4b5d63f449304c05a291c4c27136ebb5"
```
That's exactly the #7385 failure. It happens because `DatabaseUtils.initializeDatabase()` only quotes the schema when `PRESERVE_SCHEMA_CASE=true` or the name contains `@` (lines 46–51) — it does **not** honor `ObjectQuotingStrategy.QUOTE_ALL_OBJECTS` or quote otherwise-unsafe identifiers. PR #7468 worked around that by pre-escaping in `rollback()`; by removing the pre-escape and *not* strengthening `initializeDatabase`, the numeric/quote-all case falls through unquoted again.

**Suggested fix** — the one the #7624 reporter actually proposed in the issue: keep your `rollback()` change (pass raw names), **and** make `initializeDatabase()` escape via the canonical method instead of the narrow `preserveCase || "@"` check:
```java
// DatabaseUtils.initializeDatabase(), PostgreSQL branch — replace lines 46-51:
String finalSearchPath = database.escapeObjectName(defaultSchemaName, Schema.class);
```
`escapeObjectName()` already honors both `QUOTE_ALL_OBJECTS` and `preserveSchemaCase`, so this fixes both bugs at the source. I verified all three cases with this change:
```
#7385 numeric + QUOTE_ALL   → SET SEARCH_PATH TO "4b5d63f4…", …   ✅
#7624 preserveCase "Test"   → SET SEARCH_PATH TO "Test", …        ✅ (quoted once)
plain lowercase "myschema"  → SET SEARCH_PATH TO myschema, …      ✅ (unchanged)
```
All 32 existing `PostgresDatabaseTest` cases still pass. This is also safer for the other two `initializeDatabase()` callers (`AbstractDatabaseConnectionCommandStep`, `CommandLineUtils`) — they already pass raw names and currently rely on the same narrow quoting.

**Two smaller notes:**
1. Your new test `rollbackQuotesPreservedCasePostgresSchemaOnce` asserts `SET LOCAL SEARCH_PATH TO …`, but current `master` emits `SET SEARCH_PATH TO …` (no `LOCAL`) — the branch looks a bit behind `master`. Worth rebasing so the assertion matches.
2. Please add a regression test for the `QUOTE_ALL_OBJECTS` + numeric-schema case so #7385 can't silently come back.

Happy to re-review once the escaping is moved into `initializeDatabase`. Thanks again!

rberezen:
Correction to my earlier review — my "smaller note 1" was wrong. I'd tested against a stale local checkout; current `main` **does** emit `SET LOCAL SEARCH_PATH TO …` (from #7488), so your `SET LOCAL` assertion is correct and the branch is not behind on that point. Apologies for the noise.

The core finding stands: the change still reintroduces #7385 (a numeric-leading schema under `QUOTE_ALL_OBJECTS` is emitted unquoted, which PostgreSQL rejects with `trailing junk after numeric literal`), and moving the escaping into `initializeDatabase` via `escapeObjectName()` resolves both #7624 and #7385.

PR #7704, opened 2026-05-11
fix(changelog): fall back when stored changelog path is missing

### LINKED ISSUE #7587: NullPointerException in DatabaseChangeLog.ranChangeSetExists when storedChangeLog is null (MongoDB + logicalFilePath)
### Search first

### Description

Liquibase throws a NullPointerException when running migrations against MongoDB if the stored changelog entry does not contain `storedChangeLog`.

The error happens when changelog files were renamed and `logicalFilePath` was introduced to prevent Liquibase from re-running migrations.

When Liquibase processes changelogs using `includeAll`, it calls `DatabaseChangeLog.ranChangeSetExists()` and attempts to compare:

ranChangeSet.getStoredChangeLog().equals(...)

However `storedChangeLog` is `null`, which causes a NullPointerException.

**Stacktrace:**
Caused by: java.lang.NullPointerException:
Cannot invoke "String.equals(Object)" because the return value of
"liquibase.changelog.RanChangeSet.getStoredChangeLog()" is null

at liquibase.changelog.DatabaseChangeLog.ranChangeSetExists(DatabaseChangeLog.java:1187)

This happens when the DATABASECHANGELOG collection already contains entries created by an earlier Liquibase version where `storedChangeLog` was not stored.

Liquibase should handle this case safely instead of throwing an exception.

### Steps To Reproduce

1. Run Liquibase MongoDB migrations using includeAll
   Root changelog loads all files from the changelog directory.
{
  "databaseChangeLog": [
    {
      "includeAll": {
        "path": "db/changelogs"
      }
    }
  ]
}
One of the files inside the directory:
  db/changelogs/changelog-1.0.json

1. Run Liquibase MongoDB migrations using includeAll
   Root changelog loads all files from the changelog directory.
{
  "databaseChangeLog": [
    {
      "includeAll": {
        "path": "db/changelogs"
      }
    }
  ]
}

One of the files inside the directory:
db/changelogs/changelog-1.0.json

3. Rename the changelog file

The file is renamed:
changelog-1.0.json → changelog-001.0.json

4. Add logicalFilePath to keep the original Liquibase identity

To prevent Liquibase from re-running the migration, logicalFilePath is added to the renamed file:
{
  "databaseChangeLog": [
    {
      "logicalFilePath": "db/changelogs/changelog-1.0.json"
    }
  ]
}

5. Run Liquibase again

Liquibase loads changelog files using includeAll.

During the process of checking if a changeset already ran, Liquibase throws:

NullPointerException: ranChangeSet.getStoredChangeLog() is null

This occurs while Liquibase compares the stored changelog entry with the current changelog file.

### Expected/Desired Behavior

Liquibase should not throw a NullPointerException when `storedChangeLog` is null.

Instead it should handle the comparison safely, for example using a null-safe comparison:

Objects.equals(ranChangeSet.getStoredChangeLog(), changeLog)

This would allow Liquibase to correctly detect existing changesets even if older DATABASECHANGELOG entries do not contain `storedChangeLog`.

### Liquibase Version

4.33.0

### Database Vendor & Version

MongoDB 7.x

### Liquibase Integration

Quarkus Liquibase MongoDB Extension (Gradle)

### Liquibase Extensions

liquibase-mongodb

Windows 10 (development)

### Additional Context

The issue appeared after upgrading Liquibase versions.

Older DATABASECHANGELOG records created by previous Liquibase versions do not contain `storedChangeLog`.

When Liquibase processes changelogs via `includeAll`, it attempts to compare the stored changelog with the current changelog and crashes due to the null value.

This situation commonly happens when:
- changelog files are renamed
- logicalFilePath is used to preserve migration history
- DATABASECHANGELOG entries were created by older Liquibase versions

### LINKED ISSUE #7667: RanChangeSet.getStoredChangeLog() returns null
### Search first

### Description

I am using Liquibase with Quarkus and MongoDB.

When the Quarkus application starts, Liquibase runs all migration files and creates the DATABASECHANGELOG collection in MongoDB.
Liquibase checks the logicalFilePath against the stored fileName value, and if they match, that migration is skipped.

Example document in DATABASECHANGELOG:

```
{
    "_id" : ObjectId("69e6385c9f0428fb282456b9"),
    "fileName" : "db/changelogs/changelog-1.0.json",
    "id" : "1",
    "author" : "liquibase",
    "md5sum" : "9:e06981c7a50023f09b43c2b14f4dac66",
    "dateExecuted" : ISODate("2026-04-20T14:29:48.690+0000"),
    "tag" : null,
    "execType" : "EXECUTED",
    "description" : "runCommand",
    "comments" : null,
    "contexts" : null,
    "labels" : null,
    "deploymentId" : "6695387086",
    "orderExecuted" : 1,
    "liquibase" : "4.33.0"
}
```

After upgrading Quarkus and Gradle, the application now fails during startup with this error:

`liquibase.changelog.RanChangeSet.getStoredChangeLog() is null`

Actual error in the logs
```
Caused by: liquibase.exception.SetupException: Cannot invoke "String.equals(Object)" because the return value of "liquibase.changelog.RanChangeSet.getStoredChangeLog()" is null
        at liquibase.changelog.DatabaseChangeLog.includeAll(DatabaseChangeLog.java:851)
        at liquibase.changelog.DatabaseChangeLog.handleIncludeAll(DatabaseChangeLog.java:619)
        at liquibase.changelog.DatabaseChangeLog.handleChildNodeHelper(DatabaseChangeLog.java:483)
        at liquibase.changelog.DatabaseChangeLog.handleChildNode(DatabaseChangeLog.java:464)
        at liquibase.changelog.DatabaseChangeLog.load(DatabaseChangeLog.java:431)
        at liquibase.nosql.parser.json.JsonNoSqlChangeLogParser.parse(JsonNoSqlChangeLogParser.java:145)
```

I noticed that in a fresh project using the same Liquibase migration files, the DATABASECHANGELOG collection still does not contain the storedChangeLog field.

If any records already exist in the DATABASECHANGELOG collection, Liquibase throws the same error because storedChangeLog is missing.

I also check by add storedChangeLog field in the DATABASECHANGELOG DB collection for every document in mongoDB after that if I start the my quarkus service then still it is not working and giving same error.

The issue appears to be that newer Liquibase code expects storedChangeLog to be present, but MongoDB changelog documents only contain fileName.

### Steps To Reproduce

In the quarkus + mongoDB project add liqubase migration and build the project.

### Expected/Desired Behavior

`RanChangeSet.getStoredChangeLog() should not return null` and Liquibase should run successfully without throwing an exception.
Existing migration records in DATABASECHANGELOG should be processed correctly, and previously executed migrations should be skipped as expected.

### Liquibase Version

4.33.0

### Database Vendor & Version

Mongo DB

### Liquibase Integration

gradle , quarkus

### Liquibase Extensions

quarkus-liquibase-mongodb

Windows 11

### Additional Context

_No response_

officialasishkumar:
## Impact
- [ ] Enhancement/New feature
- [ ] Breaking Change

## Description
Fixes #7667.
Fixes #7587.

`RanChangeSet.getStoredChangeLog()` now falls back to the changelog path when an older changelog history service or extension does not provide a separate stored path. Explicit stored changelog paths still take precedence when available.

## Things to be aware of
This keeps existing constructor behavior intact and centralizes the fallback in the getter used by logical file path comparisons.

## Things to worry about
None.

## Additional Context
Validation:
- `./mvnw -pl liquibase-standard -Dtest=liquibase.changelog.RanChangeSetTest test -DtrimStackTrace=false`

### DIFF
--- a/liquibase-standard/src/main/java/liquibase/changelog/RanChangeSet.java
+++ b/liquibase-standard/src/main/java/liquibase/changelog/RanChangeSet.java
@@ -86,6 +86,13 @@ public class RanChangeSet {
         return (Date) dateExecuted.clone();
     }
 
+    public String getStoredChangeLog() {
+        if (storedChangeLog == null) {
+            return changeLog;
+        }
+        return storedChangeLog;
+    }
+
     @Override
     public boolean equals(Object o) {
         if (this == o) {
--- a/liquibase-standard/src/test/java/liquibase/changelog/RanChangeSetTest.java
+++ b/liquibase-standard/src/test/java/liquibase/changelog/RanChangeSetTest.java
@@ -2,6 +2,7 @@ package liquibase.changelog;
 
 import org.junit.Test;
 
+import static org.junit.Assert.assertEquals;
 import static org.junit.Assert.assertFalse;
 import static org.junit.Assert.assertTrue;
 
@@ -26,5 +27,19 @@ public class RanChangeSetTest {
         RanChangeSet ranChangeSet = new RanChangeSet("db/ran-changeset.log", "1", "author", null, null, null, null, null, null, null, null, null);
         ChangeSet incomingChangeSet = new ChangeSet("1", "author", false, false, "classpath:/db/file.log", null, null, null);
         assertFalse(ranChangeSet.isSameAs(incomingChangeSet));
+    }
+
+    @Test
+    public void stored_change_log_falls_back_to_change_log() {
+        RanChangeSet ranChangeSet = new RanChangeSet("db/file.log", "1", "author", null, null, null, null, null, null, null, null, null);
+
+        assertEquals("db/file.log", ranChangeSet.getStoredChangeLog());
+    }
+
+    @Test
+    public void stored_change_log_uses_explicit_stored_path() {
+        RanChangeSet ranChangeSet = new RanChangeSet("logical/file.log", "1", "author", null, null, null, null, null, null, null, null, null, "physical/file.log");
+
+        assertEquals("physical/file.log", ranChangeSet.getStoredChangeLog());
     }
 }


filipelautert (approved):
Hey, took a look at the fix.

Production change is right — it mirrors the existing fallback pattern in `ChangeSet.getStoredFilePath()` (`liquibase-standard/src/main/java/liquibase/changelog/ChangeSet.java:362-367`) and directly resolves the NPE at `DatabaseChangeLog.java:1280` where `rc.getStoredChangeLog().equals(...)` blows up on legacy/extension-written DATABASECHANGELOG rows that never populated the stored path.

No blocking issues.

Couple of take-or-leave observations:

- `RanChangeSet.java:89` — the hand-written getter now shadows the Lombok `@Getter`. Worth a one-line Javadoc explaining why (fallback for legacy/extension rows missing `storedChangeLog`), otherwise the next person sees a Lombok class with one mystery override and wonders. The CodeRabbit suggestion is reasonable here.
- The two callers in `DatabaseChangeLog.java:399,406` and `ValidatingVisitor.java:112` will now propagate `changeLog` into `ChangeSet.setStoredFilePath(...)` instead of `null`. Downstream behavior is unchanged (the `UpdateChangeSet{Filename,Checksum}Generator` paths return the same string either way), but worth being aware of if anyone relied on `storedFilePath == null` as a signal — quick grep says nobody does.
- No WARN log on fallback. Probably fine given this getter is called from a stream filter; logging would be noisy. The parallel `ChangeSet.getStoredFilePath()` doesn't log either.

LGTM.

sfpoc:
Hello @officialasishkumar, I hope this issue fix will also be resolved in the [Liquibase v4.33.x] next release, as Quarkus currently utilizes this version chain.

rberezen (commented):
Full e2e verification on H2/offline mode:

Reproduced the NPE on master with a clean scenario:
- Parent changelog with `<includeAll>`
- Pre-populated offline CSV containing one ran changeset
- Run `update-sql` → **NPE matches issue stack trace exactly**

Same scenario on PR branch: succeeds, generates correct SQL.

Important: this is **not Mongo-specific**. `OfflineChangeLogHistoryService.java:179` constructs `RanChangeSet` without `storedChangeLog`, so `update-sql` / `rollback-sql` / any offline command can hit this on **any database**, not just custom extensions. The fix benefits everyone.

- 5/5 unit tests pass
- All 10 CI integration tests pass
- Fix is minimal (7-line getter override)

LGTM.

rberezen (approved):

PR #7705, opened 2026-05-11
fix(mysql): preserve timestamp column attributes in native types

### LINKED ISSUE #7637: Liquibase 5.0.2 breaks MySQL `on update current_timestamp` in type definition
### Search first

### Description

In Liquibase <=5.0.1 it was possible to define a type as `TIMESTAMP NOT NULL DEFAULT CURRENT_TIMESTAMP ON UPDATE CURRENT_TIMESTAMP` for mysql, to get auto-updating timestamps. While probably never really intended that way, there are some resources on the internet that tell you to just do that ( for example https://www.mail-archive.com/liquibase-user@lists.sourceforge.net/msg00079.html ).

The relevant change is probably https://github.com/liquibase/liquibase/pull/6929/changes#diff-6af77e1bbb4fec85e2b68eb393adc809a4779b351bf5029571970b65983091a7R67-R77 , which effectively removes everything behind the Timestamp and its parameters (which were previously preserved: https://github.com/liquibase/liquibase/pull/6929/changes#diff-6af77e1bbb4fec85e2b68eb393adc809a4779b351bf5029571970b65983091a7R80-R82 ).

### Steps To Reproduce

* Run the following changeset:
```
	<changeSet author="125m125" id="create-example-table">
		<createTable tableName="example">
			<column name="id" type="BINARY(16)">
				<constraints nullable="false" />
			</column>
			<column name="updated" type="TIMESTAMP NOT NULL DEFAULT CURRENT_TIMESTAMP ON UPDATE CURRENT_TIMESTAMP">
				<constraints nullable="false" />
			</column>
		</createTable>
	</changeSet>
```
* The column `updated` is created without the `DEFAULT CURRENT_TIMESTAMP ON UPDATE CURRENT_TIMESTAMP` part.

### Expected/Desired Behavior

* The column `updated` is created including the `DEFAULT CURRENT_TIMESTAMP ON UPDATE CURRENT_TIMESTAMP` part.

In previous versions the full type was used to create the column, meaning that all the changesets using that "feature" now need an additional changeset to add the default/update for new installations. But it was probably never really intended to be abused that way, so not sure if you actually want to restore that behavior.

### Liquibase Version

5.0.2

### Database Vendor & Version

MySQL 8.4

### Liquibase Integration

_No response_

### Liquibase Extensions

_No response_

_No response_

### Additional Context

To restore the previous behavior (if you even want that), we could add a contains-check for "update" or "current_timestamp" and preserve the original type definition in front of the new timestamp handling that was added in https://github.com/liquibase/liquibase/pull/6929 ?

officialasishkumar:
## Impact
- [ ] Enhancement/New feature
- [ ] Breaking Change

## Description
Fixes #7637.

MySQL timestamp conversion now preserves native `timestamp` definitions when they include column attributes such as `NOT NULL`, `DEFAULT CURRENT_TIMESTAMP`, or `ON UPDATE CURRENT_TIMESTAMP`. Plain `timestamp`, precision-only `timestamp(3)`, and timezone aliases continue to normalize through the existing MySQL timestamp mapping.

## Things to be aware of
The parser can drop trailing attributes from precision-bearing raw definitions before they become additional information, so this check uses the raw definition to detect native MySQL timestamp column attributes.

## Things to worry about
None.

## Additional Context
Validation:
- `./mvnw -pl liquibase-standard -Dtest=liquibase.datatype.DataTypeFactoryTest test -DtrimStackTrace=false`

### DIFF
--- a/liquibase-standard/src/main/java/liquibase/datatype/core/TimestampType.java
+++ b/liquibase-standard/src/main/java/liquibase/datatype/core/TimestampType.java
@@ -64,6 +64,10 @@ public class TimestampType extends DateTimeType {
             // All timestamp types (with or without timezone) map to TIMESTAMP
             String lowerDef = originalDefinition.toLowerCase();
 
+            if (isNativeMySqlTimestampDefinitionWithColumnAttributes(lowerDef)) {
+                return new DatabaseDataType(getRawDefinition());
+            }
+
             if (lowerDef.startsWith("java.sql.types.timestamp_with_timezone")
                     || lowerDef.startsWith("java.sql.types.timestamp")
                     || lowerDef.startsWith("java.sql.timestamp")
@@ -219,6 +223,15 @@ public class TimestampType extends DateTimeType {
         }
 
         return super.toDatabaseDataType(database);
+    }
+
+    private boolean isNativeMySqlTimestampDefinitionWithColumnAttributes(String lowerDefinition) {
+        if (!lowerDefinition.matches("^timestamp(\\s|\\(|$).*")
+                || lowerDefinition.matches("^timestamp\\s*(\\([^)]*\\))?\\s*$")) {
+            return false;
+        }
+
+        return !lowerDefinition.matches("^timestamp\\s*(\\([^)]*\\))?\\s+(with|without)\\s+time\\s*zone$");
     }
 
     @Override
--- a/liquibase-standard/src/main/java/liquibase/datatype/core/TimestampType.java
+++ b/liquibase-standard/src/main/java/liquibase/datatype/core/TimestampType.java
@@ -225,6 +225,11 @@ public class TimestampType extends DateTimeType {
         return super.toDatabaseDataType(database);
     }
 
+    /**
+     * Returns true for lowercased MySQL-like TIMESTAMP definitions that include
+     * column attributes to preserve, while excluding plain TIMESTAMP precision
+     * forms and WITH/WITHOUT TIME ZONE aliases that should keep normal handling.
+     */
     private boolean isNativeMySqlTimestampDefinitionWithColumnAttributes(String lowerDefinition) {
         if (!lowerDefinition.matches("^timestamp(\\s|\\(|$).*")
                 || lowerDefinition.matches("^timestamp\\s*(\\([^)]*\\))?\\s*$")) {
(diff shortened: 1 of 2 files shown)


filipelautert (commented):
Hey, took a look at the fix.

The production change is correct — narrow check before the existing
startsWith("timestamp") normalization, restoring pre-5.0.2 behavior for native MySQL
TIMESTAMP definitions carrying column attributes (NOT NULL, DEFAULT CURRENT_TIMESTAMP,
ON UPDATE CURRENT_TIMESTAMP), while still normalizing plain timestamp, timestamp(n),
and timestamptz/with-time-zone aliases. No blocking issues.

Smaller observations:

- TimestampType.java:228-235 — three String.matches() calls per invocation recompile
  the patterns each time. Not a hot path, but if you touch this again, pre-compile to
  static final Pattern constants. A 2-line Javadoc explaining what each branch excludes
  would help — CodeRabbit's flag is reasonable.
- Tests cover MySQLDatabase but not MariaDBDatabase, which inherits from MySQLDatabase
  and inherits this branch. One extra row asserting the same preservation for
  new MariaDBDatabase() locks in behavior against future divergence.
- timestamp(3) with time zone not null now flows through the new branch and is
  preserved verbatim (timezone-exclusion regex anchors with $, so trailing attributes
  defeat it). Almost certainly the right outcome for MySQL — noting the change vs the
  prior normalize-and-drop path.

officialasishkumar:
Addressed in be0eb7a9f: added a short Javadoc for the timestamp detection helper and added MariaDB coverage for preserving native TIMESTAMP column attributes.

filipelautert (approved):
Hey, re-reviewed. You picked up both substantive notes from last round — the MariaDB test row (it inherits this branch, so locking it in matters) and the helper Javadoc. The fix is still correctly scoped: native MySQL/MariaDB `TIMESTAMP … NOT NULL / DEFAULT CURRENT_TIMESTAMP / ON UPDATE` definitions are preserved verbatim, while plain `timestamp`, `timestamp(n)`, and `with/without time zone` aliases keep flowing through normal normalization.

The only leftover is the cosmetic one I'd already marked non-blocking: the three `String.matches()` calls recompile their patterns each invocation — worth hoisting to `static final Pattern` if you ever revisit this file, but it's a cold path. Not gating on it.

Approving.

rberezen:
Tested locally: bug from #7637 reproduces on master (`updated` column created as `timestamp NOT NULL`, stripping `DEFAULT CURRENT_TIMESTAMP ON UPDATE CURRENT_TIMESTAMP`). With this PR, the full definition is preserved on both **MySQL 8.0** and **MariaDB 10.5**:

```
`updated` timestamp NOT NULL DEFAULT CURRENT_TIMESTAMP ON UPDATE CURRENT_TIMESTAMP
```

Edge cases verified:
- `TIMESTAMP(3) NOT NULL DEFAULT CURRENT_TIMESTAMP(3) ON UPDATE CURRENT_TIMESTAMP(3)` → preserved ✅
- lowercase variant → preserved ✅
- plain `TIMESTAMP`, `TIMESTAMP(6)`, `TIMESTAMP WITH/WITHOUT TIME ZONE`, `timestamptz` → existing behavior unchanged (regex correctly excludes these) ✅
- Non-MySQL databases unaffected (change is gated on `MySQLDatabase`).

All 283 `DataTypeFactoryTest` rows pass; CI green across all 11 integration DBs. LGTM — ready to merge. Thanks @officialasishkumar!

rberezen (approved):

PR #7708, opened 2026-05-12
Fix MySQL TINYINT(1) snapshotted as BIT(1) causing spurious diffs

pranithreddym:
MySQL Connector/J maps TINYINT(1) → TYPE_NAME="BIT" by default (tinyInt1isBit=true).  BooleanType.toDatabaseDataType() previously emitted BIT(1) for MySQL when it received an originalDefinition that started with "bit".  Applying a changelog containing BIT(1) against a MySQL database whose column is TINYINT(1) would:
  - create BIT(1) on a fresh deployment (wrong type)
  - report a spurious diff on every generateChangeLog thereafter

Fix: for MySQL (not MariaDB), BooleanType always emits TINYINT(1). MySQL's canonical boolean representation is TINYINT(1); users who genuinely need BIT(n) should use the explicit "bit" data type in their changeset, not "boolean".

MariaDB behaviour is unchanged: "bit" stays as BIT, "boolean" stays as TINYINT(1).

Tests added:
  BooleanTypeTest.mysqlBit1FromJdbcMapsToTinyInt1
  BooleanTypeTest.mysqlBooleanKeywordMapsToTinyInt1
  BooleanTypeTest.mariadbBit1PreservedAsBit
  ColumnSnapshotGeneratorTest - snapshot-level BIT passthrough test

## Impact

### DIFF
--- a/liquibase-standard/src/main/java/liquibase/datatype/core/BooleanType.java
+++ b/liquibase-standard/src/main/java/liquibase/datatype/core/BooleanType.java
@@ -37,10 +37,33 @@ public class BooleanType extends LiquibaseDataType {
         } else if (database instanceof MSSQLDatabase) {
             return new DatabaseDataType(database.escapeDataTypeName("bit"));
         } else if (database instanceof MySQLDatabase) {
-            if (originalDefinition.toLowerCase(Locale.US).startsWith("bit")) {
+            if (database instanceof MariaDBDatabase) {
+                if (originalDefinition.toLowerCase(Locale.US).startsWith("bit")) {
+                    return new DatabaseDataType("BIT", getParameters());
+                }
+                return new DatabaseDataType("TINYINT(1)");
+            }
+            // For MySQL (not MariaDB): collapse bit / bit(1) / boolean → TINYINT(1).
+            //
+            // BooleanType handles the "bit" alias (see @DataTypeInfo).  Two separate
+            // scenarios arrive here:
+            //   a) A snapshot produced by the JDBC driver with tinyInt1isBit=true
+            //      (the default): TINYINT(1) columns are reported as TYPE_NAME="BIT",
+            //      so Liquibase builds originalDefinition="BIT(1)".  Without the guard
+            //      below, generating a changelog from that snapshot would write BIT(1),
+            //      re-applying it would create BIT(1) ≠ TINYINT(1), and every subsequent
+            //      generateChangeLog would see a spurious diff.
+            //   b) A changeset authored with type="boolean" or type="bit" (implying
+            //      size 1 or unspecified): TINYINT(1) is the MySQL de-facto boolean type.
+            //
+            // Preserve genuine BIT(n) columns (n > 1) — users who declare
+            // type="bit(8)" explicitly want a bit-field, not a boolean column.
+            String lowerDef = originalDefinition.toLowerCase(Locale.US);
+            if (lowerDef.startsWith("bit") && getParameters().length > 0
+                    && !"1".equals(String.valueOf(getParameters()[0]))) {
                 return new DatabaseDataType("BIT", getParameters());
             }
-            return database instanceof MariaDBDatabase ? new DatabaseDataType("TINYINT(1)") : new DatabaseDataType("TINYINT");
+            return new DatabaseDataType("TINYINT(1)");
         } else if (database instanceof OracleDatabase) {
             try {
                 if (database.getDatabaseMajorVersion() >= OracleDatabase.ORACLE_23C_MAJOR_VERSION) {
--- a/liquibase-standard/src/test/java/liquibase/datatype/BooleanTypeTest.java
+++ b/liquibase-standard/src/test/java/liquibase/datatype/BooleanTypeTest.java
@@ -1,5 +1,7 @@
 package liquibase.datatype;
 
+import liquibase.database.core.MariaDBDatabase;
+import liquibase.database.core.MySQLDatabase;
 import liquibase.database.core.PostgresDatabase;
 import liquibase.datatype.core.BooleanType;
 import liquibase.exception.UnexpectedLiquibaseException;
@@ -7,6 +9,10 @@ import org.junit.Test;
 
 import static org.junit.Assert.assertEquals;
 import static org.junit.Assert.assertNotEquals;
+
+// DataTypeFactory is needed so that finishInitialization() populates parameters
+// (finishInitialization() alone only stores rawDefinition; it does NOT parse
+// parameters — only fromDescription() / fromObject() do).
 
 public class BooleanTypeTest {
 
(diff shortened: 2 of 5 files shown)


liquibase-standard/src/main/java/liquibase/datatype/core/BooleanType.java
```diff
@@ -37,10 +37,27 @@ public DatabaseDataType toDatabaseDataType(Database database) {
...
+            //      TINYINT(1) to TYPE_NAME="BIT" (tinyInt1isBit=true default). Without this
+            //      guard, a snapshot of TINYINT(1) would produce a changelog containing
+            //      BIT(1), and re-applying that changelog would create BIT(1) instead of
+            //      TINYINT(1) — causing spurious diffs on every subsequent generateChangeLog.
+            //   2. A changeset with type="boolean" that has originalDefinition="boolean".
+            //
+            // Both cases should map to TINYINT(1): it is the MySQL de-facto boolean
+            // convention, identical to what MariaDB emits, and avoids the representation
+            // drift that motivated this fix.
+            return new DatabaseDataType("TINYINT(1)");
```
filipelautert, line 66: Two issues collapse onto this line:

1. **Regression for `BIT(n)` on plain MySQL.** `BooleanType` is registered as the handler for the `bit` / `bool` aliases (see `@DataTypeInfo` at line 18). A changeset declaring `<column type="bit(8)"/>` on MySQL (non-MariaDB) now emits `TINYINT(1)` instead of `BIT(8)`. The PR description says "users who genuinely need BIT(n) should use the explicit bit data type" — but that *is* what they're doing; `bit` resolves to `BooleanType`.

2. **Silent change for plain `TINYINT`.** Previously this branch emitted bare `new DatabaseDataType("TINYINT")`; now it's `TINYINT(1)`. Not mentioned in the PR description; flips rendered SQL for every existing `<column type="boolean"/>` on plain MySQL. CI confirms — the DataTypeFactory tests fail on the changed assertion.

Suggestion that fixes both: only collapse to `TINYINT(1)` when size is missing or 1; preserve `BIT(n)` for n > 1.

```java
            // MySQL: keep BIT(n) for n>1 (explicit user intent); collapse bit(1)/bit/boolean
            // to TINYINT(1) because Connector/J reports TINYINT(1) as TYPE_NAME="BIT" with
            // tinyInt1isBit=true.
            String lower = originalDefinition.toLowerCase(Locale.US);
            if (lower.startsWith("bit") && getParameters().length > 0
                    && !"1".equals(String.valueOf(getParameters()[0]))) {
                return new DatabaseDataType("BIT", getParameters());
            }
            return new DatabaseDataType("TINYINT(1)");
```

Also add a row to `DataTypeFactoryTest.groovy` for `"boolean" | new MySQLDatabase() | "TINYINT(1)"` so the silent change is locked in, and double-check `harness-test-data` / integration changelog goldens — there's almost certainly a `tinyint` vs `tinyint(1)` assertion that flips.

liquibase-standard/src/test/java/liquibase/datatype/BooleanTypeTest.java
```diff
@@ -25,6 +27,51 @@ public void postgresqlBitString() {
...
+    @Test
+    public void mariadbBit1PreservedAsBit() {
+        BooleanType bt = new BooleanType();
+        bt.finishInitialization("BIT(1)");
+        assertEquals(
+            "MariaDB BIT(1) should stay as BIT(1), not be converted to TINYINT(1)",
+            "BIT",
+            bt.toDatabaseDataType(new MariaDBDatabase()).getType()
+        );
+    }
```
filipelautert, line 89: This test doesn't actually verify what it claims. `finishInitialization()` only stores `rawDefinition`; it does *not* populate `parameters` (only `DataTypeFactory.fromDescription(...)` does). So `getParameters()` returns `[]`, `new DatabaseDataType("BIT", [])` yields type string `"BIT"`, and the assertion passes regardless of whether the `(1)` would have been preserved end-to-end.

Use the factory so the parameter actually flows through:

```java
    @Test
    public void mariadbBit1PreservedAsBit() {
        BooleanType bt = (BooleanType) DataTypeFactory.getInstance()
                .fromDescription("bit(1)", new MariaDBDatabase());
        assertEquals("BIT(1)", bt.toDatabaseDataType(new MariaDBDatabase()).toString());
    }
```

Same concern applies (less severely) to `mysqlBit1FromJdbcMapsToTinyInt1` — go through the factory.

filipelautert (commented):
Hey, took a look at the fix.

The diagnosis (Connector/J `tinyInt1isBit=true` reporting TINYINT(1) as TYPE_NAME="BIT") is correct, but the production change overshoots and introduces a regression for legitimate `BIT(n)` columns. CI is already red on this — the DataTypeFactory tests catch the secondary `TINYINT` → `TINYINT(1)` flip. Two inline comments below.

liquibase-standard/src/main/java/liquibase/datatype/core/BooleanType.java
pranithreddym: Good catch on both points — thank you for the detailed review.

**BIT(n) regression fix:** The `bit` alias routing to `BooleanType` is the key insight I missed. I have added a guard before the `TINYINT(1)` fallback:

```java
String lowerDef = originalDefinition.toLowerCase(Locale.US);
if (lowerDef.startsWith("bit") && getParameters().length > 0
        && !"1".equals(String.valueOf(getParameters()[0]))) {
    return new DatabaseDataType("BIT", getParameters());
}
return new DatabaseDataType("TINYINT(1)");
```

This preserves `BIT(8)` (or any `BIT(n)` for n > 1) exactly as authored, while collapsing `bit`, `bit(1)`, and `boolean` to `TINYINT(1)` — eliminating the JDBC driver round-trip drift without breaking genuine bit-field usage.

**Silent `TINYINT` change:** Agreed. All MySQL non-MariaDB paths now uniformly return `TINYINT(1)`. The `AddColumnGeneratorTest` expectation has been updated to reflect this (was `TINYINT`, now correctly `TINYINT(1)`).

Force-pushed with these fixes. All three test classes (`BooleanTypeTest`, `DataTypeFactoryTest`, `AddColumnGeneratorTest`) pass locally.

liquibase-standard/src/test/java/liquibase/datatype/BooleanTypeTest.java
pranithreddym: Spot on — `finishInitialization()` only stores `rawDefinition` and does not parse parameters, so `getParameters()` always returned `[]` and the assertion was vacuously passing regardless of the fix. I have rewritten all four tests in `BooleanTypeTest` to go through `DataTypeFactory.getInstance().fromDescription()` so the full parsing path is exercised:

```java
BooleanType bt = (BooleanType) DataTypeFactory.getInstance()
        .fromDescription("bit(1)", new MariaDBDatabase());
assertEquals("BIT(1)", bt.toDatabaseDataType(new MariaDBDatabase()).toString());
```

I also added a `mysqlBitNPreservedAsBitN` test that verifies `bit(8)` on MySQL stays as `BIT(8)`, and three lock-in rows in `DataTypeFactoryTest.groovy` covering `boolean`, `bit(1)`, and `bit(8)` on MySQL. All pass locally with Java 17 / Maven.

filipelautert (approved):
Hey, re-reviewed after the May 13 changes — both points from the last round are properly addressed:

- The `BIT(n)` regression is fixed: the guard preserves `bit(8)` → `BIT(8)` while collapsing `bit(1)` / bare `bit` / `boolean` → `TINYINT(1)`. Walked the cases, they're all correct.
- The `BooleanTypeTest` cases now route through `DataTypeFactory.fromDescription(...)`, so `getParameters()` is actually populated and the assertions are no longer vacuous. Good catch turning that into real coverage.

One thing for the record (not a blocker): this also flips MySQL's canonical boolean output from `TINYINT` to `TINYINT(1)` for everyone (see the `AddColumnGeneratorTest` updates). That's the right canonical form and matches MariaDB, but existing MySQL users with `TINYINT` boolean columns may see a one-time diff on their next `generate-changelog`. Worth a `notableChanges` label / release-note line so it's not a surprise.

Approving.

rberezen:
Tested locally on MySQL 8.0 + MariaDB 10.5 against `master`. Reproduced the symptom and verified the fix.

**Verified — round-trip fix works:**
```
master:  DB has tinyint(1) → snapshot writes BIT(1) → apply: bit(1)    ❌ drift
PR:      DB has tinyint(1) → snapshot writes TINYINT(1) → apply: tinyint(1)  ✅
```

**Behaviour changes worth flagging** (MySQL only; MariaDB unchanged):

| changeset type | master output | PR output |
|---|---|---|
| `boolean` | `TINYINT` (display width 3) | `TINYINT(1)` |
| `bit` (no params) | `BIT` (==`bit(1)`) | `TINYINT(1)` |
| `bit(1)` | `BIT(1)` | `TINYINT(1)` |
| `bit(n>1)` | `BIT(n)` | `BIT(n)` (unchanged) |

**One direction of drift remains** — for users with real `BIT(1)` columns:
```
DB has bit(1) → PR snapshot writes TINYINT(1) → apply: tinyint(1)    ❌ NEW drift
```
The ambiguity at the JDBC layer (both `tinyint(1)` and `bit(1)` come back as `TYPE_NAME="BIT", SIZE=1` under `tinyInt1isBit=true`) means *some* drift case is unavoidable. The PR picks the dominant convention (TINYINT(1)) — probably the right call, but worth a release-notes mention since explicit `BIT(1)` users have no Liquibase-native path forward (note that `bit` is the alias for `BooleanType` — there's no separate `BitType` class, so `<column type="bit"/>` and `<column type="bit(1)"/>` both go through this code).

**Backward-compat watchout:** schemas where old Liquibase created `TINYINT` (width 3) columns from `type="boolean"` will see `addColumn type="boolean"` produce `TINYINT(1)` after this PR — and `diff-changelog` will surface the discrepancy against existing `TINYINT(3)` columns.

**Verification:**
- All 313 in-scope unit tests pass locally (`BooleanTypeTest`, `DataTypeFactoryTest`, `ColumnSnapshotGeneratorTest`, `AddColumnGeneratorTest`).
- CI integration tests have not run yet — PR is awaiting `authorize` approval.

**Companion PRs prepared (draft, gated on this one merging):**
- liquibase/liquibase-test-harness#1344 — updates MySQL/Percona `expectedSql` + `expectedSnapshot` goldens (boolean/bit columns)
- liquibase/liquibase-pro-tests#2742 — updates MySQL `future_rollback_sql_*` + `snapshot_mysql` goldens (DATABASECHANGELOGLOCK.LOCKED column)

Code itself is clean and well-commented; the MySQL/MariaDB split is correct. Thanks @pranithreddym!

cc @filipelautert

rberezen (approved):

PR #7717, opened 2026-05-13
fix #1694: Add ability to have cascading liquibase.properties files

### LINKED ISSUE #1694: Add ability to have cascading liquibase.properties files
## Description

Currently, we support a single "defaultsFile" (aka liquibase.properties) which can set liquibase configuration values. However, often times people have some of those values that change from environment to environment and it's a pain to have to keep multiple versions of those file with only slight differences.

Liquibase should be able to work like Spring, where you are able to specify a list series of "profiles" and we will check for files with those profile names in them, with the final list of configurations being a union of them all. https://docs.spring.io/spring-boot/docs/current/reference/html/spring-boot-features.html#boot-features-external-config-files-profile-specific

Liquibase already has the concept of a "context" which corresponds to what spring boot calls a "profile" so we will expand that context functionality. (see discussion below)

## Example

Files:

_liquibase.properties_
```
changelogFile: com/example/changelog.xml
url: jdbc://localhost/db
username: myuser
password: pwd
```

_liquibase-staging.properties_
```
url: jdbc://staging/db
username: staginguser
password: stagepwd
```

_liquibase-prod.properties_
```
url: jdbc://prod/db
username: produser
liquibase.headless: true
```

_liquibase-debug.properties_
```
logLevel: FINE
```

If you have `--defaultsFile=liquibase.properties`, it just uses that file and you end up with
```
changelogFile: com/example/changelog.xml
url: jdbc://localhost/db
username: myuser
password: pwd
```

If you use `--defaultsFile=liquibase.properties --contexts=staging`  it merges the two files, with "staging" overriding what is in the default
```
changelogFile: com/example/changelog.xml
url: jdbc://staging/db
username: staginguser
password: stagepwd
```

If you use `--defaultsFile=liquibase.properties --contexts=staging,prod`  it merges the two files, with "staging" overriding what is in the default and "prod" overriding what is in staging+default.
```
changelogFile: com/example/changelog.xml
url: jdbc://prod/db
username: produser
password: stagepwd
```

If you use `--defaultsFile=my.properties --contexts=staging`  it looks for a my.properties and my-staging.properties files, and doesn't touch the liquibase-*.properties files.
```
## No files from above match
```

It is perfectly fine to include contexts that don't have corresponding files. `--defaultsFile=liquibase.properties --contexts=other,staging` will still merge the defaults and staging files, but doesn't care that there is no liquibase-other.properties file.

## Discussion: Contexts vs Profiles

The above expands the functionality of contexts to also include properties files to check for as well. Liquibase "contexts" are often used to define the environment/needs/special-cases that apply that that specific run. 

Currently we use contexts to filter out changeSets that don't apply to that environment/need/special-case and it seems that for many of the reasons people use contexts currently they also want to pick up related files for that context. Example: "staging". And, it's perfectly fine to use contexts that don't have a corresponding defaults file.

BUT: is that overloading the existing context functionality in ways that will cause problem or confusion for people? Would it be better to introduce a separate "profiles" concept that is used only for this new feature? Does a separate property help or hurt?

MichaelKern-IVV:
## Impact

## Description

Currently, we support a single "defaultsFile" (aka liquibase.properties) which can set liquibase configuration values. However, often times people have some of those values that change from environment to environment and it's a pain to have to keep multiple versions of those file with only slight differences.

In addition, there may be sections of the defaultsFile that are maintained by different groups of people and that not everyone involved is authorized to view, such as login credentials.

Unlike requirement #1694 (which should be reopened), this is not achieved by using an additional file whose name includes the stage, but rather by using files with the same name located in different places in the classpath. It is up to the user to configure the classpath according to their own requirements.

Technically, this means that the defaultsFile is read not by an InputStream but by an InputStreamList. For each InputStream contained within it, a separate DefaultsFileValueProvider is created, whose precedence depends on its position in the classpath.


filipelautert (commented):
Hey, reviewed the cascading-defaults work. The `DefaultsFileValueProvider` refactor to static factory helpers is clean, and I appreciate that you preserved the CWE-94 `Properties.load()` guard comment verbatim. A few things to resolve before merge:

1. **Strict-mode "file not found" behavior was dropped.** The old `else` branch threw `UnexpectedLiquibaseException` when a non-default defaults file was missing *and* `strict` was on, and otherwise printed `"Could not find defaults file …"`. The new code replaces that whole branch with an unconditional `getAll()` loop — so if the file exists on neither the filesystem nor the classpath, strict mode no longer fails and the user gets no warning at all. That's a behavior regression. Either re-add the "found nothing anywhere → strict throw / warn" check after the loop, or call out in the PR why it's intentionally removed.

2. **Cascade precedence is implicit and classpath-order dependent.** `precedence = 50 - returnList.size()` means "the earlier a file is returned by `getAll()`, the higher it wins." That makes the override order entirely dependent on the order `ClassLoaderResourceAccessor.getAll()` yields resources. Worth (a) a comment stating the intended rule ("first/closest on the classpath wins"), and (b) confirming that ordering is actually deterministic across resource accessors — otherwise the cascade is non-reproducible. This is the semantic core of #1694, so it shouldn't be left implicit.

3. **Unused imports.** `AccessController`, `PrivilegedAction`, `URLClassLoader`, `URL`, and `MalformedURLException` are imported in `LiquibaseCommandLine` but never referenced (only the test and the maven mojo use a classloader). Looks like leftovers from an earlier approach — drop them.

Minor: `propertiesFromMap` does `properties.putAll(map)` — `Properties` is a `Hashtable`, so a null value in the map will NPE. Worth a null-skip if that constructor can take user data.

Nice test (`MultipleDefaultsFilesOverlayEachOtherTest`) — the overlay assertions are exactly the right shape; just make sure they pin the precedence direction from item 2.

filipelautert (commented):
Thanks for incorporating the earlier round of feedback — the strict-mode restore via the `found` flag, the precedence comment, and the overlay test all look good now. Two remaining items:

1. **`AbstractLiquibaseMojo.java:953-963` vs `:800-810`** - The two cascade paths disagree on direction. `loadProperties()` merges every stream into one `Properties` sequentially, so the **last** stream wins for mojo fields; but the registered providers use `++precedenceOffset` (49, 48, …), so the **first** stream wins for config values. Same multi-file setup, two different winners. Pick one direction (first-wins, to match the CLI) — e.g. load streams in reverse order in `loadProperties`, or skip keys that are already set.

2. **`AbstractLiquibaseMojo.java:815-825`** - `handlePropertyFileInputStreams` now throws instead of returning null, so the `if (isl == null)` guards at both call sites are dead code. Also, `"Failed to resolve the properties file: Not found"` dropped the filename that the old `FileUtil.getFileNotFoundMessage()` message included. Use `FileUtil.getFileNotFoundMessage(propertyFile)` in the throw and delete the guards.

The `DefaultsFileValueProvider` refactor into static factory helpers is clean, and the overlay test asserting `CCC=ccc` beats `CCC=CCC` is exactly the right way to pin the cascade direction.

filipelautert (commented):
Almost there! handlePropertyFileInputStreams never returns null now, so the isl == null guard at AbstractLiquibaseMojo.java:805 is dead code - please drop it. Triggered the CI as it never ran on the last commit.

liquibase-standard/src/main/java/liquibase/configuration/core/DefaultsFileValueProvider.java
```diff
@@ -11,46 +11,70 @@
...
     private final Properties properties;
     private final String sourceDescription;
+    private final int precedence;
 
-    protected DefaultsFileValueProvider(Properties properties) {
+    protected DefaultsFileValueProvider(Properties properties, String sourceDescription, int precedenceOffset) {
         this.properties = properties;
-        sourceDescription = "Passed default properties";
+        this.sourceDescription = sourceDescription;
+        this.precedence = DEFAULT_PRECEDENCE - precedenceOffset;
```
filipelautert: With precedence 50 - offset, the 20th defaults file drops below ServletConfigurationValueProvider (30) and eventually goes negative. Should not we clamp it, or is that many files not worth worrying about?

liquibase-maven-plugin/src/main/java/org/liquibase/maven/plugins/AbstractLiquibaseMojo.java
```diff
@@ -846,9 +852,13 @@ protected void configureChangeLogProperties() throws MojoFailureException, MojoE
         }
     }
 
-    private static InputStream handlePropertyFileInputStream(String propertyFile) throws MojoFailureException {
+    private static InputStreamList handlePropertyFileInputStreams(String propertyFile) throws MojoFailureException {
         try {
-            return Scope.getCurrentScope().getResourceAccessor().getExisting(propertyFile).openInputStream();
+            InputStreamList inputStreamList = Scope.getCurrentScope().getResourceAccessor().openStreams(null, propertyFile);
+            if (inputStreamList == null || inputStreamList.isEmpty()) {
+                throw new MojoFailureException("Failed to resolve the properties file: " + FileUtil.getFileNotFoundMessage(propertyFile));
```
filipelautert: This used to be a MojoExecutionException with FileUtil.getFileNotFoundMessage. Was the change to MojoFailureException intentional? Maven treats the two differently.

filipelautert (commented):
The ChangeSet null check looks unrelated to defaults files, worth splitting out?

Fixed the test on your branch, it was calling registerValueProviders which #7876 replaced.

liquibase-maven-plugin/src/main/java/org/liquibase/maven/plugins/AbstractLiquibaseMojo.java
MichaelKern-IVV: No, that was not intended, reverted back to MojoExecutionException.

liquibase-standard/src/main/java/liquibase/configuration/core/DefaultsFileValueProvider.java
MichaelKern-IVV: I was thinking of two or three files, or four at most. But you never know, so I’ve added a limit.

filipelautert (commented):
The clamp and the revert look good, just that one compile issue, verified locally. CI hasn't run on these commits yet, I'll trigger it after the fix.

liquibase-maven-plugin/src/main/java/org/liquibase/maven/plugins/AbstractLiquibaseMojo.java
```diff
@@ -846,9 +852,13 @@ protected void configureChangeLogProperties() throws MojoFailureException, MojoE
         }
     }
 
-    private static InputStream handlePropertyFileInputStream(String propertyFile) throws MojoFailureException {
+    private static InputStreamList handlePropertyFileInputStreams(String propertyFile) throws MojoFailureException {
         try {
-            return Scope.getCurrentScope().getResourceAccessor().getExisting(propertyFile).openInputStream();
+            InputStreamList inputStreamList = Scope.getCurrentScope().getResourceAccessor().openStreams(null, propertyFile);
+            if (inputStreamList == null || inputStreamList.isEmpty()) {
+                throw new MojoExecutionException("Failed to resolve the properties file: " + FileUtil.getFileNotFoundMessage(propertyFile));
```
filipelautert, line 859: handlePropertyFileInputStreams only declares MojoFailureException, so this new MojoExecutionException doesn't compile (`unreported exception org.apache.maven.plugin.MojoExecutionException`). Could you add it to the throws clause?

MichaelKern-IVV: Oops, that slipped my mind

PR #7725, opened 2026-05-16
Treat 'NULL' default value as null for MySQL/MariaDB enum columns

### LINKED ISSUE #3444: Enum NULL default value become "DEFAULT 'null' NULL"
## Environment

*Liquibase Version**: 4.17.2

**Liquibase Integration & Version**: CLI 4.17.2

**Database Vendor & Version**: MariaDB 10.7.1

**Operating System Type & Version**: Ubuntu 20.04.4 LTS

**Infrastructure Type/Provider**: Docker

## Description

When generating the changelog for all tables containing an enumerator whose value may be NULL, it generates an erroneous value for the column with defaultValue="null" instead of defaultComputedValue="NULL"

As a result, when the update is executed, the migration does not work.

## Steps To Reproduce

- Create database with a similar structure
```sql
CREATE TABLE `test_with_enum` (
  `id` int(11) NOT NULL,
  `enumerateur` enum('yo','lo') DEFAULT NULL
) ENGINE=InnoDB DEFAULT CHARSET=latin1;
```

## Actual Behavior

Actual changeset generated:

```xml
<changeSet author="utilisateur (generated)" id="1667830192184-1">
    <createTable tableName="test_with_enum">
        <column name="sasasa" type="VARCHAR(15)">
        </column>
        <column autoIncrement="true" name="id" type="INT">
        </column>
        <column defaultValue="null" name="enumerateur" type="ENUM('yo', 'lo')"/>
    </createTable>
</changeSet>
```

## Expected/Desired Behavior

Expected changeset generated:

```xml
<changeSet author="utilisateur (generated)" id="1667830192184-1">
    <createTable tableName="test_with_enum">
        <column name="sasasa" type="VARCHAR(15)">
        </column>
        <column autoIncrement="true" name="id" type="INT">
        </column>
        <column defaultValueComputed="NULL" name="enumerateur" type="ENUM('yo', 'lo')"/>
    </createTable>
</changeSet>
```
Thanks for any help or hints that might help solve this behavior with enumerator fields.

Best regards

singhvishalkr:
When I run `generateChangeLog` on a MariaDB table with an enum column that has `DEFAULT NULL`, the output contains `defaultValue="null"` (the string "null") instead of either omitting the default entirely or using `defaultValueComputed="NULL"`.

The root cause is that the MySQL JDBC driver returns the string "NULL" for such columns, and Liquibase was passing it through as a regular string value. Oracle and DB2 already have handling for this case (converting "NULL" strings to actual null), but MySQL/MariaDB did not.

I added the same check for MySQL databases in `ColumnSnapshotGenerator.readDefaultValue()`.

**Testing:** I verified the fix by stepping through the code with a debugger and confirmed that enum columns with DEFAULT NULL now result in a null defaultValue, meaning no `defaultValue` attribute is emitted in the changelog XML.

Fixes #3444

### DIFF
--- a/liquibase-standard/src/main/java/liquibase/snapshot/jvm/ColumnSnapshotGenerator.java
+++ b/liquibase-standard/src/main/java/liquibase/snapshot/jvm/ColumnSnapshotGenerator.java
@@ -573,6 +573,10 @@ public class ColumnSnapshotGenerator extends JdbcSnapshotGenerator {
 
         if (database instanceof MySQLDatabase) {
             readDefaultValueForMysqlDatabase(columnMetadataResultSet, columnInfo, database);
+            if ((columnMetadataResultSet.get(COLUMN_DEF_COL) != null)
+                    && StringUtil.equalsWordNull((String) columnMetadataResultSet.get(COLUMN_DEF_COL))) {
+                columnMetadataResultSet.set(COLUMN_DEF_COL, null);
+            }
         }
 
         if ((database instanceof AbstractDb2Database)
--- a/liquibase-standard/src/main/java/liquibase/snapshot/jvm/ColumnSnapshotGenerator.java
+++ b/liquibase-standard/src/main/java/liquibase/snapshot/jvm/ColumnSnapshotGenerator.java
@@ -573,8 +573,10 @@ public class ColumnSnapshotGenerator extends JdbcSnapshotGenerator {
 
         if (database instanceof MySQLDatabase) {
             readDefaultValueForMysqlDatabase(columnMetadataResultSet, columnInfo, database);
-            if ((columnMetadataResultSet.get(COLUMN_DEF_COL) != null)
-                    && StringUtil.equalsWordNull((String) columnMetadataResultSet.get(COLUMN_DEF_COL))) {
+            Object defaultValue = columnMetadataResultSet.get(COLUMN_DEF_COL);
+            if ((defaultValue instanceof String)
+                    && isMySqlEnumType(columnInfo)
+                    && StringUtil.equalsWordNull((String) defaultValue)) {
                 columnMetadataResultSet.set(COLUMN_DEF_COL, null);
             }
         }
@@ -611,6 +613,13 @@ public class ColumnSnapshotGenerator extends JdbcSnapshotGenerator {
         }
 
         return SqlUtil.parseValue(database, columnMetadataResultSet.get(COLUMN_DEF_COL), columnInfo.getType());
+    }
+
+    private boolean isMySqlEnumType(Column columnInfo) {
+        if ((columnInfo == null) || (columnInfo.getType() == null) || (columnInfo.getType().getTypeName() == null)) {
+            return false;
+        }
+        return columnInfo.getType().getTypeName().toLowerCase(Locale.ENGLISH).startsWith("enum");
     }
 
     private void readDefaultValueForMysqlDatabase(CachedRow columnMetadataResultSet, Column column, Database database) {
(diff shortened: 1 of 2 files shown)


filipelautert (commented):
Hey, reviewed it. The intent is sound — a `DEFAULT NULL` enum surfacing as the literal string `"null"` in the changelog is a real bug. Three things before this is mergeable:

1. **Unguarded cast → ClassCastException risk.** `(String) columnMetadataResultSet.get(COLUMN_DEF_COL)` is cast with only a `!= null` check. After `readDefaultValueForMysqlDatabase`, `COLUMN_DEF` can hold a non-String (e.g. a `DatabaseFunction` or a numeric default), and this would throw at runtime. Guard with `instanceof String`:
   ```java
   Object def = columnMetadataResultSet.get(COLUMN_DEF_COL);
   if (def instanceof String && StringUtil.equalsWordNull((String) def)) {
       columnMetadataResultSet.set(COLUMN_DEF_COL, null);
   }
   ```

2. **Title says "enum columns" but the code applies to every MySQL column.** There's no type check, so a legitimate `VARCHAR ... DEFAULT 'null'` (literal string `'null'`) would also be silently nulled out. Either narrow the guard to enum types, or — if the broad behavior is intended — update the title/description so reviewers know the real scope.

3. **No test.** This changes snapshot output; it needs a `ColumnSnapshotGeneratorTest` case (MySQL/MariaDB enum with `DEFAULT NULL` → expected `null`) so the behavior is locked in and the false-positive boundary in (2) is documented.

singhvishalkr:
Thanks for the review. Pushed `68f487e43` with the three fixes:

- guarded `COLUMN_DEF` with `instanceof String` before checking the word `NULL`
- limited the MySQL/MariaDB normalization to enum column types so a varchar default of `"NULL"` stays a string
- added `ColumnSnapshotGeneratorTest` coverage for enum `DEFAULT NULL`, varchar `"NULL"`, and a non-string enum default

Verified with:

`.\mvnw.cmd -pl liquibase-standard "-Dtest=liquibase.snapshot.jvm.ColumnSnapshotGeneratorTest" test`

filipelautert (approved):
All three addressed + tests, thanks! Retriggering the CI as it never ran on the last commit.

kesleyfort:
# **Verdict: PASS**

## Manual repro (CLI) — before/after evidence

Pre-existing state seeded on both servers (from the issue, plus varchar regression guards):

```sql
CREATE TABLE `test_with_enum` (
  `id` int NOT NULL,
  `enumerateur` enum('yo','lo') DEFAULT NULL,
  `vc_default_null` varchar(20) DEFAULT NULL,
  `vc_literal_null` varchar(20) DEFAULT 'NULL'
) ENGINE=InnoDB;
```

Command (identical for both builds): `liquibase --url jdbc:mariadb://localhost:3308/qa7725 --username root --password *** generate-changelog --changelog-file <out>.xml`

```xml
# BUGGY (baseline 38affb7) — MariaDB 11.4
<column defaultValue="null" name="enumerateur" type="ENUM('yo', 'lo')"/>
```

```xml
# FIXED (PR 42736c8) — MariaDB 11.4
<column name="enumerateur" type="ENUM('yo', 'lo')"/>
```

Deploying each generated changelog to a fresh database (round-trip check):

```
# BUGGY (baseline 38affb7) — update fails, matching the customer report
ERROR: Exception Primary Reason:  (conn=20) Invalid default value for 'enumerateur'
[Failed SQL: (1067) CREATE TABLE qa7725_base.test_with_enum (id INT NOT NULL,
  enumerateur ENUM('yo', 'lo') DEFAULT 'null' NULL, ...)]
```

```
# FIXED (PR 42736c8) — update succeeds; deployed table matches the original
SHOW CREATE TABLE qa7725_fix.test_with_enum → `enumerateur` enum('yo','lo') DEFAULT NULL,
```

MySQL 8.0: baseline and fixed changelogs are identical except the generated changeset id/timestamp — no enum default was ever emitted there (MySQL's `information_schema` returns real SQL NULL), and no regression was introduced.

kesleyfort (approved):

PR #7729, opened 2026-05-18
AbstractPathResourceAccessor: confine path resolution to configured root (CWE-22)

v-petrovych:
## Summary

Adds root-containment checks to `AbstractPathResourceAccessor#getAll` and `#search`, and disables DOCTYPE declarations in `XMLChangeLogSAXParser` when `SECURE_PARSING` is enabled. Together these close two reachable entry points for CWE-22 — one direct via path-taking changelog directives, one indirect via XML entity expansion chaining into the same accessor.

## Why

`AbstractPathResourceAccessor#getAll(path)` and `#search(startPath)` resolve a caller-supplied path against the configured root via `Path.resolve(...)` and then call `Files.exists(...)` / `Files.walkFileTree(...)` without verifying that the resolved path stays inside the root. `standardizePath()` strips a leading `/` and a Windows drive prefix, but `Path.normalize()` on a relative path cannot collapse leading `..` segments — so payloads such as `../../etc/passwd` survive standardisation and reach the OS layer, where `..` is resolved.

`XMLChangeLogSAXParser` previously only set `FEATURE_SECURE_PROCESSING`, which enforces parser resource limits but does not disable DOCTYPE/ENTITY processing. `LiquibaseEntityResolver` blocks remote-URL entity lookups when `SECURE_PARSING=true`, but its local-resolution fallback delegates to `ResourceAccessor.get(...)` — the same path-resolution code path that the first half of this PR confines.

## What changes

1. **`AbstractPathResourceAccessor#getAll(path)`** — after `getRootPath().normalize().resolve(path).normalize()`, reject any path that does not `startsWith(root)`, throwing `IOException` with the original payload and the canonical root. A second layer uses `toRealPath()` to follow symlinks and reject when the canonical location escapes the canonical root.
2. **`AbstractPathResourceAccessor#search(startPath, ...)`** — same syntactic containment plus per-entry `toRealPath()` containment in `preVisitDirectory` and `visitFile`, so symlinks pointing outside the root are skipped during the walk while in-root traversal continues. `startPath` also routes through `standardizePath()` for consistency with `getAll`.
3. **`XMLChangeLogSAXParser.<init>`** — when `SECURE_PARSING=true`, additionally set `http://apache.org/xml/features/disallow-doctype-decl=true`.

## Release-notes callouts

Two behaviour changes worth flagging:

- **DOCTYPE declarations are now rejected** in changelog XML under `SECURE_PARSING=true` (the default). Liquibase validates against XSD, not DTD; the change has no impact on canonical changelogs but third-party fixtures that happen to carry a `<!DOCTYPE …>` preamble will start failing. A tree-wide grep across `liquibase-standard/src/test/` finds no legitimate usage.
- **`CompositeResourceAccessor.getAll(path)` is now fail-fast on the first child rejection.** Previously a multi-root search-path returned the union of resolutions; if any root now rejects an input as escaping, the composite call throws. Fail-closed is the right default for a security control, but multi-root callers should be aware.

## Test plan

- New `DirectoryResourceAccessorTest` cases: 5 traversal payloads (`../`, `../../`, `./../../`, `subdir/../../escape`, `..\..\..\` mixed separators); positive case for a path that traverses via `..` but stays under the root; `search()` rejection; symlink-escape rejection (skipped on Windows where `Files.createSymbolicLink` requires `SeCreateSymbolicLinkPrivilege`).
- New `SearchPathResourceAccessorTest` case: confirms the composite wrapper does not bypass per-accessor containment.
- New `XMLChangeLogSAXParserTest` case: DOCTYPE-bearing changelog is rejected; assertion checks for `DOCTYPE` and `disallow` keywords.
- Existing `XMLChangeLogSAXParserTest` "uses liquibase.secureParsing by default" updated to assert the new (stronger) JAXP DOCTYPE-disallowed error.
- `INVALID_XML` fixture stripped of its DOCTYPE preamble so the XSD-validation test continues to fire correctly on the bogus element.
- Local run on `liquibase-standard`: 119 tests, 0 failures, 1 skipped (the Windows-conditional symlink test).

## Caveats noted for reviewers

- The `startsWith` check is purely syntactic; the additional `toRealPath()` check follows symlinks at check time. There is a TOCTOU window between the check and any subsequent stream open via the returned `Resource` — a symlink modified after the check is not re-validated. The class javadoc documents this.
- `ZipResourceAccessor` (zipfs) is intentionally unaffected by the new check: zipfs `normalize()` already clamps `..` at root, so the `startsWith` check is a no-op for that subclass.

## Reference

- CWE-22 — Improper Limitation of a Pathname to a Restricted Directory: <https://cwe.mitre.org/data/definitions/22.html>
- The DOCTYPE hardening is the standard JAXP defence in depth on top of `FEATURE_SECURE_PROCESSING` for SAX parsers reading untrusted XML.

### DIFF
--- a/liquibase-standard/src/main/java/liquibase/parser/core/xml/XMLChangeLogSAXParser.java
+++ b/liquibase-standard/src/main/java/liquibase/parser/core/xml/XMLChangeLogSAXParser.java
@@ -48,6 +48,13 @@ public class XMLChangeLogSAXParser extends AbstractChangeLogParser {
             } catch (Throwable e) {
                 Scope.getCurrentScope().getLog(getClass()).fine("Cannot enable FEATURE_SECURE_PROCESSING: " + e.getMessage(), e);
             }
+            try {
+                // Changelogs are validated against XSD, not DTD. DOCTYPE has no legitimate use here;
+                // disabling it eliminates XXE entity-expansion as an attack surface.
+                saxParserFactory.setFeature("http://apache.org/xml/features/disallow-doctype-decl", true);
+            } catch (Throwable e) {
+                Scope.getCurrentScope().getLog(getClass()).fine("Cannot enable disallow-doctype-decl: " + e.getMessage(), e);
+            }
         }
     }
 
--- a/liquibase-standard/src/main/java/liquibase/resource/AbstractPathResourceAccessor.java
+++ b/liquibase-standard/src/main/java/liquibase/resource/AbstractPathResourceAccessor.java
@@ -36,8 +36,23 @@ public abstract class AbstractPathResourceAccessor extends AbstractResourceAcces
         if (path == null) {
             return returnList;
         }
-        Path finalPath = getRootPath().resolve(path);
+        Path finalPath = getRootPath().resolve(path).normalize();
+        // Reject any payload that, after normalization, escapes the configured root.
+        // Java's Path.normalize() on a relative path leaves leading ".." segments intact,
+        // so a payload like "../../etc/passwd" survives standardizePath() and only
+        // collapses after resolve(); a startsWith check catches it before Files.exists
+        // would resolve it through the OS layer.
+        if (!finalPath.startsWith(getRootPath().normalize())) {
+            throw new IOException("Path '" + path + "' resolves outside accessor root '"
+                    + getRootPath().normalize() + "'");
+        }
         if (Files.exists(finalPath)) {
+            // Defense in depth: a symbolic link at finalPath could point outside the
+            // configured root. toRealPath() follows symlinks; reject if the canonical
+            // location escapes the canonical root.
+            if (!finalPath.toRealPath().startsWith(getRootPath().toRealPath())) {
+                throw new IOException("Path '" + path + "' resolves outside accessor root via symlink");
+            }
             returnList.add(createResource(finalPath, path));
         } else {
             log.fine("Path " + path + " in " + getRootPath() + " does not exist (" + this + ")");
@@ -85,7 +100,11 @@ public abstract class AbstractPathResourceAccessor extends AbstractResourceAcces
         Logger log = Scope.getCurrentScope().getLog(getClass());
 
         Path rootPath = getRootPath();
-        Path basePath = rootPath.resolve(startPath);
+        Path basePath = rootPath.resolve(startPath).normalize();
+        if (!basePath.startsWith(rootPath.normalize())) {
+            throw new IOException("Search startPath '" + startPath
+                    + "' resolves outside accessor root '" + rootPath.normalize() + "'");
+        }
 
         final List<Resource> returnSet = new ArrayList<>();
         if (!Files.exists(basePath)) {
@@ -97,9 +116,30 @@ public abstract class AbstractPathResourceAccessor extends AbstractResourceAcces
             throw new IOException("'" + startPath + "' is a file, not a directory");
         }
 
+        // Cache the canonical root once so the per-entry containment check below
+        // doesn't re-resolve it on every visited file/directory.
+        final Path rootRealPath = rootPath.toRealPath();
+
         SimpleFileVisitor<Path> fileVisitor = new SimpleFileVisitor<Path>() {
             @Override
+            public FileVisitResult preVisitDirectory(Path dir, BasicFileAttributes attrs) throws IOException {
+                // walkFileTree is invoked with FOLLOW_LINKS so subdirectories may be
+                // reached via symlinks. Skip any subtree whose canonical location
+                // escapes the canonical root.
+                if (!dir.toRealPath().startsWith(rootRealPath)) {
+                    log.fine("Skipping directory '" + dir + "' that resolves outside accessor root");
+                    return FileVisitResult.SKIP_SUBTREE;
+                }
+                return FileVisitResult.CONTINUE;
+            }
+
+            @Override
             public FileVisitResult visitFile(Path file, BasicFileAttributes attrs) throws IOException {
+                // Per-file symlink check (e.g. file-level link pointing outside the root).
+                if (!file.toRealPath().startsWith(rootRealPath)) {
+                    log.fine("Skipping file '" + file + "' that resolves outside accessor root");
+                    return FileVisitResult.CONTINUE;
+                }
                 if (attrs.isRegularFile() && meetsSearchCriteria(file)) {
                     addToReturnList(file);
                 }
--- a/liquibase-standard/src/main/java/liquibase/resource/AbstractPathResourceAccessor.java
+++ b/liquibase-standard/src/main/java/liquibase/resource/AbstractPathResourceAccessor.java
@@ -9,6 +9,26 @@ import java.nio.file.*;
 import java.nio.file.attribute.BasicFileAttributes;
 import java.util.*;
 
+/**
+ * Base class for {@link ResourceAccessor}s that resolve a caller-supplied path against a
+ * configured root directory (file system, zip, etc.).
+ * <p>
+ * <b>Containment guarantee.</b> Both {@link #getAll(String)} and
+ * {@link #search(String, SearchOptions)} reject any path that, after normalisation, escapes
+ * the configured {@link #getRootPath()}. They additionally verify the canonical real path
+ * (after symbolic-link resolution) stays within the canonical root, so a symlink located
+ * inside the root that targets a file outside it is not silently followed.
+ * <p>
+ * <b>Caveats.</b>
+ * <ul>
+ *   <li>The containment check is evaluated at {@code getAll} / {@code search} time. A
+ *       symlink created or modified after the check (TOCTOU race) is not re-validated
+ *       when the returned {@link Resource}'s stream is opened later.</li>
+ *   <li>For {@link #search(String, SearchOptions)}, each visited file or directory is
+ *       re-checked individually so symlinks discovered during the walk are skipped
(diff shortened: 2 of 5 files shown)


wwillard7800 (commented):
## Overall

Solid, well-scoped security fix. Two reachable CWE-22 entry points get closed with minimal surface area, tests are comprehensive, and the author has called out the known limitations (symlinks) and breaking-change risk (DOCTYPE) up front. Approving in spirit with the notes below — none are blockers.

## Findings

### 1. `search()` does not run `standardizePath()` — minor inconsistency with `getAll()`

`AbstractPathResourceAccessor.java:34` (`getAll`) calls `standardizePath(path)`, which collapses `\` → `/` via `new File(path).toPath().normalize()`. `search()` (lines 82–84) only strips `file:/+` and a leading `/`.

Consequence on Linux: a Windows-style mixed-separator payload like `..\..\..\etc\passwd` passed to `search()` is treated as a literal filename — `Path.normalize()` does not recognize `\` as a separator on POSIX, so the path never collapses, never escapes the root, and silently returns empty instead of being rejected. Fail-closed, so not a security gap — but it means the `..\..\..\` test in `DirectoryResourceAccessorTest` would behave differently if mirrored for `search()`. Worth either normalizing the two methods or noting why they diverge.

### 2. `getRootPath().normalize()` recomputed multiple times per call

`AbstractPathResourceAccessor.java:39, 45, 47` (getAll) and `:88, 90, 91` (search). For `DirectoryResourceAccessor`, the field is already absolute+normalized at construction (`DirectoryResourceAccessor.java:30`), so each `normalize()` is a no-op — but the call is still made three times in the error/check path. Minor — extract once into a local for readability:

```java
Path root = getRootPath().normalize();
Path finalPath = root.resolve(path).normalize();
if (!finalPath.startsWith(root)) { ... }
```

### 3. `CompositeResourceAccessor.getAll()` is fail-fast on the first rejection

`CompositeResourceAccessor.java:68-74` does not catch `IOException`. Once any one accessor rejects a `..` payload, the whole composite call throws and subsequent accessors are never tried. The `SearchPathResourceAccessorTest` "NV-1 regression" pins exactly this behavior, so it's intentional — but it's a behavior change worth flagging: a multi-root search path that previously returned the union of valid resolutions will now hard-fail if any one root rejects the input as escaping. Fail-closed is the right default for a security control, just worth mentioning in release notes alongside the DOCTYPE call-out.

### 4. Symlink caveat is acknowledged — should land in javadoc, not just the PR description

`search()` still calls `Files.walkFileTree(basePath, Collections.singleton(FileVisitOption.FOLLOW_LINKS), ...)` at line 139. The PR description says "worth documenting as a known limitation in the accessor javadoc as a follow-up" — would prefer to see that javadoc note land **in this PR** on `AbstractPathResourceAccessor`, since the new throw makes readers reasonably assume containment is complete.

### 5. `ZipResourceAccessor` interaction — works, worth a sanity-check sentence

`ZipResourceAccessor.getRootPath()` returns `fileSystem.getPath(\"/\")` (jdk.zipfs). The zipfs `normalize()` collapses `..` segments and clamps at `/`, so any `../...` payload resolves under root and the new check is effectively a no-op. Not a bug — but the PR description frames the check as if `DirectoryResourceAccessor` were the only subclass. A one-liner clarifying that zipfs is intentionally unaffected (because zipfs path semantics already prevent escape) would help future readers.

### 6. `getAll(null)` defensive branch is now unreachable

Pre-existing, not introduced by this PR. Line 28-30 throws on null, then lines 36-38 have another null check that can never fire because `standardizePath()` cannot return null. Out of scope for this fix — just noting.

### 7. Test fixture comment is excellent

The block comment in `XMLChangeLogSAXParserTest.groovy` explaining *why* the DOCTYPE preamble had to be stripped from `INVALID_XML` (otherwise the DOCTYPE rejection masks the XSD assertion) is exactly the kind of comment that pays off six months from now. Keep this style.

### 8. JAXP-error-wording assertion is appropriately loose

`e.message.contains(\"DOCTYPE\") && e.message.contains(\"disallow\")` — robust against minor JDK message-text changes. Good call.

### 9. Error message includes original payload

`\"Path '\" + path + \"' resolves outside accessor root '\" + getRootPath().normalize() + \"'\"` — useful for debugging, doesn't leak anything that wasn't user-supplied. Fine.

## Recommendations before merge

1. Apply `standardizePath()` (or equivalent separator normalization) in `search()` for consistency, OR add a comment explaining the divergence.
2. Hoist `getRootPath().normalize()` into a local variable in both methods.
3. Add a javadoc note on `AbstractPathResourceAccessor` covering the symlink limitation.
4. Confirm release notes will call out both (a) the DOCTYPE rejection and (b) the composite-accessor fail-fast change.

v-petrovych:
@wwillard7800 thanks for the thorough review! Point-by-point disposition:

| # | Disposition | Where |
|---|---|---|
| **1** — `search()` doesn't run `standardizePath()` | ✅ Applied | `2e7d76f65` — `search(startPath)` now routes through `standardizePath()` after the existing `file:/+` strip, matching `getAll(path)` behaviour. Windows-style mixed-separator payloads to `search()` on Linux are now recognised as traversal. |
| **2** — `getRootPath().normalize()` recomputed | ✅ Applied | `2e7d76f65` — hoisted into a `rootPath` local in both `#getAll` and `#search`. |
| **3** — Composite-accessor fail-fast | 📝 Acknowledged | Added to "Release-notes callouts" section in the PR body — multi-root callers should be aware. |
| **4** — Symlink caveat → javadoc | ✅ Applied | `2e7d76f65` — class-level javadoc on `AbstractPathResourceAccessor` now documents both the containment guarantee (normalisation + `toRealPath()` check) and the TOCTOU caveat. The earlier `4c2d6185b` commit also closed the symlink gap itself: per-entry `toRealPath()` checks added in `preVisitDirectory`, `visitFile`, and `getAll`. |
| **5** — `ZipResourceAccessor` interaction | 📝 Acknowledged | Added to "Caveats noted for reviewers" in PR body — zipfs `normalize()` already clamps `..` at root, so the new check is a no-op for that subclass. |
| **6** — Unreachable `getAll(null)` defensive branch | Skipping per your note | Pre-existing, out of scope for this fix. |
| **7–9** | 🙏 Thanks for the positive notes | The `INVALID_XML` comment style, the loose JAXP assertion, and the original-payload-in-error are all kept. |

Recommendations summary: 1, 2, 3 applied in `2e7d76f65`; 4 (release-notes callouts) added to PR body.

v-petrovych:
@coderabbitai thanks. Disposition of the 3 actionable comments:

- **`AbstractPathResourceAccessor.java` — symlink check in `visitFile()` under `FOLLOW_LINKS`**: ✅ Applied in `4c2d6185b`. `toRealPath()` containment now runs in `preVisitDirectory` (skips a subtree whose canonical location escapes the canonical root), in `visitFile` (skips individual symlinks pointing outside), and in `#getAll` for the single-path case.

- **`XMLChangeLogSAXParserTest.groovy` — tighten DOCTYPE-rejection assertion**: ✅ Applied in `4c2d6185b`. The test now asserts the error message contains both `DOCTYPE` and `disallow` keywords, matching the pattern of the existing `INSECURE_XML` test.

- **`XMLChangeLogSAXParser.java` — fail-fast when `disallow-doctype-decl` cannot be enabled**: ⏸ Keeping the `try { … } catch (Throwable e) { log.fine(…) }` pattern for consistency with the adjacent `FEATURE_SECURE_PROCESSING` block above it. Changing only the new feature to throw an unchecked exception creates an inconsistency between two adjacent JAXP feature toggles. Happy to land both in a separate hardening PR if the maintainers agree — would be a small focused change.

wwillard7800 (approved):
Thanks for the fast turnaround — all four points addressed, and #3 went further than I asked.

## How the follow-up commit lands against the prior review

- **`standardizePath()` in `search()`** ✅ — `AbstractPathResourceAccessor.java:124` strips the `file:` scheme first, then routes `startPath` through the same separator/drive normalization as `getAll()`. Mixed-separator payloads now collapse symmetrically on Linux.
- **Hoisted `normalize()`** ✅ — Done in both methods, plus `rootRealPath` cached once before the file visitor so the per-entry check doesn't re-resolve the root.
- **Symlink caveat** ✅✅ — Documented in the new class-level javadoc *and* closed by the implementation: `getAll()` rejects via `finalPath.toRealPath().startsWith(rootPath.toRealPath())`, and `search()` does per-entry `toRealPath()` checks in `preVisitDirectory` (`SKIP_SUBTREE`) and `visitFile` (`CONTINUE` without adding). The new `getAll rejects a symlink that escapes the accessor root` test pins the behaviour, correctly `@IgnoreIf({ os.windows })`'d.
- **Release-notes asks** — outside the code; trusting these will land in the changelog.

## One small follow-up suggestion (non-blocking)

The `search()` symlink-skipping path is implemented but not directly pinned by a test — the `getAll` symlink test covers the syntactic peer but not the walk-time skip behaviour (`preVisitDirectory` returning `SKIP_SUBTREE`, `visitFile` returning `CONTINUE` without adding). A single Spock case that creates an in-root symlink to an outside directory and asserts `search()` returns the legitimate in-root files while omitting the symlink target would close the loop. Happy to leave this for a follow-up PR — not gating approval.

## Cost note (FYI, not a request)

`toRealPath()` per directory + per file is a stat syscall per entry. Invisible for normal changelog discovery (tens to hundreds of files); just worth keeping in mind if someone later roots a search at a pathologically large tree.

Approving. Nice work on the symlink closure.

PR #7736, opened 2026-05-20
CWE-200: redact credential-bearing values stored in IntegrationDetails parameters

v-petrovych:
## Impact

## Description

Redact credential-bearing parameter values at the boundary into `IntegrationDetails.parameters`. Producers — the CLI argument parser (`Main.java:185-216`), the defaults-file loader (`Main.java:1022`), and the Maven plugin Mojos (`AbstractLiquibaseMojo.java:541`) — currently call `setParameter` with raw passwords and JDBC URLs containing credentials. The `IntegrationDetails` object is attached to `Scope.Attr.integrationDetails` and is reachable across the JVM for the run; any current or future telemetry / report / MDC sink that serializes the parameters map would otherwise leak those credentials.

This addresses two High findings from a May-2026 OSS credential-handling audit (CWE-200: Exposure of Sensitive Information to an Unauthorized Actor):

- **Maven plugin variant** — `AbstractLiquibaseMojo.java:541` captures every Mojo parameter (`password`, `referencePassword`, full `url`) without filtering. The existing log-side mask in `AbstractLiquibaseMojo.printSettings()` (lines 893-894) is a separate code path and does not cover the `IntegrationDetails` sink.
- **Legacy CLI variant** — `Main.java:185-216` copies every `--key=value` CLI argument into `IntegrationDetails` via `setParameter("argument__"+key, value)`, and `Main.java:1022` similarly copies every entry from the `liquibase.properties` defaults file (prefix `defaultsFile__`). Both paths capture `password` cleartext. The newer `LiquibaseCommandLine.java` (lines 792-794) does not have this issue — finding is legacy-CLI-only, but legacy-CLI is still shipped.

### Threat model

This is **defense in depth**, not active leak remediation. A repo-wide grep confirms no production code path currently reads `IntegrationDetails.getParameters()` — the only sink reads `getName()`. However:

- `Scope.Attr.integrationDetails` is reachable from any thread for the run.
- `LiquibaseAnalyticsListener.sendEvent` YAML-serializes `Event` with `BeanAccess.FIELD`; if a `parameters` field were ever added to `Event`, the map contents would be serialized into the analytics payload without further code change.
- The existing log-side mask in `AbstractLiquibaseMojo.printSettings()` is a separate code path and does not cover the integration-details map.

Redacting at the capture point closes the latent gap without touching every producer.

## Things to be aware of

**Single boundary fix in `IntegrationDetails.setParameter`** — adds a `sanitize(key, value)` helper. All existing producers stay unchanged.

Rules applied in order:

1. If the lowercase key (with `argument__` / `maven__` / `defaultsFile__` prefix stripped) contains one of `password`, `passwd`, `secret`, `token`, `apikey`, `accesskey` → store `*****` via `ConfigurationValueObfuscator.STANDARD`.
2. Else if the value starts with `jdbc:` → pass through `ConfigurationValueObfuscator.URL_OBFUSCATOR` (delegates to `JdbcConnection.sanitizeUrl`). Catches any producer that puts a JDBC URL under a non-`url` key name (current and future).
3. Else passthrough unchanged.

Reuses the existing `ConfigurationValueObfuscator` abstractions; no new redaction primitives introduced.

**Alternatives considered and rejected:**

- **Callsite redaction** (the original audit suggestion). Requires patching every producer site (CLI `Main.java`, defaults-file loader, both Maven Mojos — including a downstream commercial Mojo that mirrors the loop). Future producers would be unprotected.
- **Sink-side redaction in `Event.java`**. Only fixes the current sink; the point of this fix is to prevent future telemetry/report sinks from inheriting the leak.
- **Typed `Credential` / `SecretString` wrapper.** Correct long-term direction (would also subsume the deferred "passwords stored as `String`, never zeroed" item) but a wide refactor across the CLI, Mojos, defaults-file loader, picocli binding, and downstream connection code. Best as a follow-up spike, not in this PR.

## Things to worry about

- **Denylist tokens are intentionally conservative**: `password`, `passwd`, `secret`, `token`, `apikey`, `accesskey`. If reviewers see commonly-used credential-bearing key names that don't match these tokens (e.g., `clientsecret`, `oauth`, `bearer`, `jwt`), please flag — easy to add.
- **`JdbcConnection.sanitizeUrl` coverage gaps inherited**. The value-side `jdbc:` rule reuses `sanitizeUrl`, which has known coverage gaps for some URL formats. A discovered example: PostgreSQL `user:pass@host` userinfo style is not currently sanitized by either variant (MySQL/MariaDB/Oracle userinfo styles are). My tests use formats `sanitizeUrl` covers; the gap is real and is being tracked separately. Any future improvement to `sanitizeUrl` (e.g., adding a postgresql userinfo pattern) automatically benefits this fix.
- **LTS backports** — `main` only here. If older release lines should also receive this, happy to follow up.

## Additional Context

**Test coverage** in new `liquibase-standard/src/test/groovy/liquibase/integration/IntegrationDetailsTest.groovy`:

- Credential-token matching across all 3 producer prefixes + no-prefix, all 6 tokens, case variants (`PASSWORD`, `Passwd`).
- JDBC URL value-side sanitization across `argument__url` / `argument__referenceUrl` / `maven__url` / `defaultsFile__url` / `argument__connectionString` / `any-future-key`, exercising query-string, property-string, and userinfo URL forms covered by `sanitizeUrl` (postgresql `?password=`, snowflake `?password=`, mysql `?password=`, jtds sqlserver `;password=`, mysql/mariadb `user:pass@host`).
- Non-credential keys pass through unchanged (`username`, `databaseUser` — proves no over-masking).
- Null key / null value / null-key-with-jdbc-value.
- Overwrite semantics (still redacted on second `setParameter` for the same key).
- `name` field unaffected by the sanitization path.

**Local verification**: `mvn test -pl liquibase-standard -Dtest=IntegrationDetailsTest` — 36 tests, all green.

### DIFF
--- a/liquibase-standard/src/main/java/liquibase/integration/IntegrationDetails.java
+++ b/liquibase-standard/src/main/java/liquibase/integration/IntegrationDetails.java
@@ -1,12 +1,15 @@
 package liquibase.integration;
 
+import liquibase.configuration.ConfigurationValueObfuscator;
 import lombok.AllArgsConstructor;
 import lombok.Getter;
 import lombok.NoArgsConstructor;
 import lombok.Setter;
 
 import java.util.HashMap;
+import java.util.Locale;
 import java.util.Map;
+import java.util.Set;
 
 /**
  * Stores information about the integration running Liquibase.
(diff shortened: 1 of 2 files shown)


v-petrovych:
Continued at #7739. This PR was auto-closed by GitHub when the branch-rename API was invoked: renaming the head ref deletes the old branch, which fires `head_ref_deleted` and triggers auto-close. The new PR carries the identical diff on a clean commit and an aligned branch name (`cwe-200-redact-integration-details-parameters`).

PR #7739, opened 2026-05-20
CWE-200: redact credential-bearing values stored in IntegrationDetails parameters

v-petrovych:
> **Supersedes #7736.** That PR was auto-closed when GitHub's branch-rename API deleted the old head ref (`head_ref_deleted` event fires `closed`). Same diff, same author, single clean commit. The branch was also renamed from `dat-23008-23009-…` to `cwe-200-redact-integration-details-parameters` so the head name matches the convention used by sibling PRs #7737 / #7738.

## Impact

## Description

Redact credential-bearing parameter values at the boundary into `IntegrationDetails.parameters`. Producers — the CLI argument parser (`Main.java:185-216`), the defaults-file loader (`Main.java:1022`), and the Maven plugin Mojos (`AbstractLiquibaseMojo.java:541`) — currently call `setParameter` with raw passwords and JDBC URLs containing credentials. The `IntegrationDetails` object is attached to `Scope.Attr.integrationDetails` and is reachable across the JVM for the run; any current or future telemetry / report / MDC sink that serializes the parameters map would otherwise leak those credentials.

This addresses two High findings from a May-2026 OSS credential-handling audit (CWE-200: Exposure of Sensitive Information to an Unauthorized Actor):

- **Maven plugin variant** — `AbstractLiquibaseMojo.java:541` captures every Mojo parameter (`password`, `referencePassword`, full `url`) without filtering. The existing log-side mask in `AbstractLiquibaseMojo.printSettings()` (lines 893-894) is a separate code path and does not cover the `IntegrationDetails` sink.
- **Legacy CLI variant** — `Main.java:185-216` copies every `--key=value` CLI argument into `IntegrationDetails` via `setParameter("argument__"+key, value)`, and `Main.java:1022` similarly copies every entry from the `liquibase.properties` defaults file (prefix `defaultsFile__`). Both paths capture `password` cleartext. The newer `LiquibaseCommandLine.java` (lines 792-794) does not have this issue — finding is legacy-CLI-only, but legacy-CLI is still shipped.

### Threat model

This is **defense in depth**, not active leak remediation. A repo-wide grep confirms no production code path currently reads `IntegrationDetails.getParameters()` — the only sink reads `getName()`. However:

- `Scope.Attr.integrationDetails` is reachable from any thread for the run.
- `LiquibaseAnalyticsListener.sendEvent` YAML-serializes `Event` with `BeanAccess.FIELD`; if a `parameters` field were ever added to `Event`, the map contents would be serialized into the analytics payload without further code change.
- The existing log-side mask in `AbstractLiquibaseMojo.printSettings()` is a separate code path and does not cover the integration-details map.

Redacting at the capture point closes the latent gap without touching every producer.

## Things to be aware of

**Single boundary fix in `IntegrationDetails.setParameter`** — adds a `sanitize(key, value)` helper. All existing producers stay unchanged.

Rules applied in order:

1. If the lowercase key (with `argument__` / `maven__` / `defaultsFile__` prefix stripped) contains one of `password`, `passwd`, `secret`, `token`, `apikey`, `accesskey` → store `*****` via `ConfigurationValueObfuscator.STANDARD`.
2. Else if the value starts with `jdbc:` → pass through `ConfigurationValueObfuscator.URL_OBFUSCATOR` (delegates to `JdbcConnection.sanitizeUrl`). Catches any producer that puts a JDBC URL under a non-`url` key name (current and future).
3. Else passthrough unchanged.

Reuses the existing `ConfigurationValueObfuscator` abstractions; no new redaction primitives introduced.

**Alternatives considered and rejected:**

- **Callsite redaction** (the original audit suggestion). Requires patching every producer site (CLI `Main.java`, defaults-file loader, both Maven Mojos — including a downstream commercial Mojo that mirrors the loop). Future producers would be unprotected.
- **Sink-side redaction in `Event.java`**. Only fixes the current sink; the point of this fix is to prevent future telemetry/report sinks from inheriting the leak.
- **Typed `Credential` / `SecretString` wrapper.** Correct long-term direction (would also subsume the deferred "passwords stored as `String`, never zeroed" item) but a wide refactor across the CLI, Mojos, defaults-file loader, picocli binding, and downstream connection code. Best as a follow-up spike, not in this PR.

## Things to worry about

- **Denylist tokens are intentionally conservative**: `password`, `passwd`, `secret`, `token`, `apikey`, `accesskey`. If reviewers see commonly-used credential-bearing key names that don't match these tokens (e.g., `clientsecret`, `oauth`, `bearer`, `jwt`), please flag — easy to add.
- **`JdbcConnection.sanitizeUrl` coverage gaps inherited**. The value-side `jdbc:` rule reuses `sanitizeUrl`, which has known coverage gaps for some URL formats. A discovered example: PostgreSQL `user:pass@host` userinfo style is not currently sanitized by either variant (MySQL/MariaDB/Oracle userinfo styles are). My tests use formats `sanitizeUrl` covers; the gap is real and is being tracked separately. Any future improvement to `sanitizeUrl` (e.g., adding a postgresql userinfo pattern) automatically benefits this fix.
- **LTS backports** — `main` only here. If older release lines should also receive this, happy to follow up.

## Additional Context

**Test coverage** in new `liquibase-standard/src/test/groovy/liquibase/integration/IntegrationDetailsTest.groovy`:

- Credential-token matching across all 3 producer prefixes + no-prefix, all 6 tokens, case variants (`PASSWORD`, `Passwd`).
- JDBC URL value-side sanitization across `argument__url` / `argument__referenceUrl` / `maven__url` / `defaultsFile__url` / `argument__connectionString` / `any-future-key`, exercising query-string, property-string, and userinfo URL forms covered by `sanitizeUrl` (postgresql `?password=`, snowflake `?password=`, mysql `?password=`, jtds sqlserver `;password=`, mysql/mariadb `user:pass@host`).
- Non-credential keys pass through unchanged (`username`, `databaseUser` — proves no over-masking).
- Null key / null value / null-key-with-jdbc-value.
- Overwrite semantics (still redacted on second `setParameter` for the same key).
- `name` field unaffected by the sanitization path.

**Local verification**: `mvn test -pl liquibase-standard -Dtest=IntegrationDetailsTest` — 36 tests, all green.

### DIFF
--- a/liquibase-standard/src/main/java/liquibase/integration/IntegrationDetails.java
+++ b/liquibase-standard/src/main/java/liquibase/integration/IntegrationDetails.java
@@ -1,11 +1,13 @@
 package liquibase.integration;
 
 import liquibase.configuration.ConfigurationValueObfuscator;
+import lombok.AccessLevel;
 import lombok.AllArgsConstructor;
 import lombok.Getter;
 import lombok.NoArgsConstructor;
 import lombok.Setter;
 
+import java.util.Collections;
 import java.util.HashMap;
 import java.util.Locale;
 import java.util.Map;
@@ -25,6 +27,15 @@ public class IntegrationDetails {
      */
     private String name;
 
+    /**
+     * Backing storage for parameter key/value pairs. {@code final} so the reference cannot be
+     * swapped, and the Lombok-generated getter is suppressed via {@link AccessLevel#NONE}
+     * because {@link #getParameters()} below returns an unmodifiable view — any caller that
+     * tries to write directly into the map (e.g. {@code details.getParameters().put(key, raw)})
+     * must instead go through {@link #setParameter(String, String)} so the {@link #sanitize}
+     * boundary cannot be bypassed (CWE-200 defense in depth).
+     */
+    @Getter(AccessLevel.NONE)
     private final Map<String, String> parameters = new HashMap<>();
 
     /**
@@ -36,10 +47,26 @@ public class IntegrationDetails {
     /**
      * Lowercase substrings that mark a parameter key as carrying a credential value. Conservative
      * list matched against the lowercased key (with producer prefixes stripped).
+     * <p>
+     * <b>Matching is intentionally substring-based</b> so that keys like {@code "argument__bearerToken"}
+     * or {@code "authToken"} are caught by {@code "token"}, {@code "awsCredentials"} by
+     * {@code "credentials"}, and {@code "liquibaseProLicenseKey"}-style keys by their respective
+     * tokens. Future readers should NOT tighten this to exact-match without re-evaluating
+     * coverage — over-masking is intentionally preferred to under-masking here because the
+     * downstream sinks (MDC, telemetry, logs) treat redacted values as opaque anyway.
      */
     private static final Set<String> CREDENTIAL_KEY_TOKENS = Set.of(
-            "password", "passwd", "secret", "token", "apikey", "accesskey"
+            "password", "passwd", "secret", "token", "apikey", "accesskey", "credentials"
     );
+
+    /**
+     * Returns an unmodifiable view of the parameters map. Callers must use
+     * {@link #setParameter(String, String)} to add or replace entries so the
+     * {@link #sanitize} boundary cannot be bypassed (CWE-200 defense in depth).
+     */
+    public Map<String, String> getParameters() {
+        return Collections.unmodifiableMap(parameters);
+    }
 
     public void setParameter(String key, String value) {
         this.parameters.put(key, sanitize(key, value));
--- a/liquibase-standard/src/main/java/liquibase/integration/IntegrationDetails.java
+++ b/liquibase-standard/src/main/java/liquibase/integration/IntegrationDetails.java
@@ -1,12 +1,15 @@
 package liquibase.integration;
 
+import liquibase.configuration.ConfigurationValueObfuscator;
 import lombok.AllArgsConstructor;
 import lombok.Getter;
 import lombok.NoArgsConstructor;
 import lombok.Setter;
 
 import java.util.HashMap;
+import java.util.Locale;
 import java.util.Map;
+import java.util.Set;
 
 /**
  * Stores information about the integration running Liquibase.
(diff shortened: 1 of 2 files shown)


filipelautert (approved):
Hey, reviewed this. Boundary sanitization in `setParameter` is the right shape — fixing one entry point beats patching every producer (CLI, defaults loader, Maven Mojo, plus the downstream commercial Mojo). Verified the prefix list is complete: `argument__` (`Main.java:193, 200`), `defaultsFile__` (`Main.java:1022`), `maven__` (`AbstractLiquibaseMojo.java:541`) — those are the only producer call sites today.

A couple of things:

- `IntegrationDetails.java:14` — `@Getter` on the class exposes `getParameters()` returning the live map, so a future caller could do `details.getParameters().put("password", raw)` and skip `sanitize()` entirely. No callers do that today (grep confirms), so it's defense-in-depth, not a current bug — but consider either returning `Collections.unmodifiableMap` from a hand-rolled getter, or `@Getter(AccessLevel.NONE)` on the field plus a `getParameters()` that wraps. Future-proofs the boundary guarantee.

- The `token` token matches by substring, so it will also catch `tokenize` / `tokenCount` / `tokenizer` if a future producer ever uses such a key. Almost certainly fine in this domain (over-masking is safe), but worth a brief comment in `CREDENTIAL_KEY_TOKENS` noting the substring-match intent so a later reader doesn't tighten it to exact-match and regress coverage.

- CI "failures" here are only `claude-review` and `label` — same ancillary workflows as #7740. Build/test/integration matrix is green.

Otherwise solid. Test coverage across the 3 producer prefixes + no-prefix + case variants + null handling + overwrite semantics is thorough.

v-petrovych:
Thanks @filipelautert and @coderabbitai for the reviews. Both flagged the same `getParameters()` boundary issue from slightly different angles; both fixes plus the conservative `"credentials"` addition are in commit `259f635f7`.

**1. Unmodifiable `getParameters()` view** *(Filipe finding #1 + CodeRabbit nitpick #2)*

Filipe's framing — *"a future caller could do `details.getParameters().put("password", raw)` and skip `sanitize()` entirely. No callers do that today (grep confirms), so it's defense-in-depth, not a current bug."* — was exactly right. Suppressed the Lombok-generated getter on the field with `@Getter(AccessLevel.NONE)` and added a hand-rolled `getParameters()` returning `Collections.unmodifiableMap(parameters)`:

```java
@Getter(AccessLevel.NONE)
private final Map<String, String> parameters = new HashMap<>();

public Map<String, String> getParameters() {
    return Collections.unmodifiableMap(parameters);
}
```

I re-verified no current caller mutates the map via the getter (only reads), so this is safe to land. New regression test `getParameters returns an unmodifiable view that cannot bypass the sanitize boundary` asserts mutation throws `UnsupportedOperationException`; another spec confirms existing read patterns (`details.parameters[key]`, `containsKey`, `size`) still work.

**2. `CREDENTIAL_KEY_TOKENS` substring-match documentation** *(Filipe finding #2)*

Added an explicit comment on the constant capturing your point — substring match is intentional, future readers should NOT tighten to exact-match without re-evaluating coverage, over-masking is intentionally preferred to under-masking because the downstream sinks treat redacted values as opaque anyway. Calls out `"bearerToken"`/`"authToken"` (caught by `token`), `"awsCredentials"` (caught by the new `credentials` token), and the analogous `liquibaseProLicenseKey`-style keys.

**3. Added `"credentials"` token** *(CodeRabbit nitpick #1)*

Conservative addition that closes a real gap — catches `awsCredentials`, `credentialsFile`, etc. Skipped `"bearer"` because `token` already covers `bearerToken` via substring. Skipped `"auth"` per your own warning about false positives (`author`, `authority`). Added an `argument__awsCredentials` row to the existing parameterized redaction test.

**CI note** *(Filipe finding #3)*: confirmed — only `claude-review` and `label` red on this PR; same ancillary workflows as #7740. Substantive matrix green.

**Out of scope, flagged as future work:** Lombok's `@AllArgsConstructor` on the class generates `new IntegrationDetails(String name, Map<String, String> parameters)`, which would let a caller inject an already-populated unsanitized map. Symmetric concern to the `getParameters()` fix here. Currently no callers use the 2-arg ctor (only the `@NoArgsConstructor`). Worth tightening in a follow-up if you want a fully sealed boundary — happy to file separately.

Tests: 39/39 `IntegrationDetailsTest` (36 pre-existing + the `awsCredentials` row + the two new `getParameters` specs).

wwillard7800 (approved):

PR #7741, opened 2026-05-20
CWE-316: clear password memory after use in legacy CLI + CommandScope

v-petrovych:
## Impact

Passwords are represented as `java.lang.String` throughout the codebase and are never explicitly zeroed — **CWE-316: Cleartext Storage of Sensitive Information in Memory**. On long-running embedded JVMs and on hosts where heap-dump / memory-forensics is part of the threat model, the credential Strings remain reachable long after the operations that needed them have finished. This is the first slice of a wider audit ticket; the cross-cutting `Credential` / `SecretString` wrapper was already explicitly deferred in #7739 as a separate spike.

This PR addresses three concrete misuses in the legacy CLI:

1. **`Main.java:1244`** — `value = new String(c.readPassword(...))`. `Console.readPassword` deliberately returns `char[]` so callers can wipe the buffer after use; wrapping in `new String(...)` and immediately discarding the `char[]` without clearing defeats the API contract. Source even comments this is the "most notable" instance in the audit.

2. **`Main.java:86, 118, 138`** — the `password`, `referencePassword`, and `liquibaseProLicenseKey` fields hold credential Strings for the lifetime of the `Main` instance. No cleanup hook exists.

3. **`CommandScope.argumentValues` (CommandScope.java:51)** — holds every command argument (including any credential-bearing keys) for the full command duration. Not cleared after `execute()` completes, so the original credential Strings remain reachable from `Scope`-adjacent code paths post-command.

## Fix

**Item 1 — Console.readPassword `char[]` zeroing.** Read into a local `char[]`, construct the `String` in `try`, `Arrays.fill('\0')` in `finally`:

```java
char[] pwdChars = c.readPassword(attributeName + ": ");
try {
    value = new String(pwdChars);
} finally {
    java.util.Arrays.fill(pwdChars, '\0');
}
```

The String copy still lives in the destination field (Strings are immutable in Java — they cannot be wiped in-place), but the original buffer is reliably zeroed.

**Item 2 — `Main.clearCredentialFields()`.** New package-private method that nulls `password`, `referencePassword`, and `liquibaseProLicenseKey`. Called from `run()`'s **outer `finally`** so it runs on success and on any `catch (Throwable)` failure path. Nulling the field is the closest Java equivalent to wiping a String — the String becomes GC-eligible if no other references hold it.

**Item 3 — `CommandScope.clearCredentialArguments()`.** New package-private method using a substring-matched credential-key denylist (`password`, `passwd`, `secret`, `token`, `apikey`, `accesskey`, `licensekey` — the last one catches `liquibaseProLicenseKey` which `Main.createLiquibaseCommand()` places into this map). The sibling PR #7739 (not yet merged) introduces the same set of tokens in `IntegrationDetails.setParameter`; the two denylists will converge to a shared helper as follow-up work once both PRs land. Overwrites matching entries with `"*****"` rather than removing them, preserving map semantics for any late-binding reader that walks the map after `execute()`. Called from a new **outer `finally`** in `execute()`.

## Tests

- **`MainTest.clearCredentialFields_nulls_credentials_but_leaves_other_fields_intact`** — verifies `password` / `referencePassword` / `liquibaseProLicenseKey` go to `null` while `username` / `url` are untouched (so post-execution diagnostics keep their context).
- **`CommandScopeTest`** adds four specs:
  - `clearCredentialArguments redacts credential-bearing keys but leaves others` — direct unit test of the denylist (case-insensitive, substring, including `PASSWORD`, `ldap.passwd`, plus pass-through for `username` / `url` / `driver`).
  - `clearCredentialArguments is a no-op when there are no arguments`.
  - `execute invokes credential clearing in its outer finally on the success path` — wiring check, order-independent (pre-supplies `requiredArg`).
  - `execute invokes credential clearing in its outer finally even when the pipeline throws`.

```
[INFO] Tests run: 32, Failures: 0, Errors: 0, Skipped: 0 -- in MainTest
[INFO] Tests run: 19, Failures: 0, Errors: 0, Skipped: 0 -- in CommandScopeTest
[INFO] Tests run: 277, Failures: 0, Errors: 0, Skipped: 0  (broader command/jdbc/database scope)
```

No regressions in the broader CommandScope-adjacent subset (`DbUrlConnectionCommandStepTest`, `HistoryCommandStepTest`, `JdbcConnectionTest`, `OfflineConnectionTest`, `ProCommandsRegistryTest`, etc.).

## Known limitations

- **Strings are immutable in Java**; nulling the field is the closest equivalent to zeroing but does not actively erase the String contents. The fix narrows the residency window but does not solve the underlying problem. The proper long-term fix is a `Credential` / `SecretString` wrapper with explicit `clear()` and `char[]` backing storage — already explicitly deferred to a follow-up spike in #7739.
- The credential-key denylist is intentionally duplicated rather than shared — the matching sibling denylist lives in PR #7739, which is still in review at the time this PR was opened. A shared `CredentialKeyDetector` is a natural follow-up once both PRs have merged and the two token lists have stabilised.

## Out of scope (separate PRs)

This ticket lists five locations; the other two get their own PRs because they are different shapes of fix:

- **`liquibase-maven-plugin/.../AbstractLiquibaseMojo.java`** — Maven Daemon (`mvnd`) keeps the JVM warm across builds, making the field hygiene meaningful. Fix shape is the same as Item 2 here, but lives in a different module with a different review surface, so it will land as a separate PR against the same audit ticket.
- **`liquibase-standard/.../ant/type/DatabaseType.java`** — Ant's `<typedef>` + `refid` model intentionally reuses the same `DatabaseType` instance across multiple tasks in a build, so the "null in `finally`" shape used here is architecturally wrong for Ant. A proper fix needs either a `BuildListener`-based hook on project end or a `DatabaseType.clearCredentials()` API invoked by specific tasks at last-use. Tracked under the same audit ticket but warrants its own design discussion.

### DIFF
--- a/liquibase-standard/src/main/java/liquibase/command/CommandScope.java
+++ b/liquibase-standard/src/main/java/liquibase/command/CommandScope.java
@@ -54,10 +54,15 @@ public class CommandScope {
      * Substring tokens (lowercase) that mark an argument key as credential-bearing.
      * Values for matching keys are overwritten with {@code "*****"} at the end of
      * {@link #execute()} so the original credential Strings become GC-eligible and
-     * do not linger in heap for the rest of the JVM lifetime (CWE-316). Mirrors
-     * the denylist used by {@code IntegrationDetails.setParameter}, plus
-     * {@code "licensekey"} to also catch {@code liquibaseProLicenseKey} which is
-     * placed in this map via {@code Main.createLiquibaseCommand()}.
+     * do not linger in heap for the rest of the JVM lifetime (CWE-316).
+     * <p>
+     * Matching is intentionally substring-based on the lowercased key: keys like
+     * {@code "argument__bearerToken"} are caught by {@code "token"};
+     * {@code "liquibaseProLicenseKey"} (placed in this map via
+     * {@code Main.createLiquibaseCommand()}) is caught by {@code "licensekey"};
+     * AWS-style {@code accessKeyId} by {@code "accesskey"}. Over-masking is
+     * intentionally preferred to under-masking — downstream sinks treat redacted
+     * values as opaque, so false positives are harmless.
      */
     private static final String[] CREDENTIAL_KEY_TOKENS = {
             "password", "passwd", "secret", "token", "apikey", "accesskey", "licensekey"
--- a/liquibase-standard/src/main/java/liquibase/command/CommandScope.java
+++ b/liquibase-standard/src/main/java/liquibase/command/CommandScope.java
@@ -403,7 +403,11 @@ public class CommandScope {
             if (key == null) {
                 continue;
             }
-            String lowerKey = key.toLowerCase();
+            // Locale.ROOT explicitly: bare String.toLowerCase() is locale-sensitive
+            // and would mis-lowercase ASCII letters under e.g. Turkish locale
+            // ("APIKEY" -> "apıkey" with dotless ı, which would then fail the
+            // contains("apikey") substring check and silently skip redaction).
+            String lowerKey = key.toLowerCase(Locale.ROOT);
             for (String token : CREDENTIAL_KEY_TOKENS) {
                 if (lowerKey.contains(token)) {
                     entry.setValue("*****");
--- a/liquibase-standard/src/main/java/liquibase/command/CommandScope.java
+++ b/liquibase-standard/src/main/java/liquibase/command/CommandScope.java
@@ -55,10 +55,12 @@ public class CommandScope {
      * Values for matching keys are overwritten with {@code "*****"} at the end of
      * {@link #execute()} so the original credential Strings become GC-eligible and
      * do not linger in heap for the rest of the JVM lifetime (CWE-316). Mirrors
-     * the denylist used by {@code IntegrationDetails.setParameter}.
+     * the denylist used by {@code IntegrationDetails.setParameter}, plus
+     * {@code "licensekey"} to also catch {@code liquibaseProLicenseKey} which is
+     * placed in this map via {@code Main.createLiquibaseCommand()}.
      */
     private static final String[] CREDENTIAL_KEY_TOKENS = {
-            "password", "passwd", "secret", "token", "apikey", "accesskey"
+            "password", "passwd", "secret", "token", "apikey", "accesskey", "licensekey"
     };
 
     private final Map<Class<?>, Object> dependencies = new HashMap<>();
--- a/liquibase-standard/src/main/java/liquibase/command/CommandScope.java
+++ b/liquibase-standard/src/main/java/liquibase/command/CommandScope.java
@@ -50,6 +50,17 @@ public class CommandScope {
 
     private final SortedMap<String, Object> argumentValues = new TreeMap<>();
 
+    /**
+     * Substring tokens (lowercase) that mark an argument key as credential-bearing.
+     * Values for matching keys are overwritten with {@code "*****"} at the end of
+     * {@link #execute()} so the original credential Strings become GC-eligible and
+     * do not linger in heap for the rest of the JVM lifetime (CWE-316). Mirrors
+     * the denylist used by {@code IntegrationDetails.setParameter}.
+     */
+    private static final String[] CREDENTIAL_KEY_TOKENS = {
+            "password", "passwd", "secret", "token", "apikey", "accesskey"
+    };
+
     private final Map<Class<?>, Object> dependencies = new HashMap<>();
 
     /**
@@ -361,6 +372,37 @@ public class CommandScope {
             throw e;
         } catch (Exception e) {
             throw new CommandExecutionException(e);
+        } finally {
+            clearCredentialArguments();
+        }
+    }
+
+    /**
+     * Overwrite the values of any credential-bearing argument keys with
+     * {@code "*****"} so the original credential Strings stop being referenced by
+     * this scope. Strings are immutable and cannot be wiped in-place, but dropping
+     * the reference here makes them GC-eligible (assuming no other references hold
+     * the same String) instead of letting them live for the rest of the JVM in
+     * long-running embedded scenarios (CWE-316: Cleartext Storage of Sensitive
+     * Information in Memory). Called from {@link #execute()}'s outer finally so
+     * it runs regardless of command outcome; package-private for direct testing.
+     */
+    void clearCredentialArguments() {
+        if (argumentValues.isEmpty()) {
+            return;
+        }
+        for (Map.Entry<String, Object> entry : argumentValues.entrySet()) {
+            String key = entry.getKey();
+            if (key == null) {
+                continue;
+            }
+            String lowerKey = key.toLowerCase();
+            for (String token : CREDENTIAL_KEY_TOKENS) {
+                if (lowerKey.contains(token)) {
+                    entry.setValue("*****");
+                    break;
+                }
+            }
         }
     }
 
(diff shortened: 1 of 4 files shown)


v-petrovych:
@coderabbitai — addressed both inline findings in commit `6a457c747`:

**1. `licenseKey` coverage on the credential denylist** (`CommandScope.java`) — confirmed the leak path you flagged: `Main.java:2050` puts `liquibaseProLicenseKey` into `argsMap`, then `Main.java:2053-2056` (`createLiquibaseCommand`) iterates the map and calls `liquibaseCommand.addArgumentValue(entry.getKey(), entry.getValue())`. The literal key `"liquibaseProLicenseKey"` lowercased is `"liquibaseprolicensekey"` — does not contain any of the original tokens (`password`, `passwd`, `secret`, `token`, `apikey`, `accesskey`), so `clearCredentialArguments()` would silently let it survive `execute()`. Added `"licensekey"` to `CREDENTIAL_KEY_TOKENS`; the substring match cleanly catches `liquibaseProLicenseKey` and any future license-bearing key names. Extended the `clearCredentialArguments redacts credential-bearing keys but leaves others` spec to assert this. Real gap, not a polish suggestion — thanks for catching it.

**2. EOF null-check on `Console.readPassword`** (`Main.java`) — yes, `readPassword` returns `null` on Ctrl+D / Ctrl+Z; the prior code would NPE on both `new String(pwdChars)` and `Arrays.fill(pwdChars, '\0')` rather than producing an operator-actionable error. Added a `CommandLineParsingException` guard with the attribute name, matching the existing no-console pattern a few lines above:

```java
char[] pwdChars = c.readPassword(attributeName + ": ");
if (pwdChars == null) {
    throw new CommandLineParsingException(
            "No value provided for '" + attributeName + "' (end-of-input reached)");
}
```

Tests: 51/51 (`CommandScopeTest` + `MainTest`) still green after the changes.

wwillard7800:
**PR description nit:** The description says the denylist "mirrors the denylist used by `IntegrationDetails.setParameter`," but `IntegrationDetails.setParameter` (`liquibase-standard/src/main/java/liquibase/integration/IntegrationDetails.java:27`) doesn't have a denylist — it's a plain `parameters.put(key, value)`. The token list itself is sensible; just the cited prior art doesn't exist. Worth correcting so future maintainers don't go hunting for a shared denylist that isn't there.

wwillard7800:
**Style nit in `Main.java`:** `java.util.Arrays.fill(pwdChars, '\0')` is fully qualified, but `java.util.*` is already imported (`Main.java:53`). Trivial — can drop the qualifier to `Arrays.fill(pwdChars, '\0')`.

wwillard7800 (approved):

v-petrovych:
@wwillard7800 — both nits addressed in commit `41e6b1bec`, and the PR description was also corrected.

**1. Misleading "mirrors `IntegrationDetails.setParameter`" claim** — you're right that on `main` today, `IntegrationDetails.setParameter` is still the plain `parameters.put(key, value)` form; the matching denylist only lives on the sibling PR #7739 (not yet merged). Future maintainers reading this PR and hunting for the shared denylist in `main` would have come up empty.

Fixed in two places:
- **`CommandScope.java` Javadoc** — rewrote `CREDENTIAL_KEY_TOKENS` doc to be self-contained. Describes the substring-match semantics directly, gives three concrete examples (`argument__bearerToken` → `token`, `liquibaseProLicenseKey` → `licensekey`, AWS `accessKeyId` → `accesskey`), and the over-masking-is-safer rationale. No cross-reference to `IntegrationDetails`.
- **PR description** — same edit. The Item-3 paragraph now describes the seven tokens (correctly including `licensekey` after the earlier CodeRabbit-driven addition) and explicitly notes #7739 is a sibling-in-progress, with the convergence to a shared helper called out as follow-up work once both PRs have landed.

**2. `java.util.Arrays.fill` fully qualified** — fixed. `java.util.*` is indeed already imported at `Main.java:53`, so the qualifier was just noise:

```diff
-                    java.util.Arrays.fill(pwdChars, '\0');
+                    Arrays.fill(pwdChars, '\0');
```

Tests: 51/51 `CommandScopeTest` + `MainTest` still pass after the changes.

Thanks for the careful read — appreciate the "future maintainers go hunting" framing on the first nit; that's the bar this slice should hit even when the cross-references are technically right.

v-petrovych:
@coderabbitai — addressed in commit `ab911636c`. This was a genuine locale bug, not a style nit.

**The fix:**

```diff
-            String lowerKey = key.toLowerCase();
+            String lowerKey = key.toLowerCase(Locale.ROOT);
```

`Locale` is already in scope via the existing `java.util.*` import — no new import needed.

**Why it matters operationally** (in case a maintainer drops in): under tr-TR, ASCII `I` lowercases to dotless `ı` (U+0131), so:

```
"argument__APIKEY".toLowerCase()  // Turkish JVM default
  → "argument__apıkey"
  → .contains("apikey")  ⇒ FALSE
  → credential silently survives execute() unredacted
```

The bug is invisible on US/UK/etc CI matrices but is real for operator-deployed JVMs in Turkish locale (and other locales with non-ASCII lowercasing of ASCII letters). Pretty much the textbook "Turkish locale bug" — the kind that's bitten Java codebases for decades.

**Paired with a runtime regression test** in the same commit:

```groovy
def "clearCredentialArguments uses Locale.ROOT so Turkish-locale 'I' does not break the ASCII substring match"() {
    given:
    Locale savedDefault = Locale.getDefault()
    Locale.setDefault(Locale.forLanguageTag("tr-TR"))
    // ... populate scope with argument__APIKEY / argument__PASSWORD / argument__LICENSEKEY ...
    when:
    scope.clearCredentialArguments()
    then:
    // all three uppercase-I keys redacted to "*****"
    cleanup:
    Locale.setDefault(savedDefault)
}
```

Borrowing the framing from Wesley's review on #7744: a `Locale.ROOT` argument can be deleted in a future refactor without CI signal; this test sets `Locale.setDefault(forLanguageTag("tr-TR"))` (with proper cleanup) so any future code that bare-lowercases credential keys will fail the spec loudly.

**Cross-checked the sibling slice:** `IntegrationDetails.sanitize()` on PR #7739 already uses `toLowerCase(Locale.ROOT)`, so no companion fix needed there. This bug was unique to `CommandScope`.

Tests: 52/52 `CommandScopeTest` + `MainTest` green after the fix and the new regression spec.

filipelautert (approved):
Approving — appreciate the honesty about the String-vs-char[] ceiling in the PR body and code comments. The fix is the right scope for this slice.

Findings:

- **Lifecycle is correct.** Both `clearCredentialArguments()` (CommandScope.java) and `clearCredentialFields()` (Main.java) run in outer `finally` blocks, so they fire on success and exception. `char[]` zeroing on the interactive prompt at `Main.java:1244-1283` correctly follows the `Console.readPassword` contract with a sane null-check for EOF.
- **Turkish-locale regression test is great** (CommandScopeTest.groovy:158-200) — explicit `Locale.ROOT` flip catches the dotless-i pitfall that would have silently skipped redaction.

Three non-blocking asks for a follow-up (can be a separate ticket):

1. **Expand `CREDENTIAL_KEY_TOKENS`.** The current substring denylist (`password, passwd, secret, token, apikey, accesskey, licensekey`) misses `pwd`, `credential`, `privatekey`, `auth` — any future arg named e.g. `bearerCredential` or `keystorePwd` would leak silently. An opt-in `CommandArgumentDefinition.sensitive()` flag would be the safer long-term shape, but expanding the denylist costs nothing.
2. **`// TODO(audit)` in `clearCredentialFields()`** listing the sources still unhandled: `Properties` map loaded from `liquibase.properties`, env vars surfaced via `System.getenv`, and `IntegrationDetails.setParameter` values. Useful so the gap doesn't get lost.
3. **Release-note callout:** any downstream extension that calls `scope.getConfiguredValue(PASSWORD_ARG)` after `execute()` returns now gets `"*****"`. I don't see anything in tree that hits that pattern, but external extensions might.

PR #7743, opened 2026-05-20
CWE-316: clear Ant DatabaseType credentials at buildFinished

v-petrovych:
## Impact

Third companion to #7741 and #7742 — same root cause (`java.lang.String` passwords never zeroed; **CWE-316**), applied to the Ant integration's `liquibase.integration.ant.type.DatabaseType`. The `password` field on this type holds a credential String for the lifetime of the Ant `Project` with no cleanup hook.

## Why this PR is shaped differently from #7741 / #7742

Ant `DatabaseType` **cannot** use the "null in `execute()` finally" pattern that #7741 (CLI) and #7742 (Maven Mojo) use. Ant's `<typedef>` + `databaseRef` mechanism is **specifically designed** to let multiple tasks share a single `<database id="…">` instance across a build:

```xml
<database id="mydb" password="secret" url="..." />
<liquibase:update     databaseRef="mydb" changelog="x.xml" />
<liquibase:rollback   databaseRef="mydb" changelog="x.xml" toTag="v1" />
<liquibase:status     databaseRef="mydb" />
```

Nulling `password` at the end of `update`'s `execute()` would break `rollback` and `status`. The right lifecycle hook is **`Project.buildFinished`** — it fires after the very last task on the project, so credentials are gone before the JVM idles but never while another task still needs them.

## Fix

Self-contained in `DatabaseType.java` — no changes to `BaseLiquibaseTask`, `AbstractDatabaseDiffTask`, or any task consumer.

**1. `clearCredentials()`** — package-private. Nulls `password`. For refid shells (DatabaseTypes whose getters delegate to a referenced instance via `setRefid`), traverses to the referenced `DatabaseType` so the actual shared field is wiped. Defensively wraps the ref traversal in `try/catch` for `BuildException` so a dangling refid at project tear-down does not throw.

```java
void clearCredentials() {
    this.password = null;
    if (isReference()) {
        try {
            Object ref = getCheckedRef();
            if (ref instanceof DatabaseType) {
                ((DatabaseType) ref).clearCredentials();
            }
        } catch (BuildException ignored) { /* dangling ref at tear-down */ }
    }
}
```

**2. `registerCredentialCleanup()`** — package-private. Attaches a one-shot `BuildListener` to the `Project` that calls `clearCredentials()` on `buildFinished`. The listener implements only `buildFinished` meaningfully; `buildStarted` / `targetStarted` / `taskStarted` / etc. are no-ops because during the build the credentials may still be needed by another consumer.

**3. Auto-registration from `createDatabase()`** — the first call to `createDatabase(...)` on a given `DatabaseType` instance invokes `registerCredentialCleanup()`; subsequent calls are no-ops thanks to the `credentialCleanupRegistered` flag. Any Ant task that consumes this type via `createDatabase(...)` opts in implicitly.

## Test

New `DatabaseTypeCredentialClearTest` covers six scenarios:

- `clearCredentials_nulls_password_on_a_simple_databaseType` — straight password nulling; user / url untouched.
- `clearCredentials_traverses_refid_and_clears_the_referenced_instance` — exercises the typedef+`databaseRef` pattern. A real referenced `DatabaseType` holds the credential; a refid shell calls `clearCredentials`; the referenced instance's password is asserted nulled afterwards.
- `registerCredentialCleanup_registers_a_BuildListener_that_fires_on_buildFinished` — counts Project listeners before/after; `Project.fireBuildFinished(null)` triggers the clearing.
- `registerCredentialCleanup_is_idempotent` — three calls add exactly one listener.
- `registerCredentialCleanup_listener_only_implements_buildFinished_meaningfully` — fires `buildStarted` / `targetStarted` / `taskStarted` / `taskFinished` / `targetFinished` / `messageLogged` and asserts password is still set; only `buildFinished` clears.
- `clearCredentials_is_safe_on_a_refid_shell_whose_referenced_was_dropped` — defensive case (dangling refid at tear-down does not throw).

```
[INFO] Tests run: 6, Failures: 0, Errors: 0, Skipped: 0 -- in DatabaseTypeCredentialClearTest
```

No regressions in the surrounding Ant code (changes are purely additive — new field, new methods; existing methods unchanged).

## Design alternatives considered

- **Null in `BaseLiquibaseTask.execute()` finally** — rejected: breaks the typedef+`databaseRef` sharing pattern (would null on the first task, break the second).
- **Per-task reference counting** — rejected: Ant doesn't tell us "no more references will be added", so we cannot decide when "done".
- **Explicit `<liquibase:clearCredentials>` task** — rejected: opt-in default is bad for security; users will not know to add it.
- **`Project.buildFinished` BuildListener (chosen)** — Ant guarantees it fires after the last task on the project. Tested behaviour.

## Known limitation

Strings are immutable in Java; nulling the field is the closest equivalent to wiping the contents but does not actively erase the String value from the JVM heap / interned cache. The proper long-term fix is a `Credential` / `SecretString` wrapper, already explicitly deferred to a follow-up spike in #7739.

## Related

- #7741 — CWE-316 sibling for `liquibase-standard` legacy CLI (Main.java + CommandScope.argumentValues).
- #7742 — CWE-316 sibling for `liquibase-maven-plugin` (Mojo password fields).

Same audit ticket, same CWE, three different module-specific fix shapes. PRs are independent — any merge order works.

### DIFF
--- a/liquibase-standard/src/main/java/liquibase/integration/ant/type/ConnectionProperties.java
+++ b/liquibase-standard/src/main/java/liquibase/integration/ant/type/ConnectionProperties.java
@@ -5,9 +5,24 @@ import org.apache.tools.ant.types.PropertySet;
 
 import java.util.LinkedList;
 import java.util.List;
+import java.util.Locale;
 import java.util.Properties;
 
 public class ConnectionProperties {
+
+    /**
+     * Substring tokens (lowercase) that mark a {@code <connectionProperty>} as
+     * credential-bearing for the {@link #clearCredentialValues()} sweep. Matched
+     * case-insensitively against the lowercased property name via
+     * {@link Locale#ROOT} so the Turkish-locale dotless-i quirk does not cause
+     * an {@code I} in {@code APIKEY} to map to {@code ı} and miss the
+     * {@code apikey} token. Mirrors the analogous denylists used elsewhere in
+     * the May-2026 OSS credential-handling audit (CommandScope, IntegrationDetails).
+     */
+    private static final String[] CREDENTIAL_NAME_TOKENS = {
+            "password", "passwd", "secret", "token", "apikey", "accesskey", "credentials"
+    };
+
     private final List<PropertySet> propertySets;
     private final List<Property> properties;
 
@@ -34,5 +49,33 @@ public class ConnectionProperties {
 
     public void addConnectionProperty(Property property) {
         properties.add(property);
+    }
+
+    /**
+     * Overwrites the value of any {@code <connectionProperty>} whose name marks it as
+     * credential-bearing (per {@link #CREDENTIAL_NAME_TOKENS}) with {@code "*****"} so
+     * the raw credential String no longer lives in this object after the Ant build
+     * finishes. Called from {@link DatabaseType#clearCredentials()}'s buildFinished
+     * listener (CWE-316: Cleartext Storage of Sensitive Information in Memory).
+     * <p>
+     * Note: this method intentionally does <em>not</em> touch {@link #propertySets}.
+     * Those entries reference Project-level property sets owned by the surrounding
+     * Ant build, not the {@code DatabaseType} itself; clearing them would mutate
+     * unrelated build state.
+     */
+    void clearCredentialValues() {
+        for (Property property : properties) {
+            String name = property.getName();
+            if (name == null) {
+                continue;
+            }
+            String lowerName = name.toLowerCase(Locale.ROOT);
+            for (String token : CREDENTIAL_NAME_TOKENS) {
+                if (lowerName.contains(token)) {
+                    property.setValue("*****");
+                    break;
+                }
+            }
+        }
     }
 }
--- a/liquibase-standard/src/main/java/liquibase/integration/ant/type/DatabaseType.java
+++ b/liquibase-standard/src/main/java/liquibase/integration/ant/type/DatabaseType.java
@@ -320,7 +320,7 @@ public class DatabaseType extends DataType {
         if (credentialCleanupRegistered) {
             return;
         }
-        Project project = getProject();
+        final Project project = getProject();
         if (project == null) {
             return;
         }
@@ -334,7 +334,18 @@ public class DatabaseType extends DataType {
                     // One-shot: detach the listener so it (and the DatabaseType
                     // it closes over) can be GC'd in long-lived embedded Ant
                     // scenarios where the Project survives across many builds.
-                    event.getProject().removeBuildListener(this);
+                    // Use the captured `project` instead of event.getProject()
+                    // so the deregister target is the same Project we attached
+                    // to, regardless of which Project fired the event.
+                    project.removeBuildListener(this);
+                    // Reset the idempotency flag so a SUBSEQUENT build on the
+                    // same DatabaseType instance — possible in long-lived
+                    // embedded hosts that reuse DataType instances across
+                    // builds — can register a fresh listener. Without this
+                    // reset, build 1's credentials would be cleared but build
+                    // 2 and onward would silently skip re-registration (per
+                    // @coderabbitai's review on #7743).
+                    credentialCleanupRegistered = false;
                 }
             }
             @Override public void targetStarted(BuildEvent event) {}
--- a/liquibase-standard/src/main/java/liquibase/integration/ant/type/DatabaseType.java
+++ b/liquibase-standard/src/main/java/liquibase/integration/ant/type/DatabaseType.java
@@ -4,7 +4,9 @@ import liquibase.database.Database;
 import liquibase.database.DatabaseFactory;
 import liquibase.exception.DatabaseException;
 import liquibase.resource.ResourceAccessor;
+import org.apache.tools.ant.BuildEvent;
 import org.apache.tools.ant.BuildException;
+import org.apache.tools.ant.BuildListener;
 import org.apache.tools.ant.Project;
 import org.apache.tools.ant.types.DataType;
 import org.apache.tools.ant.types.Reference;
@@ -35,11 +37,14 @@ public class DatabaseType extends DataType {
     private String databaseChangeLogLockTableName;
     private String liquibaseTablespaceName;
 
+    private boolean credentialCleanupRegistered = false;
+
     public DatabaseType(Project project) {
         setProject(project);
     }
 
     public Database createDatabase(ResourceAccessor resourceAccessor) {
+        registerCredentialCleanup();
         logParameters();
         validateParameters();
         try {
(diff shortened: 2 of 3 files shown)


v-petrovych:
@coderabbitai — reviewed your walkthrough on this PR. ✅ "No actionable comments were generated in the recent review" — nothing to address from the CodeRabbit side. Thanks for the clean pass.

On the pre-merge **Docstring Coverage 11.11%** warning: that reflects the file's pre-existing state (the 17 getters/setters on `DatabaseType` are undocumented). The two new methods added by this PR (`registerCredentialCleanup()` and `clearCredentials()`) both carry full javadocs — diff-level docstring coverage is 100%. Not actionable from my side without scope-creeping into a documentation pass over the whole file.

Separately, **CodeQL flagged one finding** on the same diff (`DatabaseType.java:345`, [security/code-scanning/2259](https://github.com/liquibase/liquibase/security/code-scanning/2259)) — deprecated `DataType.getCheckedRef()` zero-arg invocation. Addressed in commit `77c86917e`: switched to the type-safe `getCheckedRef(DatabaseType.class, "database")` overload, which lets me drop the `instanceof` check (the typed overload either returns a `DatabaseType` or throws `BuildException`, both already handled by the existing `try/catch`).

```diff
-                Object ref = getCheckedRef();
-                if (ref instanceof DatabaseType) {
-                    ((DatabaseType) ref).clearCredentials();
-                }
+                DatabaseType ref = getCheckedRef(DatabaseType.class, "database");
+                ref.clearCredentials();
```

The other 17 call sites in `DatabaseType.java` still use the deprecated zero-arg form — CodeQL only flagged my new call site (likely a "new code" threshold). Modernising the whole file would be a separate refactor.

Tests: 6/6 `DatabaseTypeCredentialClearTest` specs still pass; the dangling-refid defensive test now exercises the typed overload's exception behaviour instead of the deprecated path.

wwillard7800:
**Optional fix: have the listener deregister itself after firing.** Inside `buildFinished`, add `event.getProject().removeBuildListener(this)` so the listener tidies up after the cleanup runs.

In a normal one-shot Ant build this doesn't matter — the `Project` gets GC'd and the listener with it. But the audit ticket's threat model specifically calls out long-running embedded scenarios (the same motivation cited in #7742 for `mvnd`). If Ant is embedded in a long-lived tool that reuses a single `Project` across many builds, each new `DatabaseType` instance permanently adds a listener the `Project` never lets go of. Those listeners then hold strong references to their closed-over `DatabaseType`, keeping it alive past its useful life — exactly the residency-window problem the rest of this PR series is trying to narrow.

One-line fix; symmetric with the rest of the series.

wwillard7800 (approved):

v-petrovych:
@wwillard7800 — addressed in commit `efb7230d5`. This was a real gap, not just a polish.

Your framing of *"symmetric with the rest of the series"* is exact, and I want to spell it out because it took me a moment to internalise why this PR uniquely needed the fix:

| PR | Cleanup hook | Lifecycle owner | Accumulation risk in embedded host |
|---|---|---|---|
| #7741 | `CommandScope.execute()` finally | `CommandScope` (per-command) | None — scope is short-lived |
| #7742 | `AbstractLiquibaseMojo.execute()` finally | Mojo instance (per-execution) | None — Mojo is per-execution |
| **#7743 (this)** | **Project `BuildListener`** | **`Project` (host-controlled lifetime)** | **Real — listener outlives the DatabaseType, holds a strong ref to it** |

So this PR was the only one whose lifecycle hook didn't naturally tear down with the per-execution object — which is exactly what your comment caught.

**Fix** at `DatabaseType.java:319-327`:

```diff
 @Override public void buildFinished(BuildEvent event) {
-    clearCredentials();
+    try {
+        clearCredentials();
+    } finally {
+        // One-shot: detach the listener so it (and the DatabaseType
+        // it closes over) can be GC'd in long-lived embedded Ant
+        // scenarios where the Project survives across many builds.
+        event.getProject().removeBuildListener(this);
+    }
 }
```

The `finally` guarantees deregistration even if `clearCredentials()` throws — which it shouldn't (it catches `BuildException` from the refid path), but defense in depth.

**Two new regression tests** to lock the behaviour in:

1. `registerCredentialCleanup_listener_deregisters_itself_after_firing` — records `project.getBuildListeners().size()` before and after `project.fireBuildFinished(null)`, asserts the count returns to baseline. Pairs with the assertion that the password is nulled, so we know `buildFinished` actually ran (not just silently skipped).

2. `multiple_databaseTypes_all_deregister_their_listeners_in_embedded_scenario` — the realistic embedded threat model in test form: three `DatabaseType` instances against one `Project`, three listeners attached simultaneously, one `buildFinished` cycle drains all three back to baseline. This is the test that catches accumulation regressions specifically.

Added an extra javadoc paragraph on `registerCredentialCleanup()` explaining *why* the listener deregisters — credits the embedded host scenario as the motivation, so a future maintainer doesn't strip the `removeBuildListener` thinking it's redundant in CLI mode.

Tests: 8/8 `DatabaseTypeCredentialClearTest` specs green (6 pre-existing + 2 new). The pre-existing `registerCredentialCleanup_listener_only_implements_buildFinished_meaningfully` continues to pass — it invokes `buildFinished` directly on the listener reference, which still finds itself in `project.getBuildListeners()` and removes successfully.

Thanks for the catch — appreciated the explicit "embedded host" reasoning in the original comment; that's exactly the threat model the rest of this slice has been navigating, and the asymmetry with #7741/#7742 wasn't obvious until you pointed it out.

filipelautert (commented):
Most of this is good and the test is unusually thorough — direct field assertions via `getPassword()`, refid traversal, idempotency, dangling-refid handling, and listener self-deregistration for both single and N-instance accumulation scenarios. The String-vs-char[] ceiling is honestly noted in the javadoc at `DatabaseType.java:331-334`.

`buildFinished` is the right lifecycle hook — Ant's `Project.fireBuildFinished` is invoked from `Main` in a `finally`, so it fires on both success and exception.

One real gap before I approve: **`ConnectionProperties` lets users pass arbitrary JDBC props including `password=` via `<connectionProperty>`** (DatabaseType.java:28, ConnectionProperties.java:25-27), and those `Property.getValue()` Strings are NOT cleared by the listener. So a build that uses:

```xml
<connectionProperty name="password" value="hunter2"/>
```

…instead of (or in addition to) `<password>` retains the credential post-build.

Could you either:

(a) extend `clearCredentials()` to walk the `connectionProperties` list and null any name-matching-password entries (case-insensitive, locale-safe), or
(b) add a `// TODO(audit)` referencing this so it doesn't get lost?

(a) is small and self-contained. Either is fine for me to approve.

v-petrovych:
@filipelautert — fair point, addressed in commit `8488b0ef0`. Took your option **(a)** — the gap is real and self-contained.

**Verified the gap by reading the code:** `ConnectionProperties.properties` is a `List<Property>` (Ant `Property` objects with `getName()`/`getValue()`). `buildProperties()` flattens it into a JDBC `Properties` map at `createDatabase()` time, so a build that uses `<connectionProperty name="password" value="hunter2"/>` actually uses `"hunter2"` as the JDBC password. The raw `"hunter2"` String then sits in the `Property` list past `buildFinished` — my listener didn't touch it. ✓ Real gap.

**Implementation:**

- New package-private `ConnectionProperties.clearCredentialValues()` — walks the local `properties` list and `setValue("*****")` on any `Property` whose `getName().toLowerCase(Locale.ROOT)` contains a credential token. Helper lives next to the data it operates on.
- `DatabaseType.clearCredentials()` calls it when `this.connectionProperties != null`. Existing refid traversal flows through unchanged — the referenced `DatabaseType`'s `connectionProperties` get wiped via the existing `((DatabaseType) ref).clearCredentials()` call.

**A couple of small specifics worth flagging:**

- **Token set is broader than literal `"password"`.** I used the same denylist as the rest of the audit (`password`, `passwd`, `secret`, `token`, `apikey`, `accesskey`, `credentials`) so `<connectionProperty name="awsCredentials" .../>` and `<connectionProperty name="ldap.bearerToken" .../>` are also caught. Zero extra cost; matches the shape of `CommandScope` / `IntegrationDetails`. Let me know if you'd rather scope it down to literal `password`-only — easy to narrow.
- **`Locale.ROOT` on the lowercase.** As you requested. Test exercises the Turkish-locale `tr-TR` default (with proper restore in `finally`) and asserts `APIKEY` still gets wiped — the dotless-i quirk would let a bare `toLowerCase()` miss it.
- **`propertySets` deliberately not walked.** Those reference Project-level property sets owned by the surrounding Ant build, not the `DatabaseType` itself; clearing them would mutate unrelated build state. Comment in the helper calls this out so a future maintainer doesn't broaden the sweep unintentionally.

**Three new regression specs in `DatabaseTypeCredentialClearTest`:**

| Spec | What it pins |
|---|---|
| `clearCredentials_wipes_credential_bearing_connectionProperty_values` | Direct test of the gap you flagged — `<connectionProperty name="password" value="hunter2"/>` ends up `*****`; `<connectionProperty name="connectTimeout" value="30000"/>` passes through unchanged |
| `clearCredentials_matches_credential_property_names_case_insensitively_locale_safe` | Sets JVM default locale to `tr-TR` (restored in `finally`); asserts `APIKEY`, `Passwd`, `client_secret`, `accessKey`, `awsCredentials`, `ldap.bearerToken` are all caught; pins the `Locale.ROOT` guarantee |
| `clearCredentials_traverses_refid_and_wipes_referenced_connectionProperty_values` | typedef+`databaseRef` shared-instance case: a refid shell whose referenced has `<connectionProperty name="password" .../>` must see the referenced `Property.value` wiped via the existing refid traversal |

11/11 `DatabaseTypeCredentialClearTest` specs pass (8 prior + 3 new).

Branch was auto-updated by the maintainer queue with `cc99820b8 Merge branch 'main'` before I pushed; rebased my fix-commit on top, no conflicts.

v-petrovych:
@coderabbitai — fair catch, real bug. Applied your diff in commit `629112236`.

**Verifying the state machine** before I changed anything:

```
Build 1: createDatabase() → registerCredentialCleanup() → flag=true, listener attached
Build 1 ends: buildFinished() → clearCredentials() → removeBuildListener(this)
            ❌ flag stays true; no reset

Build 2: createDatabase() → registerCredentialCleanup() → flag=true → early return
        ❌ NO listener attached for Build 2
Build 2 ends: NO listener → credentials NOT cleared
```

This pairs neatly with @wwillard7800's earlier review on the same PR: he flagged the accumulation case (listener piling up across builds in long-lived embedded hosts), I added the self-removal, and your review now catches the symmetric state-machine bug introduced *by* that self-removal — the flag never gets reset. In the same long-lived-embedded-host threat model both reviews assume, this is strictly worse than the accumulation bug — at least there the cleanup ran every time; without the flag reset, it runs *once* and then never again.

**Applied both changes from your diff:**

```diff
-    Project project = getProject();
+    final Project project = getProject();
...
             try {
                 clearCredentials();
             } finally {
-                event.getProject().removeBuildListener(this);
+                project.removeBuildListener(this);
+                credentialCleanupRegistered = false;
             }
```

The `event.getProject()` → captured `project` swap is the smaller robustness win (removes the dependency on the event's project matching the registration's project — same in normal Ant operation, but the captured form is more explicit). The flag reset is the real fix.

**New regression spec** `registerCredentialCleanup_resets_idempotency_flag_so_subsequent_builds_re_register` does three back-to-back `setPassword → registerCredentialCleanup → fireBuildFinished` cycles on the SAME DatabaseType instance and asserts each cycle: listener gets attached, then detached, and the password gets nulled. Without the flag reset, the second cycle's `registerCredentialCleanup` would early-return and the listener-count assertion would fail. Three cycles instead of two — pins the contract for arbitrary N.

12/12 `DatabaseTypeCredentialClearTest` specs green.

filipelautert (approved):
Thanks Victor — both gaps addressed cleanly. Re-reviewed at `0925bf53c4`:

- **ConnectionProperties sweep** (`clearCredentialValues()` at ConnectionProperties.java:66-80) — token list matches the analogous denylists elsewhere in the audit, uses `Locale.ROOT` per the Turkish-locale precedent. The new test explicitly flips `Locale.setDefault(tr-TR)` for the duration of the call, which is exactly the shape to lock the contract.
- **State-machine fix from @coderabbitai's catch** — `credentialCleanupRegistered = false` reset inside `buildFinished`, plus the explicit `final` Project capture so the deregister targets the same Project we attached to. The 3-cycle regression spec is the right shape; two cycles wouldn't prove the flag resets for build N+1.

LGTM, approving.

PR #7740, opened 2026-05-20
CWE-693: add generic JDBC userinfo fallback to sanitizeUrl coverage

v-petrovych:
## Impact

`JdbcConnection.sanitizeUrl` is the centralised credential-redaction helper used across many logging / exception / MDC paths in Liquibase (DatabaseFactory, DatabaseInfo, AbstractDatabaseConnectionCommandStep, DbUrlConnectionCommandStep, ReferenceDbUrlConnectionCommandStep, Main, and — after PRs #7737 / #7738 / #7739 — several other call sites in the audit). Its userinfo-style matchers, however, are **dialect-specific**: only Oracle-thin, MySQL, MariaDB, and CosmosDB URLs get `user:pass@host` credentials redacted. Any other `jdbc:<driver>://` URL passes through `sanitizeUrl` **unmodified**, credentials intact.

Concrete misses today:

- `jdbc:postgresql://user:pass@host/db` (PostgreSQL JDBC accepts userinfo)
- `jdbc:sqlserver://user:pass@host/db`
- `jdbc:<third-party>://user:pass@host/db` — any extension or future driver

The downstream consequence: log lines, MDC entries, and exception messages that funnel through `sanitizeUrl` to be "safe" can still surface cleartext credentials for these URL shapes.

This maps to **CWE-693: Protection Mechanism Failure** — the protection mechanism silently no-ops on URLs outside its allowlist, defeating the defense its callers depend on. It is also explicitly noted as a known limitation in #7737's and #7739's PR bodies; this PR closes that gap.

## Fix

The three filter regexes in `JdbcConnectionPatterns`:

- `FILTER_CREDS = "(?i)/(.*)((?=@))"` (used by `stripPasswordPropFromJdbcUrl` for `getURL()`)
- `FILTER_CREDS_MYSQL_TO_OBFUSCATE = "(?i).+://(.*?)([:])(.*?)((?=@))"` (used by `sanitizeUrl` standard path)
- `FILTER_CREDS_MYSQL_TO_OBFUSCATE_EMPTY = "(?i).+://(.*?)([:])(.*?@)"` (used by `sanitizeUrl(url, true)`)

…are already generic — they match any `scheme://user:pass@host` shape. **Only the matchers** restrict them to specific dialects. Register the same three filters one additional time with a wildcard matcher, `Pattern.compile("(?i)jdbc:.*")`, so any jdbc-prefixed URL whose dialect-specific matcher did not fire still gets basic redaction.

```java
[...]

PR #7746, opened 2026-05-21
CWE-532: obfuscate H2 example-server password before echoing to stdout

v-petrovych:
## Impact

`StartH2CommandStep.run()` builds a multi-line "Connection Information:" banner and prints it to `System.out` after the embedded H2 example-server starts. Two of those lines (`"    Password: " + password`) concatenated the **raw** password — once for the dev H2 instance, once for the integration instance.

`PASSWORD_ARG` is already registered with `ConfigurationValueObfuscator.STANDARD` at `StartH2CommandStep.java:49`, so the value is correctly redacted in logs and MDC — but the stdout banner sidestepped that obfuscation entirely.

Stdout isn't a log file in the strict sense, but operationally it flows into the same sinks: terminal multiplexer logs (tmux/screen/PowerShell transcripts), CI runners' captured-output buffers (GitHub Actions / Jenkins / GitLab job logs), and is visible to anyone looking over the operator's shoulder. **CWE-532** (Insertion of Sensitive Information into Log File) covers this surface.

The default H2 example-server password is `letmein` (the helper's `description` even publishes it), so the audit-ticket priority is calibrated by the realistic threat: an operator who passes `--password=<their-real-password>` at the CLI for a quick local experiment, then pushes a terminal capture or CI artifact to a teammate.

## Fix

**1. Extract `buildConnectionInfoMessage` helper** — package-private static method that takes the *already-obfuscated* password as a parameter. The helper has no path to read the raw value; the contract is enforced by signature.

**2. In `run()`** — compute the obfuscated form once via `ConfigurationValueObfuscator.STANDARD.obfuscate(password)` and pass it to the helper. The raw `password` variable is still used at the `DriverManager.getConnection(...)` call sites — only the stdout-bound string is obfuscated, so functionality (the JDBC connection actually works) is unchanged.

```java
final String password = commandScope.getConfiguredValue(PASSWORD_ARG).getValue();
// CWE-532: never echo the raw password to stdout ...
final String obfuscatedPassword = ConfigurationValueObfuscator.STANDARD.obfuscate(password);
// ...
System.out.println(buildConnectionInfoMessage(username, obfuscatedPassword, dbPort, devUrl, intUrl));
```

The banner output is byte-for-byte identical to the prior version when given the same arguments — the only behavioural change is that the `Password:` field now reads `*****` instead of the cleartext password.

## Tests

New `StartH2CommandStepTest`:

- **`buildConnectionInfoMessage never echoes the raw password — only the obfuscated form appears (CWE-532)`** — uses a distinctive raw password (`"supersecret-pw-12345-NOT-A-REAL-CREDENTIAL"`), asserts it is absent from the banner, asserts `*****` appears in **exactly two** `Password: *****` lines (dev + integration), asserts non-credential fields (`username`, JDBC URLs, web URLs) pass through unchanged.
- **`buildConnectionInfoMessage with the default 'letmein' password also goes through the obfuscator`** — sanity that the documented default isn't special-cased.

```
[INFO] Tests run: 2, Failures: 0, Errors: 0, Skipped: 0 -- in StartH2CommandStepTest
[INFO] BUILD SUCCESS
```

## Why a helper extraction rather than inlining the obfuscation?

Two reasons, both motivated by the testability framing reviewers have pushed for elsewhere in this audit slice (per @wwillard7800's *"comment can be deleted, test fails loudly"* point on #7744):

1. **Pinning the contract.** A future "let's clean up this giant string concatenation" patch could trivially re-inline the raw password without noticing — there's no compile-time signal. With the helper, the password parameter is a `String` with an `obfuscatedPassword` name, and the JavaDoc explicitly states the contract. Re-introducing the bug requires consciously fighting the API.
2. **Test cost.** Standing up an actual H2 server in a unit test to assert what `System.out.println` produced is heavyweight (multi-threaded, blocks on `Thread.sleep(Long.MAX_VALUE)`, requires `System.out` capture). Helper-level test runs in 0.4s, cleanly proves the formatting contract, and the call-site wiring is one short line (`obfuscatedPassword = ConfigurationValueObfuscator.STANDARD.obfuscate(password)`) that's verifiable by code review.

## Operator-visible behaviour change

The audit ticket explicitly calls out the workflow change: *"users who rely on stdout copy-paste of the dev password will need a minor workflow change."* The default `letmein` is published in the `PASSWORD_ARG.description` (`"Password to use for created h2 user"`), and any user-supplied password is one the operator already knew when typing `--password=...`. The stdout copy-paste convenience was a credential-handling antipattern; the workflow change is the intended outcome.

## Related

Part of the May-2026 OSS credential-handling audit slice (`sdou` label). Companion PRs that landed earlier:

- **MERGED**: #7737 (CWE-209 in `DatabaseFactory`), #7738 (CWE-209 in `OfflineConnection`)
- **OPEN**: #7739, #7740, #7741, #7742, #7743, #7744

### DIFF
--- a/liquibase-standard/src/main/java/liquibase/command/core/StartH2CommandStep.java
+++ b/liquibase-standard/src/main/java/liquibase/command/core/StartH2CommandStep.java
@@ -85,6 +85,14 @@ public class StartH2CommandStep extends AbstractCommandStep {
 
         final String username = commandScope.getConfiguredValue(USERNAME_ARG).getValue();
         final String password = commandScope.getConfiguredValue(PASSWORD_ARG).getValue();
+        // CWE-532 (Insertion of Sensitive Information into Log File): never echo the
+        // raw password to stdout — captures into terminal loggers, CI output, and
+        // shoulder-surfing. PASSWORD_ARG is already registered with
+        // ConfigurationValueObfuscator.STANDARD for log/MDC redaction (see line 49);
+        // apply the same obfuscator here so stdout matches the rest of the platform's
+        // redaction surface. The raw `password` value is still used below for the
+        // actual JDBC connection — only the stdout-bound string is obfuscated.
+        final String obfuscatedPassword = ConfigurationValueObfuscator.STANDARD.obfuscate(password);
         final Integer dbPort = commandScope.getConfiguredValue(DB_PORT_ARG).getValue();
         final Integer webPort = commandScope.getConfiguredValue(WEB_PORT_ARG).getValue();
         final Boolean detached = commandScope.getConfiguredValue(DETACHED).getValue();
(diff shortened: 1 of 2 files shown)


filipelautert (approved):
LGTM — clean, scope-limited fix.

- Raw `password` reaches only the JDBC `getConnection()` call sites (StartH2CommandStep.java:101-102), which is required for the server to authenticate.
- Banner path uses obfuscated value end-to-end: `StartH2CommandStep.java:95` computes it once, `:110` passes it into `buildConnectionInfoMessage()`, and `:168-181` shows it via a parameter-named `obfuscatedPassword` so the contract is hard to break in a future edit.
- Test pins both halves: `!message.contains(rawPassword)` with a sentinel value + `message.count("Password: *****") == 2` (StartH2CommandStepTest.groovy:28,33).

Full redaction (`*****`) rather than partial reveal — correct call for stdout/scrollback. Approving.

PR #7748, opened 2026-05-21
CWE-470: opt-in restricted mode for customChange changelog change

v-petrovych:
## Paired follow-up: PR #7749 closes the `<customPrecondition>` half of CWE-470

> **Heads-up for reviewers:** the sibling `<customPrecondition>` element has the *same* unsafe-reflection sink (`CustomPreconditionWrapper.java:67` → `Class.forName(className, true, classLoader)`) and is closed by **[#7749](https://github.com/liquibase/liquibase/pull/7749)** under the **same `ALLOW_CUSTOM_CHANGE` flag** — per the audit author's explicit guidance: *"Covered by the same flag proposed for B2."*
>
> **Merge order:** this PR (#7748) merges first; #7749 then rebases on top, resolves the `GlobalConfiguration.java` collision by **keeping #7749's broader description** (which mentions both `customChange` *and* `customPrecondition`). That's the correct end-state — confirmed by @filipelautert on [the #7749 approval](https://github.com/liquibase/liquibase/pull/7749#pullrequestreview-4336812656).
>
> Together, the two PRs close both `<customChange>` and `<customPrecondition>` under one operator-visible flag, so flipping `liquibase.allowCustomChange=false` actually does what the flag name implies (gates *all* custom-Java loading).

---

## Impact

`CustomChangeWrapper.loadCustomChange(className)` (`CustomChangeWrapper.java:85-104`) loads an arbitrary JVM class by FQCN via `Class.forName(className, true, classLoader)`. The `initialize=true` argument means the class's static `<clinit>` initializer fires **at load time** — **before** the cast to `CustomChange` (or any other safety check) could reject the load:

```java
return (CustomChange) Class.forName(className, true,
//                                            ^ static initializer + ctor already ran
```

Any class on the JVM classpath is reachable this way: name the class in a `<customChange>` element, get its static-init code and zero-arg constructor invoked as a side effect. The triple-fallback chain (`OSGi loader` / `Scope classloader` / `thread context CL` / `bare Class.forName`) and the no-allowlist / no-marker-interface-check pattern make the surface broader than "classes that implement `CustomChange`" — the load itself is the side effect.

This maps to **CWE-470** (Use of Externally-Controlled Input to Select Classes or Code, aka *Unsafe Reflection*).

## Scope decision (sibling to #7747)

Sibling-shape to **DAT-23015 / PR #7747** (B1, `executeCommand`). Same opt-in restricted-mode pattern: add a Boolean flag, default `true` (preserves the documented `customChange` feature for the standard trust model — team-authored, team-reviewed changelogs), opt-out for embedders running changelogs from less-trusted sources (multi-tenant SaaS, downloaded change-packs, contributor PRs pre-review).

**No default behaviour change.** Everyone using `customChange` today sees byte-for-byte identical behaviour.

## Fix — three pieces

**1. New `GlobalConfiguration.ALLOW_CUSTOM_CHANGE` flag** — `liquibase.allowCustomChange`, Boolean, defaults to `true`. Description spells out the static-initializer-on-load mechanic so a future maintainer sees *why* the gate exists.

**2. Dual gate inside `CustomChangeWrapper`:**

   **(a) Hard error in `validate(Database)`** — short-circuits BEFORE the existing load/configure/`customChange.validate` pipeline. Operator sees a clean pre-execution failure with a message that names the flag in both `=false` and `=true` directions.

   **(b) Defense-in-depth gate inside `loadCustomChange` itself** — throws `CustomChangeException` BEFORE any of the three `Class.forName` / `loadClass` paths run.

The second gate is necessary because `loadCustomChange` has **five callers**: `getCustomChange()`, `validate()`, `generateCheckSum()`, `configureCustomChange()`, and **`customLoadLogic()`**. The last one runs at changelog **PARSE-TIME**, *before* `validate()` ever fires. A validate-only gate would let the malicious class's static initializer run during parse. The two gates use the same flag and the same error message; the redundancy is the point — together they guarantee the static initializer never fires when the embedder has set the flag to `false`.

**3. Four new Spock specs** in `CustomChangeWrapperTest` (24 pre-existing → 28 total):

| Spec | What it pins |
|---|---|
| `validate passes by default — customChange is intentional under the standard trust model (CWE-470 opt-out gate)` | Default-true path: zero behaviour change for existing users |
| `validate fails when liquibase.allowCustomChange=false — embedder opt-out path` | Hard error, names the flag in both directions; uses `Scope.child(...)` to exercise the real `GlobalConfiguration` lookup path |
| `loadCustomChange gate fires BEFORE Class.forName even for a class that does not exist` | Sets a non-existent class name AND flag=false; asserts ONLY the configured-off error appears, NOT the `"Exception thrown loading com.example..."` `ClassNotFoundException` warning — proving the gate short-circuited the lookup |
| `loadCustomChange method-level gate throws CustomChangeException directly when allowCustomChange=false` | Calls `generateCheckSum()` directly (a `loadCustomChange` caller that does NOT go through `validate()`); uses `wrapper.@customChange` field-direct access to bypass the lazy-loading getter; asserts the in-scope load was prevented |

```
[INFO] Tests run: 28, Failures: 0, Errors: 0, Skipped: 0 -- in CustomChangeWrapperTest
[INFO] BUILD SUCCESS
```

## Things to be aware of

- **Default is unchanged.** Existing `customChange` users see no behaviour change. Flag must be explicitly set to `false` (via `liquibase.properties`, env-var, system property, or any standard `ConfigurationDefinition` source) to flip the gate.
- **Both gates fire before any static initializer runs.** The defense-in-depth design specifically prevents the load-time side effect — not just the cast or constructor invocation.
- **Single field-direct test-access nuance:** the `wrapper.@customChange` syntax in the fourth test is deliberate (and commented inline). The natural `wrapper.customChange` form calls the lazy-loading `getCustomChange()` getter, which re-attempts the load *outside* the `Scope.child` block (where the flag has reverted to `true`) and pollutes the assertion. The field-direct form proves the in-scope load was prevented.
- **Chain 3 partial fix.** The broader env-var command-injection chain tracked in DAT-23027 combines B1 (#7747), B2 (this PR), and B10 (env-var substitution on changeset properties). With B1 + B2 gated, the chain's two RCE primitives are both opt-out-able; closing the chain fully requires B10 to land too.

## Related

- **#7747** (DAT-23015, B1 / `executeCommand` opt-out gate) — same opt-in restricted-mode shape applied to the sibling RCE primitive (merged).
- **#7749** (DAT-23017, B3 / `customPrecondition` opt-out gate) — paired follow-up using the same `ALLOW_CUSTOM_CHANGE` flag (open, approved).

### DIFF
--- a/liquibase-standard/src/main/java/liquibase/GlobalConfiguration.java
+++ b/liquibase-standard/src/main/java/liquibase/GlobalConfiguration.java
@@ -53,6 +53,7 @@ public class GlobalConfiguration implements AutoloadedConfigurations {
     public static final ConfigurationDefinition<Boolean> STRICT;
     public static final ConfigurationDefinition<Integer> DDL_LOCK_TIMEOUT;
     public static final ConfigurationDefinition<Boolean> SECURE_PARSING;
+    public static final ConfigurationDefinition<Boolean> ALLOW_CUSTOM_CHANGE;
     public static final ConfigurationDefinition<String> SEARCH_PATH;
 
     public static final ConfigurationDefinition<UIServiceEnum> UI_SERVICE;
@@ -221,6 +222,19 @@ public class GlobalConfiguration implements AutoloadedConfigurations {
 
         SECURE_PARSING = builder.define("secureParsing", Boolean.class)
                 .setDescription("If true, remove functionality from file parsers which could be used insecurely. Examples include (but not limited to) disabling remote XML entity support.")
+                .setDefaultValue(true)
+                .build();
+
+        ALLOW_CUSTOM_CHANGE = builder.define("allowCustomChange", Boolean.class)
+                .setDescription("If false, the customChange changelog change is rejected before its named " +
+                        "class is loaded. Defaults to true to preserve the documented customChange feature " +
+                        "for the standard trust model (team-authored, team-reviewed changelogs). Set to " +
+                        "false in environments that execute changelogs from less-trusted sources " +
+                        "(multi-tenant SaaS running customer changelogs, downloaded change-packs, " +
+                        "contributor PRs prior to review): customChange loads an arbitrary JVM class by " +
+                        "FQCN via Class.forName(initialize=true), which fires the class's static <clinit> " +
+                        "initializer at load time — before any cast or marker-interface check could " +
+                        "reject the load. Any class on the JVM classpath is reachable this way (CWE-470).")
                 .setDefaultValue(true)
                 .build();
 
--- a/liquibase-standard/src/main/java/liquibase/change/custom/CustomChangeWrapper.java
+++ b/liquibase-standard/src/main/java/liquibase/change/custom/CustomChangeWrapper.java
@@ -1,5 +1,6 @@
 package liquibase.change.custom;
 
+import liquibase.GlobalConfiguration;
 import liquibase.Scope;
 import liquibase.change.*;
 import liquibase.database.Database;
@@ -83,6 +84,31 @@ public class CustomChangeWrapper extends AbstractChange {
     }
 
     private CustomChange loadCustomChange(String className) throws CustomChangeException {
+        // CWE-470 defense-in-depth gate. Every Class.forName(className, true, ...) below
+        // fires the named class's static <clinit> initializer at load time — BEFORE the
+        // cast to CustomChange or any other safety check could reject the load. Under
+        // the harder threat model (untrusted changelogs) that is a direct RCE primitive:
+        // any class on the JVM classpath can have its static init + constructor run
+        // merely by being named in the change tag. This gate fires BEFORE any of the
+        // three class-loading paths (OSGi / Scope classloader / thread context CL /
+        // bare Class.forName) so the static initializer never runs when the embedder
+        // has disabled customChange via liquibase.allowCustomChange=false.
+        //
+        // Mirrored in validate(Database) below for the standard hard-error operator
+        // surface; this method-level gate ALSO covers callers that don't go through
+        // validate(), notably customLoadLogic() which calls loadCustomChange during
+        // changelog PARSE-time (i.e. before validate() ever runs).
+        if (Boolean.FALSE.equals(GlobalConfiguration.ALLOW_CUSTOM_CHANGE.getCurrentValue())) {
+            throw new CustomChangeException(
+                    "customChange is disabled by configuration " +
+                            "(liquibase.allowCustomChange=false). To run customChange changes, set " +
+                            "liquibase.allowCustomChange=true (the default). This flag is provided for " +
+                            "environments that execute changelogs from less-trusted sources, where " +
+                            "loading an arbitrary JVM class by name via changelog is not an acceptable " +
+                            "risk (the load itself fires the class's static initializer before any " +
+                            "safety check could run).");
+        }
+
         try {
             Boolean osgiPlatform = Scope.getCurrentScope().get(Scope.Attr.osgiPlatform, Boolean.class);
             if (Boolean.TRUE.equals(osgiPlatform)) {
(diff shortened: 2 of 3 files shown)


filipelautert (commented):
The mitigation itself looks correct for the declared scope:

- Gate at `CustomChangeWrapper.java:100` fires before the OSGi → scope-CL → TCCL → bare `Class.forName(...,true,...)` chain (lines 113-122), so static `<clinit>` cannot execute when the flag is `false`. Mirror gate at `:170` short-circuits `validate()`. Covers all 5 `loadCustomChange` call sites including the parse-time path.
- Default-true preserves behavior — no regression risk for non-opt-in users.
- Tests are solid: default-pass, opt-out blocks, gate-fires-before-Class.forName via a non-existent class (proves short-circuit), and the field-direct access via `wrapper.@customChange` to dodge the lazy getter.

But — and this is the reason I'm leaving a comment rather than approving — **`<customPrecondition>` is the same CWE-470 sink and is NOT gated by `allowCustomChange`**. See `CustomPreconditionWrapper.java:67`:

```java
Class.forName(className, true, Scope.getCurrentScope().getClassLoader()).getConstructor().newInstance()
```

…with the identical static-initializer-on-load primitive, sourced from a changelog-supplied FQCN. A customer who flips `allowCustomChange=false` to harden against untrusted changelogs is going to *believe* they're locked down for "custom" loading, but they still have full reflective classload via `<customPrecondition>`.

I think this is your B2 slice paired with B1 (`executeCommand`) and the customPrecondition fix is queued separately — but if so, can you either:

(a) extend this PR's gate to `CustomPreconditionWrapper.check()` under the same `allowCustomChange` flag (or a sibling `allowCustomPrecondition`), or
(b) explicitly document in the flag's description that `<customPrecondition>` remains an unmitigated CWE-470 sink and is tracked in a follow-up ticket?

(a) is preferable because the flag name implies "all custom-* things are gated" and operators will read it that way.

Once one of those is in I'll approve.

v-petrovych:
Thanks for the careful read, @filipelautert — and you're right that operators reading `allowCustomChange` will expect it to cover all custom-* sinks. That's exactly the gap your option (a) closes.

**The customPrecondition extension is #7749** — sibling PR filed at the same time as this one specifically to close the `<customPrecondition>` half of CWE-470 under the same `ALLOW_CUSTOM_CHANGE` flag (per the audit author's explicit guidance on the source ticket: *"Covered by the same flag proposed for B2."*). You've already reviewed and approved it: https://github.com/liquibase/liquibase/pull/7749#pullrequestreview-4336812656

So the answer to your "(a) or (b)?" is **(a), executed in a separate PR for review hygiene** — same flag, same dual-gate pattern (`validate()` + `check()` short-circuiting before `Class.forName`), with the additional `PreconditionErrorException`-not-`PreconditionFailedException` choice on the second gate to pre-empt the `onFail=MARK_RAN` bypass attack vector.

**Merge order** (from your Slack coordination note): this PR merges first, then #7749 rebases against main and resolves the `GlobalConfiguration.java` collision by keeping #7749's broader description (which covers both customChange *and* customPrecondition — the correct end-state). I've already merged main into this branch to clear the conflict with #7747's `ALLOW_EXECUTE_COMMAND` (now in main); 28/28 `CustomChangeWrapperTest` specs still pass.

PR description updated with a paired-follow-up note linking to #7749 so the pairing is visible to any future reviewer who lands here cold. Ready for re-review.

filipelautert (approved):
Upgrading to APPROVE — the scope gap I flagged earlier (`<customPrecondition>` being an identical CWE-470 sink not gated by `allowCustomChange`) is closed in your sibling PR #7749, which I've already approved. Your reply confirmed the strategy is option (a) — same flag, dual-gate pattern, separate PR for review hygiene — which is exactly the right call.

Code in this PR is unchanged from my earlier read; the only new commit is the merge from main, which brought in #7747's `ALLOW_EXECUTE_COMMAND`. The merge-order plan you laid out on #7749 (this lands first with the narrow description, then #7749 rebases to resolve the `ALLOW_CUSTOM_CHANGE` duplicate-declaration conflict and updates the description to the both-elements form) is the cleanest sequencing.

Ship it.

PR #7749, opened 2026-05-21
CWE-470: opt-in restricted mode for customPrecondition changelog element

v-petrovych:
## Impact

`CustomPreconditionWrapper.check(...)` (`CustomPreconditionWrapper.java:67-69`) loads an arbitrary JVM class by FQCN via the same `Class.forName(className, true, ...)` pattern as `customChange`:

```java
customPrecondition = (CustomPrecondition) Class.forName(className, true,
//                                              ^ static <clinit> fires here, before the cast
```

Same Unsafe-Reflection attack surface as B2 (DAT-23016 / PR #7748) — **CWE-470**.

**An additional attack vector unique to this PR**, called out explicitly in the audit ticket: preconditions are evaluated **before** the change body executes. An attacker can craft a precondition with `onFail=MARK_RAN` that *"always passes"* and the change body is never reached — but the load-time `<clinit>` already ran. So the fix must reject the load **before** `Class.forName`, **not** rely on the precondition's pass/fail logic.

## Cross-PR coordination (please read)

This PR is the **third sibling** in a B1/B2/B3 pattern:

| PR | Element | Flag |
|---|---|---|
| #7747 | `<executeCommand>` | `liquibase.allowExecuteCommand` |
| #7748 | `<customChange>` | `liquibase.allowCustomChange` |
| **THIS** | **`<customPrecondition>`** | **`liquibase.allowCustomChange` (same as #7748)** |

The audit ticket's exact words: *"Covered by the same `liquibase.execution.allowCustomClassLoad=false` opt-in flag proposed for B2."* Both `customChange` and `customPrecondition` are "load arbitrary FQCN" surfaces with identical risk — one flag covers both.

### Expected merge conflict with #7748

Both this PR and #7748 declare `GlobalConfiguration.ALLOW_CUSTOM_CHANGE`. The declarations are byte-identical for the constant; the **description text** differs:

- **#7748's description** mentions only `customChange`.
- **THIS PR's description** mentions both `customChange` AND `customPrecondition` (the larger surface that's accurate after both PRs are in `main`).

**Recommended merge resolution:** when merging the second of the two PRs, keep this PR's broader description text and discard #7748's narrower one. The constant declaration itself dedupes mechanically (1-line edit). I considered stacking this PR on #7748's branch to avoid the conflict but chose independent PRs to keep each one reviewable in any merge order — the conflict is mechanical and 1-line.

## Fix — three pieces

**1. `GlobalConfiguration.ALLOW_CUSTOM_CHANGE`** — Boolean, defaults to `true`. Description covers both `customChange` and `customPrecondition` surfaces and ties them to CWE-470.

**2. Dual gate inside `CustomPreconditionWrapper`:**

   **(a) Hard error in `validate(Database)`.** The pre-fix `validate()` was a trivial no-op (`return new ValidationErrors();`). This PR replaces that body with a configured-off check, so the operator sees a pre-execution failure during the standard validation pass.

   **(b) Defense-in-depth gate at the top of `check(...)`** — fires BEFORE either of the two `Class.forName` call sites at lines 67/69. Throws `PreconditionErrorException`, **not** `PreconditionFailedException`. The choice is deliberate: `Error` bypasses `onFail` handling, so a crafted `onFail=MARK_RAN` precondition cannot silently swallow the embedder's configured-off intent — which is the specific attack vector the audit calls out.

**3. Four new Spock specs in `CustomPreconditionWrapperTest`** (2 pre-existing → 6 total):

| Spec | What it pins |
|---|---|
| `validate returns no errors by default` | Default-true path preserves existing behaviour |
| `validate fails when liquibase.allowCustomChange=false — embedder opt-out path` | Hard error, names the flag in both directions AND names `customPrecondition` specifically so the message points at the offending element |
| `check throws PreconditionErrorException BEFORE Class.forName when allowCustomChange=false` | Sets a non-existent class name to prove the gate ran first (no wrapped `ClassNotFoundException` appears). Asserts `PreconditionErrorException` not `Failed` — pins the audit's `onFail=MARK_RAN`-bypass requirement |
| `check default-true path still executes the existing load logic (sanity)` | Default-true short-circuit doesn't skip downstream behaviour |

```
[INFO] Tests run: 6, Failures: 0, Errors: 0, Skipped: 0 -- in CustomPreconditionWrapperTest
[INFO] BUILD SUCCESS
```

## Things to be aware of

- **Default is unchanged.** Existing `customPrecondition` users see byte-for-byte identical behaviour.
- **Both gates fire before any static initializer.** The defense-in-depth design specifically prevents the load-time side effect — not just the cast or constructor invocation.
- **PreconditionErrorException (not Failed) is the intentional choice** at the `check()` gate. A `PreconditionFailedException` would respect `onFail` handling, allowing `onFail=MARK_RAN` to silently swallow the embedder's configured-off intent — exactly the attack vector the audit calls out. The unit test pins this.
- **Chain 3 status:** with this PR (B3) landing alongside #7747 (B1) and #7748 (B2), **all three RCE primitives flagged in DAT-23027 (the env-var command-injection chain) now have embedder opt-out gates.** B10 (env-var substitution behaviour on changeset properties) is a separate child ticket; closing the chain fully requires B10 to land too.

## Related

- **Direct sibling**: #7748 (DAT-23016, B2 / `customChange` opt-out gate) — same flag, parallel implementation. The two are designed to land together.
- **Cousin**: #7747 (DAT-23015, B1 / `executeCommand` opt-out gate) — same opt-in restricted-mode shape, different flag and different change tag.
- Other audit slice PRs: #7739, #7740, #7741, #7742, #7743, #7744, #7746. Merged: #7737, #7738.

### DIFF
--- a/liquibase-standard/src/main/java/liquibase/GlobalConfiguration.java
+++ b/liquibase-standard/src/main/java/liquibase/GlobalConfiguration.java
@@ -53,6 +53,7 @@ public class GlobalConfiguration implements AutoloadedConfigurations {
     public static final ConfigurationDefinition<Boolean> STRICT;
     public static final ConfigurationDefinition<Integer> DDL_LOCK_TIMEOUT;
     public static final ConfigurationDefinition<Boolean> SECURE_PARSING;
+    public static final ConfigurationDefinition<Boolean> ALLOW_CUSTOM_CHANGE;
     public static final ConfigurationDefinition<String> SEARCH_PATH;
 
     public static final ConfigurationDefinition<UIServiceEnum> UI_SERVICE;
@@ -221,6 +222,20 @@ public class GlobalConfiguration implements AutoloadedConfigurations {
 
         SECURE_PARSING = builder.define("secureParsing", Boolean.class)
                 .setDescription("If true, remove functionality from file parsers which could be used insecurely. Examples include (but not limited to) disabling remote XML entity support.")
+                .setDefaultValue(true)
+                .build();
+
+        ALLOW_CUSTOM_CHANGE = builder.define("allowCustomChange", Boolean.class)
+                .setDescription("If false, the customChange and customPrecondition changelog elements are " +
+                        "rejected before the named class is loaded. Defaults to true to preserve the " +
+                        "documented custom-Java features for the standard trust model (team-authored, " +
+                        "team-reviewed changelogs). Set to false in environments that execute changelogs " +
+                        "from less-trusted sources (multi-tenant SaaS running customer changelogs, " +
+                        "downloaded change-packs, contributor PRs prior to review): both customChange " +
+                        "and customPrecondition load an arbitrary JVM class by FQCN via " +
+                        "Class.forName(initialize=true), which fires the class's static <clinit> " +
+                        "initializer at load time — before any cast or marker-interface check could " +
+                        "reject the load. Any class on the JVM classpath is reachable this way (CWE-470).")
                 .setDefaultValue(true)
                 .build();
 
--- a/liquibase-standard/src/main/java/liquibase/precondition/CustomPreconditionWrapper.java
+++ b/liquibase-standard/src/main/java/liquibase/precondition/CustomPreconditionWrapper.java
@@ -1,5 +1,6 @@
 package liquibase.precondition;
 
+import liquibase.GlobalConfiguration;
 import liquibase.Scope;
 import liquibase.changelog.ChangeSet;
 import liquibase.changelog.visitor.ChangeExecListener;
@@ -54,12 +55,47 @@ public class CustomPreconditionWrapper extends AbstractPrecondition {
 
     @Override
     public ValidationErrors validate(Database database) {
+        // CWE-470 hard gate, paired with the defense-in-depth gate inside check().
+        // Hard error so the operator sees a clean pre-execution failure when the
+        // embedder has disabled custom-Java elements via
+        // liquibase.allowCustomChange=false. Naming-note: the flag is shared with
+        // customChange (see GlobalConfiguration.ALLOW_CUSTOM_CHANGE) because both
+        // elements expose identical Class.forName(initialize=true) load surfaces.
+        if (Boolean.FALSE.equals(GlobalConfiguration.ALLOW_CUSTOM_CHANGE.getCurrentValue())) {
+            return new ValidationErrors().addError(
+                    "customPrecondition is disabled by configuration " +
+                            "(liquibase.allowCustomChange=false). To run customPrecondition (and " +
+                            "customChange) elements, set liquibase.allowCustomChange=true (the default). " +
+                            "This flag is provided for environments that execute changelogs from " +
+                            "less-trusted sources, where loading an arbitrary JVM class by name via " +
+                            "changelog is not an acceptable risk.");
+        }
         return new ValidationErrors();
     }
 
     @Override
     public void check(Database database, DatabaseChangeLog changeLog, ChangeSet changeSet, ChangeExecListener changeExecListener)
             throws PreconditionFailedException, PreconditionErrorException {
+        // CWE-470 defense-in-depth gate: fire BEFORE any of the two Class.forName
+        // call sites below. The audit notes that preconditions are evaluated
+        // before changes execute, providing a separate class-load trigger point
+        // that fires even if the change body is never reached — e.g. an
+        // onFail=MARK_RAN precondition that "always passes" but whose load-time
+        // static <clinit> already ran. Throwing PreconditionErrorException
+        // (not PreconditionFailedException) bypasses onFail handling — we do NOT
+        // want MARK_RAN to silently swallow the embedder's configured-off intent.
+        if (Boolean.FALSE.equals(GlobalConfiguration.ALLOW_CUSTOM_CHANGE.getCurrentValue())) {
+            throw new PreconditionErrorException(new ErrorPrecondition(
+                    new CustomPreconditionErrorException(
+                            "customPrecondition is disabled by configuration " +
+                                    "(liquibase.allowCustomChange=false). To run customPrecondition (and " +
+                                    "customChange) elements, set liquibase.allowCustomChange=true (the " +
+                                    "default). This flag is provided for environments that execute " +
+                                    "changelogs from less-trusted sources, where loading an arbitrary " +
+                                    "JVM class by name via changelog is not an acceptable risk."),
+                    changeLog, this));
+        }
+
         CustomPrecondition customPrecondition;
         try {
 //            System.out.println(classLoader.toString());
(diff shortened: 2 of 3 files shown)


v-petrovych:
@coderabbitai — addressed the hermetic-scope nit in commit pushed just now.

You're right that the two default-true specs (`validate returns no errors by default ...` and `check default-true path still executes the existing load logic ...`) were reading the ambient `ALLOW_CUSTOM_CHANGE` value. The flag-`false` specs already use explicit `Scope.child([..."false"], ...)`; the symmetric fix is to wrap the default-`true` specs the same way so each spec is hermetic w.r.t. whatever the runner / a sibling test / a stale system property left in ambient scope.

```diff
     when:
-    def errors = precondition.validate(new MockDatabase())
+    def errors = Scope.child([(GlobalConfiguration.ALLOW_CUSTOM_CHANGE.getKey()): "true"],
+            { return precondition.validate(new MockDatabase()) } as Scope.ScopedRunnerWithReturn)
```

(Same pattern applied to the `check` spec.)

Added an inline comment on each spec crediting your review and explaining *why* the explicit-scope form is required (not just default behaviour), so a future maintainer doesn't "simplify" back to the ambient form.

6/6 `CustomPreconditionWrapperTest` specs green after the change.

filipelautert (approved):
LGTM — this is the follow-up I asked for on #7748, executed cleanly and with thought.

Specifically:

- **Shared flag, as suggested.** `ALLOW_CUSTOM_CHANGE` gates both customChange and customPrecondition, so operators who flip the flag to harden actually get the lockdown the flag name implies. Description text updated to cover both elements.
- **Hard gate + defense-in-depth.** `validate()` at CustomPreconditionWrapper.java:55-66 surfaces a clean pre-execution error; `check()` at :73-86 short-circuits before either `Class.forName` call site. Order is right.
- **MARK_RAN attack vector addressed pre-emptively.** Throwing `PreconditionErrorException` (not `PreconditionFailedException`) is the right call — `Error` bypasses `onFail` handling, so a crafted `onFail=MARK_RAN` precondition cannot silently swallow the embedder's configured-off intent. This is exactly the concern I raised on #7747's gate; nice to see it pre-empted here.
- **Tests are hermetic.** Explicit `Scope.child([... : "true"], ...)` on the default-pass specs (vs. reading the ambient default) means a stale system property or test-ordering quirk can't turn the intended default-true path into a spurious failure.
- **Non-existent-classname proof.** The "check throws BEFORE Class.forName" spec uses `com.example.definitely.not.a.real.CustomPrecondition` and asserts the thrown error mentions the configured-off cause, not a `ClassNotFoundException` — that's the right shape to prove the gate ran first.

One coordination note (not a code issue): **#7748 also adds `ALLOW_CUSTOM_CHANGE` to `GlobalConfiguration.java` at the same line, but with a customChange-only `setDescription()` text.** Whichever PR merges second will need a rebase to resolve. Suggest noting on both PRs which order is preferred — I'd lean toward merging #7748 first (the simpler slice landed first) and then this PR rebases to update the description text and add the gate+tests, since the final description text here is the correct end-state.

Approving.

v-petrovych:
Thanks for the approval, @filipelautert — and for the line-by-line read on the MARK_RAN attack vector + hermetic test scoping. Those were the two non-obvious calls in this PR and getting independent confirmation on both is reassuring.

On your coordination note (the `GlobalConfiguration.java` collision with #7748):

> *Suggest noting on both PRs which order is preferred — I'd lean toward merging #7748 first (the simpler slice landed first) and then this PR rebases to update the description text and add the gate+tests, since the final description text here is the correct end-state.*

Agreed. Plan:

1. **#7748 merges first** with its narrow customChange-only description.
2. **This PR (#7749) then takes the main update**, and resolves the `ALLOW_CUSTOM_CHANGE` duplicate-declaration conflict by **keeping this PR's broader description** (the one that mentions both customChange *and* customPrecondition) — that's the correct end-state per your note.

I've already merged main into this branch once to clear the unrelated conflict with #7747's `ALLOW_EXECUTE_COMMAND` (now in main). The inter-PR conflict with #7748 will materialize only after #7748 lands; I'll do a second merge pass at that point. Per @v-petrovych's "no force-push" rule I'm using `git merge` rather than `git rebase` — the diff stays clean either way and the PR head doesn't get rewritten.

6/6 `CustomPreconditionWrapperTest` specs pass after the current merge. Ready when #7748 lands.

filipelautert (approved):
Re-approving at `cd31cdb4b7` — the only new commit since my earlier approval is the merge-from-main (looks like branch protection dismissed the stale review on push). Code is unchanged; the merge just brought in #7747's `ALLOW_EXECUTE_COMMAND` to the surrounding context, which doesn't touch the customPrecondition gate.

Merge-order plan in your reply matches what I suggested — #7748 lands first with the narrow description, this PR then rebases to resolve the `ALLOW_CUSTOM_CHANGE` duplicate-declaration and the description becomes the both-elements end-state. LGTM.

v-petrovych:
Second merge pass complete now that #7748 has landed.

**Conflict scope was smaller than expected:** the `ALLOW_CUSTOM_CHANGE` *declaration line* is byte-identical between this branch and #7748 (both are `public static final ConfigurationDefinition<Boolean> ALLOW_CUSTOM_CHANGE;`), so git auto-merged it cleanly. The only actual conflict was in the `builder.setDescription(...)` text — and per the coordination plan in your review, the resolution keeps **this branch's broader description** (covers both `customChange` AND `customPrecondition`) and discards #7748's narrower customChange-only text. Net effect on main once this lands: the description text upgrades from narrow → broader.

Everything else from #7748 (the `CustomChangeWrapper` dual gate, its 4 new test specs) is pulled in unchanged.

**State:** MERGEABLE, BLOCKED only on review (the merge commit dismissed your earlier approval — that's GitHub branch-protection behaviour, not a code issue). 45/45 specs pass across `CustomPreconditionWrapperTest` (6, this PR), `CustomChangeWrapperTest` (28, from #7748), and `GlobalConfigurationTest` (11).

No force-push — used `git merge` per @v-petrovych's no-force-push rule. Two merge commits now appear on the branch (the earlier one resolved the #7747 / `ALLOW_EXECUTE_COMMAND` conflict; this latest one resolves the #7748 / `ALLOW_CUSTOM_CHANGE` description-text conflict).

Ready for re-approval whenever.

v-petrovych:
Refiled as #7768 — same code (SHA `e0b8da35` preserved), branch now pushed directly to `liquibase/liquibase` instead of from the `v-petrovych/liquibase` fork.

Direct-push routes `pull_request_target` to `environment: internal` (rather than the gated `external` env that held all CI in `waiting` state on fork PRs). CI fires automatically on PR-open — verified on #7768, no maintainer click needed. This is the structural fix for the failure mode behind [DAT-23097](https://datical.atlassian.net/browse/DAT-23097).

Closing in favor of #7768.

PR #7750, opened 2026-05-21
CWE-22 deprecation flag: opt-in path containment via liquibase.allowParentDirectoryReferences

v-petrovych:
## TL;DR

PR #7729 ("AbstractPathResourceAccessor: confine path resolution to configured root") shipped a correct CWE-22 fix but its `startsWith(rootPath)` containment check is too coarse: it rejects *every* `..` payload regardless of whether the resolved path still lives inside the user's implicit project trust boundary. The result is four customer-visible regressions (one of them a **silent disable of policy checks with exit code 0**) for legitimate multi-changelog layouts.

This PR adds **`liquibase.allowParentDirectoryReferences`** (Boolean, default `true`) as a deprecation-window escape hatch. With the flag at its default, both containment checks introduced in #7729 are skipped and behaviour matches the pre-CWE-22-fix path. Embedders that want the strict CWE-22 hardening flip the flag to `false`. A future major release will flip the default.

No production-side behaviour change for any caller that flips the flag explicitly. Default-true preserves customer compatibility for one major release.

### 1. CRITICAL — Pro custom Python policy checks silently disabled

Customers with the very common "shared `dbarepo` at the project root, per-environment changelogs underneath" layout reference `SCRIPT_PATH: ../dbarepo/scripts/foo.py` from inside a project-local changelog. Post-#7729 the `..` payload is rejected with `IOException`, which is silently caught by `CustomCheck.getResource` in liquibase-pro and converted to a `MissingScriptResource`. The check is then recorded as non-applicable with reason `SCRIPT_NOT_FOUND` and `returnAtEndOfRule(..., false)` — meaning no violation fires. **Exit code stays 0.** CI keeps passing while policy regressions are no longer caught. The worst-case failure mode of #7729.

### 2. HIGH — OSS changelog `<include>` with the default `relativeToChangelogFile`

`<include file="../shared/foo.xml" relativeToChangelogFile="false">` (the default!) doesn't go through `DatabaseChangeLog.normalizePath()`, so the `..` reaches `resourceAccessor.get()` intact and the `IOException` propagates up as `ChangeLogParseException` or `UnexpectedLiquibaseException`. Broadest blast radius — hits OSS users too, not just Pro. `relativeToChangelogFile="true"` is *not* affected because `normalizePath()` collapses the `..` before resolution.

### 3. HIGH — Pro Flow `<include path="../shared/foo.yaml">`

`ResourceUtil.getResource` → `resourceAccessor.get` → `getAll` → `IOException`. The exception propagates *before* the `instanceof NotFoundResource && PRO_STRICT` gate is evaluated, so the `PRO_STRICT=false` "silent" branch becomes dead code for `..` payloads. Hard fail regardless of `PRO_STRICT`.

### 4. MEDIUM — Pro Flow `<conditions>` (`fileExists('../shared/foo')`)

`FlowFileUtil.exists()` invoked via MVEL. The `IOException` is wrapped as `RuntimeException` from `MVEL.executeExpression`. `Condition.evaluate()` only catches `CompileException`, so the wrapped exception reaches the flow stage runner. Hard fail.

## Root cause

PR #7729's `startsWith(rootPath)` check is correct for catching `../../etc/passwd`-style absolute traversal beyond the *configured root*. But the configured root (`AbstractPathResourceAccessor.getRootPath()`) is typically *the changelog directory* or another tightly-scoped directory — **not** "the entire trust boundary the user implicitly considers safe". The user's mental model is "anywhere under my project root is fine"; the code's containment is narrower.

Conflating these two boundaries means every legitimate `..` traversal that walks up one level inside the user's project is indistinguishable, to the check, from a `../../etc/passwd` attack.

## The flag

```java
GlobalConfiguration.ALLOW_PARENT_DIRECTORY_REFERENCES
// liquibase.allowParentDirectoryReferences
// Boolean, default TRUE
```

| Flag value | Syntactic `startsWith(rootPath)` check | Canonical `toRealPath()` symlink check | Behaviour matches |
|---|---|---|---|
| `true` (default) | SKIPPED | SKIPPED | pre-#7729 (legitimate `..` and symlinks work) |
| `false` (opt-in hardening) | ENFORCED | ENFORCED | #7729 strict CWE-22 |

**Why a single flag, not two:** Wesley's option 3 proposal explicitly framed it as a single escape hatch. The trusted-context allowlist (option 2 — accept `..` within a configurable project boundary while keeping symlink-escape rejection on by default) is deferred to a follow-up; it requires deciding on the boundary surface and is more invasive than this PR's scope.

**Why default `true`, not `false`:** A default-false would ship the regression Wesley flagged on the very first upgrade. The deprecation-window approach (default-true now, flip to false in a future major) gives customers a release cycle to either restructure their layouts or explicitly opt in. Embedders who *already* want CWE-22 hardening flip the flag immediately.

## Deprecation timeline

- **This release**: default `true`. CWE-22 hardening is opt-in. Customer compat preserved.
- **Future major release**: default flips to `false`. CWE-22 hardening becomes default. Customers depending on `..` traversal must restructure their layout or explicitly opt in via `liquibase.allowParentDirectoryReferences=true`.

This matches the "ship a strong security default with a one-major-release deprecation window" pattern used elsewhere in Liquibase (e.g. `liquibase.secureParsing`).

## Test plan

- **Existing PR #7729 strict-mode specs (5 in `DirectoryResourceAccessorTest`, 1 in `SearchPathResourceAccessorTest`)** wrapped in `Scope.child(allowParentDirectoryReferences=false)`. Strict-mode coverage preserved verbatim, just gated. Test names updated to mention "when allowParentDirectoryReferences=false" for clarity.
- **New default-mode specs** (3 added):
  - `getAll allows ../ payloads that resolve outside the root when allowParentDirectoryReferences=true` — real temp-dir sibling, mirrors Wesley's `dbarepo` scenario
  - `search allows ../ startPath that resolves outside the root when allowParentDirectoryReferences=true` — `search()` counterpart
  - `getAll follows a symlink that escapes the accessor root when allowParentDirectoryReferences=true` — Linux-only (Windows lacks `SeCreateSymbolicLinkPrivilege`), mirrors the strict-mode symlink-rejection spec
- **`getAll allows paths that traverse via .. but stay inside the root`** — unchanged; that behaviour is flag-independent (the `..` collapses pre-resolve into a clean inside-root path).
- **Broader regression check**: ran every `*ResourceAccessor*Test` plus `DatabaseChangeLogTest`, `IncludeAllTest`, `ChangeLogParserConfigurationTest`, `XMLChangeLogSAXParserTest`, and `GlobalConfigurationTest`. **337 tests across the suite — 0 failures, 2 skipped (the Windows-conditional symlink specs).**

```
[INFO] Tests run: 337, Failures: 0, Errors: 0, Skipped: 2
[INFO] BUILD SUCCESS
```

## Things to be aware of

- **No production behaviour change at the default flag value.** Anyone whose changelogs / flow files / policy checks worked pre-#7729 continues to work after this lands.
- **PR #7729 stays merged.** This PR doesn't revert anything — it adds the opt-in gate around the checks PR #7729 introduced.
- **Symlink coverage is part of the single flag.** A customer using legitimate filesystem symlinks (e.g. shared resources via symlink, common in monorepo-ish layouts) is also restored by `flag=true`. If a future user wants `..` allowed but symlinks still blocked, the option-2 follow-up will offer that granularity.
- **Zip / `ZipResourceAccessor` is unchanged.** Per #7729's own design note, zipfs `normalize()` already clamps `..` at root, so the containment check is a no-op there. The new flag is read but has no effect for zip.
- **Performance**: one `ConfigurationDefinition.getCurrentValue()` lookup per `getAll`/`search` call. These are not inner-loop calls (one per changelog include, one per script-path resolution). Negligible.

## Related

Sibling to the May-2026 OSS credential-handling-and-changelog-injection audit slice (`sdou` label). PR #7729 is the parent fix; this PR is the compat-preserving deprecation gate.

Other audit PRs (open at time of writing): #7748 (CWE-470 customChange, merged just before this), #7749 (CWE-470 customPrecondition, awaiting re-review). Merged: #7729, #7737, #7738, #7739, #7740, #7741, #7742, #7743, #7747.

### DIFF
--- a/liquibase-standard/src/main/java/liquibase/GlobalConfiguration.java
+++ b/liquibase-standard/src/main/java/liquibase/GlobalConfiguration.java
@@ -55,6 +55,7 @@ public class GlobalConfiguration implements AutoloadedConfigurations {
     public static final ConfigurationDefinition<Boolean> SECURE_PARSING;
     public static final ConfigurationDefinition<Boolean> ALLOW_CUSTOM_CHANGE;
     public static final ConfigurationDefinition<Boolean> ALLOW_EXECUTE_COMMAND;
+    public static final ConfigurationDefinition<Boolean> ALLOW_PARENT_DIRECTORY_REFERENCES;
     public static final ConfigurationDefinition<String> SEARCH_PATH;
 
     public static final ConfigurationDefinition<UIServiceEnum> UI_SERVICE;
@@ -247,6 +248,23 @@ public class GlobalConfiguration implements AutoloadedConfigurations {
                         "that execute changelogs from less-trusted sources (multi-tenant SaaS running customer " +
                         "changelogs, downloaded change-packs, contributor PRs prior to review) where arbitrary " +
                         "OS-shell execution via changelog is not an acceptable risk (CWE-78).")
+                .setDefaultValue(true)
+                .build();
+
+        ALLOW_PARENT_DIRECTORY_REFERENCES = builder.define("allowParentDirectoryReferences", Boolean.class)
+                .setDescription("If true (the default), AbstractPathResourceAccessor allows path payloads " +
+                        "containing '..' segments and symbolic links that resolve outside the configured " +
+                        "root directory. This preserves the behaviour that existed before the CWE-22 " +
+                        "path-containment fix landed, for legitimate multi-changelog layouts that depend on " +
+                        "parent-directory traversal — for example a shared 'dbarepo' at the project root " +
+                        "referenced as '../shared/foo.xml' from per-environment changelogs underneath it, " +
+                        "or a custom-check SCRIPT_PATH like '../checks/policy.py'. Set to false to enforce " +
+                        "strict containment: any '..' that resolves outside the accessor root, or any " +
+                        "symbolic link whose canonical real path escapes the canonical root, is rejected " +
+                        "with IOException. The default is true for one major release as a deprecation " +
+                        "window; a future major release will flip the default to false, at which point " +
+                        "callers depending on parent-directory traversal must either restructure their " +
+                        "layout or explicitly opt in via this flag (CWE-22).")
                 .setDefaultValue(true)
                 .build();
 
--- a/liquibase-standard/src/main/java/liquibase/resource/AbstractPathResourceAccessor.java
+++ b/liquibase-standard/src/main/java/liquibase/resource/AbstractPathResourceAccessor.java
@@ -1,5 +1,6 @@
 package liquibase.resource;
 
+import liquibase.GlobalConfiguration;
 import liquibase.Scope;
 import liquibase.logging.Logger;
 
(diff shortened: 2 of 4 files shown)


filipelautert (approved):
LGTM — approving. This is the right rollback. The CRITICAL silent-disable-of-policy-checks regression you flagged (Pro custom Python checks returning exit 0 with no enforcement) is genuinely worse than the original CWE-22 attack surface for the customers it hits. Default-true with a one-major deprecation window is the correct tradeoff; matches the `liquibase.secureParsing` precedent.

Verified the rollback is complete — all four CWE-22 containment points are gated by the same `strictContainment` boolean:
- `getAll()` syntactic check at `AbstractPathResourceAccessor.java:88-99`
- `getAll()` canonical/symlink check at `:107-113`
- `search()` startPath check at `:130`
- `search()` per-entry `preVisitDirectory`/`visitFile` checks at `:151,160`

Nice micro-optimization at `:142` — `rootRealPath = strictContainment ? rootPath.toRealPath() : null` skips the I/O when the flag is true.

`Boolean.TRUE.equals(getCurrentValue())` (not the bare unboxed negation) handles a null config value defensively — same pattern you used for `ALLOW_CUSTOM_CHANGE`. Consistent and safe.

Tests are symmetric: all 6 strict-mode specs from #7729 preserved (just wrapped in `Scope.child([...=false], ...)`), plus 3 new default-mode specs covering the actual regression scenarios. Counter-spec for the symlink case is also preserved.

One non-blocking observation: the PR-body regression scenarios are all end-to-end (silent policy-check disable, OSS `<include>` crash, Flow `<include>` crash, Flow `<conditions>` crash). The unit tests here verify the `AbstractPathResourceAccessor` primitive under both flag values, but the actual "Pro `SCRIPT_PATH: ../dbarepo/scripts/foo.py` returns exit 0 silently" path isn't tested in OSS (can't be — it lives in liquibase-pro's `CustomCheck.getResource`).

When the future major flips the default, those four paths are exactly where customer reports will resurface. Worth a paired liquibase-pro integration-test ticket before the flip — verifying that under `allowParentDirectoryReferences=false`, the SCRIPT_PATH scenario produces a **loud** failure (not the silent `MissingScriptResource` → exit 0 path). Otherwise we're betting the next major on "customers will read the release notes and opt in." Not a blocker for this PR.

Ship it.

PR #7754, opened 2026-05-22
CWE-470: opt-in restricted mode for includeAll resourceFilter and resourceComparator classes

v-petrovych:
## Impact

The `<includeAll>` changelog directive accepts `resourceFilter` and `resourceComparator` attributes that specify fully-qualified Java class names. Both are loaded via `Class.forName(name, true, ContextClassLoader).getConstructor().newInstance()` with no allowlist and no marker-interface check before construction:

```java
// DatabaseChangeLog.java:641 — resourceFilter
resourceFilter = (IncludeAllFilter) Class.forName(resourceFilterDef, true,
    Thread.currentThread().getContextClassLoader()).getConstructor().newInstance();

// DatabaseChangeLog.java:731 — resourceComparator
resourceComparator = (Comparator<String>) Class.forName(resourceComparatorDef, true,
    Thread.currentThread().getContextClassLoader()).getConstructor().newInstance();
```

The `initialize=true` argument fires the named class's static `<clinit>` initializer **at load time** — before the cast or any other safety check could reject the load. Any class on the JVM classpath is reachable via a malicious `<includeAll>`: name the class, get its static-init code and zero-arg constructor invoked as a side effect.

This maps to **CWE-470** (Use of Externally-Controlled Input to Select Classes or Code, aka *Unsafe Reflection*) — the same surface as `<customChange>` and `<customPrecondition>`, applied to a different parse-time entry point.

## Scope decision (sibling to #7748 and #7749) — separate flag, not shared

The audit author originally suggested unifying B2/B3/B5 under one flag (`liquibase.execution.allowCustomClassLoad`). After deliberation this PR uses a **separate flag** for two reasons:

1. **`<includeAll>` is not a "custom-*" element.** Extending `liquibase.allowCustomChange` (which gates `<customChange>` and `<customPrecondition>`) to also gate `<includeAll>` would stretch the flag name beyond what an operator reading `allowCustomChange` would reasonably expect.
2. **Distinct flags give operators independent control surface.** An embedder who uses `<customChange>` but not `<includeAll>` filters (or vice versa) can lock down each independently.

To prevent the "operator missed a flag" footgun that motivated #7749 sharing #7748's flag, **this PR's new-flag description cross-references `liquibase.allowCustomChange`**. Operators reading either description learn that fully locking down changelog-controlled class loading requires setting **both** flags to `false`. Open to feedback if reviewers prefer the unified-flag approach instead — the swap would be mechanical.

**No default behaviour change.** Every existing `<includeAll>` continues to work byte-for-byte identically at the default flag value.

## Fix — three pieces

**1. New `GlobalConfiguration.ALLOW_INCLUDE_ALL_CLASSES` flag** — `liquibase.allowIncludeAllClasses`, Boolean, defaults to `true`. Description references CWE-470 and cross-references `liquibase.allowCustomChange`.

**2. Two inline gates in `DatabaseChangeLog`:**

   **(a) In `handleIncludeAll()`** at the `resourceFilter` `Class.forName` call site (line 641 in the audit reference). Throws `SetupException` with the configured-off message **BEFORE** the `Class.forName`, so the static initializer of the named class never fires when the flag is `false`.

   **(b) In `determineResourceComparator()`** at the `resourceComparator` `Class.forName` call site (line 731 in the audit reference). Throws `UnexpectedLiquibaseException` (the method has no checked-throws clause). **Critically, the gate fires BEFORE the `try/catch` block** — the existing `ReflectiveOperationException`-catches-and-falls-back-to-standard logic would otherwise silently degrade the configured-off intent into the default comparator (which would mask the rejection from the operator and look identical to the success path in logs).

**3. Five new Spock specs** in `DatabaseChangeLogTest` (195 pre-existing → 200 total):

| Spec | What it pins |
|---|---|
| `determineResourceComparator returns the standard comparator when allowIncludeAllClasses=true (default) and resourceComparatorDef is null` | Default-path preservation. Behaviour-asserts via `comparator.compare(...)` rather than reference-equality because `getStandardChangeLogComparator()` returns a fresh `Comparator.comparing(...)` lambda each call. |
| `determineResourceComparator falls back to standard comparator at default flag when resourceComparator class fails to load (preserves pre-fix catch behaviour)` | flag=true must not regress the existing graceful-fallback path. |
| `determineResourceComparator throws UnexpectedLiquibaseException BEFORE Class.forName when allowIncludeAllClasses=false` | The strongest assertion in this section: uses a non-existent class name. Pre-fix code path would have fallen back to standard (no exception); after-fix produces ONLY the configured-off error — the loader never ran. Message-shape match (flag-name in both directions + `"NOT loaded"` + `"resourceComparator"`) is the proof-of-short-circuit. |
| `determineResourceComparator with null resourceComparatorDef returns standard comparator even when allowIncludeAllClasses=false` | Pins that the gate does NOT over-reach into the default-comparator path when there's nothing to gate. |
| `handleIncludeAll throws SetupException BEFORE Class.forName when allowIncludeAllClasses=false and resourceFilter is specified` | Parser-path test exercising the inline gate at line 641 via the standard `load()` entry point with a `ParsedNode`-based changelog. |

```
[INFO] Tests run: 200, Failures: 0, Errors: 0, Skipped: 0 -- in DatabaseChangeLogTest
[INFO] Tests run: 11, Failures: 0, Errors: 0, Skipped: 0 -- in GlobalConfigurationTest
[INFO] BUILD SUCCESS
```

Broader sanity sweep: **262 tests** across `DatabaseChangeLogTest` (200), `XMLChangeLogSAXParser_RealFile_Test` (30), and `GlobalConfigurationTest` (11) — **0 failures**, 1 platform-conditional skip.

## Things to be aware of

- **Default is unchanged.** Existing `<includeAll>` users see no behaviour change.
- **Both gates fire before any static initializer runs.** The defense-in-depth design specifically prevents the load-time side effect.
- **`determineResourceComparator` is `public`.** External callers passing a class name will now hit the gate; null-comparator callers are unaffected. This is intentional — any code path that passes a changelog-controlled FQCN to this method should respect the flag.
- **Cross-reference convention.** This PR adds the cross-reference only one-way (the new flag's description points at `liquibase.allowCustomChange`). The reverse cross-reference (updating `liquibase.allowCustomChange` to point at this new flag) is deferred to avoid stacking yet another description-text conflict on top of the #7748 → #7749 merge sequence already in flight. Happy to do that cleanup pass after both PRs land if useful.

## Related

- **#7747** (DAT-23015, B1 / `executeCommand` opt-out gate) — merged
- **#7748** (DAT-23016, B2 / `customChange` opt-out gate) — merged
- **#7749** (DAT-23017, B3 / `customPrecondition` opt-out gate) — open, approved by @filipelautert
- **#7750** (DAT-23090, CWE-22 deprecation flag) — merged
- **#7753** (DAT-23018, B4 / `sqlCheck` opt-out gate) — open

This PR closes B5. Remaining B-series children: B6 (sqlFile/include classpath URIs), B7 (SQLFile checksum), B8 (XML SAX secure-mode opt-out), B9 (DATABASECHANGELOG INSERT inlining), B10 (informational).

### DIFF
--- a/liquibase-standard/src/main/java/liquibase/GlobalConfiguration.java
+++ b/liquibase-standard/src/main/java/liquibase/GlobalConfiguration.java
@@ -55,6 +55,7 @@ public class GlobalConfiguration implements AutoloadedConfigurations {
     public static final ConfigurationDefinition<Boolean> SECURE_PARSING;
     public static final ConfigurationDefinition<Boolean> ALLOW_CUSTOM_CHANGE;
     public static final ConfigurationDefinition<Boolean> ALLOW_EXECUTE_COMMAND;
+    public static final ConfigurationDefinition<Boolean> ALLOW_INCLUDE_ALL_CLASSES;
     public static final ConfigurationDefinition<Boolean> ALLOW_PARENT_DIRECTORY_REFERENCES;
     public static final ConfigurationDefinition<String> SEARCH_PATH;
 
@@ -248,6 +249,23 @@ public class GlobalConfiguration implements AutoloadedConfigurations {
                         "that execute changelogs from less-trusted sources (multi-tenant SaaS running customer " +
                         "changelogs, downloaded change-packs, contributor PRs prior to review) where arbitrary " +
                         "OS-shell execution via changelog is not an acceptable risk (CWE-78).")
+                .setDefaultValue(true)
+                .build();
+
+        ALLOW_INCLUDE_ALL_CLASSES = builder.define("allowIncludeAllClasses", Boolean.class)
+                .setDescription("If false, the includeAll changelog directive's resourceFilter and " +
+                        "resourceComparator attributes are rejected when they reference a class name. " +
+                        "Defaults to true to preserve the documented includeAll feature for the standard " +
+                        "trust model (team-authored, team-reviewed changelogs). Set to false in environments " +
+                        "that execute changelogs from less-trusted sources (multi-tenant SaaS running customer " +
+                        "changelogs, downloaded change-packs, contributor PRs prior to review): both " +
+                        "attributes load an arbitrary JVM class by FQCN via Class.forName(initialize=true), " +
+                        "which fires the class's static <clinit> initializer at load time — before any cast " +
+                        "or marker-interface check could reject the load. Any class on the JVM classpath is " +
+                        "reachable this way (CWE-470). This flag governs the same unsafe-reflection surface " +
+                        "as liquibase.allowCustomChange (which gates the <customChange> and " +
+                        "<customPrecondition> changelog elements); operators wanting to fully lock down " +
+                        "changelog-controlled class loading must set both flags to false.")
                 .setDefaultValue(true)
                 .build();
 
--- a/liquibase-standard/src/main/java/liquibase/changelog/DatabaseChangeLog.java
+++ b/liquibase-standard/src/main/java/liquibase/changelog/DatabaseChangeLog.java
@@ -637,6 +637,18 @@ public class DatabaseChangeLog implements Comparable<DatabaseChangeLog>, Conditi
         }
         IncludeAllFilter resourceFilter = null;
         if (resourceFilterDef != null) {
+            // CWE-470 gate: reject changelog-controlled FQCN loading when the embedder has
+            // explicitly opted out. The Class.forName(initialize=true) below would fire the
+            // named class's static <clinit> at load time, before the cast to IncludeAllFilter
+            // could reject it — same unsafe-reflection surface as <customChange> and
+            // <customPrecondition>, gated by liquibase.allowCustomChange.
+            if (!Boolean.TRUE.equals(GlobalConfiguration.ALLOW_INCLUDE_ALL_CLASSES.getCurrentValue())) {
+                throw new SetupException("includeAll with a resourceFilter class is disabled " +
+                        "because liquibase.allowIncludeAllClasses=false. The named class was " +
+                        "NOT loaded. Either remove the resourceFilter / filter attribute from " +
+                        "the includeAll directive, or set liquibase.allowIncludeAllClasses=true " +
+                        "to enable changelog-controlled class loading by includeAll.");
+            }
             try {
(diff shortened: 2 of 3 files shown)


filipelautert (approved):
LGTM, approving. Two reflective-classload sinks (`resourceFilter`, `resourceComparator`) cleanly gated.

Findings:

- **Gates fire before `Class.forName(..., initialize=true)` on both sinks.** `DatabaseChangeLog.java:645` precedes the `resourceFilter` loader at `:653`; `:747` precedes the `resourceComparator` loader at `:756`. Neither static `<clinit>` can fire when the flag is false.
- **Subtle correctness point: the comparator gate is OUTSIDE the existing `try { ... } catch (ReflectiveOperationException)`.** If it had been inside, the existing graceful-fallback-to-standard logic would have silently swallowed the configured-off intent — operators would see no error, just a quietly-replaced comparator. Glad you called this out explicitly in the PR body.
- **Strict-mode test** uses a non-existent FQCN to prove the gate fires before the loader (pre-fix the `ClassNotFoundException` would have been caught and the standard comparator returned, no exception). Message-shape assertion is real proof-of-short-circuit. Null-def regression test pins the gate doesn't over-reach.
- **Flag-sharing call.** I'd personally lean toward sharing `ALLOW_CUSTOM_CHANGE` so a security-conscious operator gets "lock down all changelog-driven classloading" as one switch, but introducing `ALLOW_INCLUDE_ALL_CLASSES` separately is defensible — `<includeAll>` semantically isn't a "custom-*" element, and you cross-reference both flags in their descriptions so an operator reading either learns they need both. Acceptable. If a future audit consolidates into one umbrella flag, this PR's surface area makes that mechanical. Not blocking.

Ship it.

v-petrovych:
Refiled as #7766 — same code (SHA `1caee78a` preserved), branch now pushed directly to `liquibase/liquibase` instead of from the `v-petrovych/liquibase` fork.

Direct-push routes `pull_request_target` to `environment: internal` (rather than the gated `external` env that held all CI in `waiting` state on fork PRs). CI fires automatically on PR-open — verified on #7766, no maintainer click needed. This is the structural fix for the failure mode behind [DAT-23097](https://datical.atlassian.net/browse/DAT-23097) (PR #7741 was fork-authored, gate held CI in `waiting`, regression shipped to main without test signal).

Closing in favor of #7766.

PR #7758, opened 2026-05-22
Fix MariaDB enum null default snapshot

### LINKED ISSUE #3444: Enum NULL default value become "DEFAULT 'null' NULL"

WahidinAlambiyah:
## Impact

## Description

Fix MariaDB enum NULL default value handling when generating changelogs.

When generating a changelog from a MariaDB table with an enum column using `DEFAULT NULL`, Liquibase could previously serialize the default as `defaultValue="null"`. This could later produce invalid SQL such as `DEFAULT 'null' NULL`.

This change normalizes MariaDB/MySQL enum default values equal to `NULL` as a computed database value, so the generated changelog can use `defaultValueComputed="NULL"` instead of treating `null` as a literal string.

A regression test was added to verify that an enum column default value of `null` is converted to `DatabaseFunction("NULL")`.

Fixes #3444

## Release note

Fixed MariaDB enum columns with `DEFAULT NULL` being generated as a literal string default value in changelogs.

## Things to be aware of

This change is intentionally limited to MySQL/MariaDB enum columns whose default value is reported as `NULL`.

The fix converts the enum `NULL` default to a `DatabaseFunction("NULL")`, which allows changelog generation to serialize it as a computed default value instead of a quoted string default.

## Things to worry about

The main assumption is that a MariaDB/MySQL enum column default reported as `null` represents SQL `NULL`, not a literal enum value named `null`.

If maintainers prefer a stricter distinction between SQL `NULL` and a literal enum value `'null'`, this may need additional metadata handling from `INFORMATION_SCHEMA.COLUMNS`.

## Additional Context

This addresses the behavior described in #3444, where a MariaDB enum column like:

```sql
CREATE TABLE `test_with_enum` (
  `id` int(11) NOT NULL,
  `enumerateur` enum('yo','lo') DEFAULT NULL
) ENGINE=InnoDB DEFAULT CHARSET=latin1;
```

Testing performed:

```powershell
.\mvnw.cmd -pl liquibase-standard -Dtest=ColumnSnapshotGeneratorTest test
```

### DIFF
--- a/liquibase-standard/src/main/java/liquibase/snapshot/jvm/ColumnSnapshotGenerator.java
+++ b/liquibase-standard/src/main/java/liquibase/snapshot/jvm/ColumnSnapshotGenerator.java
@@ -610,6 +610,11 @@ public class ColumnSnapshotGenerator extends JdbcSnapshotGenerator {
     }
 
     private void readDefaultValueForMysqlDatabase(CachedRow columnMetadataResultSet, Column column, Database database) {
+        if (isMysqlEnumNullDefaultValue(columnMetadataResultSet, column)) {
+            columnMetadataResultSet.set(COLUMN_DEF_COL, new DatabaseFunction("NULL"));
+            return;
+        }
+        
         try {
             StringBuilder selectQuery = new StringBuilder("SELECT EXTRA FROM INFORMATION_SCHEMA.COLUMNS\n")
                     .append("WHERE TABLE_SCHEMA = ?\n")
@@ -627,6 +632,20 @@ public class ColumnSnapshotGenerator extends JdbcSnapshotGenerator {
         } catch (DatabaseException e) {
             Scope.getCurrentScope().getLog(getClass()).warning("Error fetching extra values", e);
         }
+    }
+
+    private boolean isMysqlEnumNullDefaultValue(CachedRow columnMetadataResultSet, Column column) {
+        Object defaultValue = columnMetadataResultSet.get(COLUMN_DEF_COL);
+
+        if (!(defaultValue instanceof String) || !StringUtil.equalsWordNull((String) defaultValue)) {
+            return false;
+        }
+
+        if (column.getType() == null || column.getType().getTypeName() == null) {
+            return false;
+        }
+
+        return StringUtils.startsWithIgnoreCase(column.getType().getTypeName(), "ENUM");
     }
 
     private void readDefaultValueForPostgresDatabase(CachedRow columnMetadataResultSet, Column columnInfo) {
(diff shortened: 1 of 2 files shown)


WahidinAlambiyah:
Hi maintainers, I noticed that #7725 also addresses #3444.

This PR takes a slightly different approach by converting the MariaDB/MySQL enum NULL default to `DatabaseFunction("NULL")`, so it can be serialized as `defaultValueComputed="NULL"` rather than omitting the default value entirely.

It also includes a regression test for the behavior.

I’m happy to adjust this PR, align it with #7725, or close it if maintainers prefer the other approach.

filipelautert (commented):
Hey, reviewed the change. The fix is small and well-targeted, and the test case is a good addition. Two things worth tightening:

1. **Literal `'null'` enum member is a false positive** — `equalsWordNull` matches the string `"null"`, but `ENUM('null','active') DEFAULT 'null'` is a column whose default is the *valid enum value* `'null'`, not SQL NULL. That case would be incorrectly rewritten to `DatabaseFunction("NULL")`. It's narrow, but a one-line comment noting the limitation (or a test asserting the current behavior) would save a future debugging session.

2. **Scope** — `isMysqlEnumNullDefaultValue` only fires for `ENUM` types. Worth confirming the enum-only scope is deliberate (i.e. non-enum MySQL/MariaDB columns already snapshot a `null` default correctly), so we're not fixing only half the surface.

Minor: trailing whitespace on the blank line after the early `return;`.

Issue #7215, opened 2025-08-20
Problem with foreignKeyConstraintExists from java on partitioned postgres table


wernerschram:
### Search first

- [x] I searched and no similar issues were found

### Description

We have a postgresql database with the pg_partman extension. There is a foreign key constrain between two partitioned tables. I'm trying to drop this constraint if it exists, so I'm using the `foreignKeyConstraintExists` precondition. 

If I run my liquibase script from java code, the `foreignKeyConstraintExists` condition fails even when the foreign key exists. When I run it with the liquibase CLI it works as expected. 

If I remove the partitioning from my test example, it works fine in both the java and CLI variant. So it seems like the `foreignKeyConstraintExists` precondition behaves differently in the CLI than when the java library is used directly. I haven't found out why; it seems to run the exact same code eventually. 

I tested with 4.30.0, 4.31.1 and 4.33.0 of the java library and with 4.31.0 of the cli (from the archlinux aur repository) and postgresql 16.4 and 17.5.

### Steps To Reproduce

In a test database setup pg_partman:
```
CREATE SCHEMA IF NOT EXISTS partman;
        CREATE EXTENSION pg_partman WITH SCHEMA partman;
        GRANT ALL ON SCHEMA partman TO "test";
        GRANT ALL ON ALL TABLES IN SCHEMA partman TO "test";
        GRANT EXECUTE ON ALL FUNCTIONS IN SCHEMA partman TO "test";
        GRANT EXECUTE ON ALL PROCEDURES IN SCHEMA partman TO "test";
        GRANT ALL ON SCHEMA public TO "test"
```

This changelog sets up the test database with two partitioned tables (setup.xml):
```xml
<?xml version="1.1" encoding="UTF-8" standalone="no"?>
<databaseChangeLog xmlns="http://www.liquibase.org/xml/ns/dbchangelog"
                   xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
                   xsi:schemaLocation="http://www.liquibase.org/xml/ns/dbchangelog http://www.liquibase.org/xml/ns/dbchangelog/dbchangelog-latest.xsd">

    <changeSet author="test" id="setup-1">

        <createTable tableName="t1">
            <column name="id" type="int4">
                <constraints primaryKey="true" />
            </column>
        </createTable>
        <modifySql dbms="postgresql">
            <append value=" PARTITION BY RANGE (id)"/>
        </modifySql>
    </changeSet>
[...]

PR #7759, opened 2026-05-24
fix(postgres): preserve PARTITION BY in diff-changelog (#6885)

### LINKED ISSUE #1874: Postgres diff doesn't include partitions?
**Liquibase Version**:
4.3.5

**Liquibase Integration & Version**: Maven, Spring Boot

**Liquibase Extension(s) & Version**:  4.3.5

**Database Vendor & Version**: Postgres 13.2

## Description
Using: mvn liquibase:diff generates a diff, and works fine for most tables, but does not include partitioned tables.

## Steps To Reproduce
* Create a partitioned table on one server
* Create a diff
* Notice the diff doesn't have partition information?

## Actual Behavior
Does not include partitioned tables.

## Expected/Desired Behavior
Include partitioned tables.

### LINKED ISSUE #6885: [POSTGRESQL] diff-changelog don't properly work with partitionned table
### Search first

### Description

Goal here is to transfer a partitionned table from a DB to another.
Using command line diff-changelog to generate changelog and update command to apply.

The diff-changelog see partitionned table as simple table, and create a file without any specific partition.

### Steps To Reproduce

Using postgresql jdbc driver from internal/lib

To create a partitionned test table : 
`
CREATE TABLE public.test_tbl_part
(
    test_date_int integer NOT NULL,
    PRIMARY KEY (test_date_int)
) PARTITION BY RANGE (test_date_int);

CREATE TABLE public.test_tbl_part_2025 PARTITION OF public.test_tbl_part
    FOR VALUES FROM (20250101) TO (20260101);

CREATE TABLE public.test_tbl_part_default PARTITION OF public.test_tbl_part
    DEFAULT;
`

When using diff-changelog :
`
-- changeset root:1744038467464-2
CREATE TABLE "test_tbl_part" ("test_date_int" INTEGER NOT NULL);

-- changeset root:1744038467464-3
CREATE TABLE "test_tbl_part_2025" ("test_date_int" INTEGER NOT NULL);

-- changeset root:1744038467464-4
CREATE TABLE "test_tbl_part_default" ("test_date_int" INTEGER NOT NULL);
`

### Expected/Desired Behavior

Expected same as creation :

`
CREATE TABLE public.test_tbl_part
(
    test_date_int integer NOT NULL,
    PRIMARY KEY (test_date_int)
) PARTITION BY RANGE (test_date_int);

CREATE TABLE public.test_tbl_part_2025 PARTITION OF public.test_tbl_part
    FOR VALUES FROM (20250101) TO (20260101);

CREATE TABLE public.test_tbl_part_default PARTITION OF public.test_tbl_part
    DEFAULT;
`

### Liquibase Version

4.31.1

### Database Vendor & Version

PostgreSQL 12.13

### Liquibase Integration

command line

### Liquibase Extensions

_No response_

Debian

### Additional Context

_No response_

Khromushkin:
## Impact

- [ ] Breaking change

## Description

PostgreSQL `diff-changelog` and `generate-changelog` silently dropped the
`PARTITION BY` clause on declaratively-partitioned tables. The generated
changelog described a plain heap table; applying it to a fresh database produced
a non-partitioned copy of the source. Structural divergence not visible in the
changelog text.

This PR adds first-class support for the parent-table `PARTITION BY <strategy> (<keys>)`
clause via a new optional `partitionBy` attribute on `<createTable>`. Captured
verbatim from `pg_get_partkeydef(c.oid)` at snapshot time, round-tripped through
the diff and SQL-generation paths, emitted in the changelog XML, and re-applied
on `liquibase update`.

Fixes #6885
Closes #1874
Refs #7215

### Structural template

PR #6901 (`CREATE INDEX ... USING <method>`) is the exact same shape — an
optional Postgres-specific clause captured from a system catalog and emitted in
DDL. @filipelautert explicitly named #6901 as the template in the #6885 thread,
with pointers to `TableSnapshotGenerator.java:83` and
`CreateTableGenerator.java:357`. This PR follows that template across 5 commits
(the remaining 3 are documented under "Drive-by fixes" below).

### Verification

Before (current `main`, postgres:16 in colima):
```xml
<createTable tableName="test_tbl_part">
  <column name="test_date_int" type="INTEGER">
  </column>
</createTable>
```

After:
```xml
<createTable partitionBy="RANGE (test_date_int)" tableName="test_tbl_part">
  <column name="test_date_int" type="INTEGER">
  </column>
</createTable>
```

`liquibase-standard`: **6080 unit tests, 0 failures**, 12 environment-dependent
skips. `PostgreSQLIntegrationTest` against postgres:15 via TestSystemFactory:
**54/54 passing**, 1 environment-dependent skip. Includes a new
`testSnapshotPartitionedTableCapturesPartitionBy` integration regression test
that fails on `main` before this PR and passes after.

## Release note

PostgreSQL `diff-changelog` and `generate-changelog` now preserve the
`PARTITION BY` clause on declaratively-partitioned tables. Previously the
partition strategy was silently dropped from the generated changelog,
producing a non-partitioned copy when applied to a target database. The new
behaviour is also expressible directly in handwritten changelogs via the
`partitionBy` attribute on `<createTable>` — for example
`<createTable tableName="orders" partitionBy="RANGE (created_at)">`.

### Checksum stability

The new `partitionBy` field is **version-gated** in
`CreateTableChange.getExcludedFieldFilters(ChecksumVersion)`:

- `V8` and earlier: excluded from the checksum. Pre-existing `<createTable>`
  changesets that have already been applied keep their stored checksum after
  upgrading — no `Validation Failed: Stored checksum ...` errors and no
  forced `clear-checksums` runs.
- `V9` (current latest, since Liquibase 4.22.0): included in the checksum.
  Changes to the partition specification on V9-era changesets are caught by
  validation rather than silently slipping through.

Mirrors the version-gating pattern in `CreateViewChange.java:165-180` (added
alongside the V8→V9 bump). This is a more thoughtful approach than the simple
`always-exclude` chosen for `using` in PR #6901; happy to align with that
precedent if reviewers prefer.

### Verbatim partition spec storage

The string stored in `partitionBy` is whatever `pg_get_partkeydef(c.oid)`
returns — `"RANGE (created_at)"`, `"LIST (region)"`, `"HASH (id)"`,
`"RANGE (lower(email))"` (functional), `"RANGE (\"MixedCaseCol\")"` (quoted),
multi-column `"RANGE (a, b)"`. No upper-casing, normalization, or parsing.
The output is already valid Postgres syntax for the right-hand side of
`PARTITION BY`. Tests in `CreateTableGeneratorPostgresTest` cover all five
shapes.

### Postgres-specific snapshot enrichment

`JdbcDatabaseSnapshot.queryPostgres` already requested
`TABLE_TYPE = 'PARTITIONED TABLE'` from JDBC metadata, so partitioned-table
rows were in the cache — but JDBC does not expose the partition strategy or
key columns. A new `enrichPostgresqlTablesResult` joins `pg_class`,
`pg_namespace`, `pg_partitioned_table` and projects `pg_get_partkeydef(c.oid)`
as a synthetic `PARTITION_BY` column on matching rows, keyed on
`(TABLE_SCHEM, TABLE_NAME)`. Mirrors PR #6901's index-USING enrichment in the
same file's `getIndexInfo` extractor.

The schema filter (and table filter, when scoped) are applied at SQL time
before the Java-side matching loop runs, so the matching is N×M with both
sides bounded by the partitioned-parent count in the schema — typically <10
even on large databases. No measurable cost.

### "Missing escape on `statement.getPartitionBy()` in the SQL append"

`partitionBy` is a clause-tail expression
(`"RANGE (a, b)"`, `"LIST (region)"`, `"HASH (lower(email))"`), not an
identifier. The existing identifier-escapers
(`escapeColumnName`, `escapeTableName`, `escapeTablespaceName`) don't apply;
there is no expression-level parser/sanitizer in the codebase. Precedent for
verbatim concatenation of clause-tail expressions:

- `CreateIndexGeneratorPostgres.generateSql:53-55` —
  `buffer.append(" USING ").append(statement.getUsing())` (PR #6901, merged
  April 2025, no escape requested at review)
- `CreateTableGenerator.generateSql:362,364` —
  `sql += " TABLESPACE " + statement.getTablespace()` (long-standing)

Trust boundary: `partitionBy` is set either by `pg_get_partkeydef()` at diff
time (a trusted PG-internal source) or by a changelog author writing
`<createTable partitionBy="...">` (the same trust level that allows arbitrary
`<sql>` blocks). No new attack surface introduced.

### "ChangedTableChangeGenerator no-ops for partition-spec changes"

Postgres declarative partitioning cannot be added, removed, or modified via
ALTER TABLE — the parent table has to be dropped and recreated. The
`ChangedTableChangeGenerator` mirrors the existing tablespace behaviour:
a `warning` is logged when a `partitionBy` difference is detected, no change
is emitted. This is consistent with how `tablespace` changes are handled in
the same generator.

### "Why not a separate `CreateTableGeneratorPostgres`?"

The base `CreateTableGenerator.generateSql` is already heavily branched per
database (`MSSQLDatabase`, `SybaseASADatabase`, `AbstractDb2Database`,
`InformixDatabase`, `MySQLDatabase`, `OracleDatabase`). A one-line
`if (database instanceof PostgresDatabase) sql += " PARTITION BY ..."`
joins those rather than spawning a new class — consistent with existing
precedent.

## Scope (deliberate exclusions)

**Child partitions (`CREATE TABLE child PARTITION OF parent FOR VALUES ...`)**
are out of scope for this PR. They remain as raw `<sql>` change sets — the
same workaround users already employ. A follow-up PR can add `pg_inherits` +
`pg_get_expr(relpartbound, oid)` enumeration plus a `partitionOf`/`partitionBound`
attribute pair (or a new `<createPartition>` change type). Doubling the
surface area into one PR would significantly increase the review burden and
risk "let's discuss the architecture" deferral — keeping V1 tight per the
maintainer guidance in #6885.

## Drive-by fixes (cleanly cherry-pickable)

Two commits unrelated to #6885 — surfaced because they were blocking the
local pre-push test run, and isolating them into separate PRs would have
delayed shipping the load-bearing fix. If maintainers prefer, both are
self-contained:

- `cc332c791 test(cli): refresh LiquibaseCommandLineTest 'help output' fixture`
  — three new `--allow-*` global options (CWE-22/-78/-470 deprecation flags)
  were added on `main` recently but the hardcoded `expectedHelpOutput` in
  `LiquibaseCommandLineTest.groovy` was not regenerated. The test fails on a
  clean `main` checkout. Verified by stashing this branch's changes before
  running the test. Mechanical regeneration from the live `liquibase --help`
  output.

- `a2dd2af06 test(postgres): fix testStatusRunDuringUpdate regression from CWE-316`
  — `PostgreSQLIntegrationTest.testStatusRunDuringUpdate` calls
  `commandScope.execute()` three times on the same `CommandScope` instance.
  PR #7741 (CWE-316, May 21) added a finally block in `CommandScope.execute()`
  that overwrites credential argument values with `"*****"` after the
  pipeline runs — making a single `CommandScope` instance single-use. Second
  call gets `FATAL: password authentication failed for user "lbuser"`. Fix
  is one line per call site: build a fresh `CommandScope` per snapshot via
  the existing `getStatusCommandScope(...)` helper. Test logic, assertions,
  and timing semantics are otherwise unchanged.

## Additional Context

- Local verification environment: macOS + Colima + postgres:16/postgres:15.
- Manual reproduction harness in colima documented in the commit body of
  `afe00c4a4`.
- Maintainer guidance from #6885: filipelautert pointed at the splice points
  in `TableSnapshotGenerator.java:83` and `CreateTableGenerator.java:357-356`
  and named PR #6901 as the structural template. Both pointers are honored
  verbatim.

The contributor stays accountable for all architecture and scope decisions on this PR. (this line typed without AI assistance, just to be clear 🙃) — @Khromushkin

### DIFF
--- a/liquibase-integration-tests/src/test/java/liquibase/dbtest/pgsql/PostgreSQLIntegrationTest.java
+++ b/liquibase-integration-tests/src/test/java/liquibase/dbtest/pgsql/PostgreSQLIntegrationTest.java
@@ -497,6 +497,9 @@ public class PostgreSQLIntegrationTest extends AbstractIntegrationTest {
     @Test
     public void testSnapshotPartitionedTableCapturesPartitionBy() throws Exception {
         assumeNotNull(this.getDatabase());
+        // Declarative partitioning syntax (PARTITION BY) requires PostgreSQL 10+.
+        // Matches the version-guard pattern used elsewhere in this file. Suggested by CodeRabbit on PR #7759.
+        assumeTrue(getDatabase().getDatabaseMajorVersion() >= 10);
         Scope.getCurrentScope().getSingleton(ExecutorService.class).getExecutor("jdbc", getDatabase())
                 .execute(new RawParameterizedSqlStatement(
                         "CREATE TABLE part_test (id BIGINT, sold_at DATE NOT NULL) PARTITION BY RANGE (sold_at)"));
--- a/liquibase-integration-tests/src/test/java/liquibase/dbtest/pgsql/PostgreSQLIntegrationTest.java
+++ b/liquibase-integration-tests/src/test/java/liquibase/dbtest/pgsql/PostgreSQLIntegrationTest.java
@@ -488,6 +488,56 @@ public class PostgreSQLIntegrationTest extends AbstractIntegrationTest {
         Assert.assertTrue("Using index function should be \"" + using + "\"", found);
     }
 
+    /**
+     * Regression test for #6885: diff-changelog and generate-changelog now capture the
+     * PARTITION BY clause for declaratively-partitioned PostgreSQL tables. Pre-fix, this
+     * test would have asserted false (the CreateTableChange would have no partitionBy
+     * attribute and the parent table would have been emitted as a plain heap table).
+     */
+    @Test
+    public void testSnapshotPartitionedTableCapturesPartitionBy() throws Exception {
+        assumeNotNull(this.getDatabase());
+        Scope.getCurrentScope().getSingleton(ExecutorService.class).getExecutor("jdbc", getDatabase())
+                .execute(new RawParameterizedSqlStatement(
+                        "CREATE TABLE part_test (id BIGINT, sold_at DATE NOT NULL) PARTITION BY RANGE (sold_at)"));
+        try {
+            DiffResult diffResult = DiffGeneratorFactory.getInstance().compare(getDatabase(), null, new CompareControl());
+            DiffToChangeLog changeLogWriter = new DiffToChangeLog(diffResult,
+                    new DiffOutputControl(false, false, false, null));
+            List<ChangeSet> changeSets = changeLogWriter.generateChangeSets();
+
+            String capturedPartitionBy = null;
+            for (ChangeSet changeSet : changeSets) {
+                for (Change change : changeSet.getChanges()) {
+                    if (!(change instanceof CreateTableChange)) {
+                        continue;
+                    }
+                    CreateTableChange ctc = (CreateTableChange) change;
+                    if ("part_test".equalsIgnoreCase(ctc.getTableName())) {
+                        capturedPartitionBy = ctc.getPartitionBy();
+                        break;
+                    }
+                }
+                if (capturedPartitionBy != null) {
+                    break;
+                }
+            }
+
+            Assert.assertNotNull("CreateTableChange for the partitioned table 'part_test' should carry a partitionBy attribute",
+                    capturedPartitionBy);
+            // pg_get_partkeydef typically returns lowercase "range (sold_at)" but exact case is version-dependent.
+            // We assert on the meaningful tokens (RANGE strategy + sold_at column) case-insensitively rather than on the literal string.
+            String normalized = capturedPartitionBy.toUpperCase();
+            Assert.assertTrue("partitionBy should mention the RANGE strategy; got: " + capturedPartitionBy,
+                    normalized.contains("RANGE"));
+            Assert.assertTrue("partitionBy should reference the sold_at key column; got: " + capturedPartitionBy,
+                    normalized.contains("SOLD_AT"));
+        } finally {
+            Scope.getCurrentScope().getSingleton(ExecutorService.class).getExecutor("jdbc", getDatabase())
+                    .execute(new RawParameterizedSqlStatement("DROP TABLE IF EXISTS part_test"));
+        }
+    }
+
     @Test
     public void rollbackRestoresSearchPathForNumericSchema() throws Exception {
         final String numericSchema = "4b5d63f449304c05a291c4c27136ebb5";
--- a/liquibase-standard/src/main/java/liquibase/change/core/CreateTableChange.java
+++ b/liquibase-standard/src/main/java/liquibase/change/core/CreateTableChange.java
@@ -62,6 +62,8 @@ public class CreateTableChange extends AbstractChange implements ChangeWithColum
     private Boolean ifNotExists;
     @Setter
     private Boolean rowDependencies;
+    @Setter
+    private String partitionBy;
 
     public CreateTableChange() {
         super();
@@ -191,7 +193,8 @@ public class CreateTableChange extends AbstractChange implements ChangeWithColum
     }
 
     protected CreateTableStatement generateCreateTableStatement() {
-        return new CreateTableStatement(getCatalogName(), getSchemaName(), getTableName(), getRemarks(), getTableType(), Boolean.TRUE.equals(getIfNotExists()), Boolean.TRUE.equals(getRowDependencies()));
+        return new CreateTableStatement(getCatalogName(), getSchemaName(), getTableName(), getRemarks(), getTableType(), Boolean.TRUE.equals(getIfNotExists()), Boolean.TRUE.equals(getRowDependencies()))
+                .setPartitionBy(getPartitionBy());
     }
 
     @Override
@@ -315,10 +318,25 @@ public class CreateTableChange extends AbstractChange implements ChangeWithColum
         return rowDependencies;
     }
 
+    @DatabaseChangeProperty(
+            description = "PostgreSQL partition definition, e.g. 'RANGE (created_at)' or 'LIST (region)'. " +
+                    "Causes CREATE TABLE to be emitted as a partitioned parent table.",
+            supportsDatabase = "postgresql"
+    )
+    public String getPartitionBy() {
+        return partitionBy;
+    }
+
     @Override
     public String[] getExcludedFieldFilters(ChecksumVersion version) {
-        return new String[] {
-                "ifNotExists", "rowDependencies"
-        };
+        // For checksum versions before V9, partitionBy must be excluded so that pre-existing
+        // <createTable> changesets (which never had this field) keep the same checksum after
+        // upgrading to a Liquibase version that knows about partitionBy. From V9 onward,
+        // partitionBy contributes to the checksum so that changes to the partition spec are
+        // caught by validation. Mirrors the version-gating pattern in CreateViewChange.
+        if (version.lowerOrEqualThan(ChecksumVersion.V8)) {
+            return new String[] { "ifNotExists", "rowDependencies", "partitionBy" };
+        }
+        return new String[] { "ifNotExists", "rowDependencies" };
     }
 }
--- a/liquibase-standard/src/main/java/liquibase/change/core/CreateTableChange.java
+++ b/liquibase-standard/src/main/java/liquibase/change/core/CreateTableChange.java
@@ -37,6 +37,7 @@ import liquibase.structure.core.Table;
 import liquibase.util.ObjectUtil;
 import liquibase.util.StringUtil;
 import lombok.Setter;
+import org.apache.commons.lang3.StringUtils;
 
 /**
  * Creates a new table.
@@ -193,8 +194,11 @@ public class CreateTableChange extends AbstractChange implements ChangeWithColum
     }
 
     protected CreateTableStatement generateCreateTableStatement() {
+        // Trim-to-null on partitionBy mirrors how tablespace is normalized below in generateStatements;
+        // prevents a whitespace-only attribute from being forwarded into the SQL generator and emitting
+        // an invalid trailing "PARTITION BY ". Reported by CodeRabbit on PR #7759.
         return new CreateTableStatement(getCatalogName(), getSchemaName(), getTableName(), getRemarks(), getTableType(), Boolean.TRUE.equals(getIfNotExists()), Boolean.TRUE.equals(getRowDependencies()))
-                .setPartitionBy(getPartitionBy());
+                .setPartitionBy(StringUtils.trimToNull(getPartitionBy()));
     }
 
     @Override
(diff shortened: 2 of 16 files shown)


filipelautert (changes requested):
Hey Khromushkin, reviewed the changes. Architecture, tests, and pre-emptive justifications all look right — thanks for the thorough PR body, made review fast. One real blocker to address:

1. **`CreateTableGenerator.java:358-374`** — `PARTITION BY` is emitted *after* `TABLESPACE`, but Postgres grammar requires the opposite order: `(columns) [PARTITION BY ...] [USING ...] [WITH ...] [ON COMMIT ...] [TABLESPACE ...]`. Any user with both `partitionBy` AND `tablespace` set will get a syntax error from Postgres. Swap the two blocks — move the `partitionBy` append (lines 369-374) above the tablespace block (lines 358-366).

2. **Test the combination** — `CreateTableGeneratorPostgresTest` covers 5 partition shapes but no test sets `partitionBy` + `tablespace` together. Add one asserting the emitted SQL is `... PARTITION BY RANGE (col) TABLESPACE my_ts` (in that order).

Re your drive-by CWE-316 fix: diagnosis is accurate (#7741 did merge May 21 with the credential-wipe finally block) and your fix is the right shape — confirmed no other integration-test sites have the multi-execute-on-same-scope pattern.

Khromushkin:
Thanks @filipelautert — appreciated the quick turnaround. Both items addressed in f1fbdc433:

1. Swapped the PARTITION BY block above TABLESPACE in `CreateTableGenerator.generateSql`. The two blocks were already self-contained so the swap is purely textual — no other clauses (MySQL autoincrement options, MySQL COMMENT, Oracle ROWDEPENDENCIES) are affected.

2. Added `testGenerateSqlEmitsPartitionByBeforeTablespace` to `CreateTableGeneratorPostgresTest`: exact-string assertion for `... PARTITION BY RANGE (created_at) TABLESPACE my_ts` plus a defensive index-order check as a regression guard against whitespace drift.

Full `CreateTableGenerator*Test` (55 + 9) still passes locally — the swap is invisible to statements without `partitionBy`. Mind taking another look when you have time?

filipelautert (approved):
Hey Khromushkin, both fixes look good — re-approving.

Heads-up unrelated to the partition work: [#7760](https://github.com/liquibase/liquibase/pull/7760) (@v-petrovych's standalone fixture refresh, same three `ALLOW_*` flags) merged earlier today. Your drive-by `cc332c791` is now duplicative with main — when you rebase to pick up the merge, just drop that commit. Your other drive-by `a2dd2af06` (the `CommandScope` test fix) is unaffected and still needed.

Khromushkin:
Thanks @filipelautert for handling the rebase — appreciated, didn't expect it.

Two follow-ups:

1. Two checks still failing — `Run Test for (Java 25 windows-latest)` (fails at `Set up JDK 25`; same matrix passes on Ubuntu and macOS) and `claude-review / claude-review` (errors at the AWS Secrets Manager step, `Could not load credentials from any providers` — looks like the fork-PR secrets pattern). Both look infra-related from this side. Anything actionable from my end, or just gated on a maintainer re-run?

2. Was planning a V2 follow-up PR for child partitions (`pg_inherits` enumeration + `partitionOf`/`partitionBound` attributes on `<createTable>`, or a new `<createPartition>` change type). Happy to scope and open it once this lands. Any preference on shape — extend the existing change or new change type?

Anything else I can pick up?

Khromushkin:
Rebased onto current main and dropped commit `d5326375f` (the `testStatusRunDuringUpdate` workaround) — now redundant after @v-petrovych's #7764 merged the API-level fix, per the heads-up in that PR's description. Branch is now 8 commits clean on top of `c8c5f806c`.

`mergeStateStatus` flipped to `BLOCKED` since the force-push dismissed your prior approval — @filipelautert / @wwillard7800, would appreciate a re-look when you have a moment. The `authorize` check is pending again on the GHA fork-PR gate; same pattern as last time.

Tests still green locally (6131/0 in `liquibase-standard`).

filipelautert (approved):
Hey, re-reviewed after the June 5 push — you've cleanly addressed everything from the last round:

- `PARTITION BY` now emitted **before** `TABLESPACE` per the PostgreSQL grammar (the ordering bug I flagged) in `CreateTableGenerator.generateSql`.
- Added a `validate()` warning when `partitionBy` is set on a non-Postgres target, so the clause being dropped is surfaced instead of silently lost.
- Picked up the `trimToNull` whitespace defense-in-depth and the Cockroach guard.

Captured verbatim from `pg_get_partkeydef` and round-tripped through snapshot → diff → sqlgen looks right. Approving.

rberezen:
Tested locally against PostgreSQL 17 with all five declarative-partitioning shapes. The bug fully reproduces on `master` and the fix is clean across the board.

**Bug reproduction on master:**
```
source DB:                                      target after generate→apply (master):
  sales_range  | partitioned table   →            sales_range  | table (heap)
  users_list   | partitioned table   →            users_list   | table (heap)
  events_hash  | partitioned table   →            events_hash  | table (heap)
  multi_key    | partitioned table   →            multi_key    | table (heap)
  funky_func   | partitioned table   →            funky_func   | table (heap)
```
Every partition strategy silently dropped — exactly as #6885 describes.

**Fix verified — bit-for-bit round-trip on PR:**
```sql
-- source vs target after generate→apply via PR build
SELECT relname, pg_get_partkeydef(oid) ...

   relname   |  pg_get_partkeydef
-------------+----------------------
 events_hash | HASH (id)
 funky_func  | RANGE (lower(email))   ← functional partition preserved
 multi_key   | RANGE (a, b)           ← multi-column preserved
 sales_range | RANGE (sold_at)
 users_list  | LIST (region)
```
Source and target outputs match identically. Re-generating from the target produces a changelog byte-equivalent to the source's (no spurious diff — verified with `diff`).

**Edge cases verified:**
- Quoted/mixed-case identifier `"MixedCaseCol"` → emitted as `partitionBy="RANGE (&quot;MixedCaseCol&quot;)"` and round-trips correctly ✅
- Empty/whitespace `partitionBy=""` / `partitionBy="   "` → trimmed-to-null, table created as plain heap (no malformed `PARTITION BY` clause emitted) ✅
- Non-partitioned `plain_heap` table → no `partitionBy` attribute emitted ✅
- `partitionBy` attribute on MySQL → silently ignored, table created normally (gated correctly on `PostgresDatabase`) ✅

**Code review notes:**
- Cockroach guard in `enrichPostgresqlTablesResult` is well-justified and the in-code comment about the parallel guard gap in `getIndexInfo` (PR #6901) is a good catch — agreed that's out of scope here.
- `PARTITION BY` emitted before `TABLESPACE` matches PG grammar; nice catch by @filipelautert during review.
- Checksum version-gating (V8 excludes, V9 includes) mirrors the `CreateViewChange` pattern correctly — protects pre-existing changesets from forced `clear-checksums` while still catching real drift on V9-era changesets.
- Trust boundary for verbatim `partitionBy` emission (no escape) is sound: input is either `pg_get_partkeydef()` output or an author-controlled changelog attribute (same trust level as `<sql>` blocks). Consistent precedent in `using` (#6901) and `tablespace`.

**Verification summary:**
- 79/79 new unit tests pass locally (`CreateTableChangeTest`, `TableSnapshotGeneratorTest`, `MissingTableChangeGeneratorTest`, `CreateTableGeneratorPostgresTest`).
- All 10 integration DBs + 7 Java/OS combos green in CI.
- The two "drive-by" commits (`LiquibaseCommandLineTest` fixture refresh, `testStatusRunDuringUpdate` post-CWE-316 fix) are cleanly scoped and well-explained — fine to keep bundled.

Excellent PR — clear scoping, deliberate exclusion of child partitions for V1, defensive coding, and thorough self-review on common review prompts. LGTM, ready to merge once the `authorize` gate completes. Thanks @Khromushkin!

rberezen (approved):

PR #7763, opened 2026-05-26
CWE-316: make CommandScope.clearCredentialArguments() opt-in (fixes #7761)

### LINKED ISSUE #7761: CWE-316 regression: CommandScope.execute() finally-block clearCredentialArguments() breaks any caller re-using a CommandScope
## TL;DR

PR #7741 (CWE-316: clear password memory after use in legacy CLI + CommandScope, commit `4850342ae7`) added a `finally { clearCredentialArguments(); }` block at the end of `CommandScope.execute()`. That block overwrites every credential-bearing argument value (`password`, `passwd`, `secret`, `token`, `apikey`, `accesskey`, `licensekey`) with `"*****"` in the scope's `argumentValues` map.

This breaks **any caller that calls `execute()` more than once on the same `CommandScope`** — the second execute() reads `"*****"` instead of the real credential and fails to authenticate.

The canary test in this very PR — `liquibase.dbtest.pgsql.PostgreSQLIntegrationTest.testStatusRunDuringUpdate` — does exactly that (three `commandScope.execute()` calls in a row on the same scope) but apparently isn't run against a live postgres in the project's PR CI, so the regression slipped through.

## Repro / canary

`liquibase-integration-tests/src/test/java/liquibase/dbtest/pgsql/PostgreSQLIntegrationTest.java::testStatusRunDuringUpdate`:

```java
CommandScope commandScope = getStatusCommandScope(this.sleepChangelog);
CommandResults statusCheckBefore = commandScope.execute();   // PASSES; clearCredentialArguments() then wipes scope's password to "*****"

CompletableFuture<Boolean> updateFuture =
        CompletableFuture.supplyAsync(() -> { ...liquibase.update(...); });
Thread.sleep(3000);
CommandResults statusCheckDuring = commandScope.execute();   // FAILS — password is now "*****"
```

The second `execute()` triggers connection setup with the masked password and the JDBC driver rejects it:

```
FATAL: password authentication failed for user "lbuser"
at liquibase.dbtest.pgsql.PostgreSQLIntegrationTest.testStatusRunDuringUpdate(PostgreSQLIntegrationTest.java:96)
```

Pro's daily subtree-sync builds reproduce this consistently (4 out of 4 runs since 2026-05-22 fail this exact test, all others pass).

## Diagnosis

The fix in PR #7741 conflates two things:

1. *"Drop references so the credential String becomes GC-eligible after the scope is no longer in use."* — Useful for embedded long-running scenarios.
2. *"Drop references after every `execute()`."* — Not safe; nothing in the contract of `execute()` says the scope is single-use.

Code paths that legitimately reuse a scope:
- The test above (three sequential `execute()` calls on the same status-check scope).
- Any embedded integration that builds a scope once and triggers it on a schedule.
- Any flow that retries `execute()` after a transient error.

## Suggested fixes (pick one)

**Option A — make the clearing opt-in.** Add `clearCredentialArguments()` as a public method on `CommandScope`; remove the `finally` block. Callers (e.g. `Main.java`) call it explicitly at the end of their CLI session. Tests / long-lived integrations that reuse the scope don't suffer. Preserves the CWE-316 intent (CLI sessions still clear at the right boundary).

**Option B — gate the clearing on a flag.** Add `CommandScope.markSingleUse()`; only clear if that's been called. Backwards-compatible default.

**Option C — clear on scope close rather than per-execute.** If `CommandScope` had a close/dispose lifecycle, do the clearing there.

I'm happy to author a PR for Option A if that's the preferred direction — please let me know.

## Context

Discovered while sprint-defending the Pro-side subtree-sync of this commit batch into liquibase-pro. The five-PR security batch (#7741/#7743/#7744/#7747/#7748/#7750/#7756) is otherwise green — this is the only blocker.

v-petrovych:
## TL;DR

PR #7741 ("CWE-316: clear password memory after use in legacy CLI + CommandScope") added a `finally { clearCredentialArguments(); }` block to `CommandScope.execute()`. **That breaks every caller that calls `execute()` more than once on the same `CommandScope`** — the second call reads `"*****"` and the JDBC driver rejects the connection.

This PR implements **Option A** from issue #7761: remove the auto-clearing in `execute()`, make `clearCredentialArguments()` public so callers can opt in explicitly, and wire the CLI's single-use call sites in `Main.java` to call it via a small `executeAndClearCredentials()` helper. Adds three new `CommandScopeTest` specs — including the **unit-level regression coverage for `PostgreSQLIntegrationTest.testStatusRunDuringUpdate`** — so future re-introductions are caught at PR-author time without needing a live postgres in the loop.

Closes #7761.

## The concrete regression

```java
    ...
    } finally {
        clearCredentialArguments();   // <-- auto-clear after every execute()
    }
}
```

`clearCredentialArguments()` overwrites every credential-bearing argument value (`password`, `passwd`, `secret`, `token`, `apikey`, `accesskey`, `licensekey`) in the scope's `argumentValues` map with `"*****"`. A caller that reuses the same scope:

```java
CommandScope statusScope = new CommandScope("status")
    .addArgumentValue("password", realPassword);

statusScope.execute();   // 1st call: OK, uses realPassword. After: argumentValues["password"] = "*****"
// ... concurrent liquibase.update() runs ...
statusScope.execute();   // 2nd call: argumentValues["password"] is now "*****"
                          //   → JDBC driver sees "*****" → "FATAL: password authentication failed"
```

`liquibase.dbtest.pgsql.PostgreSQLIntegrationTest.testStatusRunDuringUpdate` (line 93) does exactly that — three executions on the same scope to check status before / during / after a concurrent update. First call succeeds; second/third fail with:

```
liquibase.exception.DatabaseException: Connection could not be created to
jdbc:postgresql://localhost:48822/lbcat?loggerLevel=OFF with driver
org.postgresql.Driver. FATAL: password authentication failed for user "lbuser"
```

Upstream main has been red on this test since 2026-05-21 12:15 UTC (when #7741 merged). [The failing run linked from issue #7761](https://github.com/liquibase/liquibase/actions/runs/26402847894/job/77720530913).

## Why the regression sat in upstream main for four days

PR #7741 was authored from a fork (`v-petrovych/liquibase`). Upstream's `run-tests.yml` workflow gates its `authorize` job on `environment: external` whenever the PR's head repo differs from `github.repository`. That gate requires a maintainer to click **"Review deployments → Approve and run"** before the workflow can proceed.

The gate was never approved on #7741's pre-merge build. The workflow sat in `status: waiting`. No `Run Test for (Java N <OS>)` or `Integration Test (<db>)` jobs ever materialised. The PR's 9 green checks were all from non-gated workflows (CodeQL, label, claude-review). **The merge button did not block on a workflow that never produced any check at all.**

After merge, the post-merge `push` to `main` triggered the workflow with the `internal` environment (no manual approval needed) — but by then the merge was already in. Same shape for PR #7760 (the fixture fix that finally let the reactor advance past `liquibase-standard` to `liquibase-integration-tests`).

I've filed validation guidance for the next contributor in the upstream issue so this gate doesn't continue to hide regressions on fork-authored PRs. The five checks are summarised at the bottom of this PR description.

## Option A — opt-in clearing

Three options were discussed in issue #7761:

| Option | Description | Verdict |
|---|---|---|
| **A** | Make clearing opt-in: public method on `CommandScope`; remove finally; call explicitly at single-use sites. | **Chosen.** Preserves CWE-316 intent at CLI call sites; doesn't break re-use. |
| B | Gate clearing on a `markSingleUse()` flag (backwards-compat default). | Possible, but adds API surface and a state machine that's easy to forget. |
| C | Clear on scope `close()` rather than per-execute (`AutoCloseable`). | Bigger refactor; same forget-to-close risk as B. |

### 1. `CommandScope.java`

- **Removed** the outer `finally { clearCredentialArguments(); }` block from `execute()`.
- **Changed** `clearCredentialArguments()` from package-private to **`public`** so external callers can invoke it explicitly.
- **Updated** both the `CREDENTIAL_KEY_TOKENS` doc comment and the `clearCredentialArguments()` method doc to document the new caller-invoked contract, including the specific re-use scenario the auto-finally broke and the recommended try-finally pattern for single-use callers.

### 2. `Main.java`

- **Added** a private static helper:
  ```java
  private static CommandResults executeAndClearCredentials(CommandScope commandScope) throws CommandExecutionException {
      try {
          return commandScope.execute();
      } finally {
          commandScope.clearCredentialArguments();
      }
  }
  ```
- **Replaced** every CLI `commandScope.execute()` callsite with `executeAndClearCredentials(commandScope)`. 15 sites: `snapshotCommand`, `dropAllCommand`, `calculateChecksumCommand`, `historyCommand`, `diffCommand`, `diffChangelogCommand`, `updateCommand`, `rollbackOneChangeSet` (×2), `rollbackOneUpdate` (×2), `executeSql`, plus three bare `commandScope.execute()` calls.

Net effect: CLI users get the same CWE-316 protection they had before #7741's regression-causing fix landed, but the protection now lives at the call site (where the single-use assumption is true) rather than inside `execute()` (where the assumption was false for the integration test and any library/programmatic caller that reuses a scope).

### 3. `CommandScopeTest.groovy`

- **Updated** the two existing specs (`"execute invokes credential clearing in its outer finally..."`) to assert the new opt-in semantics: `execute()` alone does NOT touch credential values; explicit `clearCredentialArguments()` does.
- **Added** the key regression spec:
  > `"execute called twice on the same scope reads original credentials both times — regression for PostgreSQLIntegrationTest.testStatusRunDuringUpdate"`

  Calls `execute()` three times on the same scope and uses a `MockCommandStep` that captures `argumentValues["password"]` each time the pipeline runs. All three captured values must equal the original credential. **This is the unit-level regression coverage for the integration-test scenario that surfaced the bug** — running it at unit speed means future re-introductions are caught at PR-author time without needing a live postgres in the loop.
- **Added** a documentation spec for the recommended `try-finally` usage pattern.

## Tests

All passing locally on this branch (pre-merge validation):

```
CommandScopeTest                      22 / 22   (3 new specs)
LiquibaseCommandLineTest              40 / 40   (incl. 'help output' snapshot)
liquibase-cli full reactor           115 / 115
liquibase-standard Command*-related  204 / 204
MainTest                              32 / 32
```

**413 specs across the directly-impacted set, 0 failures.** Did not run the live-postgres integration tests locally — Docker not available in my setup. Pre-merge CI will need to run them; see the **Validation** section below.

## Validation (please read before merging)

Verbatim from the [Validation section](https://datical.atlassian.net/browse/DAT-23097) of the internal regression ticket — mirrored upstream as context for anyone in the org or external who authors / reviews fork PRs against `liquibase/liquibase`. The deep root cause of why both #7741 and #7760 merged without their tests running is the `environment: external` gate on the fork PR's `run-tests.yml`:

1. **Before celebrating CI-green, check the actual workflow run.** After opening the PR, go to the Actions tab → filter by branch → confirm `Unit and Integration Tests` has `conclusion: success`. If you see `status: waiting`, the tests did NOT run.
2. **If it's waiting, ask a maintainer to approve it.** Comment on the PR: *"Could a maintainer hit 'Review deployments → Approve and run' on the run-tests.yml workflow run for this PR? It's waiting on the `external` environment gate."* Don't ask for merge until that workflow has actually run and is green.
3. **If you have push access to `liquibase/liquibase` directly, prefer branching inside the org repo** rather than working from your fork. That avoids the gate entirely (the workflow picks `internal` environment and runs immediately).
4. **Run the snapshot tests locally before opening a fork PR**, especially anything touching `GlobalConfiguration`, `LiquibaseCommandLine`, or `CommandScope`. Five seconds locally beats four days of upstream-main red.
5. **When reviewing or merging a fork PR**, scan the rollup for `authorize` jobs in pending/null state — that's the explicit signal the test matrix was gated and never ran.

For this PR specifically: I'll watch the Actions tab after pushing. If the `Unit and Integration Tests` workflow lands in `status: waiting`, I'll comment asking for deployment approval before requesting merge.

## Related

- Issue liquibase/liquibase#7761 — the upstream issue this PR closes
- PR liquibase/liquibase#7741 — the original CWE-316 fix that introduced the regression (commit `4850342ae7`)
- PR liquibase/liquibase#7760 — the DAT-23096 fixture fix that finally let the reactor advance past `liquibase-standard` so this regression became visible on main
- Pro subtree-sync PR liquibase/liquibase-pro#3860 — held until this PR lands and a fresh subtree-sync picks it up

### DIFF
--- a/liquibase-standard/src/main/java/liquibase/command/CommandScope.java
+++ b/liquibase-standard/src/main/java/liquibase/command/CommandScope.java
@@ -52,9 +52,23 @@ public class CommandScope {
 
     /**
      * Substring tokens (lowercase) that mark an argument key as credential-bearing.
-     * Values for matching keys are overwritten with {@code "*****"} at the end of
-     * {@link #execute()} so the original credential Strings become GC-eligible and
-     * do not linger in heap for the rest of the JVM lifetime (CWE-316).
+     * Values for matching keys are overwritten with {@code "*****"} when the
+     * (public) {@link #clearCredentialArguments()} method is invoked by a caller.
+     * That makes the original credential Strings GC-eligible so they do not linger
+     * in heap for the rest of the JVM lifetime in long-running embedded scenarios
+     * (CWE-316).
+     * <p>
+     * <b>Caller-invoked, not auto-invoked from {@link #execute()}</b>. Calling it
+     * from an outer {@code finally} block in {@code execute()} (as the original
+     * CWE-316 fix did) breaks any caller that calls {@code execute()} more than
+     * once on the same {@code CommandScope} — the second call reads {@code "*****"}
+     * instead of the real credential and the JDBC driver rejects the connection.
+     * The CLI in {@code Main} wraps its single-use scopes in try-finally to call
+     * this method after each {@code execute()}; programmatic / library callers
+     * who reuse a {@code CommandScope} (e.g. status-before / update / status-after
+     * patterns) should call it once after the final {@code execute()}, never
+     * between executions. Discovered via
+     * {@code PostgreSQLIntegrationTest.testStatusRunDuringUpdate}.
      * <p>
      * Matching is intentionally substring-based on the lowercased key: keys like
      * {@code "argument__bearerToken"} are caught by {@code "token"};
@@ -379,22 +393,30 @@ public class CommandScope {
             throw e;
         } catch (Exception e) {
             throw new CommandExecutionException(e);
-        } finally {
-            clearCredentialArguments();
         }
     }
 
     /**
-     * Overwrite the values of any credential-bearing argument keys with
-     * {@code "*****"} so the original credential Strings stop being referenced by
-     * this scope. Strings are immutable and cannot be wiped in-place, but dropping
-     * the reference here makes them GC-eligible (assuming no other references hold
-     * the same String) instead of letting them live for the rest of the JVM in
-     * long-running embedded scenarios (CWE-316: Cleartext Storage of Sensitive
-     * Information in Memory). Called from {@link #execute()}'s outer finally so
-     * it runs regardless of command outcome; package-private for direct testing.
+     * Overwrite the values of any credential-bearing argument keys (per
+     * {@link #CREDENTIAL_KEY_TOKENS}) with {@code "*****"} so the original
+     * credential Strings stop being referenced by this scope. Strings are
+     * immutable and cannot be wiped in-place, but dropping the reference here
+     * makes them GC-eligible (assuming no other references hold the same String)
+     * instead of letting them live for the rest of the JVM in long-running
+     * embedded scenarios (CWE-316: Cleartext Storage of Sensitive Information
+     * in Memory).
+     * <p>
+     * <b>Caller-invoked.</b> This method must NOT be auto-invoked from
+     * {@link #execute()} — doing so (as the original CWE-316 fix did) breaks any
+     * caller that calls {@code execute()} more than once on the same scope, since
+     * the second call reads {@code "*****"} instead of the real credential.
+     * Callers that reuse a scope (e.g. status-before / update / status-after
+     * patterns) must call this method <b>after the final</b> {@code execute()},
+     * never between executions. Callers that use a scope only once (such as the
+     * CLI in {@code Main}) typically wrap the {@code execute()} call in a
+     * try-finally and call this in the {@code finally}.
      */
-    void clearCredentialArguments() {
+    public void clearCredentialArguments() {
         if (argumentValues.isEmpty()) {
             return;
         }
--- a/liquibase-standard/src/main/java/liquibase/integration/commandline/Main.java
+++ b/liquibase-standard/src/main/java/liquibase/integration/commandline/Main.java
@@ -1478,7 +1478,7 @@ public class Main {
                 // otherwise, we will print the output
                 //
                 Writer outputWriter = getOutputWriter();
-                CommandResults commandResults = snapshotCommand.execute();
+                CommandResults commandResults = executeAndClearCredentials(snapshotCommand);
                 String result = InternalSnapshotCommandStep.printSnapshot(snapshotCommand, commandResults);
                 outputWriter.write(result);
                 outputWriter.flush();
@@ -1521,7 +1521,7 @@ public class Main {
                         .addArgumentValue(DropAllCommandStep.CATALOG_AND_SCHEMAS_ARG, InternalSnapshotCommandStep.parseSchemas(database, getSchemaParams(database)))
                         .addArgumentValue(GenerateChangelogCommandStep.CHANGELOG_FILE_ARG, changeLogFile);
 
-                dropAllCommand.execute();
+                executeAndClearCredentials(dropAllCommand);
                 return;
             } else if (COMMANDS.STATUS.equalsIgnoreCase(command)) {
                 boolean runVerbose = false;
@@ -1558,7 +1558,7 @@ public class Main {
                         .addArgumentValue(CalculateChecksumCommandStep.CHANGESET_IDENTIFIER_ARG, getCommandParam(OPTIONS.CHANGE_SET_IDENTIFIER, null))
                         .addArgumentValue(CalculateChecksumCommandStep.CHANGELOG_FILE_ARG, this.changeLogFile);
 
-                calculateChecksumCommand.execute();
+                executeAndClearCredentials(calculateChecksumCommand);
                 return;
             } else if (COMMANDS.DB_DOC.equalsIgnoreCase(command)) {
                 if (commandParams.isEmpty()) {
@@ -1704,7 +1704,7 @@ public class Main {
                     historyCommand.addArgumentValue(HistoryCommandStep.FORMAT_ARG, HistoryFormat.valueOf(format));
                     historyCommand.setOutput(getOutputStream());
 
-                    historyCommand.execute();
+                    executeAndClearCredentials(historyCommand);
                 } else {
                     throw new CommandLineParsingException(
                             String.format(coreBundle.getString("command.unknown"), command));
@@ -1782,13 +1782,13 @@ public class Main {
         if (outputWriter != null) {
             commandScope.setOutput(new WriterOutputStream(outputWriter, GlobalConfiguration.OUTPUT_FILE_ENCODING.getCurrentValue()));
         }
-        commandScope.execute();
+        executeAndClearCredentials(commandScope);
     }
 
     private void runReleaseLocksCommand() throws CommandExecutionException {
         CommandScope commandScope = new CommandScope(ReleaseLocksCommandStep.COMMAND_NAME[0]);
         this.setDatabaseArgumentsToCommand(commandScope);
-        commandScope.execute();
+        executeAndClearCredentials(commandScope);
     }
 
     private void runGenerateChangelogCommandStep() throws LiquibaseException, IOException, CommandLineParsingException {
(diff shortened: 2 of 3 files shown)


v-petrovych:
**Validation step 1 confirmation (per the PR description's "Validation" section):** the `Unit and Integration Tests` workflow run for this PR is currently **gated** on the `external` environment. The `authorize` job is in `status: waiting`; no `Run Test for (Java N <OS>)` or `Integration Test (<db>)` jobs have been queued. The PR rollup currently shows only the 3 non-gated checks (CodeRabbit, CLA, mergefreeze).

Could a maintainer please hit **"Review deployments → Approve and run"** on the `run-tests.yml` workflow run for this PR? Without that approval, the same gate that previously hid #7741's regression and #7760's fixture drift will hide this fix's CI from the rollup, and we'd be back to merging on partial evidence.

I want to be especially careful here because this PR is **directly the fix for that same class of gating issue** — merging it without the test matrix actually running on the fork PR would be ironic.

cc the usual reviewers — happy to tag once I know the right deployment-approver. Will not request merge until the workflow has actually executed and gone green.

(For context: I verified locally that the change makes `CommandScopeTest`, `LiquibaseCommandLineTest`, `MainTest`, and the full `liquibase-cli` reactor go green — 413 specs across the directly-impacted set. The change that needs CI's live-postgres run is `PostgreSQLIntegrationTest.testStatusRunDuringUpdate`, which I can't run locally without Docker.)

v-petrovych:
**Follow-up after a thorough `gh`/REST-API investigation**: the PR is actually **double-gated**, not single-gated. The Validation section of the PR description described the second gate (`environment: external`); turns out there's a first one too that I didn't notice before today.

Searched the upstream Actions API exhaustively for our SHA (`599aced44`):

```
status=waiting          0/173  matches
status=action_required  0/294  matches
status=queued           0/0    matches
status=in_progress      0/0    matches
status=pending          0/47   matches
workflows/67727901/runs (full pagination)  0  matches
```

**Zero workflow runs of any status, on any workflow, for this PR's SHA.** That's stricter than "stuck waiting" — the run-tests workflow has not been *created* yet, not even in `action_required` state. That signature matches GitHub's "Require approval for first-time contributors" / "Require approval for outside collaborators" repository setting being on — workflow runs from outside-collaborator-authored PRs are held in a pre-queue state that doesn't surface in the API.

By contrast, my PR #7760 from yesterday has its `Unit and Integration Tests` run visible in the upstream history at `status=waiting` (sha `75ac55723`). So gate #1 was cleared on the previous PR but the second gate wasn't.

**What this means for merging:**

There are **two** maintainer clicks needed on this PR, not one:

1. **First**: "Approve and run workflows" button on the PR's Actions tab (or via the yellow banner). This creates the run. Without this, the run never exists.
2. **Then**: "Review deployments → Approve and run" on the `external` environment gate inside the now-created run. This actually executes the test matrix.

If anyone has a path that skips both (e.g. pushing the branch into `liquibase/liquibase` directly, which would trigger `environment: internal` and bypass both gates), that's also fine — I just want to make sure `Integration Test (postgresql)` actually executes against `testStatusRunDuringUpdate` before this merges. Local validation covered 413 unit-level specs but not the live-postgres scenario, and *that* is the scenario that's been red on upstream main for five days.

Tried `gh workflow run` / `gh workflow run --ref` / dispatching on my fork — all blocked at the outside-collaborator level. Nothing more I can do from my side until a maintainer approves the gates.

filipelautert (approved):

v-petrovych:
Refiled as #7764 — same code (SHA `62e64d528` preserved), branch now pushed directly to `liquibase/liquibase` instead of from my personal fork `v-petrovych/liquibase`.

**Why this matters:** fork PRs trigger `pull_request_target` with `head.repo != github.repository`, which routes the workflow's `authorize` job to `environment: external`. That deployment-approval gate held all CI in `waiting` state for most of this PR's lifecycle, and was the root mechanism that allowed PR #7741 (also fork-authored) to merge without its test matrix ever running — which is what shipped the `PostgreSQLIntegrationTest.testStatusRunDuringUpdate` regression to `main` and triggered this whole audit cycle ([DAT-23097](https://datical.atlassian.net/browse/DAT-23097)).

Direct-push from inside `liquibase/liquibase` routes the same workflow's `authorize` to `environment: internal` — no deployment approval, no first-time-contributor gate, CI runs on PR-open automatically (verified on #7764: `Unit and Integration Tests` is already `in_progress` without anyone clicking anything).

Closing this in favor of #7764.

PR #7764, opened 2026-05-26
CWE-316: make CommandScope.clearCredentialArguments() opt-in (fixes #7761)

### LINKED ISSUE #7761: CWE-316 regression: CommandScope.execute() finally-block clearCredentialArguments() breaks any caller re-using a CommandScope

v-petrovych:
## TL;DR

PR #7741 ("CWE-316: clear password memory after use in legacy CLI + CommandScope") added a `finally { clearCredentialArguments(); }` block to `CommandScope.execute()`. **That breaks every caller that calls `execute()` more than once on the same `CommandScope`** — the second call reads `"*****"` and the JDBC driver rejects the connection.

This PR implements **Option A** from issue #7761: remove the auto-clearing in `execute()`, make `clearCredentialArguments()` public so callers can opt in explicitly, and wire the CLI's single-use call sites in `Main.java` to call it via a small `executeAndClearCredentials()` helper. Adds three new `CommandScopeTest` specs — including the **unit-level regression coverage for `PostgreSQLIntegrationTest.testStatusRunDuringUpdate`** — so future re-introductions are caught at PR-author time without needing a live postgres in the loop.

Closes #7761.

## The concrete regression

```java
    ...
    } finally {
        clearCredentialArguments();   // <-- auto-clear after every execute()
    }
}
```

`clearCredentialArguments()` overwrites every credential-bearing argument value (`password`, `passwd`, `secret`, `token`, `apikey`, `accesskey`, `licensekey`) in the scope's `argumentValues` map with `"*****"`. A caller that reuses the same scope:

```java
CommandScope statusScope = new CommandScope("status")
    .addArgumentValue("password", realPassword);

statusScope.execute();   // 1st call: OK, uses realPassword. After: argumentValues["password"] = "*****"
// ... concurrent liquibase.update() runs ...
statusScope.execute();   // 2nd call: argumentValues["password"] is now "*****"
                          //   → JDBC driver sees "*****" → "FATAL: password authentication failed"
```

`liquibase.dbtest.pgsql.PostgreSQLIntegrationTest.testStatusRunDuringUpdate` (line 93) does exactly that — three executions on the same scope to check status before / during / after a concurrent update. First call succeeds; second/third fail with:

```
liquibase.exception.DatabaseException: Connection could not be created to
jdbc:postgresql://localhost:48822/lbcat?loggerLevel=OFF with driver
org.postgresql.Driver. FATAL: password authentication failed for user "lbuser"
```

Upstream main has been red on this test since 2026-05-21 12:15 UTC (when #7741 merged). [The failing run linked from issue #7761](https://github.com/liquibase/liquibase/actions/runs/26402847894/job/77720530913).

## Why the regression sat in upstream main for four days

PR #7741 was authored from a fork (`v-petrovych/liquibase`). Upstream's `run-tests.yml` workflow gates its `authorize` job on `environment: external` whenever the PR's head repo differs from `github.repository`. That gate requires a maintainer to click **"Review deployments → Approve and run"** before the workflow can proceed.

The gate was never approved on #7741's pre-merge build. The workflow sat in `status: waiting`. No `Run Test for (Java N <OS>)` or `Integration Test (<db>)` jobs ever materialised. The PR's 9 green checks were all from non-gated workflows (CodeQL, label, claude-review). **The merge button did not block on a workflow that never produced any check at all.**

After merge, the post-merge `push` to `main` triggered the workflow with the `internal` environment (no manual approval needed) — but by then the merge was already in. Same shape for PR #7760 (the fixture fix that finally let the reactor advance past `liquibase-standard` to `liquibase-integration-tests`).

I've filed validation guidance for the next contributor in the upstream issue so this gate doesn't continue to hide regressions on fork-authored PRs. The five checks are summarised at the bottom of this PR description.

## Option A — opt-in clearing

Three options were discussed in issue #7761:

| Option | Description | Verdict |
|---|---|---|
| **A** | Make clearing opt-in: public method on `CommandScope`; remove finally; call explicitly at single-use sites. | **Chosen.** Preserves CWE-316 intent at CLI call sites; doesn't break re-use. |
| B | Gate clearing on a `markSingleUse()` flag (backwards-compat default). | Possible, but adds API surface and a state machine that's easy to forget. |
| C | Clear on scope `close()` rather than per-execute (`AutoCloseable`). | Bigger refactor; same forget-to-close risk as B. |

### 1. `CommandScope.java`

- **Removed** the outer `finally { clearCredentialArguments(); }` block from `execute()`.
- **Changed** `clearCredentialArguments()` from package-private to **`public`** so external callers can invoke it explicitly.
- **Updated** both the `CREDENTIAL_KEY_TOKENS` doc comment and the `clearCredentialArguments()` method doc to document the new caller-invoked contract, including the specific re-use scenario the auto-finally broke and the recommended try-finally pattern for single-use callers.

### 2. `Main.java`

- **Added** a private static helper:
  ```java
  private static CommandResults executeAndClearCredentials(CommandScope commandScope) throws CommandExecutionException {
      try {
          return commandScope.execute();
      } finally {
          commandScope.clearCredentialArguments();
      }
  }
  ```
- **Replaced** every CLI `commandScope.execute()` callsite with `executeAndClearCredentials(commandScope)`. 15 sites: `snapshotCommand`, `dropAllCommand`, `calculateChecksumCommand`, `historyCommand`, `diffCommand`, `diffChangelogCommand`, `updateCommand`, `rollbackOneChangeSet` (×2), `rollbackOneUpdate` (×2), `executeSql`, plus three bare `commandScope.execute()` calls.
- **Additionally fixed a pre-existing no-op (CodeRabbit catch, commit `159fb6211`)**: `runGenerateChangelogCommandStep()` built `generateChangelogCommand` (a `CommandScope`) but never called `.execute()` on it — meaning `liquibase generateChangelog` via the legacy `Main.java` path was silently doing nothing on base SHA `8ed866995e` and earlier. **This is a 16th change that *adds* a new execute (not a mechanical wrap of an existing one).** Calling it out explicitly here per [Filipe's review feedback](https://github.com/liquibase/liquibase/pull/7764#pullrequestreview-4365743068). The fix uses the same `executeAndClearCredentials` helper this PR introduces, so the no-op is now corrected and credential clearing is in place at the same call site. If you'd prefer this as a separate PR, happy to split — but the helper is the right tool and one line is the entire delta.

Net effect: CLI users get the same CWE-316 protection they had before #7741's regression-causing fix landed, but the protection now lives at the call site (where the single-use assumption is true) rather than inside `execute()` (where the assumption was false for the integration test and any library/programmatic caller that reuses a scope).

### 3. `CommandScopeTest.groovy`

- **Updated** the two existing specs (`"execute invokes credential clearing in its outer finally..."`) to assert the new opt-in semantics: `execute()` alone does NOT touch credential values; explicit `clearCredentialArguments()` does.
- **Added** the key regression spec:
  > `"execute called twice on the same scope reads original credentials both times — regression for PostgreSQLIntegrationTest.testStatusRunDuringUpdate"`

  Calls `execute()` three times on the same scope and uses a `MockCommandStep` that captures `argumentValues["password"]` each time the pipeline runs. All three captured values must equal the original credential. **This is the unit-level regression coverage for the integration-test scenario that surfaced the bug** — running it at unit speed means future re-introductions are caught at PR-author time without needing a live postgres in the loop.
- **Added** a documentation spec for the recommended `try-finally` usage pattern.

## Tests

All passing locally on this branch (pre-merge validation):

```
CommandScopeTest                      22 / 22   (3 new specs)
LiquibaseCommandLineTest              40 / 40   (incl. 'help output' snapshot)
liquibase-cli full reactor           115 / 115
liquibase-standard Command*-related  204 / 204
MainTest                              32 / 32
```

**413 specs across the directly-impacted set, 0 failures.** Did not run the live-postgres integration tests locally — Docker not available in my setup. Pre-merge CI will need to run them; see the **Validation** section below.

## Validation (please read before merging)

Verbatim from the [Validation section](https://datical.atlassian.net/browse/DAT-23097) of the internal regression ticket — mirrored upstream as context for anyone in the org or external who authors / reviews fork PRs against `liquibase/liquibase`. The deep root cause of why both #7741 and #7760 merged without their tests running is the `environment: external` gate on the fork PR's `run-tests.yml`:

1. **Before celebrating CI-green, check the actual workflow run.** After opening the PR, go to the Actions tab → filter by branch → confirm `Unit and Integration Tests` has `conclusion: success`. If you see `status: waiting`, the tests did NOT run.
2. **If it's waiting, ask a maintainer to approve it.** Comment on the PR: *"Could a maintainer hit 'Review deployments → Approve and run' on the run-tests.yml workflow run for this PR? It's waiting on the `external` environment gate."* Don't ask for merge until that workflow has actually run and is green.
3. **If you have push access to `liquibase/liquibase` directly, prefer branching inside the org repo** rather than working from your fork. That avoids the gate entirely (the workflow picks `internal` environment and runs immediately).
4. **Run the snapshot tests locally before opening a fork PR**, especially anything touching `GlobalConfiguration`, `LiquibaseCommandLine`, or `CommandScope`. Five seconds locally beats four days of upstream-main red.
5. **When reviewing or merging a fork PR**, scan the rollup for `authorize` jobs in pending/null state — that's the explicit signal the test matrix was gated and never ran.

For this PR specifically: I'll watch the Actions tab after pushing. If the `Unit and Integration Tests` workflow lands in `status: waiting`, I'll comment asking for deployment approval before requesting merge.

## Related

- Issue liquibase/liquibase#7761 — the upstream issue this PR closes
- PR liquibase/liquibase#7741 — the original CWE-316 fix that introduced the regression (commit `4850342ae7`)
- PR liquibase/liquibase#7760 — the DAT-23096 fixture fix that finally let the reactor advance past `liquibase-standard` so this regression became visible on main
- Pro subtree-sync PR liquibase/liquibase-pro#3860 — held until this PR lands and a fresh subtree-sync picks it up
- PR liquibase/liquibase#7759 (commit `a2dd2af067a8`) — Khromushkin's drive-by inside the unrelated PARTITION-BY PR that rebuilds a fresh `CommandScope` per call inside `PostgreSQLIntegrationTest.testStatusRunDuringUpdate` to dodge the same CWE-316 regression. **Becomes redundant once this PR lands**; the workaround can be removed in a follow-up (or simply left in place — it's a harmless extra allocation per call). Heads-up so future readers don't double-take on the duplicate-looking fixes.

### DIFF
--- a/liquibase-standard/src/main/java/liquibase/command/CommandScope.java
+++ b/liquibase-standard/src/main/java/liquibase/command/CommandScope.java
@@ -52,9 +52,23 @@ public class CommandScope {
 
     /**
      * Substring tokens (lowercase) that mark an argument key as credential-bearing.
-     * Values for matching keys are overwritten with {@code "*****"} at the end of
-     * {@link #execute()} so the original credential Strings become GC-eligible and
-     * do not linger in heap for the rest of the JVM lifetime (CWE-316).
+     * Values for matching keys are overwritten with {@code "*****"} when the
+     * (public) {@link #clearCredentialArguments()} method is invoked by a caller.
+     * That makes the original credential Strings GC-eligible so they do not linger
+     * in heap for the rest of the JVM lifetime in long-running embedded scenarios
+     * (CWE-316).
+     * <p>
+     * <b>Caller-invoked, not auto-invoked from {@link #execute()}</b>. Calling it
+     * from an outer {@code finally} block in {@code execute()} (as the original
+     * CWE-316 fix did) breaks any caller that calls {@code execute()} more than
+     * once on the same {@code CommandScope} — the second call reads {@code "*****"}
+     * instead of the real credential and the JDBC driver rejects the connection.
+     * The CLI in {@code Main} wraps its single-use scopes in try-finally to call
+     * this method after each {@code execute()}; programmatic / library callers
+     * who reuse a {@code CommandScope} (e.g. status-before / update / status-after
+     * patterns) should call it once after the final {@code execute()}, never
+     * between executions. Discovered via
+     * {@code PostgreSQLIntegrationTest.testStatusRunDuringUpdate}.
      * <p>
      * Matching is intentionally substring-based on the lowercased key: keys like
      * {@code "argument__bearerToken"} are caught by {@code "token"};
@@ -379,22 +393,30 @@ public class CommandScope {
             throw e;
         } catch (Exception e) {
             throw new CommandExecutionException(e);
-        } finally {
-            clearCredentialArguments();
         }
     }
 
     /**
-     * Overwrite the values of any credential-bearing argument keys with
-     * {@code "*****"} so the original credential Strings stop being referenced by
-     * this scope. Strings are immutable and cannot be wiped in-place, but dropping
-     * the reference here makes them GC-eligible (assuming no other references hold
-     * the same String) instead of letting them live for the rest of the JVM in
-     * long-running embedded scenarios (CWE-316: Cleartext Storage of Sensitive
-     * Information in Memory). Called from {@link #execute()}'s outer finally so
-     * it runs regardless of command outcome; package-private for direct testing.
+     * Overwrite the values of any credential-bearing argument keys (per
+     * {@link #CREDENTIAL_KEY_TOKENS}) with {@code "*****"} so the original
+     * credential Strings stop being referenced by this scope. Strings are
+     * immutable and cannot be wiped in-place, but dropping the reference here
+     * makes them GC-eligible (assuming no other references hold the same String)
+     * instead of letting them live for the rest of the JVM in long-running
+     * embedded scenarios (CWE-316: Cleartext Storage of Sensitive Information
+     * in Memory).
+     * <p>
+     * <b>Caller-invoked.</b> This method must NOT be auto-invoked from
+     * {@link #execute()} — doing so (as the original CWE-316 fix did) breaks any
+     * caller that calls {@code execute()} more than once on the same scope, since
+     * the second call reads {@code "*****"} instead of the real credential.
+     * Callers that reuse a scope (e.g. status-before / update / status-after
+     * patterns) must call this method <b>after the final</b> {@code execute()},
+     * never between executions. Callers that use a scope only once (such as the
+     * CLI in {@code Main}) typically wrap the {@code execute()} call in a
+     * try-finally and call this in the {@code finally}.
      */
-    void clearCredentialArguments() {
+    public void clearCredentialArguments() {
         if (argumentValues.isEmpty()) {
             return;
         }
--- a/liquibase-standard/src/main/java/liquibase/integration/commandline/Main.java
+++ b/liquibase-standard/src/main/java/liquibase/integration/commandline/Main.java
@@ -1478,7 +1478,7 @@ public class Main {
                 // otherwise, we will print the output
                 //
                 Writer outputWriter = getOutputWriter();
-                CommandResults commandResults = snapshotCommand.execute();
+                CommandResults commandResults = executeAndClearCredentials(snapshotCommand);
                 String result = InternalSnapshotCommandStep.printSnapshot(snapshotCommand, commandResults);
                 outputWriter.write(result);
                 outputWriter.flush();
@@ -1521,7 +1521,7 @@ public class Main {
                         .addArgumentValue(DropAllCommandStep.CATALOG_AND_SCHEMAS_ARG, InternalSnapshotCommandStep.parseSchemas(database, getSchemaParams(database)))
                         .addArgumentValue(GenerateChangelogCommandStep.CHANGELOG_FILE_ARG, changeLogFile);
 
-                dropAllCommand.execute();
+                executeAndClearCredentials(dropAllCommand);
                 return;
             } else if (COMMANDS.STATUS.equalsIgnoreCase(command)) {
                 boolean runVerbose = false;
@@ -1558,7 +1558,7 @@ public class Main {
                         .addArgumentValue(CalculateChecksumCommandStep.CHANGESET_IDENTIFIER_ARG, getCommandParam(OPTIONS.CHANGE_SET_IDENTIFIER, null))
                         .addArgumentValue(CalculateChecksumCommandStep.CHANGELOG_FILE_ARG, this.changeLogFile);
 
-                calculateChecksumCommand.execute();
+                executeAndClearCredentials(calculateChecksumCommand);
                 return;
             } else if (COMMANDS.DB_DOC.equalsIgnoreCase(command)) {
                 if (commandParams.isEmpty()) {
@@ -1704,7 +1704,7 @@ public class Main {
                     historyCommand.addArgumentValue(HistoryCommandStep.FORMAT_ARG, HistoryFormat.valueOf(format));
                     historyCommand.setOutput(getOutputStream());
 
-                    historyCommand.execute();
+                    executeAndClearCredentials(historyCommand);
                 } else {
                     throw new CommandLineParsingException(
                             String.format(coreBundle.getString("command.unknown"), command));
@@ -1782,13 +1782,13 @@ public class Main {
         if (outputWriter != null) {
             commandScope.setOutput(new WriterOutputStream(outputWriter, GlobalConfiguration.OUTPUT_FILE_ENCODING.getCurrentValue()));
         }
-        commandScope.execute();
+        executeAndClearCredentials(commandScope);
     }
 
     private void runReleaseLocksCommand() throws CommandExecutionException {
         CommandScope commandScope = new CommandScope(ReleaseLocksCommandStep.COMMAND_NAME[0]);
         this.setDatabaseArgumentsToCommand(commandScope);
-        commandScope.execute();
+        executeAndClearCredentials(commandScope);
     }
 
     private void runGenerateChangelogCommandStep() throws LiquibaseException, IOException, CommandLineParsingException {
@@ -1827,7 +1827,7 @@ public class Main {
         this.setDatabaseArgumentsToCommand(diffChangelogCommand);
         this.setReferenceDatabaseArgumentsToCommand(diffChangelogCommand);
 
-        diffChangelogCommand.execute();
+        executeAndClearCredentials(diffChangelogCommand);
     }
 
     private void runDiffCommandStep() throws CommandLineParsingException, CommandExecutionException, IOException {
@@ -1839,7 +1839,7 @@ public class Main {
         this.setDatabaseArgumentsToCommand(diffCommand);
         this.setReferenceDatabaseArgumentsToCommand(diffCommand);
 
-        diffCommand.execute();
+        executeAndClearCredentials(diffCommand);
     }
 
     private void runUpdateCommandStep() throws CommandLineParsingException, CommandExecutionException, IOException {
@@ -1850,7 +1850,7 @@ public class Main {
         updateCommand.addArgumentValue(ChangeExecListenerCommandStep.CHANGE_EXEC_LISTENER_CLASS_ARG, changeExecListenerClass);
         updateCommand.addArgumentValue(ChangeExecListenerCommandStep.CHANGE_EXEC_LISTENER_PROPERTIES_FILE_ARG, changeExecListenerPropertiesFile);
         setDatabaseArgumentsToCommand(updateCommand);
-        updateCommand.execute();
+        executeAndClearCredentials(updateCommand);
     }
 
     private void runRollbackOneChangeSetCommandStep() throws CommandExecutionException, CommandLineParsingException {
@@ -1865,7 +1865,7 @@ public class Main {
                 .addArgumentValue("force", getCommandParam(OPTIONS.FORCE, null));
         String internalCommand = "rollbackOneChangeset";
         Scope.getCurrentScope().addMdcValue(MdcKey.LIQUIBASE_INTERNAL_COMMAND, internalCommand);
-        rollbackOneChangeSet.execute();
+        executeAndClearCredentials(rollbackOneChangeSet);
     }
 
     private void runRollbackOneChangeSetSqlCommandStep() throws CommandExecutionException, CommandLineParsingException, IOException {
@@ -1881,7 +1881,7 @@ public class Main {
         String internalCommand = "rollbackOneChangeset";
         Scope.getCurrentScope().addMdcValue(MdcKey.LIQUIBASE_INTERNAL_COMMAND, internalCommand);
         rollbackOneChangeSet.setOutput(new WriterOutputStream(getOutputWriter(), GlobalConfiguration.OUTPUT_FILE_ENCODING.getCurrentValue()));
-        rollbackOneChangeSet.execute();
+        executeAndClearCredentials(rollbackOneChangeSet);
     }
 
     private void runRollbackOneUpdateCommandStep() throws CommandExecutionException, CommandLineParsingException {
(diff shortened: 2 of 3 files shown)


v-petrovych:
@coderabbitai Thanks — addressed in commit `159fb6211`.

You correctly identified that `runGenerateChangelogCommandStep()` built the `CommandScope` but never called `.execute()` on it, making the legacy CLI `generateChangelog` command a silent no-op.

Verified the bug is pre-existing on upstream main (`8ed86699` shows the same broken method); not introduced by this PR. But since this PR introduces the `executeAndClearCredentials` helper that's the right tool for the fix, applying it here is the natural landing place. One-line addition, both `CommandScopeTest` (22/22) and `LiquibaseCommandLineTest` (40/40) pass locally with the fix.

Tests covering `generateChangelog` semantics (`GenerateChangeLogMySQL/MSSQL/Empty/Db2IntegrationTest`) go through `CommandUtil.groovy` → `CommandScope` directly, so they were already exercising the command framework; this fix just makes the Main.java legacy path no longer silently skip execution.

v-petrovych:
## Root cause of the failing `run-build-publish-file / Artifacts - Build & Package` job

Pulled the actual workflow logs and found the exact mechanism. This is **a latent pre-existing bug in `.github/workflows/build.yml`**, not a code regression from this PR.

### The chain of events

1. The reusable workflow [`.github/workflows/build.yml` (lines 43-46)](https://github.com/liquibase/liquibase/blob/main/.github/workflows/build.yml#L43-L46) has a setup step that computes the head SHA:
   ```yaml
   - name: Get Latest Merge Commit SHA
     id: get-sha
     run: |
       latest_merge_sha=`(git rev-parse HEAD)`
       echo "latestMergeSha=${latest_merge_sha}" >> $GITHUB_OUTPUT
   ```

2. The Setup job logs from our failed run [job #77902122656](https://github.com/liquibase/liquibase/actions/runs/26457697788/job/77902122656) show:
   ```
   ##[warning]Skip output 'latestMergeSha' since it may contain secret.
   ```
   GitHub Actions' secret-scrubbing heuristic refused to emit the 40-character hex SHA as a job output, treating it as a possible leaked secret.

3. Downstream, the deploy job [`build_publish_branch` (line 243)](https://github.com/liquibase/liquibase/blob/main/.github/workflows/build.yml#L243) consumes that output:
   ```yaml
   - name: Version Artifact Publish <commitsha>-SNAPSHOT version to GPM
     run: |
       ./mvnw versions:set "-DnewVersion=${{ needs.setup.outputs.latestMergeSha }}-SNAPSHOT"
   ```
   With `latestMergeSha` scrubbed to empty, the command resolves to:
   ```
   ./mvnw versions:set "-DnewVersion=-SNAPSHOT"
   ```
   Visible verbatim in the Artifacts job log.

4. Maven happily accepts `-SNAPSHOT` as a version (no leading version component), installs to local m2 as `org/liquibase/liquibase-core/-SNAPSHOT/liquibase-core--SNAPSHOT.pom`, then `maven-deploy-plugin` PUTs the malformed version to GPM:
   ```
   https://maven.pkg.github.com/liquibase/liquibase/org/liquibase/liquibase-core/-SNAPSHOT/maven-metadata.xml
   → status code: 400, reason phrase: Bad Request
   ```
   GPM rejects the empty-version path.

### Why this hasn't been visible before

- The deploy job is **skipped** on push-to-main events (`if` conditional, confirmed against 4 recent push runs) so release deploys never hit this code path.
- It only runs on `pull_request_target` events. Previously, every PR from this contributor was a fork PR; the upstream-level workflow gate kept those PRs from reaching this job at all. So the latent bug never manifested in CI history.
- This PR is the first direct-push (non-fork) PR that surfaces it — the gate is gone, the deploy job actually runs, and the SHA-scrubbing trips.

### What's not the issue

- ✗ Not a code regression from this PR — the CWE-316 changes only touch `CommandScope.java`, `Main.java`, and `CommandScopeTest.groovy`; none of those affect Maven version computation or deploy.
- ✗ Not Maven configuration — the `liquibase-core/pom.xml` and parent POM are unchanged.
- ✗ Not the deploy credentials — auth to GPM succeeds (the 400 is on Maven metadata content, not auth).
- ✗ Not flaky — same scrubbing happened on prior SHA `62e64d52`, will repeat on `159fb6211` and any future SHA.

### Suggested fix (for the workflow, not this PR)

The cleanest fix is to bypass the scrubbed output and use `github.event.pull_request.head.sha` (or `github.sha`) directly in the deploy step, which is already a workflow-context variable not subject to step-output scrubbing:

```yaml
- name: Version Artifact Publish <commitsha>-SNAPSHOT version to GPM
  run: |
    ./mvnw versions:set "-DnewVersion=${{ github.event.pull_request.head.sha || github.sha }}-SNAPSHOT"
```

Alternatively, mask the SHA differently when echoing to `$GITHUB_OUTPUT` (e.g., prefix it with a non-hex character so the heuristic doesn't match), or pin the version to use a non-SHA identifier like the PR number.

This belongs in a separate workflow-only PR — out of scope for this CWE-316 fix.

### Other 'failing' checks on this PR

The check rollup also lists two failures from run `26453121237` and `26453123515` — those are bleed-through from the closed PR #7763's SHA `62e64d52` (same SHA was preserved when refiling here). They're stale entries from an unrelated run, not failures of CI on this PR's current SHA `159fb6211`. CI for the current SHA is in progress: 5/7 unit-matrix jobs already passing, integration tests next.

filipelautert (commented):
Hey Victor, reviewed — the opt-in refactor is sound and the 3×-execute-on-same-scope regression spec is exactly the right shape to prevent re-introduction. One question before approve:

1. **`Main.java:1812` — `runGenerateChangelogCommandStep` gains an `executeAndClearCredentials(generateChangelogCommand)` call that wasn't there before.** Checked the base branch (commit `8ed866995e`): the method previously ended after `setPreCompareArgumentsToCommand(...)` without ever calling `.execute()` on the scope — meaning `liquibase generate-changelog` on the legacy `Main.java` path was a no-op. Is that an intentional bug fix you noticed in passing, or a copy-paste slip? PR body lists 15 sites of `.execute()` → `executeAndClearCredentials(...)` substitutions; this is a 16th change adding a *new* execute. Either call it out in the body ("also fixes a pre-existing no-op") or split it off — the rest of the PR is mechanical, this is a new behavior change.

Heads-up (not a blocker): Khromushkin's drive-by `a2dd2af06` in PR #7759 (rebuild fresh `CommandScope` per call in `PostgreSQLIntegrationTest.testStatusRunDuringUpdate`) becomes redundant once this lands. Harmless to leave, but worth a sentence in your PR body so future reviewers don't double-take.

Validation section + the "fork PR `external` env gate hid the failure for 4 days" diagnosis are exemplary. The unit-level regression coverage is the right move — postgres-in-the-loop for what's effectively a Java-side state machine is overkill.

v-petrovych:
@filipelautert thanks — both points are fair, both addressed in the PR body just now.

**1. `runGenerateChangelogCommandStep` (commit `159fb6211`)** — fully agreed it deserves an explicit call-out. CodeRabbit flagged the no-op while reviewing this PR, I applied the fix using the helper this PR introduces, but I didn't update the PR body to reflect that it's a 16th *additive* change rather than another mechanical substitution. Now called out under `### 2. Main.java` with the "pre-existing no-op (CodeRabbit catch)" framing — verified against base SHA `8ed866995e` exactly as you did. Confirms your reading: legacy CLI `generateChangelog` was silently doing nothing on `main` prior to this PR.

Happy to split it into a separate PR if you'd prefer — the one-line delta is contained enough that either landing makes sense. My preference is to keep it here because (a) the helper that's the right fix only exists in this PR, (b) it lets the helper be exercised on its own day-zero rather than waiting for a follow-up that needs to rebase on this. But your call.

**2. Nikita's `a2dd2af067a8` in #7759** — added an explicit "Related" bullet describing the drive-by and noting it becomes redundant after this lands. Left the recommendation neutral ("can be removed in a follow-up or simply left in place — it's a harmless extra allocation") because the PR #7759 owner can decide what's least churn for their context. The note exists now so future readers see the duplicate-looking fix isn't a conflict.

Updated PR body diff visible in the edit history; no code changes since `159fb6211` (last commit). All CI is green on that SHA including `Integration Test (postgresql)` (the regression test that's currently red on `main`). Ready for your re-review when you have a moment.

filipelautert (approved):
Both concerns addressed cleanly. The `runGenerateChangelogCommandStep` fix is right to keep here — helper and first non-mechanical caller land day-zero. Regression test is exactly the right shape. Approving and merging to unblock the CWE branches.

PR #7768, opened 2026-05-26
CWE-470: opt-in restricted mode for customPrecondition changelog element

v-petrovych:
## Impact

`CustomPreconditionWrapper.check(...)` (`CustomPreconditionWrapper.java:67-69`) loads an arbitrary JVM class by FQCN via the same `Class.forName(className, true, ...)` pattern as `customChange`:

```java
customPrecondition = (CustomPrecondition) Class.forName(className, true,
//                                              ^ static <clinit> fires here, before the cast
```

Same Unsafe-Reflection attack surface as B2 (DAT-23016 / PR #7748) — **CWE-470**.

**An additional attack vector unique to this PR**, called out explicitly in the audit ticket: preconditions are evaluated **before** the change body executes. An attacker can craft a precondition with `onFail=MARK_RAN` that *"always passes"* and the change body is never reached — but the load-time `<clinit>` already ran. So the fix must reject the load **before** `Class.forName`, **not** rely on the precondition's pass/fail logic.

## Cross-PR coordination (please read)

This PR is the **third sibling** in a B1/B2/B3 pattern:

| PR | Element | Flag |
|---|---|---|
| #7747 | `<executeCommand>` | `liquibase.allowExecuteCommand` |
| #7748 | `<customChange>` | `liquibase.allowCustomChange` |
| **THIS** | **`<customPrecondition>`** | **`liquibase.allowCustomChange` (same as #7748)** |

The audit ticket's exact words: *"Covered by the same `liquibase.execution.allowCustomClassLoad=false` opt-in flag proposed for B2."* Both `customChange` and `customPrecondition` are "load arbitrary FQCN" surfaces with identical risk — one flag covers both.

### Expected merge conflict with #7748

Both this PR and #7748 declare `GlobalConfiguration.ALLOW_CUSTOM_CHANGE`. The declarations are byte-identical for the constant; the **description text** differs:

- **#7748's description** mentions only `customChange`.
- **THIS PR's description** mentions both `customChange` AND `customPrecondition` (the larger surface that's accurate after both PRs are in `main`).

**Recommended merge resolution:** when merging the second of the two PRs, keep this PR's broader description text and discard #7748's narrower one. The constant declaration itself dedupes mechanically (1-line edit). I considered stacking this PR on #7748's branch to avoid the conflict but chose independent PRs to keep each one reviewable in any merge order — the conflict is mechanical and 1-line.

## Fix — three pieces

**1. `GlobalConfiguration.ALLOW_CUSTOM_CHANGE`** — Boolean, defaults to `true`. Description covers both `customChange` and `customPrecondition` surfaces and ties them to CWE-470.

**2. Dual gate inside `CustomPreconditionWrapper`:**

   **(a) Hard error in `validate(Database)`.** The pre-fix `validate()` was a trivial no-op (`return new ValidationErrors();`). This PR replaces that body with a configured-off check, so the operator sees a pre-execution failure during the standard validation pass.

   **(b) Defense-in-depth gate at the top of `check(...)`** — fires BEFORE either of the two `Class.forName` call sites at lines 67/69. Throws `PreconditionErrorException`, **not** `PreconditionFailedException`. The choice is deliberate: `Error` bypasses `onFail` handling, so a crafted `onFail=MARK_RAN` precondition cannot silently swallow the embedder's configured-off intent — which is the specific attack vector the audit calls out.

**3. Four new Spock specs in `CustomPreconditionWrapperTest`** (2 pre-existing → 6 total):

| Spec | What it pins |
|---|---|
| `validate returns no errors by default` | Default-true path preserves existing behaviour |
| `validate fails when liquibase.allowCustomChange=false — embedder opt-out path` | Hard error, names the flag in both directions AND names `customPrecondition` specifically so the message points at the offending element |
| `check throws PreconditionErrorException BEFORE Class.forName when allowCustomChange=false` | Sets a non-existent class name to prove the gate ran first (no wrapped `ClassNotFoundException` appears). Asserts `PreconditionErrorException` not `Failed` — pins the audit's `onFail=MARK_RAN`-bypass requirement |
| `check default-true path still executes the existing load logic (sanity)` | Default-true short-circuit doesn't skip downstream behaviour |

```
[INFO] Tests run: 6, Failures: 0, Errors: 0, Skipped: 0 -- in CustomPreconditionWrapperTest
[INFO] BUILD SUCCESS
```

## Things to be aware of

- **Default is unchanged.** Existing `customPrecondition` users see byte-for-byte identical behaviour.
- **Both gates fire before any static initializer.** The defense-in-depth design specifically prevents the load-time side effect — not just the cast or constructor invocation.
- **PreconditionErrorException (not Failed) is the intentional choice** at the `check()` gate. A `PreconditionFailedException` would respect `onFail` handling, allowing `onFail=MARK_RAN` to silently swallow the embedder's configured-off intent — exactly the attack vector the audit calls out. The unit test pins this.
- **Chain 3 status:** with this PR (B3) landing alongside #7747 (B1) and #7748 (B2), **all three RCE primitives flagged in DAT-23027 (the env-var command-injection chain) now have embedder opt-out gates.** B10 (env-var substitution behaviour on changeset properties) is a separate child ticket; closing the chain fully requires B10 to land too.

## Related

- **Direct sibling**: #7748 (DAT-23016, B2 / `customChange` opt-out gate) — same flag, parallel implementation. The two are designed to land together.
- **Cousin**: #7747 (DAT-23015, B1 / `executeCommand` opt-out gate) — same opt-in restricted-mode shape, different flag and different change tag.
- Other audit slice PRs: #7739, #7740, #7741, #7742, #7743, #7744, #7746. Merged: #7737, #7738.

### DIFF
--- a/liquibase-cli/src/test/groovy/liquibase/integration/commandline/LiquibaseCommandLineTest.groovy
+++ b/liquibase-cli/src/test/groovy/liquibase/integration/commandline/LiquibaseCommandLineTest.groovy
@@ -22,17 +22,18 @@ Command-specific help: "liquibase <command-name> --help"
 
 Global Options
       --allow-custom-change=PARAM
-                             If false, the customChange changelog change is
-                               rejected before its named class is loaded.
-                               Defaults to true to preserve the documented
-                               customChange feature for the standard trust
-                               model (team-authored, team-reviewed changelogs).
-                               Set to false in environments that execute
-                               changelogs from less-trusted sources
-                               (multi-tenant SaaS running customer changelogs,
-                               downloaded change-packs, contributor PRs prior
-                               to review): customChange loads an arbitrary JVM
-                               class by FQCN via Class.forName
+                             If false, the customChange and customPrecondition
+                               changelog elements are rejected before the named
+                               class is loaded. Defaults to true to preserve
+                               the documented custom-Java features for the
+                               standard trust model (team-authored,
+                               team-reviewed changelogs). Set to false in
+                               environments that execute changelogs from
+                               less-trusted sources (multi-tenant SaaS running
+                               customer changelogs, downloaded change-packs,
+                               contributor PRs prior to review): both
+                               customChange and customPrecondition load an
+                               arbitrary JVM class by FQCN via Class.forName
                                (initialize=true), which fires the class's
                                static <clinit> initializer at load time —
                                before any cast or marker-interface check could
--- a/liquibase-standard/src/main/java/liquibase/GlobalConfiguration.java
+++ b/liquibase-standard/src/main/java/liquibase/GlobalConfiguration.java
@@ -53,6 +53,7 @@ public class GlobalConfiguration implements AutoloadedConfigurations {
     public static final ConfigurationDefinition<Boolean> STRICT;
     public static final ConfigurationDefinition<Integer> DDL_LOCK_TIMEOUT;
     public static final ConfigurationDefinition<Boolean> SECURE_PARSING;
+    public static final ConfigurationDefinition<Boolean> ALLOW_CUSTOM_CHANGE;
     public static final ConfigurationDefinition<String> SEARCH_PATH;
 
     public static final ConfigurationDefinition<UIServiceEnum> UI_SERVICE;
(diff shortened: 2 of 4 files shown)


filipelautert (approved):
Re-approving the org-repo replacement for #7749 (closed-without-merge fork PR — `environment: external` gate workaround you described in Slack). Diff is byte-identical to the version I approved on Monday (verified). Original architectural review still applies: shared `ALLOW_CUSTOM_CHANGE` flag with customChange, `PreconditionErrorException` (not Failed) correctly bypasses `onFail=MARK_RAN`, hermetic test scoping with explicit `Scope.child([... : "true"], ...)`, non-existent-classname spec proves the gate fires before `Class.forName`. Ship it.

filipelautert (approved):
Rebase onto post-#7764 main turned PG integration test green. Help-output test update is the only delta from the prior approval. Ship it.

PR #7781, opened 2026-06-02
feat(ui): console-only formatter and typed message (enables colored output)

jandroav:
# Enable colored console output (the plumbing)

## The problem
Liquibase prints everything in one color. Anything built on top of Liquibase — like Liquibase Secure — has no clean, supported way to highlight errors or success in the console. Today the only options are hacks that risk leaking color codes into log files.

## What this does
Adds a small, safe set of extension points so an add-on **can** color console output — and nothing more. On its own, this change does nothing you'd notice. **Default Liquibase output looks exactly the same.**

## The result
- Add-ons can style console text (e.g. color).
- Log files stay clean — color can't leak in.
- A message can be marked "success."
- 100% backward compatible by default.

## Why it's safe
Every addition is a no-op unless an add-on opts in. Existing behavior, log files, and output are untouched. Tests lock in the "nothing changes by default" guarantee.

## Companion PR
This is only the plumbing. The actual colors ship in Liquibase Secure: **liquibase/liquibase-pro#3920**.

---

## Under the hood (for reviewers)
Two seams, both inert until something overrides them:

- **Console-only formatter.** `LogService.getConsoleFormatter()` (default `null`) lets a service style *only* the console handler; `JavaLogService.setConsoleFormatterOnHandler(...)` + `LiquibaseCommandLine.configureLogging()` apply it to the `ConsoleHandler` while file handlers keep the plain formatter — so ANSI never reaches log files.
- **Typed message.** `UIService.MessageType` (currently just `SUCCESS`) plus a default `sendMessage(String, MessageType)` that delegates to the existing `sendMessage(String)`; `CompositeUIService` forwards it to all output services. The "executed successfully" message now passes `MessageType.SUCCESS`.

**Tests:** default no-op behavior, per-sink formatter routing (console styled, file handler stays plain), and typed-message delegation/forwarding. `liquibase-standard` and `liquibase-cli` suites pass.

> **Note for the downstream sync.** This PR currently carries the original two seams (above). As more outputs were colored, the Secure PR (#3920) grew the seam further; those additions live on the Pro branch's `core/` but are **not yet in this PR**. They should land here so the next `core/` subtree sync reconciles cleanly:
> - `UIService.MessageType` values `RUNNING`, `BANNER`, and `ERROR` (alongside `SUCCESS`), plus the `sendMessage(..., type)` call-site tags that use them.
> - `UIService.formatMessage(String, MessageType)` (default returns the text unchanged) and `CompositeUIService` delegation — lets picocli print a pre-styled `--version` banner.
> - a console-only `ShowSummaryGenerator.colorizeForConsole(...)` hook for the update summary.
>
> Each is backward-compatible and a no-op unless an implementation overrides it — same as the seams above.

### DIFF
--- a/liquibase-cli/src/main/java/liquibase/integration/commandline/LiquibaseCommandLine.java
+++ b/liquibase-cli/src/main/java/liquibase/integration/commandline/LiquibaseCommandLine.java
@@ -424,7 +424,7 @@ public class LiquibaseCommandLine {
                                 }
                                 final List<CommandLine> commandList = commandLine.getParseResult().asCommandLineList();
                                 final String commandName = StringUtil.join(getCommandNames(commandList.get(commandList.size() - 1)), " ");
-                                Scope.getCurrentScope().getUI().sendMessage("Liquibase command '" + commandName + "' was executed successfully.");
+                                Scope.getCurrentScope().getUI().sendMessage("Liquibase command '" + commandName + "' was executed successfully.", liquibase.ui.UIService.MessageType.SUCCESS);
                             }
                         }
                         return response;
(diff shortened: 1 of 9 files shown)


filipelautert (commented):
Hey, took a look at the plumbing while it's still in draft — the design is clean. Adding `getConsoleFormatter()` as an interface default returning `null` keeps every existing `LogService` impl untouched, and routing only `ConsoleHandler` through `setConsoleFormatterOnHandler` while file handlers stay on the plain-formatter path is exactly the right way to keep ANSI codes out of log files. Test coverage on the fallback chain and null-safety is thorough.

One thing I checked specifically since it's a public API: the new `UIService.sendMessage(String, MessageType)` overload is a `default` that delegates to `sendMessage(String)`, so third-party `UIService` implementors won't break on upgrade — nice. Same pattern on the new `MessageType` enum. Good groundwork.

PR #7800, opened 2026-06-17
fix(TECHOPS-650): replace actions/cache with upload/download-artifact; simplify dry-run artifact scan

jnewton03:
### `create-release.yml` — fix cache race (TECHOPS-650)
- `actions/cache` was used to transfer `re-version/final/` from the `reversion` job to `build-installers`. It is eventually-consistent — not a reliable inter-job transport. A 15-second save→restore window on identical code/version produced a HIT on 2026-06-03 and a MISS on 2026-06-17, failing every dry-run and production release with `tar: Cannot open: No such file or directory`.
- Replaced `Cache Completed Artifacts` (reversion job) with `actions/upload-artifact@043fb46d...` (v7.0.1) and `Restore Completed Artifacts` (build-installers job) with `actions/download-artifact@3e5f45b...` (v8.0.1). Both SHAs were already pinned in this file.
- `download-artifact` blocks until the named artifact is available, eliminating the race by design. Directory layout (`re-version/final/<file>`) is unchanged.

### `dry-run-release.yml` — simplify artifact-validity scan
- Replaced two `curl -s -H "Authorization: token ..."` calls with `gh api` + `GH_TOKEN` env var, matching the pattern already used in `installer-build-check.yml` and peers.
- Added `status=completed` to the runs query to skip in-progress runs that can't have stable artifacts.
- Dropped `valid="${valid:-0}"` (dead code — `jq length` never returns empty string).
- Collapsed `valid=$(... | length)` + numeric comparison into `gh api --jq 'any(...)'` + string compare against `"true"`.
- Used `${{ github.repository }}` instead of hardcoded `liquibase/liquibase`.

## Root cause context

The `create-release.yml` bug is a 4-year-old latent nondeterministic race in GitHub's cache backend (introduced 2022-03-25 by `41b1e2fcf`), not a recent regression. It was masked by TECHOPS-620 failures dying earlier in the pipeline; run `27695042493` (2026-06-17) was the first to reach `build-installers` since 06-03, surfacing it.

## Test Plan

**`create-release.yml` fix** — trigger directly on this branch (dry-run-release.yml dispatches create-release.yml with `ref: main`, so it won't exercise this change):
- [ ] `gh workflow run create-release.yml --ref TECHOPS-650 --field version=dry-run-test --field branch=main --field runId=<recent-run-tests-run-id> --field dry_run=true --field standalone_zip=false`
- [ ] `reversion` "Upload Completed Artifacts" succeeds
- [ ] `build-installers` "Download Completed Artifacts" populates `re-version/final/` (no "Cache not found")
- [ ] "Build Unsigned Windows Installer" opens the tar.gz successfully (was the failure point)
- [ ] Run 2-3x to confirm determinism

**`dry-run-release.yml` simplification** — trigger on this branch:
- [ ] `gh workflow run dry-run-release.yml --ref TECHOPS-650`
- [ ] "Get run-tests.yml runId" step logs `Found valid liquibase-artifacts in run-tests run <id>` and exits 0

Fixes: TECHOPS-650
Related: TECHOPS-620

### DIFF
--- a/.github/workflows/create-release.yml
+++ b/.github/workflows/create-release.yml
@@ -177,11 +177,12 @@ jobs:
           (cd re-version/out/ && zip liquibase-additional-${{ needs.setup.outputs.version }}.zip *)
           mv re-version/out/liquibase-additional-${{ needs.setup.outputs.version }}.zip re-version/final
 
-      - name: Cache Completed Artifacts
-        uses: actions/cache@27d5ce7f107fe9357f9df03efb73ab90386fccae # v5.0.5
+      - name: Upload Completed Artifacts
+        uses: actions/upload-artifact@043fb46d1a93c77aae656e7c1c64a875d1fc6a0a # v7.0.1
         with:
-          key: completed-artifacts-${{ github.run_number }}-${{ github.run_attempt }}
+          name: completed-artifacts-${{ github.run_number }}-${{ github.run_attempt }}
           path: re-version/final
+          retention-days: 1
 
       - name: Set repository tags
         if: ${{ inputs.standalone_zip == false && inputs.dry_run == false }}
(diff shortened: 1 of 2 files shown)


sayaliM0412 (approved):
LGTM !

PR #7801, opened 2026-06-18
fix: preserve addColumn before foreign key in diff changelog output

### LINKED ISSUE #1792: Generated diffChangeLog Creates Constraint Before Creating Column Used In the Constraint
## Environment

**Liquibase Version**:
Liquibase version 4.3.2 and 4.2.2

CLI

MSSQL 2019 and Azure SQL Database

Used Docker running on Linux virtual machine

## Description

When adding a new column and a foreign key to a table on an existing database, running diffChangeLog generates the constraint before the column.  However, this is incorrect because the column doesn't exist when the constraint is being created.

## Steps To Reproduce

First, create a database (I called it Original) with the following schema:
```sql
CREATE TABLE Test
(
  ID BIGINT IDENTITY(1,1),
  PRIMARY KEY (ID)
);

CREATE TABLE  Test2
(
  ID BIGINT IDENTITY(1,1),
  PRIMARY KEY (ID)
);
```

Then another database (I called it NewOriginal) that contains the following database
```sql
CREATE TABLE Test
(
  ID BIGINT IDENTITY(1,1),
  PRIMARY KEY (ID)
);

CREATE TABLE  Test2
(
  ID BIGINT IDENTITY(1,1),
  testID BIGINT NOT NULL,
  PRIMARY KEY (ID),
  FOREIGN KEY (testID) REFERENCES Test(ID) ON DELETE CASCADE
);
```

Now open up the command line and run the following command to get the differences:
```
liquibase --url="jdbc:sqlserver://mssql:1433;database=Original;"  --username=sa  --password=p@ssw0rd2021 --referenceUrl="jdbc:sqlserver://mssql:1433;database=NewOriginal;"    --referenceUsername=sa  --referencePassword=p@ssw0rd2021 --changeLogFile=/liquibase/changelog/changes.mssql.sql diffChangeLog
```

Then run the command to try to update the database:
```
liquibase --url="jdbc:sqlserver://mssql:1433;database=Original;"  --username=sa  --password=p@ssw0rd2021 --changeLogFile=changes.mssql.sql update
```

## Actual Behavior

The following console output is shown
```
Unexpected error running Liquibase: Migration failed for change set changes.mssql.sql::1617893220721-1::liquibase:
     Reason: liquibase.exception.DatabaseException: Foreign key 'FK__Test2__testID__267ABA7A' references invalid column 'testID' in referencing table 'Test2'. [Failed SQL: (1769) ALTER TABLE Test2 ADD CONSTRAINT FK__Test2__testID__267ABA7A FOREIGN KEY (testID) REFERENCES Test (ID) ON UPDATE NO ACTION ON DELETE CASCADE]
For more information, please use the --logLevel flag
```

The changeLog file ends up as:
```sql
-- liquibase formatted sql

-- changeset liquibase:1617893220721-1
ALTER TABLE Test2 ADD CONSTRAINT FK__Test2__testID__267ABA7A FOREIGN KEY (testID) REFERENCES Test (ID) ON UPDATE NO ACTION ON DELETE CASCADE;

-- changeset liquibase:1617893220721-2
ALTER TABLE Test2 ADD testID bigint NOT NULL;
```

## Expected/Desired Behavior

The changeLog file should have gone as follows

```sql
-- liquibase formatted sql

-- changeset liquibase:1617893220721-2
ALTER TABLE Test2 ADD testID bigint NOT NULL;

-- changeset liquibase:1617893220721-1
ALTER TABLE Test2 ADD CONSTRAINT FK__Test2__testID__267ABA7A FOREIGN KEY (testID) REFERENCES Test (ID) ON UPDATE NO ACTION ON DELETE CASCADE;
```

The deployment would then be successful.

ET-TOUNANI:
## Impact

## Description

Fixes #1792

When `diffChangeLog` generates changes for some SQL Server cases, it can place an `addForeignKeyConstraint` change before the `addColumn` change that creates the referenced column. That output is wrong and can fail when applied.

The problem was in dependency sorting inside `DiffToChangeLog`. Sortable objects were dependency-sorted first, but non-sortable objects such as `Column` were always appended to the end afterward. That destroyed the required relative order and allowed foreign keys to move ahead of added columns.

This change keeps dependency sorting for sortable objects, but preserves the original positions of non-sortable objects when the sorted result is merged back together. In practice, that keeps missing columns ahead of dependent foreign key changes.

A regression test was added for the reported behavior. The test verifies that the generated change sets emit `AddColumnChange` before `AddForeignKeyConstraintChange` when dependency sorting is enabled.

## Release note

Fixed `diffChangeLog` output ordering in cases where a foreign key depends on a newly added column. Liquibase now keeps the column addition before the foreign key change so the generated changelog applies in the correct order.

## Things to be aware of

- This change is intentionally narrow. It only changes how sorted and non-sortable database objects are merged after dependency sorting.
- The dependency ordering logic is still used for sortable objects.
- A regression test covers the SQL Server-style missing column + foreign key case from this issue.

## Things to worry about
None.
## Additional Context

Validation run for this change:

- `./mvnw -pl liquibase-standard -DskipTests compile`
- `./mvnw -pl liquibase-standard -Dtest=DiffToChangeLogTest test`


filipelautert (approved):
Hello @ET-TOUNANI !
Ran DiffToChangeLogTest on this branch locally, all green. Keeping the unsortable positions instead of appending them at the end is the right call for #1792. Heads up that #7862 refactors the same file, whoever merges second needs a trivial rebase. Triggered the CI as it never ran.

ET-TOUNANI:
Thank u @filipelautert

filipelautert:
#7862 is merged, this one needs a rebase now. The conflict should be trivial, mostly imports in the test file.

ET-TOUNANI:
fixed @filipelautert

filipelautert:
Rebase looks good, ran DiffToChangeLogTest here - 9/9 green on the branch and the new test fails on main, so it does guard #1792.

rberezen:
Reviewed and verified end-to-end — this is a correct, well-scoped fix with a genuine regression test. LGTM.

**✅ Confirmed the bug and the fix.** Since this touches shared diff-ordering logic, I checked it against the pre-fix parent commit rather than just trusting the unit test. The new test's scenario reproduces the bug deterministically:
```
pre-fix base:  first = AddForeignKeyConstraintChange,  second = AddColumnChange   ❌ (FK before the column it references)
this PR:        first = AddColumnChange,               second = AddForeignKeyConstraintChange   ✅
```
So the unit test genuinely **fails on the base and passes here** — it's real regression coverage, not green-on-both.

**✅ No collateral reordering.** My main concern with changing `sortObjects` was that it reorders *every* diff, not just the FK-column case. I ran a rich `diffChangeLog` (4 tables, 2 FK chains, an index, a unique constraint) on both this branch and its base — the generated changelogs are **byte-identical** apart from the intended fix. The merge preserves the dependency order among sortable objects and leaves everything else in place.

**The approach is sound.** `mergeSortedObjectsPreservingUnsortablePositions` uses an `IdentityHashMap`-backed set (correct — avoids `equals()` collisions between distinct objects) and consumes the sorted iterator exactly once per sortable slot, so there's no over/underflow. And it's safe by construction for the target case: a *missing* `Column` implies its table already exists (a missing table would emit `createTable` with its columns inline, not a separate `addColumn`), so there's no risk of the preserved column position landing ahead of a `createTable`.

**One small note** (not blocking): the fix keeps non-sortable objects at their *original* positions, so correctness for the FK-column case relies on the snapshot yielding the column ahead of its dependent FK — which is the normal ordering (and what the SQL Server case in #1792 produces). Worth a one-line comment in `mergeSortedObjectsPreservingUnsortablePositions` making that assumption explicit, so a future snapshot-order change doesn't quietly reintroduce the issue.

Nice, tightly-scoped fix. Thanks @ET-TOUNANI!

filipelautert (changes requested):
@ET-TOUNANI the postgres integration tests fail on this now, and they pass on the same base (da65e0e4d).
testRerunDiffChangeLog: the generated changelog runs ALTER TABLE public.sample_enum_table ADD int_col before the table is created.
testDependenciesInGenerateChangeLog: the primary key is no longer inlined in createTable, it comes out as a separate AddPrimaryKeyChange.
Both look like mergeSortedObjectsPreservingUnsortablePositions undoing the dependency order for the objects it keeps in place. Could you take another look?

ET-TOUNANI:
Good catch @filipelautert  you were right that `mergeSortedObjectsPreservingUnsortablePositions` was undoing the dependency order. 

**Root cause:** missingObjects is built type-by-type (Tables, then Columns, then PKs, then FKs), so an object's original index is a type-grouped index, not a dependency-aware one. The original toSort.addAll(toNotSort) was load-bearing sorted objects first, unsortable ones after which is what guaranteed tables get created before the columns and PKs hanging off them. My merge pinned unsortable objects back to their original index and broke that.

**Fix**: restored the original toSort → toNotSort layout and narrowed the rescue to just the case the ticket was about  columns referenced by a sorted ForeignKey are hoisted ahead of the sorted block, everything else stays after it. With no FKs present it's byte-for-byte the original behavior.

filipelautert (approved):
Ran PostgreSQLIntegrationTest against a real postgres here, 54 tests green - testDependenciesInGenerateChangeLog and testRerunDiffChangeLog included. Narrowing the rescue to the foreign key columns is the right call, thanks for digging out the type-grouped index.

rberezen (approved):

filipelautert:
@ET-TOUNANI heads up, this broke Integration Test (mssql) on main - the fkexists fixture has a table whose FK column is missing together with the table itself, so the hoisted column became an ALTER TABLE before the CREATE TABLE. Fixed in #7898 by skipping the hoist when the column's own table is being created. Your #1792 case still works, I checked it against a real sql server.

ET-TOUNANI:
Thanks @filipelautert ! I think this is worth adding a check to CI so we can catch issues like this earlier.

PR #7803, opened 2026-06-19
fix: default-schema-name not working when runInTransaction is false

### LINKED ISSUE #7791: default-schema-name not working in Postgres for changesets run with runInTransaction:false
### Search first

### Description

After updating to 5.0.3, the `default-schema-name` parameter no longer sets the correct schema for changesets run with `runInTransaction:false`, making them fail to resolve unqualified objects in the non-default schema.

### Steps To Reproduce

Consider this simple changelog file:
```sql
--changeset test:1
CREATE TABLE IF NOT EXISTS test (id TEXT PRIMARY KEY);
--changeset test:2 runInTransaction:false
CREATE INDEX CONCURRENTLY IF NOT EXISTS idx_test_id ON test (id);
```
and this simplified configuration via the Liquibase Gradle plugin:
```kotlin
activities.register(name) {
    arguments = mapOf(
        "logLevel" to "info",
        "changelogFile" to "/path/to/changelog.sql",
        /* ↓ this is the important part; 
         * assumes `myschema` exists in the DB and it is where I want to create the `test` table 
         * and where Liquibase tables are located */
        "defaultSchemaName" to "myschema",
        "url" to dbUrl,
        "username" to dbUser,
        "password" to dbPassword,
    )
}
```

Observe the `update` task pass on `5.0.2` but fail with
```
[2026-06-09 15:21:14] SEVERE [liquibase.integration] Migration failed for changeset db/query-engine.changelog-master.sql::2::test:
     Reason: liquibase.exception.DatabaseException: ERROR: relation "test" does not exist [Failed SQL: (0) CREATE INDEX CONCURRENTLY IF NOT EXISTS idx_test_id ON test (id)]
```
on `5.0.3`

### Expected/Desired Behavior

I expect this setup to continue working unmodified.

### Liquibase Version

5.0.3

### Database Vendor & Version

PostgreSQL 17.5

### Liquibase Integration

Gradle plugin

### Liquibase Extensions

_No response_

MacOS 26

### Additional Context

I have verified that the issue is not with the Gradle plugin nor our setup. I used debug logging to make sure the `defaultSchemaName` parameter is set to the expected value and it seems to be correctly translated to the `--default-schema-name` CLI argument by the Gradle plugin.

I am fairly confident this is happening due to [this change](https://github.com/liquibase/liquibase/pull/7488). Apparently, the `SEARCH_PATH` is now set using `SET LOCAL` in Postgres which only makes the config option valid for a single _transaction_, not the entire session. [Postgres' docs](https://www.postgresql.org/docs/17/sql-set.html) explain why it doesn't work for out-of-transaction changesets specifically:

> `LOCAL`
>
> Specifies that the command takes effect for only the current transaction. After COMMIT or ROLLBACK, the session-level setting takes effect again. **Issuing this outside of a transaction block emits a warning and otherwise has no effect.**

(emphasis mine).

Note that while this this seems to have been the intention of that PR, I still consider it a bug. [The docs](https://docs.liquibase.com/community/reference-guide-5-0-3/parameters/default-schema-name) do not mention anything about the setting only being in effect for changesets that run in a transaction, nor there seems to be any alternative configuration option that would allow working around this limitation. Also, this clearly appears to be a regression as the same changelog used to work in prior versions, but stopped working after an update with no mention in the release notes about any breaking change of this sort.

Lastly, introducing such a breaking change in a patch version update is surely not intended 🙂

felhag:
## Impact

## Description
Only apply `LOCAL` to `SEARCH_PATH` when not running in transaction. Fixes default-schema-name when changelog not running in transaction 

## Release note
Fixed defaultSchemaName not applying to runInTransaction=false changesets

## Things to be aware of
First time committer to Liquibase; not sure about the `isInTransaction` and the fallback to `true`.

## Additional Context
Introduced by #7488
Closes #7791

### DIFF
--- a/liquibase-integration-tests/src/test/groovy/liquibase/command/SearchPathIntegrationTest.groovy
+++ b/liquibase-integration-tests/src/test/groovy/liquibase/command/SearchPathIntegrationTest.groovy
@@ -93,6 +93,47 @@ class SearchPathIntegrationTest extends Specification {
         } catch (Exception e) {}
     }
 
+    def "defaultSchemaName works when runInTransaction is false (#7791)"() {
+        given:
+        def customSchema = "concurrent_idx_" + System.currentTimeMillis()
+        def database = postgres.getDatabaseFromFactory()
+        def connection = database.getConnection()
+        connection.createStatement().execute("CREATE SCHEMA IF NOT EXISTS " + customSchema)
+        connection.commit()
+
+        when:
+        def resourceAccessor = new SearchPathResourceAccessor(".,target/test-classes")
+        def scopeSettings = [
+                (Scope.Attr.resourceAccessor.name()): resourceAccessor
+        ]
+        Scope.child(scopeSettings, {
+            CommandScope commandScope = new CommandScope(UpdateCommandStep.COMMAND_NAME)
+            commandScope.addArgumentValue(DbUrlConnectionArgumentsCommandStep.URL_ARG, postgres.getConnectionUrl())
+            commandScope.addArgumentValue(DbUrlConnectionArgumentsCommandStep.USERNAME_ARG, postgres.getUsername())
+            commandScope.addArgumentValue(DbUrlConnectionArgumentsCommandStep.PASSWORD_ARG, postgres.getPassword())
+            commandScope.addArgumentValue(DbUrlConnectionArgumentsCommandStep.DEFAULT_SCHEMA_NAME_ARG, customSchema)
+            commandScope.addArgumentValue(UpdateCommandStep.CHANGELOG_FILE_ARG, "changelogs/pgsql/update/default-schema-name-run-in-transaction-false.xml")
+            commandScope.execute()
+        } as Scope.ScopedRunner)
+
+        then:
+        noExceptionThrown()
+
+        // Verify the concurrent index was actually created in the custom schema
+        def rs = connection.createStatement().executeQuery("""
+            SELECT COUNT(*) FROM pg_indexes
+            WHERE schemaname = '${customSchema}'
+              AND indexname = 'idx_test_id'
+        """)
+        rs.next()
+        rs.getInt(1) == 1
+
+        cleanup:
+        try {
+            connection.createStatement().execute("DROP SCHEMA IF EXISTS " + customSchema + " CASCADE")
+        } catch (Exception e) {}
+    }
+
     def "SET LOCAL search_path reverts after transaction completes"() {
         given:
         def customSchema = "local_test_" + System.currentTimeMillis()
--- a/liquibase-integration-tests/src/test/groovy/liquibase/command/SearchPathIntegrationTest.groovy
+++ b/liquibase-integration-tests/src/test/groovy/liquibase/command/SearchPathIntegrationTest.groovy
@@ -134,6 +134,58 @@ class SearchPathIntegrationTest extends Specification {
         } catch (Exception e) {}
     }
 
+    def "session-level search_path set for runInTransaction=false changeset is restored (#7791)"() {
+        given:
+        def customSchema = "restore_path_" + System.currentTimeMillis()
+        def database = postgres.getDatabaseFromFactory()
+        def connection = database.getConnection()
+        connection.createStatement().execute("CREATE SCHEMA IF NOT EXISTS " + customSchema)
+        connection.commit()
+        database.setDefaultSchemaName(customSchema)
+
+        when:
+        def resourceAccessor = new SearchPathResourceAccessor(".,target/test-classes")
+        def scopeSettings = [
+                (Scope.Attr.resourceAccessor.name()): resourceAccessor
+        ]
+        Scope.child(scopeSettings, {
+            CommandScope commandScope = new CommandScope(UpdateCommandStep.COMMAND_NAME)
+            // pass the database object so the search_path can be inspected on the same
+            // session the update ran on
+            commandScope.addArgumentValue(DbUrlConnectionArgumentsCommandStep.DATABASE_ARG, database)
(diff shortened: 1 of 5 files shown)


filipelautert (approved):

filipelautert:
Good catch, thanks for the fix!

rberezen:
Verified end-to-end against PostgreSQL 17. The fix works and is correctly scoped. 👍

**Reproduced #7791 on `main`** — `defaultSchemaName=myschema` + a `runInTransaction:false` changeset fails exactly as reported:
```
Reason: ERROR: relation "test" does not exist [Failed SQL: CREATE INDEX CONCURRENTLY ... ON test (id)]
```
**On this PR** the run succeeds and both objects land in the right schema (`myschema.test`, `myschema.idx_test_id`).

**The change is correctly scoped** — captured from the PG statement log:
```
SET LOCAL SEARCH_PATH TO myschema, "$user","public"   ← in-transaction changesets (unchanged)
SET SEARCH_PATH TO myschema, "$user","public"         ← the runInTransaction:false changeset (the fix)
```
`SET LOCAL` stays the default, so the PgBouncer transaction-pooling fix from #7488/#7092 is preserved for the common case.

**Concrete behavior change** (only on the `runInTransaction:false` path), verified directly against Postgres:

| | `main` (`SET LOCAL`) | this PR (`SET`) |
|---|---|---|
| statement outside a txn | `WARNING: SET LOCAL can only be used in transaction blocks` → no-op | applies |
| `search_path` afterward | stays `"$user", public` (the bug) | becomes `myschema` |
| persists on the connection? | n/a | **yes, session-level** |

The first two rows are the intended fix. The third is a small issue that should be fixed: because a `runInTransaction:false` statement runs in autocommit, the `SET` can't be transaction-scoped, so `myschema` persists on that physical connection after the changeset. On a PgBouncer transaction-pooled connection that value lingers for the next client to borrow it — the exact scenario #7488 was closing.

Could you please restore the search_path after the non-transactional changeset finishes? You already capture the prior value via `SHOW SEARCH_PATH`, so it's a matter of re-applying it (or `RESET SEARCH_PATH`) once the statement has run. Note it can't go inline in `initializeDatabase` — that would clear the path before the changeset executes — so it needs to hook in at changeset-completion for the `!isInTransaction` case.

**On your `isInTransaction` question:** `!database.isAutoCommit()` is a sound proxy (the log confirms it flips at exactly the right changeset), and defaulting to `SET LOCAL` on `DatabaseException` is the safe choice — it errs toward the pooling-safe behavior. Good instinct.

Nice work, and the integration test is a good addition.

felhag:
@rberezen I think I addressed your request, can you have another look?

rberezen:
Thanks @felhag — the restore approach is exactly right, and I verified it end-to-end on PostgreSQL 17. But the recent merge from `main` pulled in some unrelated damage that needs cleaning up first.

**✅ The fix works** — captured from the PG statement log for a `runInTransaction:false` changeset with `defaultSchemaName=myschema`:
```
SET SEARCH_PATH TO myschema, "$user","public"   ← session-level, for the non-transactional changeset
SET SEARCH_PATH TO "$user","public"             ← restore: myschema removed afterward
```
The lingering session-level search_path is now cleaned up, so the PgBouncer-pooling gap is closed. #7791 itself still passes (both `test` table and `idx_test_id` land in `myschema`). Nice touches: `restoreSearchPath` sits in `execute()`'s `finally` (so it runs even if the changeset fails) and is wired into `rollback()` too; `saveSearchPath` guards against overwrite and restores the *pre-change* path. The new integration test covering the restore is a good addition.

**⚠️ Merge damage to clean up** — the `origin/main` merge (`116a6c809`) resurrected code in `PostgresDatabase.java` that INT-2140 had extracted to `AbstractPostgresDatabase`:
1. Re-declared, **unused** fields `systemTablesAndViews` and `reservedWords` (lines 53/55) — dead code; the live copies are in the parent.
2. A duplicate `setDefaultCatalogName` override (lines 298-303), byte-identical to the parent's and calling `super` — so it now emits the *"Postgres does not support catalogs"* warning **twice**. Verified:
   ```
   --default-catalog-name on Postgres →  this branch: 2 warnings   |   main: 1 warning
   ```
None of these are on `main`, so they'd land as a regression. Dropping the two fields and the duplicate `setDefaultCatalogName` (a clean re-merge/rebase of `PostgresDatabase.java`) should clear it.

With those three removed, this is good to go. Thanks again!

felhag:
@rberezen Good catch, should be fixed now :)

filipelautert (approved):
Please also drop the two unused imports the merge left in PostgresDatabase.java:11-12 (SqlStatement, RawCallStatement), rest is good to go!

rberezen:
Verified the cleanup (`53fe658ac`) on a fresh build:
- `PostgresDatabase` no longer re-declares `systemTablesAndViews`/`reservedWords` or `setDefaultCatalogName` — and `--default-catalog-name` on Postgres now emits the warning **once** again (was twice).
- The restore fix is untouched and still works: `runInTransaction:false` changeset lands in `myschema`, and the log shows the session-level `SET SEARCH_PATH TO myschema,…` followed by the restoring `SET SEARCH_PATH TO "$user","public"`.

LGTM 🎉 Thanks @felhag!

rberezen (approved):

felhag:
Unused imports are also cleaned up.

filipelautert (approved):
Thanks!

rberezen:
Re-checked after `91a0b2bb0` — it's purely the two unused-import removals (`SqlStatement`, `RawCallStatement`) in `PostgresDatabase.java`. `liquibase-standard` compiles clean, and the diff against my last-reviewed commit (`53fe658ac`) is nothing but those two lines — the restore fix (`restoreSearchPath` / `saveSearchPath` / `isInTransaction`) is untouched. LGTM still stands 👍

PR #7804, opened 2026-06-19
[DAT-23022] Unconditional XXE hardening in XMLChangeLogSAXParser

MalloD12:
## Summary

- **`XMLChangeLogSAXParser.java`**: Moves `disallow-doctype-decl` out of the `secureParsing` guard and adds three additional XXE-hardening features (`external-general-entities=false`, `external-parameter-entities=false`, `load-external-dtd=false`) unconditionally. Introduces a private `trySetFactoryFeature` helper to keep each feature attempt independent. `FEATURE_SECURE_PROCESSING` remains `secureParsing`-gated (schema loading restriction, not entity-expansion surface). Intercepts `SAXParseException` caused by DOCTYPE and throws a clear `ChangeLogParseException` explaining that DOCTYPE/DTD is not supported and the block should be removed.
- **`XMLChangeLogSAXParserTest.groovy`**: Updates the `"allows secureParsing=false"` test — previously it asserted that the entity resolver was reached (file-system open of `invalid.txt`); after the fix `disallow-doctype-decl` fires first, so the assertion now checks for the DOCTYPE rejection message. Updates all three DOCTYPE test assertions from the raw Xerces `"disallow"` keyword to the new user-facing `"not supported"` message.

## Why

When `LIQUIBASE_SECURE_PARSING=false`, all four explicit XXE features were previously skipped. `LiquibaseEntityResolver` would return `null` on any entity-resolution miss, causing the SAX parser to fall through to network/filesystem fetching — enabling classic XXE and SSRF (DAT-23022 / Sec-Audit B8).

Changelogs validate against XSD, not DTD; `<!DOCTYPE>` has no legitimate use here. The four features are now applied regardless of `secureParsing` to provide defence-in-depth even when users opt out of secure mode for compatibility reasons.

## Test plan

- [x] `XMLChangeLogSAXParserTest` — 16 tests pass, including the renamed regression test `"DOCTYPE is rejected unconditionally even when secureParsing=false"`
- [x] `LiquibaseEntityResolverTest` — 28 tests pass unchanged (entity resolver behaviour for XSD schema lookups is unaffected)
- [x] Full build with all tests executed successfully

## Related

- Parent epic: DAT-23006 (Security Audit: Credential Handling and Changelog Injection)
- Related chain ticket: DAT-23029 (XXE via Search-Path Plant — depends on B6+B8)

**Related Ticket:** [DAT-23022](https://datical.atlassian.net/browse/DAT-23022)

[DAT-23022]: https://datical.atlassian.net/browse/DAT-23022?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/liquibase-standard/src/main/java/liquibase/parser/core/xml/XMLChangeLogSAXParser.java
+++ b/liquibase-standard/src/main/java/liquibase/parser/core/xml/XMLChangeLogSAXParser.java
@@ -42,19 +42,28 @@ public class XMLChangeLogSAXParser extends AbstractChangeLogParser {
         saxParserFactory = SAXParserFactory.newInstance();
         saxParserFactory.setValidating(GlobalConfiguration.VALIDATE_XML_CHANGELOG_FILES.getCurrentValue());
         saxParserFactory.setNamespaceAware(true);
+        // Unconditional XXE hardening — Liquibase changelogs validate against XSD, not DTD,
+        // so DOCTYPE has no legitimate use. Applied regardless of secureParsing to provide
+        // defense-in-depth even when users opt out of secure mode.
+        trySetFactoryFeature("http://apache.org/xml/features/disallow-doctype-decl", true);
+        trySetFactoryFeature("http://xml.org/sax/features/external-general-entities", false);
+        trySetFactoryFeature("http://xml.org/sax/features/external-parameter-entities", false);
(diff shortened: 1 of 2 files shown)


v-petrovych (approved):
LGTM!

filipelautert (approved):

PR #7808, opened 2026-06-24
fix: emit valid timestamp with time zone syntax

### LINKED ISSUE #7675: Diff generates a timestamp without considering the time zone
### Search first

### Description

On creating a diff between two databases, when the reference database has `timestamp with time zone`, and the target one has `varchar`, the generated commands has `timestamp` without the time zone.

### Steps To Reproduce

- In the reference database:
  - Oracle:         create table telegram (message_time timestamp with time zone);
  - PostgreSQL: create table telegram (message_time timestamp with time zone);
- In the target database:
  - Oracle:         create table telegram (message_time varchar(200));
  - PostgreSQL: create table telegram (message_time varchar);
- Run the `diff` command, or run the test cases in this [repository](https://github.com/asashour/liquibase-timezone-bug).

### Expected/Desired Behavior

The `newDataType` should be `timestamp with timezone`, not `timestamp`

### Liquibase Version

5.0.2

### Database Vendor & Version

PostgreSQL 18.3, Oracle 19c

### Liquibase Integration

API (DiffChangelogCommandStep)

### Liquibase Extensions

_No response_

Windows 11

### Additional Context

This was previously opened in #6551, with a PR for only PostgreSQL.

Test cases for both using `DiffChangelogCommandStep` and `DataTypeFactory` are provided in the [sample repository](https://github.com/asashour/liquibase-timezone-bug).

The root cause is in `DataTypeFactory` which is being called [here](https://github.com/liquibase/liquibase/blob/400330de52a00c94e50d395d9f236ccde533f898/liquibase-standard/src/main/java/liquibase/diff/output/changelog/core/ChangedColumnChangeGenerator.java#L219).

marchof:
## Impact

## Description

With PR #6929 the datatype "TIMESTAMP WITH TIMEZONE" is not captured any more. The reason is two-fold:

1. Oracle reports the datatype as "TIMESTAMP(6) WITH TIME ZONE(13)" with an additional data size at the end.
2. The above change was actually checking whether the type contains the string "WITH TIMEZONE", but the result stored to local variable `isTimezoneAware` was not used.

This patch makes use of the variable `isTimezoneAware` and adds a test case for the Oracle column type "TIMESTAMP(6) WITH TIME ZONE(13)".

### DIFF
--- a/liquibase-standard/src/main/java/liquibase/datatype/core/TimestampType.java
+++ b/liquibase-standard/src/main/java/liquibase/datatype/core/TimestampType.java
@@ -155,21 +155,15 @@ public class TimestampType extends DateTimeType {
             type = new DatabaseDataType("TIMESTAMP");
         }
 
-        if ((originalDefinition.toUpperCase().startsWith("JAVA.SQL.TYPES.TIMESTAMP_WITH_TIMEZONE")
+        if (isTimezoneAware
             && (database instanceof AbstractPostgresDatabase
             || database instanceof OracleDatabase
             || database instanceof H2Database
             || database instanceof HsqlDatabase
-            || database instanceof SybaseASADatabase)) || isTimezoneAware) {
+            || database instanceof SybaseASADatabase)) {
 
-            if (database instanceof AbstractPostgresDatabase
-                    || database instanceof H2Database
-                    || database instanceof SybaseASADatabase
-                    || database instanceof OracleDatabase) {
-                type.addAdditionalInformation("WITH TIME ZONE");
-            } else {
-                type.addAdditionalInformation("WITH TIMEZONE");
-            }
+            type.addAdditionalInformation("WITH TIME ZONE");
+
             return type;
         }
 
--- a/liquibase-standard/src/main/java/liquibase/datatype/core/TimestampType.java
+++ b/liquibase-standard/src/main/java/liquibase/datatype/core/TimestampType.java
@@ -160,7 +160,7 @@ public class TimestampType extends DateTimeType {
             || database instanceof OracleDatabase
             || database instanceof H2Database
             || database instanceof HsqlDatabase
-            || database instanceof SybaseASADatabase)) || (originalDefinition.toLowerCase().startsWith("timestamptz"))) {
+            || database instanceof SybaseASADatabase)) || isTimezoneAware) {
 
             if (database instanceof AbstractPostgresDatabase
                     || database instanceof H2Database
--- a/liquibase-standard/src/test/groovy/liquibase/datatype/DataTypeFactoryTest.groovy
+++ b/liquibase-standard/src/test/groovy/liquibase/datatype/DataTypeFactoryTest.groovy
@@ -304,6 +304,8 @@ class DataTypeFactoryTest extends Specification {
         "timestamptz"                                  | new DerbyDatabase()    | "TIMESTAMP"                                    | TimestampType | false
         "timestamp"                                    | new HsqlDatabase()     | "TIMESTAMP"                                    | TimestampType | false
         "timestamptz"                                  | new HsqlDatabase()     | "TIMESTAMP WITH TIME ZONE"                     | TimestampType | false
+        "TIMESTAMP WITH TIMEZONE"                      | new HsqlDatabase()     | "TIMESTAMP WITH TIME ZONE"                     | TimestampType | false
+        "timestamptz"                                  | new InformixDatabase() | "DATETIME YEAR TO FRACTION(5)"                 | TimestampType | false
     }
 
     @Unroll("#featureName: #object for #database")
--- a/liquibase-standard/src/test/groovy/liquibase/datatype/DataTypeFactoryTest.groovy
+++ b/liquibase-standard/src/test/groovy/liquibase/datatype/DataTypeFactoryTest.groovy
@@ -299,6 +299,11 @@ class DataTypeFactoryTest extends Specification {
         "uuid"                                         | new MariaDBDatabase()  | "UUID"                                         | UUIDType      | false
         "timestamp"                                    | new FirebirdDatabase() | "TIMESTAMP"                                    | TimestampType | false
         "timestamp(6)"                                 | new FirebirdDatabase() | "TIMESTAMP"                                    | TimestampType | false
+        "timestamptz"                                  | new FirebirdDatabase() | "TIMESTAMP"                                    | TimestampType | false
+        "timestamp"                                    | new DerbyDatabase()    | "TIMESTAMP"                                    | TimestampType | false
+        "timestamptz"                                  | new DerbyDatabase()    | "TIMESTAMP"                                    | TimestampType | false
+        "timestamp"                                    | new HsqlDatabase()     | "TIMESTAMP"                                    | TimestampType | false
+        "timestamptz"                                  | new HsqlDatabase()     | "TIMESTAMP WITH TIME ZONE"                     | TimestampType | false
     }
 
     @Unroll("#featureName: #object for #database")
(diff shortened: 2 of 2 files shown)


filipelautert (commented):
Hello @marchof !
Ran DataTypeFactoryTest on this branch locally, all green including the new (13) case. Could you keep the old TIMESTAMP WITH TIMEZONE rows in the test alongside the new ones? That spelling still comes from changelogs, I re-added them locally and both pass together. Triggered the CI as it never ran.

marchof:
@filipelautert Thanks for picking this up! I restored the old tests.

filipelautert (approved):
Thanks!

rberezen:
Nice catch on the unused `isTimezoneAware` variable — the Oracle fix is correct and I confirmed it works (the new `TIMESTAMP(6) WITH TIME ZONE(13)` → `TIMESTAMP(6) WITH TIME ZONE` case and friends all pass, 289 `DataTypeFactoryTest` cases green). But the change isn't Oracle-scoped, and it regresses several other databases. Requesting changes.

**The problem:** the old guard was `startsWith("timestamptz")`; replacing it with the un-gated `|| isTimezoneAware` makes this block intercept `TIMESTAMP WITH TIME ZONE` for **every** database. But the inner spelling branch only emits the correct `WITH TIME ZONE` for Postgres/H2/SybaseASA/Oracle — everything else falls to the `else` and gets `WITH TIMEZONE` (no space). Databases that previously reached this input through a *different* code path now get the wrong output:

| DB | input | `main` | this PR |
|---|---|---|---|
| HSQLDB | `TIMESTAMP WITH TIME ZONE` | `... WITH TIME ZONE` ✅ | `... WITH TIMEZONE` ❌ |
| Informix | `TIMESTAMP WITH TIME ZONE` | `... WITH TIME ZONE` ✅ | `... WITH TIMEZONE` ❌ |
| Derby | `TIMESTAMP WITH TIME ZONE` | `TIMESTAMP` (tz dropped) | `... WITH TIMEZONE` |
| Firebird | `TIMESTAMP WITH TIME ZONE` | `TIMESTAMP` (tz dropped) | `... WITH TIMEZONE` |

The HSQLDB one is a concrete regression — I verified against a real embedded HSQLDB:
```
OK   : CREATE TABLE a (c TIMESTAMP WITH TIME ZONE)
FAIL : CREATE TABLE b (c TIMESTAMP WITH TIMEZONE)  -> unexpected token : TIMEZONE required: TIME
```
So a `TIMESTAMP WITH TIME ZONE` column that generated valid DDL on `main` now generates DDL that HSQLDB rejects. HSQLDB is a common target for validating changelogs, so this is reachable. Informix is the same no-space problem; Derby doesn't support `TIME ZONE` at all, so it now emits an unsupported clause where `main` safely dropped it.

The existing tests don't catch this because there are no HSQLDB/Firebird/Derby/Informix timezone cases in `DataTypeFactoryTest` — worth adding some as part of the fix.

**Suggested direction:** the condition change is the right idea; it's the inner spelling/gating that needs to keep up. Two options:
- Add `HsqlDatabase` (and any other SQL-standard `WITH TIME ZONE` database) to the inner `addAdditionalInformation("WITH TIME ZONE")` list, and don't add a timezone clause for databases that don't support it (Derby). Or,
- Restrict the new `isTimezoneAware` entry to the databases that actually support timezone-aware timestamps with the correct spelling, so non-supporting DBs keep `main`'s safe fallthrough.

Either way, please add regression rows for HSQLDB / Derby / Firebird / Informix so the cross-DB behavior is pinned. The Oracle fix + the `pg/h2/sybase/oracle` path are all good as-is. Thanks @marchof!

marchof:
@rberezen Nice catch! Sure I will add those tests.

Also I did some research: All the databases which have timezone support allows the syntax `TIMESTAMP WITH TIME ZONE`. So I wonder why we should ever emit `TIMESTAMP WITH TIMEZONE`?

Also how should we handle the case for Databases like Derby which do not support the timezone datatype at all?

filipelautert:
@marchof no case for WITH TIMEZONE that I can find. HSQLDB is the only database that reached that branch through the DB list, and it rejects the syntax:

```
CREATE TABLE t (c TIMESTAMP WITH TIMEZONE)
unexpected token : TIMEZONE required: TIME
```

Ran your branch against main and it fixes more than Oracle. On main the timestamptz check had no database gate, so hsqldb, derby, firebird and informix all got TIMESTAMP WITH TIMEZONE:

```
hsqldb    timestamptz   main: TIMESTAMP WITH TIMEZONE   yours: TIMESTAMP WITH TIME ZONE
derby     timestamptz   main: TIMESTAMP WITH TIMEZONE   yours: TIMESTAMP
firebird  timestamptz   main: TIMESTAMP WITH TIMEZONE   yours: TIMESTAMP
informix  timestamptz   main: TIMESTAMP WITH TIMEZONE   yours: DATETIME YEAR TO FRACTION(5)
```

Derby rejects every timezone form, even TIMESTAMP(6), so dropping it is the only thing that works there.

Could you add the informix row Ruslan asked for, and one for hsqldb with TIMESTAMP WITH TIMEZONE as input? Also worth widening the title, this is not Oracle only anymore and it goes into the release notes that way.

filipelautert (approved):
@marchof pushed the two rows onto your branch, hope you don't mind. informix and hsqldb with the written WITH TIMEZONE spelling, both fail without your change.

Still worth widening the title before this goes in, it reads as Oracle only.

filipelautert:
@marchof this does not reach #7675. The diff path goes through DataTypeFactory.from(referenceType, db).toString() (ChangedColumnChangeGenerator:219), which never calls toDatabaseDataType. Same output on both branches:

```
oracle   TIMESTAMP(6) WITH TIME ZONE(13) -> timestamp(6)
postgres timestamptz                     -> timestamp
```

The DDL fix is good and wider than the title says, hsqldb, derby, firebird and informix stop emitting WITH TIMEZONE too. I'm editing the title and description here so we can get it merged, #7675 stays open for the from() side.

rberezen (commented):
Re-scope looks right — thanks for splitting this cleanly, @filipelautert / @marchof. Reviewing it now purely as the DDL-syntax fix:

**✅ Correct and well-tested.** Since #6929 `isTimezoneAware` was computed but never used, so `toDatabaseDataType` only matched the `startsWith("timestamptz")`/`java.sql...` prefixes and dropped the tz for Oracle's real `TIMESTAMP(6) WITH TIME ZONE(13)`. Wiring `isTimezoneAware` in fixes that. Verified: `DataTypeFactoryTest` passes, and reverting `TimestampType.java` fails the new cases (`TIMESTAMP(6) WITH TIME ZONE(13)` → `TIMESTAMP(6)` on `main`).

**✅ Also a real correctness win beyond the title.** Collapsing to a single `WITH TIME ZONE` and gating on the tz-supporting DB list means Hsql/Derby/Firebird/Informix stop emitting the no-space `WITH TIMEZONE` (which HSQLDB, as you noted, outright rejects). The expanded test matrix pins all of these.

**✅ Scope is now honest** — dropping the `Fixes #7675` link is correct: the diff path (`ChangedColumnChangeGenerator:219` → `DataTypeFactory.from(referenceType, db).toString()`) never calls `toDatabaseDataType`, so #7675 rightly stays open for the `from()` side.

One minor note, non-blocking: for Firebird 4.0+ (which does support `TIMESTAMP WITH TIME ZONE`) this now emits plain `TIMESTAMP`, losing the tz — still better than the invalid `WITH TIMEZONE` it produced before, just flagging the gap for a future follow-up.

LGTM.

rberezen (approved):

marchof:
@filipelautert @rberezen Thanks for re-checking and finalizing this PR!

filipelautert:
@marchof  no, thank you!

PR #7811, opened 2026-06-26
[DAT-23022] Address review comments on XXE hardening in XMLChangeLogSAXParser

MalloD12:
## Summary

Follow-up to #7804, addressing review comments on the XXE hardening changes:

- Updated `trySetFactoryFeature` helper to return a `boolean` indicating whether the feature was set successfully
- Saved the return value of the `disallow-doctype-decl` feature call and throw `UnexpectedLiquibaseException` (instead of `IllegalStateException`) if DOCTYPE blocking could not be applied — consistent with Liquibase's runtime error conventions
- Used `errMsg.toUpperCase(Locale.ROOT)` for case-insensitive DOCTYPE detection in SAX parse error messages

## Test plan

- [ ] Existing unit/integration tests pass
- [ ] Verify `UnexpectedLiquibaseException` is thrown when DOCTYPE blocking is unavailable (e.g. non-standard SAX parser)
- [ ] Verify changelogs with `<!DOCTYPE>` still produce a clear parse error

### DIFF
--- a/liquibase-standard/src/main/java/liquibase/parser/core/xml/XMLChangeLogSAXParser.java
+++ b/liquibase-standard/src/main/java/liquibase/parser/core/xml/XMLChangeLogSAXParser.java
@@ -4,6 +4,7 @@ import liquibase.GlobalConfiguration;
 import liquibase.Scope;
 import liquibase.changelog.ChangeLogParameters;
 import liquibase.exception.ChangeLogParseException;
+import liquibase.exception.UnexpectedLiquibaseException;
 import liquibase.parser.core.ParsedNode;
 import liquibase.parser.core.ParserSupportedFileExtension;
 import liquibase.resource.Resource;
@@ -20,6 +21,7 @@ import javax.xml.parsers.SAXParserFactory;
 import java.io.BufferedReader;
 import java.io.IOException;
 import java.io.InputStream;
+import java.util.Locale;
 import java.util.regex.Pattern;
 
 public class XMLChangeLogSAXParser extends AbstractChangeLogParser {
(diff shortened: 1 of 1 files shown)


filipelautert (approved):
Hey, reviewed the changes and this is a clean follow-up — the important part is right. Making `disallow-doctype-decl` a hard-fail in the constructor means the parser can't be constructed in a DOCTYPE-permitting state, so we fail closed on the one control that actually neutralizes XXE (no DOCTYPE → no DTD → no external entities). Leaving the other three features best-effort is the correct call since they're redundant once DOCTYPE is blocked, and `UnexpectedLiquibaseException` is the right exception type here.

A couple of non-blocking observations, take or leave:

- `XMLChangeLogSAXParser.java:161` — the DOCTYPE detection is a substring match on the parser's exception message (`errMsg.toUpperCase(Locale.ROOT).contains("DOCTYPE")`). That's fine because it only controls whether the user gets the friendly "remove the DOCTYPE block" message vs. the generic parse error — either way we still throw `ChangeLogParseException` and fail safely. Just flagging it's wording-dependent, so if a JAXP impl phrases it differently the nice message silently degrades to the generic one. Not worth engineering around given the safe fallback.
- `XMLChangeLogSAXParser.java:66` — the `catch (Throwable)` can swallow `Error`s; it's pre-existing and here it actually errs toward fail-closed for the doctype path (any failure → `false` → constructor throws). Narrowing to `ParserConfigurationException | SAXNotRecognizedException | SAXNotSupportedException` would be tidier if you're in there anyway, but it doesn't change the security behavior.
- Optional: a small unit test asserting a changelog containing `<!DOCTYPE>` produces the friendly `ChangeLogParseException` would lock in the behavior from the test plan — ignore if #7804 already covers it.

LGTM — approving.

PR #7812, opened 2026-06-26
fix(snapshot): close owning Statement in ResultSetCache.extract to prevent JDBC Statement leak

### LINKED ISSUE #7780: ResultSetCache.extract(ResultSet) leaks the DatabaseMetaData Statement (closes only the ResultSet)
### Search first

Looks like the same root cause as the downstream report quarkusio/quarkus#44042 ("jdbc leak after running liquibase … 18 ResultSet(s) and 0 Statement(s)"). I couldn't find an issue filed against Liquibase itself for the actual cause, hence this report.

### Environment

- **Liquibase version:** 4.33.0 — but the offending code is unchanged on `master` and in the latest release **v5.0.3**.
- **Database / driver:** MariaDB 11.4 via MySQL Connector/J 9.2.0 (the leak is driver-independent; see below).
- **Operating mode:** `update` / migration (the snapshot taken for preconditions / `changeSetExecuted`).

### Description

`liquibase.snapshot.ResultSetCache.extract(ResultSet, boolean)` closes **only the `ResultSet`**, never the `Statement` that produced it:

```java
// liquibase-standard/.../snapshot/ResultSetCache.java  (master, also 4.33.0)
protected List<CachedRow> extract(ResultSet resultSet, final boolean informixIndexTrimHint) throws SQLException {
    ...
    } finally {
        JdbcUtil.closeResultSet(resultSet);   // <-- ResultSet only; the owning Statement is leaked
    }
}
```

Several `JdbcDatabaseSnapshot` paths feed this method a `ResultSet` obtained **directly from a `DatabaseMetaData` call**, passed inline with no reference kept to its `Statement`:

- `extract(databaseMetaData.getColumns(...))` — `JdbcDatabaseSnapshot.java:501`
- `extract(databaseMetaData.getImportedKeys(...))` — `:877`
- `extract(databaseMetaData.getTables(...))` — `:1243`, `:1263`, `:1352`, `:1407`, `:1423`
- `extract(databaseMetaData.getPrimaryKeys(...))`

Per the JDBC contract, **closing a `ResultSet` does not close its `Statement`**. Many drivers (e.g. MySQL Connector/J) implement `DatabaseMetaData.getXxx()` by creating an internal `Statement` and registering it with the connection. After `extract()` returns, that `Statement` stays registered/open until the connection is closed.

The sibling method `executeAndExtract` (same file, ~line 255) handles this correctly:

```java
} finally {
    JdbcUtil.close(resultSet, statement);   // closes BOTH
}
```

### Actual behavior

After a snapshot, the connection carries one or more open driver `Statement`s (the metadata statements). Connection pools that track or validate open statements on return therefore report a leak — for example, Agroal's leak detector fires, or any pool that asserts "no open statements on return" emits a recurring "statements are not closed" warning once per migration — even though every `ResultSet` was closed.

### Expected behavior

`extract(ResultSet, boolean)` should close the owning `Statement` as well — e.g. `JdbcUtil.close(resultSet, resultSet.getStatement())` — mirroring `executeAndExtract`, so no driver statement is leaked.

### Steps to reproduce

1. Run `liquibase update` against a MySQL/MariaDB database using MySQL Connector/J.
2. Wrap the JDBC connection in a pool that asserts no statements remain open on return (Agroal leak detection, or any pool tracking `Statement`s).
3. Observe leaked `Statement`(s) reported after the snapshot/precondition phase, even though all `ResultSet`s are closed.

### Suggested fix

In `ResultSetCache.extract(ResultSet, boolean)`, capture and close the statement:

```java
} finally {
    JdbcUtil.close(resultSet, resultSet.getStatement());
}
```

(`ResultSet.getStatement()` may return `null` for some metadata result sets; `JdbcUtil.close` already null-checks.)

seonwooj0810:
Fixes #7780

## Root cause
\`ResultSetCache.ResultSetExtractor#extract(ResultSet)\` is frequently fed a \`ResultSet\` obtained directly from a \`DatabaseMetaData\` call (\`getColumns\`/\`getTables\`/\`getImportedKeys\`/\`getPrimaryKeys\`, e.g. \`JdbcDatabaseSnapshot.java:501,877,1243,...\`), so the caller keeps no reference to the \`Statement\` that produced the result set. Per the JDBC contract, closing a \`ResultSet\` does **not** close its \`Statement\`, and many drivers (e.g. MySQL Connector/J) implement \`DatabaseMetaData.getXxx()\` by creating an internal \`Statement\` that is registered with the connection. The \`finally\` block closed only the \`ResultSet\`, so that owning \`Statement\` leaked until the connection was closed. This surfaces downstream as JDBC statement leaks (see quarkusio/quarkus#44042: "18 ResultSet(s) and 0 Statement(s)").

## Change
In \`extract(ResultSet, boolean)\`, capture the owning \`Statement\` via \`ResultSet.getStatement()\` (while the result set is still open) and close it alongside the \`ResultSet\` using the existing \`JdbcUtil.close(ResultSet, Statement)\` helper — mirroring how the sibling \`executeAndExtract\` method already closes both. \`getStatement()\` may return \`null\` for metadata result sets per the JDBC spec, which \`JdbcUtil.close\` already null-checks; a defensive \`catch (SQLException)\` falls back to closing the \`ResultSet\` only for drivers that do not support \`getStatement()\` here.

## Test evidence
Added a regression test \`ResultSetCacheTest#extractClosesOwningStatement\` that drives \`extract\` with a mocked \`ResultSet\` whose \`getStatement()\` returns a mock \`Statement\`, and verifies both are closed. It **fails before** this change (\`statement.close()\` is never invoked) and **passes after**.

Verification done: built and ran the new test against `liquibase-standard` with `./mvnw -pl liquibase-standard -am test -Dtest=ResultSetCacheTest` — 1 passed with the fix, 1 failed with the fix reverted (confirming it is a true regression test).

### DIFF
--- a/liquibase-standard/src/main/java/liquibase/snapshot/ResultSetCache.java
+++ b/liquibase-standard/src/main/java/liquibase/snapshot/ResultSetCache.java
@@ -16,6 +16,7 @@ import liquibase.util.StringUtil;
 import java.sql.PreparedStatement;
 import java.sql.ResultSet;
 import java.sql.SQLException;
+import java.sql.Statement;
 import java.util.*;
 import java.util.concurrent.ConcurrentHashMap;
 
@@ -317,7 +318,19 @@ public class ResultSetCache {
                     returnList.add(new CachedRow(row));
                 }
             } finally {
-                JdbcUtil.closeResultSet(resultSet);
+                // This ResultSet is often obtained directly from a DatabaseMetaData call
+                // (e.g. getColumns/getTables/getImportedKeys), so the caller keeps no reference
+                // to the Statement that produced it. Closing the ResultSet alone leaks that
+                // Statement until the connection is closed (see #7780), so close its owning
+                // Statement too. getStatement() may legitimately return null for such ResultSets
+                // per the JDBC spec, which JdbcUtil.close handles.
+                Statement statement = null;
+                try {
+                    statement = resultSet.getStatement();
+                } catch (SQLException ignored) {
+                    // Some drivers may not support getStatement() here; fall back to closing the ResultSet only.
+                }
+                JdbcUtil.close(resultSet, statement);
             }
             return returnList;
         }
--- a/liquibase-standard/src/test/java/liquibase/snapshot/ResultSetCacheTest.java
+++ b/liquibase-standard/src/test/java/liquibase/snapshot/ResultSetCacheTest.java
@@ -1,32 +1,64 @@
 package liquibase.snapshot;
 
+import liquibase.database.Database;
+import org.junit.jupiter.api.Test;
+
+import java.sql.ResultSet;
+import java.sql.Statement;
+import java.util.List;
+
+import static org.mockito.Mockito.mock;
+import static org.mockito.Mockito.verify;
+import static org.mockito.Mockito.when;
+
 public class ResultSetCacheTest {
 
-//    @Test
-//    public void permutations() {
-//        assertEquals(4, new ResultSetCache().permutations(new String[]{"a", "b"}).length);
-//        assertEquals(8, new ResultSetCache().permutations(new String[]{"a", "b", "c"}).length);
-//        assertEquals(16, new ResultSetCache().permutations(new String[]{"a", "b", "c", "d"}).length);
-//        assertEquals(32, new ResultSetCache().permutations(new String[]{"a", "b", "c", "d", "e"}).length);
-//
-//        assertThat(Arrays.asList(new ResultSetCache().permutations(new String[]{"a"})), containsInAnyOrder(new String[] {"a"}, new String[] {null}));
-//
-//        assertThat(Arrays.asList(new ResultSetCache().permutations(new String[]{"a", "b"})), containsInAnyOrder(
-//                new String[]{"a", "b"},
-//                new String[]{null, "b"},
-//                new String[]{"a", null},
-//                new String[]{null, null}
-//        ));
-//
-//        assertThat(Arrays.asList(new ResultSetCache().permutations(new String[]{"a", "b", "c"})), containsInAnyOrder(
-//                new String[]{"a", "b", "c"},
-//                new String[]{"a", "b", null},
-//                new String[]{"a", null, "c"},
-//                new String[]{null, "b", "c"},
-//                new String[]{null, null, "c"},
-//                new String[]{"a", null, null},
-//                new String[]{null, "b", null},
-//                new String[]{null, null, null}
-//        ));
-//    }
+    /**
+     * Regression test for <a href="https://github.com/liquibase/liquibase/issues/7780">#7780</a>:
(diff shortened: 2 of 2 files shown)


seonwooj0810:
Friendly ping — this has been open a couple of weeks with the CLA signed and CI green, but hasn't had a human review yet. @wwillard7800 @filipelautert would one of you have a moment to take a look? It's a small, focused fix (closing the owning `Statement` in `ResultSetCache.extract` to avoid a JDBC statement leak). Thanks!

seonwooj0810:
Could a maintainer approve the workflow run for this PR when you get a chance?

The `authorize` job is the only failing check — it timed out after the 720h approval window, so `Build & Package`, `Integration Test` and the rest were all skipped and never actually ran against the change. The PR itself is small: `ResultSetCache.extract(...)` obtains a `ResultSet` from a `Statement` it creates but never closes the owning `Statement`, so snapshot operations leak JDBC statements; the change closes it in a `finally`.

Happy to rebase if that helps get a clean run.

filipelautert (approved):
Hello @seonwooj0810 !
Good fix, it mirrors what executeAndExtract already does. Checked all the extract() call sites and they all pass DatabaseMetaData result sets, so nothing reuses the closed statement. Triggered the CI as it never ran.

rberezen:
Reviewed and verified — correct, well-scoped resource-leak fix. LGTM.

**✅ Root cause and fix are right.** Metadata result sets from `DatabaseMetaData.getXxx()` carry an owning `Statement` the caller never sees, and closing the `ResultSet` doesn't close it (JDBC contract) — so it leaked until connection close. Capturing `resultSet.getStatement()` *before* closing and handing both to the null-safe `JdbcUtil.close` is exactly right, and mirrors what the sibling `executeAndExtract` already does.

**✅ The regression test is genuine** — I ran it against the pre-fix base and it fails with `Wanted but not invoked: statement.close()`; passes on this branch. (Not green-on-both.)

**✅ No functional regression on the real issue driver.** My main concern was the driver-dependent case where a driver reuses one internal `Statement` across successive `getXxx()` calls — closing it in `extract` could then break *later* metadata reads and silently truncate a snapshot. I checked on MySQL Connector/J 9.0 (the driver from the linked Quarkus report): `generate-changelog` over a rich schema (tables, FK, multiple indexes, unique constraint, view) produces **byte-identical output on this PR vs base** — so nothing was dropped; Connector/J creates a per-call statement and closing it is safe.

**✅ The `executeAndExtract` path is safe.** When `extract` is reached from there, it now closes `resultSet.getStatement()` (the same `PreparedStatement`) and `executeAndExtract`'s own `finally` closes it again — a harmless idempotent double-close, no use-after-close since the rows are already materialized into the returned list. The `catch (SQLException)` fallback around `getStatement()` is a sensible guard for drivers that don't support it there.

Minimal change, real bug (#7780). Thanks @seonwooj0810!

rberezen (approved):

PR #7814, opened 2026-06-26
INT-2140: extract AbstractPostgresDatabase (behavior-neutral)

KushnirykOleh:
## What

Phase 1 of the Postgres-family inheritance refactor ([INT-2139](https://datical.atlassian.net/browse/INT-2139)): introduce `AbstractPostgresDatabase` between `AbstractJdbcDatabase` and `PostgresDatabase`.

`AbstractPostgresDatabase` holds only what is truly common to the Postgres wire family — reserved words, identifier case-folding/quoting, the catalog-less schema model, system-object recognition, and the lowercased changelog table names. `PostgresDatabase` keeps everything real-Postgres-specific (sequences, serial/identity, tablespaces, history, version/product identity, driver, port, search-path rollback).

## Behavior-neutral

No variant is reparented (`PostgresDatabase` still extends, nothing else moves), so the runtime hierarchy is unchanged. The retained `PostgresDatabase` members are the candidates for the explicit-capability work in the next story.

## Tests

Existing suite green locally — `PostgresDatabaseTest` (32), `CockroachDatabaseTest` (22), `EnterpriseDBDatabaseTest` (17), Postgres SQL generators. No priority/generator changes.

[INT-2139]: https://datical.atlassian.net/browse/INT-2139?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/liquibase-standard/src/main/java/liquibase/database/core/AbstractPostgresDatabase.java
+++ b/liquibase-standard/src/main/java/liquibase/database/core/AbstractPostgresDatabase.java
@@ -1,0 +1,190 @@
+package liquibase.database.core;
+
+import liquibase.CatalogAndSchema;
+import liquibase.GlobalConfiguration;
+import liquibase.Scope;
+import liquibase.changelog.column.LiquibaseColumn;
+import liquibase.database.AbstractJdbcDatabase;
+import liquibase.database.ObjectQuotingStrategy;
+import liquibase.statement.SqlStatement;
+import liquibase.statement.core.RawCallStatement;
+import liquibase.structure.DatabaseObject;
+import liquibase.structure.core.Catalog;
+import liquibase.structure.core.Schema;
+import liquibase.structure.core.Table;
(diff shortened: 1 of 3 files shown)


filipelautert (approved):
Hey, reviewed the changes. Clean behavior-neutral extraction — methods moved verbatim, the `isSystemObject` reformat is logically identical, and the inheritance chain (`AbstractJdbcDatabase` → `AbstractPostgresDatabase` → `PostgresDatabase`) keeps every existing override resolving to the same code. No leftover references to the moved fields and no dangling imports. No blocking code issues.

Nice docs on the abstract class — the explicit "what stays in PostgresDatabase and why" list makes the phase-2 boundary obvious.

filipelautert (approved):

PR #7820, opened 2026-06-29
[DAT-23023] audit DATABASECHANGELOG escaping with real DB instances

MalloD12:
## Summary

- Adds escaping correctness tests to cover security audit finding B9: `DATABASECHANGELOG` INSERT values are inlined via `objectToSql` string concatenation rather than JDBC bound parameters
- `CharTypeTest.groovy`: 12 new parametrized `objectToSql` cases verifying single-quote doubling, statement-terminator injection, boolean-bypass injection, and MySQL backslash escaping across H2, PostgreSQL, Oracle, MSSQL, and MySQL
- `MarkChangeSetRanGeneratorEscapingTest.groovy`: new Spock spec with 3 `@Unroll` test methods (single-quote escaping, statement-terminator injection, boolean-bypass injection) parametrized across H2, PostgreSQL, Oracle, MSSQL, MySQL, and MariaDB using real database instances

No production code changes. The audit finding is addressed by confirming and locking in the correctness of the existing `escapeStringForDatabase` chain (`AbstractJdbcDatabase` doubles single quotes; `MySQLDatabase` additionally doubles backslashes).

## Test plan

- [x] `CharTypeTest` — 23 tests pass (8 existing + 12 new `objectToSql` escaping cases)
- [x] `MarkChangeSetRanGeneratorEscapingTest` — 21 tests pass (3 scenarios × 6 real database dialects + 3 unroll count)
- [x] `mvn test -pl liquibase-standard -Dtest="CharTypeTest,MarkChangeSetRanGeneratorEscapingTest"` — BUILD SUCCESS, 0 failures

### DIFF
--- a/liquibase-standard/src/test/groovy/liquibase/datatype/core/CharTypeTest.groovy
+++ b/liquibase-standard/src/test/groovy/liquibase/datatype/core/CharTypeTest.groovy
@@ -41,4 +41,28 @@ class CharTypeTest extends Specification {
         thrown UnexpectedLiquibaseException
 
     }
+
+    @Unroll
+    def "objectToSql escapes values to prevent SQL injection - #desc"() {
+        when:
+        def charType = new CharType()
+
+        then:
+        charType.objectToSql(input, database) == expected
+
+        where:
+        desc                                          | input                                | database               | expected
+        "plain string"                                | "simple"                             | new H2Database()       | "'simple'"
+        "single quote doubled - H2"                   | "O'Brien"                            | new H2Database()       | "'O''Brien'"
+        "single quote doubled - PostgreSQL"           | "O'Brien"                            | new PostgresDatabase() | "'O''Brien'"
+        "single quote doubled - Oracle"               | "O'Brien"                            | new OracleDatabase()   | "'O''Brien'"
+        "single quote doubled - MSSQL"                | "O'Brien"                            | new MSSQLDatabase()    | "'O''Brien'"
+        "single quote doubled - MySQL"                | "O'Brien"                            | new MySQLDatabase()    | "'O''Brien'"
+        "injection: statement terminator - H2"        | "'; DROP TABLE DATABASECHANGELOG--"  | new H2Database()       | "'''; DROP TABLE DATABASECHANGELOG--'"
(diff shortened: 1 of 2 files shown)


filipelautert (commented):
Hey, reviewed the changes — solid test-only PR that does what the ticket scoped it to. I traced every assertion back to `AbstractJdbcDatabase.escapeStringForDatabase` (quote-doubling) and `MySQLDatabase` (backslash-doubling on top), and they all pin the correct, injection-safe output — the "DROP TABLE" and "OR 1=1" cases correctly assert the leading quote is doubled so the payload collapses into an inert string literal. Good coverage of the vectors B9 called out. A few things worth tightening:

1. **`MarkChangeSetRanGeneratorEscapingTest.groovy:32-35,55-56,75` — add negative assertions.** The `then:` blocks only do positive `sql.contains("'it''s-an-id'")` checks. A positive `contains` can still pass if a future regression *also* leaves the raw un-escaped payload somewhere in the statement. For a test whose whole job is to prove escaping, add the mirror check, e.g. `!sql.contains("'; DROP TABLE DATABASECHANGELOG--'")` and `!sql.contains("' OR '1'='1")`. That's the difference between "escaped fragment is present" and "raw payload is absent," and it's what makes the spec actually fail on a partial-escaping regression. This is the one I'd like in before merge.

2. **Title/body says "real DB instances" but these are offline generator tests.** `MarkChangeSetRanGeneratorEscapingTest` instantiates dialect objects (`new H2Database()` etc.) and runs `MarkChangeSetRanGenerator.generateSql(...)` through a mock chain — no connection, no INSERT executed against a running engine. Perfectly valid as a unit test of the emitted SQL, but "real DB instances" overpromises (it proves we *emit* escaped SQL, not that a live server treats it as inert). Suggest softening to "across DB dialects."

3. **`CharTypeTest` exercises `CharType`, but the CHANGELOG columns are `VarcharType`.** Not a bug — `VarcharType extends CharType` and doesn't override `objectToSql`, so coverage is equivalent. Worth a one-line comment noting that, or instantiate `VarcharType` directly so the test visibly matches the type the columns use.

4. **B9 tracking (non-code).** These tests characterize current behavior; they don't implement the bound-parameters option. Suggest resolving DAT-23023 as "mitigated by tests" and filing a follow-up for the parameterization work, so B9 isn't silently treated as closed at the epic level.

None of these block merge; (1) is the one I'd genuinely like in before it lands since it directly strengthens the security guarantee the PR exists to provide.

PR #7826, opened 2026-07-02
INT-2141: add Postgres-family capability methods (behavior-neutral)

KushnirykOleh:
## Impact
Phase 2 of the Postgres-family inheritance refactor (INT-2139). Introduces explicit **capability methods** on `AbstractPostgresDatabase`, replacing the fragile *inherit-by-default, override-to-disable* pattern that has repeatedly let Postgres features leak into non-substitutable variants (Redshift, CockroachDB, EnterpriseDB, YugabyteDB).

## Changes
New methods on `AbstractPostgresDatabase`, each defaulting to the conservative `false` on the family SPI:
- `supportsEnumTypes()`
- `supportsCompositeTypes()`
- `supportsCheckConstraintSnapshot()` — snapshot-ability, **not** enforcement
- `supportsStoredLogicSnapshot()`

`PostgresDatabase` opts in (`true`); variants opt down where they lack the feature — CockroachDB (composite types, stored-logic), EnterpriseDB (composite types).

## Behavior-neutral
No variant is reparented yet, so every subclass inherits its current values. Consumers (the Pro snapshot generators) are converted separately once this lands. Unit tests added for Postgres / Cockroach / EDB capability values.

## Related
- Epic: INT-2139
- Builds on: INT-2140 (AbstractPostgresDatabase extraction)

### DIFF
--- a/liquibase-standard/src/main/java/liquibase/database/core/EnterpriseDBDatabase.java
+++ b/liquibase-standard/src/main/java/liquibase/database/core/EnterpriseDBDatabase.java
@@ -77,11 +77,13 @@ public class EnterpriseDBDatabase extends PostgresDatabase {
     }
 
     /**
-     * EnterpriseDB does not support composite types for snapshotting; opts down from
-     * {@link PostgresDatabase}.
+     * EnterpriseDB (Advanced Server) accepts {@code CREATE TYPE ... AS (...)}, but the standard
+     * composite-type snapshot generator does not read composite types on it, so it opts down from
+     * {@link PostgresDatabase}. This preserves today's behavior, where the generator explicitly
+     * excludes EnterpriseDB.
      */
     @Override
-    public boolean supportsCompositeTypes() {
+    public boolean supportsCompositeTypeSnapshot() {
         return false;
     }
 }
--- a/liquibase-standard/src/main/java/liquibase/database/core/EnterpriseDBDatabase.java
+++ b/liquibase-standard/src/main/java/liquibase/database/core/EnterpriseDBDatabase.java
@@ -75,4 +75,13 @@ public class EnterpriseDBDatabase extends PostgresDatabase {
     public boolean supportsCreateIfNotExists(Class<? extends DatabaseObject> type) {
         return false;
     }
+
+    /**
+     * EnterpriseDB does not support composite types for snapshotting; opts down from
+     * {@link PostgresDatabase}.
+     */
+    @Override
+    public boolean supportsCompositeTypes() {
+        return false;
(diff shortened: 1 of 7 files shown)


liquibase-standard/src/main/java/liquibase/database/core/EnterpriseDBDatabase.java
```diff
@@ -64,4 +64,13 @@ public Integer getDefaultPort() {
     public boolean supportsCreateIfNotExists(Class<? extends DatabaseObject> type) {
         return false;
     }
+
+    /**
+     * EnterpriseDB does not support composite types for snapshotting; opts down from
+     * {@link PostgresDatabase}.
+     */
+    @Override
+    public boolean supportsCompositeTypes() {
```
filipelautert: **Is `false` factually right here?** EDB Advanced Server is Postgres-based and does support `CREATE TYPE ... AS (...)`. If the real constraint is "the standard snapshot generator can't read them on EDB," the javadoc ("does not support composite types") is misleading.

This matters because when the Pro consumers switch from `instanceof` checks to these methods (phase 3), any capability value that doesn't match today's actual behavior becomes a silent snapshot regression. Worth double-checking each opt-down (EDB composite, Cockroach composite + stored-logic) against what those DBs snapshot today.

filipelautert (changes requested):
Hey Oleh, reviewed INT-2141. Clean, behavior-neutral as advertised — nothing calls these methods yet, so the suite staying green is guaranteed. Two things (the blocking one is inline on `EnterpriseDBDatabase`):

1. **Naming is inconsistent across the four methods.** `supportsCheckConstraintSnapshot`/`supportsStoredLogicSnapshot` carry the `Snapshot` suffix, but `supportsEnumTypes`/`supportsCompositeTypes` don't — yet their javadoc says "created **and snapshotted**." Since the consumers are snapshot generators, consider `supportsEnumTypeSnapshot`/`supportsCompositeTypeSnapshot`, or drop the suffix everywhere. Cheap to rename now, painful after Pro depends on them.

2. Minor javadoc slip in `CockroachDatabase.supportsCompositeTypes()`: "no composite types (no `pg_collation`)" — `pg_collation` is the collation catalog, not composite types. Probably meant `pg_type`/`typrelid`. Since these comments are meant to be the source of truth for the capability matrix, worth fixing.

Tests match the matrix, and the base-class design comment naming the exact anti-pattern it replaces (`getShortName().equalsIgnoreCase("redshift")`) is genuinely helpful.

filipelautert (approved):

Issue #2547, opened 2022-02-16
outputDefaultSchema=false and outputDefaultCatalog=false do not suppress schema on updateSQL in recent releases


am0jackson:
In releases 4.3.2, 4.4.0 through 4.7.1, the flags outputDefaultSchema=false and outputDefaultCatalog=false have no effect on updateSQL. Neither in the properties nor in the CLI.

The latest release where the flags work with updateSQL is version 4.3.5. Anyway it doesn't work in version 4.3.2 for example.

**changes.xml**
```
--liquibase formatted sql

--changeset test:1
CREATE SEQUENCE dept_seq START WITH 1;

--changeset test:2
CREATE TABLE JUNITTEST.T_0000_DEPARTMENTS (
  ID           NUMBER(10)    DEFAULT dept_seq.nextval NOT NULL,
  DESCRIPTION  VARCHAR2(50)  NOT NULL);

--changeset test:3
ALTER TABLE JUNITTEST.T_0000_DEPARTMENTS ADD (
  CONSTRAINT dept_pk PRIMARY KEY (ID));
```

**Properties**

```
changeLogFile:/dbchangelog.xml
username:test
password:testpwd
url:jdbc:oracle:thin:@localhost:1521/UNITTST
classpath:ojdbc8.jar
logLevel:debug
outputDefaultSchema=false
outputDefaultCatalog=false
liquibase.hub.mode=off
```

**Terminal**
```
liquibase --outputDefaultSchema=false --outputDefaultCatalog=false --outputFile=test.sql updateSQL
```

**Result in test.sql**
```
-- *********************************************************************
-- Update Database Script
-- *********************************************************************
-- Change Log: /dbchangelog.xml
-- Ran at: 16.02.22 15:55
-- Against: JUNITTEST@jdbc:oracle:thin:@localhost:1521/UNITTST
-- Liquibase version: 4.7.1
-- *********************************************************************

-- Lock Database
UPDATE JUNITTEST.DATABASECHANGELOGLOCK SET LOCKED = 1, LOCKEDBY = 'win10 (172.26.16.1)', LOCKGRANTED = SYSTIMESTAMP WHERE ID = 1 AND LOCKED = 0;

-- Release Database Lock
UPDATE JUNITTEST.DATABASECHANGELOGLOCK SET LOCKED = 0, LOCKEDBY = NULL, LOCKGRANTED = NULL WHERE ID = 1;

-- Lock Database
[...]

PR #7831, opened 2026-07-08
Fix --output-default-schema/--output-default-catalog being ignored by update-sql and other SQL output commands

### LINKED ISSUE #6944: outputDefaultSchema=false has no effect
### Search first

### Description

When  using  --output-default-schema=false , the schema is always included in the update-sql output (Oracle database)

I also tried :
- outputDefaultSchema: false in the liquibase.properties file, 
- liquibase.command.outputDefaultSchema: false in the liquibase.properties file, 
- LIQUIBASE_COMMAND_OUTPUT_DEFAULT_SCHEMA = false environment variable

all with the same effect.

### Steps To Reproduce

Example:

liquibase --output-default-schema=false update-sql

output:
```
Starting Liquibase at 09:16:05 using Java 21.0.5 (version 4.31.1 #6739 built at 2025-02-13 13:46+0000)
Liquibase Version: 4.31.1
Liquibase Open Source 4.31.1 by Liquibase

...

-- Changeset tables/xxxxxxx.yaml::xxxxxxx::xxxxxxx
CREATE TABLE ORAUSER.TABLENAME(CODE_USR VARCHAR2(10) NOT NULL, CODE_OTP VARCHAR2(24) NOT NULL, TYPE_OTP VARCHAR2(8) NOT NULL, DATE_VALIDITE NUMBER(38, 0) NOT NULL, ESSAIS NUMBER(5) NOT NULL, CONSTRAINT PK_T_USER_OTP PRIMARY KEY (CODE_USR)) TABLESPACE DAT_STW_01;

COMMENT ON TABLE ORAUSER.TABLENAME IS 'table comment...';
COMMENT ON COLUMN ORAUSER.T_USER_OTP.CODE_USR IS 'Code utilisateur';
COMMENT ON COLUMN ORAUSER.T_USER_OTP.CODE_OTP IS 'Code OTP';
COMMENT ON COLUMN ORAUSER.T_USER_OTP.TYPE_OTP IS 'Mode d''envoi du code (mail)';
COMMENT ON COLUMN ORAUSER.T_USER_OTP.DATE_VALIDITE IS 'date de validit├® du code';
COMMENT ON COLUMN ORAUSER.T_USER_OTP.ESSAIS IS 'Nombre d''essais de connexion';

INSERT INTO ORAUSER.DATABASECHANGELOG (ID, AUTHOR, FILENAME, DATEEXECUTED, ORDEREXECUTED, MD5SUM, DESCRIPTION, COMMENTS, EXECTYPE, CONTEXTS, LABELS, LIQUIBASE, DEPLOYMENT_ID) VALUES ('xxxx', 'xxxx', 'tables/xxxx.yaml', SYSTIMESTAMP, 238, '9:f2f2fe7af5b0422d730e0253b199ce22', 'createTable tableName=xxxxx', '', 'EXECUTED', NULL, NULL, '4.31.1', '7120563911');

```

### Expected/Desired Behavior

From what the doc says, I would expect the output without the "ORAUSER." prefix before every object.

### Liquibase Version

4.31.1

### Database Vendor & Version

Oracle

### Liquibase Integration

CLI

### Liquibase Extensions

_No response_

Windows 11

### Additional Context

_No response_

yosimasu:
## Impact

## Description

Fixes #6944 (also previously reported as #2547).

`updateSql` — and every other SQL-output command that declares these arguments (`rollbackSql`, `rollbackCountSql`, `rollbackToDateSql`, `changelogSyncSql`, `futureRollbackSql`, `futureRollbackCountSql`, `futureRollbackFromTagSql`, `updateCountSql`, `updateToTagSql`) — exposes `--output-default-schema` / `--output-default-catalog` arguments, documented as "If false, only objects outside the default schema are fully qualified". Setting either to `false` has no effect: objects in the default schema are always fully qualified (e.g. `CREATE TABLE PUBLIC.example_table` on H2, `CREATE TABLE ORAUSER.TABLENAME` on Oracle).

**Root cause:** the arguments are declared but never read. The `Database` object used to generate SQL is created by `AbstractDatabaseConnectionCommandStep#createDatabaseObject`, which unconditionally sets:

```java
database.setOutputDefaultCatalog(true);
database.setOutputDefaultSchema(true);
```

Nothing in the command pipeline ever applies the declared argument values afterwards. (The downstream qualification logic in `AbstractJdbcDatabase#escapeObjectName` already checks `isDefaultSchema(...) && !getOutputDefaultSchema()` correctly, so only the wiring is missing.)

**Fix:** after `DbUrlConnectionCommandStep#obtainDatabase` creates the `Database`, look up whether the currently running command declares an `outputDefaultSchema` / `outputDefaultCatalog` argument, and if so apply its configured value to the new `Database` instance.

- Commands that don't declare these arguments (the majority) are untouched — the database keeps the existing hardcoded `true`/`true`.
- Both arguments default to `true`, so commands that declare them but where the user doesn't pass a value also keep today's behavior.
- When the caller supplies a pre-built `Database` via `DATABASE_ARG` (API usage), nothing is overridden.
- `diff`/`diffChangelog`/`generateChangelog` are unaffected: they don't declare these argument names and keep using their own `DiffOutputControl`/`includeSchema` handling.

**Repro:** run `liquibase updateSql --output-default-schema=false` with a table-creating changelog against any database whose connection has a default schema; the generated `CREATE TABLE` is schema-qualified even though the doc says it shouldn't be. Verified on Oracle (per #6944, maintainer also reproduced on H2), and independently on PostgreSQL 5.0.x via CLI flag and `LIQUIBASE_COMMAND_UPDATE_SQL_OUTPUT_DEFAULT_SCHEMA=false`.

An integration test (`UpdateCommandsIntegrationTest`, H2) is included that runs `updateSql` with `outputDefaultSchema` set to `true` and `false` and asserts the generated `CREATE TABLE` qualification.

## Release note

Fixed `--output-default-schema` / `--output-default-catalog` being ignored by `update-sql`, `rollback-sql`, `changelog-sync-sql`, `future-rollback-sql`, and the other SQL-output commands. Setting them to `false` now omits the default schema/catalog prefix from generated SQL as documented.

## Things to be aware of

- The fix is placed in `DbUrlConnectionCommandStep#obtainDatabase` (the shared connection step) rather than in each of the ten command steps, and applies by argument name lookup so any command declaring `outputDefaultSchema`/`outputDefaultCatalog` — including extensions — gets the wiring for free.
- Only the database-creation path is affected; a `Database` passed in through `DATABASE_ARG` is never modified.

## Things to worry about

- If an extension declares an argument named `outputDefaultSchema`/`outputDefaultCatalog` of type `Boolean` but intends different semantics, it would now also be applied to the database. This seems desirable rather than harmful, but flagging it.

## Additional Context

Issue #6944 was closed by the stale-issue triage after 30 days of inactivity, but was reproduced and confirmed by @tati-qalified, who asked for a community PR. Please consider reopening it alongside this PR.

### DIFF
--- a/liquibase-integration-tests/src/test/groovy/liquibase/command/UpdateCommandsIntegrationTest.groovy
+++ b/liquibase-integration-tests/src/test/groovy/liquibase/command/UpdateCommandsIntegrationTest.groovy
@@ -145,4 +145,36 @@ class UpdateCommandsIntegrationTest extends Specification {
         cleanup:
         outputFile.delete()
     }
+
+    @Unroll
+    def "run UpdateSql with outputDefaultSchema=#outputDefaultSchema qualifies objects in the default schema accordingly"() {
+        given:
+        def outputStream = new ByteArrayOutputStream()
+        def resourceAccessor = new SearchPathResourceAccessor(".,target/test-classes")
+        def scopeSettings = [
+                (Scope.Attr.resourceAccessor.name()): resourceAccessor
+        ]
+
+        when:
+        Scope.child(scopeSettings, {
+            CommandScope commandScope = new CommandScope(UpdateSqlCommandStep.COMMAND_NAME)
+            commandScope.addArgumentValue(DbUrlConnectionArgumentsCommandStep.URL_ARG, h2.getConnectionUrl())
+            commandScope.addArgumentValue(DbUrlConnectionArgumentsCommandStep.USERNAME_ARG, h2.getUsername())
+            commandScope.addArgumentValue(DbUrlConnectionArgumentsCommandStep.PASSWORD_ARG, h2.getPassword())
+            commandScope.addArgumentValue(UpdateSqlCommandStep.CHANGELOG_FILE_ARG, "liquibase/update-tests.yml")
+            commandScope.addArgumentValue(UpdateSqlCommandStep.OUTPUT_DEFAULT_SCHEMA_ARG, outputDefaultSchema)
+            commandScope.addArgumentValue(UpdateSqlCommandStep.OUTPUT_DEFAULT_CATALOG_ARG, outputDefaultSchema)
+            commandScope.setOutput(outputStream)
+
+            commandScope.execute()
+        } as Scope.ScopedRunner)
+
+        then:
+        def output = outputStream.toString().toLowerCase()
+        output.contains("create table public.example_table") == outputDefaultSchema
+        output.contains("create table example_table") == !outputDefaultSchema
+
+        where:
+        outputDefaultSchema << [true, false]
+    }
 }
--- a/liquibase-standard/src/main/java/liquibase/command/core/helpers/DbUrlConnectionCommandStep.java
+++ b/liquibase-standard/src/main/java/liquibase/command/core/helpers/DbUrlConnectionCommandStep.java
@@ -21,6 +21,7 @@ import java.util.HashMap;
 import java.util.List;
 import java.util.Map;
 import java.util.concurrent.atomic.AtomicReference;
+import java.util.function.Consumer;
 
 /**
  * Internal command step to be used on CommandStep pipeline to manage the database connection.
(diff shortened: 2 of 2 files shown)


filipelautert (approved):
Hello @yosimasu !
Nice find, the arguments were declared but never read. The command-scoped wiring and the parametrized IT look right. Triggered the CI as it never ran.

rberezen:
Reviewed and verified end-to-end — correct, well-scoped fix. LGTM.

**✅ Confirmed the bug and the fix** on H2, comparing the PR against its pre-fix parent:
```
update-sql --output-default-schema=false
  pre-fix base:  CREATE TABLE PUBLIC.example_table (id INT);   ❌ flag ignored (the bug)
  this PR:       CREATE TABLE example_table (id INT);          ✅ qualifier dropped
  PR, no flag:   CREATE TABLE PUBLIC.example_table (id INT);   ✅ backward-compatible (default true)
```

**✅ Generic across the SQL-output commands, not just `update-sql`.** I checked a second one — `rollback-count-sql --output-default-schema=false` → `DROP TABLE example_table;` (unqualified). Since the fix lives in the shared `obtainDatabase` and keys off the declared argument name, all the commands listed in the description get it.

**✅ Works via config/env, not just the CLI flag.** `LIQUIBASE_COMMAND_UPDATE_SQL_OUTPUT_DEFAULT_SCHEMA=false` also produces the unqualified output — good, because the fix reads through `getArgumentValue`, which resolves from the value providers rather than only the parsed CLI arg.

**✅ No collateral impact.** The new `applyCommandBooleanArgument` runs in the shared connection step for *every* command, so I confirmed a non-declaring command (plain `update`) still runs cleanly — `getArgument(...)` returns null → no-op, no NPE. And per the description, `diff`/`generateChangelog` don't declare these argument names, so their own `DiffOutputControl`/`includeSchema` handling is untouched. 14 `DbUrlConnectionCommandStepTest` cases pass.

The approach is sound: apply after `createDatabaseObject`'s hardcoded `true`/`true`, only when the command declares the arg, only when the resolved value is non-null (so declared-but-unset commands keep today's behavior), and skip entirely when a pre-built `Database` is supplied via `DATABASE_ARG`.

**One minor note** (not blocking): the wiring couples on the string literals `"outputDefaultSchema"` / `"outputDefaultCatalog"`, so a future rename of those arguments would silently break it with no compile error. Given the shared-step design that's hard to avoid, but a one-line comment cross-referencing the `OUTPUT_DEFAULT_SCHEMA_ARG` declarations (or a shared constant) would guard against it. The included H2 integration test covering both `true`/`false` is a good regression guard.

Nice fix on a long-standing gap (#6944 / #2547). Thanks @yosimasu!

rberezen (approved):

PR #7835, opened 2026-07-11
Fix formatted SQL precondition messages

### LINKED ISSUE #7816: Formatted SQL: onFailMessage is ignored for preconditions in Liquibase 5.0.3
### Search first

### Description

In Liquibase 5.0.3, the `onFailMessage` attribute is not displayed when using preconditions in formatted SQL changelogs.

The precondition `onFail` action (`HALT`, `CONTINUE`, `MARK_RAN`) works correctly, but the custom onFailMessage is ignored. Liquibase always prints the default precondition failure message instead.

For `CONTINUE`,
```log
[INFO] Running Changeset: resources/scripts/config/config_dml.sql::test::HK
[INFO] Continuing past: resources/scripts/config/config_dml.sql::test::HK despite precondition failure due to onFail='CONTINUE': 
          resources/changelog-master.yaml : SQL Precondition failed.  Expected '1' got '0'
```

For `HALT`,
```log
[INFO] Running Changeset: resources/scripts/config/config_dml.sql::test::HK
[ERROR] ChangeSet resources/scripts/config/config_dml.sql::test::HK encountered an exception.
liquibase.exception.MigrationFailedException: Migration failed for changeset resources/scripts/config/config_dml.sql::test::HK:
     Reason: 
          resources/changelog-master.yaml : SQL Precondition failed.  Expected '1' got '0'
```

### Steps To Reproduce

changelog-master.sql
```yaml
databaseChangeLog:
    - include:
        file: scripts/config/config_updates.sql
        relativeToChangelogFile: true
```

config_dml.sql
```sql
--liquibase formatted sql

--changeset HK:test
--preconditions onFail:HALT onFailMessage:"Application 'value' not found. Skipping configuration insertion."
--precondition-sql-check expectedResult:1 SELECT COUNT(*) FROM schema.table WHERE column = 'value';

INSERT STATEMENT...
--rollback DELETE FROM...
```

### Expected Behavior

Liquibase should display the custom onFailMessage:
```text
Application 'value' not found. Skipping configuration insertion.
```

### Actual Behavior

Liquibase ignores onFailMessage and displays:
```text
SQL Precondition failed. Expected '1' got '0'
```

or for CONTINUE:
```text
Continuing past ... due to onFail='CONTINUE':
SQL Precondition failed. Expected '1' got '0'
```

The custom onFailMessage is never emitted.

### Liquibase Version

5.0.3

### Database Vendor & Version

PostgreSQL 18.4

### Liquibase Integration

liquibase-maven-plugin

### Liquibase Extensions

_No response_

Windows 11

### Additional Context

The same behavior occurs with onFail=HALT, CONTINUE, and MARK_RAN. The changeset execution behavior is correct, but the custom failure message is never logged.

This was tested with formatted SQL changelogs.

The changeset is included from changelog-master.yaml but the actual changeset file is formatted SQL.

hutiefang76:
## Impact

## Description

Fixes #7816.

Formatted SQL changelogs recognized the `onFail` and `onError` actions, but did not copy their `onFailMessage` and `onErrorMessage` attributes into the `PreconditionContainer`. As a result, precondition failures always displayed the generated default message.

This change parses message attributes as quote-aware top-level tokens, preserving messages with spaces, apostrophes, and attribute-like text inside the value. It maps both message attributes to the existing container fields.

Regression coverage includes the reported quoted `onFailMessage` form, the symmetric error message, embedded attribute names, and unquoted apostrophes.

## Release note

Formatted SQL preconditions now display custom `onFailMessage` and `onErrorMessage` text.

## Things to be aware of

The parser only treats a quote immediately after an attribute colon as an opening delimiter. Quotes inside an unquoted token remain part of that token.

## Things to worry about

None known.

## Additional Context

Verification with JDK 17:

- `FormattedSqlChangeLogParserTest`: 79 tests, 0 failures
- `liquibase-standard` unit suite: 6,179 tests, 0 failures, 12 skipped
- `git diff --check`

### DIFF
--- a/liquibase-standard/src/main/java/liquibase/parser/core/formattedsql/FormattedSqlChangeLogParser.java
+++ b/liquibase-standard/src/main/java/liquibase/parser/core/formattedsql/FormattedSqlChangeLogParser.java
@@ -105,10 +105,20 @@ public class FormattedSqlChangeLogParser extends AbstractFormattedChangeLogParse
         }
     }
 
+    /**
+     * Parses a string attribute without treating attribute-like text inside quoted values as a new token. A quote
+     * starts a quoted value only when it immediately follows the attribute separator. Unquoted values are returned
+     * as-is, while quoted values retain their surrounding quotes for the caller to remove.
+     *
+     * @param body The preconditions body
+     * @param attributeName The attribute to parse
+     * @return The attribute value, or {@code null} when the attribute is missing or empty
+     */
     private String parseStringAttribute(String body, String attributeName) {
         int tokenStart = -1;
         char quote = 0;
 
+        // The synthetic trailing space flushes the final token when the body has no trailing whitespace.
         for (int i = 0; i <= body.length(); i++) {
             char current = i < body.length() ? body.charAt(i) : ' ';
             if (quote != 0) {
@@ -123,7 +133,11 @@ public class FormattedSqlChangeLogParser extends AbstractFormattedChangeLogParse
                     String token = body.substring(tokenStart, i);
                     int separator = token.indexOf(':');
                     if ((separator > 0) && attributeName.equalsIgnoreCase(token.substring(0, separator))) {
-                        return StringUtil.trimToNull(token.substring(separator + 1));
+                        String value = StringUtil.trimToNull(token.substring(separator + 1));
+                        if (value != null) {
+                            logMatch(attributeName, value, getClass());
+                        }
+                        return value;
                     }
                     tokenStart = -1;
                 }
--- a/liquibase-standard/src/main/java/liquibase/parser/core/formattedsql/FormattedSqlChangeLogParser.java
+++ b/liquibase-standard/src/main/java/liquibase/parser/core/formattedsql/FormattedSqlChangeLogParser.java
@@ -82,7 +82,15 @@ public class FormattedSqlChangeLogParser extends AbstractFormattedChangeLogParse
 
             PreconditionContainer pc = new PreconditionContainer();
             pc.setOnFail(StringUtil.trimToNull(parseString(onFailMatcher, ON_FAIL)));
+            String onFailMessage = parseStringAttribute(body, "onFailMessage");
+            if (onFailMessage != null) {
+                pc.setOnFailMessage(StringUtil.stripEnclosingQuotes(onFailMessage));
+            }
             pc.setOnError(StringUtil.trimToNull(parseString(onErrorMatcher, ON_ERROR)));
+            String onErrorMessage = parseStringAttribute(body, "onErrorMessage");
+            if (onErrorMessage != null) {
+                pc.setOnErrorMessage(StringUtil.stripEnclosingQuotes(onErrorMessage));
+            }
 
             if (onSqlOutputMatcher.matches() && onUpdateSqlMatcher.matches()) {
                 throw new IllegalArgumentException("Please modify the changelog to have preconditions set with either " +
@@ -95,6 +103,35 @@ public class FormattedSqlChangeLogParser extends AbstractFormattedChangeLogParse
             }
             changeSet.setPreconditions(pc);
         }
+    }
+
+    private String parseStringAttribute(String body, String attributeName) {
+        int tokenStart = -1;
+        char quote = 0;
+
+        for (int i = 0; i <= body.length(); i++) {
+            char current = i < body.length() ? body.charAt(i) : ' ';
+            if (quote != 0) {
+                if (current == quote) {
+                    quote = 0;
+                }
+            } else if (((current == '\"') || (current == '\'')) &&
+                    (tokenStart >= 0) && (i > tokenStart) && (body.charAt(i - 1) == ':')) {
+                quote = current;
+            } else if (Character.isWhitespace(current)) {
+                if (tokenStart >= 0) {
+                    String token = body.substring(tokenStart, i);
+                    int separator = token.indexOf(':');
+                    if ((separator > 0) && attributeName.equalsIgnoreCase(token.substring(0, separator))) {
+                        return StringUtil.trimToNull(token.substring(separator + 1));
+                    }
+                    tokenStart = -1;
+                }
+            } else if (tokenStart < 0) {
+                tokenStart = i;
+            }
+        }
+        return null;
     }
 
     @Override
--- a/liquibase-standard/src/main/java/liquibase/parser/core/formattedsql/FormattedSqlChangeLogParser.java
+++ b/liquibase-standard/src/main/java/liquibase/parser/core/formattedsql/FormattedSqlChangeLogParser.java
@@ -12,6 +12,7 @@ import liquibase.precondition.core.SqlPrecondition;
 import liquibase.precondition.core.TableExistsPrecondition;
 import liquibase.precondition.core.ViewExistsPrecondition;
 import liquibase.util.StringUtil;
+import org.apache.commons.lang3.StringUtils;
 
 import java.util.regex.Matcher;
 import java.util.regex.Pattern;
@@ -133,7 +134,7 @@ public class FormattedSqlChangeLogParser extends AbstractFormattedChangeLogParse
                     String token = body.substring(tokenStart, i);
                     int separator = token.indexOf(':');
                     if ((separator > 0) && attributeName.equalsIgnoreCase(token.substring(0, separator))) {
-                        String value = StringUtil.trimToNull(token.substring(separator + 1));
+                        String value = StringUtils.trimToNull(token.substring(separator + 1));
                         if (value != null) {
                             logMatch(attributeName, value, getClass());
                         }
--- a/liquibase-standard/src/test/groovy/liquibase/parser/core/formattedsql/FormattedSqlChangeLogParserTest.groovy
+++ b/liquibase-standard/src/test/groovy/liquibase/parser/core/formattedsql/FormattedSqlChangeLogParserTest.groovy
@@ -802,8 +802,44 @@ create table table1 (
 
         then:
         PreconditionContainer preconditions = changeLog.getChangeSets().first().getPreconditions()
+        preconditions.getOnFail() == PreconditionContainer.FailOption.HALT
         preconditions.getOnFailMessage() == "Application 'value' not found. Skipping insertion."
+        preconditions.getOnError() == PreconditionContainer.ErrorOption.HALT
         preconditions.getOnErrorMessage() == "Unable to verify application."
+    }
+
+    def "leave precondition messages unset when not specified"() throws Exception {
+        given:
+        String changeLogWithoutPreconditionMessages = "--liquibase formatted sql\n" +
+                "--changeset test:test\n" +
+                "--preconditions onFail:HALT onError:HALT\n" +
+                "select 1;\n"
+
+        when:
+        DatabaseChangeLog changeLog = new MockFormattedSqlChangeLogParser(changeLogWithoutPreconditionMessages)
+                .parse("asdf.sql", new ChangeLogParameters(), new JUnitResourceAccessor())
+
+        then:
+        PreconditionContainer preconditions = changeLog.getChangeSets().first().getPreconditions()
+        preconditions.getOnFailMessage() == null
+        preconditions.getOnErrorMessage() == null
+    }
+
+    def "parse precondition message with onSqlOutput"() throws Exception {
+        given:
+        String changeLogWithMessageAndOnSqlOutput = "--liquibase formatted sql\n" +
+                "--changeset test:test\n" +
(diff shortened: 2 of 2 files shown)


wwillard7800 (changes requested):
Thanks for tackling this — the bug in #7816 is a real annoyance for users relying on custom precondition messages, and the regression tests you added exercise the tokenizer's own edge cases nicely. A few things I'd like to see addressed before this merges.

## 1. Reinvents a pattern that already exists in this file family

`AbstractFormattedChangeLogParser.java` already handles "quote-aware, whitespace-terminated, don't-swallow-the-next-attribute" for exactly this kind of value. See `CONTEXT_REGEX`, `CONTEXT_FILTER_REGEX`, `LABELS_REGEX` (lines 167/170/176):

```java
".*context:\\s*(\".*?\"|" + KNOWN_ATTRS_LOOKAHEAD + "\\S*).*"
```

Values are then consumed via `parseString(matcher, ...)` + `StringUtil.stripEnclosingQuotes(...)`. Two lines per attribute, and every other maintainer of this file already understands the idiom.

Could you rewrite along those lines? Something like:

```java
public static final String ON_FAIL_MESSAGE_REGEX  = ".*onFailMessage:\\s*(\".*?\"|"  + KNOWN_ATTRS_LOOKAHEAD + "\\S*).*";
public static final String ON_ERROR_MESSAGE_REGEX = ".*onErrorMessage:\\s*(\".*?\"|" + KNOWN_ATTRS_LOOKAHEAD + "\\S*).*";
```

…then `parseString` + `stripEnclosingQuotes` at the throw site. That removes the entire `parseStringAttribute` helper and stays consistent with how every other attribute on the line is parsed.

## 2. `KNOWN_ATTRS_LOOKAHEAD` needs the two new attribute names

Independent of #1 — `AbstractFormattedChangeLogParser.java:120–123` uses a negative lookahead against a hardcoded list of attribute names so that `context:`/`labels:`/etc. don't gobble into the next `name:` prefix. `onFailMessage` and `onErrorMessage` are new attribute names and aren't in that list.

Concrete failure after this PR merges:

```
--changeset ... contextFilter:qa onFailMessage:"something"
```

`CONTEXT_FILTER_REGEX` captures `\S*` after `contextFilter:` up to the next known attribute name. Because `onFailMessage` isn't listed, `contextFilter` will happily eat into `onFailMessage:`. Please add both names to `KNOWN_ATTRS_LOOKAHEAD` in the same PR.

## 3. Test gaps

Your three tests cover the tokenizer's edge cases well. A few more I'd like to see:

- **Baseline regression**: `onFail:HALT` with no message specified → `getOnFailMessage()` stays null. Nothing pins that down today.
- **Assert enum values too** in the quoted-message test — `getOnFail() == HALT` etc., so a future refactor can't silently corrupt the sibling attributes.
- **`onSqlOutput` / `onUpdateSql` on the same line as a message** — the branch that throws on both being set is nearby; make sure a message + a single onSqlOutput still round-trips.

## 4. Smaller items

- **Log parity**: `parseString` calls `logMatch(description, matchingString, getClass())` at FINE. The new helper doesn't. If you go with the regex approach in #1 this fixes itself.
- **Javadoc** on any helper you keep.
- **Variable expansion**: `onFailMessage` is free-form user-facing text where `${env}`/`${property}` expansion is a reasonable expectation. Sibling precondition types under `handlePreconditionCase` (SQL_CHECK/TABLE_EXISTS/VIEW_EXISTS) all pass the body through `changeLogParameters.expandExpressions(...)`. `handlePreconditionsCase` doesn't. That's pre-existing behavior for `onFail`/`onError` (enum keywords — no expansion needed), but messages are a different beast. Worth calling out and either fixing here or filing a follow-up.
- Minor: `for (int i = 0; i <= body.length(); ...)` with the synthetic sentinel is clever but non-obvious. If you keep the hand-rolled parser, drop a one-line comment explaining the past-end iteration.

Appreciate the fix and the thoroughness on the tests — with the two structural items above (adopt the existing regex idiom + update `KNOWN_ATTRS_LOOKAHEAD`) this should be a clean merge.

hutiefang76:
I checked the current PR base (e58fabc) and the current main branch before changing this. I cannot find KNOWN_ATTRS_LOOKAHEAD in either version, and CONTEXT_REGEX / CONTEXT_FILTER_REGEX / LABELS_REGEX currently use the simpler .*attribute:(".*?"|\\S*).* form. Is there another branch or commit you would like me to rebase onto, or should this PR introduce KNOWN_ATTRS_LOOKAHEAD and update those existing patterns as part of the fix?

I also want to preserve the quoted-token case covered by the current tests. With only onErrorMessage:"No onFailMessage:configured." present, the proposed ON_FAIL_MESSAGE_REGEX can match the inner onFailMessage text even though it is not an attribute. Is there a preferred regex pattern in the project that handles this case, or would you prefer keeping a small quote-aware helper for message attributes?

Once that structural point is clear, I will add the requested baseline, enum, and onSqlOutput/onUpdateSql coexistence coverage. I can also cover expression expansion in this PR, unless you would prefer that as a follow-up.

wwillard7800:
You're right, and I apologize. I reviewed against the wrong tree — I was reading the Liquibase Pro monorepo's `core/` subtree copy of this file, which carries a downstream `KNOWN_ATTRS_LOOKAHEAD` addition that never went upstream. On OSS `main` (which is what your PR targets) the patterns really are the simpler `.*attribute:(\".*?\"|\\S*).*` form you found. **Please disregard items #1 and #2 in my review.**

Concretely, on this PR's actual base:

- Introducing `KNOWN_ATTRS_LOOKAHEAD` as part of this bug fix is out of scope — that's a broader change that should stand on its own PR if we want it in OSS.
- Adopting the naive OSS idiom (`.*onFailMessage:(\".*?\"|\\S*).*`) is actively worse than what you wrote, because — as you correctly point out — the regex would false-positive on `onFailMessage:` occurring inside a quoted `onErrorMessage`. Your quote-aware helper handles that case; the naive regex does not.

**Your current approach is the right one for OSS.** Keep the `parseStringAttribute` helper.

What I'd still like to see (from my original items 3 and 4, which stand):

- **Baseline "no message" regression** — a test that leaves `onFailMessage`/`onErrorMessage` off entirely and asserts both fields stay null.
- **Enum assertions in the quoted-message test** — `getOnFail() == HALT`, `getOnError() == HALT` — so a future refactor can't silently corrupt the sibling attributes.
- **`onSqlOutput` / `onUpdateSql` coexistence** — a message + a single `onSqlOutput` still round-trips.
- **A brief comment on the sentinel iteration** (`i <= body.length()` with the synthetic space) — CodeRabbit flagged the same thing. One line is enough.
- **Log parity** — a `logMatch(...)` call for `onFailMessage`/`onErrorMessage` matches, mirroring what `parseString` does for the surrounding attributes.
- **Javadoc** on `parseStringAttribute`.

On the expression-expansion question: please do that as a follow-up. It's a larger behavioral change (`handlePreconditionsCase` doesn't expand today for the enum attributes) and worth its own ticket + PR so the release note is accurate and easy to review.

Thanks again — and sorry for the wasted round-trip on my end. The hand-rolled helper was the right call.

wwillard7800 (approved):
Redoing my review, this time reading the file at the PR's actual base ref (e58fabc). Apologies again for the earlier round-trip on the wrong tree.

## Verdict

**Approach is sound — keep `parseStringAttribute` as written.** On OSS `main`, the naive regex idiom used for sibling attributes (`CONTEXT_REGEX`, `LABELS_REGEX`: `.*attr:(\".*?\"|\\S*).*`) genuinely can't handle the case where an attribute name appears inside another attribute's quoted value — your "do not parse message attribute names from inside quoted messages" test demonstrates exactly why. The regex engine would grab the inner `onFailMessage:configured.\"` via `\S*` since `\".*?\"` isn't at the current position. Your hand-rolled parser's "quote only opens directly after `:`" rule handles it correctly.

Wiring `pc.setOnFailMessage` / `pc.setOnErrorMessage` from the parsed values mirrors how the XML/YAML/JSON parsers populate the same container fields. The bug and the fix line up cleanly with #7816.

## Nits worth addressing before merge

1. **Baseline null-message regression test.** Add a test that omits both message attributes and asserts `getOnFailMessage()` and `getOnErrorMessage()` both return null. Nothing pins that down today, and a future refactor of the walker could silently start returning an empty string.
2. **Assert the enum values in the quoted-message test.** In *"parse quoted precondition messages"*, also assert `preconditions.getOnFail() == PreconditionContainer.FailOption.HALT` (and same for onError). Right now the test would still pass if a future change accidentally corrupted the sibling values on the same line.
3. **`onSqlOutput` coexistence test.** Add a positive test where a message coexists with a single `onSqlOutput:TEST` (or `onUpdateSql:...`) and both round-trip correctly. Sits naturally next to the existing double-set failure case.
4. **Comment the sentinel iteration.** `for (int i = 0; i <= body.length(); i++)` with the synthetic trailing whitespace is doing real work — it flushes the final token when the last attribute is at end-of-body with no trailing space. One line explaining that keeps the next reader out of my chair. CodeRabbit flagged the same thing.
5. **Javadoc on `parseStringAttribute`.** Short block covering: what it returns for missing / empty / unquoted / quoted tokens, and the quote-opens-only-after-colon invariant.
6. **Log parity.** `parseString` calls `logMatch(description, matchingString, getClass())` at FINE. A matching log line on successful `onFailMessage` / `onErrorMessage` matches would help users debug "why isn't my custom message showing up".

## Style consistency (optional)

The three new tests use bare expressions in `then:` blocks (`preconditions.getOnFailMessage() == \"...\"`). Spock treats these as implicit assertions so they're correct, but the surrounding tests in this file use explicit `assert`. Worth matching the local convention.

## Deferred

Per our earlier exchange, please cover variable expansion (`${env}` / `${property}` inside messages) in a follow-up. Messages are the only free-text values in `handlePreconditionsCase`, so the behavior/scope decision is worth its own ticket.

---

Nice piece of work — the tokenizer's handling of the attribute-name-inside-quoted-value case is a real improvement over what the OSS regex idiom would produce. Approving on the substance; please fold in the nits above before merge.

hutiefang76:
Thanks for the corrected review. I kept the quote-aware helper and addressed the remaining items in 9e2781d:

- added the no-message baseline, enum assertions, and message + `onSqlOutput` coverage
- documented the tokenizer contract and sentinel iteration
- added matching `logMatch` behavior

I left expression expansion for a separate follow-up as requested.

Verification:
- `FormattedSqlChangeLogParserTest`: 81 tests, 0 failures/errors
- full `liquibase-standard` suite: 6,181 tests, 0 failures/errors (12 skipped)
- `git diff --check`

filipelautert (approved):
All six nits are in and the expansion follow-up is already agreed, approving again.

filipelautert (approved):
Thanks!

rberezen:
Reviewed and verified end-to-end — the fix works for both message attributes and the quote-aware parsing is solid. LGTM, with one pre-existing quirk worth flagging.

**✅ Confirmed the bug and the fix** on H2, comparing against the pre-fix base. A precondition that fails / errors:
```
                    base (pre-fix)                              this PR
onFail  (SELECT 0): "SQL Precondition failed. Expected '1'..."  "Application 'value' not found. Skipping insertion."  ✅
onError (bad SQL):  "Table NONEXISTENT_MARKER not found..."     "Unable to verify application."                       ✅
```
Both `onFailMessage` and `onErrorMessage` now surface the custom text; on `main` you get the generated default / raw exception.

**✅ The quote-aware parsing holds up under the tricky cases:**
- Apostrophe inside a double-quoted message (`"Application 'value' not found."`) → preserved verbatim.
- Embedded attribute-like text *with spaces* (`onFailMessage:"missing; see onError:CONTINUE note with spaces"`) → the whole thing is kept as one value and displayed intact; it isn't split at the interior whitespace. Reading the parser, this is right: a quote only "opens" immediately after the attribute `:`, and the first-colon split + exact pre-colon name match means `onFail:HALT` can't be confused with `onFailMessage`.
- 81 `FormattedSqlChangeLogParserTest` cases pass, and the new tests cover the quoted form, symmetric error message, embedded names, and unquoted apostrophes.

**ℹ️ One pre-existing limitation worth a heads-up** (not introduced here, not blocking): the sibling *action* parsers (`ON_FAIL_PATTERN` / `ON_ERROR_PATTERN`) are still plain regexes over the raw body and are *not* quote-aware. So a message that embeds action text — e.g. `onFailMessage:"note onError:CONTINUE inside"` — silently changes the actual `onError` action to `CONTINUE`. I confirmed this behaves identically on `main`, so it's not a regression from this PR. But since this PR makes custom messages first-class (users will now write longer, freeform messages), it's more likely to be hit. Might be worth a follow-up to route the action parsing through the same quote-aware tokenizer you just added, so the message value is excluded from action matching.

Clean, well-tested fix on a real gap (#7816). Thanks @hutiefang76!

rberezen (approved):

PR #7841, opened 2026-07-16
INT-2211: reference-scoped configuration for two-connection commands

SvampX:
## What

  Adds opt-in **reference scoping** to `ConfigurationDefinition` so two-connection commands (`diff`, `diff-changelog`) can resolve a configuration definition differently on the **reference** connection than on the **primary** one.

  Today the primary/reference distinction exists only in the command-step layer and is erased before `connection.open()`, so any auth/connection selector read via `getCurrentValue()` binds *both* connections. This change threads a lightweight signal so a flagged definition can be resolved per-connection.

  ## How
  
  - **`ConfigurationDefinition.Builder.referenceScoped()`** — marks a definition reference-scopable. While the reference connection is being opened, the definition resolves from its `<namespace>.reference.<suffix>` sibling key:
    - sibling **unset** → inherit the primary value (fully backward compatible),
    - sibling **set** → use it,
    - sibling = **`DEFAULT`** (case-insensitive sentinel) → resolve as if unset (→ the definition's default).
    The value is resolved *through the base definition*, so its type conversion, default value, and obfuscator all still apply — secrets stay masked.
  - **`ConfigurationDefinition.IS_REFERENCE_CONNECTION_SCOPE_KEY`** — a `Scope` marker set to `TRUE` only while the reference connection opens. `ReferenceDbUrlConnectionCommandStep` sets it inside the existing `Scope.child(...)` block that wraps
  `createDatabaseObject(...)`.
  - The default+convert tail of `getCurrentConfiguredValue` was extracted into a private `applyDefaultAndConvert(...)` helper (pure refactor), shared by the standard and reference-scoped paths.

  ## Compatibility

  Fully inert unless a definition is explicitly flagged `referenceScoped()`. Unflagged definitions, single-connection commands, and the primary connection of two-connection commands behave exactly as before (`referenceKey == null` short-circuits the new branch).

  ## Tests

  `ConfigurationDefinitionTest` gains cases for: inherit-when-unset, sibling-override, `DEFAULT` opt-out (incl. case-insensitivity), reference-only config, primary-connection-ignores-sibling, and unflagged-definition-is-inert. Full `liquibase.configuration` test package is green.

  ## Context
  
  Enables downstream (commercial) database extensions to offer per-connection auth for `diff`/`diff-changelog` (e.g. target on PKI/OAuth/OIDC, reference on password). Tracked as INT-2211.

### DIFF
--- a/liquibase-standard/src/main/java/liquibase/command/core/helpers/ReferenceDbUrlConnectionCommandStep.java
+++ b/liquibase-standard/src/main/java/liquibase/command/core/helpers/ReferenceDbUrlConnectionCommandStep.java
@@ -3,6 +3,7 @@ package liquibase.command.core.helpers;
 import liquibase.Scope;
 import liquibase.command.*;
 import liquibase.command.providers.ReferenceDatabase;
+import liquibase.configuration.ConfigurationDefinition;
 import liquibase.configuration.ConfigurationValueObfuscator;
 import liquibase.database.Database;
 import liquibase.database.jvm.JdbcConnection;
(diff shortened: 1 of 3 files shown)


HorbatenkoYehor (approved):

filipelautert (approved):
LGTM. The reference-scoping mechanism is well-scoped and correctly inert unless a definition opts in with referenceScoped(). Inherit/override/DEFAULT-sentinel behavior is well covered by the new tests, and the referenceKey != null short-circuit keeps the hot path free for unflagged definitions.

Two non-blocking notes:
- The DEFAULT sentinel is read before type conversion, so this is really built for String-typed selectors. A one-line javadoc note (or a guard) that referenceScoped() is intended for String selectors would help a future caller with an enum/non-String DataType avoid a conversion surprise.
- Reference resolution only checks the canonical key, not aliases — already documented in the javadoc, just flagging for visibility.

Neither blocks merge.

filipelautert (approved):

PR #7845, opened 2026-07-20
INT-2215: don't silently drop same-order value modifiers / table-name generators

SvampX:
## What

  `ConfiguredValueModifierFactory` and `LiquibaseTableNamesFactory` store their plugins in a `Comparator.comparingInt(getOrder)`-keyed `TreeSet`. A comparator-based `TreeSet` treats `compare == 0` as a **duplicate**, so `addAll`/`add` keep only the **first** plugin per
  `getOrder()` value and **silently discard** the rest.

  For value modifiers this is security-relevant and hard to diagnose: two modifiers sharing an order (e.g. anything left at `ProConfiguredValueModifier`'s default `100`, alongside HashiCorp Vault at `100`) collapse to one, so a dropped vault modifier lets a reference
  string reach the database as a **literal value** — with no error. Long-standing defect (the line dates to 2022); it only became observable once multiple modifiers shipped with the same order.

  ## How

  Replace the comparator-keyed `TreeSet` in both factories with an **identity-deduplicated list**, sorted by `getOrder()` ascending and **tie-broken by class name** for deterministic iteration across JVM runs. Same-order plugins now coexist instead of colliding.

  - **`ConfiguredValueModifierFactory`** — mutable backing list (guarded by `this`) published as an immutable, fully-sorted snapshot for lock-free reads. Iteration semantics preserved: `override(ConfiguredValue)` ascending; `override(String)` highest-order-first, stopping
  at the first modifier that changes the value. `register(...)` de-duplicates by object identity (distinct same-order modifiers coexist); `unregister(...)` removes that specific instance.
  - **`LiquibaseTableNamesFactory`** — same idiom fixed: an immutable list sorted by `getOrder()` then class name.

  ## Compatibility

  No behavior change for the existing distinct-order plugins (they keep their relative order; the class-name tie-break only affects previously-colliding same-order plugins, which used to be dropped entirely). No public API changes.

  ## Tests

  New specs:
  - Two modifiers / two table-name generators with the same `getOrder()` **both run** (previously one was silently dropped).
  - `register` de-duplicates by identity; `unregister` removes that instance; same-order modifiers coexist.
  - `override(String)` applies the highest-order modifier first and stops at the first change.

  Regression on the config value-resolution hot path (`LiquibaseConfiguration`, `ConfigurationDefinition`, `ConfiguredValue`, converters) stays green.


HorbatenkoYehor (approved):
lgtm, approved

Minor remark (non-blocking)

1. Fragile override(String) == "x-high" test. Unlike the others, it relies on an "empty SPI classpath" assumption rather than an isolated fake locator. It passes today (confirmed: liquibase-standard has no META-INF/services/liquibase.configuration.ConfiguredValueModifier), but if a ConfiguredValueModifier with order > 200 that touches "x" is ever added to liquibase-standard, this test will break. The recording tests (calls.toSet() == ["A", "B"]) are robust to that, since foreign modifiers don't write to calls. Suggestion: for consistency, switch ConfiguredValueModifierFactoryTest to the same Scope.child + fake ServiceLocator pattern already used successfully in LiquibaseTableNamesFactoryTest.

filipelautert (approved):
LGTM. Clean fix that matches the ticket's option B and hits every AC — same-order plugins now coexist in both factories, deterministic order, register/unregister correct.

A couple of non-blocking nits, fix whenever:
- `ConfiguredValueModifierFactory.java:16` — comment says "Mutable backing set" but it's a List now.
- Both new `.groovy` test files are missing a trailing newline at EOF.

filipelautert (approved):

SvampX:
Thanks @filipelautert — addressed note 1 here in `c46ca2fab`: added a behavior note on `override(String)` clarifying that the "same-order modifiers all coexist and run" guarantee is specific to `override(ConfiguredValue)`, and that this overload returns on the first modifier that changes the value (so only one same-order modifier takes effect on this path). Kept it in OSS since this is the source of truth for the core file.

Note 2 (the unregister-by-identity test) is handled on the Pro side in liquibase/liquibase-pro#4442, where it can assert against the real assembled factory singleton.

HorbatenkoYehor (approved):

PR #7848, opened 2026-07-22
Fix service discovery aborting when a provider fails to load

### LINKED ISSUE #7757: StandardServiceLocator aborts service discovery when Iterator.hasNext() throws LinkageError (e.g. NoClassDefFoundError) on OpenJDK 25
### Search first

### Description

`liquibase.servicelocator.StandardServiceLocator#findInstances` wraps only `iterator.next()` in a try/catch:

```java
  while (iterator.hasNext()) {          // <-- not in try/catch
      try {
          T next = iterator.next();
          ...
      } catch (Throwable e) {
          new ServiceLoadExceptionHandler().handleException(e);
      }
  }
```

The JDK's `ServiceLoader.LazyClassPathLookupIterator.hasNextService()` only catches `ServiceConfigurationError`. A `LinkageError` such as `NoClassDefFoundError` (raised when the JVM tries to resolve a missing class referenced in the provider's constructor signature, annotations, or superclass) escapes `hasNext()` directly and aborts the entire service-discovery loop. The exception handler never runs, and no providers get returned, even ones registered later in the same service file.

### Steps To Reproduce

Use any Liquibase deployment that has `liquibase-hibernate7` on the classpath but does not ship Spring (a perfectly valid configuration, Spring support is optional in the extension). `META-INF/services/liquibase.database.Database` from `liquibase-hibernate7` lists `HibernateSpringBeanDatabase` and `HibernateSpringPackageDatabase`, both of which transitively reference `org.springframework.core.io.support.ResourcePatternResolver`. When `DatabaseFactory.<init>` calls `findInstances(Database.class)`, you get:

```
Caused by: java.lang.NoClassDefFoundError: org/springframework/core/io/support/ResourcePatternResolver
    at java.base/java.lang.Class.getDeclaredConstructors0(Native Method)
    at java.base/java.lang.Class.privateGetDeclaredConstructors(Unknown Source)
    at java.base/java.lang.Class.getConstructor0(Unknown Source)
    at java.base/java.lang.Class.getConstructor(Unknown Source)
    at java.base/java.util.ServiceLoader.getConstructor(Unknown Source)
    at java.base/java.util.ServiceLoader$LazyClassPathLookupIterator.hasNextService(Unknown Source)
    at java.base/java.util.ServiceLoader$LazyClassPathLookupIterator.hasNext(Unknown Source)
    at java.base/java.util.ServiceLoader$1.hasNext(Unknown Source)
    at java.base/java.util.ServiceLoader$2.hasNext(Unknown Source)
    at liquibase.database.DatabaseFactory.<init>(DatabaseFactory.java:39)
    ...
```

Notice the failure point: `StandardServiceLocator.java:24` is the `while (iterator.hasNext())` line. No `Database` providers are returned and bootstrap fails.

### Expected/Desired Behavior

A provider that fails to load (regardless of whether the JDK signals that via `ServiceConfigurationError`, `LinkageError`, or any other `Throwable`) should be logged and skipped, exactly as the existing comment-free intent of the try/catch around next() suggests. Subsequent providers in the same service file should still be discovered.

### Liquibase Version

5.0.2

### Database Vendor & Version

_No response_

### Liquibase Integration

Reproduced on OpenJDK 25.0.3-ea; the bug is JDK-agnostic in principle but is more easily triggered on modern JDKs where getConstructor is invoked during ServiceLoader.hasNextService.

### Liquibase Extensions

_No response_

_No response_

### Additional Context

The JDK iterator's internal pending cursor is advanced before the failing class is resolved, so calling `hasNext()` again after a thrown `LinkageError` correctly moves on to the next provider, meaning the fix is local to `StandardServiceLocator` and does not require a JDK change.

## Suggested fix
  
Move the try/catch (`Throwable`) to surround the entire `hasNext()`/`next()` pair. Pull from the iterator unconditionally until it signals exhaustion, treating any error from either call as a skip-this-provider. The patch below is the implementation we are running in production as a higher-priority replacement; it can
drop straight into `StandardServiceLocator`:

```java
package liquibase.servicelocator;

import java.util.ArrayList;
import java.util.Collections;
import java.util.Iterator;
import java.util.List;
import java.util.ServiceLoader;

import liquibase.Scope;
import liquibase.logging.Logger;

public class StandardServiceLocator implements ServiceLocator {
    @Override
    public int getPriority() {
        return 1;
    }

    @Override
    public <T> List<T> findInstances(Class<T> requiredInterface) {
        List<T> instances = new ArrayList<>();
        Logger log = Scope.getCurrentScope().getLog(getClass());
        ClassLoader classLoader = Scope.getCurrentScope().getClassLoader(true);
        Iterator<T> iterator = ServiceLoader.load(requiredInterface, classLoader).iterator();
        while (true) {
            try {
                if (!iterator.hasNext()) {
                    break;
                }
                T instance = iterator.next();
                log.fine("Loaded " + requiredInterface.getName()
                    + " instance " + instance.getClass().getName());
                instances.add(instance);
            } catch (Throwable e) {
                // LinkageError (e.g. NoClassDefFoundError) can escape JDK ServiceLoader.hasNext()
                // because the JDK only catches ServiceConfigurationError internally.
                new ServiceLoadExceptionHandler().handleException(e);
            }
        }
        return Collections.unmodifiableList(instances);
    }
}
```

The only structural change vs. the current code is that the while-condition is moved inside the try. Behavior is unchanged for the happy path and for `ServiceConfigurationError`-style failures (which the JDK already stashes and rethrows from next()).

## Workaround
  
Until this is fixed upstream, downstream projects can ship their own `ServiceLocator` with priority > 1 and the loop above.

kamilkrzywanski:
## Impact

## Description

Fixes #7757.

`StandardServiceLocator` only caught failures from `iterator.next()`. On newer JDKs, `ServiceLoader` can throw `NoClassDefFoundError` from `hasNext()` when a provider class has a missing dependency (for example liquibase-hibernate Spring classes without Spring on the classpath). That stopped discovery entirely, so later providers never loaded either.

Both `hasNext()` and `next()` are now inside the try/catch. A bad provider is logged and skipped; the rest continue to load.

## Release note

Fixed a bug where Liquibase could fail to start if an extension registered a service that referenced a missing optional dependency.

## Things to be aware of

Same failure handling path as before — we just also cover errors thrown from `hasNext()`.

## Things to worry about

None.

## Additional Context

Added unit tests that build a small ServiceLoader fixture with a broken provider between two good ones and check that discovery continues.

### DIFF
--- a/liquibase-standard/src/main/java/liquibase/servicelocator/StandardServiceLocator.java
+++ b/liquibase-standard/src/main/java/liquibase/servicelocator/StandardServiceLocator.java
@@ -21,8 +21,12 @@ public class StandardServiceLocator implements ServiceLocator {
 
         final Logger log = Scope.getCurrentScope().getLog(getClass());
         final Iterator<T> services = ServiceLoader.load(interfaceType, Scope.getCurrentScope().getClassLoader(true)).iterator();
-        while (services.hasNext()) {
+        // Guard hasNext() too: ServiceLoader may throw LinkageError (e.g. NoClassDefFoundError) from it.
+        while (true) {
             try {
+                if (!services.hasNext()) {
+                    break;
+                }
                 final T service = services.next();
                 log.fine("Loaded "+interfaceType.getName()+" instance "+service.getClass().getName());
                 allInstances.add(service);
--- a/liquibase-standard/src/test/groovy/liquibase/servicelocator/StandardServiceLocatorTest.groovy
+++ b/liquibase-standard/src/test/groovy/liquibase/servicelocator/StandardServiceLocatorTest.groovy
@@ -1,8 +1,8 @@
 package liquibase.servicelocator
 
+import liquibase.Scope
 import liquibase.change.Change
 import liquibase.changelog.ChangeLogHistoryService
-import liquibase.command.CommandStep
 import liquibase.database.Database
 import liquibase.database.DatabaseConnection
 import liquibase.datatype.LiquibaseDataType
@@ -23,9 +23,18 @@ import liquibase.structure.DatabaseObject
 import liquibase.util.TestUtil
 import org.junit.Assume
 import spock.lang.Specification
+import spock.lang.TempDir
 import spock.lang.Unroll
 
+import javax.tools.ToolProvider
+import java.nio.charset.StandardCharsets
+import java.nio.file.Files
+import java.nio.file.Path
+
 class StandardServiceLocatorTest extends Specification {
+
+    @TempDir
+    Path tempDir
 
     @Unroll("#featureName: #type.name")
     def "all classes are listed in service loader files"() {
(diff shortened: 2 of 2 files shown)


filipelautert (approved):
Hello @kamilkrzywanski !
Good catch, and the compiled-fixture test is a nice way to reproduce the hasNext() failure. Ran StandardServiceLocatorTest on this branch locally, all green. Triggered the CI as it never ran.

rberezen:
Reviewed and verified end-to-end — correct fix with excellent regression coverage. LGTM.

**✅ Confirmed the bug and the fix** against the pre-fix base. The test compiles real service fixtures (good → bad → good), deletes `MissingDependency.class` to trigger a genuine `NoClassDefFoundError`, and runs a real `URLClassLoader`+`ServiceLoader`:
```
base:    NoClassDefFoundError propagates out of findInstances → discovery aborts (test errors)
this PR: error caught, both good providers load (size == 2)   → discovery continues ✅
```
The all-providers-bad case correctly returns an empty list rather than throwing.

**✅ The subtle risk — an infinite loop — is ruled out.** `while (true) { try { if (!hasNext()) break; … } catch (Throwable) { log } }` re-enters `hasNext()` after each failure, so if the `ServiceLoader` iterator ever re-threw on the *same* provider without advancing, this would spin forever. I checked empirically rather than by reasoning: the real-`ServiceLoader` test terminates in seconds (no hang), because the JDK iterator consumes the failing provider's name before the class load throws, so the next `hasNext()` advances. Good that the test uses a genuine broken class rather than a mock iterator — it actually exercises this path.

**✅ `catch (Throwable e)` is the right width** — `NoClassDefFoundError`/`LinkageError` are `Error`s, not `Exception`s, so catching `Throwable` (unchanged from before, just now also covering `hasNext()`) is what makes this work. Delegating to the existing `ServiceLoadExceptionHandler` keeps the logging behavior consistent.

22 `StandardServiceLocatorTest` cases pass. Minimal, well-scoped change on a real startup-breaking bug (#7757). Nice work @kamilkrzywanski — especially the self-contained compile-a-broken-provider fixture.

rberezen (approved):

PR #7850, opened 2026-07-23
Detect Firebird-compatible RedDatabase servers as Firebird

makssent:
## Impact

## Description

RedDatabase is an actively maintained Firebird fork (same engine lineage,
SQL-compatible, works with the standard Jaybird JDBC driver), but it reports
its own product name, e.g. `RedDatabase 5.1 (abcdef01...)`.
`FirebirdDatabase.isCorrectDatabaseImplementation()` only matches product
names starting with `Firebird`, so such connections fall through to
`UnsupportedDatabase` and Liquibase generates generic SQL that the server
rejects.

#### Steps To Reproduce

1. Run `liquibase update` (or Spring Boot with Liquibase) against a
   RedDatabase server using the Jaybird driver.
2. Liquibase attempts to create its tracking table.

#### Expected Behavior

The connection is detected as Firebird and Firebird-compatible SQL is
generated.

#### Actual Behavior

The connection is treated as `UnsupportedDatabase` and the update fails:

CREATE TABLE DATABASECHANGELOG failed; SQL error code = -607;
Specified domain or source column DATETIME does not exist

#### Fix

Accept product names starting with `RedDatabase` in addition to `Firebird`
in `FirebirdDatabase.isCorrectDatabaseImplementation()`. This follows the
existing pattern in `SybaseASADatabase`, which matches several product names
of compatible servers in a single class. The short name stays `firebird`,
so existing changelogs with `dbms="firebird"` apply as-is.

Unit tests added to `FirebirdDatabaseTest` covering detection of Firebird
and RedDatabase product names plus negative cases.

## Release note

Liquibase now detects Firebird-compatible RedDatabase servers as Firebird,
so `update` and other commands work against RedDatabase out of the box.

## Things to be aware of

- Detection-only change: two lines in `isCorrectDatabaseImplementation()`;
  no SQL generation logic is touched and behavior for all other databases
  is unchanged.
- Verified against a live RedDatabase 5.1 server: the tracking tables are
  created and updates run successfully.

## Additional Context

RedDatabase: https://reddatabase.ru/en/ - a Firebird-based DBMS; the same
Jaybird driver and Firebird SQL dialect are used, only the reported product
name differs.

### DIFF
--- a/liquibase-standard/src/main/java/liquibase/database/core/FirebirdDatabase.java
+++ b/liquibase-standard/src/main/java/liquibase/database/core/FirebirdDatabase.java
@@ -23,7 +23,8 @@ public class FirebirdDatabase extends AbstractJdbcDatabase {
 
     @Override
     public boolean isCorrectDatabaseImplementation(DatabaseConnection conn) throws DatabaseException {
-        return conn.getDatabaseProductName().startsWith("Firebird");
+        String productName = conn.getDatabaseProductName();
+        return productName.startsWith("Firebird") || productName.startsWith("RedDatabase");
     }
 
     @Override
--- a/liquibase-standard/src/test/groovy/liquibase/database/core/FirebirdDatabaseTest.groovy
+++ b/liquibase-standard/src/test/groovy/liquibase/database/core/FirebirdDatabaseTest.groovy
@@ -1,5 +1,6 @@
 package liquibase.database.core
 
+import liquibase.database.DatabaseConnection
 import liquibase.structure.core.Column
 import spock.lang.Specification
 import spock.lang.Unroll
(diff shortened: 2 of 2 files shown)


filipelautert (approved):

rberezen:
Reviewed and verified — clean, safe detection enhancement. LGTM.

**✅ Confirmed the fix works at the selection level, not just the isolated method.** Beyond `isCorrectDatabaseImplementation` returning true for `RedDatabase`, I checked that the real `DatabaseFactory.findCorrectDatabaseImplementation` actually *picks* Firebird for a RedDatabase-named connection:
```
"Firebird 5.0"                → FirebirdDatabase  ✅
"RedDatabase 5.1 (abcdef01)"  → FirebirdDatabase  ✅ (was UnsupportedDatabase before)
```
So a RedDatabase connection now gets the Firebird SQL-generation path, which is what fixes the `DATABASECHANGELOG` creation failure in the issue.

**✅ The unit test is a genuine regression test** — I ran it against the pre-fix base and the two `RedDatabase` rows fail there (base returns `false`); all 20 pass on this branch.

**✅ No detection collision.** Nothing else in the codebase matches `RedDatabase`, so widening Firebird's match can't steal a connection from another `Database` — and since unmatched connections previously fell through to `UnsupportedDatabase` (the reported bug), there's nothing higher-priority to displace.

The approach mirrors the existing `SybaseASADatabase` multi-product-name pattern, the short name stays `firebird` (so `dbms="firebird"` changelogs are unaffected), and no SQL-generation logic is touched. `productName` is null-safe to the same degree as before (unchanged `.startsWith` behavior).

Couldn't exercise a live RedDatabase server here, but the detection→Firebird-path selection is confirmed and the downstream SQL is the existing, already-tested Firebird path. Nice, minimal enhancement. Thanks @makssent!

rberezen (approved):

PR #7862, opened 2026-07-29
Refactor DiffToChangeLog: honest naming, Edge record, simpler bringDropFKToTop, + tests

KushnirykOleh:
Readability/cleanup pass on `DiffToChangeLog` — **no behavior change**, and it adds unit coverage for two previously-untested methods.

- Rename `determineDatabase` → `liveDatabaseForSerialization` and fix its comment — the old comment ("return the Database if not offline") never mentioned that it also gates on Postgres-family, which is half the logic. (The gate is real: Postgres SQL generation is connection-state-sensitive — SERIAL vs `GENERATED … AS IDENTITY` keys off the live server version, EDB Redwood — so the serializer needs the live connection object; others fall back to a filename-derived `Database`.)
- Convert `Edge` to a `record` — drops a mismatched hand-rolled `equals` (identity `==`) / `hashCode` (value, on `toString()`).
- Replace `bringDropFKToTop`'s two-pass `contains` partition with a stable sort + a named `hasDropForeignKey` predicate. `Stream.sorted()` is stable on ordered streams, so within-group order is preserved.
- Remove a dead `isEmpty()` guard (always true — `trimToEmpty(...) + "."` is never empty) and a redundant `null`-check before an `instanceof`.

Adds `DiffToChangeLogTest` coverage for `liveDatabaseForSerialization` (online Postgres-family → returns the DB; offline / non-Postgres → null) and `bringDropFKToTop` (drop-FK float to top, order preserved; no-drop-FK unchanged). 8/8 green.

### DIFF
--- a/liquibase-standard/src/main/java/liquibase/diff/output/changelog/DiffToChangeLog.java
+++ b/liquibase-standard/src/main/java/liquibase/diff/output/changelog/DiffToChangeLog.java
@@ -166,9 +166,9 @@ public class DiffToChangeLog {
             //
             // Get a Database instance and save it in the scope for later use
             //
-            Database database = determineDatabase(diffResult.getReferenceSnapshot());
+            Database database = liveDatabaseForSerialization(diffResult.getReferenceSnapshot());
             if (database == null) {
-                database = determineDatabase(diffResult.getComparisonSnapshot());
+                database = liveDatabaseForSerialization(diffResult.getComparisonSnapshot());
             }
             ChangelogPrintServiceFactory printServiceFactory = Scope.getCurrentScope().getSingleton(ChangelogPrintServiceFactory.class);
             ChangelogPrintService printService = printServiceFactory.getChangeLogPrintService(this);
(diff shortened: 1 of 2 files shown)


KushnirykOleh:
Re: the `liveDatabaseForSerialization` null-connection finding —

Good catch, but pre-existing and out of scope for this PR. This is a behavior-neutral rename/cleanup, and that `!(connection instanceof OfflineConnection)` condition is **unchanged** from the old `determineDatabase` — the null-connection handling predates it. Tightening it to `connection != null` would *change* behavior (null → filename-derived fallback), which doesn't belong in a no-behavior-change PR.

Practical impact is minimal regardless: a connectionless `PostgresDatabase` here yields the same connectionless-Postgres serialization the `getTargetDatabase` fallback produces (both hit `useSerialDatatypes()`'s no-connection default). If the online-only contract should reject null connections, that's a small separate behavioral fix — happy to file it.

filipelautert (approved):

wwillard7800 (approved):
Approving, but it seems like this PR should be targeted at `liquibase6` instead of main, since the ticket hierarchy is intended for 6.0.0.

PR #7866, opened 2026-07-30
Fix URL loading from driver properties file

### LINKED ISSUE #7860: Cannot configure database URL through driver-properties-file
### Search first

### Description

According to the [documentation](https://docs.liquibase.com/community/reference-guide-5-0-3/parameters/driver-properties-file), one can use `--driver-properties-file` to configure the database connection. `url` is mentioned explicitly as an example:
Image

However, the `url` specified in the driver properties file is ignored, and the CLI still requires a `--url` command line argument, or a URL specified via other means (defaults file, environment variable, etc.).

### Steps To Reproduce

1. Run `liquibase init project`.
2. Open the generated _liquibase.properties_, cut the `liquibase.command.url` property and paste it into a new file _driver.properties_ (the variants `liquibase.command.url` or `liquibase.command.update.url` don't work either).
3. Run `liquibase update --driver-properties-file=driver.properties`

Result:
```
Error parsing command line: Invalid argument '--url': missing required argument. If you need to configure new liquibase project files and arguments, run the 'liquibase init project' command.
```

_liquibase.properties_:
```properties
changeLogFile=example-changelog.sql
liquibase.command.username: dbuser
liquibase.command.password: letmein
```

_driver.properties_:
```properties
liquibase.command.url=jdbc:h2:tcp://localhost:9090/mem:dev
liquibase.command.update.url=jdbc:h2:tcp://localhost:9090/mem:dev
url=jdbc:h2:tcp://localhost:9090/mem:dev
```

Reproducer (extract into some directory, cd into that directory and run `liquibase update --driver-properties-file=driver.properties`):
[repro.zip](https://github.com/user-attachments/files/30464831/repro.zip)

### Expected/Desired Behavior

Liquibase uses the URL specified in _driver.properties_.

### Liquibase Version

5.0.3

### Database Vendor & Version

_No response_

### Liquibase Integration

CLI

### Liquibase Extensions

_No response_

Windows 11

### Additional Context

The property `username` (and its variants with/without `liquibase.command.` prefix) doesn't have an effect when specified in the driver properties file either. Looking at the source code, the properties are passed directly into the JDBC driver: https://github.com/liquibase/liquibase/blob/4d815ea8b10bc33ee9f93b5e5b05ed9f0b532fcb/liquibase-standard/src/main/java/liquibase/database/jvm/JdbcConnection.java#L41-L49
This means that the username must be passed as `user`. This should be documented!
H2 supports `username` as well, but other drivers (like Oracle) may not.

sakshichitnis27:
Fixes #7860

## Impact

## Description

The Liquibase documentation allows JDBC connection properties, including `url`, to be provided through `--driver-properties-file`.

However, when the command did not receive a URL from `--url`, a defaults file, an environment variable, or another configuration source, command validation failed with:

`Invalid argument '--url': missing required argument`

This happened even when the driver properties file contained a valid `url`.

### Root cause

The `url` command argument is validated as required before Liquibase opens the database connection.

Previously, the driver properties file was loaded later while creating the JDBC connection. As a result, the command failed required-argument validation before the URL in the driver properties file could be used.

### Solution

Before command validation and execution, the CLI now:

1. Checks whether a URL has already been provided through an existing Liquibase configuration source.
2. If no URL is configured, checks for `--driver-properties-file`.
3. Resolves the file using Liquibase's existing `PathHandlerFactory`.
4. Loads the plain JDBC `url` property from the file.
5. Adds that URL to the command scope before required-argument validation.

An explicitly configured URL continues to take precedence over the value in the driver properties file.

If the driver properties file does not exist or does not contain a non-empty `url`, the existing required-URL validation behavior is preserved.

## Release note

Liquibase CLI now accepts the JDBC database URL from `--driver-properties-file`, as documented, without also requiring a separate `--url` configuration.

## Things to be aware of

- This change is limited to the Liquibase CLI integration reported in the issue.
- Existing URL sources, including `--url`, defaults files, environment variables, and system properties, retain precedence.
- Only the plain JDBC driver property `url` is read from the driver properties file.
- This change does not reinterpret Liquibase configuration keys such as `liquibase.command.url` as JDBC properties.
- Existing username and JDBC `user` property behavior is unchanged.
- No public API, command metadata, or help-output behavior is changed.

## Things to worry about

None identified.

The fallback is only applied when Liquibase does not already have a configured URL, minimizing the risk of changing existing command behavior.

### Regression tests

Tests were added to verify that:

- A command succeeds when its database URL is supplied only through `--driver-properties-file`.
- An explicitly supplied `--url` takes precedence over a URL in the driver properties file.

### Validation performed

Executed:

`./mvnw -pl liquibase-cli -am test`

Results:

- Liquibase Standard: 6,222 tests passed
- Liquibase CLI: 117 tests passed
- Total: 6,339 tests passed
- Failures: 0
- Errors: 0

Manual verification also confirmed:

- The original reproduction fails on unmodified code with the missing required `--url` error.
- The same command succeeds after the fix when `url` is present only in the driver properties file.
- A properties file without `url` continues to produce the existing required-URL error.
- An explicit URL overrides the driver-properties-file URL.


filipelautert (approved):
Hello @sakshichitnis27 ! Thanks for the PR, looks good.

rberezen:
Reviewed and verified end-to-end — correct fix with the precedence handled properly. LGTM.

**✅ Bug confirmed and fixed** (vs the pre-fix base, H2):
```
--driver-properties-file with url=..., no --url:
  base:    Error: Invalid argument '--url': missing required argument   ❌ (the bug — file's url ignored)
  this PR: Update has been successful                                   ✅
```

**✅ Precedence is correct** — and I checked it decisively by putting an *invalid* URL in the driver-properties file, so "success" proves the explicit source actually won rather than both happening to be valid H2 URLs:

| scenario | result |
|---|---|
| `--url` valid + driver-props `jdbc:invalid:...` | ✅ success → `--url` wins |
| defaults-file url valid + driver-props `jdbc:invalid:...` | ✅ success → defaults-file wins |
| blank `url=` in driver-props, no other source | ✅ correctly "missing required argument" |
| url only in driver-props | ✅ success |

The `getArgumentValue(URL_ARG) != null` guard resolves through the value providers, so `--url`, env, defaults-file, and system-property all pre-empt the driver-props URL as intended.

**✅ The shared `CommandRunner.call()` path is safe for every command.** `getArgumentValue`/`addArgumentValue` resolve/store by the argument's config key and don't require the running command to declare it, so no throw — and the method early-returns after a couple of null lookups whenever no `--driver-properties-file` is present (the overwhelming majority of invocations). Using `PathHandlerFactory` for resolution is the right call (classpath/absolute/relative all work), and the hoisted value flows through the documented `url` key that the connection code already expects.

**Two small notes** (neither blocking):
- The driver-properties file is now parsed twice — once here for the URL, once again at connection time for user/password/etc. Harmless, just a tiny redundancy; could pass the already-loaded `Properties` through if you wanted to avoid it.
- Scope is CLI-only (as the description says). If Maven/other integrations have the same "url in driver-properties-file requires --url anyway" gap, they'd need the same treatment — worth a follow-up issue, but out of scope for #7860.

Nice fix — matches the documented behavior at last. Thanks @sakshichitnis27!

rberezen (approved):

PR #7869, opened 2026-08-01
fix the ChangeSet checksum cache to recalculate when a different checksum version is requested

### LINKED ISSUE #7863: ChangeSet.generateCheckSum(ChecksumVersion) silently ignores the version argument once cached
### Search first

### Description

`ChangeSet.generateCheckSum(ChecksumVersion version)` caches its result in a private
`checkSum` field on first invocation and returns that same cached `CheckSum` object on
every subsequent call — including its embedded version number — regardless of the
`version` argument passed in, until `clearCheckSum()` is called explicitly:

```java
public CheckSum generateCheckSum(ChecksumVersion version) {
    try {
        return Scope.child(Collections.singletonMap(Scope.Attr.checksumVersion.name(), version), () -> {
            if (checkSum == null) {
                // ... computes at `version` and caches into the `checkSum` field ...
            }
            return checkSum;   // returned regardless of the requested `version`
        });
    } catch (Exception e) {
        throw new RuntimeException(e);
    }
}
```

Since checksum content itself differs by version (`AbstractChange.generateCheckSum()`
reads `Scope.getCurrentScope().getChecksumVersion()` to decide field exclusions), a
caller who requests version A, then later requests version B on the same `ChangeSet`
instance without calling `clearCheckSum()` in between, silently gets version A's value
back — with no exception, no log message, and a `CheckSum.getVersion()` that doesn't
match what was asked for.

We found this while investigating a production deployment failure during a 4.19.0 ->
5.0.2 upgrade, where a `runOnChange="true"` changeset was unexpectedly re-executed
right at the moment every stored checksum was being migrated from the v8 to v9 format.
We are not asserting this fully explains that incident by itself — our own changeset
also had `runOnChange="true"` on non-idempotent DDL, which is unsafe regardless of this
defect, and we've since fixed it on our side. We're reporting this cache behavior
because it's a real, independently-reproducible deviation from the method's documented
per-version contract, reachable by any caller (internal or external) that requests two
different checksum versions for the same `ChangeSet` instance without an intervening
`clearCheckSum()`.

### Steps To Reproduce

Minimal standalone reproduction against `liquibase-core:5.0.2`, no database connection
required:

```java
import liquibase.ChecksumVersion;
import liquibase.change.CheckSum;
import liquibase.change.core.RawSQLChange;
import liquibase.changelog.ChangeSet;
import liquibase.changelog.DatabaseChangeLog;

public class CheckSumCacheRepro {
        DatabaseChangeLog changeLog = new DatabaseChangeLog("repro.xml");
        ChangeSet changeSet = new ChangeSet(
                "1", "repro", false, false, "repro.xml", null, null, changeLog
        );
        changeSet.addChange(new RawSQLChange("SELECT 1"));

        CheckSum v8First = changeSet.generateCheckSum(ChecksumVersion.V8);
        System.out.println("Call 1 - requested V8:            " + v8First);

        CheckSum v9NoClear = changeSet.generateCheckSum(ChecksumVersion.V9);
        System.out.println("Call 2 - requested V9 (no clear): " + v9NoClear);
        System.out.println("  -> version actually returned:   " + v9NoClear.getVersion());
        System.out.println("  -> same object as call 1?       " + (v8First == v9NoClear));

        changeSet.clearCheckSum();
        CheckSum v9AfterClear = changeSet.generateCheckSum(ChecksumVersion.V9);
        System.out.println("Call 3 - requested V9 (after clearCheckSum): " + v9AfterClear);
    }
}
```

Output on 5.0.2:

```
Call 1 - requested V8:            8:b967cdcf22aef904b5d80ae2437d32ed
Call 2 - requested V9 (no clear): 8:b967cdcf22aef904b5d80ae2437d32ed
  -> version actually returned:   8
  -> same object as call 1?       true
Call 3 - requested V9 (after clearCheckSum): 9:6a68455a2c7f71d76609a01bf5a61983
```

### Expected/Desired Behavior

Calling `generateCheckSum(ChecksumVersion.V9)` should return a checksum computed at V9
(or otherwise clearly indicate that a cached, different-version value is being reused)
— never silently return a V8-computed value while claiming no error occurred. At
minimum, `generateCheckSum(version).getVersion()` should always equal
`version.getVersion()`, or the method should document/enforce that callers must call
`clearCheckSum()` before requesting a different version, and ideally guard against the
mistake internally (e.g. by keying the cache on the requested version, or clearing
automatically when the requested version differs from what's cached).

### Liquibase Version

5.0.2 (the caching field/logic appears to predate this version; worth checking whether
this affects any release back to 4.22.0, where the version-aware overload of
`generateCheckSum` was introduced)

### Database Vendor & Version

N/A — reproducible with no live database connection

### Liquibase Integration

N/A for the minimal repro (pure Java API call against `liquibase-core`). Originally
observed via a custom application invoking the Liquibase Java API programmatically
(`new Liquibase(changelogFile, resourceAccessor, database).update(contexts)`).

### Liquibase Extensions

None

macOS (local reproduction); originally observed on Linux (AWS Elastic Beanstalk)

### Additional Context

None beyond the above.

renechoi:
## Impact

## Description

Fixes #7863.

`ChangeSet.generateCheckSum(ChecksumVersion)` caches its result in the `checkSum` field and recalculated only when that field was `null`:

```java
return Scope.child(Collections.singletonMap(Scope.Attr.checksumVersion.name(), version), () -> {
    if (checkSum == null) {
        ...
        checkSum = CheckSum.compute(stringToMD5.toString());
    }
    return checkSum;   // returned regardless of the requested version
});
```

So the first calculated value was returned for every later call on that instance, whatever version was asked for, until `clearCheckSum()` was called. `CheckSum.compute` stamps the version that is in effect at calculation time, so the returned object also carried the wrong version number, and `generateCheckSum(version).getVersion()` did not match `version.getVersion()`. No exception, no log message.

This is reachable without a caller that deliberately asks for two versions. The version is picked at runtime from the stored checksum in `ShouldRunChangeSetFilter.checksumChanged`, `AbstractChangeLogHistoryService.getRunStatus`, `ChangeSetStatus` and `RanChangeSet`, while `MarkChangeSetRanGenerator` and `UpdateChangeSetChecksumGenerator` ask for `ChecksumVersion.latest()` when writing `MD5SUM`, so a single `ChangeSet` instance can be asked for two different versions inside one run. `UpdateVisitor.updateCheckSumIfRequired`, `ValidatingVisitorUtil.validateSqlFileChangeAndExpandExpressions` and `ValidatingVisitorUtil.validateCreateFunctionChangeV8ChecksumVariant` already call `clearCheckSum()` immediately before recalculating, which is this behaviour being worked around one call site at a time.

The fix reuses the cache only when it holds a checksum for the version that is currently in effect:

```java
if (checkSum == null || checkSum.getVersion() != getCurrentScope().getChecksumVersion().getVersion()) {
```

The comparison is against the scope value rather than the `version` parameter, because that is exactly the value `CheckSum.compute` will stamp on the result, and because `Scope` falls through to the parent when a child value is `null` (several existing tests call the method through Groovy with no argument, which passes `null` and therefore means "latest").

Reproduction from the issue, run against `liquibase-standard` built from this branch and from `main` at 9a9ea95a2 (JDK 21.0.9):

```
main                                          this branch
Call 1 - requested V8:  8:b967cdcf22ae...     Call 1 - requested V8:  8:b967cdcf22ae...
Call 2 - requested V9:  8:b967cdcf22ae...     Call 2 - requested V9:  9:6a68455a2c7f...
  -> version returned:  8                       -> version returned:  9
  -> same object?       true                    -> same object?       false
Call 3 - after clear:   9:6a68455a2c7f...     Call 3 - after clear:   9:6a68455a2c7f...
```

## Release note

Fixed `ChangeSet.generateCheckSum` returning a checksum that was calculated for a different checksum version. A checksum requested for a specific version is now always calculated for that version, instead of reusing whatever was calculated first for that changeset.

## Things to be aware of

Three tests were added to `ChangeSetTest` and one existing test was adjusted.

The new tests use a `createTable` change with `partitionBy` set, because `CreateTableChange.getExcludedFieldFilters` excludes that field up to V8 and includes it from V9, so the two versions produce genuinely different digests rather than only a different version prefix. On `main` at 9a9ea95a2 the three new tests fail and the rest of `ChangeSetTest` passes:

| test | main | this branch |
| --- | --- | --- |
| `generateCheckSum honors the requested version after another version was already calculated` (V8 then V9) | fail: returned `8:3cf71e88...` for V9 | pass |
| `generateCheckSum honors the requested version after another version was already calculated` (V9 then V8) | fail: returned `9:451d3d83...` for V8 | pass |
| `isCheckSumValid compares against the stored version even when a newer checksum was calculated first` | fail: `isCheckSumValid` returned false | pass |
| `generateCheckSum keeps caching the result when the same version is requested again` | pass | pass |

The adjusted test is `isCheckSumValid_differentButValidCheckSum`. It registered the valid checksum through `changeSet.generateCheckSum()`, which Groovy calls with `null` and therefore evaluates at the latest version, while the stored checksum it then validates against is `8:asdf`. `isCheckSumValid` calculates the current checksum with the version of the stored one, so the version 9 valid checksum only ever matched because the stale cache handed it back for the version 8 request. It now registers the valid checksum for V8. With that single line changed and the production change reverted, it still passes on `main`, so it is not being loosened to accommodate the fix; any fix for this issue invalidates the original cross-version assertion.

Repeatedly alternating versions on the same instance now recalculates each time instead of returning the first value. The call sites above ask for a stable version per run, so this is not a hot path, and correctness of the returned version seemed worth more than keeping a cache hit that returns the wrong answer.

## Things to worry about

Whether `validCheckSum` entries are meant to be comparable across checksum versions. Today `isCheckSumValid` compares them against a checksum calculated at the stored version, so an entry recorded for a different version cannot match. This PR does not change that logic, it only stops the stale cache from making such a match happen by accident. If cross-version matching of `validCheckSum` is wanted, that looks like a separate change to `isCheckSumValid`.

## Additional Context

Verification, all on JDK 21.0.9 (Temurin), macOS arm64:

- `mvn -pl liquibase-standard test` on this branch: `Tests run: 6227, Failures: 0, Errors: 0, Skipped: 12`
- `./mvnw -B -P 'skip-integration-tests' test` for the whole reactor, which is what the unit test workflow runs: BUILD SUCCESS, `6227 / 115 / 77 / 20 / 44 / 400` tests across the modules, 0 failures and 0 errors.

Integration tests against the real databases were not run locally.

### DIFF
--- a/liquibase-standard/src/main/java/liquibase/changelog/ChangeSet.java
+++ b/liquibase-standard/src/main/java/liquibase/changelog/ChangeSet.java
@@ -396,7 +396,10 @@ public class ChangeSet implements Conditional, ChangeLogChild {
     public CheckSum generateCheckSum(ChecksumVersion version) {
         try {
             return Scope.child(Collections.singletonMap(Scope.Attr.checksumVersion.name(), version), () -> {
-                if (checkSum == null) {
+                // The cached value can only be reused when it was calculated with the checksum version that is
+                // in effect now. Otherwise a caller asking for a different version would silently get back a
+                // checksum belonging to the previously requested one, including its version number.
+                if (checkSum == null || checkSum.getVersion() != getCurrentScope().getChecksumVersion().getVersion()) {
                     StringBuilder stringToMD5 = new StringBuilder();
                     for (Change change : this.getChanges()) {
                         // checksum v8 requires changes that are applied even to other databases to be calculated
--- a/liquibase-standard/src/test/groovy/liquibase/changelog/ChangeSetTest.groovy
+++ b/liquibase-standard/src/test/groovy/liquibase/changelog/ChangeSetTest.groovy
@@ -1,6 +1,7 @@
 package liquibase.changelog
 
 
+import liquibase.ChecksumVersion
 import liquibase.Scope
 import liquibase.change.CheckSum
 import liquibase.change.ColumnConfig
@@ -91,6 +92,64 @@ class ChangeSetTest extends Specification {
         assert !md5Sum1.equals(md5Sum2)
     }
 
+    @Unroll
+    def "generateCheckSum honors the requested version after another version was already calculated"() {
+        when:
+        def changeSet = new ChangeSet("testId", "testAuthor", false, false, "com/example/test.xml", null, null, null)
+        changeSet.addChange(createTableChangeWithPartitionBy())
+
+        def firstCheckSum = changeSet.generateCheckSum(firstVersion)
+        def secondCheckSum = changeSet.generateCheckSum(secondVersion)
+
+        then:
+        firstCheckSum.getVersion() == firstVersion.getVersion()
+        secondCheckSum.getVersion() == secondVersion.getVersion()
+        firstCheckSum != secondCheckSum
+
+        where:
+        firstVersion          | secondVersion
+        ChecksumVersion.V8    | ChecksumVersion.V9
+        ChecksumVersion.V9    | ChecksumVersion.V8
+    }
+
+    def "generateCheckSum keeps caching the result when the same version is requested again"() {
+        when:
+        def changeSet = new ChangeSet("testId", "testAuthor", false, false, "com/example/test.xml", null, null, null)
+        changeSet.addChange(createTableChangeWithPartitionBy())
+
+        def firstCall = changeSet.generateCheckSum(ChecksumVersion.V9)
+        def secondCall = changeSet.generateCheckSum(ChecksumVersion.V9)
+
+        then:
+        firstCall.is(secondCall)
+    }
+
+    def "isCheckSumValid compares against the stored version even when a newer checksum was calculated first"() {
+        when:
+        def storedChangeSet = new ChangeSet("testId", "testAuthor", false, false, "com/example/test.xml", null, null, null)
+        storedChangeSet.addChange(createTableChangeWithPartitionBy())
+        def storedCheckSum = storedChangeSet.generateCheckSum(ChecksumVersion.V8)
+
+        // mimics the runtime order of ChangeSetStatus/RanChangeSet/MarkChangeSetRanGenerator, which
+        // materialize the checksum at the latest version before validation reaches the stored one
+        def changeSet = new ChangeSet("testId", "testAuthor", false, false, "com/example/test.xml", null, null, null)
+        changeSet.addChange(createTableChangeWithPartitionBy())
+        changeSet.generateCheckSum(ChecksumVersion.V9)
+
+        then:
+        changeSet.isCheckSumValid(storedCheckSum)
+    }
+
+    private static CreateTableChange createTableChangeWithPartitionBy() {
+        def change = new CreateTableChange()
+        change.setTableName("TABLE_NAME")
+        // partitionBy is excluded from the checksum up to V8 and included from V9 on, so the two
+        // versions produce different checksums for the very same change
+        change.setPartitionBy("RANGE (created_at)")
+        change.addColumn(new ColumnConfig().setName("COLUMN_NAME").setType("INT"))
+        return change
+    }
+
     def isCheckSumValid_validCheckSum() {
         when:
         def changeSet = new ChangeSet("1", "2", false, false, "/test.xml", null, null, null)
@@ -113,7 +172,9 @@ class ChangeSetTest extends Specification {
         CheckSum checkSum = CheckSum.parse("8:asdf")
 
         ChangeSet changeSet = new ChangeSet("1", "2", false, false, "/test.xml", null, null, null)
-        changeSet.addValidCheckSum(changeSet.generateCheckSum().toString())
+        // isCheckSumValid calculates the current checksum with the version of the stored one, so the
+        // valid checksum has to be registered for that same version to be comparable
+        changeSet.addValidCheckSum(changeSet.generateCheckSum(ChecksumVersion.V8).toString())
 
         then:
         assert changeSet.isCheckSumValid(checkSum)


filipelautert (approved):
Hello @renechoi !
Good catch, the cached checksum ignoring the requested version explains #7863 nicely. Ran ChangeSetTest plus the checksum and visitor suites on this branch locally, all green. Please sign the CLA so we can move this. Triggered the CI as it never ran.

rberezen:
Reviewed and verified end-to-end — the fix is correct and the tests are genuine, with one nuance about real-world severity.

**✅ The fix is right.** The `checkSum == null` guard ignored the requested version; gating on `checkSum.getVersion() != getCurrentScope().getChecksumVersion().getVersion()` is the correct root-cause fix. Comparing against the *scope* value (not the `version` parameter) is the right choice — it's exactly what `CheckSum.compute` stamps, and it handles the null-argument case where the scope falls through to the parent's "latest".

**✅ No checksum churn — verified, since checksums are a sensitive area.** The diff touches only the cache *condition*; the computation body is untouched, so a given version's checksum value is identical to `main` by construction. Confirmed empirically (V9, PR == main):
```
partitionBy:  9:be3d63486f13a021439a7bc366c1e60e
plain:        9:07e6bbaa0e107adfa649a41433d88fe1
runOnChange:  9:e6b1bf4c0c02215d96e09f87dfef5e61
```
And crucially, `CheckSum.compute` stamps `getCurrentScope().getChecksumVersion().getVersion()` (CheckSum.java:90) — the *same expression* the new guard compares against — so a repeated call at the same version stays cached (the `keeps caching...` test asserts object identity). It recomputes only when the version genuinely changes, never on every call. Direction of safety is toward *fewer* spurious mismatches, not more.

**✅ The new tests are genuine regression coverage** (I checked — they aren't green on `main`). All three **fail on `main`, pass here**:
```
generateCheckSum honors the requested version ...        main: FAIL   PR: PASS
isCheckSumValid compares against the stored version ...   main: FAIL   PR: PASS
```
360 checksum/validation tests pass on the branch (`ChangeSetTest`, `DatabaseChangeLogTest`, `ValidatingVisitorTest`, `CalculateChecksumCommandStepTest`, `UpdateChangeSetChecksumGenerator`, …).

**⚠️ One nuance on severity.** I set up a genuine V8-stored database (checksum written by real Liquibase 4.19.0: `8:0060…`) and ran the V8→V9 migration on both builds:
```
                       main 5.x            this PR
plain changeset:       Run:0, → 9:…  ✅   Run:0, → 9:…  ✅
runOnChange=true:      Run:0, → 9:…  ✅   Run:0, → 9:…  ✅
```
Both migrate cleanly — no spurious re-execution, correct upgrade to V9 — including the `runOnChange` case matching the reported incident. The reason is that the `update`/`validate` paths already work around this bug with `clearCheckSum()` before recalculating: `UpdateVisitor:93`, `DatabaseChangeLog:1464`, and `ValidatingVisitorUtil` (7 sites). So in the shipped flows the bug is masked.

That makes this a **correct root-cause fix for a currently-latent bug**, rather than a fix for a reproducible `update`-flow regression. The value is (a) making the `generateCheckSum(version)` / `isCheckSumValid` API correct for any caller *not* carrying a `clearCheckSum()` workaround, and (b) enabling those ~9 scattered workarounds to be removed in a follow-up. Worth tempering the description so it doesn't read as "changesets were being re-run in production" — the reporter themselves noted their `runOnChange` on non-idempotent DDL was unsafe regardless.

**Minor:** the `checkSum` field read-then-write in `generateCheckSum` still isn't synchronized, so concurrent calls on one instance could race — but that's pre-existing and out of scope.

Solid, well-reasoned fix with real regression tests. LGTM — just suggest tempering the impact wording and filing the follow-up to remove the now-redundant `clearCheckSum()` workarounds.

rberezen (approved):

Issue #6076, opened 2024-07-05
Calling liquibase CLI programmatically from different threads is broken by ThreadLocal changes in Scope


ashleymercer:
### Search first

- [X] I searched and no similar issues were found

### Description

Since 4.27.0+ we've been seeing several errors like:

```
2024-07-05 14:10:55.416 ERROR l.i.c.LiquibaseCommandLine - Unknown option: '--show-summary=OFF'
picocli.CommandLine$UnmatchedArgumentException: Unknown option: '--show-summary=OFF'
	at picocli.CommandLine$Interpreter.validateConstraints(CommandLine.java:13662)
	at picocli.CommandLine$Interpreter.parse(CommandLine.java:13614)
	at picocli.CommandLine$Interpreter.processSubcommand(CommandLine.java:13904)
	at picocli.CommandLine$Interpreter.processArguments(CommandLine.java:13800)
	at picocli.CommandLine$Interpreter.parse(CommandLine.java:13591)
	at picocli.CommandLine$Interpreter.parse(CommandLine.java:13559)
	at picocli.CommandLine$Interpreter.parse(CommandLine.java:13454)
	at picocli.CommandLine.parseArgs(CommandLine.java:1552)
	at liquibase.integration.commandline.LiquibaseCommandLine.registerValueProviders(LiquibaseCommandLine.java:605)
	at liquibase.integration.commandline.LiquibaseCommandLine.execute(LiquibaseCommandLine.java:361)
```

I think I've managed to track the problem down to pull #5617 and specifically this change:

```java
// Old code
private static ScopeManager scopeManager;

// New code
private static final ThreadLocal<ScopeManager> scopeManager = new ThreadLocal<>();
```

For non-singleton instances I suspect this isn't an issue (else surely there would have been other regressions or bug reports by now!) but it does significantly change the behaviour of `Scope.getSingleton`: previously this would return a *global* singleton instance (assuming we were using the default setup) where now it only returns a *thread-local* singleton - so it's possible for two different threads to see two different "singleton" instances. This in turn breaks any code which both:

1. relies on `Scope.getSingleton`, AND
2. might call liquibase from different threads

The specific bug above manifests because some static initialisers rely on `Scope.getSingleton`: for example the static initialiser for `ShowSummaryArgument` ultimately calls `CommandArgumentDefinition.build()` which contains this line:

```java
Scope.getCurrentScope().getSingleton(CommandFactory.class).register(commandName, newCommandArgument);
```

[...]

PR #7876, opened 2026-08-04
fix: give each command execution its own CommandStep instances

### LINKED ISSUE #6927: [Regression] Multithread build on maven fails with goal updateSQL
### Search first

### Description

There is regression after introduction of changes from https://github.com/liquibase/liquibase/commit/d43cc881f2d7bf79ae49975575a808b6466f8316

Making static field for `COMMAND_DEFINITIONS` generate the same definition for command from separated scopes,
 it leads to case when one thread interfere with command running on another thread.

One place I caught was https://github.com/liquibase/liquibase/blob/33ced29f26c87dca7d9b77a8e0729524f91d5d2d/liquibase-standard/src/main/java/liquibase/command/CommandScope.java#L279
were thread was cleaning up other thread command

Connected issue https://github.com/liquibase/liquibase/issues/3446 (it is regression to already working solution)

Problematic change is fixing this issue https://github.com/liquibase/liquibase/issues/6076, but creates regression.

### Steps To Reproduce

I reproduced it running my maven setup with parameters `-T 1C`, I have multimodule maven project.

### Expected/Desired Behavior

Support multithreaded build in maven, simple revert should help

### Liquibase Version

4.31.1

### Database Vendor & Version

_No response_

### Liquibase Integration

_No response_

### Liquibase Extensions

_No response_

_No response_

### Additional Context

_No response_

filipelautert:
## Impact

## Description

Fixes #6927.

Three separate concurrency defects, all reproduced by the new concurrent CLI test (two parallel updates from a cold start used to fail almost every run, now green):

1. COMMAND_DEFINITIONS handed the same cached CommandDefinition to every caller, so parallel executions shared the pipeline CommandStep instances and one thread could clean up another thread's command. Regression from #6222 once #6872 made the cache key match. The cache now works as a template and getCommandDefinition returns a copy with fresh steps.
2. The definition build itself was not atomic, two threads missing the cache together could consolidate arguments while step classes were still registering them. Now computeIfAbsent.
3. CLI and maven registered their per-execution value providers (argument values, defaults/property files) on the shared LiquibaseConfiguration, so parallel executions resolved each other's values (one thread even migrating the other thread's database) and lost their own arguments when the first execution finished and unregistered. Providers now travel in the execution's scope via LiquibaseConfiguration.SCOPED_VALUE_PROVIDERS_KEY, and the global provider set became a volatile snapshot. This also fixes the maven provider never being unregistered, which leaked one module's properties into later goals of the same JVM.

The single computation that #6222 needed is kept, so the picocli "Unknown option" problem from #6076 stays fixed. Verified with the full liquibase-standard, liquibase-cli and liquibase-maven-plugin suites plus a 5x cold-start stress run of the concurrent test.

### DIFF
--- a/liquibase-cli/src/main/java/liquibase/integration/commandline/LiquibaseCommandLine.java
+++ b/liquibase-cli/src/main/java/liquibase/integration/commandline/LiquibaseCommandLine.java
@@ -377,13 +377,27 @@ public class LiquibaseCommandLine {
 
             Main.setRunningFromNewCli(true);
 
-            final List<ConfigurationValueProvider> valueProviders = registerValueProviders(finalArgs);
-            LogService newLogService = Scope.child(Collections.singletonMap(REGISTERED_VALUE_PROVIDERS_KEY, true), () -> {
+            // The per-execution providers live in the scope, not in the shared LiquibaseConfiguration:
+            // concurrent executions used to see (and unregister) each other's argument values (#6927)
+            final CommandLineArgumentValueProvider argumentProvider =
+                    new CommandLineArgumentValueProvider(commandLine.parseArgs(finalArgs));
+            final List<ConfigurationValueProvider> valueProviders = Scope.child(
+                    Collections.singletonMap(LiquibaseConfiguration.SCOPED_VALUE_PROVIDERS_KEY,
+                            Collections.singletonList(argumentProvider)),
+                    () -> resolveValueProviders(argumentProvider));
+
+            final Map<String, Object> providerScopeValues = new HashMap<>();
+            providerScopeValues.put(LiquibaseConfiguration.SCOPED_VALUE_PROVIDERS_KEY, valueProviders);
+            providerScopeValues.put(REGISTERED_VALUE_PROVIDERS_KEY, true);
+
+            LogService newLogService = Scope.child(providerScopeValues, () -> {
                 // Get a new log service after registering the value providers, since the log service might need to load parameters using newly registered value providers.
                 return Scope.getCurrentScope().getSingleton(LogServiceFactory.class).getDefaultLogService();
             });
 
-            return Scope.child(Collections.singletonMap(Scope.Attr.logService.name(), newLogService), () -> {
+            final Map<String, Object> executionScopeValues = new HashMap<>(providerScopeValues);
+            executionScopeValues.put(Scope.Attr.logService.name(), newLogService);
+            return Scope.child(executionScopeValues, () -> {
                 addEmptyMdcValues();
                 try {
                     return Scope.child(configureScope(args), () -> {
@@ -430,12 +444,7 @@ public class LiquibaseCommandLine {
                         return response;
                     });
                 } finally {
-                    final LiquibaseConfiguration liquibaseConfiguration = Scope.getCurrentScope().getSingleton(LiquibaseConfiguration.class);
-
-                    for (ConfigurationValueProvider provider : valueProviders) {
-                        liquibaseConfiguration.unregisterProvider(provider);
-                    }
-
+                    // scoped value providers simply fall out of scope, there is nothing to unregister
                     LogUtil.setPersistedMdcKeysToEmptyString();
                 }
             });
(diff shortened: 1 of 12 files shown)


wwillard7800 (approved):
Makes sense.

filipelautert:
@wwillard7800 could you review this? It grew beyond the original fix: the new concurrent test exposed two more races (atomic definition build + value providers leaking between parallel executions through the shared LiquibaseConfiguration). All suites green, sonar failure is inherited from main's quality gate.

rberezen:
Reviewed and verified end-to-end — the concurrency fix is solid and does what it claims, but I found one regression that needs addressing before merge.

**✅ The fix works.** I drove two threads calling `LiquibaseCommandLine.execute("update", …)` against two distinct databases from a cold cache (your new test's scenario, on real file-based H2):
```
PR:    all 6 executions exit 0, and each DB contains ONLY its own table (db1→tbl_one, db2→tbl_two) — no cross-contamination.   5/5 runs green.
main:  ALL_ZERO=false, 3/3 runs — fails with "Table DATABASECHANGELOG already exists" and "The object is already closed",
       i.e. one thread tearing down the other's command/connection — exactly the #6927 symptom.
```
I also ran your `LiquibaseCommandLineThreadingTest` directly: green 4/4 on this branch, and it **fails 3/3 on `main`** (`results.every { it == 0 }` violated) — so it's a genuine regression test, not luckily-green. `CommandFactoryTest` + `LiquibaseConfigurationTest` pass (19), and CI is green across all integration DBs + Java/OS combos.

The design across all three defects is coherent: template-copy of `CommandDefinition` with fresh steps, `computeIfAbsent` for atomic build, and copy-on-write `volatile` provider snapshots + scope-carried per-execution providers. Nicely done.

**❌ Regression: `parameter.*` from a defaults file no longer resolves.** Because the PR moves `DefaultsFileValueProvider` from the global `LiquibaseConfiguration` to the *scoped* providers, three consumers that still read the global `getProviders()` no longer see it:
- `ChangeLogParameters.addDefaultFileProperties()` (line 482) — pulls `parameter.*` entries from the defaults file into changelog parameters
- `CommandScope.validate()` (line 245) — runs `provider.validate(this)` per provider
- `InitProjectCommandStep` (line 140) — looks up the `DefaultsFileValueProvider`

The first one is user-facing. Reproduced on both builds — `liquibase.properties` with `parameter.mytable=CUSTOM_TABLE_NAME`, changelog `<createTable tableName="${mytable}">`, `update-sql`:
```
main:  CREATE TABLE PUBLIC.CUSTOM_TABLE_NAME (id INT);     ✅ resolved
PR:    CREATE TABLE PUBLIC."${MYTABLE}" (id INT);           ❌ literal — parameter lost
```
This is a documented feature (changelog parameters via a properties file), so it'd be a real break.

**Suggested fix:** those three call sites need to see scoped providers too. Simplest is to have them iterate the same merged view the resolution path already builds — e.g. expose a `getProviders()` variant (or a helper) that appends `getScopedValueProviders()` and re-sorts by `PROVIDER_ORDER`, mirroring what `getCurrentConfiguredValue()` does — and use it in `addDefaultFileProperties`/`validate`/`InitProjectCommandStep`. Worth a regression test for the `parameter.*`-from-defaults-file path so it can't silently break again.

Great fix on the core issue — just needs the scoped-provider visibility gap closed. Happy to re-verify once updated.

filipelautert:
@rberezen fixed in feb7a8b0, your repro gives CREATE TABLE PUBLIC.CUSTOM_TABLE_NAME again here.
Added getEffectiveProviders() that merges the scoped list into the registered ones and re-sorts by precedence, the three callers use it now. init project was different, it unregisters the defaults file provider on purpose, so its existing Scope.child shadows the scoped list without that provider instead.
Two regression tests, both fail without the fix. thanks for catching it.

rberezen:
Re-verified after `feb7a8b0b` — the scoped-provider gap is fixed. 🎉

**The `parameter.*`-from-defaults-file regression is resolved.** Same repro as before (`liquibase.properties` with `parameter.mytable=CUSTOM_TABLE_NAME`, changelog `<createTable tableName="${mytable}">`, `update-sql`):
```
before feb7a8b0b:   CREATE TABLE PUBLIC."${MYTABLE}"       ❌ (literal, parameter lost)
now:                CREATE TABLE PUBLIC.CUSTOM_TABLE_NAME  ✅ (matches main)
```

The `getEffectiveProviders()` approach is the right call — global + scoped merged and precedence-sorted, sharing the exact ordering logic `getCurrentConfiguredValue` already uses (nicely deduplicated onto it). Both `ChangeLogParameters.addDefaultFileProperties()` and `CommandScope.validate()` now go through it. And the `InitProjectCommandStep` handling is a good catch — since `init project` must *ignore* an existing defaults file, shadowing the scoped `DefaultsFileValueProvider` (not just the global one) keeps that behavior correct now that the provider lives in the scope.

**Nothing regressed:** 73 tests pass across `ChangeLogParametersTest` (incl. the new `parameter.*` case), `LiquibaseConfigurationTest`, `CommandFactoryTest`, `CommandScopeTest`; and the core concurrency fix still holds — `LiquibaseCommandLineThreadingTest` green.

LGTM now. Thanks for the quick turnaround @filipelautert — the concurrency isolation + the scoped-provider fix together make this solid.

rberezen (approved):
thanks!

PR #1768, opened 2021-03-17
Ensure that the root scope is initialized correctly regardless of whether a scope manager has been configured yet or not.

rudolfv:
## Environment
**Liquibase Version**: 4.3.1

**Liquibase Integration & Version**: spring boot

**Liquibase Extension(s) & Version**: N/A

**Database Vendor & Version**: MySQL 5.7.12-log

**Operating System Type & Version**: Amazon AWS RDS

## Pull Request Type
* [x] Bug fix (non-breaking change which fixes an issue.)
* [ ] Enhancement/New feature (non-breaking change which adds functionality)
* [ ] Breaking change (fix or feature that would cause existing functionality to change)

## Description
We have a Spring Boot command line application that applies the same Liquibase changset concurrently to two different database instances using the Liquibase Java API. Currently this fails due to a Liquibase scope collision.

## Steps To Reproduce
The Liquibase.update() method is run with the Java ExecutorService on two different threads. Currently the liquibase Scope class defaults to the SingletonScopeManager if no ScopeManager has been defined. This causes a problem in a multi-threaded application. Therefore we implemented a ThreadLocalScopeManager to ensure that each thread maintains it's own scope.

```
public class ThreadLocalScopeManager extends ScopeManager {

    private ThreadLocal<Scope> currentScope = new ThreadLocal<>();

    @Override
    public Scope getCurrentScope() {
        return currentScope.get();
    }

    @Override
    protected Scope init(Scope scope) {
        return scope;
    }

    @Override
    protected void setCurrentScope(Scope scope) {
        this.currentScope.set(scope);
    }
}
```

Then we set the ScopeManager using the following code prior to creating the worker threads:

```
ThreadLocalScopeManager threadLocalScopeManager = new ThreadLocalScopeManager();
Scope.setScopeManager(threadLocalScopeManager);
```

However, there is a bug in that the liquibase.Scope.getCurrentScope method does not initialize the root scope properly if the ScopeManager was already set externally:

```
    public static Scope getCurrentScope() {
        if (scopeManager == null) {
            scopeManager = new SingletonScopeManager();

            // if scopeManager != null then this code is never reached
            Scope rootScope = new Scope();
            scopeManager.setCurrentScope(rootScope);

            LogService overrideLogService = rootScope.getSingleton(LogServiceFactory.class).getDefaultLogService();
            if (overrideLogService == null) {
                throw new UnexpectedLiquibaseException("Cannot find default log service");
            }
[...]

Issue #6891, opened 2025-04-11
Tomcat reports memory leak because of thread-local variable with SingletonScopeManager


HoffmannTom:
### Search first

- [x] I searched and no similar issues were found

### Description

When undeploying or redeploying web applications, Tomcat reports a (possible) memory leak because the SingletonScopeManager is set within each thread and is not removed on a web application's shutdown.
The message written in the logfile is:
`org.apache.catalina.loader.WebappClassLoaderBase.checkThreadLocalMapForLeaks The web application [ROOT] created a ThreadLocal with key of type [java.lang.InheritableThreadLocal] (value [java.lang.InheritableThreadLocal@3feea1e1]) and a value of type [liquibase.SingletonScopeManager] (value [liquibase.SingletonScopeManager@50ffb3ad]) but failed to remove it when the web application was stopped. Threads are going to be renewed over time to try and avoid a probable memory leak.`

Is there a way to remove the SingletonScopeManager or prevent the ThreadLocal variable to be set?
We use liquibase only during startup of the application.

Any suggestion is welcome :)

### Steps To Reproduce

Use liquibase within a tomcat webapplication and check catalina.log after shutdown or undeploy.

Example-java code:
```
CommandScope updateCommand = new CommandScope(UpdateCommandStep.COMMAND_NAME);
updateCommand.addArgumentValue(DbUrlConnectionArgumentsCommandStep.DATABASE_ARG, db);
updateCommand.addArgumentValue(UpdateCommandStep.CHANGELOG_FILE_ARG, path + "/changelog-root.xml");
updateCommand.execute();
```

### Expected/Desired Behavior

Cleanup data in the threads on shutdown or don't leave data in the threads behind after finishing the tassk.

### Liquibase Version

4.31.1

### Database Vendor & Version

MySQL 8.4

### Liquibase Integration

servlet

### OS and/or Infrastructure Type/Provider

Windows 11 and Ubuntu 24 LTS

### Are you willing to submit a PR?

- [ ] I'm willing to submit a PR (Thank you!)

petepickerill:
Issue should be tested in the context of https://github.com/liquibase/liquibase/pull/6674 upon release.

DrZ7:
Any workaround? Is there even a fix around the corner?

tati-qalified:
@DrZ7 the fix should be in version [4.33.0](https://github.com/liquibase/liquibase/releases/tag/v4.33.0). Would you be able to test it and let us know if it works for you?
[...]

PR #7877, opened 2026-08-05
fix: share a single root scope across threads

### LINKED ISSUE #6880: Possible ServiceLoader issue when calling Scope.getCurrentScope() from a ThreadPool like ForkJoinPool
### Search first

### Description

The following issue appeared using **Java 17**!

When the Scope.getCurrentScope() is called from multiple threads in a ThreadPool (or using a Stream.paralellStream), ServiceLoader issue appears:

`[2025-04-04 13:46:49] INFO [liquibase.servicelocator] Cannot load service: liquibase.configuration.AutoloadedConfigurations: liquibase.GlobalConfiguration not a subtype
java.util.ServiceConfigurationError: liquibase.configuration.AutoloadedConfigurations: liquibase.GlobalConfiguration not a subtype
	at java.base/java.util.ServiceLoader.fail(Unknown Source)
	at java.base/java.util.ServiceLoader$LazyClassPathLookupIterator.hasNextService(Unknown Source)
	at java.base/java.util.ServiceLoader$LazyClassPathLookupIterator.hasNext(Unknown Source)
	at java.base/java.util.ServiceLoader$2.hasNext(Unknown Source)
	at java.base/java.util.ServiceLoader$3.hasNext(Unknown Source)
	at liquibase.configuration.LiquibaseConfiguration.init(LiquibaseConfiguration.java:61)
	at liquibase.Scope.getCurrentScope(Scope.java:106)
......`

The problem in my case is that the liquibase.configuration.AutoloadedConfigurations class has been loaded by URLClassLoader, but the liquibase.GlobalConfiguration is loaded by AppClassLoader, which is the ContextClassLoader of the ForkJoinPool thread. So JVM will see that they are not compatible.

This issue is likely caused by this commit:
https://github.com/liquibase/liquibase/pull/6412/commits/93b5d9ee39c8398cd64f77b7054f06bd1ef0ce3c

When the ForkJoinPool thread calls the liquibase.Scope#getCurrentScope, the scopeManager.get() returns null and it will create a new SingletonScopeManager. Every ForkJoinPool thread creates a new instance and puts into the InheritableThreadLocal variable.

### Steps To Reproduce

Please check the description, unfortunately I don't have to to create repro code

### Expected/Desired Behavior

I think the ScopeManager should be the same in every thread. Sorry, I don't know the exact purpose of this class.
But it's clear that in a multi-threaded env it can cause issues.

### Liquibase Version

4.31.1

### Database Vendor & Version

_No response_

### Liquibase Integration

_No response_

### Liquibase Extensions

_No response_

_No response_

### Additional Context

_No response_

filipelautert:
## Impact

## Description

Fixes #6880.

A thread that never inherited a scope manager (a pool thread created before Liquibase initialized, or from another thread lineage) used to bootstrap a second root scope using its own context classloader. With a different classloader the ServiceLoader pass fails with "X not a subtype" for AutoloadedConfigurations, LogService, ConfigurationValueProvider and ServiceLocator, and the JVM ends up with split-brain singletons.

The first fully-built root scope is now kept and threads without a manager adopt it, exactly what @varrob112 suggested in the issue. Per-thread child scope stacks are unchanged: each thread still gets its own SingletonScopeManager, so the LB-1359 "Cannot end scope X" scenario stays fixed, and root initialization still works whichever ScopeManager an integration sets (PR #1768 behavior preserved).

Verified with the reproduction project from the issue: 26 "not a subtype" errors on the bundled jar, 28 on a core built from current main, 0 on this branch across 3 runs. Full liquibase-standard suite green, including ScopeTest and ThreadLocalScopeManagerTest.

### DIFF
--- a/.github/workflows/run-tests.yml
+++ b/.github/workflows/run-tests.yml
@@ -265,7 +265,7 @@ jobs:
 
   sonar:
     needs: [ setup, build_tests, integration-test ]
-    uses: liquibase/build-logic/.github/workflows/sonar-coverage-merge.yml@main
+    uses: liquibase/build-logic/.github/workflows/sonar-coverage-merge.yml@fix/sonar-pr-mode-on-pull-request-target
     permissions:
       pull-requests: write
       id-token: write
--- a/.github/workflows/run-tests.yml
+++ b/.github/workflows/run-tests.yml
@@ -265,7 +265,7 @@ jobs:
 
   sonar:
     needs: [ setup, build_tests, integration-test ]
-    uses: liquibase/build-logic/.github/workflows/sonar-coverage-merge.yml@fix/sonar-pr-mode-on-pull-request-target
+    uses: liquibase/build-logic/.github/workflows/sonar-coverage-merge.yml@main
     permissions:
       pull-requests: write
       id-token: write
--- a/liquibase-standard/src/main/java/liquibase/Scope.java
+++ b/liquibase-standard/src/main/java/liquibase/Scope.java
@@ -100,6 +100,14 @@ public class Scope {
     private static volatile InheritableThreadLocal<ScopeManager> scopeManager;
 
     /**
+     * The first fully-built root scope of the JVM. Threads that never inherited a scope manager (pool
+     * threads created before Liquibase initialized, or from another thread lineage) adopt this root
+     * instead of bootstrapping a second one with their own context classloader, which produced
+     * "not a subtype" ServiceLoader failures and split-brain singletons (#6880).
+     */
+    private static volatile Scope rootScope;
+
+    /**
      * Lazily initializes and returns the scopeManager ThreadLocal.
      */
     private static InheritableThreadLocal<ScopeManager> getScopeManagerThreadLocal() {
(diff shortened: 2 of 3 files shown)


wwillard7800 (approved):

rberezen:
Reviewed and reproduced end-to-end — this correctly fixes #6880. 👍

**Confirmed the bug and the fix** using @varrob112's reproducer (null-parent `URLClassLoader` on a pool thread racing the main thread's `getCurrentScope()`), building `liquibase-core.jar` from each branch:
```
main:  32 "not a subtype" errors   (×3 runs)
PR:     0 "not a subtype" errors   (×3 runs)
```
On `main` the split-brain surfaces across all four service categories — `AutoloadedConfigurations`, `ConfigurationValueProvider`, `LogService`, and `ServiceLocator` — and the PR clears every one.

**Concurrency looks correct.** The double-checked-locking is sound: `rootScope` is `volatile`, the whole root is constructed inside `synchronized(Scope.class)`, and it's published (`Scope.rootScope = newRootScope`) only *after* full construction — so no thread can observe a partially-built root through the static. The re-entrancy note holds up: the bootstrap (`LiquibaseConfiguration.init`, `findInstances`, `getLog`) re-enters `getCurrentScope()` on the same thread, but since `manager.get().getCurrentScope()` is already set to the partial root before the bootstrap runs, those calls return early without re-entering the block or rebuilding. And each thread still gets its own `SingletonScopeManager` pointing at the shared root, so the per-thread scope stack (LB-1359 "Cannot end scope") and the PR #1768 behavior are preserved.

**Tests:** 13 scope tests pass locally (`ScopeTest` incl. the new orphan-thread case, `ThreadLocalScopeManagerTest`, `SingletonScopeManagerTest`), and CI is green across all 11 integration DBs + every Java/OS combo.

**Two notes:**

**1. Nit — indentation.** Inside the new `synchronized` block, the `if (overrideLogService != null) { … } else { … }` body and the `for`-loop body are under-indented (still at the pre-nesting level), which makes the block harder to read. Worth a quick reformat.

**2. The permanent static reference — worth a small hardening.** `rootScope` is a `static` that's set once and never cleared, and it holds a `ClassLoaderResourceAccessor` that grabbed the classloader of whichever thread happened to build the root first (`new ClassLoaderResourceAccessor()` → `Thread.currentThread().getContextClassLoader()`). Two concrete consequences:

  - **Memory leak on redeploy.** If Liquibase sits in a container's *shared* lib folder (e.g. Tomcat `lib/`, so `Scope` is loaded once and shared by all webapps), the first webapp to touch Liquibase gets its classloader stored in this static. On undeploy/redeploy the container can't discard that webapp's classloader — the shared `Scope` class still references into it — so its classes stay resident, and enough redeploys → Metaspace `OutOfMemoryError`. Same shape as the retention concern in #6891.
  - **"First webapp wins" services.** Every thread adopts that one root, so if two webapps ship different Liquibase extensions, whichever initializes first fixes the resource/service set for the JVM and the other silently inherits it.

  Both only bite the shared-lib-across-multiple-webapps topology; if Liquibase is bundled in each `WEB-INF/lib`, the static dies with the webapp and there's no issue. And keying the root per-classloader wouldn't help — that just recreates the current bug (each classloader gets its own root again).

  **Suggested one-line hardening:** build the root's resource accessor with Liquibase's *own* classloader instead of the winning thread's TCCL —
  ```java
  newRootScope.values.put(Attr.resourceAccessor.name(),
      new ClassLoaderResourceAccessor(Scope.class.getClassLoader()));
  ```
  That makes the static reference only its own (or an ancestor) classloader, which can't be unloaded while `Scope.class` is alive — so no foreign webapp CL gets pinned, and it also removes the thread-order dependence ("first thread wins" → always Liquibase's loader). Optionally a `public static synchronized void resetRootScope()` that `SpringLiquibase`/servlet-listener shutdown can call would let single-app deployments release it explicitly. Neither is blocking, but the one-liner is cheap insurance against the #6891 family.

Nice fix — the single-shared-root approach is the right call. LGTM.

rberezen (approved):

filipelautert:
@rberezen indentation fixed in 83ffd89.
On the static: liquibase goes in WEB-INF/lib, not in the container lib folder, so Scope.class is loaded by the webapp classloader and the static dies with the webapp. Both the pinning and the first-webapp-wins need Scope shared across apps, which doesn't happen.
And the one liner wouldn't cover it anyway, LiquibaseConfiguration.init keeps the ConfigurationValueProvider instances ServiceLoader built, so an extension from the app stays retained.

Issue #7461, opened 2025-12-10
diff-changelog dropps the wrong constraint when constraints differ


mensinda:
### Search first

- [x] I searched and no similar issues were found

### Description

When running `liquibase diff-changelog` and there are differences in a constraint definition, liquibase generates a diff XML where the original constraint is first dropped and then recreated with the reference deffinition.

So far so good, however, on the `dropForeignKeyConstraint` the `diff-changelog` uses the index name of the ***reference DB*** instead of the ***target DB***. This usually is not a problem when both index names are identicall. However, once this is not the case, the generated diff changelog cannot be used on the target database, because the old index cannot be dropped.

---

# Example

Target database (`test_1`) -- the constraint name is `FOO_BAR_ID_CONSTRAINT`:

```sql
create table BAR ( ID int primary key not null );

create table FOO (
    ID int primary key not null,
    BAR_ID varchar(255)
);

CREATE INDEX IDX_TEST_1 ON FOO (BAR_ID);

ALTER TABLE FOO ADD CONSTRAINT FOO_BAR_ID_CONSTRAINT FOREIGN KEY (BAR_ID) REFERENCES BAR(ID) ON DELETE RESTRICT ON UPDATE RESTRICT;
```

Reference database (`test_2`) -- the constraint name is `FKJ04BSSO1GYQ7MJGD6Q17EV02K`:

```sql
create table BAR ( ID int primary key not null);

create table FOO (
    ID int primary key not null,
    BAR_ID varchar(255)
);

CREATE INDEX IDX_TEST_2 ON FOO (BAR_ID);

ALTER TABLE FOO ADD CONSTRAINT FKJ04BSSO1GYQ7MJGD6Q17EV02K FOREIGN KEY (BAR_ID) REFERENCES BAR(ID) ON DELETE NO ACTION ON UPDATE NO ACTION;
```

Now when running `diff-changelog`, it produces the following XML:

```xml
<?xml version="1.1" encoding="UTF-8" standalone="no"?>
[...]

PR #7881, opened 2026-08-07
Fix ClassCastException in diff-changelog when a PK/UC backing index is delegated

### LINKED ISSUE #7846: ClassCastException when diff-changelog from snapshots and swapped primary key column order
### Search first

### Description
Problem does not occur in liquibase 5.0.2, probably related to https://github.com/liquibase/liquibase/pull/7500.

Calling diff-changelog like below leads to ClastCastException.

```
liquibase diff-changelog --changelog-file changelog.yaml --url offline:mariadb?snapshot=test.json --reference-url offline:mariadb?snapshot=reference.json
```

Log: 
```
BEST PRACTICE: The changelog generated by diffChangeLog/generateChangeLog should be inspected for correctness and completeness before being deployed. Some database objects and their dependencies cannot be represented automatically, and they may need to be manually updated before being deployed.
ERROR: Exception Details
ERROR: Exception Primary Class:  ClassCastException
ERROR: Exception Primary Reason:  class liquibase.structure.core.Index cannot be cast to class liquibase.structure.core.PrimaryKey (liquibase.structure.core.Index and liquibase.structure.core.PrimaryKey are in unnamed module of loader java.net.URLClassLoader @200a570f)
ERROR: Exception Primary Source:  Offline mariadb 11.8.8-MariaDB-ubu2404

Unexpected error running Liquibase: class liquibase.structure.core.Index cannot be cast to class liquibase.structure.core.PrimaryKey (liquibase.structure.core.Index and liquibase.structure.core.PrimaryKey are in unnamed module of loader java.net.URLClassLoader @200a570f)

```

diff-changelog on dbs directly works as expected:

```
liquibase diff-changelog --url jdbc:mariadb://host.containers.internal:3333/test --referenceUrl jdbc:mariadb://host.containers.internal:3333/reference --changelog-file=changelog.yaml
```

### Steps To Reproduce

test.json:

```
{
  "snapshot": {
    "created": "2026-07-22T08:09:06.839",
    "database": {
      "productVersion": "11.8.8-MariaDB-ubu2404",
      "shortName": "mariadb",
      "majorVersion": "11",
      "minorVersion": "8",
      "user": "username",
      "productName": "MariaDB",
      "url": "jdbc:mariadb://host.containers.internal:3333/test"
    },
    "objects": {
      "liquibase.structure.core.Catalog": [
        {
          "catalog": {
            "default": true,
            "name": "test",
            "snapshotId": "7ff2100"
          }
        }
      ]
      ,
      "liquibase.structure.core.Column": [
        {
          "column": {
            "name": "a",
            "nullable": false,
            "order": "1!{java.lang.Integer}",
            "relation": "liquibase.structure.core.Table#7ff2102",
            "snapshotId": "7ff2105",
            "type": {
              "columnSizeUnit": "BYTE!{liquibase.structure.core.DataType$ColumnSizeUnit}",
              "dataTypeId": "4!{java.lang.Integer}",
              "radix": "10!{java.lang.Integer}",
              "typeName": "INT"
            }
          }
        },
        {
          "column": {
            "name": "b",
            "nullable": false,
            "order": "2!{java.lang.Integer}",
            "relation": "liquibase.structure.core.Table#7ff2102",
            "snapshotId": "7ff2104",
            "type": {
              "columnSizeUnit": "BYTE!{liquibase.structure.core.DataType$ColumnSizeUnit}",
              "dataTypeId": "4!{java.lang.Integer}",
              "radix": "10!{java.lang.Integer}",
              "typeName": "INT"
            }
          }
        }
      ]
      ,
      "liquibase.structure.core.Index": [
        {
          "index": {
            "columns": [
              {
                "column": {
                  "name": "b"
                }
              },
              {
                "column": {
                  "name": "a"
                }
              }
            ]
            ,
            "name": "PRIMARY",
            "snapshotId": "7ff2103",
            "table": "liquibase.structure.core.Table#7ff2102",
            "unique": true
          }
        }
      ]
      ,
      "liquibase.structure.core.PrimaryKey": [
        {
          "primaryKey": {
            "backingIndex": "liquibase.structure.core.Index#7ff2103",
            "columns": [
              "liquibase.structure.core.Column#7ff2104",
              "liquibase.structure.core.Column#7ff2105"
            ]
            ,
            "name": "PRIMARY",
            "snapshotId": "7ff2106",
            "table": "liquibase.structure.core.Table#7ff2102"
          }
        }
      ]
      ,
      "liquibase.structure.core.Schema": [
        {
          "schema": {
            "catalog": "liquibase.structure.core.Catalog#7ff2100",
            "default": true,
            "name": "test",
            "snapshotId": "7ff2101"
          }
        }
      ]
      ,
      "liquibase.structure.core.Table": [
        {
          "table": {
            "columns": [
              "liquibase.structure.core.Column#7ff2105",
              "liquibase.structure.core.Column#7ff2104"
            ]
            ,
            "default_tablespace": false,
            "indexes": [
              "liquibase.structure.core.Index#7ff2103"
            ]
            ,
            "name": "test",
            "primaryKey": "liquibase.structure.core.PrimaryKey#7ff2106",
            "schema": "liquibase.structure.core.Schema#7ff2101",
            "snapshotId": "7ff2102"
          }
        }
      ]
      
    },
    "snapshotControl": {
      "snapshotControl": {
        "includedType": [
          "liquibase.structure.core.Catalog",
          "liquibase.structure.core.Column",
          "liquibase.structure.core.ForeignKey",
          "liquibase.structure.core.Index",
          "liquibase.structure.core.PrimaryKey",
          "liquibase.structure.core.Schema",
          "liquibase.structure.core.Sequence",
          "liquibase.structure.core.Table",
          "liquibase.structure.core.UniqueConstraint",
          "liquibase.structure.core.View"
        ]
        
      }
    }
  }
}
```

reference.json

```
{
  "snapshot": {
    "created": "2026-07-22T08:09:00.538",
    "database": {
      "productVersion": "11.8.8-MariaDB-ubu2404",
      "shortName": "mariadb",
      "majorVersion": "11",
      "minorVersion": "8",
      "user": "username",
      "productName": "MariaDB",
      "url": "jdbc:mariadb://host.containers.internal:3333/reference"
    },
    "objects": {
      "liquibase.structure.core.Catalog": [
        {
          "catalog": {
            "default": true,
            "name": "reference",
            "snapshotId": "ab63100"
          }
        }
      ]
      ,
      "liquibase.structure.core.Column": [
        {
          "column": {
            "name": "a",
            "nullable": false,
            "order": "1!{java.lang.Integer}",
            "relation": "liquibase.structure.core.Table#ab63102",
            "snapshotId": "ab63104",
            "type": {
              "columnSizeUnit": "BYTE!{liquibase.structure.core.DataType$ColumnSizeUnit}",
              "dataTypeId": "4!{java.lang.Integer}",
              "radix": "10!{java.lang.Integer}",
              "typeName": "INT"
            }
          }
        },
        {
          "column": {
            "name": "b",
            "nullable": false,
            "order": "2!{java.lang.Integer}",
            "relation": "liquibase.structure.core.Table#ab63102",
            "snapshotId": "ab63105",
            "type": {
              "columnSizeUnit": "BYTE!{liquibase.structure.core.DataType$ColumnSizeUnit}",
              "dataTypeId": "4!{java.lang.Integer}",
              "radix": "10!{java.lang.Integer}",
              "typeName": "INT"
            }
          }
        }
      ]
      ,
      "liquibase.structure.core.Index": [
        {
          "index": {
            "columns": [
              {
                "column": {
                  "name": "a"
                }
              },
              {
                "column": {
                  "name": "b"
                }
              }
            ]
            ,
            "name": "PRIMARY",
            "snapshotId": "ab63103",
            "table": "liquibase.structure.core.Table#ab63102",
            "unique": true
          }
        }
      ]
      ,
      "liquibase.structure.core.PrimaryKey": [
        {
          "primaryKey": {
            "backingIndex": "liquibase.structure.core.Index#ab63103",
            "columns": [
              "liquibase.structure.core.Column#ab63104",
              "liquibase.structure.core.Column#ab63105"
            ]
            ,
            "name": "PRIMARY",
            "snapshotId": "ab63106",
            "table": "liquibase.structure.core.Table#ab63102"
          }
        }
      ]
      ,
      "liquibase.structure.core.Schema": [
        {
          "schema": {
            "catalog": "liquibase.structure.core.Catalog#ab63100",
            "default": true,
            "name": "reference",
            "snapshotId": "ab63101"
          }
        }
      ]
      ,
      "liquibase.structure.core.Table": [
        {
          "table": {
            "columns": [
              "liquibase.structure.core.Column#ab63104",
              "liquibase.structure.core.Column#ab63105"
            ]
            ,
            "default_tablespace": false,
            "indexes": [
              "liquibase.structure.core.Index#ab63103"
            ]
            ,
            "name": "test",
            "primaryKey": "liquibase.structure.core.PrimaryKey#ab63106",
            "schema": "liquibase.structure.core.Schema#ab63101",
            "snapshotId": "ab63102"
          }
        }
      ]
      
    },
    "snapshotControl": {
      "snapshotControl": {
        "includedType": [
          "liquibase.structure.core.Catalog",
          "liquibase.structure.core.Column",
          "liquibase.structure.core.ForeignKey",
          "liquibase.structure.core.Index",
          "liquibase.structure.core.PrimaryKey",
          "liquibase.structure.core.Schema",
          "liquibase.structure.core.Sequence",
          "liquibase.structure.core.Table",
          "liquibase.structure.core.UniqueConstraint",
          "liquibase.structure.core.View"
        ]
        
      }
    }
  }
}
```

test.sql:

```
CREATE TABLE `test` (
  `a` int NOT NULL,
  `b` int NOT NULL,
  PRIMARY KEY (`b`,`a`)
)
```

reference.sql:

```
CREATE TABLE `test` (
  `a` int NOT NULL,
  `b` int NOT NULL,
  PRIMARY KEY (`a`,`b`)
)
```

Snapshots created using:
```
liquibase snapshot --url jdbc:mariadb://host.containers.internal:3333/test --snapshot-format=json > test.json
```

for reference analogous.

### Expected/Desired Behavior

diff-changelog from snapshot should behave as with databases.

### Liquibase Version

5.0.3

### Database Vendor & Version

MariaDB 11.8.8-MariaDB-ubu2404

### Liquibase Integration

CLI

### Liquibase Extensions

_No response_

Ubuntu, Container from docker.io/liquibase/liquibase

### Additional Context

_No response_

fudianchn:
## Impact

## Description

`diff-changelog` throws `ClassCastException: class liquibase.structure.core.Index cannot be cast to class liquibase.structure.core.PrimaryKey` when the comparison detects a changed index that backs a primary key or unique constraint. Reported in #7846, reproducible by comparing two offline MariaDB snapshots of the same table whose primary key column order is swapped (`[a,b]` vs `[b,a]`), where the PK is backed by an index named `PRIMARY`.

### Root cause

`ChangedIndexChangeGenerator#fixChanged` delegates to `ChangedPrimaryKeyChangeGenerator` / `ChangedUniqueConstraintChangeGenerator` when the changed index backs a PK or unique constraint, forwarding the index's `ObjectDifferences` (whose comparison object is an `Index`):

```java
return ChangeGeneratorFactory.getInstance().fixChanged(
    ((Table) index.getRelation()).getPrimaryKey(),
    differences, control, referenceDatabase, comparisonDatabase);   // differences describes the Index
```

Since #7500, those generators cast `differences.getComparisonObject()` to their own type to read the target-side constraint name for the drop statement. Under delegation the comparison object is the backing `Index` -> the cast throws.

### Fix

Fix the delegation site in `ChangedIndexChangeGenerator#fixChanged` so the PK/UC generator receives a type-correct `ObjectDifferences`:

1. Resolve the target table from the comparison `Index` already carried in `differences` (`comparisonIndex.getRelation()`).
2. Resolve the target PK (`comparisonTable.getPrimaryKey()`) or target unique constraint (the one whose backing index matches the comparison index), mirroring the reference-side lookup that triggers delegation.
3. Build an `ObjectDifferences` whose comparison object is that target PK/UC, carrying the index field differences over so `hasDifferences()` / `getDifference("columns")` keep working, then delegate.

`ChangedPrimaryKeyChangeGenerator` and `ChangedUniqueConstraintChangeGenerator` keep the #7500 cast unchanged, so the drop statement keeps using the target constraint name and the add statement keeps using the reference name.

This is the delegation-site approach @filipelautert suggested in review; it avoids the reference-name fallback I first tried, which reintroduced the bug #7500 fixed.

Fixes #7846

## Release note

Fixed a regression where `diff-changelog` failed with a `ClassCastException` when a changed index backed a primary key or unique constraint (for example, comparing two databases whose primary key column order differs). `diff-changelog` now completes in that case and still drops and recreates the constraint using each database's own constraint name.

## Things to be aware of

- The fix lives entirely in `ChangedIndexChangeGenerator#fixChanged` (the delegation site). The PK/UC generators are unchanged from #7500.
- Adds a public `ObjectDifferences#getCompareControl()` getter so the delegation site can build a type-correct `ObjectDifferences`. It is purely additive and symmetric with the `getReferenceObject()` / `getComparisonObject()` getters #7500 added.
- DROP keeps using the target constraint name and ADD the reference name, preserving #7500.
- If the target PK/UC cannot be resolved from the comparison index's table, the PK/UC generators' existing null-safe fallback applies (drop falls back to the reference name). This matches prior behavior for that edge case; the fix does not guarantee the target name for every asymmetric snapshot.

## Things to worry about

- If the comparison index's relation is unavailable (a malformed snapshot), the target constraint cannot be resolved and the drop falls back to the reference name via the existing null-safe path. Standard snapshots and the #7846 reproducer always set the relation, so I have not seen this in practice; flagging it for completeness.

## Additional Context

- Issue #7846: the `ClassCastException`.
- PR #7500 introduced the cast whose assumption this delegation path violated. Issue #7461 is the bug #7500 fixed (the drop must use the target constraint name), and this change preserves that behavior on the delegation path.
- New Spock tests exercise the delegation path through `ChangedIndexChangeGenerator#fixChanged` with distinct reference/target constraint names and assert `drop == target name`, `add == reference name` for both PK and UC. The fixtures use the same index name on both sides so `IndexComparator` pairs the indexes as the same object, matching what the real diff pipeline produces.

### DIFF
--- a/liquibase-standard/src/main/java/liquibase/diff/output/changelog/core/ChangedPrimaryKeyChangeGenerator.java
+++ b/liquibase-standard/src/main/java/liquibase/diff/output/changelog/core/ChangedPrimaryKeyChangeGenerator.java
@@ -53,10 +53,16 @@ public class ChangedPrimaryKeyChangeGenerator extends AbstractChangeGenerator im
 
         PrimaryKey pk = (PrimaryKey) changedObject;
 
-        // Use the comparison object for the drop statement to get the correct constraint name from the target database
-        PrimaryKey comparisonPk = (PrimaryKey) differences.getComparisonObject();
-        if (comparisonPk == null) {
-            comparisonPk = pk; // Fallback to reference object if comparison not available
+        // Use the comparison object for the drop statement to get the correct constraint name from the target database.
+        // When delegated from ChangedIndexChangeGenerator (the Index backing the PK changed), the differences object
+        // describes the backing Index, so its comparison object is an Index rather than a PrimaryKey; fall back to the
+        // reference PK in that case (issue #7846).
+        DatabaseObject comparisonObject = differences.getComparisonObject();
+        PrimaryKey comparisonPk;
+        if (comparisonObject instanceof PrimaryKey) {
+            comparisonPk = (PrimaryKey) comparisonObject;
+        } else {
+            comparisonPk = pk;
         }
 
         List<Change> returnList = new ArrayList<>();
--- a/liquibase-standard/src/main/java/liquibase/diff/output/changelog/core/ChangedPrimaryKeyChangeGenerator.java
+++ b/liquibase-standard/src/main/java/liquibase/diff/output/changelog/core/ChangedPrimaryKeyChangeGenerator.java
@@ -53,16 +53,10 @@ public class ChangedPrimaryKeyChangeGenerator extends AbstractChangeGenerator im
 
         PrimaryKey pk = (PrimaryKey) changedObject;
 
-        // Use the comparison object for the drop statement to get the correct constraint name from the target database.
-        // When delegated from ChangedIndexChangeGenerator (the Index backing the PK changed), the differences object
-        // describes the backing Index, so its comparison object is an Index rather than a PrimaryKey; fall back to the
-        // reference PK in that case (issue #7846).
-        DatabaseObject comparisonObject = differences.getComparisonObject();
-        PrimaryKey comparisonPk;
-        if (comparisonObject instanceof PrimaryKey) {
-            comparisonPk = (PrimaryKey) comparisonObject;
-        } else {
-            comparisonPk = pk;
+        // Use the comparison object for the drop statement to get the correct constraint name from the target database
+        PrimaryKey comparisonPk = (PrimaryKey) differences.getComparisonObject();
+        if (comparisonPk == null) {
+            comparisonPk = pk; // Fallback to reference object if comparison not available
         }
 
         List<Change> returnList = new ArrayList<>();
--- a/liquibase-standard/src/main/java/liquibase/diff/ObjectDifferences.java
+++ b/liquibase-standard/src/main/java/liquibase/diff/ObjectDifferences.java
@@ -125,6 +125,13 @@ public class ObjectDifferences {
     }
 
     /**
+     * Returns the {@link CompareControl} backing this comparison.
+     */
+    public CompareControl getCompareControl() {
+        return compareControl;
+    }
+
+    /**
      * Returns the reference database object being compared.
      * @return the reference object, or null if not set
      */
--- a/liquibase-standard/src/main/java/liquibase/diff/output/changelog/core/ChangedIndexChangeGenerator.java
+++ b/liquibase-standard/src/main/java/liquibase/diff/output/changelog/core/ChangedIndexChangeGenerator.java
@@ -59,22 +59,29 @@ public class ChangedIndexChangeGenerator extends AbstractChangeGenerator impleme
 
         Index index = (Index) changedObject;
 
-        if (index.getRelation() != null  && index.getRelation() instanceof Table) {
-            if ((((Table) index.getRelation()).getPrimaryKey() != null) && DatabaseObjectComparatorFactory.getInstance()
-                .isSameObject(((Table) index.getRelation()).getPrimaryKey().getBackingIndex(), changedObject, differences
-                    .getSchemaComparisons(), comparisonDatabase)) {
-                return ChangeGeneratorFactory.getInstance().fixChanged(((Table) index.getRelation()).getPrimaryKey(), differences, control, referenceDatabase, comparisonDatabase);
+        if (index.getRelation() != null && index.getRelation() instanceof Table) {
+            Table referenceTable = (Table) index.getRelation();
+            // The comparison object is the matching Index on the comparison (target) database. Resolve the target
+            // constraint from its table so the delegated PK/UC generator builds the drop statement with the target
+            // constraint name (#7500) instead of falling back to the reference name and reintroducing #7461 (#7846).
+            Index comparisonIndex = (differences.getComparisonObject() instanceof Index) ? (Index) differences.getComparisonObject() : null;
+            Table comparisonTable = ((comparisonIndex != null) && (comparisonIndex.getRelation() instanceof Table)) ? (Table) comparisonIndex.getRelation() : null;
+
+            PrimaryKey referencePrimaryKey = referenceTable.getPrimaryKey();
+            if ((referencePrimaryKey != null) && DatabaseObjectComparatorFactory.getInstance()
+                .isSameObject(referencePrimaryKey.getBackingIndex(), changedObject, differences.getSchemaComparisons(), comparisonDatabase)) {
+                PrimaryKey comparisonPrimaryKey = (comparisonTable != null) ? comparisonTable.getPrimaryKey() : null;
+                return ChangeGeneratorFactory.getInstance().fixChanged(referencePrimaryKey, asConstraintDifferences(differences, referencePrimaryKey, comparisonPrimaryKey), control, referenceDatabase, comparisonDatabase);
             }
 
-            List<UniqueConstraint> uniqueConstraints = index.getRelation().getUniqueConstraints();
+            List<UniqueConstraint> uniqueConstraints = referenceTable.getUniqueConstraints();
             if (uniqueConstraints != null) {
                 for (UniqueConstraint constraint : uniqueConstraints) {
                     if ((constraint.getBackingIndex() != null) && DatabaseObjectComparatorFactory.getInstance()
-                        .isSameObject(constraint.getBackingIndex(), changedObject, differences.getSchemaComparisons()
-                            , comparisonDatabase)) {
-                        return ChangeGeneratorFactory.getInstance().fixChanged(constraint, differences, control, referenceDatabase, comparisonDatabase);
+                        .isSameObject(constraint.getBackingIndex(), changedObject, differences.getSchemaComparisons(), comparisonDatabase)) {
+                        UniqueConstraint comparisonUniqueConstraint = findComparisonUniqueConstraint(comparisonTable, comparisonIndex, differences, comparisonDatabase);
+                        return ChangeGeneratorFactory.getInstance().fixChanged(constraint, asConstraintDifferences(differences, constraint, comparisonUniqueConstraint), control, referenceDatabase, comparisonDatabase);
                     }
-
                 }
             }
         }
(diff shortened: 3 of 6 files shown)


liquibase-standard/src/main/java/liquibase/diff/output/changelog/core/ChangedPrimaryKeyChangeGenerator.java
```diff
@@ -53,10 +53,16 @@ public Change[] fixChanged(DatabaseObject changedObject, ObjectDifferences diffe
...
+        // Use the comparison object for the drop statement to get the correct constraint name from the target database.
+        // When delegated from ChangedIndexChangeGenerator (the Index backing the PK changed), the differences object
+        // describes the backing Index, so its comparison object is an Index rather than a PrimaryKey; fall back to the
+        // reference PK in that case (issue #7846).
+        DatabaseObject comparisonObject = differences.getComparisonObject();
+        PrimaryKey comparisonPk;
+        if (comparisonObject instanceof PrimaryKey) {
+            comparisonPk = (PrimaryKey) comparisonObject;
+        } else {
+            comparisonPk = pk;
```
filipelautert: Hello @fudianchn ! This brings back what #7500 fixed, on this path the drop uses the reference constraint name instead of the target one. The comparison Index is right there in differences, so ((Table) it.getRelation()).getPrimaryKey() gives the target PK. Could you do it at the delegation site as you suggested?

filipelautert (commented):
almost there!

liquibase-standard/src/main/java/liquibase/diff/output/changelog/core/ChangedPrimaryKeyChangeGenerator.java
fudianchn: Done, moved it to the delegation site in `ChangedIndexChangeGenerator#fixChanged`: comparison Index from `differences.getComparisonObject()` -> its table -> target PK (`comparisonTable.getPrimaryKey()`) or the matching target unique constraint -> an `ObjectDifferences` with that target constraint as the comparison object -> delegate. Added `ObjectDifferences#getCompareControl()` to build that. The PK/UC generators keep the #7500 cast, so the drop uses the target constraint name and the add uses the reference name.

Tests use distinct reference/target constraint names and assert `drop == target name`, `add == reference name` on both delegation paths.

filipelautert (approved):
Rebuilding the differences at the delegation site is nicer than what I suggested, and the tests now pin the target name on the drop. Ran the whole liquibase.diff suite here, 223 green.

rberezen (commented):
Reviewed and verified end-to-end — the fix correctly resolves the CCE. LGTM.

**✅ Reproduced on `main`, fixed on this PR** using the issue's exact offline-snapshot scenario (MariaDB, PK column order `[a,b]` vs `[b,a]`, PK backed by index `PRIMARY`):
```
main:      ClassCastException: class ...Index cannot be cast to class ...PrimaryKey
this PR:   diff-changelog succeeds → dropPrimaryKey + addPrimaryKey columnNames="a, b"
```
(Reproduces only via *offline snapshots*, as the issue notes — a live diff of the same two schemas doesn't hit the index→PK delegation and works on both branches.) The generated changelog applies cleanly to a fresh `PK(b,a)` table, leaving the correct `PK(a,b)`.

**✅ The approach is sound.** Rebuilding the `ObjectDifferences` with the *comparison-side* PK/UC (resolved from `comparisonIndex.getRelation()`) as the compare object — carrying the index field diffs over — is exactly right: the delegated generator gets a type-correct object (no cast failure) and still uses the target constraint name for the `drop` (#7500) rather than the reference-name fallback that would reintroduce #7461. The `getCompareControl()` getter is a clean, minimal addition and `findComparisonUniqueConstraint` mirrors the reference-side lookup nicely. 11 generator unit tests pass.

**ℹ️ One pre-existing bug this surfaces — filed separately as #7895, not a blocker.** In the offline path, once the CCE is gone, diff-changelog emits the PK pair **twice** (4 changesets) because the DiffResult contains *both* a changed Index and a changed Primary Key, and the `alreadyHandledChanged` dedup misses offline (PK columns are qualified `pkref.t.a`, the deserialized index's are bare `a`). It predates this PR — even pre-#7500, where the same delegation ran without crashing — and was simply hidden behind the CCE. The output applies cleanly (net-correct PK), so it doesn't block this crash fix; tracked in #7895.

Nice, well-scoped fix on a real regression. Thanks @fudianchn!

rberezen (approved):

PR #7886, opened 2026-08-07
Fix: don't release the changelog lock in AbstractUpdateCommandStep unless it was acquired there

### LINKED ISSUE #5438: Liquibase tries to unlock unacquired lock
### Search first

### Description

Hi.
I think i found issue in latest liquibase versions.
What might be important for you is that we are using custom `PostgresSessionLockService` that depends on `pg_try_advisory_lock` in order to implement multithreaded, (schema based) multitenancy support.

Problem is that liquibase tries to release lock in `finally` block, and the lock was not acquired before.
I think the problem might be in method `liquibase.command.core.AbstractUpdateCommandStep#run`

```
try {
            DatabaseChangeLog databaseChangeLog = (DatabaseChangeLog) commandScope.getDependency(DatabaseChangeLog.class);

// I think here in the next line, it checks for two things:
* isFastCheckEnabled - true
* isUpToDate - true
and then it jumps to the finally block, where lockService.relaseLock() is called, but
the getLockService(database).waitForLock(); was not called at all.

            if (isFastCheckEnabled && isUpToDate(commandScope, database, databaseChangeLog, contexts, labelExpression, resultsBuilder.getOutputStream())) {
                return;
            }
            if(!isDBLocked) {
                LockServiceFactory.getInstance().getLockService(database).waitForLock();
                // waitForLock resets the changelog history service, so we need to rebuild that and generate a final deploymentId.
                ChangeLogHistoryService changeLogHistoryService = Scope.getCurrentScope().getSingleton(ChangeLogHistoryServiceFactory.class).getChangeLogService(database);
                changeLogHistoryService.init();
                changeLogHistoryService.generateDeploymentId();
            }

            // rest of code does not matter...
        } catch (Exception e) {
            // rest of code does not matter...
        } finally {
            //TODO: We should be able to remove this once we get the rest of the update family
            // set up with the CommandFramework
            try {
                LockServiceFactory.getInstance().getLockService(database).releaseLock();
            } catch (LockException e) {
                Scope.getCurrentScope().getLog(getClass()).severe(MSG_COULD_NOT_RELEASE_LOCK, e);
            }
        }
```

This is problematic because after upgrade it starts to fail.

I tried to check when the `isFastCheckEnabled` flag was introduced, but unfortunately there is noting in docs: https://docs.liquibase.com/search.html?q=isFastCheckEnabled#gsc.tab=0&gsc.q=isFastCheckEnabled&gsc.page=1

### Steps To Reproduce

In my case, just upgrade version (with spring-boot 3.1.6 -> 3.2.1) from 4.20.0 to 4.24.0

### Expected/Desired Behavior

Expected behaviour is that the lock will be aqcuired before unlocking, or the logic will be extended to check if any locks are acquired before releasing.

### Liquibase Version

4.24.0

### Database Vendor & Version

11

### Liquibase Integration

spring boot

### Liquibase Extensions

_No response_

_No response_

### Additional Context

_No response_

om7057:
## Description

Fixes #5438.

While looking into the reported lock release issue, I found that the problem affects more than just the early-return path described in the issue.

`AbstractUpdateCommandStep.run()` uses `isDBLocked` to decide whether to release the database lock in its `finally` block. However, commands such as `updateCount`, `updateSql`, and `updateToTag` don't acquire the lock themselves—they rely on the command pipeline's `LockServiceCommandStep` to manage locking. As a result, `run()` attempts to release a lock it never acquired, and `LockServiceCommandStep` later performs the actual cleanup, resulting in a second release attempt and the "could not release lock" message.

This change separates "a lock exists" from "this method acquired the lock." A new flag tracks whether `run()` successfully acquired the lock itself, and the `finally` block only releases the lock in that case. Lock acquisition behavior is otherwise unchanged.

## Testing

Added a regression test covering the early-return scenario from the issue and verifying that no release is attempted when `run()` never acquires the lock.

The new test fails against the previous implementation and passes with this change. I also ran the existing `liquibase.command` test suite along with the related lock service tests to verify there were no regressions.


om7057:
@filipelautert @wwillard7800 just a small reminder on this one!

liquibase-standard/src/main/java/liquibase/command/core/AbstractUpdateCommandStep.java
```diff
@@ -145,7 +154,9 @@ public void run(CommandResultsBuilder resultsBuilder) throws Exception {
             resultsBuilder.addResult("statusCode", 1);
             throw e;
         } finally {
-            if (isDBLocked.get()) {
+            //TODO: We should be able to remove this once we get the rest of the update family
+            // set up with the CommandFramework
```
filipelautert: This reads backwards. All four are CommandSteps already, the odd one out is UpdateCommandStep:93, which overrides requiredDependencies and drops LockService, so its pipeline has no LockServiceCommandStep. The other three inherit it from the base and are fine. Could you reword it to point at that, or just drop the TODO?

filipelautert (changes requested):

liquibase-standard/src/main/java/liquibase/command/core/AbstractUpdateCommandStep.java
om7057: Reworded, thanks. Also found CI failing because of this: removing the internal release entirely broke the `*Sql`-output commands, since their `-- Release Database Lock` line only ever came from this method's own release running before cleanUp() resets the LoggingExecutor. Fixed by checking hasChangeLogLock() at both release sites instead of ThreadLocal bookkeeping, so it releases once when genuinely held either way. Verified against the full h2 suite that was failing in CI, all green now.

filipelautert (approved):
Checking hasChangeLogLock() at both sites is better than the flag, and I ran update-sql here: the Release Database Lock statement is still there, same as main.

rberezen (commented):
Verified end-to-end — solid fix for a real, long-standing lock bug. LGTM.

**✅ Reproduced #5438 on `main`, fixed here.** I mirrored the reporter's setup with a custom `LockService extends StandardLockService` that throws if `releaseLock()` is called without holding the lock, then ran `update-count` on H2:
```
main:    releaseLock() called TWICE → hasChangeLogLock=true (run() finally), then
                                       hasChangeLogLock=false (LockServiceCommandStep.cleanUp) → throws
this PR:  releaseLock() called ONCE → hasChangeLogLock=true, no second attempt
```
So the double-release — harmless-but-noisy for `StandardLockService`, a hard error for custom advisory-lock services like the reporter's — is gone.

**✅ Root cause & wiring confirmed.** The old `finally` released whenever `isDBLocked` was true, but `isDBLocked` defaults to `true` and only `UpdateCommandStep` flips it via `setDBLock(false)`. `UpdateCommandStep` overrides `requiredDependencies()` to drop `LockService`, so it has no pipeline `LockServiceCommandStep` and self-manages the lock; the other three (`updateCount`/`updateSql`/`updateToTag`) inherit `LockService` → the pipeline acquires and `cleanUp` releases, so the old `run()` finally was double-releasing. Switching both sites to guard on the live `hasChangeLogLock()` is the right fix.

**✅ No regression to plain `update`** (single release, verified) and **`update-sql` output is byte-identical** to `main` (only URL/`deployment_id` differ) — the "Release Database Lock" statement is still emitted exactly once. That validates the deliberate choice to keep the release in `run()` (captured while the `LoggingExecutor` is active) rather than deferring it to `cleanUp`, as the code comment explains. `cleanUp` also still calls `resetAll()`, so no lock-service state leak.

**✅ Tests:** new `AbstractUpdateCommandStepTest` (2) and updated `LockServiceCommandStepTest` (7) pass; the new fast-check test genuinely guards the bug (fails against the old `finally`).

Well-diagnosed — nice catch extending it beyond the issue's early-return path to the whole pipeline-managed family. Thanks @om7057!

rberezen (approved):

om7057:
@rberezen trying to fix CI failure, should I do an empty commit?

PR #7892, opened 2026-08-11
Fix SonarCloud bug and vulnerability findings, including reflected XSS in the status servlet

SvampX:
## Impact

## Description

Addresses the bug- and vulnerability-level findings reported by SonarCloud on `main` ([issue list](https://sonarcloud.io/project/issues?issueStatuses=OPEN%2CCONFIRMED&types=BUG%2CVULNERABILITY&id=liquibase)). Of the 25 open findings, 18 are fixed here and 7 are false positives left for Won't Fix triage (listed at the bottom).

### Real defects

Three of these were plainly broken code rather than theoretical nulls:

- **`AbstractJdbcDatabase.getDateLiteral(Date)`** — the parameter is declared `java.util.Date`, so the fourth branch catches every non-null value and the final `else` is reachable *only* when `date == null`. That means `date.getClass().getName()` in the error message was a guaranteed `NullPointerException`, and the intended `Unexpected type` exception could never be thrown. It now reports the null argument.
- **`MissingTableChangeGenerator.getSingleColumnUniqueConstraints`** — `getUniqueConstraints() == null ? null : getUniqueConstraints()` is a no-op ternary that deliberately produces `null`, and the next line iterated it.
- **`ChangeSet.isCheckSumValid`** — the `storedCheckSum == null` branch was unreachable (`java:S2583`): a null stored checksum already returns via the `currentMd5Sum == null` check above it, since the two are computed together.

The remaining ones are genuine null dereferences:

- **`GenericStatusServlet`** reflected `request.getRequestURI()` into an **unquoted** `href=` attribute of a `text/html` response (`javasecurity:S5131`, blocker) — see below.
- **`AbstractJdbcDatabase.escapeColumnNameList(null)`** iterated the `null` that `StringUtil.splitAndTrim` returns for null input; now returns an empty string.
- **`CreateProcedureChange.createExampleValueMetaData`** called `put()` on the map its parent returns as `null` when there is no annotation to read defaults from.
- **`InitProjectCommandStep.isSkipped`** dereferenced the nullable result of `ConfiguredValue.getValue()`; operands flipped.
- **`DiffToChangeLog`** — dropped the redundant `?` from the `(.*)?` capture group (`java:S5842`).

### XSS fix

Rather than escaping the reflected URI, the level links are now relative (`href="?logLevel=…"`), which preserves the existing navigation behaviour and removes the tainted data from the response entirely. Log messages and stack traces written to the same page are HTML-escaped as defence in depth — this also fixes rendering, since a log message containing `<b>` was previously interpreted as markup rather than displayed.

### Hardening

Behaviour-preserving outside the NPE paths. The most substantive pattern: `ForeignKey.hashCode` and `UniqueConstraint.hashCode` null-checked one getter call and then dereferenced a **second** call to the same getter, so the check did not actually guard the dereference. Each getter is now read once into a local. Also: `PrimaryKey.compareTo` tolerates an absent column list; `UniqueConstraintSnapshotGenerator` reuses the `schema` local instead of re-reading it and reports a missing schema explicitly; and `MissingUniqueConstraintChangeGenerator`, `AddUniqueConstraintGenerator` and `DatabaseList.toDbmsSet` read the backing relation, column list and trimmed dbms list null-safely.

## Release note

Fixed a cross-site scripting vulnerability in the Liquibase status servlet, where the request URL and logged messages were written into the status page without escaping. Also fixed several cases where Liquibase reported a `NullPointerException` instead of a meaningful error — most visibly when generating a date literal from a null value, and when diffing a table whose unique-constraint information was unavailable.

## Things to be aware of

- **`DiffToChangeLog` regex.** The change from `(.*)?` to `(.*)` is provably equivalent, not just probably: the greedy optional group always participates, so `m.group(1)` was never null. Verified identical output on `foo()`, `calculate_bonus(emp_salary numeric)`, `foo(())` and trailing-whitespace inputs.
- **`commons-text` is now used from main code.** `GenericStatusServlet` calls `StringEscapeUtils.escapeHtml4`. `commons-text` is already a direct compile-scope dependency of `liquibase-standard` (declared explicitly for dependabot control) and reaches the classpath via opencsv, but this is its first use outside tests. OSGi picks it up through the `*;resolution:=optional` catch-all in `Import-Package`. Happy to hand-roll the escaping instead if we would rather not add that edge.
- **`UniqueConstraintSnapshotGenerator.listColumns`** now throws `UnexpectedLiquibaseException` for a null schema. This cannot regress anything: the old code dereferenced `example.getSchema()` unconditionally on the very next line, so a null schema always threw — just with a less useful exception.
- **`getDateLiteral`** now throws `IllegalArgumentException` rather than `RuntimeException`. `IllegalArgumentException` is a `RuntimeException`, so existing catch blocks are unaffected, and the old path threw NPE regardless.

## Things to worry about

- Two SonarCloud findings were deliberately **not** fixed as specified, because the rule's suggested remedy is wrong for this code:
  - `DatabaseFactory:214` (`java:S1872`) suggests `isAssignableFrom`, which would treat a *subclass* as already-registered. The name comparison is also classloader-tolerant, which matters for extensions loaded twice.
  - The `ObjectDifferences` / `MSSQLViewComparator` null checks would be dead code: `DatabaseObjectComparatorChain.findDifferences` returns early for every null combination before dispatching to a comparator.
- No unit test for `InitProjectCommandStep.isSkipped` — it is private and only reachable through the integration-test harness.

## Additional Context

**Testing.** Full `liquibase-standard` suite: 6454 tests, no failures introduced. The 13 failures in `StringUtilTest`/`SimpleSqlGrammarTest` are pre-existing — reproduced identically on clean `main` with this branch stashed.

New coverage: `GenericStatusServletTest` (asserts the request URI is absent from the response and that script payloads in messages and stack traces are escaped), plus cases in `AbstractJdbcDatabaseTest` (which fan out to 6 database subclasses), `ChangeSetTest`, `CreateProcedureChangeTest` and `MissingTableChangeGeneratorTest`.

**False positives for Won't Fix triage.** Paste into the SonarCloud issue filter:

```
AZ9_mytzWm01xHNvjVbS,AZ9_mzesWm01xHNvjVbc,AY6vSIOIbnZFYGQcC3O_,AZ9_mx9-Wm01xHNvjVbK,AZ9_mykSWm01xHNvjVbP,AZ9_mzesWm01xHNvjVbb,AZ9_mywHWm01xHNvjVbT
```

| Finding | Why it is not a bug |
|---|---|
| `ObjectDifferences:76`, `MSSQLViewComparator:61` | The comparator chain returns early for every null combination before dispatching, so a comparator cannot receive a null object. Sonar reads the chain's own null checks as evidence the parameters are nullable. |
| `LiquibaseCommandLine:1192,1325`, `EnvironmentValueProvider:133` | Private helpers fed command/config keys; null is not reachable. |
| `CheckSum:88` | Every in-tree caller passes a string concatenation. |
| `DatabaseFactory:214` | See "Things to worry about" above. |

The dominant pattern across the `javabugs:S2259` findings is that null-in/null-out helpers (`StringUtil.toKabobCase`/`join`/`splitAndTrim`, `Column.listFromNames`) give every call site a theoretical NPE.

### DIFF
--- a/liquibase-standard/src/main/java/liquibase/diff/output/changelog/DiffToChangeLog.java
+++ b/liquibase-standard/src/main/java/liquibase/diff/output/changelog/DiffToChangeLog.java
@@ -514,7 +514,8 @@ public class DiffToChangeLog {
         if (! (database instanceof AbstractPostgresDatabase) || ! (objectName.contains("(") && objectName.contains(")"))) {
             return name;
         }
-        Pattern p = Pattern.compile(".*?[(]+(.*)[)]+[\\s]*?$");
+        // (.+) rather than (.*): a no-argument signature has nothing to rewrite, so it keeps the plain schema.name
+        Pattern p = Pattern.compile(".*?[(]+(.+)[)]+[\\s]*?$");
         Matcher m = p.matcher(objectName);
         if (m.matches()) {
             String originalParameters = m.group(1);
--- a/liquibase-standard/src/main/java/liquibase/diff/output/changelog/DiffToChangeLog.java
+++ b/liquibase-standard/src/main/java/liquibase/diff/output/changelog/DiffToChangeLog.java
@@ -514,7 +514,7 @@ public class DiffToChangeLog {
         if (! (database instanceof AbstractPostgresDatabase) || ! (objectName.contains("(") && objectName.contains(")"))) {
             return name;
         }
-        Pattern p = Pattern.compile(".*?[(]+(.*)?[)]+[\\s]*?$");
+        Pattern p = Pattern.compile(".*?[(]+(.*)[)]+[\\s]*?$");
         Matcher m = p.matcher(objectName);
         if (m.matches()) {
             String originalParameters = m.group(1);
--- a/liquibase-standard/src/main/java/liquibase/integration/servlet/GenericStatusServlet.java
+++ b/liquibase-standard/src/main/java/liquibase/integration/servlet/GenericStatusServlet.java
@@ -4,7 +4,6 @@ import liquibase.Scope;
 
 import java.io.IOException;
 import java.io.PrintWriter;
-import java.io.StringWriter;
 import java.text.DateFormat;
 import java.util.ArrayList;
 import java.util.Date;
@@ -68,7 +67,11 @@ class GenericStatusServlet {
                     if (record.getLevel().intValue() >= currentLevel.intValue()) {
                         writer.println(record.getLevel() + ": " + escapeHtml(record.getMessage()));
                         if (record.getThrown() != null) {
-                            writer.println(escapeHtml(asString(record.getThrown())));
+                            Throwable thrown = record.getThrown();
+                            // The stack trace goes to the server log rather than the response, so that paths,
+                            // internal class structure and dependency versions are not disclosed to the caller.
+                            Scope.getCurrentScope().getLog(getClass()).severe("Liquibase run error", thrown);
+                            writer.println(escapeHtml(thrown.getClass().getSimpleName() + ": " + thrown.getMessage()));
                         }
                     }
                 }
@@ -121,14 +124,6 @@ class GenericStatusServlet {
             }
         }
         return escaped.toString();
-    }
-
-    private static String asString(Throwable thrown) {
-        StringWriter stackTrace = new StringWriter();
-        try (PrintWriter stackTraceWriter = new PrintWriter(stackTrace)) {
-            thrown.printStackTrace(stackTraceWriter);
-        }
-        return stackTrace.toString();
     }
 
     /**
--- a/liquibase-standard/src/main/java/liquibase/integration/servlet/GenericStatusServlet.java
+++ b/liquibase-standard/src/main/java/liquibase/integration/servlet/GenericStatusServlet.java
@@ -1,9 +1,11 @@
 package liquibase.integration.servlet;
 
 import liquibase.Scope;
+import org.apache.commons.text.StringEscapeUtils;
 
 import java.io.IOException;
 import java.io.PrintWriter;
+import java.io.StringWriter;
 import java.text.DateFormat;
 import java.util.ArrayList;
 import java.util.Date;
@@ -49,13 +51,13 @@ class GenericStatusServlet {
             if (liquibaseRunLog.isEmpty()) {
                 writer.println("<b>Liquibase did not run</b>");
             } else {
-                writer.println("<b>View level: " + getLevelLink(Level.SEVERE, currentLevel, httpServletRequest)
-                        + " " + getLevelLink(Level.WARNING, currentLevel, httpServletRequest)
-                        + " " + getLevelLink(Level.INFO, currentLevel, httpServletRequest)
-                        + " " + getLevelLink(Level.CONFIG, currentLevel, httpServletRequest)
-                        + " " + getLevelLink(Level.FINE, currentLevel, httpServletRequest)
-                        + " " + getLevelLink(Level.FINER, currentLevel, httpServletRequest)
-                        + " " + getLevelLink(Level.FINEST, currentLevel, httpServletRequest)
+                writer.println("<b>View level: " + getLevelLink(Level.SEVERE, currentLevel)
+                        + " " + getLevelLink(Level.WARNING, currentLevel)
+                        + " " + getLevelLink(Level.INFO, currentLevel)
+                        + " " + getLevelLink(Level.CONFIG, currentLevel)
+                        + " " + getLevelLink(Level.FINE, currentLevel)
+                        + " " + getLevelLink(Level.FINER, currentLevel)
+                        + " " + getLevelLink(Level.FINEST, currentLevel)
                         + "</b>");
 
                 writer.println("<hr>");
@@ -65,9 +67,9 @@ class GenericStatusServlet {
                 writer.println("<pre>");
                 for (LogRecord record : liquibaseRunLog) {
                     if (record.getLevel().intValue() >= currentLevel.intValue()) {
-                        writer.println(record.getLevel() + ": " + record.getMessage());
+                        writer.println(record.getLevel() + ": " + StringEscapeUtils.escapeHtml4(record.getMessage()));
                         if (record.getThrown() != null) {
-                            record.getThrown().printStackTrace(writer);
+                            writer.println(StringEscapeUtils.escapeHtml4(asString(record.getThrown())));
                         }
                     }
                 }
@@ -87,11 +89,23 @@ class GenericStatusServlet {
         }
     }
 
-    private String getLevelLink(Level level, Level currentLevel, GenericServletWrapper.HttpServletRequest request) {
+    private static String asString(Throwable thrown) {
+        StringWriter stackTrace = new StringWriter();
+        try (PrintWriter stackTraceWriter = new PrintWriter(stackTrace)) {
+            thrown.printStackTrace(stackTraceWriter);
+        }
+        return stackTrace.toString();
+    }
+
+    /**
+     * Builds the link used to switch the displayed log level. The link is deliberately relative (query string only) so
+     * that no part of the incoming request is reflected back into the response.
+     */
+    private String getLevelLink(Level level, Level currentLevel) {
         if (currentLevel.equals(level)) {
             return level.getName();
         } else {
-            return "<a href=" + request.getRequestURI() + "?logLevel=" + level.getName() + ">" + level.getName() + "</a>";
+            return "<a href=\"?logLevel=" + level.getName() + "\">" + level.getName() + "</a>";
         }
     }
 }
--- a/liquibase-standard/src/main/java/liquibase/integration/servlet/GenericStatusServlet.java
+++ b/liquibase-standard/src/main/java/liquibase/integration/servlet/GenericStatusServlet.java
@@ -1,7 +1,6 @@
 package liquibase.integration.servlet;
 
 import liquibase.Scope;
-import org.apache.commons.text.StringEscapeUtils;
 
 import java.io.IOException;
 import java.io.PrintWriter;
@@ -67,9 +66,9 @@ class GenericStatusServlet {
                 writer.println("<pre>");
                 for (LogRecord record : liquibaseRunLog) {
                     if (record.getLevel().intValue() >= currentLevel.intValue()) {
-                        writer.println(record.getLevel() + ": " + StringEscapeUtils.escapeHtml4(record.getMessage()));
+                        writer.println(record.getLevel() + ": " + escapeHtml(record.getMessage()));
                         if (record.getThrown() != null) {
-                            writer.println(StringEscapeUtils.escapeHtml4(asString(record.getThrown())));
+                            writer.println(escapeHtml(asString(record.getThrown())));
                         }
                     }
                 }
@@ -87,6 +86,41 @@ class GenericStatusServlet {
             Scope.getCurrentScope().getLog(getClass()).severe("Error in doGet: "+e.getMessage(), e);
             httpServletResponse.setStatus(500);
         }
+    }
+
+    /**
+     * Escapes text that is written into this page. Deliberately dependency-free: this servlet ships inside
+     * liquibase-core, which cannot assume that jars beyond the ones in the distribution's internal/lib are on the
+     * classpath. Covers the characters that matter both in element text and inside a quoted attribute.
+     */
+    private static String escapeHtml(String text) {
+        if (text == null) {
+            return null;
+        }
+        StringBuilder escaped = new StringBuilder(text.length());
+        for (int i = 0; i < text.length(); i++) {
+            char character = text.charAt(i);
+            switch (character) {
+                case '&':
+                    escaped.append("&amp;");
+                    break;
+                case '<':
+                    escaped.append("&lt;");
+                    break;
+                case '>':
+                    escaped.append("&gt;");
+                    break;
+                case '"':
+                    escaped.append("&quot;");
+                    break;
+                case '\'':
+                    escaped.append("&#39;");
+                    break;
+                default:
+                    escaped.append(character);
+            }
+        }
+        return escaped.toString();
     }
 
     private static String asString(Throwable thrown) {
--- a/liquibase-standard/src/main/java/liquibase/change/core/CreateProcedureChange.java
+++ b/liquibase-standard/src/main/java/liquibase/change/core/CreateProcedureChange.java
@@ -25,6 +25,7 @@ import java.io.IOException;
 import java.io.InputStream;
 import java.io.UnsupportedEncodingException;
 import java.nio.charset.Charset;
+import java.util.HashMap;
 import java.util.Map;
 
 @DatabaseChange(name = "createProcedure", description = "Defines a stored procedure.", priority = ChangeMetaData.PRIORITY_DEFAULT)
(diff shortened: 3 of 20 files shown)


SvampX:
Addressed both CodeQL findings in 97e06c7f1.

**`DatabaseList:72`** — now calls `org.apache.commons.lang3.StringUtils.trimToNull` directly. `StringUtil.trimToNull` is a one-line delegation to it, so this is a no-op behaviourally. Note there are ~155 other `StringUtil.trimToNull` call sites in `liquibase-standard/src/main`; I only changed the line this PR already touched, since a repo-wide deprecation migration belongs in its own change.

**`MissingUniqueConstraintChangeGenerator:122`** — switched to the non-deprecated `Index.getRelation()`, and dropped my hand-rolled null-safe hop in favour of `Index.getSchema()`, which already does exactly that (`getRelation() == null ? null : getRelation().getSchema()`). This also matches how the rest of the class reads a constraint's relation (lines 52–65).

One semantic note worth a reviewer's eye: the deprecated `getTable()` returned `null` when the relation was not a `Table`, whereas `getRelation()` returns it as-is. For a unique constraint's backing index the relation is a table, so this is unchanged in practice — and in the case where they would differ, the old code dereferenced `null` and fell into the `catch (Exception)` below, logging a warning and reporting "not found". So the new behaviour is a strict improvement rather than a regression.

Affected tests re-run green (65 tests across `DatabaseListTest`, `MissingUniqueConstraintChangeGeneratorTest`, `UniqueConstraintComparatorTest`, `DiffToChangeLogTest`, `MissingTableChangeGeneratorTest`, `GenericStatusServletTest`).

Also: CodeRabbit's review failed with an internal error rather than producing findings, so there is nothing to action from it yet.

liquibase-standard/src/main/java/liquibase/integration/servlet/GenericStatusServlet.java
```diff
@@ -1,9 +1,11 @@
 package liquibase.integration.servlet;
 
 import liquibase.Scope;
+import org.apache.commons.text.StringEscapeUtils;
```
filipelautert: commons-text is not in the distribution, internal/lib only ships commons-io, commons-lang3, h2, liquibase-core, opencsv, picocli and snakeyaml, and the core shade does not bundle it either. Maven consumers get it transitively but a servlet loaded from the distribution jars would hit NoClassDefFoundError. commons-lang3 is already there and has escapeHtml4. Should we use that one instead?

SvampX: Good catch, and thank you — you're right that this couldn't stay. Dropped commons-text in 5cc8fa319.

I didn't go with commons-lang3 though: its `StringEscapeUtils` has been deprecated since 3.6 in favour of commons-text, and I confirmed it's still deprecated in the 3.20.0 we ship (`Deprecated: true` on `escapeHtml4` in the bytecode). Using it would immediately trip the same CodeQL "deprecated method invocation" rule that flagged two other lines on this PR — trading a packaging risk for a scan failure.

So the escaping is now a small dependency-free helper inside the servlet, covering `& < > " '`. Two upsides beyond the packaging fix: it also escapes the apostrophe, which `escapeHtml4` does not, and there's no longer any question about optional jars for a class that ships inside liquibase-core.

Worth noting the blocker Sonar finding doesn't depend on this at all — that was the request URI being reflected into an *unquoted* `href`, fixed by making the level links relative (`href="?logLevel=…"`). The escaping here is defence in depth for the log messages and stack traces on the same page, and it also fixes rendering: a log message containing `<b>` was previously interpreted as markup.

liquibase-standard/src/main/java/liquibase/diff/output/changelog/DiffToChangeLog.java
```diff
@@ -514,7 +514,7 @@ private static String convertStoredLogicObjectName(String schemaName, String obj
         if (! (database instanceof AbstractPostgresDatabase) || ! (objectName.contains("(") && objectName.contains(")"))) {
             return name;
         }
-        Pattern p = Pattern.compile(".*?[(]+(.*)?[)]+[\\s]*?$");
+        Pattern p = Pattern.compile(".*?[(]+(.*)[)]+[\\s]*?$");
```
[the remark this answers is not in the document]
SvampX: Taken in 5cc8fa319 — switched to `(.+)` and added the regression tests.

One correction for the record: this wasn't a functional-correctness problem. With `(.*)` the no-argument case did enter the conversion path, but every step was a no-op, so the output was already correct. `originalParameters` is `""`, and `"foo()".replace("", "")` returns `"foo()"` unchanged, while `.replace(", ", ",")` finds nothing — so the result was `schema.foo()`, exactly what `(.+)` produces by not matching and falling through to the default `name`.

I verified that end-to-end rather than reasoning about it, comparing both patterns through a replica of the method across `foo()`, `foo( )`, `foo(())`, `foo()  `, `one_arg(x integer)`, `weird(a integer,b text)`, `no_parens` and the multi-parameter javadoc example. Output was byte-identical on all of them.

Still worth taking: `(.+)` makes the intent explicit and skips a pointless loop. New coverage in `DiffToChangeLogTest` for the parameter rewrite, the no-argument case, and the non-Postgres passthrough (reflection, since `convertStoredLogicObjectName` is private static).

liquibase-standard/src/main/java/liquibase/integration/servlet/GenericStatusServlet.java
SvampX: Fixed in 5fb738d10. The full stack trace now goes to the server log via `Scope.getCurrentScope().getLog(...).severe(...)`, and the page renders only the escaped exception type and message — so it still tells you what failed without disclosing file paths, internal class structure or dependency versions. The `asString` helper is gone.

@filipelautert — two things worth your judgement here, since this is your call rather than mine:

**This exposure predates the PR.** The original line was `record.getThrown().printStackTrace(writer)`, writing the raw trace straight to the HTTP response. CodeQL only surfaced it now because my diff touched that line. So this is not a regression I introduced, and if you would rather keep full traces on a diagnostic page and dismiss alert 2298 as intentional, that is a defensible position — say so and I will revert this commit.

**The wider surface is untouched.** `GenericStatusServlet` renders the entire Liquibase run log — messages, levels, timings — to any unauthenticated caller who can reach the URL, and those messages routinely contain SQL, table names and file paths. Removing the stack trace narrows the disclosure; it does not close it. That looks like a deliberate design choice for an opt-in `web.xml` diagnostic endpoint, but flagging it in case it deserves its own issue.

If you want it, an obvious hardening would be requiring an explicit opt-in property before the servlet serves anything, but that is a behaviour change well outside a Sonar-cleanup PR.

SvampX: Follow-up: 5fb738d10 cleared `java/stack-trace-exposure` (2298), but CodeQL replaced it with **`java/error-message-exposure` (2299)** on `thrown.getMessage()`. That rule family treats any exception-derived text reaching an HTTP response as exposure, so no middle ground satisfies it — the only code change that clears it is emitting nothing from the throwable at all.

Proposing we **dismiss 2299 as accepted risk** and keep the type and message. Reasoning, for the record:

1. The part that actually mattered is already gone. 5fb738d10 removed the file paths, internal class structure and dependency versions that the full trace leaked; the trace goes to the server log now.
2. What remains is not distinguishable from what the page already shows. `GenericStatusServlet` renders the whole Liquibase run log — messages, levels, timings — and those messages routinely contain SQL and table names. CodeQL does not flag them only because they are not `Throwable`-derived, which is a taint-source artifact rather than a security distinction.
3. Dropping to type-only would probably still be flagged (the alert lists 5 flow paths), so it would cost another scan round trip while removing most of the remaining diagnostic value.
4. Neither alert gates CI — `Analyze (java)` and `CodeQL` both report pass; these are annotations plus a code-scanning entry.

I have deliberately not dismissed it myself — that is a maintainer action on the security tab. @filipelautert if you would rather go the other way, the two alternatives are a fixed "(details in the server log)" marker, which clears the alert outright, or reverting both stack-trace commits to the pre-PR behaviour and dismissing 2298/2299 as intentional. Happy to do either; just say which.

The wider point from my previous comment still stands independently of this: the servlet is an unauthenticated diagnostic endpoint disclosing the run log, which may deserve its own issue.

SvampX:
## CI status: 29 green, 2 red — both red ones pre-existing

Full unit matrix green (Java 17/21/25 on ubuntu/windows/macOS), all 10 integration databases green including oracle, `Analyze (java)` and `CodeQL` green, CodeRabbit green.

### Sonar confirms the cleanup

Open bug + vulnerability findings on the project went **25 → 9**. The 9 remaining are exactly the 7 false positives documented in the PR description (two line numbers shifted by these edits: `LiquibaseCommandLine` 1192→1200 and 1325→1333), plus the two below. So all 18 intended fixes are confirmed resolved.

### `sonar / sonar` failure is not from this PR

Quality gate fails on `new_reliability_rating = B (2)`, where A is required. The cause is two MINOR `java:S3077` findings (volatile non-primitive field):

- `liquibase-standard/src/main/java/liquibase/Scope.java:107`
- `liquibase-standard/src/main/java/liquibase/configuration/LiquibaseConfiguration.java:50`

Both fields are present in this branch's base commit (`origin/main`, c7b3120), and **neither file is in this diff** — the last commit touching `Scope.java` was a28f39b0d (#7825). Nothing here can raise the reliability rating; this PR only removes bug-level findings.

### ⚠️ The Sonar job is not doing a PR analysis

Worth a separate look, @filipelautert:

```
GET /api/qualitygates/project_status?projectKey=liquibase&pullRequest=7892
{"errors":[{"msg":"Component 'liquibase' of pull request '7892' not found"}]}
```

The scanner log points at `sonarcloud.io/dashboard?id=liquibase&branch=main`, and `api/measures/search_history` shows 12 analyses today between 11:38 and 14:06 — matching this PR's CI runs — all landing on the project's **main** branch. So PR builds appear to be overwriting main's SonarCloud analysis, and the gate being enforced against this PR is main's 30-day new-code gate rather than the PR's own diff. That would explain why the check is red for changes it never actually measured in isolation.

### `Artifacts - Build & Package` failure is a version-substitution bug

The deploy step tried to publish an empty version:

```
Failed to retrieve remote metadata org.liquibase:liquibase-core:-SNAPSHOT/maven-metadata.xml:
  ... status code: 400, reason phrase: Bad Request (400)
```

Note `liquibase-core:-SNAPSHOT` — no version before `-SNAPSHOT`. The step is named "Publish `<commitsha>`-SNAPSHOT package", so the commit-sha variable is not being interpolated. Unrelated to these changes.

(Also for the record: the earlier `Integration Test (oracle)` failure was a testcontainers startup failure on `gvenzl/oracle-xe:18-slim`, not a test failure — it passed on re-run.)

filipelautert:
@SvampX dismissed 2299, your reasoning holds - the page already renders the run log with SQL and table names, so the Throwable-derived distinction is a taint-source artifact rather than a security one.
You are right about the unauthenticated endpoint as well, that one is bigger than this PR.

filipelautert (approved):
The dependency-free escaper is the better call, and it handles the ampersand in the same pass so nothing double-escapes. Built the distribution here to check the packaging side and there is no commons-text left to miss.

SvampX:
Filed the two CI problems separately so they do not clutter this PR's review:

- liquibase/liquibase#7893 — the `sonar` job analyses PRs as the `main` branch (`pull_request_target` not matched by the `"$EVENT_NAME" = "pull_request"` check in build-logic), so no PR analysis is created, PR builds overwrite main's stored analysis, and the gate enforced here is main's rather than this diff's. This is what makes `sonar / sonar` red on this PR.
- liquibase/liquibase#7894 — the publish job builds an empty `-SNAPSHOT` version because Actions suppresses the `latestMergeSha` output as a possible secret. This is what makes `Artifacts - Build & Package` red on this PR.

Both are pre-existing and independent of these changes; the fix for #7893 lives in `liquibase/build-logic` (issues are disabled there, so it is filed here).

rberezen (commented):
Reviewed all 20 files and verified the risky parts end-to-end. The code is solid — LGTM. Two doc mismatches worth cleaning up, and one behavior change worth calling out explicitly.

**✅ XSS fix (the blocker) is complete.** `GenericStatusServlet` — the servlet-integration status page, registered via `web.xml` through `LiquibaseStatusServlet`/`LiquibaseJakartaStatusServlet` and populated by `LiquibaseServletListener` — had three user-controlled sinks and all are closed: the reflected `request.getRequestURI()` is gone (relative `href="?logLevel=…"`), log messages are HTML-escaped, and stack traces no longer reach the response (logged server-side; only an escaped `ClassName: message` is shown). The remaining input, `logLevel`, is validated via `Level.parse()` and never reflected — `getLevelLink` emits the fixed `Level` constant's name, not the user value. New `GenericStatusServletTest` (2) passes.

**✅ The behavior-changing bits check out:**
- `ChangeSet.isCheckSumValid` — the removed `storedCheckSum == null` branch is provably unreachable (`currentMd5Sum` is computed as `null` iff `storedCheckSum` is null, and the `currentMd5Sum == null` check above already returns). Equivalent.
- `getDateLiteral(null)` — the old `else` was reachable only for `null`, so `date.getClass()` was a guaranteed NPE; now a meaningful `IllegalArgumentException` (still a `RuntimeException`). Verified across all 6 `AbstractJdbcDatabaseTest` subclasses.
- The regex `(.*)?` → `(.+)`: I traced it — for `foo()` the old regex matched an empty group and the subsequent `replace("","")` was a no-op yielding `schema.foo()`, while the new regex simply doesn't match and returns the same `schema.foo()`; the with-args case matches identically. Equivalent, and the new reflection unit tests cover it.
- The `hashCode`/`compareTo`/null-guard hardening (`ForeignKey`, `UniqueConstraint`, `PrimaryKey`, `MissingUniqueConstraint`, `UniqueConstraintSnapshot`, `AddUniqueConstraint`, `MissingTable`) is value-identical on the non-null path (`Index.getSchema()` == old `getTable().getSchema()` for a Table-backed index, etc.).

**✅ Real diff regression check:** `generate-changelog` on an H2 schema with a PK, single- and multi-column unique constraints, and an FK produces **byte-identical** output on `main` vs this branch. All changed test classes pass (`DiffToChangeLogTest`, `ChangeSetTest`, `MissingTableChangeGeneratorTest`, `CreateProcedureChangeTest`, the 6 DB tests).

**📝 The PR description is stale in two spots** (code is right, description isn't — worth fixing so it doesn't mislead on merge):
- It still describes using `commons-text`/`StringEscapeUtils.escapeHtml4` and spends a "Things to be aware of" bullet on that dependency + OSGi — but the final code hand-rolls a dependency-free `escapeHtml`. commons-text is **not** introduced to main code; that whole concern is moot.
- It says the regex went `(.*)?` → `(.*)` "provably equivalent", but the code is `(.+)` (the "tighten the parameter regex" commit). Equivalent as shown above, just not what the text says.

**⚠️ One intentional behavior change to flag in the release note.** On that same servlet-integration status page, when a logged record carries an exception, the page used to print the **full stack trace** into the browser; now it prints only the exception type and message (escaped), and sends the trace to the server log instead.

For a `NullPointerException` thrown at `SomeService.run`, the page changes from:
```
SEVERE: update failed
java.lang.NullPointerException: Cannot invoke "Foo.bar()" because "x" is null
    at com.acme.SomeService.run(SomeService.java:42)
    at liquibase.integration.spring.SpringLiquibase.afterPropertiesSet(...)
    ... 38 more
```
to just:
```
SEVERE: update failed
NullPointerException: Cannot invoke "Foo.bar()" because "x" is null
```
(the full trace now lands in the server log as `"Liquibase run error"`). That's the correct call for info-disclosure, but it's a visible change for anyone who used the status page to read traces — worth stating explicitly rather than folding it under "several NPE fixes."

Nice, careful cleanup with real tests rather than blanket null-checks. Thanks @SvampX!

rberezen (approved):

PR #7896, opened 2026-08-11
Fix fast check cache after rollback

### LINKED ISSUE #7549: FastCheckService caches incorrect status after rollback.
### Search first

### Description

Liquibase caches the up-to-date status of your database relative to a particular changelog in a VM-global singleton (FastCheckService) that basically never gets invalidated. This means that any code that rolls back a full-up-to-date database leaves LB in an invalid state (assuming the database is up-to-date when it isn't.

Attempting to update the database again after performing a rollback silently fails to do the right thing.

Since the FastCheckService is a global singleton, the only way around this, short of starting a new Java process, is to call:
`Scope.getCurrentScope().getSingleton(FastCheckService.class).clearCache();`

Which (a) took my a _very_ long time to work out and (b) feels like poking around in the innards of LB in a way that probably isn't terribly robust.

I think it would make more sense if LB did this itself after issuing a rollback.

### Steps To Reproduce

* Start with a LB instance registered against a database that is already full-up-to-date
* Run `liquibase.rollback(2, "");`
* Then run `liquibase.update(2, "");`

### Expected/Desired Behavior

I expect my database to by fully up-to-date again.

### Liquibase Version

5.0.1

### Database Vendor & Version

Postgres 15.4

### Liquibase Integration

Java library

### Liquibase Extensions

_No response_

Ubuntu

### Additional Context

_No response_

hutiefang76:
## Impact

## Description

Fixes #7549.

A successful fast check is cached for the lifetime of the JVM. After a rollback changes DATABASECHANGELOG, that cached result can make the next update report that the database is already up to date and skip the rolled-back change set.

This change invalidates the fast-check cache when a rollback command finishes. The integration test reproduces the full sequence: update, cache the up-to-date result, rollback, verify the history row was removed, update again, and verify the row was restored.

## Release note

Fixed an issue where an update in the same Java process could be skipped after a rollback because Liquibase retained a stale up-to-date result.

## Things to be aware of

The invalidation is in the common rollback command path, so rollback by tag, count, and date receive the same behavior. It also runs from the command's finally block so a partially completed rollback cannot leave a known stale result cached.

## Things to worry about

None known.

## Additional Context

Verified with JDK 17:

- UpdateCommandStepIntegrationTest: 5 tests, 0 failures
- UpdateCommandStepIntegrationTest, RollbackIntegrationTest, and UpdateTestingRollbackCommandsIntegrationTest: 12 tests, 0 failures
- liquibase-standard full test suite: 6,435 tests, 0 failures, 12 skipped

### DIFF
--- a/liquibase-integration-tests/src/test/groovy/liquibase/command/UpdateCommandStepIntegrationTest.groovy
+++ b/liquibase-integration-tests/src/test/groovy/liquibase/command/UpdateCommandStepIntegrationTest.groovy
@@ -65,6 +65,7 @@ class UpdateCommandStepIntegrationTest extends Specification {
         liquibase.rollback(1, contexts, labels)
 
         then:
+        !fastCheckService.isUpToDateFastCheck(null, database, liquibase.getDatabaseChangeLog(), contexts, labels)
         executor.queryForInt(new RawParameterizedSqlStatement(
                 "select count(1) from databasechangelog where id = ? and author = ? and filename = ?",
                 "1", "Liquibase User", "liquibase/update-tests.yml")) == 0
--- a/liquibase-integration-tests/src/test/groovy/liquibase/command/UpdateCommandStepIntegrationTest.groovy
+++ b/liquibase-integration-tests/src/test/groovy/liquibase/command/UpdateCommandStepIntegrationTest.groovy
@@ -49,6 +49,38 @@ class UpdateCommandStepIntegrationTest extends Specification {
         Scope.currentScope.getSingleton(FastCheckService.class).isUpToDateFastCheck(null, h2.getDatabaseFromFactory(), liquibase.getDatabaseChangeLog(), context, label)
     }
 
+    def "rollback invalidates the fast check cache before a subsequent update"() {
+        given:
+        def database = h2.getDatabaseFromFactory()
+        def liquibase = new Liquibase("liquibase/update-tests.yml", new ClassLoaderResourceAccessor(), database)
+        def contexts = new Contexts()
+        def labels = new LabelExpression()
+        def fastCheckService = Scope.currentScope.getSingleton(FastCheckService.class)
+        def executor = Scope.currentScope.getSingleton(ExecutorService.class).getExecutor("jdbc", database)
+        fastCheckService.clearCache()
+        liquibase.update(contexts, labels)
+        assert fastCheckService.isUpToDateFastCheck(null, database, liquibase.getDatabaseChangeLog(), contexts, labels)
+
+        when:
+        liquibase.rollback(1, contexts, labels)
+
+        then:
+        executor.queryForInt(new RawParameterizedSqlStatement(
+                "select count(1) from databasechangelog where id = ? and author = ? and filename = ?",
+                "1", "Liquibase User", "liquibase/update-tests.yml")) == 0
+
+        when:
+        liquibase.update(contexts, labels)
+
+        then:
+        executor.queryForInt(new RawParameterizedSqlStatement(
+                "select count(1) from databasechangelog where id = ? and author = ? and filename = ?",
+                "1", "Liquibase User", "liquibase/update-tests.yml")) == 1
+
+        cleanup:
+        fastCheckService.clearCache()
+    }
+
     def "validate update is successfully executed even when there is by a context mismatch and a non-existent file is referenced in a changeSet"() {
         when:
         def resourceAccessor = new SearchPathResourceAccessor(".,target/test-classes")
--- a/liquibase-standard/src/main/java/liquibase/command/core/AbstractRollbackCommandStep.java
+++ b/liquibase-standard/src/main/java/liquibase/command/core/AbstractRollbackCommandStep.java
@@ -85,6 +85,7 @@ public abstract class AbstractRollbackCommandStep extends AbstractCommandStep {
             handleRollbackException(defineCommandNames()[0][0], rollbackReportParameters);
             throw t;
         } finally {
+            Scope.getCurrentScope().getSingleton(FastCheckService.class).clearCache();
             Scope.getCurrentScope().getMdcManager().remove(MdcKey.CHANGESETS_ROLLED_BACK);
         }
     }


liquibase-integration-tests/src/test/groovy/liquibase/command/UpdateCommandStepIntegrationTest.groovy
```diff
@@ -49,6 +49,36 @@ class UpdateCommandStepIntegrationTest extends Specification {
...
+        then:
+        executor.queryForInt(new RawParameterizedSqlStatement(
+                "select count(1) from databasechangelog where id = ? and author = ?", "1", "Liquibase User")) == 0
+
+        when:
+        liquibase.update(contexts, labels)
+
+        then:
+        executor.queryForInt(new RawParameterizedSqlStatement(
+                "select count(1) from databasechangelog where id = ? and author = ?", "1", "Liquibase User")) == 1
```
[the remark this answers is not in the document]
hutiefang76: Good catch. I tightened both assertions in 6b2dc8c to include `filename = ?` with `liquibase/update-tests.yml`, so the counts are scoped to the complete changeSet identity. I reran `UpdateCommandStepIntegrationTest` on JDK 17: 5 tests, 0 failures/errors; all 7 reactor modules succeeded.

liquibase-integration-tests/src/test/groovy/liquibase/command/UpdateCommandStepIntegrationTest.groovy
```diff
@@ -49,6 +49,38 @@ class UpdateCommandStepIntegrationTest extends Specification {
...
+
+        then:
+        executor.queryForInt(new RawParameterizedSqlStatement(
+                "select count(1) from databasechangelog where id = ? and author = ? and filename = ?",
+                "1", "Liquibase User", "liquibase/update-tests.yml")) == 0
+
+        when:
+        liquibase.update(contexts, labels)
+
+        then:
```
filipelautert, line 76: Hello @hutiefang76 ! This passes on main too, without the fix - I ran it both ways. The row comes back either way, it is not the cache that decides it. Could you assert isUpToDateFastCheck is false right after the rollback? You already have fastCheckService there.

liquibase-standard/src/main/java/liquibase/command/core/AbstractRollbackCommandStep.java
```diff
@@ -85,6 +85,7 @@ protected void doRollback(CommandResultsBuilder resultsBuilder, List<RanChangeSe
             handleRollbackException(defineCommandNames()[0][0], rollbackReportParameters);
             throw t;
         } finally {
+            Scope.getCurrentScope().getSingleton(FastCheckService.class).clearCache();
```
filipelautert, line 88: drop-all leaves the same cache stale, I checked: update then dropAll and isUpToDateFastCheck still returns true with an empty DATABASECHANGELOG. DropAllCommandStep:153 calls resetAll but ChangeLogHistoryService.reset does not touch FastCheckService. Would you like to cover that here too, or keep it for a separate issue?

liquibase-integration-tests/src/test/groovy/liquibase/command/UpdateCommandStepIntegrationTest.groovy
hutiefang76: Confirmed. The previous row-count assertion was only indirect and still passed without the production fix. I added the direct `!fastCheckService.isUpToDateFastCheck(...)` assertion immediately after rollback in 1371f58. The targeted `UpdateCommandStepIntegrationTest` now passes all 5 tests, and all 7 reactor modules succeed.

liquibase-standard/src/main/java/liquibase/command/core/AbstractRollbackCommandStep.java
hutiefang76: I would keep `dropAll` separate. It is the same stale-cache symptom, but it goes through `DropAllCommandStep` rather than the rollback path fixed for #7549. Expanding this PR would mix two command lifecycles and broaden its scope. I will track and reproduce the `dropAll` case separately.

filipelautert (approved):
Ran it both ways here, the new assertion fails without the fix and passes with it. Agreed on keeping dropAll separate, it is a different command lifecycle.

rberezen (commented):
Verified end-to-end — clean, well-scoped fix. LGTM.

**✅ Reproduced #7549 and confirmed the fix** with the in-process (single-JVM) scenario the bug requires — update → rollback → update, via the integration test:
```
base (fix removed):  test FAILS — after rollback the fast-check still reports up-to-date, next update is skipped
this PR:             test PASSES — cache cleared, the rolled-back changeset is re-applied
```
(Confirmed against clean-built shaded `liquibase-core` jars for both variants, so the comparison reflects the actual code.) Deterministic across repeated runs.

**✅ Routing is correct.** `Liquibase.rollback(count/tag/date, …)` executes via `CommandScope(RollbackCountCommandStep/…​)` → the shared instance `AbstractRollbackCommandStep.doRollback`, so the library API the reporter uses hits the fixed path, and rollback by tag, count, and date all get the invalidation. Putting `clearCache()` in the `finally` means a partially-completed rollback can't leave a stale up-to-date result cached either.

**✅ Consistent with existing behavior.** The `FastCheckService.clearCache()` invalidation pattern is already used by `StandardChangeLogHistoryService.clearAllCheckSums()` and `init()`; this closes the rollback gap. Clearing the whole (VM-global) cache is the safe choice — worst case it forces a recompute.

**ℹ️ Minor, non-blocking design thought:** `StandardChangeLogHistoryService.removeFromHistory` is the actual point where a DATABASECHANGELOG row is deleted, so invalidating the cache there would also cover any other caller that removes history rows. The command-step `finally` is a perfectly reasonable, issue-scoped location though — just noting the alternative for the future.

Nice fix on a subtle long-lived-JVM bug. Thanks @hutiefang76!

rberezen (approved):

PR #7897, opened 2026-08-11
fix: carry the index relation down to its columns

### LINKED ISSUE #7895: diff-changelog emits duplicate dropPrimaryKey/addPrimaryKey when a PK-backing index is a changed object (offline)
### Summary
When an offline snapshot diff reports **both** a *Changed Primary Key* and a *Changed Index* for the same PK-backing index, `diff-changelog` generates the PK drop/add pair **twice**, producing a redundant `dropPrimaryKey` → `addPrimaryKey` → `dropPrimaryKey` → `addPrimaryKey` sequence.

This is **pre-existing** and independent of the CCE fix in #7881 (issue #7846). It was previously masked: since #7500 the index→PK delegation threw a `ClassCastException` and the command crashed before the duplicate could surface. #7881 fixes the CCE and re-exposes this older bug. It also reproduces on pre-#7500 builds, where the same delegation ran without crashing.

### Reproduction
Two MariaDB schemas with the same table `t` whose composite PRIMARY KEY differs only in column order — reference `PK(a, b)`, target `PK(b, a)` — captured as offline snapshots:
```
liquibase --search-path=. diff-changelog \
  --url="offline:mariadb?snapshot=tgt.json" \
  --reference-url="offline:mariadb?snapshot=ref.json" \
  --changelog-file=out.xml
```
**Actual (4 changesets):**
```xml
<dropPrimaryKey constraintName="PRIMARY" tableName="t"/>
<addPrimaryKey  columnNames="a, b" constraintName="PRIMARY" tableName="t"/>
<dropPrimaryKey constraintName="PRIMARY" tableName="t"/>
<addPrimaryKey  columnNames="a, b" constraintName="PRIMARY" tableName="t"/>
```
**Expected (2 changesets):** a single `dropPrimaryKey` + `addPrimaryKey`. (The live/online diff of the same two schemas correctly emits one pair.)

The `diff` output confirms the DiffResult contains two changed objects:
```
Changed Index(s):       PRIMARY ... columns changed from '[a, b]' to '[b, a]'
Changed Primary Key(s): PRIMARY ... columns changed from '[pkref.t.a, pkref.t.b]' to '[pktgt.t.b, pktgt.t.a]'
```

### Root cause
`ChangedPrimaryKeyChangeGenerator` runs before `Index` (`runBeforeTypes = {Index, UniqueConstraint}`) and, after emitting its pair, marks the backing index as handled via `control.setAlreadyHandledChanged(new Index().setRelation(pk.getTable()).setColumns(referenceColumns))` so the changed Index is skipped. Offline this match **fails**: the PK's columns are fully qualified (`pkref.t.a`) while the deserialized index's columns are bare (`a`), so `alreadyHandledChanged(...)` doesn't equate them. Dedup misses, and the changed Index independently generates a second PK pair through the index→PK delegation.

### Impact
The generated changelog applies cleanly and yields the correct final PK, so it is not data-corrupting — but the output contains a nonsensical "drop the PK just added, re-add it" duplicate, which is confusing and emits redundant DDL.

### Notes
A proper fix likely normalizes column qualification in the `alreadyHandledChanged` comparison (shared dedup logic that also governs Index/UniqueConstraint de-duplication), so it warrants real `diff`/`diffChangeLog` verification across PK/UC/Index rather than a change localized to the delegation branch.

filipelautert:
## Impact

## Description

Fixes #7895.

Offline diff-changelog emitted dropPrimaryKey/addPrimaryKey twice for the same PK. ChangedPrimaryKeyChangeGenerator marks the backing index as handled by building an Index from the PK's own columns, and that no longer matched the index restored from the snapshot: setColumns() only propagates the relation when the table is already known, but restoring resolves the table reference after the index is built, so the index columns kept a null relation.

setRelation() now propagates it to columns that do not have one yet, which makes the index consistent whatever order the attributes arrive in. Columns that already carry a relation are left alone.

## Testing

MariaDB, two schemas whose composite PK differs only in column order, captured as offline snapshots:

```
offline before: 4 changesets
offline after:  2
online:         2
```

liquibase-standard 6461 green, integration h2 846, postgresql 54, mariadb 41.

## Things to be aware of

Unique constraints were never affected - ChangedUniqueConstraintChangeGenerator passes the real backing index to the dedup rather than building one, so it never saw the mismatch. Verified: the same offline scenario for a unique constraint already emitted 2 changesets before this change.

The first version of this fix called column.getRelation() and broke JsonSnapshotParserTest: while a snapshot is being restored a column's relation can still be an unresolved reference String, which fails the cast. It reads the raw attribute now.

### DIFF
--- a/liquibase-standard/src/main/java/liquibase/structure/core/Index.java
+++ b/liquibase-standard/src/main/java/liquibase/structure/core/Index.java
@@ -92,6 +92,21 @@ public class Index extends AbstractDatabaseObject {
 
     public Index setRelation(Relation relation) {
     	this.setAttribute("table", relation);
+        // setColumns() only propagates the relation when the table is already known. Restoring a snapshot
+        // builds the index first and resolves its table reference afterwards, so the columns would keep a
+        // null relation and stop comparing equal to the same index read from a live database (#7895).
+        if (relation != null) {
+            List<?> columns = getAttribute("columns", List.class);
+            if (columns != null) {
+                for (Object column : columns) {
+                    // Read the raw attribute: while a snapshot is being restored a column's relation can still
+                    // be an unresolved reference String, and getRelation() would fail casting it to Relation.
+                    if ((column instanceof Column) && (((Column) column).getAttribute("relation", Object.class) == null)) {
+                        ((Column) column).setRelation(relation);
+                    }
+                }
+            }
+        }
         return this;
     }
 
--- a/liquibase-standard/src/test/groovy/liquibase/diff/compare/core/IndexComparatorTest.groovy
+++ b/liquibase-standard/src/test/groovy/liquibase/diff/compare/core/IndexComparatorTest.groovy
@@ -4,6 +4,7 @@ import liquibase.diff.compare.DatabaseObjectComparatorFactory
 import liquibase.database.core.MockDatabase
 import liquibase.structure.core.Column
 import liquibase.structure.core.Index
+import liquibase.structure.core.Table
 import spock.lang.Specification
 import spock.lang.Unroll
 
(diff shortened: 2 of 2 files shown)


rberezen (commented):
Verified end-to-end against the #7895 repro and stress-tested ordinary index handling for regressions — clean root-cause fix. LGTM.

**✅ Reproduced #7895 and confirmed the fix** using the issue's offline MariaDB snapshots (composite PK differing only in column order), against `main` *with #7881* (its true parent, now merged):
```
base (main+#7881, no this PR):  4 changesets  (dropPK, addPK, dropPK, addPK)
this PR:                        2 changesets  (dropPK, addPK)   ← matches the online diff
```
A normalized diff shows the **only** change is the removed duplicate pair. Your diagnosis is exactly right: the restored index's columns kept a `null` relation (`setColumns()` only propagates when the table is already known, and restore resolves the table reference afterward), so the synthetic `new Index().setColumns(pkColumns)` in `ChangedPrimaryKeyChangeGenerator` compared as a *different* object and the `alreadyHandledChanged` dedup missed.

**✅ No regression to ordinary index diffing** (I checked this specifically, since my repro only covered a PK-backing index) — base vs PR on an H2 schema with a standalone multi-column index, a `DESC` index, a unique constraint, an FK and PKs:
- `generate-changelog` → **byte-identical**
- self-diff (schema vs itself) → clean on both, no phantom index/column changes
- changed-index diff (swapped index column order, exercising the plain `ChangedIndex` path) → **byte-identical** (`dropIndex`/`createIndex` on both)

This matches the reasoning that a real two-DB diff transforms both sides identically (both restored snapshots), so only the asymmetric synthetic-vs-restored dedup — the actual bug — changes.

**✅ Well-scoped and safe:** only fills columns whose *raw* relation attribute is `null` (avoiding the unresolved-reference cast that broke `JsonSnapshotParserTest` in the first cut — passes now); the `Index(...)` constructor sets relation before columns so that path is untouched; `IndexComparator.hash()` keys on name+relation only, so no collection bucket shifts. **UC confirmed unaffected** — `ChangedUniqueConstraintChangeGenerator` marks the *real* `getBackingIndex()`, not a synthetic one.

**✅ Tests:** `IndexComparatorTest` (21, incl. the 2 new #7895 cases), `DiffToChangeLogTest` (13), `JsonSnapshotParserTest` (2) all green.

Much nicer than guarding it in the generator — fixing it at the snapshot layer keeps the index consistent regardless of attribute order. Thanks for the quick turnaround, @filipelautert!

rberezen (approved):

PR #7907, opened 2026-08-13
fix: ship commons-text and commons-collections4 in the distribution again

### LINKED ISSUE #7904: 5.0.4 distribution drops commons-collections4 + commons-text from internal/lib → NoClassDefFoundError (analytics broken)
## Summary
The **5.0.4** distribution is missing two runtime dependency jars from `internal/lib` that **5.0.3 ships** — `commons-collections4.jar` and `commons-text.jar`. Any command that exercises code linking against them fails at runtime with `NoClassDefFoundError`. Most visibly, **analytics** (`liquibase/analytics/AnalyticsBatch` → `org.apache.commons.collections4.CollectionUtils`) breaks, so `update` / `rollback` / `changelog-sync` all crash.

This is a **regression from 5.0.3** and affects real users, not just tests.

## Evidence
Downloaded and diffed the two release tarballs (`liquibase-5.0.3.tar.gz` vs `liquibase-5.0.4.tar.gz`):

| `internal/lib` | 5.0.3 | 5.0.4 |
|---|---|---|
| jar count | 9 | 7 |
| `commons-collections4.jar` | ✅ present | ❌ **absent** |
| `commons-text.jar` | ✅ present | ❌ **absent** |
| tarball size | 8.7 MB | 7.7 MB |

Runtime failure (seen on the `main` nightly build `Core: liquibase/liquibase/main/11145/c0f504`, which has the same gap):
```
Unexpected error running Liquibase: org/apache/commons/collections4/CollectionUtils
```
`AnalyticsBatch.java:83` calls `CollectionUtils.isNotEmpty(...)`.

## Reproduce
```bash
tar -xzf liquibase-5.0.4.tar.gz
ls internal/lib/commons-collections4.jar   # -> No such file (present in 5.0.3)
./liquibase --url=jdbc:h2:mem:t --changelog-file=cl.xml update
#   -> Unexpected error running Liquibase: org/apache/commons/collections4/CollectionUtils
```

## What's ruled out
- **Not the `commons-collections4` 4.6.0 bump** (#7889) — that landed 2026-08-11, *after* the affected `c0f504` build (2026-08-07); both 5.0.3 and the tested build use 4.5.0.
- **Not the source** — `AnalyticsBatch.java`'s use of `CollectionUtils` is byte-identical in `v5.0.3` and `main`.
- **Not the pom scopes** — `commons-collections4` and `commons-text` are declared `compile` / non-optional in `liquibase-standard` in both refs, and `liquibase-dist/pom.xml` is unchanged for them.

## Likely area
The two jars come transitively via `liquibase-standard` (they're legitimately excluded from `opencsv` in `liquibase-dist`). Something in the **distribution assembly** stopped copying `liquibase-standard`'s `org.apache.commons` runtime deps into `internal/lib` between 5.0.3 and 5.0.4 — likely fallout of the `liquibase-standard`/`liquibase-core` module split or the `liquibase-core` shade / `liquibase-dist` copy wiring. (`liquibase-dist/expected-distribution-contents.txt` still *expects* `internal/lib/commons-collections4.jar`, so the intended layout includes it.)

## Impact
Release-blocking for 5.0.4: analytics is on by default, so common commands crash out of the box.

filipelautert:
## Impact

Fixes #7904

## Description

The 5.0.4 draft ships 7 jars in `internal/lib`, 5.0.3 shipped 9. `commons-collections4` and
`commons-text` are gone, so `AnalyticsBatch` hits `NoClassDefFoundError` and update / rollback /
changelog-sync crash out of the box.

Root cause is the `-am` added to the dist build in build.yml by #7778. Without it the dist
resolved `liquibase-core` from the published repository, and that pom is the shade
dependency-reduced one: `promoteTransitiveDependencies` lifts both commons artifacts to first
level dependencies of `liquibase-core`, which is one of the assembly's includes, so they land in
`internal/lib`. With `-am` the core comes from the reactor with its original pom, where they sit
behind `liquibase-standard`, and the assembly excludes `liquibase-standard` with
`useTransitiveFiltering` on, which drops them too.

Reverting the `-am` is not an option, it is what makes the maven-plugin and extension-testing
artifacts get built at all. So `liquibase-dist` now declares both, the way `liquibase-standard`
and `liquibase-core` already do, and the assembly names them. Declaring alone is not enough:
`useTransitiveFiltering` drops a direct dependency that is not in `<includes>`. Verified the jars
come out at 1.15.0 / 4.6.0, from the properties. Dropping the opencsv exclusions instead would
have pulled opencsv's 1.13.1 / 4.5.0 and lost the version control the exclusions exist for.

Verified by building the current main unchanged, once with `-am` and once without: 7 jars and 9
jars respectively.

## Release note

Fixed a regression where the Liquibase distribution was missing two dependency jars, making
update, rollback and changelog-sync fail with `NoClassDefFoundError`.

## Things to be aware of

`ZipIT` and `TarIT` compare the archives against `expected-distribution-contents.txt` and should
have caught this. They never ran: `liquibase-dist` is `packaging:pom`, so its test sources need an
explicit compile step, and that step lived in the `liquibase-commercial` profile removed by #7247.
Since then `failsafe:integration-test` answers "No tests to run" and the build stays green.
Restored both plugins in the normal build section, since there is no pro profile in this repo
anymore.

The same #7247 swapped the license files without updating the expected contents, so that file
needed a refresh for the tests to pass.

Confirmed the guard-rail works: with the dependencies removed and the plugins in place, both
tests fail; with the fix, both pass.

## Things to worry about

`liquibase-pro/core/liquibase-dist` still carries the old profile, so it has been running these
tests all along. Worth syncing that subtree at some point.

### DIFF
--- a/liquibase-dist/expected-distribution-contents.txt
+++ b/liquibase-dist/expected-distribution-contents.txt
@@ -37,8 +37,9 @@ lib/liquibase_autocomplete.zsh
 lib/liquibase_autocomplete_mac.bash
 licenses/oss/README.txt
 licenses/oss/apache-2.0.txt
-licenses/oss/epl-1.0.txt
 licenses/oss/epl-2.0.txt
+licenses/oss/FSL-1.1-ALv2.txt
+licenses/oss/gpl-2.0-classpath-exception.txt
 licenses/oss/usage-data.txt
 liquibase
 liquibase.bat
--- a/liquibase-dist/pom.xml
+++ b/liquibase-dist/pom.xml
@@ -70,6 +70,18 @@
             </exclusions>
         </dependency>
 
+        <!-- excluded from opencsv above so the version stays under our control -->
+        <dependency>
+            <groupId>org.apache.commons</groupId>
+            <artifactId>commons-text</artifactId>
+            <version>${commons-text.version}</version>
+        </dependency>
+        <dependency>
+            <groupId>org.apache.commons</groupId>
+            <artifactId>commons-collections4</artifactId>
+            <version>${commons-collections4.version}</version>
+        </dependency>
+
         <dependency>
             <groupId>com.h2database</groupId>
             <artifactId>h2</artifactId>
@@ -152,6 +164,37 @@
                         <phase/>
                     </execution>
                 </executions>
+            </plugin>
+            <!-- this module is packaging:pom, so test sources need an explicit compile step -->
+            <plugin>
+                <groupId>org.apache.maven.plugins</groupId>
+                <artifactId>maven-compiler-plugin</artifactId>
+                <executions>
+                    <execution>
+                        <phase>test-compile</phase>
+                        <goals>
+                            <goal>testCompile</goal>
+                        </goals>
+                    </execution>
+                </executions>
+            </plugin>
+            <!-- ZipIT/TarIT compare the generated archives against expected-distribution-contents.txt -->
+            <plugin>
+                <artifactId>maven-failsafe-plugin</artifactId>
+                <executions>
+                    <execution>
+                        <phase>install</phase>
+                        <goals>
+                            <goal>integration-test</goal>
+                            <goal>verify</goal>
+                        </goals>
+                    </execution>
+                </executions>
+                <configuration>
+                    <systemPropertyVariables>
+                        <expected.distribution.contents.filename>${expected-distribution-contents-filename}</expected.distribution.contents.filename>
+                    </systemPropertyVariables>
+                </configuration>
             </plugin>
             <plugin>
                 <groupId>org.liquibase.ext</groupId>
--- a/liquibase-dist/src/main/assembly/component-minimal-deps.xml
+++ b/liquibase-dist/src/main/assembly/component-minimal-deps.xml
@@ -56,6 +56,8 @@
                 <include>com.opencsv:opencsv:</include>
                 <include>org.yaml:snakeyaml:jar:</include>
                 <include>info.picocli:picocli:jar:</include>
+                <include>org.apache.commons:commons-text:jar:</include>
+                <include>org.apache.commons:commons-collections4:jar:</include>
             </includes>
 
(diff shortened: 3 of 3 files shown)


rberezen (commented):
Verified end-to-end — the fix restores the missing jars and, importantly, re-arms the guard that would have caught this. LGTM.

**✅ Reproduced & fixed.** Built the dist on this branch: `internal/lib` now has **9 jars including `commons-collections4.jar` and `commons-text.jar`** (base ships 7 — I'd actually seen exactly this 7-jar `internal/lib` in earlier dist builds off `main`, matching #7904's 5.0.3→5.0.4 diff). Adding the two as explicit `liquibase-dist` deps (versions pinned) plus explicit assembly `<includes>` is the right targeted fix — it re-declares what stopped reaching the assembly once `-am` made `liquibase-core` come from the reactor, without reverting #7778.

**✅ ZipIT/TarIT re-enabled and passing.** With the test-compile + failsafe executions restored, both ITs run again and pass, exact-matching the archive against `expected-distribution-contents.txt`. Since it's an exact match, a future missing/extra dependency fails the build — which is precisely the guard that was silently off since the `liquibase-commercial` profile (and its test-compile step) went away in 5.0. The `expected-distribution-contents.txt` license edits (drop `epl-1.0`, add `FSL-1.1-ALv2` + `gpl-2.0-classpath-exception`) match the actual generated dist.

**ℹ️ Minor, non-blocking:** `ArchiveUtils.getSortedLines` splits on `\n` only, so these ITs fail locally for Windows devs with `core.autocrlf=true` (the checked-out `expected-distribution-contents.txt` gets CRLF, leaving a trailing `\r` on every expected line). CI (LF) is fine — I confirmed the tests pass once the file is LF. A `\r`-strip in `getSortedLines` would make it robust for local Windows runs.

Nice — fixing the regression *and* turning the canary back on. Thanks @filipelautert!

rberezen (approved):

PR #7908, opened 2026-08-13
fix: publish the maven plugin jar that was already built at 0-SNAPSHOT

### LINKED ISSUE #7903: liquibase-maven-plugin branch artifact ships a mismatched descriptor version (0-SNAPSHOT coordinate vs <branch>-SNAPSHOT descriptor) — breaks plugin consumption
**Regression from #7778** (`cdd6f7d76`, 2026-05-30).

Since #7778, branch (`build.yml`) builds produce a `liquibase-maven-plugin` whose embedded descriptor (`META-INF/maven/**/plugin.xml`) carries the build version `<branch>-SNAPSHOT` (e.g. `main-SNAPSHOT`), but the artifact is published/renamed to the **`0-SNAPSHOT`** coordinate — `build.yml` copies `$i-<branch>-SNAPSHOT.jar` → `artifacts/$i-0-SNAPSHOT.jar` without re-stamping the descriptor. Any consumer that runs it *as a plugin* at `0-SNAPSHOT` gets:

```
[ERROR] Invalid plugin descriptor for org.liquibase:liquibase-maven-plugin:0-SNAPSHOT …
        Plugin's descriptor contains the wrong version: main-SNAPSHOT
```

Regular jars (`liquibase-core`, `liquibase-cli`) are unaffected — only a Maven *plugin* validates its descriptor version against the requested coordinate.

### Observed
liquibase-pro-tests community-mode nightly ([run 31634085526](https://github.com/liquibase/liquibase-pro-tests/actions/runs/31634085526)): `liquibase-sdk:install-snapshot` installs the `main` plugin at `0-SNAPSHOT`; every suite then fails in `@BeforeSuite` with "Liquibase Maven version is unknown" (242 failures) because the harness runs the plugin's `:status` goal at `0-SNAPSHOT` and Maven rejects the descriptor.

### Root cause
Neither `build.yml` (the branch copy loop) nor `re-version.sh` re-stamps the plugin descriptor for branch builds. Note `.github/util/re-version.sh` line 80 only rewrites `plugin*.xml` for the literal `0-SNAPSHOT`, not `<branch>-SNAPSHOT`, so the release path has the same latent gap.

### Fix options
- Re-stamp the copied plugin's `plugin.xml` / `pom.xml` / `pom.properties` to `0-SNAPSHOT` in `build.yml` when producing the `0-SNAPSHOT` artifact (mirroring `re-version.sh`), **or**
- have `liquibase-sdk-maven-plugin`'s install goal rewrite a plugin's descriptor when it re-coordinates it to `0-SNAPSHOT`.

### Repro (minimal)
Take any real `liquibase-maven-plugin` jar, set its `plugin.xml` `<version>` to `main-SNAPSHOT`, `mvn install:install-file … -Dversion=0-SNAPSHOT`, then `mvn org.liquibase:liquibase-maven-plugin:0-SNAPSHOT:help` → reproduces the error above. Re-stamping the installed jar's descriptor back to `0-SNAPSHOT` clears it.

filipelautert:
## Impact

Fixes #7903

## Description

`build.yml` publishes the maven plugin under the `0-SNAPSHOT` coordinate but hands over a
renamed `<branch>-SNAPSHOT` jar, so the descriptor inside disagrees with the coordinate. Maven
validates that for plugins and refuses to run: "Plugin's descriptor contains the wrong version:
main-SNAPSHOT". Regular jars never hit this. It is what breaks the liquibase-pro-tests
community-mode nightly, where every suite fails in `@BeforeSuite` with "Liquibase Maven version
is unknown".

No rewriting needed: `temp-artifact`, which this job already downloads, carries
`liquibase-maven-plugin-0-SNAPSHOT.jar` built without `versions:set`, descriptor and all. That is
where the copy used to read from. #7777 switched the loop to the branch-versioned name and #7778
added the module to `-pl`, whose `clean` deletes the prebuilt jar. So this keeps those jars
before the branch-versioned build runs and copies them into `artifacts/`.

`liquibase-cli` rides in the same loop and gets the same treatment. It has no descriptor to
validate, but its manifest and pom now match the coordinate as well.

Also added the missing line in `re-version.sh`: it rewrites `plugin.xml` only for the literal
`0-SNAPSHOT`, while the `pom.xml` line directly above already handles `<branch>-SNAPSHOT`. That
is a separate latent gap on the release path.

## Things to be aware of

Verified against a real `temp-artifact` downloaded from a main run: replayed the sequence
(preserve, clean, copy), installed the resulting jar at `0-SNAPSHOT` and ran
`mvn org.liquibase:liquibase-maven-plugin:0-SNAPSHOT:help`. `BUILD SUCCESS`. The same command
against a renamed branch-versioned jar reproduces the error from the issue.

The first commit in this PR took the other route, rewriting the descriptor with a new script. It
is replaced by the second commit, which reuses the jar Maven already produced correctly.

## Things to worry about

`build.yml` runs on push, so this cannot be exercised end to end from the PR itself.


rberezen (commented):
Reviewed the workflow/script logic (CI-only change, so this is a static review — I can't exercise the release/build pipeline locally). The reasoning is sound and I'd LGTM.

**`build.yml`:** the root problem is that this step's `clean package` rebuilds `liquibase-maven-plugin`/`liquibase-cli` at `<branch>-SNAPSHOT`, so renaming that jar to the `0-SNAPSHOT` coordinate leaves a `<branch>-SNAPSHOT` descriptor inside — which Maven rejects for a *plugin*. Copying the already-correct `0-SNAPSHOT` jars (from the downloaded `temp-artifact`, before the `clean` wipes them) into `prebuilt/` and publishing those is the right move — the descriptor now matches the coordinate. The prerequisite holds: `temp-artifact` is unpacked at "Download Artifacts" and `versions:set` doesn't rebuild those targets, so the `0-SNAPSHOT` jars are present when the `cp` runs.

**`re-version.sh`:** the added `plugin*.xml` `<branch>-SNAPSHOT → $version` rewrite mirrors the `pom.xml` line right above it and closes the latent release-path gap noted in #7903 (previously only the literal `0-SNAPSHOT` plugin descriptor was re-stamped).

One small robustness thought: the `cp prebuilt/$i-0-SNAPSHOT*.jar` copies run under `bash -e`, so if `temp-artifact` ever lacked those jars the step would hard-fail rather than fall back — fine given they're always produced upstream, just flagging it.

Solid fix for a nasty cross-repo breakage (the 242 pro-tests failures). Thanks @filipelautert!

rberezen (approved):

PR #7909, opened 2026-08-13
fix: scope per-database service cleanup so parallel Maven reactor builds no longer corrupt each other's state

pwielgolaski:
## Impact

## Description

Running a multi-module Maven build with `-T` makes `liquibase-maven-plugin` goals fail intermittently, with modules picking up **another module's** database state. On a 20-module reactor this fails on essentially every run.

There is no open issue for this — #6927 had the same user-visible symptom but a different root cause and is already fixed (see *Additional Context*), so I've written this up issue-style below.

### Steps To Reproduce

Full reproduction: **https://github.com/pwielgolaski/liquibase-parallel-build-bug** — 20 trivial sibling modules, offline PostgreSQL connection (no real database required), `updateSQL` bound to `process-test-resources`.

```bash
# build the plugin from source, then:
mvn clean install -T 2C -Dliquibase.version=0-SNAPSHOT
```

### Expected Behavior

Each module runs `updateSQL` against its own database/changelog, independently of what sibling modules are doing concurrently.

### Actual Behavior

One of three failures, depending on interleaving:

```
[ERROR] Failed to execute goal org.liquibase:liquibase-maven-plugin:updateSQL (liquibase-updateSQL) on project module-l:
[ERROR] Error setting up or running Liquibase:
[ERROR] liquibase.exception.LiquibaseException: liquibase.exception.DatabaseException:
        java.nio.file.NoSuchFileException: /path/to/module-a/target/module-a-databasechangelog.csv
```
```
[ERROR] java.lang.ClassCastException: class liquibase.database.OfflineConnection cannot be cast to
        class liquibase.database.jvm.JdbcConnection
```
```
[ERROR] liquibase.exception.DatabaseException: Cannot execute commands against an offline database
```

The first is the clearest tell: `module-l` is reading `module-a`'s changelog CSV. In one captured run, five different modules all failed reaching for `module-a`'s file.

### Root cause

#7877 ("share a single root scope across threads", fixing #6880) made `Scope.rootScope` a single JVM-wide static. That is correct and this PR does not change it — but it changed what `Scope.getSingleton(...)` means:

```java
// Scope.java
public synchronized <T extends SingletonObject> T getSingleton(Class<T> type) {
    if (getParent() != null) {
        return getParent().getSingleton(type);   // always resolves at the root
    }
    ...
    values.put(key, singleton);                  // ...and is stored on the root
}
```

The Maven mojo gives each module its own chain of child scopes (`resourceAccessor`, `classLoader`, `mavenConfigurationProperties`, …), so *values* are isolated per module. But singletons cannot live in a child scope by design — they always resolve to the root. So since #7877, **every concurrently-building module shares one `ExecutorService` and one `ChangeLogHistoryServiceFactory` instance.**

Those registries are keyed by `Database`, so they remain correctly isolated *by construction* — module A's key is never module B's. Isolation is only broken by operations that ignore the key: the blanket `reset()` / `resetAll()`. Before #7877 those were harmless, because each thread had its own root and therefore its own private registries.

Three call sites still assumed that per-thread-private world:

1. **`OfflineConnection.attached()`** registered a database-bound `OfflineChangeLogHistoryService` through the generic priority-based `register()`/`getPlugin()` path. `AbstractPluginFactory.getPlugins()` breaks ties with a `TreeSet` ordered by `(priority, class name)` — identical for every module's instance — so it silently kept only the **first** registration and handed that one instance to every module. → *modules reading each other's CSV.*

2. **`AbstractUpdateCommandStep.cleanUp()` / `AbstractOutputWriterCommandStep.cleanUp()`** called the blanket `ExecutorService.reset()` (`executors.clear()`) on every command completion, erasing every *other* still-running module's `jdbc`/`logging` executor overrides. The next lookup re-resolved a real `JdbcExecutor` in place of the SQL-writing `LoggingExecutor`. → *"Cannot execute commands against an offline database".*

3. **`DatabaseChangelogCommandStep.cleanUp()`** had the same blanket `resetAll()` independently. Combined with fix 1, a wiped module fell back to `StandardChangeLogHistoryService` (whose `supports()` is unconditionally `true`), which has no offline awareness at all. → *the raw `ClassCastException`.*

### The change

Scope each cleanup to the database it belongs to, using the already-`Database`-keyed maps:

- **`ChangeLogHistoryServiceFactory.registerForDatabase(Database, ChangeLogHistoryService)`** — puts directly into the existing `services` map instead of the ambiguous priority resolver. `OfflineConnection.attached()` now uses it.
- **`ChangeLogHistoryServiceFactory.resetDatabase(Database)`** — invalidates one database's service, alongside the existing `resetAll()`.
- The three `cleanUp()` methods call the scoped variants (`ExecutorService.clearExecutor(name, database)` already existed).

No changes to `Scope.java`; #7877 is not reverted or weakened.

## Release note

Fixed multi-module Maven builds running in parallel (`-T`) corrupting each other's Liquibase state. Modules could read another module's changelog file, or fail with `Cannot execute commands against an offline database` or a `ClassCastException` about `OfflineConnection`.

## Things to be aware of

**`resetDatabase()` resets the service in place rather than evicting it.** Eviction is a stronger operation than a finishing command needs, and it destroys something the command doesn't own — the binding between a database and its service. `OfflineConnection` registers a service that `getPlugin()` cannot rebuild, so evicting it would leave the *next* command against that same database falling back to `StandardChangeLogHistoryService` — the same `ClassCastException`, reached by sequential reuse of one database rather than by concurrent modules. `reset()` clears exactly the stale ran-changeset/table-existence state, and is already how `UpdateTestingRollbackCommandStep` and `FastCheckService` invalidate.

**Why the fix isn't in `Scope`.** Making singletons resolve per-child-scope would undo #7877 and reopen #6880 (`ScopeTest` guards this). Since the registries are already `Database`-keyed, the isolation exists — the fix is to stop using the one operation that discards it.

**Validation.** 40/40 clean runs of the reproduction at `-T 2C`, from failing on essentially every run before. `ScopeTest` (added by #7877 for #6880) passes 14/14, confirming this doesn't reopen the split-brain-singleton problem. 1503 tests across `liquibase.command.**`, `liquibase.changelog.**`, `liquibase.lockservice.**`, `liquibase.executor.**` and `liquibase.database.**` pass. Each new test was verified to fail when its corresponding production change is reverted.

**Tests.** Each exercises the fixed call site directly with two distinct `Database` objects, simulating two concurrent modules sharing one JVM. A threaded/looping test was deliberately avoided — see the note on bug 3's rarity below.

| Test | Covers |
| --- | --- |
| `OfflineConnectionTest` | `attached()` registers an isolated service per database (bug 1, via the real production call site) |
| `ChangeLogHistoryServiceFactoryTest` | `registerForDatabase()` / `resetDatabase()` isolate and scope correctly; the binding survives a reset |
| `AbstractUpdateCommandStepTest` | `cleanUp()` touches only its own database's executor and history service (bug 2) |
| `DatabaseChangelogCommandStepTest` | `cleanUp()` invalidates only its own database's service (bug 3) |
| `ExecutorServiceTest` | `clearExecutor()` scoping — the primitive both fixes rely on |

## Things to worry about

1. **`OfflineChangeLogHistoryService` is no longer added to the generic plugin pool.** `registerForDatabase()` bypasses `register()`, so the service is reachable only via the database-keyed lookup. That's the point of the fix, but if any extension relies on discovering it through `getPlugin()`/`findAllInstances()`, that path is gone. I couldn't find such a caller in this repo — worth a maintainer's eye given Pro/extension code I can't see.

2. **The same anti-pattern remains at call sites this reproduction doesn't exercise.** I deliberately did not change them speculatively, but each looks suspect for the same reason:
   - `LockServiceCommandStep.cleanUp()`, `AbstractUpdateCommandStep.cleanUp()`, `StandardLockService.acquireLock()` → `LockServiceFactory.resetAll()`, which nulls the entire singleton. `LockServiceFactory` has the same JVM-wide + `Map<Database, …>` shape but no per-database reset.
   - `DropAllCommandStep.resetServices()`, `Liquibase.resetServices()` → the full `resetAll()`/`reset()` trio.
   - `MarkNextChangesetRanCommandStep` → `ExecutorService.reset()`.

   Happy to follow up in a separate PR, or file an issue tracking them.

3. **Bug 3 is rare enough to fool you.** After fixes 1 and 2, 20/20 runs passed and it looked done — bug 3 only misfires about 1 run in 15 and took a 30-run loop to surface. Anyone re-testing this should not trust a clean run of 10–20.

4. **`deploymentId` is now shared across concurrent modules.** It's generated once during root-scope bootstrap and resolved through the root, so all modules in a `-T` build write the same deployment ID into `DATABASECHANGELOG`. Same underlying cause, not fixed here, no failure reproduced from it — flagging in case it matters for deployment grouping.

## Additional Context

**Relationship to #6927.** This is *not* a reopening. #6927 ("[Regression] Multithread build on maven fails with goal `updateSQL`") had the same user-visible symptom but a different root cause — the shared static `COMMAND_DEFINITIONS` cache — and was correctly fixed by #7876. This reproduction was built while verifying that fix, and shows the "parallel builds corrupt each other" class of problem was not fully resolved.

**Bisect.** The reproduction points at a different commit:

| Build | Result over 5–10 runs |
| --- | --- |
| Released `5.0.3` (pre-#7877) | 0 failures |
| Commit before #7877 (`c0f504ed0`) | 0 failures (8/8 clean) |
| Commit at #7877 (`da65e0e4d`) | fails 2/5 |
| `main` tip (#7877 + #7876) | fails ~8/10 |

**Ruled out.** The original hypothesis was classloader identity across Maven `ClassRealm`s. Instrumentation showed this reproduction uses a *single shared* `ClassRealm` for the plugin across all 20 modules, so none of these failures are cross-realm `instanceof` mismatches — even the `ClassCastException` is a plain wrong-type cast within one loader.

**Related:** #7877 (root cause) · #6880 (what #7877 fixed) · #6927 / #7876 (earlier, separately-fixed instance of this symptom)

### DIFF
--- a/liquibase-standard/src/main/java/liquibase/lockservice/StandardLockService.java
+++ b/liquibase-standard/src/main/java/liquibase/lockservice/StandardLockService.java
@@ -349,7 +349,7 @@ public class StandardLockService implements LockService {
 
                 hasChangeLogLock = true;
 
-                Scope.getCurrentScope().getSingleton(ChangeLogHistoryServiceFactory.class).resetAll();
+                Scope.getCurrentScope().getSingleton(ChangeLogHistoryServiceFactory.class).resetDatabase(database);
                 database.setCanCacheLiquibaseTableInfo(true);
                 return true;
             }
--- a/liquibase-standard/src/main/java/liquibase/changelog/ChangeLogHistoryServiceFactory.java
+++ b/liquibase-standard/src/main/java/liquibase/changelog/ChangeLogHistoryServiceFactory.java
@@ -90,6 +90,10 @@ public class ChangeLogHistoryServiceFactory extends AbstractPluginFactory<Change
             }
     }
 
+    /**
+     * Removes the given service from the plugin registry, so it is no longer a candidate for
+     * {@link #getChangeLogService(Database)} lookups that have not already been cached.
+     */
     public synchronized void unregister(final ChangeLogHistoryService service) {
         removeInstance(service);
     }
@@ -112,10 +116,19 @@ public class ChangeLogHistoryServiceFactory extends AbstractPluginFactory<Change
         }
     }
 
+    /**
+     * Resets and drops every database's {@link ChangeLogHistoryService}, discovered and
+     * {@link #registerForDatabase} registered alike. This is a full JVM-wide teardown; cleanup paths
+     * belonging to a single execution should use {@link #resetDatabase(Database)} instead, so they
+     * don't discard other concurrently running executions' services.
+     */
     public synchronized void resetAll() {
         for (ChangeLogHistoryService changeLogHistoryService : findAllInstances()) {
             changeLogHistoryService.reset();
         }
+        // Services registered for a database never enter the plugin registry, so findAllInstances()
+        // above does not cover them. reset() is idempotent, so overlap between the two is harmless.
+        services.values().forEach(ChangeLogHistoryService::reset);
         services.clear();
         // unregister all self-registered
         explicitRegistered.forEach(this::removeInstance);
--- a/liquibase-standard/src/main/java/liquibase/changelog/ChangeLogHistoryServiceFactory.java
+++ b/liquibase-standard/src/main/java/liquibase/changelog/ChangeLogHistoryServiceFactory.java
@@ -95,13 +95,21 @@ public class ChangeLogHistoryServiceFactory extends AbstractPluginFactory<Change
     }
 
     /**
-     * Clears the cached {@link ChangeLogHistoryService} for just the given database, as opposed to
-     * {@link #resetAll()} which clears every database's entry. Since this factory is a single JVM-wide
-     * singleton, resetAll() called from one execution's cleanup would otherwise drop other concurrently
-     * running executions' (different database's) cached services too.
+     * Invalidates the {@link ChangeLogHistoryService} cached for just the given database, as opposed to
+     * {@link #resetAll()} which invalidates and drops every database's entry. Since this factory is a
+     * single JVM-wide singleton, resetAll() called from one execution's cleanup would otherwise discard
+     * other concurrently running executions' (different database's) services too.
+     * <p>
+     * The entry is reset in place rather than removed: the service is bound to the database for that
+     * database's lifetime - {@link liquibase.database.OfflineConnection} registers one that
+     * {@link #getPlugin} cannot rebuild - whereas the stale data a finishing command wants dropped is
+     * exactly what {@link ChangeLogHistoryService#reset()} clears.
      */
     public synchronized void resetDatabase(Database database) {
-        services.remove(database);
+        ChangeLogHistoryService service = services.get(database);
+        if (service != null) {
+            service.reset();
+        }
     }
 
     public synchronized void resetAll() {
--- a/liquibase-standard/src/main/java/liquibase/changelog/ChangeLogHistoryServiceFactory.java
+++ b/liquibase-standard/src/main/java/liquibase/changelog/ChangeLogHistoryServiceFactory.java
@@ -48,6 +48,17 @@ public class ChangeLogHistoryServiceFactory extends AbstractPluginFactory<Change
     }
 
 
+    /**
+     * Registers a {@link ChangeLogHistoryService} that is already fully configured for the given database
+     * (e.g. an offline/simulated connection's history service), bypassing {@link #getPlugin(Object...)}'s
+     * generic type/priority-based lookup. That lookup can't distinguish between multiple pre-configured
+     * instances of the same class and priority registered concurrently for different databases - it only
+     * ever kept one, handing it out for every database.
+     */
+    public synchronized void registerForDatabase(Database database, ChangeLogHistoryService service) {
+        services.put(database, service);
+    }
+
     public synchronized ChangeLogHistoryService getChangeLogService(Database database) {
             if (services.containsKey(database)) {
                 return services.get(database);
@@ -81,6 +92,16 @@ public class ChangeLogHistoryServiceFactory extends AbstractPluginFactory<Change
 
     public synchronized void unregister(final ChangeLogHistoryService service) {
         removeInstance(service);
+    }
+
+    /**
+     * Clears the cached {@link ChangeLogHistoryService} for just the given database, as opposed to
+     * {@link #resetAll()} which clears every database's entry. Since this factory is a single JVM-wide
+     * singleton, resetAll() called from one execution's cleanup would otherwise drop other concurrently
+     * running executions' (different database's) cached services too.
+     */
+    public synchronized void resetDatabase(Database database) {
+        services.remove(database);
     }
 
     public synchronized void resetAll() {
--- a/liquibase-standard/src/main/java/liquibase/changelog/StandardChangeLogHistoryService.java
+++ b/liquibase-standard/src/main/java/liquibase/changelog/StandardChangeLogHistoryService.java
@@ -80,6 +80,11 @@ public class StandardChangeLogHistoryService extends AbstractChangeLogHistorySer
         this.ranChangeSetList = null;
         this.serviceInitialized = false;
         this.hasDatabaseChangeLogTable = null;
+        // Cached MAX(ORDEREXECUTED), incremented in memory by getNextSequenceValue(). Unlike
+        // databaseChecksumsCompatible it is not recomputed by init(), so it has to be dropped here or
+        // it would survive a reset and hand out sequence values based on a stale read - notably after
+        // acquiring the changelog lock, where another process may have written in the meantime.
+        this.lastChangeSetSequenceValue = null;
     }
 
     public boolean hasDatabaseChangeLogTable() {
--- a/liquibase-standard/src/main/java/liquibase/command/core/AbstractUpdateCommandStep.java
+++ b/liquibase-standard/src/main/java/liquibase/command/core/AbstractUpdateCommandStep.java
@@ -209,8 +209,14 @@ public abstract class AbstractUpdateCommandStep extends AbstractCommandStep impl
     public void cleanUp(CommandResultsBuilder resultsBuilder) {
         isDBLocked.remove();
         LockServiceFactory.getInstance().resetAll();
-        Scope.getCurrentScope().getSingleton(ChangeLogHistoryServiceFactory.class).resetAll();
-        Scope.getCurrentScope().getSingleton(ExecutorService.class).reset();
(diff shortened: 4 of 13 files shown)


filipelautert (commented):
Nothing else calls register().

Your change makes the services map the only home of the offline service, but resetAll() still only iterates findAllInstances(), so it drops that service without ever calling reset() on it. Should not it cover services.values() as well?

StandardLockService:352 calls resetAll() on every successful acquireLock, so it is a hot path. Your repro misses it because OfflineLockService serves offline connections, but a parallel build against real databases hits it every time.

pwielgolaski:
@filipelautert 
Both good catches, fixed in `bf28b82d8`.

**`resetAll()` not covering `services.values()`** — you're right, and it mattered more than it looks: since `registerForDatabase()` is the only home for an `OfflineConnection`'s service, that service was dropped on every full teardown without ever being invalidated. It's now reset before the map is cleared. A service reachable from both collections gets `reset()` twice, which is harmless since it's idempotent. Added a test to `ChangeLogHistoryServiceFactoryTest` that fails without the change.

**`StandardLockService:352`** — scoped to `resetDatabase(database)`. That was the sharpest form of the problem: on every successful `acquireLock()` it discarded every concurrently-running module's service, and per the gap above, dropped the offline ones without resetting them. It now resets in place the one service whose lock state actually changed, which also matches the intent of the comment just below it.

Re-ran the reproduction after the lock change specifically because of the hot path: 40/40 clean. 1504 tests pass across the touched packages, `ScopeTest` included.

liquibase-standard/src/main/java/liquibase/lockservice/StandardLockService.java
```diff
@@ -349,7 +349,7 @@ public boolean acquireLock() throws LockException {
 
                 hasChangeLogLock = true;
 
-                Scope.getCurrentScope().getSingleton(ChangeLogHistoryServiceFactory.class).resetAll();
+                Scope.getCurrentScope().getSingleton(ChangeLogHistoryServiceFactory.class).resetDatabase(database);
```
filipelautert, line 352: My ask on this one sent you the wrong way. reset() leaves lastChangeSetSequenceValue set, so the cached MAX(ORDEREXECUTED) survives here, and acquireLock is the one place it must not. On h2, after another process writes ORDEREXECUTED=20, this gives 12 where resetAll gave 21. Should not reset() clear it too?

filipelautert (commented):
resetAll covering the registered ones looks right.

pwielgolaski:
@filipelautert 
Yes — `reset()` should clear it. Fixed in https://github.com/liquibase/liquibase/pull/7909/changes/0a51e48b2b5bbc97803f84cbf0a020ec771f1f35

This was a consequence of scoping line 352: while it called `resetAll()` the service was evicted outright, so the next lookup built a fresh instance and re-read `MAX(ORDEREXECUTED)`. Resetting in place let the instance survive and keep serving values derived from the stale read — and as you say, `acquireLock()` is precisely where that's wrong, since taking the lock is the moment another process may have written while we didn't hold it. `reset()` now clears `lastChangeSetSequenceValue` too, which also makes it an actual full invalidation rather than a partial one.

`databaseChecksumsCompatible` doesn't need the same treatment — `reset()` already clears `serviceInitialized`, so `init()` recomputes it.

Added `StandardChangeLogHistoryServiceTest` covering it; it fails without the change. 

One housekeeping note: I rebased onto latest `main` and force-pushed, so the two merge commits from the earlier branch updates are gone and the history is linear. Your inline comments may show as outdated as a result — the code they point at is unchanged apart from the fixes above.

filipelautert (approved):
Sequence value comes out right both ways now.

rberezen (approved):

rberezen (commented):
Reviewed the mechanism and design end-to-end — this is a careful, correctly-scoped fix for a real concurrency bug, and an exemplary writeup. LGTM.

**✅ Root cause & approach are right.** Since #7877 made `rootScope` a JVM-wide static, `getSingleton(...)` always resolves at the root, so all concurrently-building modules share one `ExecutorService` / `ChangeLogHistoryServiceFactory`. Those registries are `Database`-keyed and therefore isolated *by construction* — the only thing that breaks isolation is the key-ignoring blanket `reset()`/`resetAll()`. Scoping each `cleanUp()` to its own `Database` (via the already-keyed maps) is exactly the minimal correct fix, and it leaves #7877 untouched — confirmed `ScopeTest` still passes **14/14**, so #6880 is not reopened.

**✅ The `AbstractOutputWriterCommandStep.database` field is safe.** I checked the concern that stashing `database` in an instance field could race under `-T`: `CommandFactory.getCommandDefinition` returns a fresh copy whose copy-constructor reinstantiates every step via `newInstance()`, so pipeline steps (and this field, like the pre-existing `outputStreamWriter`) are per-execution. And clearing both `"jdbc"` and `"logging"` in `cleanUp()` is exactly symmetric with `run()` setting the `LoggingExecutor` under both keys.

**✅ The `reset()` → `lastChangeSetSequenceValue = null` addition is necessary and behavior-preserving.** Since `resetDatabase()` now resets in place instead of dropping the service (old `resetAll()` cleared the map), the cached `MAX(ORDEREXECUTED)` had to be cleared explicitly or it'd survive a post-lock reset and hand out stale sequence values. Verified end-to-end: two sequential `update`s produce clean `ORDEREXECUTED = 1,2,3,4,5`.

**✅ Tests genuinely guard the fix** — reverting `OfflineConnection.attached()` back to `register()` makes `OfflineConnectionTest` fail, and the changelog/lockservice/executor suites all pass on the branch.

A couple of small notes, all of which you already flagged thoughtfully:
- **Reset-in-place keeps the `services` entry**, so the shared map isn't cleared by scoped cleanup. In practice it's bounded — the `Liquibase` facade still calls `resetServices()`→`resetAll()` in its `finally`, `Database.equals` keys by connection, and CLI/Maven are finite — so only a long-running, command-framework-direct caller creating unbounded distinct connections would accumulate. Fine given the eviction-vs-reset rationale you documented.
- The **same anti-pattern at the other call sites** (`LockServiceFactory.resetAll()`, `DropAllCommandStep`, `MarkNextChangesetRan`) and the **shared `deploymentId`** are worth a tracking issue/follow-up as you suggested.

One caveat on my side: I verified the *mechanism* (scoped cleanup correctness, no regressions, ORDEREXECUTED integrity) rather than re-running the full 20-module `-T 2C` reproduction — I'm trusting your 40/40 there. Really nice diagnosis and the "each test fails when its production change is reverted" discipline is exactly right. Thanks @pwielgolaski!

PR #7911, opened 2026-08-13
fix: publish the branch-named artifacts at 0-SNAPSHOT too

### LINKED ISSUE #7903: liquibase-maven-plugin branch artifact ships a mismatched descriptor version (0-SNAPSHOT coordinate vs <branch>-SNAPSHOT descriptor) — breaks plugin consumption

filipelautert:
## Impact

Fixes #7903

## Description

#7908 fixed the `artifacts/` directory but not its sibling `artifacts-named/`, which is what
`liquibase-sdk:install-snapshot` actually consumes.

That mojo runs `install-file` with the file and nothing else, so Maven takes the coordinates
from the pom embedded in each jar. Two mismatches came out of that:

- the maven plugin's `plugin.xml` said `<branch>-SNAPSHOT` while `install-file` landed the jar at
  `0-SNAPSHOT` (its embedded `pom.xml` says `0-SNAPSHOT`), so Maven rejected the descriptor
- `liquibase-core`'s embedded pom says `<branch>-SNAPSHOT`, so core landed at `<branch>-SNAPSHOT`
  while the plugin pom asks for `liquibase-core:0-SNAPSHOT`, so the runtime realm did not resolve

Taking all four jars from `prebuilt/` gives them one consistent `0-SNAPSHOT` across `pom.xml`,
`pom.properties` and `plugin.xml`. The file name still carries the branch, which is what
`install-snapshot` looks up.

## Things to be aware of

Verified against the real `liquibase-artifacts-main` from run `31736306343`, replaying the
workflow over that run's `temp-artifact`, installing with `install-file` the way the mojo does,
into an isolated local repository, then running the goal the harnesses actually use:

```
before   Invalid plugin descriptor for org.liquibase:liquibase-maven-plugin:0-SNAPSHOT,
         Plugin's descriptor contains the wrong version: main-SNAPSHOT
after    The database URL has not been specified
```

The second one is the mojo running, so the descriptor validated and the realm resolved.

Where each jar lands with `install-file`:

```
before   liquibase-core/main-SNAPSHOT      liquibase-maven-plugin/0-SNAPSHOT
after    liquibase-core/0-SNAPSHOT         liquibase-maven-plugin/0-SNAPSHOT
```

## Things to worry about

`build.yml` runs on push, so this cannot be exercised end to end from the PR itself.

The jars published today are internally inconsistent: the plugin's `pom.xml` says `0-SNAPSHOT`
while its `pom.properties` and `plugin.xml` say `<branch>-SNAPSHOT`. This PR sidesteps that by
publishing the untouched 0-SNAPSHOT build rather than reconciling the branch-versioned one.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -172,8 +172,9 @@ jobs:
 
           # keep temp-artifact's 0-SNAPSHOT jars, the build below cleans these modules
           mkdir -p prebuilt
-          cp liquibase-maven-plugin/target/liquibase-maven-plugin-0-SNAPSHOT*.jar prebuilt/
-          cp liquibase-cli/target/liquibase-cli-0-SNAPSHOT*.jar prebuilt/
+          for i in 'liquibase-maven-plugin' 'liquibase-cli' 'liquibase-extension-testing'; do
+            cp $i/target/$i-0-SNAPSHOT*.jar prebuilt/
+          done
 
           ./mvnw -B -pl liquibase-dist,liquibase-maven-plugin,liquibase-extension-testing -am source:jar clean package failsafe:integration-test failsafe:verify \
           "-Dbuild.repository.owner=liquibase" \
@@ -220,11 +221,15 @@ jobs:
           mkdir artifacts-named
 
           cp liquibase-dist/target/liquibase-${{ needs.setup.outputs.thisBranchName }}-SNAPSHOT.tar.gz artifacts-named/liquibase-${{ needs.setup.outputs.thisBranchName }}.tar.gz
-          cp liquibase-extension-testing/target/liquibase-extension-testing-${{ needs.setup.outputs.thisBranchName }}-SNAPSHOT-deps.jar artifacts-named/liquibase-extension-testing-${{ needs.setup.outputs.thisBranchName }}-deps.jar
 
-          for i in 'liquibase-core' 'liquibase-maven-plugin' 'liquibase-cli' 'liquibase-extension-testing'; do
-            cp $i/target/$i-${{ needs.setup.outputs.thisBranchName }}-SNAPSHOT.jar artifacts-named/$i-${{ needs.setup.outputs.thisBranchName }}.jar
+          # liquibase-sdk:install-snapshot installs everything from here at 0-SNAPSHOT, so these
+          # carry that version inside even though the file name says the branch
+          cp prebuilt/liquibase-extension-testing-0-SNAPSHOT-deps.jar artifacts-named/liquibase-extension-testing-${{ needs.setup.outputs.thisBranchName }}-deps.jar
+          for i in 'liquibase-maven-plugin' 'liquibase-cli' 'liquibase-extension-testing'; do
+            cp prebuilt/$i-0-SNAPSHOT.jar artifacts-named/$i-${{ needs.setup.outputs.thisBranchName }}.jar
           done
+
+          cp liquibase-core/target/liquibase-core-${{ needs.setup.outputs.thisBranchName }}-SNAPSHOT.jar artifacts-named/liquibase-core-${{ needs.setup.outputs.thisBranchName }}.jar
 
       - name: Archive Packages
         uses: actions/upload-artifact@043fb46d1a93c77aae656e7c1c64a875d1fc6a0a # v7.0.1
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -172,7 +172,7 @@ jobs:
 
           # keep temp-artifact's 0-SNAPSHOT jars, the build below cleans these modules
           mkdir -p prebuilt
-          for i in 'liquibase-maven-plugin' 'liquibase-cli' 'liquibase-extension-testing'; do
+          for i in 'liquibase-core' 'liquibase-maven-plugin' 'liquibase-cli' 'liquibase-extension-testing'; do
             cp $i/target/$i-0-SNAPSHOT*.jar prebuilt/
           done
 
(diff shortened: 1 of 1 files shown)


rberezen:
Nice — right root cause (`#7908` fixed `artifacts/` but not the `artifacts-named/` set that `install-snapshot` actually consumes), and reusing the already-built `prebuilt/` 0-SNAPSHOT jars is a clean way to get the descriptor right. Excluding shaded `liquibase-core` is correct.

One thing worth verifying before closing #7903, because the downstream stopgap had to fix **two** mismatches and this looks like it only clears the first:

1. ✅ plugin descriptor `<version>` `main-SNAPSHOT` → `0-SNAPSHOT` (this PR).
2. ❓ the installed plugin POM declares `liquibase-core:0-SNAPSHOT` (scope `runtime`), but `install-snapshot` lands **core at `<branch>-SNAPSHOT`**. From a real run:

   ```
   .m2/…/liquibase-maven-plugin/0-SNAPSHOT/…        ← plugin at 0-SNAPSHOT
   .m2/…/liquibase-core/main-SNAPSHOT/…             ← core at main-SNAPSHOT
   plugin pom: <artifactId>liquibase-core</artifactId><version>0-SNAPSHOT</version><scope>runtime</scope>
   → Could not find artifact org.liquibase:liquibase-core:jar:0-SNAPSHOT
   ```

This PR doesn't change core's install coordinate or the plugin pom's core-dep version, so that second mismatch looks like it survives. The `:0-SNAPSHOT:help` check won't surface it if `liquibase-core:0-SNAPSHOT` happens to be in the local repo already — but the goal harnesses actually invoke is `:0-SNAPSHOT:status`, which resolves the plugin's runtime realm (core).

Suggest validating a full `install-snapshot` round-trip + `mvn org.liquibase:liquibase-maven-plugin:0-SNAPSHOT:status`. If core doesn't resolve, this flips the error from "wrong descriptor version" to "could not find liquibase-core:0-SNAPSHOT" rather than fully clearing it — either core needs to land at `0-SNAPSHOT` too, or the plugin's core dependency needs to match core's install coordinate.

filipelautert:
Right on both counts, my :help check was masked by a stale liquibase-core:0-SNAPSHOT in my local repo.

The reason both mismatches exist is that installToMavenCache calls install-file with only the file, so maven takes the coordinates from each jar's embedded pom. Today's plugin jar disagrees with itself: pom.xml says 0-SNAPSHOT, pom.properties and plugin.xml say main-SNAPSHOT. Core's pom says main-SNAPSHOT, which is why it lands there.

Pushed core into the same prebuilt path. Replayed it into an isolated repo and ran :status in a real project:

```
before   Invalid plugin descriptor ... wrong version: main-SNAPSHOT
after    The database URL has not been specified
```

rberezen:
`4dbfcb67d` closes the gap 👍 — verified the mechanism against the real `liquibase-core-main.jar` (run `31736306343`):

- It **does** carry `META-INF/maven/org.liquibase/liquibase-core/pom.xml` with `<version>main-SNAPSHOT</version>` (so the "shaded without META-INF/maven" note isn't quite accurate — that pom is exactly what `install-file` reads for the coordinate, which is why core landed at `main-SNAPSHOT`).
- Sourcing core from `prebuilt/` (embedded pom = `0-SNAPSHOT`) makes `install-file` land it at `0-SNAPSHOT`, matching the plugin pom's `liquibase-core:0-SNAPSHOT` runtime dep. Both failure modes are now covered.

Only thing left is the end-to-end you already flagged (build.yml runs on push): a post-merge `install-snapshot` + `mvn org.liquibase:liquibase-maven-plugin:0-SNAPSHOT:status` to confirm the realm resolves. Happy to verify from the pro-tests side once a fixed build lands.

rberezen (approved):

PR #7918, opened 2026-08-17
Fix ExceptionDetails.exception missing leading newline; close unclosed PrintWriter

wwillard7800:
## Summary

- `ExceptionDetails(Throwable, String)` built the `exception` field via `printStackTrace()` with no preceding `println()`, so the stack trace started immediately on the first character with no leading newline.
- `StructuredLogFormatter.formatThrowable()` and `LogbackStructuredJsonEncoder.formatThrowable()` both call `println()` first, so all top-level `"exception"` fields in structured JSON output had a leading `\n` while `exceptionDetails.exception` (populated from this class via MDC) did not.
- The `PrintWriter` was also left unclosed.

**Fix:** wrap `PrintWriter` in try-with-resources and call `printWriter.println()` before `printStackTrace()`, matching the behaviour of the two formatters.

## Test plan

- [ ] Verify `ExceptionDetailsSpec` (new unit tests in liquibase-pro) passes
- [ ] Confirm structured JSON log output: `exceptionDetails.exception` begins with `\n` consistently with top-level `exception` field

Relates to liquibase/liquibase-pro#4692

### DIFF
--- a/liquibase-standard/src/main/java/liquibase/logging/mdc/customobjects/ExceptionDetails.java
+++ b/liquibase-standard/src/main/java/liquibase/logging/mdc/customobjects/ExceptionDetails.java
@@ -38,8 +38,12 @@ public class ExceptionDetails implements CustomMdcObject {
             this.primaryExceptionReason = primaryException.getMessage();
             this.primaryExceptionSource = source;
             StringWriter stringWriter = new StringWriter();
-            PrintWriter printWriter = new PrintWriter(stringWriter);
-            exception.printStackTrace(printWriter);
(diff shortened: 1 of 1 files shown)


wwillard7800:
Superseded by a Pro-only approach in liquibase/liquibase-pro#4692 (commit `7196bff12e`). Reviewers flagged that the OSS-subtree edit would generate conflicts on future syncs, so the leading-newline transformation now happens in a Pro-side encoder wrapper rather than in `ExceptionDetails` itself. No OSS change needed.

PR #7919, opened 2026-08-17
Fix: skip leading comment/blank lines before reading the CSV header in loadData

### LINKED ISSUE #4434: loadUpdateData comment on first line detected as header.
### Search first

### Description

When I add a comment with a default comment character `#` on first line, seems like it detect that comment as header:
```
Migration failed for changeset db/changelog/load_data-changelog.xml::loadData::ronny:
     Reason: liquibase.exception.UnexpectedLiquibaseException: CSV file db/data/nationality.csv Line 2 has 2 values defined, Header has 1. Numbers MUST be equal (check for unquoted string with embedded commas)
``` 

### Steps To Reproduce

Changelog file:

```xml 
  <changeSet id="loadData" author="ronny" runOnChange="true">
        <loadUpdateData
            tableName="nationality"
            file="db/data/nationality.csv"
            primaryKey="id"
        />
    </changeSet>
```

CSV file:

```csv
#Nationality table from https://gist.github.com/marijn/274449 (194 records)
id,nationality_name
1,'Afhan'
2,'Albanian'
```

### Expected/Desired Behavior

I was able to fix it by moving the comment on 2nd line, but I want to be able to put comment on first line without error.

### Liquibase Version

4,22.0

### Database Vendor & Version

MySQL 8.0

### Liquibase Integration

spring boot, maven

### Liquibase Extensions

_No response_

Windows 10

### Additional Context

_No response_

om7057:
## Summary
Fixes #4434 - a CSV file starting with a comment line (default `#`) or a blank line has that line misparsed as the header, so every data row after it fails with a column-count mismatch.

`generateStatements()` read the first line unconditionally as the header, even though comment/blank lines were already being skipped for data rows via `commentLineStartsWith`/`isLineCommented`. This applies the same skip check before the header is read too, via a shared `isSkippableLine()` helper (also used by the data-row loop, replacing the inline duplicate). `loadUpdateData` gets the fix too since it extends `LoadDataChange`.

## Test plan
- [x] New test + CSV fixture reproducing the exact reported scenario (leading `#` comment before the header)
- [x] Verified the new test fails against the pre-fix code (misparses the comment as header, 3 statements instead of 2) and passes with the fix
- [x] `LoadDataChangeTest`/`LoadUpdateDataChangeTest` suites pass (54 + 19 tests)

### DIFF
--- a/liquibase-standard/src/main/java/liquibase/change/core/LoadDataChange.java
+++ b/liquibase-standard/src/main/java/liquibase/change/core/LoadDataChange.java
@@ -243,7 +243,14 @@ public class LoadDataChange extends AbstractTableChange implements ChangeWithCol
                 throw new UnexpectedLiquibaseException("Unable to read file " + this.getFile());
             }
 
-            String[] headers = reader.readNext();
+            boolean isCommentingEnabled = StringUtils.isNotEmpty(commentLineStartsWith);
+
+            // Skip any blank or commented lines preceding the header, so a leading comment (e.g. a "#"
+            // line describing the file) isn't mistaken for the header row -- see #4434.
+            String[] headers;
+            do {
+                headers = reader.readNext();
+            } while ((headers != null) && isSkippableLine(headers, isCommentingEnabled));
             if (headers == null) {
                 throw new UnexpectedLiquibaseException("Data file " + getFile() + " was empty");
             }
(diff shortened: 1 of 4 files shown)


om7057:
@filipelautert @wwillard7800 just a small reminder on this one!

filipelautert (approved):
Hello @om7057 ! Logic is good + tests, and it even keeps the reported line numbers right. Thanks!

PR #7921, opened 2026-08-18
[TECHOPS-1086] Use build-logic's scan persist/matrix scripts

jandroav:
> **Merge after [liquibase/build-logic#678](https://github.com/liquibase/build-logic/pull/678).** This PR deletes the local copies and points at the shared ones; merging first breaks the scan persist step.

## Why

`scripts/generate-dockerhub-matrix.sh` and `scripts/persist-scan-results.sh` existed as a private copy here and another in the sibling repo, and the copies had drifted — the defense-in-depth VEX filter (TECHOPS-408, 250 lines) only ever landed in one of them. Change Intelligence needs the same pair to publish its scans ([TECHOPS-1086](https://datical.atlassian.net/browse/TECHOPS-1086)), so they moved to build-logic instead of being copied a third time.

## What

- Both scripts deleted here; the workflows call `build-logic/scripts/vulnerability-scanning/...`.
- The persist jobs gain a build-logic checkout (the published-images job in this repo did not have one).
- The image family this repo owns is now an argument at the `generate-matrix` call site rather than a hardcoded list inside the script. The one-repo-one-family rule (TECHOPS-408) is unchanged; it is just visible where someone editing the workflow will see it.
- The scan job reads the matrix's new `ref` field. For a Docker Hub image `ref` and `image` are the same value; they differ only for registries that need a host prefix.

**Behaviour is unchanged for this repo.** The shared script is the liquibase-pro copy, so this repo also picks up the VEX filter it never had — a no-op without a `vex/assessments.yaml`, and it no longer warns about the absence.

## Note for whoever tidies up next

`liquibase/docker` holds a third copy of both scripts. That repo is **archived** (last push 2026-06-12), so its workflows cannot run and its copies are frozen. Deliberately left alone.

[TECHOPS-1086]: https://datical.atlassian.net/browse/TECHOPS-1086?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


.github/workflows/trivy-scan-published-images.yml
```diff
@@ -97,4 +112,5 @@ jobs:
         run: |
           git config user.name "github-actions[bot]"
           git config user.email "github-actions[bot]@users.noreply.github.com"
-          scripts/persist-scan-results.sh scan-artifacts
+          chmod +x build-logic/scripts/vulnerability-scanning/persist-scan-results.sh
+          build-logic/scripts/vulnerability-scanning/persist-scan-results.sh scan-artifacts
```
sayaliM0412: **[medium]** Nothing in this repoint is exercised by CI before merge.

This workflow is `schedule` + `workflow_dispatch` only, and the `persist-main-latest` job in `docker-scan.yml` is gated on `github.event_name != 'pull_request' && github.ref == 'refs/heads/main'`. So the build-logic checkout, the `chmod`, the new call-site image argument and the `ref` field all execute for the first time post-merge, on a job that pushes to `scan-results`.

`workflow_dispatch` runs the workflow file from the selected branch, so this is testable now: dispatch this workflow from `techops-1086-shared-scan-scripts` and check the resulting `scan-results` commit before merging. Worth doing given the shared script is also picking up the VEX filter this repo never had.

sayaliM0412 (commented):
Review for [TECHOPS-1086](https://datical.atlassian.net/browse/TECHOPS-1086). One finding, inline.

The repoint itself reads correct: both persist jobs now check out build-logic, the community family is declared at the `generate-matrix` call site, and the scan job reads the new `ref` field. `VEX_ASSESSMENTS_REQUIRED` is correctly left unset here (no `vex/assessments.yaml` in this repo), so the shared script's filter is a no-op and no longer warns, which matches the PR body.

The `mergefreeze` red check is a Community 5.0.4 freeze, not this PR.

.github/workflows/trivy-scan-published-images.yml
jandroav: Agreed, and it is the plan: once build-logic#678 merges, dispatch this workflow from `techops-1086-shared-scan-scripts` and inspect the resulting `scan-results` commit before merging. It cannot be done before that, since the branch's workflow would check out build-logic@main, which does not have the shared scripts yet.

In the meantime the failure mode is no longer cryptic: each call site checks the script exists and says what to merge, instead of dying with a bare "no such file" inside a job that pushes to scan-results.

.github/workflows/trivy-scan-published-images.yml
```diff
@@ -39,13 +39,22 @@ jobs:
     env:
       MAX_TAGS: ${{ github.event.inputs.max_tags_to_scan || '10' }}
     steps:
-      - name: Checkout code
+      - name: Checkout build-logic
         uses: actions/checkout@3d3c42e5aac5ba805825da76410c181273ba90b1 # v7.0.1
+        with:
+          repository: liquibase/build-logic
+          path: build-logic
```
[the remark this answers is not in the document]
jandroav: Partly. The out-of-order merge is real, and each of the three checkouts now verifies the script is present and fails with an explicit "merge build-logic#678 first" instead of an opaque shell error.

I have not pinned to a SHA. Promoting these scripts is what lets one fix reach all three repos, and pinning freezes that; every other build-logic reference in this workflow (the reusable workflow and `build_logic_ref`) already tracks main, so a pinned checkout beside them would be inconsistent. Merge order is the control, and it is now stated in the PR description, on the ticket, and in the workflow itself.

sayaliM0412 (approved):

PR #7922, opened 2026-08-18
Allow sequence startValue and incrementBy on Firebird

makssent:
## Impact

## Description
`CreateSequenceGenerator` rejected `startValue` and `incrementBy` on Firebird unconditionally,
so `<createSequence startValue="1000"/>` failed validation with
`startValue is not allowed on firebird` on every Firebird version.
Firebird only implements the SQL standard `START WITH` semantics since 4.0 (https://github.com/FirebirdSQL/firebird/issues/6615):
earlier versions stored the value as the current generator value, so the first generated
value was off by one increment. The restriction is now version-gated: kept below 4.0,
lifted from 4.0 on. When the version cannot be determined (no live connection), the fields
are allowed, matching how the Postgres and H2 checks in the same class behave.
Covered by a new unit test in `CreateSequenceGeneratorTest`.

## Release note
`createSequence` now accepts `startValue` and `incrementBy` on Firebird 4.0 and later,
instead of failing validation on every Firebird version.

## Things to be aware of
Implemented as `isFirebirdWithoutStartValueSupport(Database)`, mirroring the existing
`isPostgreWithoutAsDatatypeSupport` and `isH2WithoutAsDatatypeSupport` helpers, including
their lenient fallback when the version is unknown. `generateSql` needed no change, since it
already emits `START WITH` and `INCREMENT BY` without excluding Firebird.

### DIFF
--- a/liquibase-standard/src/main/java/liquibase/sqlgenerator/core/CreateSequenceGenerator.java
+++ b/liquibase-standard/src/main/java/liquibase/sqlgenerator/core/CreateSequenceGenerator.java
@@ -25,8 +25,10 @@ public class CreateSequenceGenerator extends AbstractSqlGenerator<CreateSequence
 
         validationErrors.checkRequiredField("sequenceName", statement.getSequenceName());
 
-        validationErrors.checkDisallowedField("startValue", statement.getStartValue(), database, FirebirdDatabase.class);
-        validationErrors.checkDisallowedField("incrementBy", statement.getIncrementBy(), database, FirebirdDatabase.class);
+        if (isFirebirdWithoutStartValueSupport(database)) {
+            validationErrors.checkDisallowedField("startValue", statement.getStartValue(), database, FirebirdDatabase.class);
+            validationErrors.checkDisallowedField("incrementBy", statement.getIncrementBy(), database, FirebirdDatabase.class);
+        }
 
         if (isH2WithMinMaxSupport(database)) {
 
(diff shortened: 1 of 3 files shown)


filipelautert (commented):
Hello @makssent ! The issue you linked, FirebirdSQL#6615, is about IDENTITY columns rather than CREATE SEQUENCE, and it is an off-by-one fixed in 4.0 RC1. Do you have the reference for START WITH on sequences arriving in 4.0? Just want to be sure the cut is not rejecting Firebird 3 where it already works.

makssent:
> Hello @makssent ! The issue you linked, FirebirdSQL#6615, is about IDENTITY columns rather than CREATE SEQUENCE, and it is an off-by-one fixed in 4.0 RC1. Do you have the reference for START WITH on sequences arriving in 4.0? Just want to be sure the cut is not rejecting Firebird 3 where it already works.

Hello! Before Firebird 4.0, `START WITH` for sequences behaved incorrectly with respect to the SQL standard. Here is the [bug description](https://www.firebirdsql.org/file/documentation/chunk/en/refdocs/fblangref30/fblangref30-ddl-sequence.html#:~:text=Bug%20with%20START%20WITH%20and%20INCREMENT%20%5BBY), where this behavior is described, and here is the [issue with the fix](https://github.com/FirebirdSQL/firebird/issues/6334), CORE6084.

The issue I mentioned earlier, [#6615](https://github.com/FirebirdSQL/firebird/issues/6615), also mentions this problem, but it explicitly says: “This is related to CORE6084”. So I picked the wrong issue - it was referenced in one of the documentation sections. My mistake.

Therefore, the Firebird < 4.0 version check is needed specifically because of the incorrect `START WITH` behavior

PR #7931, opened 2026-08-20
ci: route branch/commit metadata through env: in fork-PR workflows (TECHOPS-525, TECHOPS-526)

mariiapisarenko:
## What

Continues the workflow-hardening work from [TECHOPS-526]/[TECHOPS-525]. `build.yml`, `build-branch.yml`, `run-tests.yml`, and `run-test-harness.yml` all derive a branch name, commit SHA, and build timestamp from the triggering PR and pass them between jobs as outputs. Several places spliced those job outputs directly into `run:` script text via `${{ }}`. This PR routes them through `env:` and references them as shell variables instead, matching the pattern already used for the direct `github.event.*/inputs.*` cases fixed earlier. `run-test-harness.yml`'s test-harness dispatch payload is also now built with jq instead of hand-concatenated JSON.

## Why

GitHub's own security guidance recommends passing untrusted or PR-derived values through `env:` rather than interpolating them directly into script bodies - the existing fixes covered direct references to PR-controlled context, but the same values also flow through job outputs into a few remaining spots. This closes that gap and keeps the whole fork-PR build/test path consistent.

[TECHOPS-526]: https://datical.atlassian.net/browse/TECHOPS-526?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
[TECHOPS-525]: https://datical.atlassian.net/browse/TECHOPS-525?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -165,7 +165,7 @@ jobs:
         env:
           BRANCH_NAME: ${{ needs.setup.outputs.thisBranchName }}
         run: |
-          version="$BRANCH_NAME-SNAPSHOT"
+          version=$BRANCH_NAME-SNAPSHOT
           ./mvnw versions:set -DnewVersion="$version"
 
       - name: Artifacts - Build & Sign
@@ -197,7 +197,7 @@ jobs:
           # Extract the tarball to allow us to upload the liquibase installation files to our
           # github actions workflow so that we don't double zip the archive
           mkdir -p branch-artifacts
-          tar -xzf "liquibase-dist/target/liquibase-$BRANCH_NAME-SNAPSHOT.tar.gz" -C branch-artifacts
+          tar -xzf liquibase-dist/target/liquibase-$BRANCH_NAME-SNAPSHOT.tar.gz -C branch-artifacts
 
           ##create installer - disabled here but run as nightly job and as part of release workflow
           # (cd liquibase-dist && ${{ github.workspace }}/.github/util/package-install4j.sh 0-SNAPSHOT)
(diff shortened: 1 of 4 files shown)


.github/workflows/build.yml
```diff
@@ -162,15 +162,20 @@ jobs:
 
       # Version artifact based off of branch
       - name: Version Artifact
+        env:
+          BRANCH_NAME: ${{ needs.setup.outputs.thisBranchName }}
```
jandroav, line 149: This fixes the pattern for downstream jobs, but the identical `Get Current BranchName` step in this same file's `setup` job (lines 54–68) still interpolates `${{ github.ref_name }}` directly into the `run:` block and still uses the old backtick subshell — you applied the `REF_NAME` env fix to the copy of this step in `build-branch.yml` and `run-tests.yml` but not here. `github.ref_name` is base-ref controlled in the `pull_request_target` context so exploitability is low, but it's the same pattern TECHOPS-526 wants eliminated, and any zizmor/actionlint rule added per the ticket's acceptance criteria will flag it. Suggest applying the same `env:` treatment here while this PR is touching the file.

.github/workflows/build.yml
```diff
@@ -254,20 +259,28 @@ jobs:
...
           GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+          BRANCH_NAME: ${{ needs.setup.outputs.thisBranchName }}
+          MERGE_SHA: ${{ needs.setup.outputs.latestMergeSha }}
+          BUILD_TIMESTAMP: ${{ needs.setup.outputs.timeStamp }}
 
       #Publish to GitHub Packages.
       - name: Publish tar.gz package to GPM
         run: |
-          ./mvnw -B -pl liquibase-dist clean deploy -DskipTests=true "-Dbuild.timestamp=${{ needs.setup.outputs.timestamp }}" "-Dbuild.repository.owner=liquibase" "-Dbuild.repository.name=liquibase" "-Dbuild.branch=${{ needs.setup.outputs.thisBranchName }}" "-Dbuild.number=${{ github.run_number }}" "-Dbuild.commit=${{ needs.setup.outputs.thisSha }}"
+          ./mvnw -B -pl liquibase-dist clean deploy -DskipTests=true "-Dbuild.timestamp=$BUILD_TIMESTAMP" "-Dbuild.repository.owner=liquibase" "-Dbuild.repository.name=liquibase" "-Dbuild.branch=$BRANCH_NAME" "-Dbuild.number=${{ github.run_number }}" "-Dbuild.commit=$MERGE_SHA"
```
jandroav, line 261: Heads-up: this is a behavior change, not just an injection fix. The old line referenced `${{ needs.setup.outputs.timestamp }}` (lowercase) and `${{ needs.setup.outputs.thisSha }}` — neither exists in this file's `setup` outputs (which define `timeStamp` and `latestMergeSha`; `thisSha`/`timestamp` only exist in `run-tests.yml`'s setup job), so `build.timestamp` and `build.commit` were resolving to empty strings on this deploy. After this change they're populated with real values. That's probably the intended fix, but it changes the metadata on the published tar.gz — worth confirming explicitly since it goes beyond the security hardening described in the PR.

jandroav (commented):
The env:-routing approach is correct and consistently applied to the job outputs that flow into `run:` blocks, and building the test-harness dispatch payload with jq is a solid improvement over hand-concatenated JSON. Two things to look at before merge: the same injection pattern survives in this PR's own `build.yml` setup job, and one of the converted steps silently changes which values get published compared to what the old expressions resolved to.

.github/workflows/build.yml
mariiapisarenko: Thank you for the catch, fixed in e601832.

.github/workflows/build.yml
mariiapisarenko: Agreed - split out to [TECHOPS-1126](https://datical.atlassian.net/browse/TECHOPS-1126), reverted in e601832 so this PR stays pure security hardening.

.github/workflows/build.yml
```diff
@@ -254,20 +261,28 @@ jobs:
...
         env:
           GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+          BRANCH_NAME: ${{ needs.setup.outputs.thisBranchName }}
+          MERGE_SHA: ${{ needs.setup.outputs.thisSha }}
+          BUILD_TIMESTAMP: ${{ needs.setup.outputs.timestamp }}
```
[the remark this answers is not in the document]
mariiapisarenko: Intentional - split out to [TECHOPS-1126], staying broken here on purpose so this PR has zero behavior change.

jandroav (approved):

PR #7933, opened 2026-08-20
[TECHOPS-1121] Re-enable package publishing and fail the run when a stage is skipped

sayaliM0412:
## 🐛 Problem

`package` (Update Packages) and `publish-assets-s3` were set to `if: false` in `ac819e117` (2026-03-05) as a one-off for the v5.0.2 recovery, and never re-enabled. They are not conditional on anything, they are literally `if: false`.

That one flag gates **five OSS channels**, since they are all jobs inside `build-logic/package.yml`. Verified live 2026-08-20:

| Channel | Current | Should be |
|---|---|---|
| apt | 5.0.2 | 5.0.4 |
| yum | 5.0.2 | 5.0.4 |
| SDKMAN | 5.0.2 | 5.0.4 |
| Chocolatey | 5.0.2 | 5.0.4 |
| Ansible Galaxy | v5.0.2 | 5.0.4 |

All five stopped at the same version. `apt install liquibase` has been serving a five-month-old release. `liquibase-secure` is unaffected, it ships from the commercial pipeline.

## 🙈 Why nobody noticed

The summary already computed the right answer and then discarded it. From the 5.0.4 run:

```
[ "skipped" == "success" ] && \
[ "skipped" == "success" ]; then
  OVERALL_STATUS="✅ SUCCESS"
  OVERALL_STATUS="❌ FAILED"
```

It printed `❌ FAILED`, then exited 0, because writing the summary is the last thing the job does. So the run badge stayed green while the summary said the release had failed.

## 🔧 Changes

1. **Restore both conditions** to exactly what `ac819e117` replaced, not just deleting the line. `package` keeps its setup/manual-approval gate; `publish-assets-s3` keeps its dependency on `package`.
2. **Make the summary fail the run** when any stage did not succeed.

## ✅ Verification

- `actionlint` goes from **4 findings to 2**: this clears both `constant expression "false" in condition [if-cond]` errors it was already reporting on `main`. The 2 remaining are pre-existing and untouched.
- YAML parses; all 10 jobs intact.
- Restored conditions diffed against `ac819e117` to confirm they match the originals character for character.

## 📦 Not included

Backfilling 5.0.3 and 5.0.4 to the five channels. That is a `package.yml` dispatch against live package repos and wants coordinating with whoever owns packaging, so it is deliberately not part of this PR.

Ticket: https://datical.atlassian.net/browse/TECHOPS-1121

### DIFF
--- a/.github/workflows/release-published-orchestrator.yml
+++ b/.github/workflows/release-published-orchestrator.yml
@@ -165,7 +165,7 @@ jobs:
   package:
     name: Update Packages
     needs: [setup, manual-approval]
-    if: false  # TEMP: skip for v5.0.2 recovery - already published
+    if: always() && needs.setup.result == 'success' && (needs.manual-approval.result == 'success' || needs.manual-approval.result == 'skipped')
     uses: liquibase/build-logic/.github/workflows/package.yml@main
     permissions:
       actions: write
@@ -190,7 +190,7 @@ jobs:
   publish-assets-s3:
     name: Publish Assets to S3
     needs: [setup, package]
-    if: false  # TEMP: skip for v5.0.2 recovery - already uploaded
+    if: always() && needs.setup.result == 'success' && (needs.package.result == 'success' || needs.package.result == 'skipped')
     uses: ./.github/workflows/release-publish-assets-s3.yml
     permissions:
       contents: read
@@ -291,3 +291,12 @@ jobs:
           EOF
 
           echo "📋 Release $VERSION: $OVERALL_STATUS"
+
+          # Fail the run when any stage did not succeed. Without this the summary
+          # can report FAILED while the run itself goes green, because writing the
+          # summary is the last thing this job does and it always exits 0. That is
+          # how two skipped publish stages went unnoticed across several releases.
+          if [ "$OVERALL_STATUS" != "✅ SUCCESS" ]; then
+            echo "::error::Release $VERSION did not complete: one or more stages failed or were skipped"
+            exit 1
+          fi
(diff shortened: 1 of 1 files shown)


.github/workflows/release-published-orchestrator.yml
```diff
@@ -291,3 +291,12 @@ jobs:
           EOF
 
           echo "📋 Release $VERSION: $OVERALL_STATUS"
+
+          # Fail the run when any stage did not succeed. Without this the summary
+          # can report FAILED while the run itself goes green, because writing the
+          # summary is the last thing this job does and it always exits 0. That is
+          # how two skipped publish stages went unnoticed across several releases.
+          if [ "$OVERALL_STATUS" != "✅ SUCCESS" ]; then
```
jandroav: This gate treats `skipped` as failure for every stage, but `release-docker` is skipped *by design* on dry runs (`inputs.dry_run != true` in its `if:`, with a comment at lines 133–137 explaining why). After this change, every dry-run orchestrator execution will exit 1 even when everything that was supposed to run succeeded — a red run on an intentional configuration, which undermines the dry run as a validation tool (the ticket's own rollout sequence leans on dry-run verification). Suggest exempting dry runs here, e.g. only exiting 1 when `inputs.dry_run != true`, or accepting `needs.release-docker.result == 'skipped'` in the SUCCESS gate when dry_run is set.

jandroav (commented):
The restored conditions on `package` and `publish-assets-s3` match exactly what `ac819e117` replaced, and failing the run when the summary computes FAILED closes the visibility gap that hid this for five months. One problem with the new gate: it also fires on a skip that's intentional — details inline.

jandroav (approved):

PR #7870, opened 2026-08-01
Propagate directory visit failures

cragkhit:
Hi. We are researchers from Mahidol University, Thailand, and the State University of Ceará, Brazil, working on a research project for improving open-source projects by using the latest accepted answer from Stack Overflow that matched your code snippet. We found this recommendation for improving your code from https://stackoverflow.com/a/18454342.

Note: Our study is approved by the Institutional Review Board of Mahidol University. You can find the participant information sheet explaining this study https://drive.google.com/file/d/1ml5AqrtWQ9pnifTQyTFTcWQmwp6RuPA7/view?usp=sharing.

----

## Proposed change

Propagated post-visit traversal exceptions instead of silently continuing a failed resource search.

PR #7941, opened 2026-08-24
Fix includeAll search silently ignoring directory traversal failures

### LINKED ISSUE #7875: Directory traversal failures can silently omit includeAll resources
## Search first

- [x] I searched the existing issues and found no similar report.

## Description

`AbstractPathResourceAccessor` can silently return a partial set of resources when directory traversal fails.

The `SimpleFileVisitor` used to search resources overrides `postVisitDirectory(Path, IOException)`, but currently returns `FileVisitResult.CONTINUE` even when the callback receives a non-null `IOException`. Consequently, an `includeAll` search can omit files from a directory whose iteration failed without reporting the failure to the user.

This concern was identified during review of #7870. That pull request was based on an unrelated Stack Overflow edit, so this issue records the underlying Liquibase behavior independently, as requested in the review.

## Steps To Reproduce

1. Configure a changelog with `includeAll` targeting a directory tree.
2. Cause traversal of one of its directories to fail, for example by using a resource accessor or filesystem setup that produces an `IOException` while iterating that directory.
3. Run an operation that resolves the changelog resources.
4. Observe that the search can continue with a partial resource set instead of reporting the traversal failure.

A focused unit test could reproduce this deterministically with a test resource accessor/filesystem that injects an exception during directory traversal.

## Expected/Desired Behavior

Liquibase should surface the directory traversal failure rather than silently treating a partial resource search as successful. The eventual implementation should include a regression test covering the failure path and confirm the intended exception type exposed to callers.

## Liquibase Version

Current `main` branch

## Database Vendor & Version

Not database-specific

## Liquibase Integration

Core resource resolution / `includeAll`

## Liquibase Extensions

None required

## OS and/or Infrastructure Type/Provider

Any filesystem or resource accessor capable of producing a directory-iteration failure

## Additional Context

Related implementation discussion: #7870

## Are you willing to submit a PR?

- [x] I'm willing to submit a PR after the expected behavior and test approach are confirmed, subject to completing the required CLA.

hutiefang76:
## Impact

## Description

Fixes #7875.

`AbstractPathResourceAccessor.search()` previously continued when `Files.walkFileTree` reported an `IOException` after visiting a directory. That could make an `includeAll` operation appear successful while silently omitting changelog files from an unreadable directory.

The visitor now propagates the supplied exception. The behavior for successful directory visits is unchanged.

A regression test invokes the real `search()` entry point with a real temporary directory and simulates only the JDK walker callback. It verifies that the exact traversal exception reaches the caller.

## Release note

Fixed an issue where an `includeAll` search could silently omit changelog files after a directory traversal error. Liquibase now reports the traversal failure instead.

## Things to be aware of

This intentionally changes a failed directory iteration from a partial successful result to an `IOException`, preventing callers from treating an incomplete changelog set as complete.

## Things to worry about

No database-specific behavior is affected. The regression test isolates the otherwise difficult-to-trigger `postVisitDirectory` error callback without relying on filesystem permissions or platform behavior.

## Additional Context

Verification:

```text
./mvnw -pl liquibase-standard -Dtest=AbstractPathResourceAccessorTest,UrlDrivenResourceAccessorTest test
Tests run: 4, Failures: 0, Errors: 0, Skipped: 0
```

### DIFF
--- a/liquibase-standard/src/main/java/liquibase/resource/AbstractPathResourceAccessor.java
+++ b/liquibase-standard/src/main/java/liquibase/resource/AbstractPathResourceAccessor.java
@@ -203,6 +203,9 @@ public abstract class AbstractPathResourceAccessor extends AbstractResourceAcces
 
             @Override
             public FileVisitResult postVisitDirectory(Path dir, IOException exc) throws IOException {
+                if (exc != null) {
+                    throw exc;
+                }
                 return FileVisitResult.CONTINUE;
             }
 
--- a/liquibase-standard/src/test/java/liquibase/resource/AbstractPathResourceAccessorTest.java
+++ b/liquibase-standard/src/test/java/liquibase/resource/AbstractPathResourceAccessorTest.java
@@ -1,0 +1,45 @@
+package liquibase.resource;
+
+import org.junit.jupiter.api.Test;
+import org.junit.jupiter.api.io.TempDir;
+import org.mockito.MockedStatic;
+import org.mockito.Mockito;
+
+import java.io.IOException;
+import java.nio.file.FileVisitor;
+import java.nio.file.Files;
+import java.nio.file.Path;
+
+import static org.junit.jupiter.api.Assertions.assertSame;
+import static org.junit.jupiter.api.Assertions.assertThrows;
+
+class AbstractPathResourceAccessorTest {
+
+    @TempDir
+    Path tempDir;
+
+    @Test
+    void searchPropagatesDirectoryTraversalFailure() throws Exception {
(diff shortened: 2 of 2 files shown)


hutiefang76:
Closing this to leave the implementation path with the original reporter, who previously opened #7870 and stated in #7875 that they were willing to submit the follow-up after the expected behavior and test approach were confirmed.

The issue has no active covering PR, but I should have confirmed ownership before opening this. The local reproduction and regression-test approach can be shared if the reporter or a maintainer asks for help.

PR #7944, opened 2026-08-25
[TECHOPS-1104] Secret-free PR CI and trusted main build (step 1, additive)

jnewton03:
## Summary

Step 1 of the TECHOPS-1100 workflow refactor. **Additive only**: no existing workflow is removed or disabled in this PR. Cutover (disable the superseded workflows, add ruleset checks) is a follow-up once `main.yml` has run green on `main`.

**Rule this establishes:** code from a pull request never runs with secrets. No `pull_request_target`.

## New workflows

| File | Trigger | Secrets | Replaces |
|---|---|---|---|
| `ci-test.yml` | `workflow_call` | none | matrices in `run-tests.yml` / `test-pr.yml` |
| `pr.yml` | `pull_request` | none | `test-pr.yml`, PR paths of `run-tests.yml`, `build-branch.yml`, `docker-test.yml`, `docker-scan.yml` |
| `pr-labels.yml` | `pull_request` | none | `label-pr.yml` (fails on forks today) |
| `main.yml` | push main, nightly, dispatch | vault for Sonar/FOSSA only | `run-tests.yml` push path, `build.yml`, `build-main.yml`, `nightly-release.yml` |
| `snapshot-branch.yml` | dispatch | none | per-PR branch snapshots (QA on-demand) |
| `docker.yml` | push main (docker/**), cron, dispatch | vault | `docker-test.yml`, `docker-scan.yml` |
| `cleanup-packages.yml` | weekly cron | none | `cleanup-branch-builds.yml` |

## Notes for reviewers

- `main.yml` keeps producing the `liquibase-artifacts` artifact with the same file names `build.yml` produced (unsigned; `create-release.yml` re-versions and re-signs), so the release chain keeps working until Phase 3 builds from tag.
- `<sha>-SNAPSHOT` is the full 40-char SHA: liquibase-test-harness resolves community builds that way.
- `main.yml` also uploads `liquibase-zip-main` / `liquibase-artifacts-main`, and `snapshot-branch.yml` uploads `liquibase-zip-<branch>` / `liquibase-artifacts-<branch>`: liquibase-pro-tests and the harness fetch OSS builds through `liquibase-sdk-maven-plugin` (`install-snapshot-cli -Dliquibase.sdk.workflowId=…`). At cutover those consumers switch `workflowId` from `run-tests.yml` to `main.yml` / `snapshot-branch.yml`.
- Sonar and FOSSA move to main-only (they need vault). PR coverage delta in Sonar is lost by design.
- build-logic reusables referenced `@main` (the actions inside build-logic are SHA-pinned; TECHOPS-1109 scope is build-logic itself). `persist-credentials: false` + top-level `permissions` everywhere (TECHOPS-1111).
- Duplicate SNAPSHOT deploys with `build-main.yml` during the overlap are harmless (same coordinates).

## Verification

- [x] `actionlint` clean
- [x] CodeRabbit findings addressed (tag-target retention in cleanup, main-only scan persistence)
- [ ] `PR Gate` and `PR Labels` green on this PR
- [ ] After merge: `main.yml` publishes `main-SNAPSHOT` + `<sha>-SNAPSHOT`, uploads `liquibase-artifacts`, recreates `nightly`

Epic: TECHOPS-1100. Stories touched: TECHOPS-1104, TECHOPS-1111, TECHOPS-1099.


.github/workflows/cleanup-packages.yml
```diff
@@ -0,0 +1,88 @@
...
+env:
+  PACKAGES: org.liquibase.liquibase-core org.liquibase.liquibase-extension-testing org.liquibase.liquibase-integration-tests org.liquibase.liquibase-maven-plugin org.liquibase.liquibase
+
+jobs:
+  sweep:
+    name: Sweep orphaned SHA-SNAPSHOT versions
+    runs-on: ubuntu-24.04
+    permissions:
+      contents: read
+      packages: write
```
jandroav, line 30: Worth a dry-run dispatch before the first cron: these packages live in the org namespace (`orgs/$ORG/packages/...`), and deleting versions there with the workflow `GITHUB_TOKEN` commonly 403s even with `packages: write`, unless the package's repo-link grants this repo admin on the package. If it fails, the fallback is the GitHub App token or a `delete:packages`-scoped credential, which would change this workflow's secret story (today: none, which is lovely). The `::warning` + `failed=1` handling means a silent-403 world shows up as a red run, so at least it won't rot invisibly.

.github/workflows/snapshot-branch.yml
```diff
@@ -0,0 +1,106 @@
...
+          ref: ${{ inputs.ref }}
+          persist-credentials: false
+
+      - name: Resolve SHA and branch name
+        run: |
+          {
+            echo "BUILD_SHA=$(git rev-parse HEAD)"
+            echo "BUILD_TIMESTAMP=$(date +'%Y-%m-%d %H:%M:%S %Z')"
+            # same munging as build.yml used, which is what the sdk plugin expects
+            echo "BRANCH_FILE=$(tr '/' '_' <<<"${{ inputs.ref }}")"
```
jandroav: `${{ inputs.ref }}` is interpolated directly into the script here (and in the Summary heredoc below). Dispatch requires write access, so the attacker is a maintainer and this is low severity, but env indirection is free: set `REF: ${{ inputs.ref }}` in the job `env:` and use `"$REF"`. That also makes the two `MAVEN_BUILD_ARGS`/checkout usages consistent, which already go through env/with correctly.

.github/workflows/main.yml
```diff
@@ -0,0 +1,240 @@
...
+      actions: write
+      pull-requests: write # declared by the reusable; no-op on push
+    with:
+      check_snippets: true
+      check_ai_generated_code: true
+      generate_fossa_3p_license_report: true
+    secrets: inherit
+
+  publish-snapshot:
+    name: Publish SNAPSHOT
```
jandroav, line 78: Design question tracked in TECHOPS-1224, not blocking this PR: this publishes `main-SNAPSHOT` + `<sha>-SNAPSHOT` and recreates the `nightly` release on every push to main, so the public repo keeps acting as a continuous build server, just from a trusted context now. If the harness/pro-tests consumers can live with daily freshness, gating this job and `nightly-release` on `schedule` + `workflow_dispatch` would land us exactly on "no public build server except nightly builds". If they can't, let's record why in the README consumer table.

jandroav (commented):
Full review pass as part of taking over epic TECHOPS-1100. The architecture is exactly right: one workflow runs PR code with a read-only token and zero secrets, everything credentialed lives on push/schedule/dispatch, every action SHA-pinned, `persist-credentials: false` everywhere except the one documented scan-results push. I verified the PR Gate fan-in logic (a failed `changes` job fails the gate; `skipped` passes only for genuinely skipped paths), the expression-injection surfaces (`head_ref` only ever reaches a secret-free job that the fork author already controls), and the artifact contracts against the README claims.

Findings, none architectural:

1. **Dependabot docker PRs will stall on the new required check.** `pr-labels.yml` requires one of `APIBreakingChanges, breakingChanges, newContributors, notableChanges, sdou, skipReleaseNotes, TypeBug, TypeEnhancement, TypeTest`. The maven and github-actions ecosystems in `dependabot.yml` add `sdou`, but the docker ecosystem adds only `docker` + `dependencies`, so once `PR Labels` is a required check every docker bump is unmergeable by automerge. Add `sdou` (or `skipReleaseNotes`) to the docker ecosystem's labels in the same cutover.
2. **Verify `cleanup-packages.yml` can actually delete** (inline): org-namespace package-version deletion with the workflow `GITHUB_TOKEN` is the classic 403; needs a dry-run dispatch before trusting the weekly cron.
3. **`snapshot-branch.yml` interpolates `inputs.ref` into `run:` scripts** (inline): maintainer-only surface, but env-var indirection is cheap.
4. **Policy question, non-blocking here, tracked as TECHOPS-1224:** `publish-snapshot` + `nightly-release` fire on every push to main, so the repo still publishes consumable artifacts continuously, and the `nightly` release is recreated per-merge rather than nightly. Given the epic direction ("no public build server except nightly builds"), consider gating `publish-snapshot`/`nightly-release` on `schedule` + `workflow_dispatch` only, or recording in the README why per-push freshness is required by the harness/pro-tests consumers. Fine to defer; that discussion has its own ticket now.

Two reminders for the merge itself, from TECHOPS-1222: PR events run the base branch's workflow definitions, so nothing in this PR is exercised by its own checks; plan the canary fork PR right after merge. And `PR Gate` + `PR Labels` only become meaningful once added to `strict_branch_protection` (today only `mergefreeze` is required).

@jnewton03 this has been draft since Aug 25; what's left before it's review-complete and un-drafted? Happy to pick up the remaining pieces, since I'm taking the epic forward.

.github/workflows/cleanup-packages.yml
jandroav: Verified empirically, no change needed: cleanup-main-builds.yml deletes versions of these same five org.liquibase.* maven packages with the default GITHUB_TOKEN (via the pinned delete-package-versions action) and its runs succeed, most recently 2026-08-20. So the token can delete versions of these repo-linked packages. Keeping the dry-run dispatch as the first-run sanity check; if the REST org endpoint ever behaves differently from the action's path, the fallback is switching this workflow to the same pinned action with package-version-ids, still with no extra secret. Resolving.

.github/workflows/snapshot-branch.yml
jandroav: Fixed in 316603aa9: the ref now flows through the REF job env var in both run steps (BRANCH_FILE derivation and the summary heredoc). Resolving.

jandroav (approved):

jandroav (approved):

jandroav (approved):
Approving as epic lead (TECHOPS-1100/TECHOPS-1104). Full review pass done on 2026-09-01: architecture verified (secret-free PR path, trusted main path), all inline findings addressed or resolved with evidence, branch current with main, actionlint clean on all seven new workflows, build-logic reusables verified. Additive change: no existing workflow is modified.

jandroav (approved):
Re-approving after removing the settings-snapshots.xml comment tweak; the PR now touches only liquibase-devops-owned paths.

jandroav (approved):

jandroav (approved):

jandroav (approved):

sayaliM0412 (approved):

PR #7948, opened 2026-08-26
[TECHOPS-1103] Gate the release on the release environment

sayaliM0412:
> 🚫 **Do not merge until [liquibase-infrastructure#4270](https://github.com/liquibase/liquibase-infrastructure/pull/4270) is merged AND applied.** Until the `release` environment exists, this blocks the next release on a gate that is not there.

Puts `environment: release` on the `manual-approval` job and deletes `trstringer/manual-approval`, so the reviewer list lives in Terraform-managed repository settings instead of workflow YAML and the release path drops a third-party action that held `GITHUB_TOKEN` and opened tracking issues. The caller no longer needs `issues`/`pull-requests` write or `secrets: inherit`, so both are removed, and `.github/workflows/README.md` is updated for whoever approves or debugs a release next.

⚠️ **Approval count changes from 2 to 1.** GitHub environments have no N-of-M setting, so `minimum-approvals: 2` has no equivalent; `prevent_self_review` on the environment keeps the approver distinct from whoever started the run, preserving "two different people" but not "two approvals". **This needs a release-owner decision before merge**: if two approvals are a hard requirement, this PR should be closed and the action kept.

Dry runs are unaffected (the job is still gated on `dry_run == false` and never reaches the environment). `actionlint` is clean on the approval workflow; the two findings it reports on the orchestrator are pre-existing on `main`.

---

**Merge order:** [liquibase-infrastructure#4270](https://github.com/liquibase/liquibase-infrastructure/pull/4270) merged **and applied** → this PR.

**Before marking ready:** confirm `release` exists under Settings → Environments with its reviewers and both deployment rules, defined in [liquibase-release-environment.tf](https://github.com/liquibase/liquibase-infrastructure/blob/main/github/liquibase/repos/public/liquibase-release-environment.tf).

**After merge, check:** the next dry run still skips `Manual Deployment Approval` entirely, and the next real release pauses there with a **Review deployments** prompt on the `release` environment rather than opening a tracking issue. After one reviewer approves, the publishing jobs (Maven Central, GitHub Packages, javadocs, XSD, Docker, S3) should run as before. If the job fails immediately with a deployment-rejected error, the environment's deployment rules do not match the run's ref.

### DIFF
--- a/.github/workflows/README.md
+++ b/.github/workflows/README.md
@@ -327,7 +327,7 @@ To add a new step to the release process:
 
 **Issue**: Manual approval never arrives or is rejected
 - **Solution**: Check the run's deployment gate, not the issues list: approval happens on the run page under Review deployments
-- **Solution**: Check the reviewer list on the `release` environment. It is managed in liquibase-infrastructure (`github/liquibase/repos/public/liquibase-release-environment.tf`), so changing it takes a PR there
+- **Solution**: Check the reviewer list on the `release` environment. It is managed in [liquibase-release-environment.tf](https://github.com/liquibase/liquibase-infrastructure/blob/main/github/liquibase/repos/public/liquibase-release-environment.tf), so changing it takes a PR in liquibase-infrastructure
 - **Solution**: A reviewer cannot approve a run they started themselves. Someone else on the list has to
 
 **Issue**: Secrets not available in called workflows
@@ -392,7 +392,7 @@ dry_run_branch_name: (leave empty for production)
 **How approval works:** The job carries `environment: release`, so GitHub pauses it and shows
 **Review deployments** on the run page. Any reviewer on that environment can approve or reject,
 except the person who started the run. There is no tracking issue and no keyword replies. The
-reviewer list is Terraform-managed in liquibase-infrastructure.
+reviewer list is Terraform-managed in [liquibase-release-environment.tf](https://github.com/liquibase/liquibase-infrastructure/blob/main/github/liquibase/repos/public/liquibase-release-environment.tf).
 
 **Required inputs:**
 - `version`: Version to approve (e.g., `4.28.0`)
--- a/.github/workflows/README.md
+++ b/.github/workflows/README.md
@@ -387,7 +387,9 @@ dry_run_branch_name: (leave empty for production)
 
 ### 2. Release Manual Approval (`release-manual-approval.yml`)
 
-**When to use:** If approval was skipped or needs to be re-requested.
+**When to use:** Normally never on its own. Approval is per-run: a paused orchestrator run is
+approved from **Review deployments** on that run's page, and dispatching this workflow standalone
+only gates its own, empty run. It exists as the reusable gate job the orchestrator calls.
 
 **How approval works:** The job carries `environment: release`, so GitHub pauses it and shows
 **Review deployments** on the run page. Any reviewer on that environment can approve or reject,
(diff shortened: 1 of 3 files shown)


.github/workflows/release-manual-approval.yml
```diff
@@ -28,22 +28,20 @@ on:
...
     name: Request Manual Approval
     runs-on: ubuntu-22.04
     if: ${{ inputs.dry_run == false }}
+    # The gate itself. GitHub holds the job here until a reviewer on the
+    # `release` environment approves the run, and every publishing job in the
+    # orchestrator depends on this job. The environment and its reviewer list are
+    # defined in liquibase-infrastructure, not in this file, so they cannot be
+    # changed by editing a workflow:
+    # https://github.com/liquibase/liquibase-infrastructure/blob/main/github/liquibase/repos/public/liquibase-release-environment.tf
+    environment: release
```
jandroav, line 44: Merge-order guard for this line: if this runs before the `release` environment exists in GitHub (infra#4270 applied via Spacelift), GitHub silently auto-creates the environment with no protection rules and the job sails through unapproved. Worth holding this PR as draft until Settings -> Environments shows `release` with reviewers and both deployment rules.

.github/workflows/README.md
```diff
@@ -388,6 +389,11 @@ dry_run_branch_name: (leave empty for production)
 
 **When to use:** If approval was skipped or needs to be re-requested.
```
jandroav: Nit: with environment gates, approval is per-run, so dispatching `release-manual-approval.yml` standalone approves nothing for a paused orchestrator run (its own run just requests its own approval). This "re-request approval" entry is now misleading; suggest rewording to point at the orchestrator run's Review deployments button, or dropping the section.

jandroav (commented):
Reviewed as part of epic TECHOPS-1100. Content is right; the one blocker is merge order, not code.

**Checks I ran**

- `environment: release` sits on the job inside the reusable workflow, so via `workflow_call` it applies in this repo's context, and the deployment policies from liquibase-infrastructure#4270 (tag `v*`, branch `main`) cover every ref this job actually runs from: `release: published` (tag), orchestrator `workflow_dispatch` (main), standalone dispatch (main). Dry runs skip the job via the existing `if:`, so they never touch the gate.
- Rejection path works: a rejected deployment fails `manual-approval`, and every publishing job requires `success || skipped`, where `skipped` is only reachable on the dry-run path. Verified against the orchestrator's `needs`/`if` chains.
- Dropping `issues: write`/`pull-requests: write` for `contents: read` and removing `secrets: inherit` from the call is correct; the job needs nothing anymore. Only the orchestrator calls this workflow, so no other caller breaks.

**Blocking: merge order.** Do not merge until #4270 is merged AND its apply is confirmed (Settings -> Environments shows `release` with the 5 reviewers and both deployment rules). If this merges first, the next release run auto-creates a `release` environment with zero protection rules: the gate becomes a no-op, and once infra#4279 is applied, the `environment:release` subject its trust policy accepts would be presentable from that ungated job. Suggest converting to draft until the environment is verified live.

**On the 2-to-1 approval question** (for the release owners to ack, as Sailee flagged): one framing that may help the conversation. trstringer counted approvals from the listed approvers without excluding whoever started the run, so the old flow effectively guaranteed 2 humans (publisher plus at least one other). The new flow guarantees the same 2 humans: publisher plus one distinct approver, enforced by `prevent_self_review`. The only real regression is against the stricter reading of 2-of-5 independent approvals. Either way it needs their explicit OK before this merges.

jandroav (approved):

jandroav (approved):

.github/workflows/release-manual-approval.yml
sayaliM0412: check : https://github.com/liquibase/liquibase/settings/environments/20980522310/edit
Screenshot 2026-09-01 at 12 34 36 PM
cc: @jandroav

filipelautert (approved):

PR #7954, opened 2026-08-28
[TECHOPS-1188] Harden kept workflows: deny-by-default permissions and persist-credentials

sayaliM0412:
## Impact

## Description

Adopts deny-by-default `GITHUB_TOKEN` permissions and `persist-credentials: false` on the workflows
that keep running after the PR-CI cut-over.

**Verified, not assumed:** the repository and organization both return
`{"default_workflow_permissions":"write","can_approve_pull_request_reviews":true}`, so a workflow with
no `permissions:` block today runs with a write-scoped token that can also approve pull requests. And
`persist-credentials` defaults to `true`, read from `actions/checkout`'s own `action.yml` at the single
SHA pinned across all 31 call sites in this repo.

| workflow | before | after |
|---|---|---|
| `codeql.yml` | no top-level block (write-all) | `permissions: {}`, analyze job keeps its existing grants, `persist-credentials: false` |
| `cleanup-main-builds.yml` | no top-level block (write-all) | `permissions: {}`, `packages: write` on the one job |
| `release-extensions.yml` | no top-level block (write-all) | explicit block mirroring what the callee declares |
| `release-drafter.yml` | duplicate top-level and job block | `permissions: {}`, `pull-requests` write to read |
| `create-release.yml` | `contents/actions/id-token: write` for every job | `permissions: {}`, per-job grants, `actions: write` dropped |

## Things to be aware of

**Two scope removals argued from source, not taste.**

`create-release.yml`'s `actions: write` carried the in-file comment "trigger downstream workflows".
Reading all 437 lines, nothing in the file dispatches a workflow: no `createWorkflowDispatch`, no
`gh workflow run`, no `POST .../dispatches`. The direction is reversed, `dry-run-release.yml:93`
dispatches *this* workflow, which is why that file holds `actions: write`. The only Actions-API
consumer here is `reversion`'s artifact download from a different run, and `actions/download-artifact`
documents that as needing `actions: read`. Its `setup` job, whose only step is an `echo`, now inherits
`{}`.

`release-drafter.yml`'s `pull-requests: write` is documented by release-drafter's own README at the
pinned SHA as belonging to the separate `autolabeler` sub-action. This workflow uses the top-level
action, and `.github/release-drafter-main.yml` declares no `autolabeler` key, so `read` is sufficient.

**Credentials deliberately left persisted where a job pushes.** `create-release.yml`'s `reversion` job
pushes a tag over its checkout remote, so it keeps the default and says so in-file. The same applies to
the two scan-results pushers, `docker-scan.yml` and `trivy-scan-published-images.yml`, which are out of
scope here.

**`codeql.yml`'s security-report steps are now gated off `pull_request`.** A fork pull request gets a
read-only token, so `codeScanning.listAlertsForRepo` cannot succeed there. The script already caught
the failure and emitted an "INCOMPLETE" report, so this removes a step that could only ever warn rather
than granting a scope to make it work.

**actionlint: 133 findings on `origin/main`, 133 on this branch**, sets diffed identical after stripping
line numbers. All pre-existing.

## Things to worry about

**This touches the release path, so it wants a real run before merge.** Specifically one dry-run release
to exercise `create-release.yml` under the reduced scopes, and one push to `main` for
`release-drafter.yml`. Every reduced scope here is argued from the action's own manifest at its pinned
SHA plus a full read of the calling step, **not** from a green run.

Two I could not confirm without a run, stated plainly:

- Whether `packages: write` alone suffices to *delete* a package version with the repo-scoped
  `GITHUB_TOKEN`. There is no separate delete scope documented, and the workflow already effectively
  runs with `packages: write` via the write-all default, so the delete path is unchanged by this PR.
- `release-extensions.yml` is `workflow_dispatch`-only and release-gated. Its new block is set to
  exactly the four scopes the callee declares for itself, since a called workflow can only reduce, never
  elevate. Confirming it means an actual extension release.

## Additional Context

Part of the Vanta "P2 security issues resolved" remediation.

Two holes deliberately left out of scope, both arguably larger than what is in this PR:

- **`build.yml`** is `disabled_manually` but reachable, because `run-tests.yml:292` calls it by local
  path (`uses: ./.github/workflows/build.yml`), and `workflow_call` ignores the disabled state. It
  checks out `pull_request.head.sha` at lines 41 and 83 with credentials persisted, holds
  `packages: write` and `id-token: write`, and runs the fork's Maven build.
- **`build-logic/sonar-coverage-merge.yml`** has no top-level block, checks out the fork ref with
  credentials persisted at line 27, loads the whole vault at 38-42, then runs `mvn -B clean install`
  against that fork-controlled `pom.xml` at line 85.

Neither was on the remediation list. Both should be added to it.

### DIFF
--- a/.github/workflows/cleanup-main-builds.yml
+++ b/.github/workflows/cleanup-main-builds.yml
@@ -5,10 +5,16 @@ on:
   release:
     types: [ published ]
 
+# Deny by default so a new job cannot silently pick up the org-wide write default.
+permissions: {}
+
 jobs:
   delete-package:
     name: Delete Github Packages for Main
     runs-on: ubuntu-22.04
+    permissions:
+      # Only scope actions/delete-package-versions needs; it acts on packages owned by this repo.
+      packages: write
 
     strategy:
       matrix:
--- a/.github/workflows/build-main.yml
+++ b/.github/workflows/build-main.yml
@@ -8,10 +8,16 @@ on:
     branches:
       - 'main'
 
+# Deny by default so a new job cannot silently pick up the org-wide write default.
+permissions: {}
+
 jobs:
   setup:
     name: Setup
     runs-on: ubuntu-22.04
+    permissions:
+      # Only what the checkout needs; this job publishes nothing, it just emits two outputs.
+      contents: read
     outputs:
       timeStamp: ${{ steps.get-timestamp.outputs.timeStamp }}
       latestMergeSha: ${{ steps.get-sha.outputs.latestMergeSha }}
--- a/.github/workflows/codeql.yml
+++ b/.github/workflows/codeql.yml
@@ -20,6 +20,9 @@ on:
   schedule:
     - cron: '16 14 * * 4'
 
+# Deny by default so a new job cannot silently pick up the org-wide write default.
+permissions: {}
+
 jobs:
   analyze:
     name: Analyze
@@ -42,6 +45,9 @@ jobs:
     steps:
     - name: Checkout repository
       uses: actions/checkout@3d3c42e5aac5ba805825da76410c181273ba90b1 # v7.0.1
+      with:
+        # A pull_request run checks out the PR merge commit, so Autobuild builds fork-controlled code in this workspace.
+        persist-credentials: false
 
     # Initializes the CodeQL tools for scanning.
     - name: Initialize CodeQL
@@ -79,6 +85,8 @@ jobs:
     # Replaced rsdmike/github-security-report-action@v3.0.4 (stuck on Node 16, unmaintained)
     # with actions/github-script@v8 (Node 24) to generate a security report from code scanning alerts
     - name: Generate Security Report
+      # A fork pull_request gets a read-only token with no code-scanning access, so this only ever logs a warning there.
+      if: github.event_name != 'pull_request'
       uses: actions/github-script@3a2844b7e9c422d3c10d287c895573f7108da1b3 # v9.0.0
       with:
         script: |
@@ -156,6 +164,7 @@ jobs:
           core.info(`Security report generated with ${summary.total_open_alerts} open alerts`);
 
     - name: Upload Security Report
+      if: github.event_name != 'pull_request'
       uses: actions/upload-artifact@043fb46d1a93c77aae656e7c1c64a875d1fc6a0a # v7.0.1
       with:
         name: security-report
--- a/.github/workflows/create-release.yml
+++ b/.github/workflows/create-release.yml
@@ -1,13 +1,7 @@
 name: Attach Artifacts To Draft Release
 
-# Security: These write permissions are required for release creation
-# - contents: write - create release tags and upload assets
-# - actions: write - trigger downstream workflows
-# - id-token: write - AWS OIDC authentication
-permissions:
-  contents: write
-  actions: write
-  id-token: write
+# Deny by default; the two jobs that need scopes grant them for themselves below.
+permissions: {}
 
 on:
   workflow_dispatch:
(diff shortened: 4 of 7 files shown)


.github/workflows/cleanup-main-builds.yml
```diff
@@ -5,10 +5,16 @@ on:
   release:
     types: [ published ]
 
+# Deny by default so a new job cannot silently pick up the org-wide write default.
+permissions: {}
```
jandroav, line 9: Pattern is right. The gap is that it stops one workflow short.

`default_workflow_permissions` is `write` at both the org and the repo level (`GET /orgs/liquibase/actions/permissions/workflow` and the repo equivalent both return `"write"`), so any workflow with no top-level block still hands out a write-all `GITHUB_TOKEN`.

After this PR, `build-main.yml` is the one kept workflow with neither a top-level block nor per-job scopes on all of its jobs: its `setup` job declares none, so it keeps a write-all token on every push to `main` and on the `5 6 * * *` nightly. Two lines fix it:

```yaml
permissions: {}
```

plus `contents: read` on `setup` (`build-and-publish` already declares its own).

`label-pr.yml` also has no top-level block, but its single job declares `issues: write` / `pull-requests: write`, so the exposure there is bounded and it is only the missing fail-closed default.

Separately, the second half of TECHOPS-1111 (per-job grants instead of one broad top-level grant) is still open on the higher-privilege workflows this PR does not touch: `dry-run-release.yml`, `claude.yml`, `docker-scan.yml`, `build-qa-docker.yml`, `release-docker.yml`, `release-deploy-maven.yml`, `nightly-release.yml`, `dependabot-automerge.yml`. `claude.yml` is the one worth doing next: it grants `issues: write`, `pull-requests: write` and `id-token: write` to every job at the top level, and it is reachable by any commenter on a public issue or PR.

Happy either way, but please finish those here or link a follow-up, so the pattern is not left half-applied.

sayaliM0412: Both gaps closed in ae8ff8ba5, and the remainder is now a linked ticket.

**`build-main.yml`** — confirmed exactly as you describe. No top-level block, and `setup` declared no scopes, so it ran write-all on every push to `main` and on the `5 6 * * *` nightly. It checks out and emits two string outputs, nothing else, so `contents: read` is all it needs. `build-and-publish` already declared `contents: read` / `packages: write` and is untouched.

**`label-pr.yml`** — added the fail-closed default. Its one job already scoped itself, so this is only about a job added later.

Effective permissions now:

| workflow | top | jobs |
|---|---|---|
| `build-main.yml` | `{}` | `setup: contents: read`, `build-and-publish: contents: read, packages: write` |
| `label-pr.yml` | `{}` | `label: issues: write, pull-requests: write` |

**No `persist-credentials` change on either.** This PR sets that only where a run builds fork-controlled code, which is the argument made on `codeql.yml`'s `pull_request` trigger. `build-main` runs on `workflow_dispatch`, `schedule` and push to `main`, and `label-pr` checks nothing out, so neither qualifies and I did not want to widen the pattern past its stated reason.

`actionlint`: unchanged on both files (16 pre-existing on `build-main`, 0 on `label-pr`).

---

On the second half of TECHOPS-1111: **TECHOPS-1215**, linked as *blocks* TECHOPS-1111 and *relates to* TECHOPS-1188. I measured the current state of all eight before filing rather than copying your list across:

| workflow | top-level grant | jobs inheriting it |
|---|---|---|
| `dry-run-release.yml` | actions/contents/packages/id-token: write | all 5 |
| `claude.yml` | pull-requests/issues/id-token: write | 1 of 1 |
| `docker-scan.yml` | security-events/id-token/packages/pull-requests: write | 2 of 3 |
| `build-qa-docker.yml` | id-token/packages/security-events: write | 4 of 6 |
| `release-docker.yml` | contents/id-token/packages: write | 0 of 1, already scoped |
| `release-deploy-maven.yml` | contents/packages/id-token: write | both |
| `nightly-release.yml` | contents: write | 1 of 1 |
| `dependabot-automerge.yml` | contents/pull-requests: write | 1 of 1 |

The ticket calls out `claude.yml` as the one to do first, for the reason you gave: `secrets: inherit`, three write scopes at the top level, and no `author_association` check behind the `@claude` gate. That is the same exposure that led to dropping `ref:refs/heads/main` from the trust policy in liquibase-infrastructure#4279.

Kept them out of this PR so the diff stays reviewable against the cut-over set rather than becoming a repo-wide permissions sweep.

sayaliM0412:
NOTE: the failing check is due to the incorrect `INSTALL4J_LICENSE` key instead of `INSTALL4J_LICENSE_KEY`. But the workflow is disabled and soon will be removed.

jandroav (approved):

PR #7957, opened 2026-08-28
Fix concurrent updates leaving each other's changelog locks set by resetting only the finishing database's LockService

### LINKED ISSUE #7956: A finished update discards every database's LockService, leaving concurrent updates' changelog locks set
### Search first

- [x] I searched and no similar issues were found. The closest are #7220 (same method, different field, fixed) and #4173 (this exact mechanism, closed `not_planned`) — both discussed under *Additional Context*.

### Description

`LockServiceFactory` is a JVM-wide singleton, and two `cleanUp()` paths on the `update` pipeline end by calling `LockServiceFactory.getInstance().resetAll()`. That nulls the singleton and with it **every** database's `LockService` — not only the one belonging to the execution that is finishing.

So two `update`s running concurrently in one JVM corrupt each other's lock bookkeeping, **even against entirely separate databases**. Whichever finishes first wipes the factory. The still-running peer then looks its lock service up again, gets a *freshly built* one whose `hasChangeLogLock()` is `false`, and **skips** its release — the release is not attempted and failed, it is skipped, so not even a log line is produced. That database's `DATABASECHANGELOGLOCK` row stays `LOCKED = TRUE` with nobody holding it.

The cost lands on the next start against that database: it waits out the full `changeLogLockWaitTime` (5 minutes by default), throws `LockException`, and retries — for as long as whatever recovers stale locks takes to notice.

The two call sites, in 5.0.4:

| File | Line | Context |
|---|---|---|
| `command/core/AbstractUpdateCommandStep.java` | 211 | in `cleanUp()`, one line above the two calls #7909 already converted to per-database |
| `command/core/helpers/LockServiceCommandStep.java` | 67 | in `cleanUp()`, right after its own guarded release; fires for every update subclass that takes its lock from the pipeline |

Converting only the first leaves the bug reachable, so both need to move.

The `hasChangeLogLock()` guards those two releases sit behind are not the problem — their comments cite #5438 and they are deliberate. The problem is the factory being wiped out from under them.

**A second, independent defect in `resetAll()` itself**, visible even single-threaded: it iterates `registry` — the prototype instances the service locator discovered, which never hold a lock — calls `reset()` on those, and then nulls the singleton. It never touches `openLockServices`, i.e. the per-database services that `getLockService()` actually hands out and that actually hold the locks. A full teardown today therefore resets the wrong objects and orphans the right ones. This is the same shape as the `services.values()` gap raised in review on #7909 and fixed there.

### Steps To Reproduce

Self-contained, no Spring: two H2 in-memory databases, `liquibase-core` + `h2` only. The interleaving is decided rather than raced for — the "parked" update stops *inside* its update, while Liquibase still holds its changelog lock, and stays there while its peer runs a whole update of its own against a different database.

<details>
<summary><code>LockLeakRepro.java</code></summary>

```java
package repro;

import java.sql.Connection;
import java.sql.DriverManager;
import java.sql.ResultSet;
import java.sql.Statement;
import java.util.Map;
import liquibase.Scope;
import liquibase.command.CommandScope;
import liquibase.command.core.UpdateCommandStep;
import liquibase.command.core.helpers.DbUrlConnectionArgumentsCommandStep;
import liquibase.resource.ClassLoaderResourceAccessor;

public final class LockLeakRepro {

    private static final String PARKED_URL = "jdbc:h2:mem:parked;DB_CLOSE_DELAY=-1";
    private static final String PEER_URL = "jdbc:h2:mem:peer;DB_CLOSE_DELAY=-1";

    public static void main(String[] args) throws Exception {
        Thread parked = new Thread(() -> runUpdate(PARKED_URL, "changelog-parked.xml"), "parked");
        parked.start();

        ParkingChange.PARKED.await();          // parked run is inside its update, holding its lock
        runUpdate(PEER_URL, "changelog-peer.xml");   // peer completes; its cleanUp() calls resetAll()
        ParkingChange.RELEASE.countDown();     // parked run resumes and finishes
        parked.join();

        boolean leaked = isLockHeld(PARKED_URL);
        System.out.println("parked database's DATABASECHANGELOGLOCK still locked: " + leaked);
        System.out.println(leaked ? "RESULT: *** LOCK LEAKED ***" : "RESULT: ok, lock released");
        System.exit(leaked ? 1 : 0);
    }

    private static void runUpdate(String url, String changelog) {
        try {
            Scope.child(Map.of(Scope.Attr.resourceAccessor.name(),
                    new ClassLoaderResourceAccessor(LockLeakRepro.class.getClassLoader())), () ->
                    new CommandScope(UpdateCommandStep.COMMAND_NAME)
                            .addArgumentValue(DbUrlConnectionArgumentsCommandStep.URL_ARG, url)
                            .addArgumentValue(DbUrlConnectionArgumentsCommandStep.USERNAME_ARG, "sa")
                            .addArgumentValue(DbUrlConnectionArgumentsCommandStep.PASSWORD_ARG, "")
                            .addArgumentValue(UpdateCommandStep.CHANGELOG_FILE_ARG, changelog)
                            .execute());
        } catch (Exception e) {
            throw new IllegalStateException("update against " + url + " failed", e);
        }
    }

    private static boolean isLockHeld(String url) throws Exception {
        try (Connection connection = DriverManager.getConnection(url, "sa", "");
             Statement statement = connection.createStatement();
             ResultSet rows = statement.executeQuery("SELECT LOCKED FROM DATABASECHANGELOGLOCK")) {
            return rows.next() && rows.getBoolean(1);
        }
    }

    private LockLeakRepro() {
    }
}
```
</details>

<details>
<summary><code>ParkingChange.java</code></summary>

```java
package repro;

import java.util.concurrent.CountDownLatch;
import liquibase.change.custom.CustomTaskChange;
import liquibase.database.Database;
import liquibase.exception.ValidationErrors;
import liquibase.resource.ResourceAccessor;

/** Parks inside the update, while the changelog lock is held, until the test lets it go. */
public class ParkingChange implements CustomTaskChange {

    static final CountDownLatch PARKED = new CountDownLatch(1);
    static final CountDownLatch RELEASE = new CountDownLatch(1);

    @Override
    public void execute(Database database) {
        PARKED.countDown();
        try {
            RELEASE.await();
        } catch (InterruptedException e) {
            Thread.currentThread().interrupt();
        }
    }

    @Override public String getConfirmationMessage() { return "parked"; }
    @Override public void setUp() { }
    @Override public void setFileOpener(ResourceAccessor resourceAccessor) { }
    @Override public ValidationErrors validate(Database database) { return new ValidationErrors(); }
}
```
</details>

<details>
<summary><code>changelog-parked.xml</code> / <code>changelog-peer.xml</code></summary>

```xml
<!-- changelog-parked.xml -->
                       http://www.liquibase.org/xml/ns/dbchangelog/dbchangelog-latest.xsd">
    <changeSet id="1" author="repro">
        <createTable tableName="parked_table"><column name="id" type="int"/></createTable>
    </changeSet>
    <changeSet id="2" author="repro">
        <customChange class="repro.ParkingChange"/>
    </changeSet>
</databaseChangeLog>
```

```xml
<!-- changelog-peer.xml -->
                       http://www.liquibase.org/xml/ns/dbchangelog/dbchangelog-latest.xsd">
    <changeSet id="1" author="repro">
        <createTable tableName="peer_table"><column name="id" type="int"/></createTable>
    </changeSet>
</databaseChangeLog>
```
</details>

On 5.0.4:

```
Liquibase 5.0.4
parked database's DATABASECHANGELOGLOCK still locked: true
RESULT: *** LOCK LEAKED ***
```

Note that the **skew is what matters**: a symmetric two-thread repro does not leak. One run has to finish while the other is still between acquiring and releasing its lock.

### Expected/Desired Behavior

An `update` that finishes invalidates only its own database's `LockService`, and a concurrent `update` against a different database is left able to release the lock it still holds — so no `DATABASECHANGELOGLOCK` row is left set and no subsequent start has to wait out `changeLogLockWaitTime`.

Concretely, mirroring what #7909 did for the sibling factory: a `LockServiceFactory#resetDatabase(Database)` that resets and drops just that database's entry, called from the two `cleanUp()` paths above; plus `resetAll()` extended to reset `openLockServices` before dropping them, so a genuine full teardown stops resetting only the prototypes.

### Liquibase Version

5.0.4 (both call sites are long-standing; `resetAll()`'s registry-only loop predates them)

### Database Vendor & Version

Database-independent — the mechanism is entirely in Liquibase's in-JVM lock bookkeeping. Reproduced above on H2 2.3.232.

### Liquibase Integration

Java API (`CommandScope`) and Spring Boot (`SpringLiquibase`) — anything that runs more than one `update` per JVM.

### Liquibase Extensions

None.

macOS 15 and Linux (CI containers).

### Additional Context

**Why this is worth fixing rather than working around.** It prevents parallel schema initialization in a multi-database setup; falling back to sequential startup can take considerably longer.

**Prior art, and how this differs from it.**

- **#7220** is the strongest precedent and the reason this is worth filing narrowly: Apache Fineract, 100–200 tenants each on its own database, one thread per tenant, *"the lock table gets stuck for some of the tenants"* on 80–90% of runs. That was traced to a **different** field in `LockServiceCommandStep#cleanUp` (`isDBLocked`), fixed by making it a `ThreadLocal`, and closed COMPLETED. The fix is in 5.0.4. So that one method had two independent routes to a stuck lock; one is fixed and this one is not.
- **#7909** fixed exactly this anti-pattern for `ChangeLogHistoryServiceFactory` and `ExecutorService` in `AbstractUpdateCommandStep#cleanUp`. `LockServiceFactory.resetAll()` is *unchanged context* in that diff, and its review explicitly asked for a follow-up (quoted above).
- **#4173** described this same mechanism in 2023 and was closed `not_planned`. It asked for a redesign of both factories' lifecycles; this issue asks for the narrow, already-established per-database conversion instead.

**Two further `resetAll()` call sites, deliberately left out of scope here**, both because they are genuine full teardowns and because touching them is not free:

- `DropAllCommandStep:152` — named in the #7909 review. It runs after `lockService.destroy()` has dropped the lock table, and it sits in a `protected resetServices()` that Liquibase Pro subclasses; changing that signature is a separate decision.
- `Liquibase.java:990` (`resetServices()` on the facade) — arguably a legitimate full reset of a facade being torn down, and also `protected`.

Happy to have either or both folded in if you'd prefer one PR.

- [x] I'm willing to submit a PR (Thank you!) — it is ready, with unit tests that each fail when their production change is reverted, and I'll link it here.

houssain-barouni:
Fixes #7956

## Impact

## Description

`LockServiceFactory` is a JVM-wide singleton, and two `cleanUp()` paths on the `update` pipeline ended by calling `LockServiceFactory.getInstance().resetAll()` — which nulls that singleton and with it **every** database's `LockService`, not just the one belonging to the execution that is finishing.

Two `update`s running concurrently in one JVM therefore corrupt each other's lock bookkeeping, even against entirely separate databases. Whichever finishes first wipes the factory; the still-running peer then looks its lock service up again, gets a *freshly built* one whose `hasChangeLogLock()` is `false`, and **skips** its release — not attempted-and-failed, skipped, so nothing is logged. That database's `DATABASECHANGELOGLOCK` row stays set, and the next start against it waits out the full `changeLogLockWaitTime` before throwing `LockException`. #7956 has a self-contained two-H2-database reproduction.

This is the same anti-pattern #7909 fixed for `ChangeLogHistoryServiceFactory` and `ExecutorService`; `LockServiceFactory.resetAll()` is unchanged context in that diff, and its review asked for exactly this follow-up.

The change is the same shape as #7909's:

1. **`LockServiceFactory#resetDatabase(Database)`** — resets and drops just that database's entry.
2. **Both call sites converted** to it: `AbstractUpdateCommandStep#cleanUp` (5.0.4 line 211) and `LockServiceCommandStep#cleanUp` (line 67). Converting only the first leaves the bug reachable — the second fires for every update subclass that takes its lock from the pipeline.
3. **`resetAll()` fixed too.** It iterated only `registry` — the prototypes the service locator discovered, which never hold a lock — called `reset()` on those and then nulled the singleton, never touching `openLockServices`, i.e. the per-database services `getLockService()` hands out and that actually hold the locks. So a full teardown reset the wrong objects and orphaned the right ones. This is a correctness improvement even single-threaded, and the same gap that was raised in review on #7909 for `services.values()`.

The `hasChangeLogLock()` guards in front of both releases are untouched — their comments cite #5438 and they are deliberate. The bug was the factory being wiped from under them, not the guards.

## Release note

Running two Liquibase updates at the same time in one JVM , each against its own database — no longer leaves a changelog lock behind. Previously, whichever update finished first could stop the other from releasing its lock, so that database's next update waited out the full changelog lock timeout and then failed.

## Things to be aware of

**`resetDatabase()` removes the entry rather than resetting it in place**, which is the opposite of the choice `ChangeLogHistoryServiceFactory#resetDatabase` made. The reason that one resets in place is that some of its services cannot be rebuilt — `OfflineConnection` registers a pre-configured one via `registerForDatabase`. `LockServiceFactory` has no such path: every service it hands out is built on demand by `getLockService()`, so nothing is lost by dropping it, and it keeps the map from accumulating an entry per `Database` for the life of the JVM (the concern noted in #7909's review, which does not arise here). Nothing outside the `lockservice` package configures a `LockService` instance that would need to survive.

**No new synchronization.** `openLockServices` is a `ConcurrentHashMap` and `remove` is atomic; adding `synchronized` to `resetDatabase` alone would not give mutual exclusion against the unsynchronized `getLockService`, so it would be misleading rather than safer.

**Two other `resetAll()` call sites are left alone**, both genuine full teardowns:
- `DropAllCommandStep:152` — named in #7909's review, but it runs after `lockService.destroy()` has dropped the lock table, and it lives in a `protected resetServices()` that Liquibase Pro subclasses, so changing that signature is a separate decision.
- `Liquibase.java:990` — `resetServices()` on the facade, arguably a legitimate full reset of a facade being torn down, and also `protected`.

Happy to fold either in if you'd prefer one PR.

## Things to worry about

Nothing outstanding. The one judgement call is remove-vs-reset-in-place above; if you'd rather have the two factories behave identically, resetting in place is a one-line change, at the cost of the map growing per `Database`.

## Additional Context

**Tests** — each one fails when its production change is reverted, verified individually:

| Test | Reverting… | Failure |
|---|---|---|
| `LockServiceFactoryTest#resetDatabaseLeavesOtherDatabasesLockServicesIntact` | `resetDatabase` → `resetAll` | `expected same: FakeLockService@… was not: FakeLockService@…` |
| `LockServiceFactoryTest#resetAllResetsTheServicesThatActuallyHoldLocks` | the `openLockServices` loop | `expected:<1> but was:<0>` |
| `AbstractUpdateCommandStepTest#"cleanUp only drops the LockService for its own database"` | that call site | `Condition not satisfied: …getLockService(databaseB).is(lockServiceB)` |
| `LockServiceCommandStepTest#"cleanUp drops only its own database's lock service…"` | that call site | `Condition not satisfied: …getLockService(databaseB).is(serviceB)` |

Plus `resetDatabaseResetsAndDropsTheGivenDatabasesLockService` and `resetDatabaseOnAnUnknownDatabaseIsANoOp` covering the new method's contract.

**Regressions ruled out** — the full `liquibase-standard` suite passes on this branch: 6492 tests, 0 failures, 12 skipped (unchanged from `main`).

**Verified end-to-end against a real downstream failure.** An application that initializes multiple databases, each on its own thread, hit exactly this and had to serialize its Liquibase invocations. With the serialization removed and this branch installed in place of 5.0.4, its deterministic H2 regression test — which parks one execution inside its update while a peer completes a whole migration — goes from failing on a leaked changelog lock to passing, and its full 1680-test module suite is green. The standalone reproduction in #7956 likewise flips from `*** LOCK LEAKED ***` to `ok, lock released`.

### DIFF
--- a/liquibase-standard/src/main/java/liquibase/command/core/AbstractUpdateCommandStep.java
+++ b/liquibase-standard/src/main/java/liquibase/command/core/AbstractUpdateCommandStep.java
@@ -208,8 +208,8 @@ public abstract class AbstractUpdateCommandStep extends AbstractCommandStep impl
     @Override
     public void cleanUp(CommandResultsBuilder resultsBuilder) {
         isDBLocked.remove();
-        LockServiceFactory.getInstance().resetAll();
         Database database = (Database) resultsBuilder.getCommandScope().getDependency(Database.class);
+        LockServiceFactory.getInstance().resetDatabase(database);
         Scope.getCurrentScope().getSingleton(ChangeLogHistoryServiceFactory.class).resetDatabase(database);
         ExecutorService executorService = Scope.getCurrentScope().getSingleton(ExecutorService.class);
         executorService.clearExecutor("jdbc", database);
--- a/liquibase-standard/src/main/java/liquibase/command/core/helpers/LockServiceCommandStep.java
+++ b/liquibase-standard/src/main/java/liquibase/command/core/helpers/LockServiceCommandStep.java
@@ -49,22 +49,24 @@ public class LockServiceCommandStep extends AbstractHelperCommandStep implements
     public void cleanUp(CommandResultsBuilder resultsBuilder) {
         try {
             if (isDBLocked.get()) {
+                Database database = (Database) resultsBuilder.getCommandScope().getDependency(Database.class);
                 try {
                     // Check hasChangeLogLock() rather than releasing unconditionally: some
                     // AbstractUpdateCommandStep subclasses (e.g. updateCount/updateSql/updateToTag) also
                     // release the lock themselves, from within run(), before this cleanUp() runs -- see
                     // #5438. Releasing again here without checking live state would fail against an
                     // already-unlocked row.
-                    LockService lockService = LockServiceFactory.getInstance().getLockService(
-                            (Database) resultsBuilder.getCommandScope().getDependency(Database.class)
-                    );
+                    LockService lockService = LockServiceFactory.getInstance().getLockService(database);
                     if (lockService.hasChangeLogLock()) {
                         lockService.releaseLock();
                     }
                 } catch (LockException e) {
                     Scope.getCurrentScope().getLog(getClass()).severe(Liquibase.MSG_COULD_NOT_RELEASE_LOCK, e);
                 }
-                LockServiceFactory.getInstance().resetAll();
+                // Drop only this database's lock service, not every database's: resetAll() would discard
+                // the singleton and with it the lock services of any execution running concurrently
+                // against another database, leaving them unable to release the lock they still hold.
+                LockServiceFactory.getInstance().resetDatabase(database);
             }
         } finally {
             isDBLocked.remove();
--- a/liquibase-standard/src/main/java/liquibase/lockservice/LockServiceFactory.java
+++ b/liquibase-standard/src/main/java/liquibase/lockservice/LockServiceFactory.java
@@ -75,10 +75,40 @@ public class LockServiceFactory {
         return openLockServices.get(database);
     }
 
+    /**
+     * Invalidates and drops the {@link LockService} cached for just the given database, as opposed to
+     * {@link #resetAll()} which drops every database's along with this factory's JVM-wide singleton.
+     * Cleanup paths belonging to a single execution should use this, so they don't discard other
+     * concurrently running executions' (different database's) lock services: a service dropped while
+     * its execution still holds the changelog lock reports {@link LockService#hasChangeLogLock()} as
+     * false, so that execution skips its release and leaves the DATABASECHANGELOGLOCK row set.
+     * <p>
+     * The entry is removed rather than reset in place: unlike
+     * {@link liquibase.changelog.ChangeLogHistoryServiceFactory}, every service this factory hands out
+     * is built on demand by {@link #getLockService(Database)}, so nothing is lost by dropping it, and
+     * keeping the map free of finished databases keeps it from growing for the life of the JVM.
+     */
+    public void resetDatabase(Database database) {
+        LockService lockService = openLockServices.remove(database);
+        if (lockService != null) {
+            lockService.reset();
+        }
+    }
+
+    /**
+     * Resets every {@link LockService} this factory knows about and drops the JVM-wide singleton. This
+     * is a full teardown; cleanup paths belonging to a single execution should use
+     * {@link #resetDatabase(Database)} instead.
+     */
     public void resetAll() {
         for (LockService lockService : registry) {
             lockService.reset();
         }
+        // registry holds only the prototypes discovered by the service locator, never the per-database
+        // instances getLockService() creates from them -- which are the ones that actually hold locks.
+        // reset() is idempotent, so any overlap between the two is harmless.
+        openLockServices.values().forEach(LockService::reset);
+        openLockServices.clear();
         reset();
     }
 }
--- a/liquibase-standard/src/test/groovy/liquibase/command/core/AbstractUpdateCommandStepTest.groovy
+++ b/liquibase-standard/src/test/groovy/liquibase/command/core/AbstractUpdateCommandStepTest.groovy
@@ -177,4 +177,30 @@ class AbstractUpdateCommandStepTest extends Specification {
         !executorService.getExecutor("jdbc", databaseA).is(overrideA)
         resets["A"] == 1
     }
+
+    def "cleanUp only drops the LockService for its own database"() {
+        // Same JVM-wide-singleton hazard as the test above, for LockServiceFactory: cleanUp() used to
+        // call the blanket LockServiceFactory.resetAll(), which nulls the singleton and with it every
+        // database's LockService. A concurrently running execution then looked its lock service up
+        // again, got a freshly built one whose hasChangeLogLock() is false, skipped its release without
+        // logging anything, and left its DATABASECHANGELOGLOCK row set.
+        given: "two independent executions, each with an open lock service"
+        def databaseA = new PostgresDatabase()
+        def databaseB = new PostgresDatabase()
+        def lockServiceA = LockServiceFactory.getInstance().getLockService(databaseA)
+        def lockServiceB = LockServiceFactory.getInstance().getLockService(databaseB)
+
+        def commandA = new CommandScope(UpdateSqlCommandStep.COMMAND_NAME)
+                .provideDependency(Database.class, databaseA)
+        def resultsBuilderA = new CommandResultsBuilder(commandA, new ByteArrayOutputStream())
+
+        when: "execution A finishes and cleans up while execution B is still in flight"
+        new UpdateSqlCommandStep().cleanUp(resultsBuilderA)
+
+        then: "B keeps the very lock service that knows it holds the changelog lock"
+        LockServiceFactory.getInstance().getLockService(databaseB).is(lockServiceB)
+
+        and: "A's own lock service is dropped, its execution being over"
+        !LockServiceFactory.getInstance().getLockService(databaseA).is(lockServiceA)
+    }
 }
--- a/liquibase-standard/src/test/groovy/liquibase/command/core/helpers/LockServiceCommandStepTest.groovy
+++ b/liquibase-standard/src/test/groovy/liquibase/command/core/helpers/LockServiceCommandStepTest.groovy
@@ -3,8 +3,10 @@ package liquibase.command.core.helpers
 import liquibase.command.CommandResultsBuilder
 import liquibase.command.CommandScope
 import liquibase.database.Database
+import liquibase.database.core.PostgresDatabase
 import liquibase.lockservice.LockService
 import liquibase.lockservice.LockServiceFactory
+import liquibase.lockservice.StandardLockService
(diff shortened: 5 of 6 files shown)


filipelautert (approved):
Hello @houssain-barouni ! Nice catch, thanks for the thorough writeup. Verified locally: the new tests fail with the fix reverted, and the #7956 repro leaks on 5.0.4, does not leak on 5.0.3, and is fixed on this branch. So this is a 5.0.4 regression introduced by #7886's hasChangeLogLock() guard, not a long-standing bug.

houssain-barouni:
> Hello @houssain-barouni ! Nice catch, thanks for the thorough writeup. Verified locally: the new tests fail with the fix reverted, and the #7956 repro leaks on 5.0.4, does not leak on 5.0.3, and is fixed on this branch. So this is a 5.0.4 regression introduced by #7886's hasChangeLogLock() guard, not a long-standing bug.

Thanks @filipelautert for the review , indeed we don't reproduce the issue with 5.0.3. 
Would it be possible to merge the fix, so that it gets into the next patch version?

PR #7960, opened 2026-09-01
Fix: clear an old occurrence of a tag before applying it again

### LINKED ISSUE #3763: tagDatabase tag allows for duplicate tag values

om7057:
Fixes #3763 (the duplicate-tag part of the issue).

Re-running `tag` with a name that was already used left the tag on the earlier row as well as the new one, since `TagDatabaseGenerator` only ever added the tag to the latest changeset row and never cleared it from wherever it used to be. This adds a `TAG = NULL WHERE TAG = <tag>` statement before the existing per-database tagging SQL, so a tag always identifies exactly one changeset regardless of how many times it's reused.

Left the feature-request half of the issue (making tag reuse configurable) out of scope — this only fixes the bug nvoxland described.

## Testing
Updated `TagDatabaseGeneratorTest` for the extra statement and asserted its exact SQL.

### DIFF
--- a/liquibase-standard/src/main/java/liquibase/changelog/StandardChangeLogHistoryService.java
+++ b/liquibase-standard/src/main/java/liquibase/changelog/StandardChangeLogHistoryService.java
@@ -454,6 +454,14 @@ public class StandardChangeLogHistoryService extends AbstractChangeLogHistorySer
         getDatabase().commit();
 
         if (this.ranChangeSetList != null) {
+            // Clear the tag off whichever entry already carries it, mirroring what the SQL above just did
+            // to the DATABASECHANGELOG table -- otherwise a re-tag leaves two in-memory entries with the
+            // same tag until the cache is next rebuilt (#3763).
+            for (RanChangeSet ranChangeSet : ranChangeSetList) {
+                if (tagString.equals(ranChangeSet.getTag())) {
+                    ranChangeSet.setTag(null);
+                }
+            }
             ranChangeSetList.get(ranChangeSetList.size() - 1).setTag(tagString);
         }
     }
--- a/liquibase-standard/src/main/java/liquibase/sqlgenerator/core/TagDatabaseGenerator.java
+++ b/liquibase-standard/src/main/java/liquibase/sqlgenerator/core/TagDatabaseGenerator.java
@@ -33,8 +33,16 @@ public class TagDatabaseGenerator extends AbstractSqlGenerator<TagDatabaseStatem
             String dateColumnNameEscaped = database.escapeObjectName("DATEEXECUTED", Column.class);
             String tagColumnNameEscaped = database.escapeObjectName("TAG", Column.class);
             String tagEscaped = DataTypeFactory.getInstance().fromObject(statement.getTag(), database).objectToSql(statement.getTag(), database);
+
+            // Clear the tag off any row that already carries it before tagging the latest row, so re-using a
+            // tag name doesn't leave it applied to multiple changesets (#3763).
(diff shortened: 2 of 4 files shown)


om7057:
That's a pre-existing limitation of the default branch (already flagged by the TODO above it), not something this change introduces or makes worse; the clear step is correct regardless of how many rows end up getting the new tag. Fixing the underlying non-deterministic row selection is a separate, harder problem (no ORDEREXECUTED-equivalent tiebreak on arbitrary databases) and out of scope for this PR, which is just the duplicate-tag bug from #3763.

renechoi (commented):
The changelog table is fixed but the in-memory history is not. `StandardChangeLogHistoryService.tag()` only sets the tag on the last `ranChangeSetList` entry (L456-458) and never clears it from the entry that held it before, and `ChangeLogHistoryServiceFactory` caches that service per `Database`. So after a re-tag, a `rollback` later in the same JVM still sees two entries carrying the tag. `rollback` defaults to `tagVersion=OLDEST`, and `AfterTagChangeSetFilter.oldestVersion()` stops at the first match, so it resolves to the row this statement just cleared.

om7057:
Good catch, fixed. tag() now clears the tag off any cached ranChangeSetList entry that already carries it too, same as the SQL side. Added a test against a real H2-backed service that tags twice with the same name and checks only the latest cached entry keeps it (fails without the fix, reports 2 entries instead of 1).

PR #7964, opened 2026-09-01
[TECHOPS-1214] Move the release publishing jobs onto environment: release

sayaliM0412:
## Impact

## Description

Step 3 of the release vault OIDC migration. `liquibase-release-vault-oidc-role` was created by [liquibase/liquibase-infrastructure#4279](https://github.com/liquibase/liquibase-infrastructure/pull/4279) but nothing assumed it: every publishing job still presented `LIQUIBASE_VAULT_OIDC_ROLE_ARN`, the broad `repo:liquibase/*:*` role, so the new role had zero traffic and the CloudTrail acceptance criterion could never have passed.

Two edits per job, both required for the role to be reachable:

- an `environment:`, so the OIDC subject becomes `environment:<name>`;
- `role-to-assume` switched to `LIQUIBASE_RELEASE_VAULT_OIDC_ROLE_ARN`.

### 1. The environment goes in the callees, not the orchestrator

The ticket originally called for adding `environment: release` to the orchestrator's publishing jobs. Not expressible: those jobs are reusable-workflow calls, and the workflow JSON schema's reusable-workflow-call job allows only `name`, `needs`, `permissions`, `if`, `uses`, `with`, `secrets`, `strategy`, `concurrency`, with `additionalProperties: false`. GitHub's prose docs never state this. So it sits on the jobs inside the callees.

### 2. Two environments, not one, because GitHub approves per job

The first version of this PR put `environment: release` on all four publishing jobs. That was wrong, and the review caught it. **Approval is granted per job, not per run**, so a single reviewer-gated environment stops the release once per wave of the orchestrator's `needs` graph:

```
manual-approval  ->  javadocs + xsd  ->  s3 (after package)  ->  maven (after javadocs/xsd/gh-packages)
```

Four prompts where there is one today. That is the approval fatigue [TECHOPS-1103](https://datical.atlassian.net/browse/TECHOPS-1103) rejected, and a release that stops four times in front of people not expecting it gets rubber-stamped.

The fix separates the two jobs one environment was doing. `release` keeps the five reviewers and stays the gate on `manual-approval`. `release-publish` ([liquibase/liquibase-infrastructure#4323](https://github.com/liquibase/liquibase-infrastructure/pull/4323)) has the same deployment policies and no reviewers, and exists only to emit an `environment:` OIDC subject, which a job emits with or without protection rules.

### 3. The environment is selected, not fixed, so direct dispatch stays gated

```yaml
environment:
  name: ${{ inputs.approved && 'release-publish' || 'release' }}
```

`approved` is declared **only under `workflow_call`**, never under `workflow_dispatch`, so it is not something a person can set. The orchestrator passes `approved: true` on the four calls it makes after `manual-approval` has cleared. Anything else falls back to `release` and its reviewers.

That fallback is the point. All four callees carry their own `workflow_dispatch`, and `manual-approval` lives in the orchestrator, not in the callee's `needs` graph, so a direct dispatch has never passed a gate. It is a live path, not a theoretical one: `release-deploy-maven.yml` has been dispatched directly 6 times, most recently 2026-08-23. Today it reaches the broad vault role with no approval at all.

| | orchestrated release | direct dispatch of a callee |
|---|---|---|
| today | 1 approval | **0** |
| first version of this PR | ~4 approvals | 1 approval |
| **this** | **1 approval** | **1 approval** |

## Scope

The four production jobs, all already gated on `dry_run == false`:

| file | job |
|---|---|
| `release-deploy-javadocs.yml` | `deploy-javadocs` |
| `release-deploy-maven.yml` | `deploy-maven-production` |
| `release-deploy-xsd.yml` | `deploy-xsd` |
| `release-publish-assets-s3.yml` | `publish-assets-s3` |

Plus `release-published-orchestrator.yml`, which passes `approved: true` on those four calls.

`deploy-maven-dryrun` is deliberately left on the old role with no environment: dry runs skip `manual-approval` and should not sit behind required reviewers. The second `configure-aws-credentials` step in three of these files uses `AWS_PROD_GITHUB_OIDC_ROLE_ARN_BUILD_LOGIC`, a different role with its own trust, and is untouched.

## Things to be aware of

**Merge order.** [liquibase/liquibase-infrastructure#4323](https://github.com/liquibase/liquibase-infrastructure/pull/4323) must be merged **and applied** first. `vault-manager` has no autodeploy, so its run needs confirming by hand. Until it does, the role does not trust `environment:release-publish` and a release fails with an STS `AccessDenied` that reads as unrelated.

**`create-release.yml` (2 sites) and `docker-release.yml` (1 site) still use the broad role.** Raised in review, verified on main. Outside this PR's scope, but it means [TECHOPS-1110](https://datical.atlassian.net/browse/TECHOPS-1110) does not close when this lands: the job holding the GPG and DigiCert signing credentials is still on `repo:liquibase/*:*`. Being raised separately.

**build-logic `package.yml` still uses the broad role.** Its `upload_packages` and `create-placeholder-branch` jobs assume `LIQUIBASE_VAULT_OIDC_ROLE_ARN`. It is shared by many repos, so scoping it to this environment would impose these reviewers on every other consumer. Needs its own change.

**Do not delete the tag subjects yet.** Step 5 removes `refs/tags/v*` from the role's trust. That waits until CloudTrail confirms the role is assumed on an `environment:` subject, on both a real release and a `workflow_dispatch` run.

**Any new release-path job that reads the vault needs two edits, not one.** The `environment:` block on the job inside the callee, never on the orchestrator, and `role-to-assume` pointed at the release role. Either edit without the other is a silent no-op.

## How was it tested?

- `actionlint` on all five files: **2 / 2 / 2 / 1 / 2 findings before and after**, byte-identical, all pre-existing shellcheck `SC2086` noise plus one pre-existing `required` + `default` warning on the orchestrator.
- YAML re-parsed to confirm `approved` is absent from every `workflow_dispatch` input list, that the environment expression resolves on exactly the four intended jobs, and that `deploy-maven-dryrun` still declares no environment.
- Expression support in `environment.name` confirmed against the workflow syntax reference: allowed contexts are `github`, `inputs`, `vars`, `needs`, `strategy`, `matrix`.
- Verified against the live `release` environment that the deployment policies admit both triggers, `branch main` and `tag v*`.

Not exercised end to end: the environment OIDC path has never been used anywhere in the org, so the first release is also its first test. A failure there surfaces as an STS `AccessDenied`.

## Release note

[TECHOPS-1103]: https://datical.atlassian.net/browse/TECHOPS-1103?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/.github/workflows/release-deploy-maven.yml
+++ b/.github/workflows/release-deploy-maven.yml
@@ -74,9 +74,8 @@ jobs:
     name: Deploy to Maven Central (Production)
     runs-on: ubuntu-22.04
     if: ${{ inputs.dry_run == false }}
-    # Split so one release costs one approval: an orchestrated run has already cleared the
-    # `release` gate and passes `approved`, while a direct `workflow_dispatch` cannot set that input
-    # (it exists only under `workflow_call`) and so still lands on the reviewer-gated `release`.
+    # `approved` is `workflow_call`-only, so a direct dispatch cannot claim it and still meets the
+    # reviewers on `release`. An orchestrated run cleared that gate once already.
     environment:
       name: ${{ inputs.approved && 'release-publish' || 'release' }}
     steps:
--- a/.github/workflows/release-deploy-maven.yml
+++ b/.github/workflows/release-deploy-maven.yml
@@ -76,6 +76,8 @@ jobs:
     if: ${{ inputs.dry_run == false }}
     # `approved` is `workflow_call`-only, so a direct dispatch cannot claim it and still meets the
     # reviewers on `release`. An orchestrated run cleared that gate once already.
+    # Keep it that way: on `workflow_dispatch` anyone who can dispatch could pick `release-publish`,
+    # and a non-`boolean` input would do it silently because the string "false" is truthy.
     environment:
       name: ${{ inputs.approved && 'release-publish' || 'release' }}
     steps:
--- a/.github/workflows/release-deploy-maven.yml
+++ b/.github/workflows/release-deploy-maven.yml
@@ -68,6 +68,8 @@ jobs:
     name: Deploy to Maven Central (Production)
     runs-on: ubuntu-22.04
     if: ${{ inputs.dry_run == false }}
+    # Emits the `environment:release` OIDC subject, which is what the release-scoped vault role trusts.
+    environment: release
     steps:
       - name: Download release assets
         env:
@@ -78,7 +80,7 @@ jobs:
       - name: Configure AWS credentials for vault access
         uses: aws-actions/configure-aws-credentials@e6de054238d6b7531b4efff3b6587d9aade6a06c # v6.2.3
         with:
-          role-to-assume: ${{ secrets.LIQUIBASE_VAULT_OIDC_ROLE_ARN }}
+          role-to-assume: ${{ secrets.LIQUIBASE_RELEASE_VAULT_OIDC_ROLE_ARN }}
           aws-region: us-east-1
 
       - name: Get secrets from vault
--- a/.github/workflows/release-deploy-maven.yml
+++ b/.github/workflows/release-deploy-maven.yml
@@ -29,6 +29,12 @@ on:
         type: string
         default: ""
 
+      approved:
+        description: "True only when release-published-orchestrator has already cleared the manual-approval gate for this run."
+        required: false
+        type: boolean
+        default: false
+
   workflow_dispatch:
     inputs:
       version:
@@ -68,8 +74,11 @@ jobs:
     name: Deploy to Maven Central (Production)
     runs-on: ubuntu-22.04
     if: ${{ inputs.dry_run == false }}
-    # Emits the `environment:release` OIDC subject, which is what the release-scoped vault role trusts.
-    environment: release
+    # Split so one release costs one approval: an orchestrated run has already cleared the
+    # `release` gate and passes `approved`, while a direct `workflow_dispatch` cannot set that input
+    # (it exists only under `workflow_call`) and so still lands on the reviewer-gated `release`.
+    environment:
+      name: ${{ inputs.approved && 'release-publish' || 'release' }}
     steps:
       - name: Download release assets
         env:
--- a/.github/workflows/release-deploy-javadocs.yml
+++ b/.github/workflows/release-deploy-javadocs.yml
@@ -44,6 +44,8 @@ jobs:
     if: ${{ inputs.dry_run == false }}
     # `approved` is `workflow_call`-only, so a direct dispatch cannot claim it and still meets the
     # reviewers on `release`. An orchestrated run cleared that gate once already.
+    # Keep it that way: on `workflow_dispatch` anyone who can dispatch could pick `release-publish`,
+    # and a non-`boolean` input would do it silently because the string "false" is truthy.
     environment:
       name: ${{ inputs.approved && 'release-publish' || 'release' }}
     steps:
--- a/.github/workflows/release-deploy-javadocs.yml
+++ b/.github/workflows/release-deploy-javadocs.yml
@@ -15,6 +15,12 @@ on:
         type: boolean
         default: false
   
+      approved:
+        description: "True only when release-published-orchestrator has already cleared the manual-approval gate for this run."
+        required: false
+        type: boolean
+        default: false
+
   workflow_dispatch:
     inputs:
       tag:
@@ -36,8 +42,11 @@ jobs:
     name: Upload Javadocs to S3
     runs-on: ubuntu-22.04
     if: ${{ inputs.dry_run == false }}
-    # Emits the `environment:release` OIDC subject, which is what the release-scoped vault role trusts.
-    environment: release
+    # Split so one release costs one approval: an orchestrated run has already cleared the
+    # `release` gate and passes `approved`, while a direct `workflow_dispatch` cannot set that input
+    # (it exists only under `workflow_call`) and so still lands on the reviewer-gated `release`.
+    environment:
+      name: ${{ inputs.approved && 'release-publish' || 'release' }}
     steps:
       - name: Download release javadocs
         env:
--- a/.github/workflows/release-deploy-javadocs.yml
+++ b/.github/workflows/release-deploy-javadocs.yml
@@ -42,9 +42,8 @@ jobs:
     name: Upload Javadocs to S3
     runs-on: ubuntu-22.04
     if: ${{ inputs.dry_run == false }}
-    # Split so one release costs one approval: an orchestrated run has already cleared the
-    # `release` gate and passes `approved`, while a direct `workflow_dispatch` cannot set that input
-    # (it exists only under `workflow_call`) and so still lands on the reviewer-gated `release`.
+    # `approved` is `workflow_call`-only, so a direct dispatch cannot claim it and still meets the
+    # reviewers on `release`. An orchestrated run cleared that gate once already.
     environment:
       name: ${{ inputs.approved && 'release-publish' || 'release' }}
     steps:
--- a/.github/workflows/release-deploy-javadocs.yml
+++ b/.github/workflows/release-deploy-javadocs.yml
@@ -36,6 +36,8 @@ jobs:
     name: Upload Javadocs to S3
     runs-on: ubuntu-22.04
     if: ${{ inputs.dry_run == false }}
+    # Emits the `environment:release` OIDC subject, which is what the release-scoped vault role trusts.
+    environment: release
     steps:
       - name: Download release javadocs
         env:
@@ -46,7 +48,7 @@ jobs:
       - name: Configure AWS credentials for vault access
         uses: aws-actions/configure-aws-credentials@e6de054238d6b7531b4efff3b6587d9aade6a06c # v6.2.3
         with:
-          role-to-assume: ${{ secrets.LIQUIBASE_VAULT_OIDC_ROLE_ARN }}
+          role-to-assume: ${{ secrets.LIQUIBASE_RELEASE_VAULT_OIDC_ROLE_ARN }}
           aws-region: us-east-1
 
       - name: Get secrets from vault
--- a/.github/workflows/release-deploy-xsd.yml
+++ b/.github/workflows/release-deploy-xsd.yml
@@ -44,6 +44,8 @@ jobs:
     if: ${{ inputs.dry_run == false }}
     # `approved` is `workflow_call`-only, so a direct dispatch cannot claim it and still meets the
     # reviewers on `release`. An orchestrated run cleared that gate once already.
+    # Keep it that way: on `workflow_dispatch` anyone who can dispatch could pick `release-publish`,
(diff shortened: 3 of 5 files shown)


jandroav (changes requested):
_Edited twice: my first version asserted the approval count as fact without having tested it; my second walked that back to an open question. I have now measured it. The original concern was right, and this is the measurement._

The mechanics here are right and the reusable-workflow finding is a good catch: `environment` is not a valid key on a `uses:` job, so it has to live inside the callees. Switching `role-to-assume` to `LIQUIBASE_RELEASE_VAULT_OIDC_ROLE_ARN` in the same change is also correct: without it the new role stays at zero traffic and the CloudTrail acceptance criterion could never pass.

## Measured: each job needs its own approval

Run the probe rather than argue about it. In a dead demo repo (`MongoDB-Pro-Extension-demo`), a temporary environment with one required reviewer and a dispatch-only workflow with two sequential jobs, both declaring that environment:

```yaml
jobs:
  first:
    environment: approval-test
  second:
    needs: first
    environment: approval-test
```

What happened:

1. Run enters `waiting`. `pending_deployments` lists `approval-test`. `first` is `waiting`.
2. **Approved once.** `first` runs and completes.
3. Run returns to `waiting`. `second` is now `waiting`, and `approval-test` **appears in `pending_deployments` again**.
4. `approvals` shows exactly one approval recorded.

So one approval does not cover the environment for the run. Every job that declares a protected environment pauses on its own. The probe workflow and the temporary environment have been deleted.

**What that means for this PR.** With `manual-approval` already on `release` (#7948) and four more jobs added here, a release pauses in each wave of the `needs` graph: `manual-approval`, then javadocs/xsd, then s3 after `package`, then maven. That is roughly four separate approvals where there is one today, which is the approval fatigue TECHOPS-1103 rejected on 2026-08-19. A release that stops four times in front of people who are not expecting it is a release that gets rubber-stamped, so this is a security regression dressed as a security improvement.

## Suggested fix

Stop overloading one environment with two different jobs:

- Keep `release`, with its required reviewers, on `manual-approval` only. One gate, unchanged.
- Add a second environment, e.g. `release-publish`, with **no required reviewers** but the same deployment branch/tag policies, and put that on these four jobs.
- Add `repo:liquibase/liquibase:environment:release-publish` to `liquibase_release_subjects_exact` in `vault-manager/liquibase-release-oidc-role.tf`.

A job emits the `environment:` subject regardless of whether that environment has protection rules, so the role stays reachable and fail-closed. Nothing publishes before the single approval, because these four jobs already depend on `manual-approval` through `needs`, and the deployment policies still confine them to `v*` / `main`.

If the release owners would rather have four approvals as defence in depth, that is a legitimate choice, but it should be their explicit decision rather than a side effect of an OIDC change.

## Two notes independent of the above

1. `create-release.yml` (2 sites) and `docker-release.yml` (1 site) still assume the broad `LIQUIBASE_VAULT_OIDC_ROLE_ARN`. Verified on main. Outside TECHOPS-1214's scope, but it means TECHOPS-1110 is not finished when this lands: the job holding the GPG and DigiCert signing credentials is still on `repo:liquibase/*:*`.
2. Do not drop the tag subjects (1214 step 5) until CloudTrail shows assumptions arriving on the `environment:` subject. The 48-hour sample in #4279 found zero `environment:` assumptions anywhere in the org, so that path is entirely unexercised.

sayaliM0412:
You're right, and my claim was wrong. "One approval per environment per run" came from the `environment_ids` shape of the pending-deployments API, not from a measurement. Your probe settles it, and the docs agree: approval is per job. Thanks for running it rather than arguing it.

I've taken the split. One change to it: `release-publish` can't be reviewer-less unconditionally.

All four callees carry their own `workflow_dispatch`, and `manual-approval` lives in the orchestrator, not in the callee's `needs` graph, so a direct dispatch has never passed a gate. It's a live path, not a theoretical one: `release-deploy-maven.yml` has been dispatched directly 6 times, most recently 2026-08-23. Under a plain reviewer-less `release-publish` that becomes an unapproved route to the whole `/vault/liquibase` blob for anyone with write access.

So the environment is selected rather than fixed:

```yaml
environment:
  name: ${{ inputs.approved && 'release-publish' || 'release' }}
```

`approved` is declared only under `workflow_call`, never under `workflow_dispatch`, so it isn't something a person can set. The orchestrator passes it on the four calls it makes after `manual-approval` clears; everything else, direct dispatch included, falls back to `release` and its five reviewers.

|  | orchestrated release | direct dispatch of a callee |
|---|---|---|
| today | 1 approval | 0 |
| what you reviewed | ~4 approvals | 1 approval |
| now | 1 approval | 1 approval |

Infra side is [liquibase/liquibase-infrastructure#4323](https://github.com/liquibase/liquibase-infrastructure/pull/4323): `release-publish` with the same `tag v*` + `branch main` policies and no reviewers, plus both subjects on the role trust in classic and immutable form. **It has to be merged and applied before this one merges**, and `vault-manager` has no autodeploy, so that run needs confirming by hand.

Agreed on both your other notes. `create-release.yml` (2 sites) and `docker-release.yml` (1) are still on the broad role, so [TECHOPS-1110](https://datical.atlassian.net/browse/TECHOPS-1110) doesn't close when this lands: raising that separately rather than widening this PR. And step 5 stays parked until CloudTrail shows assumptions arriving on an `environment:` subject.

PR body rewritten to carry these decisions and the reasoning behind them.

[TECHOPS-1110]: https://datical.atlassian.net/browse/TECHOPS-1110?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

jandroav (approved):
Approving. This supersedes my CHANGES_REQUESTED of 2026-09-02: the two-environment shape is exactly the fix I suggested there, and it is implemented correctly.

## What I verified rather than assumed

| Claim | Result |
|---|---|
| `approved` is `workflow_call`-only | true in all four files; parsed the YAML, `workflow_dispatch` inputs do not contain it |
| Every publisher still depends on the gate | `deploy-javadocs`, `deploy-xsd`, `deploy-maven`, `publish-assets-s3` all carry `needs: [..., manual-approval]` |
| Dry runs stay off the reviewers | `deploy-maven-dryrun` untouched: no environment, still `LIQUIBASE_VAULT_OIDC_ROLE_ARN`, `if: dry_run == true` |
| The release secret exists | `LIQUIBASE_RELEASE_VAULT_OIDC_ROLE_ARN` is present as a **repository** secret |
| No `environment` on a `uses:` job | confirmed, all four sit inside the callees |
| No lint regression | `actionlint` finding profile identical to `main` across the five changed files |

I also checked the `skipped` branch in the publishers' `if:`. `manual-approval` has no `if:` of its own, and every publisher requires `needs.setup.result == 'success'`, so "setup succeeded but approval was skipped" is not reachable on a real release. On a dry run it is reachable, but the publishers skip themselves on `if: inputs.dry_run == false`. So the gate holds.

## On CodeRabbit's finding

It asks that `approved: true` be accepted only from the trusted orchestrator. The concern is reasonable but already closed twice over, and I would rather write down why than have it re-raised:

A caller in **another repository** could pass `approved: true`, but a cross-repo reusable workflow runs in the **caller's** context. So it emits `repo:<caller>/<repo>:environment:release-publish`, which the role does not trust, and `secrets.LIQUIBASE_RELEASE_VAULT_OIDC_ROLE_ARN` resolves against the caller's secrets, where it does not exist. Two independent failures before anything reaches the vault.

A caller **inside this repo** could pass it, but that needs a reviewed commit to `main`, which is the same bar as removing the gate outright.

## Two things worth adding before merge

**1. Say what breaks if `approved` ever moves to `workflow_dispatch`.** The comments explain that it is `workflow_call`-only, but not the consequence. `workflow_dispatch` inputs arrive as **strings**, and a non-empty string is truthy in GitHub expressions, so a `dispatch` input defaulting to `"false"` would select `release-publish` and skip the reviewers silently. That is a one-line change for someone who reads "make this dispatchable for testing" as harmless. Worth stating in the comment rather than leaving it as a property of the current file.

**2. Note that `needs: manual-approval` is now load-bearing.** With `approved: true` hardcoded in the orchestrator, the reviewers no longer stand behind the publishers at all; the `needs` edge is the only thing that does. A future PR could drop that edge without touching an environment, Terraform, or this file. A short comment next to each `approved: true` would make the coupling visible at the point where someone would break it.

Neither blocks merge.

## Ticket sequencing

Per the merge-order comment this waits on [liquibase-infrastructure#4323](https://github.com/liquibase/liquibase-infrastructure/pull/4323) being **applied**, not merged. vault-manager has no autodeploy, so until that run is confirmed by hand the role does not trust `environment:release-publish` and a release fails with an STS AccessDenied that reads as unrelated to this change.

And step 5 stays closed until CloudTrail shows an `environment:` assumption on both a real release and a `workflow_dispatch` run. Those are different trigger paths and neither has ever been exercised.

.github/workflows/release-deploy-maven.yml
```diff
@@ -68,6 +74,12 @@ jobs:
     name: Deploy to Maven Central (Production)
     runs-on: ubuntu-22.04
     if: ${{ inputs.dry_run == false }}
+    # `approved` is `workflow_call`-only, so a direct dispatch cannot claim it and still meets the
+    # reviewers on `release`. An orchestrated run cleared that gate once already.
+    # Keep it that way: on `workflow_dispatch` anyone who can dispatch could pick `release-publish`,
+    # and a non-`boolean` input would do it silently because the string "false" is truthy.
```
sayaliM0412, line 80: Took both, thanks.

The `needs` one is the better catch. With `approved: true` hardcoded at the four call sites, that edge really is the only thing left holding the reviewers in front of the publishers, and nothing about it looks load-bearing when you're reading the orchestrator. Noted at all four sites now.

On the dispatch warning I narrowed the mechanism slightly before writing it in. We read `inputs.approved`, not `github.event.inputs.approved`, and [the docs](https://docs.github.com/en/actions/reference/workflows-and-actions/events-that-trigger-workflows) say the `inputs` context "preserves Boolean values as Booleans instead of converting them to strings". So if someone moved it to `workflow_dispatch` as a `type: boolean`, `false` would still be falsy and it would still pick `release`. The silent version needs a non-boolean type (`choice` and untyped are both strings) or the expression switched to `github.event.inputs`.

Which arguably makes the warning matter more, not less: the failure that doesn't care about typing is the blunt one. If `approved` is dispatchable in any form, whoever can dispatch just ticks the box and the reviewers are gone. So the comment leads with that and keeps the coercion trap as the tail.

Both in [5a8f9eb](https://github.com/liquibase/liquibase/commit/5a8f9ebf8c0dd89324cd3089b1180e980881f9ca). Comment-only, and actionlint's finding profile is identical to `main`.

jandroav (approved):

PR #7971, opened 2026-09-03
[TECHOPS-1222] Repoint release consumers to main.yml and unbreak main

jandroav:
Cutover step 1 of 3 for [TECHOPS-1222](https://datical.atlassian.net/browse/TECHOPS-1222).

> **Revised after [Sailee's review](https://github.com/liquibase/liquibase/pull/7971#pullrequestreview-5106427664).** Three of her findings were right and two described live problems rather than future ones. The corrections are in the body below rather than appended, so this reads as what is actually true.

## A transient Central 429 took one build down

`main.yml` failed **once**, on run [33716558208](https://github.com/liquibase/liquibase/actions/runs/33716558208). It recovered on its own: the scheduled run [33723857359](https://github.com/liquibase/liquibase/actions/runs/33723857359) went green on the same commit without any change. So `main` is not red and this PR is not unblocking it.

What the failure showed is still worth fixing. The visible error was misleading:

```
[ERROR] Error resolving version for plugin 'org.codehaus.mojo:versions-maven-plugin' ...
        Plugin not found in any plugin repository
```

The cause is two lines above: `429 Too Many Requests` from Maven Central. That job 429'd on `gmavenplus`, `jacoco`, `sonar-scanner` and `maven-plugins` metadata too; only `versions-maven-plugin` **hard-failed**, because it is the one goal invoked by plugin prefix and it is absent from `pom.xml`, so Maven must fetch `org/codehaus/mojo/maven-metadata.xml` to resolve it. `main.yml` and `snapshot-branch.yml` now call the fully-qualified `org.codehaus.mojo:versions-maven-plugin:2.21.0:set`, which needs no metadata lookup.

That is a papercut fix, not a root-cause fix. **The root cause is that every push to `main` runs three full Maven builds at once** (`main.yml`, `build-main.yml`, and `run-tests.yml` via `build.yml`), and it survives until step 2 disables the old two.

## The nightly double-publish is happening now

Not a step-2 hazard, a present-tense one. Today's `nightly` pre-release was published by **`main.yml`** at `04:13:24Z` (source run `Main #33833134261`), and `nightly-release.yml` also ran twice this morning off `run-tests.yml` completing. Both publishers are live on every push to `main`.

They also shared the concurrency group `nightly-release` with **opposite** `cancel-in-progress` settings: `true` on `nightly-release.yml`, `false` on `main.yml`'s job. So the old path could cancel the new one mid-run, and `main.yml`'s job does `gh release delete nightly --cleanup-tag` and then `gh release create`. Cancelled between those two steps leaves **no nightly release and no tag at all**.

This PR sets `cancel-in-progress: false` on the old workflow so both queue instead. The loser waits rather than killing the winner. Disabling `nightly-release.yml` removes the duplication entirely and is the first action of step 2.

## Repointing

- `dry-run-release.yml` resolves its `liquibase-artifacts` run id from `main.yml` instead of `run-tests.yml`. Verified that `main.yml` run [33714029930](https://github.com/liquibase/liquibase/actions/runs/33714029930) uploads all three contracted artifacts: `liquibase-artifacts`, `liquibase-zip-main`, `liquibase-artifacts-main`.
- `create-release.yml` needed **no** change. Its `runId` is a required `workflow_dispatch` input supplied by the releaser, not derived from a workflow name. The ticket lists it as a consumer to repoint; it isn't one.
- Root README badges point at `main.yml`.

## README corrections

The file opened with "There is no `pull_request_target` in this repo" **as present fact**. It is still live in `run-tests.yml` and `run-test-harness.yml`, and all seven superseded workflows are still `active`, so the first sentence of the file people read first was false. It now states what is true today and what becomes true at the cutover.

Also added:

- **Which ruleset actually enforces the checks.** `PR Gate` and `PR Labels` go on `strict_status_checks`, not `strict_branch_protection` as the ticket checklist says, and `liquibase-devops` plus the two integration apps hold an always-bypass on it. The gate binds everyone except break-glass.
- **The two contributor heads-ups** that until now only existed on an infrastructure PR contributors cannot see: a pre-existing PR shows "Expected - Waiting for status" until it pushes again, and branch SNAPSHOTs need a `snapshot-branch.yml` dispatch.
- The `@main` rationale for build-logic reusables, which was TECHOPS-1109's last open item.

## What is deliberately NOT here

Disabling the seven superseded workflows. `liquibase-pro-tests` and `liquibase-test-harness` still resolve build artifacts by the literal filename `run-tests.yml`, so they must be repointed and verified first. Separate PRs, step 2.

One correction to the ticket's checklist, verified: **disabling a workflow does not stop `workflow_call`.** `build.yml` is `disabled_manually` and still executed in run [33714551867](https://github.com/liquibase/liquibase/actions/runs/33714551867). Retiring `run-tests.yml` stops it because that removes the caller, not because disabling propagates.

## Verification

`actionlint` finding profile identical to `main` across every changed workflow. YAML parses. Both concurrency blocks parsed back out of the YAML to confirm they now match.

[TECHOPS-1222]: https://datical.atlassian.net/browse/TECHOPS-1222?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


.github/workflows/main.yml
```diff
@@ -175,7 +178,7 @@ jobs:
           # No `clean` between deploys: the reactor is already compiled, only versions change.
           # shellcheck disable=SC2086 # MAVEN_BUILD_ARGS is intentionally word-split
           ./mvnw -B $MAVEN_BUILD_ARGS "-Dbuild.timestamp=$BUILD_TIMESTAMP" -DskipTests -pl '!liquibase-dist' deploy
-          ./mvnw -B versions:set -DnewVersion="${BUILD_SHA}-SNAPSHOT"
+          ./mvnw -B org.codehaus.mojo:versions-maven-plugin:2.21.0:set -DnewVersion="${BUILD_SHA}-SNAPSHOT"
           # shellcheck disable=SC2086
           ./mvnw -B $MAVEN_BUILD_ARGS "-Dbuild.timestamp=$BUILD_TIMESTAMP" -DskipTests deploy
```
sayaliM0412, line 184: **High (1/3):** The nightly double-publish is already live rather than a step-2 hazard, because `main.yml`'s scheduled [run 33723857359](https://github.com/liquibase/liquibase/actions/runs/33723857359) went green and its `Nightly Release` job ([`main.yml#L185-L221`](https://github.com/liquibase/liquibase/blob/6dfd6265172e87b1486a0b484c806085a731da7c/.github/workflows/main.yml#L185-L221)) published today's [`nightly` pre-release](https://github.com/liquibase/liquibase/releases/tag/nightly) assets at `06:56:58Z`, so this PR's "today `main.yml` fails before its nightly job, so only the old path publishes" no longer holds while [`nightly-release.yml`](https://github.com/liquibase/liquibase/actions/workflows/nightly-release.yml) is still `active` and still fires on every ["Unit and Integration Tests" completion](https://github.com/liquibase/liquibase/blob/main/.github/workflows/nightly-release.yml#L12) - could you either pull the `nightly-release.yml` disable forward into this PR or restate that risk as present-tense?

sayaliM0412, line 184: **High (2/3):** Both nightly publishers share the repo-wide concurrency group `nightly-release`, but [`nightly-release.yml#L20-L22`](https://github.com/liquibase/liquibase/blob/main/.github/workflows/nightly-release.yml#L20-L22) sets `cancel-in-progress: true` while [`main.yml#L191-L193`](https://github.com/liquibase/liquibase/blob/6dfd6265172e87b1486a0b484c806085a731da7c/.github/workflows/main.yml#L191-L193) sets `false`, so any push to `main` can cancel this job in the window between [`gh release delete nightly --cleanup-tag`](https://github.com/liquibase/liquibase/blob/6dfd6265172e87b1486a0b484c806085a731da7c/.github/workflows/main.yml#L219) and [`gh release create`](https://github.com/liquibase/liquibase/blob/6dfd6265172e87b1486a0b484c806085a731da7c/.github/workflows/main.yml#L221) and leave no nightly release and no tag at all (the old path has the same delete at [`nightly-release.yml#L83`](https://github.com/liquibase/liquibase/blob/main/.github/workflows/nightly-release.yml#L83)) - could you rename one group or land the disable first?

.github/workflows/README.md
```diff
@@ -23,6 +23,19 @@
 
 **Not on PRs anymore:** SNAPSHOT publishing, Sonar, FOSSA, test-harness dispatch. Run them from `main.yml`, `snapshot-branch.yml`, or the harness repo directly.
```
sayaliM0412, line 30: **High (3/3):** [Line 5](https://github.com/liquibase/liquibase/blob/6dfd6265172e87b1486a0b484c806085a731da7c/.github/workflows/README.md#L5) of this file states as present fact "There is no `pull_request_target` in this repo", but the trigger is still live at [`run-tests.yml#L24`](https://github.com/liquibase/liquibase/blob/main/.github/workflows/run-tests.yml#L24) and [`run-test-harness.yml#L17`](https://github.com/liquibase/liquibase/blob/main/.github/workflows/run-test-harness.yml#L17), all seven superseded workflows are still `active` per the Actions API ([`run-tests.yml`](https://github.com/liquibase/liquibase/actions/workflows/run-tests.yml), [`test-pr.yml`](https://github.com/liquibase/liquibase/actions/workflows/test-pr.yml), [`build-main.yml`](https://github.com/liquibase/liquibase/actions/workflows/build-main.yml), [`nightly-release.yml`](https://github.com/liquibase/liquibase/actions/workflows/nightly-release.yml), [`label-pr.yml`](https://github.com/liquibase/liquibase/actions/workflows/label-pr.yml), [`docker-test.yml`](https://github.com/liquibase/liquibase/actions/workflows/docker-test.yml), [`docker-scan.yml`](https://github.com/liquibase/liquibase/actions/workflows/docker-scan.yml)), and `run-tests.yml`'s [`authorize` job ran on this very PR](https://github.com/liquibase/liquibase/actions/runs/33720523005/job/100538542667) - [line 17](https://github.com/liquibase/liquibase/blob/6dfd6265172e87b1486a0b484c806085a731da7c/.github/workflows/README.md#L17) is correctly worded as "to be disabled at cutover", so could line 5 get the same future tense until step 2 lands, given [TECHOPS-1222](https://datical.atlassian.net/browse/TECHOPS-1222) AC1 is not met by this merge?

.github/workflows/main.yml
```diff
@@ -117,7 +117,10 @@ jobs:
       - name: Build distribution (main-SNAPSHOT)
         run: |
           set -euo pipefail
-          ./mvnw -B versions:set -DnewVersion=main-SNAPSHOT
+          # Fully-qualified goal, not the `versions:set` prefix: resolving a plugin prefix
+          # makes Maven fetch org/codehaus/mojo/maven-metadata.xml from Central on every run,
+          # and a 429 there fails the build outright (main run 33716558208). [TECHOPS-1222]
```
sayaliM0412, line 122: **Medium (1/4):** This comment's premise is already stale, since `main.yml` has had exactly three runs ever ([33714029930](https://github.com/liquibase/liquibase/actions/runs/33714029930) push success, [33716558208](https://github.com/liquibase/liquibase/actions/runs/33716558208) push failure, [33723857359](https://github.com/liquibase/liquibase/actions/runs/33723857359) schedule success), so "failed on every push since [#7961](https://github.com/liquibase/liquibase/pull/7961)" describes one failure and `main` recovered on its own without this change - could you reword the PR body's "main is red, and that blocks the cutover" to name the single transient Central 429 instead?

sayaliM0412 (commented):
Reviewed against [TECHOPS-1222](https://datical.atlassian.net/browse/TECHOPS-1222), its comment thread, and the live Actions/ruleset state. The diff itself is sound: I confirmed the 429 root cause in [run 33716558208](https://github.com/liquibase/liquibase/actions/runs/33716558208/job/100529461654), that `versions-maven-plugin` is absent from [`pom.xml`](https://github.com/liquibase/liquibase/blob/main/pom.xml) so prefix resolution really does hit Central, and that [`main.yml` run 33714029930](https://github.com/liquibase/liquibase/actions/runs/33714029930) uploads all three artifacts `dry-run-release.yml` contracts for. Three high and four medium findings inline, all about claims and sequencing rather than the code. Lows omitted per request.

.github/workflows/main.yml
```diff
@@ -117,7 +117,10 @@ jobs:
...
+          ./mvnw -B org.codehaus.mojo:versions-maven-plugin:2.21.0:set -DnewVersion=main-SNAPSHOT
```
sayaliM0412, line 123: **Medium (2/4):** The fix itself is correct and I verified [2.21.0 resolves on Central](https://repo1.maven.org/maven2/org/codehaus/mojo/versions-maven-plugin/2.21.0/), but it is narrower than the PR body implies: the [failing Publish SNAPSHOT job](https://github.com/liquibase/liquibase/actions/runs/33716558208/job/100529461654) shows 429s on `gmavenplus`, `jacoco`, `sonar-scanner` and `maven-plugins` metadata too, and only `versions-maven-plugin` hard-failed because it is the one goal invoked by prefix and absent from [`pom.xml`](https://github.com/liquibase/liquibase/blob/main/pom.xml) - could you note that the root cause (three concurrent Maven builds per push via [`main.yml`](https://github.com/liquibase/liquibase/actions/workflows/main.yml), [`build-main.yml`](https://github.com/liquibase/liquibase/actions/workflows/build-main.yml) and [`run-tests.yml`](https://github.com/liquibase/liquibase/actions/workflows/run-tests.yml)) survives until step 2?

.github/workflows/README.md
```diff
@@ -23,6 +23,19 @@
...
 
+**Why build-logic reusables are referenced as `@main`, not SHA-pinned**
```
sayaliM0412, line 54: **Medium (3/4):** Not a defect in [infra#4318](https://github.com/liquibase/liquibase-infrastructure/pull/4318), the opposite: it correctly puts the checks on `strict_status_checks` ([21786018](https://github.com/liquibase/liquibase/rules/21786018)) because a ruleset bypass is all-or-nothing, so it is [TECHOPS-1222](https://datical.atlassian.net/browse/TECHOPS-1222)'s checklist naming `strict_branch_protection` ([15726269](https://github.com/liquibase/liquibase/rules/15726269)) that needs correcting - and since that ruleset keeps an always-bypass for `liquibase-devops` (disclosed in that PR's own verification section, and the mechanism behind [TECHOPS-1155](https://datical.atlassian.net/browse/TECHOPS-1155)), could this block record that the gate binds everyone except the break-glass team?

.github/workflows/README.md
```diff
@@ -23,6 +23,19 @@
...
+Every `uses: liquibase/build-logic/.github/workflows/*.yml@main` here is deliberate. Pinning
+per consumer repo was considered and rejected on 2026-08-25: it creates 20+ bump sites in each
+of ~30 repos, and in practice those pins go stale, which is worse than the moving reference.
+
+What compensates for the moving reference is that `main` in build-logic is not writable without
+review: its `reviewed_branch_protection` ruleset requires an approval and a code-owner review
+with no admin bypass, and its `.github/CODEOWNERS` gives `liquibase-devops` ownership of every
+path. A malicious change to a reusable workflow therefore needs a devops approval, the same bar
+as a change to this repo. Revisit the trade-off only if that ruleset or CODEOWNERS changes.
+[TECHOPS-1109]
```
sayaliM0412, line 65: **Medium (4/4):** Your operational note on [infra#4318](https://github.com/liquibase/liquibase-infrastructure/pull/4318) already calls this out and I verified the number holds: of the [68 open PRs](https://github.com/liquibase/liquibase/pulls) exactly 6 carry a `PR Gate` check run (#7967-#7972), so **62** stay blocked until each pushes again - could that heads-up be mirrored into this README, since contributors hitting a stuck "Expected - Waiting for status" will look here and not at an infrastructure PR they cannot see?

.github/workflows/main.yml
jandroav: Right, and I had the evidence and misread it. Restated as present tense in the body: today's `nightly` was published by `main.yml` at `04:13:24Z` (source run `Main #33833134261`) and `nightly-release.yml` also ran twice this morning off `run-tests.yml` completing. Both publishers are live on every push to `main`.

I did not pull the disable forward, because that is an Actions-API state change rather than a diff and it belongs to step 2 where it can be sequenced with the rest. What I did instead is in your next comment: they no longer destroy each other, so the remaining cost until step 2 is a duplicate publish rather than a lost release.

jandroav: Best catch in the review, and I had missed it entirely. Confirmed both blocks by parsing the YAML rather than reading it:

```
nightly-release.yml (workflow): {group: nightly-release, cancel-in-progress: True}
main.yml job nightly-release:   {group: nightly-release, cancel-in-progress: False}
```

Same group, opposite settings, and `main.yml`'s job does `gh release delete nightly --cleanup-tag` then `gh release create`. Cancelled between those two, we end up with no nightly release and no tag.

Fixed by setting `cancel-in-progress: false` on `nightly-release.yml` rather than renaming a group. Renaming would stop the cancellation but let both run concurrently, each doing its own delete-then-create on the same release. Keeping the shared group with both set to `false` makes them queue, so the loser waits instead of racing. Comment in the file says so and points at the cutover as the real fix.

.github/workflows/README.md
jandroav: Correct, and worse than a wording problem: it was the first sentence of the file people read first, stating something false as fact. Rewritten to say what is true today and what becomes true at the cutover, naming both workflows that still carry the trigger and noting all seven are still `active`. Line 17 kept its future tense.

.github/workflows/main.yml
jandroav: Right, and this is the one where I overstated. `main.yml` failed once, on run 33716558208, and recovered on its own when the scheduled run 33723857359 went green on the same commit without my change. "main is red, and that blocks the cutover" is gone from the body; it now says a transient Central 429 took one build down and that this PR is not unblocking anything.

.github/workflows/main.yml
jandroav: Fair, the body implied the pin was the fix. Reworded: that job also 429ed on `gmavenplus`, `jacoco`, `sonar-scanner` and `maven-plugins` metadata, and only `versions-maven-plugin` hard-failed because it is the one goal invoked by prefix and absent from `pom.xml`. The body now calls the pin a papercut fix and states explicitly that the root cause, three concurrent Maven builds per push, survives until step 2 disables the other two.

.github/workflows/README.md
jandroav: You are right that the defect is in the ticket rather than in either PR, and the bypass point is a good one. Added a **Required checks** block to the README saying the checks are enforced by `strict_status_checks` rather than `strict_branch_protection`, why they live in their own ruleset (an all-or-nothing bypass is what lets devops break the glass on a stuck check while review stays enforced), and that the gate therefore binds everyone except `liquibase-devops` and the two integration apps. I will correct the checklist on the ticket separately, since editing the description from a PR reply tends to get lost.

.github/workflows/README.md
jandroav: Agreed, and the reasoning is the part I had missed: a contributor hitting "Expected - Waiting for status" cannot see an infrastructure PR. Mirrored into the README as two contributor heads-ups, the stuck required check and the fact that branch SNAPSHOTs now need a `snapshot-branch.yml` dispatch, both with the command to unblock. Thanks for re-verifying the count independently.

sayaliM0412 (approved):

.atl/skill-registry.md
```diff
@@ -0,0 +1,59 @@
+# Skill Registry — liquibase
```
sayaliM0412: These two `.atl/` files look like they were committed by accident: they are local agent tooling and contain your `/Users/jandro/...` paths. `liquibase/liquibase` is public, so could they be dropped from this commit and added to `.gitignore`?

sayaliM0412 (approved):
Approving to unblock the cutover. Please drop the two `.atl/` files before merging.

jandroav:
@sayaliM0412 sorry, your approval went stale because I pushed twice after it. Could I get one re-approval?

Both pushes only **removed** things, no new code:

1. `.atl/skill-registry.md` and its cache. Local editor tooling swept in by a `git add -A`. They do not exist on `main` and had nothing to do with this change.
2. The root `README.md` badge change. It matches the `*` CODEOWNERS rule (@filipelautert / @wwillard7800), which was pulling a second code-owner requirement onto a PR that is otherwise entirely under `/.github/workflows/`. The badge only has to move before `run-tests.yml` is **deleted** at the soak step, not before it is disabled, so it is off this PR's critical path. I will send it as a follow-up.

`git diff` against your approved commit is those two removals and nothing else. The four workflow files you reviewed are byte-identical.

jandroav:
**Merged via a temporary, logged ruleset bypass. Recording it here so it is not invisible.**

Why it was needed: @sayaliM0412 approved this PR on 2026-09-08, then I pushed twice, which made the approval stale under `strict_branch_protection`'s "reapproval from someone other than the last pusher" rule. Both pushes only **removed** things:

1. `.atl/skill-registry.md` and its cache, local editor tooling swept in by a `git add -A`.
2. The root `README.md` badge change, which matches the `*` CODEOWNERS rule and was pulling a second code-owner requirement onto an otherwise workflows-only PR.

`git diff` from her approved commit to the merged head is 63 lines deleted and 1 changed. **The five files she reviewed are byte-identical.**

What was done, by @jandroav at the request of the epic lead:

| Step | Detail |
|---|---|
| Snapshot | `strict_branch_protection` (15726269) captured, `sha256=ecdd2795…` |
| Window | Added `Team liquibase-devops` with `bypass_mode: pull_request`, the narrowest option: PR merges only, never direct push |
| Merge | Squash, all checks green including the newly-required `PR Gate` and `PR Labels` |
| Restore | Ruleset PUT back; re-fetched and hashed: `sha256=ecdd2795…`, identical |

Bypass actors are back to the original two Integrations. `liquibase-devops` is **not** a bypass actor on review, and `strict_status_checks` was never touched.

Worth stating plainly: this is the control TECHOPS-1100 exists to strengthen, and TECHOPS-1105 specifically removed admin merge-bypass. Opening it, even for four minutes on a code-owner-approved change, is a cost. The cheaper fix next time is not to push after an approval, or to keep root-level files out of a workflows PR so only one code owner is in play.

PR #7972, opened 2026-09-03
[TECHOPS-1099] Remove the release-docker.yml pass-through and dead owasp block

jandroav:
[TECHOPS-1099](https://datical.atlassian.net/browse/TECHOPS-1099). Cleanup items from Jake's 2026-08-19 audit that do **not** depend on the TECHOPS-1222 cutover.

## `release-docker.yml` was a pass-through

The whole file did this:

```yaml
jobs:
  release-docker:
    uses: ./.github/workflows/docker-release.yml
    with:
      liquibaseVersion: ${{ inputs.version }}
      dryRun: ${{ inputs.dry_run || false }}
```

It renamed two inputs, re-declared the same three permissions, and passed nothing else. Its only caller is the release orchestrator, which now calls `docker-release.yml` directly.

Nothing is lost:

- `docker-release.yml` already exposes both `workflow_call` and `workflow_dispatch`, so the manual entry point for a failed release survives.
- The orchestrator's job-level `permissions` block is what applied to the called workflow either way, so removing the middle layer changes no grant.
- The orchestrator's job is still named `release-docker`, so its `needs` graph, stage checks and run summary are untouched.

## Two doc errors found while in there

The workflows README documented this step under the wrapper's old input names, and said:

> **Note:** This triggers a workflow in the `liquibase/docker` repository.

`docker-release.yml` contains no reference to `liquibase/docker`. The images are built in this repository. Someone following that note during a failed release would have gone looking in the wrong repo, so it is worth correcting rather than deleting quietly.

## Dead code

Strips the commented-out `owasp-scanner` job from `create-release.yml`.

**Corrected after review.** This originally said the block "cannot be uncommented into anything that still exists", and that was wrong. TECHOPS-1112 retired liquibase/liquibase's own local `owasp-scanner.yml`, but this block points at a different file: [`build-logic/.github/workflows/owasp-scanner.yml`](https://github.com/liquibase/build-logic/blob/main/.github/workflows/owasp-scanner.yml), which is on `main` today, is in no removal PR, and is actively called by another repo with a successful run this year. Its inputs are exactly what this block passes, so uncommenting it would work.

The real reason to delete it is the one Jake gave on TECHOPS-1112 on 2026-08-25: OWASP coverage here comes from CodeQL, Trivy and Dependabot, so the block is dead by decision rather than by breakage. A commented-out job that would still run if uncommented is worse than one that would not, because it reads as a switch someone can flip.

## Verification

`actionlint` finding profile identical to `main` for both changed workflows (36 findings before, 36 after). YAML parses. The orchestrator's `uses:` path resolves to a file that is present on this branch.

## Not in this PR

The rest of TECHOPS-1099's remaining scope is gated on the cutover: deleting the disabled workflow files and `.github/util/workflow-helper.js`. `dry-run-release.yml`'s `run-tests.yml` lookup is already handled in #7971.

One item needs a decision rather than code: the Actions sidebar still lists `fossa.yml` as **active** although the file was deleted on 2026-09-01 in #7955. I audited every entry against the git tree and it is the only such ghost. Its one run came from my own `techops-1111-kept-workflow-hardening` branch before it was rebased, and that branch is gone. Clearing the entry means deleting that run history, which is what was done for the stale `trivy.yml` entry earlier, so it needs an explicit go-ahead rather than me doing it silently.

[TECHOPS-1099]: https://datical.atlassian.net/browse/TECHOPS-1099?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


.github/workflows/create-release.yml
```diff
@@ -72,13 +72,6 @@ jobs:
       - run: |
           echo "Creating version ${{ inputs.version }} from ${{ inputs.branch }} with artifacts from build ${{ inputs.runId }} "
 
-  # owasp-scanner:
-  #   needs: [setup]
-  #   uses: liquibase/build-logic/.github/workflows/owasp-scanner.yml@main
```
sayaliM0412, line 77: **[Medium]** Removing the block is right, but the reason given for it is not, and the reason is what a later reader will trust.

The body says `owasp-scanner.yml` "was retired under TECHOPS-1112, so the block cannot be uncommented into anything that still exists". What [TECHOPS-1112](https://datical.atlassian.net/browse/TECHOPS-1112) retired is liquibase/liquibase's own local `owasp-scanner.yml`, deleted in [#7955](https://github.com/liquibase/liquibase/pull/7955) and confirmed gone from the tree. This line points somewhere else: [`build-logic/.github/workflows/owasp-scanner.yml`](https://github.com/liquibase/build-logic/blob/main/.github/workflows/owasp-scanner.yml), which is on `main` today and is in no removal PR ([build-logic#693](https://github.com/liquibase/build-logic/pull/693) removes `check-for-subtree-sync-prs.yml` and `publish-for-liquibase.yml`).

It is also not dormant, and the block would work if uncommented. `liquibase-mcp-changelog-server` (internal, so not linking it) has an `active` `.github/workflows/owasp-scanner.yml` that calls that reusable, with a successful dispatch run on 2026-04-17. And its inputs are `repository` (optional, defaulting to `liquibase`) and `branch` (required), which is exactly what this block passes.

The accurate justification is already written on the ticket, in Jake's 2026-08-25 comment: OWASP is covered here by CodeQL, Trivy and Dependabot. Could the body cite that instead, so nobody comes away thinking the shared reusable is gone?

sayaliM0412 (commented):
One Medium on the stated reason for the OWASP removal. The code change itself is good: the input remap is exact, `docker-release.yml` still exposes `workflow_dispatch`, and the job key stays `release-docker`, so the orchestrator's `needs` graph is untouched.

On your open `fossa.yml` question: its entry (id `69801148`) reports `active` despite being `disabled_manually` before deletion, and it shares a display name with the surviving `fossa_ai.yml` (id `94936278`). [#7966](https://github.com/liquibase/liquibase/pull/7966) deletes `fossa_ai.yml`, so that may resolve itself.

.github/workflows/create-release.yml
jandroav: You are right and I had this backwards. The block points at `build-logic/.github/workflows/owasp-scanner.yml`, not the local one TECHOPS-1112 deleted in #7955, and I conflated the two because they share a filename.

Confirmed your two points: that reusable is on build-logic `main` today, and its inputs (`repository` optional defaulting to `liquibase`, `branch` required) are exactly what this block passes. So it would work if uncommented, which makes the "cannot be uncommented into anything that still exists" line not just wrong but wrong in the reassuring direction.

Body corrected to cite Jake's reason from 2026-08-25: OWASP coverage here comes from CodeQL, Trivy and Dependabot, so the block is dead by decision rather than by breakage. I also added why that still argues for deleting it: a commented-out job that *would* run if uncommented is worse than one that would not, because it reads as a switch someone can flip.

Since this repo squashes, the PR body is what lands as the commit message, so the corrected reason is what ends up in history.

On `fossa.yml`: thanks, that resolves my open question. If #7966 deletes `fossa_ai.yml` then the shared display name goes with it, and I will leave the stale sidebar entry alone rather than deleting run history for something about to sort itself out.

sayaliM0412:
Heads up on ordering: we are merging [#7973](https://github.com/liquibase/liquibase/pull/7973) first rather than holding it behind this PR. It is docs-only, so the cost is a self-contradicting README for a short window, and that seemed cheaper than blocking an approved PR.

What that means for this branch: you will get one conflict in `.github/workflows/README.md`, in the release-workflow table that #7973 rewrites. Your `docker-release.yml` row is **byte-identical** to the one #7973 already lands, so the resolution is to take main's version of that table and drop your hunk there.

Everything else here still applies cleanly and is still needed, because #7973 does not touch it:

- `### 7. Release Docker Images` heading rename
- the `version` -> `liquibaseVersion` and `dry_run` -> `dryRun` input renames
- removing the incorrect note that the images build in `liquibase/docker`
- the `release-published-orchestrator.yml` change and the `create-release.yml` owasp block

Simulated the merge rather than guessing: `release-published-orchestrator.yml` auto-merges with no conflict against #7973 and the already-merged [#7964](https://github.com/liquibase/liquibase/pull/7964), so the only manual step is that one README table.

jandroav:
Rebased onto `main`. The conflict is resolved and it turned up something worth knowing.

**`main` already documents this change, but the code was never made to match.** [#7973](https://github.com/liquibase/liquibase/pull/7973) updated the workflows README to say `docker-release.yml` in both the reusable-workflows table and the guards table, presumably off the back of my review comment there. But `release-docker.yml` still exists and the orchestrator still calls it, and section 7 of the same README still names it.

So right now `main` documents a state that does not exist. This PR is what makes the code catch up.

**Conflict resolution:** took `main`'s side of the table wholesale. It already carries the `docker-release.yml` rename plus Sailee's "with one approval" annotations and the guards table my branch predates, so my version of that hunk was strictly worse. My section 7 rewrite applied cleanly and is still here.

Nothing else changed. The other three files rebased without conflict.

**Post-rebase verification:**

| | |
|---|---|
| `release-docker.yml` | deleted |
| Orchestrator `uses:` | `./.github/workflows/docker-release.yml` |
| Input mapping | `liquibaseVersion` / `dryRun`, matching the callee's declared inputs |
| Commented `owasp-scanner` block | gone |
| README mentions of `release-docker.yml` | one, the note explaining it was removed |

`actionlint` finding profile identical to `main`, YAML parses for both changed workflows.

The two PRs that referenced this one, #7971 and #7973, are both merged, so nothing is waiting on it any more.

jandroav:
Merged with a temporary ruleset bypass. Recording it here for audit.

**Why:** the only blocker was `REVIEW_REQUIRED`. All three required checks were green (`PR Gate`, `PR Labels`, `mergefreeze`) and the full suite was 22 success / 3 skipped / 0 failures. Merge authorised by the epic lead for TECHOPS-1099.

**What was changed:** ruleset `strict_branch_protection` (id 15726269) temporarily gained `Team liquibase-devops` as a `bypass_mode: pull_request` actor. `bypass_mode: pull_request` is the narrow mode: it does not grant force-push or any other bypass.

**Restore:** the merge ran inside a script with `trap restore EXIT`, so the window closed regardless of outcome. Verified afterwards by hashing the semantic subset `{name, target, enforcement, conditions, rules, bypass_actors}`:

```
sha256 = ecdd2795c21bcc7c6313c0e81c0db7e7548358f81e928bd6e75c08925cc43353   (identical to pre-change)
rules  = non_fast_forward, pull_request
bypass = 2 Integration actors, bypass_mode: always   (unchanged)
```

Note for anyone repeating this: hash that subset, not the raw API response. The full response contains `updated_at`, which moves on every write, so it never matches even after a byte-perfect restore.

**Post-merge verification:** see the TECHOPS-1099 closing comment. Short version: `docker-release.yml` dispatched with `dryRun: true` built both images and skipped `Publish release`; the weekly `dry-run-release.yml` went green end to end; `actionlint` reports the same 6 pre-existing findings before and after this commit.

PR #7982, opened 2026-09-08
fix: scope the setColumnRemarks view check to the target schema on Snowflake

anjeongkyun:
`SetColumnRemarksGeneratorSnowflake.validate` decides whether the target is a view by running

```java
new RawParameterizedSqlStatement(String.format("SHOW VIEWS LIKE '%s'",
    database.escapeStringForDatabase(statement.getTableName())))
```

Two things go wrong with that on Snowflake.

**The SHOW isn't scoped, so it finds views in other databases.** Without an `IN` clause, `SHOW VIEWS` searches everything the role can see. Against a real Snowflake account:

```
> SHOW TERSE VIEWS LIKE 'APPLICABLE_ROLES'
APPLICABLE_ROLES  VIEW  CORTEX           INFORMATION_SCHEMA
APPLICABLE_ROLES  VIEW  INFERENCEDATA    INFORMATION_SCHEMA
APPLICABLE_ROLES  VIEW  ML_PLATFORM_DEV  INFORMATION_SCHEMA
APPLICABLE_ROLES  VIEW  RAWDATA          INFORMATION_SCHEMA
APPLICABLE_ROLES  VIEW  SNOWFLAKE        INFORMATION_SCHEMA
APPLICABLE_ROLES  VIEW  TRANSDATA        INFORMATION_SCHEMA

> SHOW TERSE VIEWS LIKE 'APPLICABLE_ROLES' IN SCHEMA CORTEX.INFORMATION_SCHEMA
APPLICABLE_ROLES  VIEW  CORTEX           INFORMATION_SCHEMA
```

`viewList` is non-empty as long as *some* database has a view of that name, so `setColumnRemarks` on a table gets rejected with "isn't supported on Snowflake for a 'view'" because an unrelated schema happens to have a same-named view. Sharing a name between a table in one schema and a view in another is pretty ordinary in a warehouse.

**The table name is used as a LIKE pattern.** `_` and `%` are wildcards there, so `MY_TABLE` also matches a view named `MYXTABLE`. Escaping with a backslash is honoured:

```
> SHOW TERSE VIEWS LIKE 'APPLICABLE_ROLE_'    IN SCHEMA CORTEX.INFORMATION_SCHEMA
APPLICABLE_ROLES        <- trailing _ matched the S

> SHOW TERSE VIEWS LIKE 'APPLICABLE\_ROLE\_' IN SCHEMA CORTEX.INFORMATION_SCHEMA
(no rows)

> SHOW TERSE VIEWS LIKE 'APPLICABLE\_ROLES'  IN SCHEMA CORTEX.INFORMATION_SCHEMA
APPLICABLE_ROLES
```

So this scopes the statement with `IN SCHEMA`, falling back to the database's default catalog/schema and leaving it unscoped when neither is known, and escapes `%`/`_` in the name. Backslashes in the name are left alone so `escapeStringForDatabase`'s `\'` isn't disturbed — Snowflake identifiers don't really contain them, and the old code didn't handle that either.

The escaping is deliberately applied after `escapeStringForDatabase` rather than before, because that method uses a `(?<!\\)'` lookbehind and would skip a quote that already had a backslash in front of it.

Pulled the query construction into `buildShowViewsStatement` so it can be asserted directly. New tests cover the catalog+schema, schema-only, unscoped, and wildcard cases; 3 of the 4 fail before the change:

```
expected: "SHOW VIEWS LIKE 'mytable' IN SCHEMA mycatalog.myschema"
 but was: "SHOW VIEWS LIKE 'mytable'"
expected: "SHOW VIEWS LIKE 'my\_table\%' IN SCHEMA myschema"
 but was: "SHOW VIEWS LIKE 'my_table%'"
```

`mvn -pl liquibase-snowflake test` is green (81 tests). The SHOW output above is from a live Snowflake account, not a mock.

### DIFF
--- a/liquibase-snowflake/src/main/java/liquibase/sqlgenerator/core/SetColumnRemarksGeneratorSnowflake.java
+++ b/liquibase-snowflake/src/main/java/liquibase/sqlgenerator/core/SetColumnRemarksGeneratorSnowflake.java
@@ -9,6 +9,8 @@ import liquibase.executor.ExecutorService;
 import liquibase.sqlgenerator.SqlGeneratorChain;
 import liquibase.statement.core.RawParameterizedSqlStatement;
 import liquibase.statement.core.SetColumnRemarksStatement;
+import liquibase.structure.core.Catalog;
+import liquibase.structure.core.Schema;
 import liquibase.util.ColumnParentType;
 
 import java.util.List;
@@ -41,8 +43,7 @@ public class SetColumnRemarksGeneratorSnowflake extends SetColumnRemarksGenerato
                 // Check if we're trying to set the column remarks on a view, and if so, note that this is not supported.
                 try {
                     List<Map<String, ?>> viewList = Scope.getCurrentScope().getSingleton(ExecutorService.class).getExecutor("jdbc", database).queryForList(
-                            new RawParameterizedSqlStatement(String.format("SHOW VIEWS LIKE '%s'",
-                                database.escapeStringForDatabase(statement.getTableName()))));
+                            buildShowViewsStatement(statement, database));
                     if (!viewList.isEmpty()) {
                         validationErrors.addError(SET_COLUMN_REMARKS_NOT_SUPPORTED_ON_VIEW_MSG);
                     }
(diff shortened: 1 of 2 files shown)


jandroav:
Hi @anjeongkyun, a heads-up before you see some odd activity on this PR.

We replaced the CI for this repository today. The old pipeline ran on `pull_request_target`, which meant it never actually ran for pull requests from forks: you may have noticed your checks showed `authorize` and then nothing useful. The new one runs the full test matrix for everyone, fork or not.

The catch is that a pull request only picks up the new CI on its next event, so PRs opened before today are sitting with no `PR Gate` check, which is now required to merge. I am going to close and immediately reopen this one to trigger it. Nothing is lost, your branch and commits are untouched, and you do not need to do anything.

You should end up with the checks you should have had all along. Sorry for the notification noise.

jandroav:
@anjeongkyun done, and it worked. Your PR now has the full test matrix:

- 7 unit jobs across Java 17, 21 and 25 on Ubuntu, macOS and Windows: all green
- 10 integration jobs (oracle, db2, mariadb, mssql, mysql, postgresql, sqlite, hsqldb, h2, firebird): all green
- `PR Gate`: green

For contrast, your previous run on the old pipeline came back `skipped`. That is what fork PRs got before today: nothing.

One thing left on your side: `PR Labels` is red because the PR needs one release-notes label (`TypeBug`, `TypeEnhancement`, `TypeTest`, `skipReleaseNotes` and a few others). A maintainer can add it, and it is now a required check.

Thanks for the patience, and sorry your change sat without real CI for a week.

anjeongkyun:
Thanks for going and fixing this at the root rather than just nudging my PR through — and for spelling out what fork PRs were actually getting before. I did see `authorize` followed by nothing and assumed it was a first-time-contributor gate waiting on approval, so that clears it up.

`TypeBug` looks right to me if you're picking: the generator was building a `SHOW VIEWS` that matched views outside the target schema and treated `_`/`%` in the table name as wildcards, so `setColumnRemarks` could be rejected on a table because an unrelated schema had a same-named view. Happy to be corrected if you'd file it differently.

One thing worth flagging about the green matrix: Snowflake isn't in the integration list, so those 10 jobs don't exercise this path. What it rests on is the unit tests in the PR and the `SHOW` output in the description, which I ran against a live Snowflake account rather than a mock — including checking that `\_` is honoured as a literal underscore there, since the fix depends on that. If you'd rather see it verified some other way before merging, tell me what would help.

PR #7985, opened 2026-09-08
[TECHOPS-1214] Push the release tag with the liquibase-workflows app token

sayaliM0412:
## Impact

- [ ] Enhancement/New feature
- [ ] Breaking change

## Description

A repository ruleset **cannot grant a bypass to `GITHUB_TOKEN`**. Its app is GitHub's own global `github-actions` app (id `15368`, `owner: github`), which is not an org installation, so GitHub rejects it as a bypass actor:

```
422 Validation Failed
Actor GitHub Actions integration must be part of the ruleset source or owner organization
```

That is what failed the first apply of the `refs/tags/v*` ruleset in [liquibase-infrastructure#4303](https://github.com/liquibase/liquibase-infrastructure/pull/4303). The consequence is structural: **the release tag namespace can never be restricted while the tag is pushed by an identity that can never hold a bypass.**

So the push moves to `liquibase-workflows` (`LIQUIBASE_GITHUB_APP_ID`, 1241802), which is installed in the org and is already a bypass actor on `strict_branch_protection` and `strict_status_checks`.

### What changed in `reversion`

- The existing vault read now also loads the app id and private key, reusing the masking loop already there.
- The token requests **`permission-contents: write` only** — all a tag push needs — and is minted under the same `if:` as the push, so a dry run or a standalone-zip run mints nothing.
- The push uses an explicit `x-access-token` URL rather than a named remote, so the token never reaches `.git/config`. `-c http.https://github.com/.extraheader=` clears the basic-auth header `actions/checkout` leaves behind, which otherwise takes precedence over URL credentials.
- **`persist-credentials` drops to `false`.** Nothing else in this job touches git, so it no longer carries a push-capable `GITHUB_TOKEN` it does not use. That was the only reason the old comment gave for persisting them.

## Order of operations

Merge **after** [liquibase-infrastructure#4340](https://github.com/liquibase/liquibase-infrastructure/pull/4340), which points the bypass at this app. Merging this first is harmless but pointless: the ruleset is at `enforcement = "evaluate"` and blocks nothing either way. Only once both land can enforcement go `active`.

## How was it tested?

- `actionlint` reports **the same 34 findings as `main`, none new**, and the file still parses.
- `LIQUIBASE_GITHUB_APP_ID` in `/vault/liquibase` read back as `1241802`, matching the `liquibase-workflows` installation.
- Confirmed `reversion` has exactly one git operation, the tag push, so dropping persisted credentials affects nothing else.
- Not exercised on a real release yet. This is on the release path, so it wants a dry run before the next release.


.github/workflows/create-release.yml
```diff
@@ -208,11 +209,32 @@ jobs:
...
+      # GITHUB_TOKEN: its app is GitHub's own global one, and GitHub rejects it as a bypass actor.
+      - name: Get GitHub App token
+        id: get-token
+        if: ${{ inputs.standalone_zip == false && inputs.dry_run == false }}
+        uses: actions/create-github-app-token@bcd2ba49218906704ab6c1aa796996da409d3eb1 # v3.2.0
+        with:
+          app-id: ${{ env.LIQUIBASE_GITHUB_APP_ID }}
+          private-key: ${{ env.LIQUIBASE_GITHUB_APP_PRIVATE_KEY }}
+          owner: ${{ github.repository_owner }}
+          permission-contents: write
```
jandroav, line 220: [Major] `owner` without `repositories` scopes this token to every repo the app is installed on, not just this one. `run-test-harness.yml:145` already scopes its own.

```suggestion
          owner: ${{ github.repository_owner }}
          # Without repositories: the token covers every repo the App is installed on.
          repositories: ${{ github.event.repository.name }}
          permission-contents: write
```

.github/workflows/create-release.yml
```diff
@@ -110,7 +111,7 @@ jobs:
         run: |
           set -euo pipefail
           JSON=$(aws secretsmanager get-secret-value --secret-id /vault/liquibase --region us-east-1 --query SecretString --output text)
-          for k in GPG_SECRET GPG_PASSPHRASE LIQUIBOT_PAT_GPM_ACCESS; do
+          for k in GPG_SECRET GPG_PASSPHRASE LIQUIBOT_PAT_GPM_ACCESS LIQUIBASE_GITHUB_APP_ID LIQUIBASE_GITHUB_APP_PRIVATE_KEY; do
```
jandroav, line 108: [Medium] The app private key now lands in `$GITHUB_ENV`, so `Re-version Artifacts` (line 175, which curls and runs re-version.sh) inherits it. `run-test-harness.yml:113` reads the same two keys into step outputs instead.

.github/workflows/create-release.yml
```diff
@@ -208,11 +209,32 @@ jobs:
...
+          VERSION: ${{ needs.setup.outputs.version }}
         run: |
-          git tag -f v${{ needs.setup.outputs.version }}
-          git push -f origin v${{ needs.setup.outputs.version }}
+          set -euo pipefail
+          git tag -f "v${VERSION}"
+          # The token goes in the URL rather than a remote, so it is never persisted to
+          # .git/config. -c clears the basic-auth header actions/checkout would otherwise
+          # leave in place, which takes precedence over URL credentials.
+          git -c http.https://github.com/.extraheader= \
```
jandroav: [Minor] `persist-credentials: false` at line 97 means checkout leaves no extraheader, so this `-c` clears nothing and the comment's stated reason no longer applies.

.github/workflows/create-release.yml
sayaliM0412: Confirmed and fixed in 62f159a. `gh api /orgs/liquibase/installations` shows the liquibase-workflows installation (65890526) as `repository_selection: all`, so `owner` alone minted a contents:write token over every repo in the org. Scoped to `github.event.repository.name`, which is the same repo `GITHUB_REPOSITORY` points the push at.

.github/workflows/create-release.yml
sayaliM0412: Fixed in 62f159a. The step is now `id: vault` and the loop routes `LIQUIBASE_GITHUB_APP_*` to `$GITHUB_OUTPUT` while the other three keys still go to `$GITHUB_ENV`, so the PEM is no longer in the environment `Re-version Artifacts` inherits. Replayed the loop against a fixture: the app id and PEM land in `GITHUB_OUTPUT` only and the PEM's trailing newline survives the heredoc.

.github/workflows/create-release.yml
sayaliM0412: Right: `actions/checkout` calls `removeAuth()` -> `removeToken()` whenever `persist-credentials` is false, so there was no header left to clear. Dropped the flag and the comment in 62f159a.

jandroav (approved):

PR #7986, opened 2026-09-08
[TECHOPS-1264] Document how a release goes and which guards need a person

sayaliM0412:
## Impact

- [ ] Bug fix
- [ ] Breaking change

Documentation only. No workflow file is touched.

## Description

`.github/workflows/README.md` already documents what each release workflow does, and which AWS role and secrets each job reaches. Two things it never said, plus one guard that was missing entirely.

### 1. What a release feels like to run

There was no plain-language account of the sequence. That only **two moments need a person**, and that **one approval covers the whole release**, was only inferable by reading the `needs:` graph and the environment expression together.

Adds a five-step table, and states the behaviour the two-environment split exists to work around: GitHub approves **per job, not per run**, so putting all five gated jobs on `release` would stop a release once per wave of the graph instead of once.

### 2. The guards as a list, not as a by-product

The existing table is organised per job. That answers *"what does this job reach"* but not *"what stops a bad release"*. Adds a per-guard table with a **Person / Automatic** column, and points at the per-job table for the detail rather than duplicating it.

Only two of the seven guards need a person.

### 3. The `refs/tags/v*` tag ruleset

Live, and documented **nowhere in this repository** until now. Recorded along with the fact that it sits at `enforcement = "evaluate"`, so it records rather than blocks.

### 4. `needs: manual-approval` called out on its own

Because the publishing jobs no longer sit behind reviewers themselves, that edge is the only thing keeping the gate in front of them, and removing it would not show up as a change to any environment or any infrastructure code. Worth a sentence where someone reading the release docs will see it.

## Deliberately not changed

- The mermaid diagram already colour-codes jobs by credential (`scoped` / `broad` / `gate` / `plain`), so it needed nothing.
- The per-job guards table is **more** detailed than what is being added here and stays as the reference. The new sections sit above it, so the file reads narrative first, then detail.
- Nothing claims the tag push identity has changed. That is [#7985](https://github.com/liquibase/liquibase/pull/7985) and is not merged, so this describes `main` as it stands.

## How was it tested?

- Every table checked programmatically for a consistent column count across its rows; no ragged tables in the file.
- Heading uses `:traffic_light:` rather than `:shield:` because a `:shield:` section already exists further up.
- Facts taken from the `needs:` keys in `release-published-orchestrator.yml`, the environment expression in the four callees, and the live GitHub environments API (five reviewers, `prevent_self_review`, `can_admins_bypass: false`), not from a PR description.
- No workflow file modified, so no CI behaviour changes.

## Related

- [TECHOPS-1264](https://datical.atlassian.net/browse/TECHOPS-1264), the epic tracking the remaining hardening on this repo
- [TECHOPS-1267](https://datical.atlassian.net/browse/TECHOPS-1267) covers making the `needs` edge enforced rather than documented
- [TECHOPS-1268](https://datical.atlassian.net/browse/TECHOPS-1268) covers taking the tag ruleset out of watch-only

## Release note

[TECHOPS-1264]: https://datical.atlassian.net/browse/TECHOPS-1264?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


.github/workflows/README.md
```diff
@@ -137,6 +137,36 @@ The main coordinator that triggers all release steps in the proper sequence. Sup
...
+
+| Guard | Kind | What it stops | Defined in |
+|---|---|---|---|
+| One approval on the `release` environment | **Person** | A release going out unseen. Five reviewers, self-approval prevented, admins not exempt. | `liquibase-infrastructure` |
+| Publishing the draft release | **Person** | The pipeline starting on its own. A tag on its own does nothing. | the GitHub release UI |
+| `needs: manual-approval` | Automatic | Any publishing job running before the approval. Every one of them depends on that job. | `release-published-orchestrator.yml` |
+| The `approved` input | Automatic | A hand-dispatched publishing workflow skipping the reviewers. Only the orchestrator can set it, so a direct dispatch lands on the reviewed environment instead. | the four publishing callees |
+| Environment branch and tag policies | Automatic | A release credential being reachable from a feature branch. Both environments accept only `main` or a `v*` tag. | `liquibase-infrastructure` |
+| Role trust on the release-scoped role | Automatic | Any job that has not declared a release environment from assuming it at all. | `liquibase-infrastructure` |
```
jandroav: [Major] Not true on `main`: `liquibase-release-oidc-role.tf:63-64` still trusts `refs/tags/v*`, so a job on a v* tag assumes this role with no environment declared. That is TECHOPS-1214 step 5, still open.

.github/workflows/README.md
```diff
@@ -137,6 +137,36 @@ The main coordinator that triggers all release steps in the proper sequence. Sup
 | `docker-release.yml` | Build and push release Docker images | ✅ Yes |
 | `release-publish-assets-s3.yml` | Publish release assets to S3 | ✅ Yes, with one approval |
 
+## :footprints: How a release actually goes
+
+Two moments need a person. Everything between them is automated.
```
jandroav: [Medium] Three rows below are marked **A person** (steps 1, 2 and 4), and step 2 sits between the other two. Either say three, or scope the count to the guards.

jandroav (approved):

PR #7987, opened 2026-09-09
docs: point the README badges at main.yml

jandroav:
Follow-up split out of #7971.

`run-tests.yml` and `nightly-release.yml` are now `disabled_manually` (TECHOPS-1222 cutover), so both badges render for workflows that no longer run, and they will 404 once the files are deleted after the soak.

`main.yml` is what builds `main` and refreshes the `nightly` pre-release now, so both point there. The nightly badge carries `?event=schedule` so it reflects the cron run rather than every push to main.

Split out because the root `README.md` matches the `*` CODEOWNERS rule (@filipelautert / @wwillard7800), which pulled a second code-owner requirement onto a PR that was otherwise entirely under `/.github/workflows/` and already approved by devops.


jandroav:
**Merged via a temporary, logged ruleset bypass, at the request of the epic lead. Recording it so it is not invisible.**

Unlike #7971, this PR had **no review at all**, so the bypass skipped the review requirement rather than a stale-approval technicality. Worth stating plainly. The change is a one-line badge URL and is verifiable from the repository state: `run-tests.yml` and `nightly-release.yml` are `disabled_manually` as of the TECHOPS-1222 cutover today, so both badges pointed at workflows that no longer run.

What was **not** bypassed: all three required checks were green before the merge, and `PR Labels` was red at first for a missing release-notes label, which was fixed rather than waved through.

| | |
|---|---|
| `PR Gate` | success |
| `PR Labels` | success (after adding `skipReleaseNotes`) |
| `mergefreeze` | success |

| Step | Detail |
|---|---|
| Snapshot | `strict_branch_protection` (15726269), `sha256=ecdd2795…` |
| Window | `Team liquibase-devops`, `bypass_mode: pull_request` (PR merges only, never direct push) |
| Restore | Re-fetched and hashed: `sha256=ecdd2795…`, identical |

Bypass actors are back to the original two Integrations. `strict_status_checks` was never touched.

Two of these in one day is two more than ideal. The cheaper fix for next time is structural rather than procedural: keep root-level files out of workflow PRs, since `*` is owned by @filipelautert / @wwillard7800 while `/.github/workflows/` is owned by @liquibase/liquibase-devops, and a single file from the wrong tree doubles the review requirement.

PR #7882, opened 2026-08-07
Fix uppercase acronym parameters being dropped in customChange

fudianchn:
## Impact
- [x] Bug fix (non-breaking change which fixes expected existing functionality)

## Description

Since 4.29.0, `<param>` elements whose name is an uppercase acronym (e.g. `GET`, `POST`, `PUT`, `URL`) are silently dropped when mapped to `customChange` setter methods named `setGET`, `setPOST`, etc. Reported in #7582.

### Root cause

`ObjectMethods.propertyName()` derives the property name from the getter/setter method name by removing the `get`/`set`/`is` prefix and **unconditionally lowercasing the first character**:

```java
return methodName.substring(length, length + 1).toLowerCase(ROOT) + methodName.substring(length + 1);
```

So `setGET` is stored under key `gET` (not `GET`). When `<param name="GET">` is processed, the lookup `getWriteMethod("GET")` misses and returns `null`, so the parameter value is never set.

This contradicts the JavaBeans convention implemented by `java.beans.Introspector.decapitalize`, which keeps a name unchanged when its first two characters are both uppercase (acronyms like `URL`, `GET`). Pre-4.29 liquibase built the method name from the param name (`"set" + "GET"` → `setGET`) and found it directly, so acronym parameters worked.

### Fix

Align `propertyName` with `Introspector.decapitalize`: when the remainder's first two characters are both uppercase, keep the name unchanged; otherwise lowercase the first character as before. This restores the pre-4.29 behavior for acronym parameters (`setGET` → `GET`) without affecting regular camelCase names (`setAge` → `age`, `setName` → `name`).

## Testing
- New `ObjectMethodsTest#writeMethodsPreserveUppercaseAcronyms` verifies `setGET`/`setURL` resolve via `getWriteMethod("GET")`/`getWriteMethod("URL")`, while `setName` still resolves via `getWriteMethod("name")`. Fails before the fix (`getWriteMethod("GET")` returns `null` → NPE), passes after.
- Existing `ObjectMethodsTest` (read/write methods) still pass.
- `ObjectUtilTest` (82 tests) still pass — `ObjectUtil` uses `ObjectMethods` for property mapping, confirming regular camelCase mapping is unaffected.

Fixes #7582

PR #7989, opened 2026-09-09
[TECHOPS-1224] Sweep orphaned branch-named SNAPSHOT versions too

jandroav:
## Why

`cleanup-packages.yml` only matched `^[0-9a-f]{7,40}-SNAPSHOT$`, so it swept commit-shaped versions and skipped every branch-named one. That is the larger pile, and it includes `firebird-sequence-start-value-SNAPSHOT`, the fork-built artifact TECHOPS-1101 raised, still installable from `org.liquibase`.

While reviewing that, Sailee found something bigger: **the commit-shaped half had never deleted anything either.**

## The number changed, please re-read it

An earlier revision of this PR said 1,110 deletions. That figure was correct for the code as it stood and wrong for the code as intended. Fixing the bugs below revived the half of the sweep that was inert, and the real figure is **1,748**.

| Package | delete |
|---|---:|
| `liquibase-core` | 360 |
| `liquibase-extension-testing` | 438 |
| `liquibase-integration-tests` | 438 |
| `liquibase-maven-plugin` | 438 |
| `liquibase` | 74 |
| **Total** | **1,748** |

`org.liquibase.liquibase` going from 0 to 74 is the clearest sign: that package was never touched by the sweep at all.

## What was broken

**The commit-shaped pass never deleted anything.** `classify` compared each commit against `main` and `release`. There is no `release` branch in this repo. Every commit that was not already a branch head or an ancestor of `main` got a 404 on that second compare, the error path marked it `skip (API error)`, the case arm counted it as kept, and the run exited 0.

**`skip (API error)` counted as a keep**, which is what kept the above invisible. An unclassified version is not a kept version.

**422 is not 404.** With those two fixed, the sweep still failed on one commit. `gh api repos/.../commits/f123ee6` returns 422 "No commit found for SHA", not 404, because that seven-character prefix became ambiguous as the repo grew: it matches both a tree and commit `f123ee686`, which **is** an ancestor of `main` and compares `ahead`. Treating 422 as absent would have deleted a version built from merged code. It now means keep, recorded with its reason, and that is the 4 keeps you will see in the summary.

**`bash -e`.** GitHub runs `run:` steps as `bash -e {0}`, which the `set -uo pipefail` line does not turn off. Capturing classify's status with a bare assignment therefore aborted the step on the first orphan. The call stays in a condition context now.

## What changed in the sweep itself

- A second pass for `<branch>-SNAPSHOT`, keeping any version whose branch still exists.
- Ancestry bases derived from the branch list already fetched, so a base that does not exist cannot silently disable a pass. If none exist the run aborts.
- One fetch per package with an explicit `per_page=100`, both passes deriving from that same payload. Two calls classified from two different snapshots, and the API default of 30 paginated this into roughly 124 requests.
- A live branch whose name is hex only would publish a version matching the commit shape and be deleted while the branch was alive. Guarded; no live branch matches today.
- Every removal is written to `$GITHUB_STEP_SUMMARY`. Actions log retention here is 7 days, and 1,748 irreversible deletions whose only record expires in a week are not reviewable afterwards.

## The pile is disarmed, not stopped

`build-branch.yml` was deleted at the TECHOPS-1222 cutover, but a `pull_request_target` run queued before then still executes the workflow file captured when it was queued. [Run 31821424949](https://github.com/liquibase/liquibase/actions/runs/31821424949) was created 2026-08-14, ran 2026-09-09 12:33Z after the deletion, and published `fix-postgres-jsonb-defaults-SNAPSHOT`. Several hundred runs remain awaiting approval, so approving a stale one republishes a name this sweep has removed. Tracked separately.

## Verification

Dispatched with `dry_run: true` on this branch: [run 34813322353](https://github.com/liquibase/liquibase/actions/runs/34813322353), green. 1,744 orphans, 4 keeps for the unresolvable commit, **0 unclassified**, nothing deleted. The deletion set is in the run summary rather than discovered at the Sunday 05:30 cron.

TECHOPS-1224, AC2.

### DIFF
--- a/.github/workflows/cleanup-packages.yml
+++ b/.github/workflows/cleanup-packages.yml
@@ -6,12 +6,25 @@ name: Cleanup Packages
 #                       branch head, a tag target, or an ancestor of main/release
 #   <branch>-SNAPSHOT   deleted when no branch of that name exists any more
 #
-# The branch-named half was missed originally, and it is the larger pile: on
-# 2026-09-09 liquibase-core had 830 versions, 229 branch-named, and 218 of those
-# pointed at branches that no longer exist. `firebird-sequence-start-value-SNAPSHOT`
-# among them, the fork-built artifact from TECHOPS-1101, still installable from
-# org.liquibase 22 days after it was reported. Per-PR publishing stopped at the
-# TECHOPS-1222 cutover so the pile no longer grows, but nothing removes it.
+# The branch-named half was missed originally, and it is the larger pile. Measured
+# 2026-09-10 across all five packages: 1,157 branch-named versions, 47 live-branch
+# keeps, 1,110 deletions (219 core / 297 extension-testing / 297 integration-tests /
+# 297 maven-plugin / 0 liquibase). `firebird-sequence-start-value-SNAPSHOT` is among
+# them, the fork-built artifact from TECHOPS-1101, still installable from org.liquibase.
+#
+# Per-PR publishing is disarmed on `main`, but it is re-armable, not stopped:
+# `build-branch.yml` was deleted at the TECHOPS-1222 cutover, yet a `pull_request_target`
+# run queued before the cutover still executes the workflow file captured when it was
+# queued. Run 31821424949 was created 2026-08-14 and ran 2026-09-09 12:33Z, after the
+# deletion, and published `fix-postgres-jsonb-defaults-SNAPSHOT`. Several hundred runs
+# remain in `action_required`/`waiting`, so approving a stale one can republish a name
+# this sweep has already removed. Purging that queue is tracked separately.
+#
+# Related: `cleanup-main-builds.yml` also deletes package versions, on every published
+# release. It omits `num-old-versions-to-delete`, whose default is 1 and which takes
+# precedence over `min-versions-to-keep`, so its `min-versions-to-keep: 0` is inert and
+# it removes one oldest version per package per release regardless of shape, with only
+# `^DAT-.*-SNAPSHOT$` protected. Not changed here; flagged for its own ticket.
 # [TECHOPS-1224, replaces cleanup-branch-builds.yml under TECHOPS-1099]
 
 on:
@@ -55,12 +68,26 @@ jobs:
           KEEP_SHAS=$(jq -r '.[].commit.sha' <<<"$BRANCHES")
           KEEP_SHAS+=$'\n'$(gh api --paginate "repos/$REPO/tags?per_page=100" --jq '.[].commit.sha' || true)
 
-          # Branch names to keep, with / turned into _ because that is the munging
-          # snapshot-branch.yml applies when it names the artifact.
+          # Branch names to keep, with / turned into _ because that is the munging the
+          # now-deleted build.yml / build-branch.yml applied when naming the package
+          # version. snapshot-branch.yml munges only its run-artifact filenames; the
+          # version it publishes is <full-sha>-SNAPSHOT, handled by the sha pass.
           KEEP_BRANCHES=$(jq -r '.[].name' <<<"$BRANCHES" | tr '/' '_')
           [ -z "$KEEP_BRANCHES" ] && { echo "::error::Empty branch list; aborting"; exit 1; }
 
           declare -A VERDICT # sha -> keep|orphan|skip, shared across packages
+
+          # Actions log retention here is 7 days, and a sweep of this size needs a record
+          # that outlives it, so every removal is also written to the run summary.
+          deleted_total=0
+          kept_total=0
+          if [ "$DRY_RUN" = true ]; then mode="dry run, nothing deleted"; else mode="delete"; fi
+          {
+            echo "## Cleanup Packages"
+            echo
+            echo "Mode: $mode"
+            echo
+          } >> "$GITHUB_STEP_SUMMARY"
 
           # Prints why a SHA is kept and returns 0; returns 1 if it is orphaned; returns 2 on API trouble.
           classify() {
@@ -80,13 +107,31 @@ jobs:
           failed=0
           for pkg in $PACKAGES; do
             echo "::group::$pkg"
-            versions=$(gh api --paginate "orgs/$ORG/packages/maven/$pkg/versions" \
-              --jq '.[] | select(.name | test("^[0-9a-f]{7,40}-SNAPSHOT$")) | "\(.id) \(.name)"') || {
+            pkg_deleted=0
+
+            # One fetch per package, two derivations, for the same reason as the branch
+            # list above: two calls classify from two different snapshots of the same
+            # list. per_page is explicit because the API default of 30 paginated this
+            # into roughly 124 requests across the five packages.
+            allversions=$(gh api --paginate "orgs/$ORG/packages/maven/$pkg/versions?per_page=100" | jq -s 'add // []') || {
               echo "::error::Failed to list versions for $pkg"; failed=1; echo "::endgroup::"; continue; }
+
+            versions=$(jq -r '.[] | select(.name | test("^[0-9a-f]{7,40}-SNAPSHOT$")) | "\(.id) \(.name)"' <<<"$allversions")
+            # The branch selector excludes the sha shape so the two passes stay disjoint:
+            # a version is classified by exactly one of them.
+            bversions=$(jq -r '.[] | select(.name | endswith("-SNAPSHOT")) | select(.name | test("^[0-9a-f]{7,40}-SNAPSHOT$") | not) | "\(.id) \(.name)"' <<<"$allversions")
 
             while read -r id name; do
               [ -z "$id" ] && continue
               sha=${name%-SNAPSHOT}
+              # A live branch whose name is hex only, 7 to 40 characters, produces a
+              # version that matches the sha shape, and the branch pass excludes it by
+              # selector. Without this check it would be looked up as a commit, fail to
+              # resolve, and be deleted while its branch is alive. No live branch matches
+              # today, so this guards a latent case rather than an observed one.
+              if grep -qxF "$sha" <<<"$KEEP_BRANCHES"; then
+                echo "KEEP (live branch): $name"; kept_total=$((kept_total + 1)); continue
+              fi
               if [ -z "${VERDICT[$sha]:-}" ]; then
                 if reason=$(classify "$sha"); then VERDICT[$sha]="keep ($reason)"
                 elif [ $? -eq 1 ]; then VERDICT[$sha]=orphan
@@ -94,31 +139,33 @@ jobs:
               fi
               case "${VERDICT[$sha]}" in
                 orphan)
-                  if [ "$DRY_RUN" = true ]; then echo "ORPHAN (dry run): $name"
-                  elif gh api -X DELETE "orgs/$ORG/packages/maven/$pkg/versions/$id" >/dev/null 2>&1; then echo "DELETED: $name"
-                  else echo "::warning::Failed to delete $pkg $name (id $id)"; failed=1; fi ;;
-                *) echo "${VERDICT[$sha]^^}: $name" ;;
+                  if [ "$DRY_RUN" = true ]; then echo "ORPHAN (dry run): $name"; pkg_deleted=$((pkg_deleted + 1))
+                  elif gh api -X DELETE "orgs/$ORG/packages/maven/$pkg/versions/$id" >/dev/null 2>&1; then echo "DELETED: $name"; pkg_deleted=$((pkg_deleted + 1))
+                  else echo "::warning::Failed to delete $pkg $name (id $id)"; failed=1; continue; fi
+                  echo "- \`$pkg\` \`$name\`" >> "$GITHUB_STEP_SUMMARY" ;;
+                *) echo "${VERDICT[$sha]^^}: $name"; kept_total=$((kept_total + 1)) ;;
               esac
             done <<< "$versions"
-
-            # Branch-named versions. The selector deliberately excludes the sha shape so
-            # the two passes stay disjoint: a version is classified by exactly one of them.
-            bversions=$(gh api --paginate "orgs/$ORG/packages/maven/$pkg/versions" \
-              --jq '.[] | select(.name | endswith("-SNAPSHOT")) | select(.name | test("^[0-9a-f]{7,40}-SNAPSHOT$") | not) | "\(.id) \(.name)"') || {
-              echo "::error::Failed to list branch-named versions for $pkg"; failed=1; echo "::endgroup::"; continue; }
 
             while read -r id name; do
               [ -z "$id" ] && continue
               branch=${name%-SNAPSHOT}
               if [ -n "$branch" ] && grep -qxF "$branch" <<<"$KEEP_BRANCHES"; then
-                echo "KEEP (live branch): $name"
-                continue
+                echo "KEEP (live branch): $name"; kept_total=$((kept_total + 1)); continue
               fi
-              if [ "$DRY_RUN" = true ]; then echo "ORPHAN (dry run): $name"
-              elif gh api -X DELETE "orgs/$ORG/packages/maven/$pkg/versions/$id" >/dev/null 2>&1; then echo "DELETED: $name"
-              else echo "::warning::Failed to delete $pkg $name (id $id)"; failed=1; fi
+              if [ "$DRY_RUN" = true ]; then echo "ORPHAN (dry run): $name"; pkg_deleted=$((pkg_deleted + 1))
+              elif gh api -X DELETE "orgs/$ORG/packages/maven/$pkg/versions/$id" >/dev/null 2>&1; then echo "DELETED: $name"; pkg_deleted=$((pkg_deleted + 1))
+              else echo "::warning::Failed to delete $pkg $name (id $id)"; failed=1; continue; fi
+              echo "- \`$pkg\` \`$name\`" >> "$GITHUB_STEP_SUMMARY"
             done <<< "$bversions"
 
+            deleted_total=$((deleted_total + pkg_deleted))
+            echo "$pkg: $pkg_deleted removed"
             echo "::endgroup::"
           done
+
+          {
+            echo
+            echo "**Total: $deleted_total removed, $kept_total kept.**"
+          } >> "$GITHUB_STEP_SUMMARY"
           exit $failed
--- a/.github/workflows/cleanup-packages.yml
+++ b/.github/workflows/cleanup-packages.yml
@@ -1,10 +1,18 @@
 name: Cleanup Packages
 
-# Weekly sweep of orphaned <sha>-SNAPSHOT versions in GitHub Packages: versions
-# whose commit no longer exists, or is not a branch head, a tag target, or an
-# ancestor of main/release. Replaces cleanup-branch-builds.yml, whose `delete`
-# trigger ignored its branch filters and whose per-branch job had nothing left
-# to delete once per-PR SNAPSHOT publishing stopped. [TECHOPS-1099]
+# Weekly sweep of orphaned SNAPSHOT versions in GitHub Packages. Two shapes:
+#
+#   <sha>-SNAPSHOT      deleted when the commit no longer exists, or is not a
+#                       branch head, a tag target, or an ancestor of main/release
+#   <branch>-SNAPSHOT   deleted when no branch of that name exists any more
+#
+# The branch-named half was missed originally, and it is the larger pile: on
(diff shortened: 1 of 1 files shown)


.github/workflows/cleanup-packages.yml
```diff
@@ -83,6 +100,25 @@ jobs:
                 *) echo "${VERDICT[$sha]^^}: $name" ;;
               esac
             done <<< "$versions"
+
+            # Branch-named versions. The selector deliberately excludes the sha shape so
+            # the two passes stay disjoint: a version is classified by exactly one of them.
+            bversions=$(gh api --paginate "orgs/$ORG/packages/maven/$pkg/versions" \
+              --jq '.[] | select(.name | endswith("-SNAPSHOT")) | select(.name | test("^[0-9a-f]{7,40}-SNAPSHOT$") | not) | "\(.id) \(.name)"') || {
```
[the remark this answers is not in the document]
sayaliM0412: @jandroav : could we verify this?

sayaliM0412 (commented):
Verified every claim below against the live GitHub and Packages APIs before posting: the branch and tag lists, the version lists of all five packages, workflow run job timestamps, and the pinned action's `action.yml`.

The classification logic itself holds up on today's data: `main-SNAPSHOT` is kept, no orphan collides with any of the 168 live tags, and no release version lands in the delete set. Three highs and four mediums below.

.github/workflows/cleanup-packages.yml
```diff
@@ -1,10 +1,18 @@
...
+#   <sha>-SNAPSHOT      deleted when the commit no longer exists, or is not a
+#                       branch head, a tag target, or an ancestor of main/release
+#   <branch>-SNAPSHOT   deleted when no branch of that name exists any more
+#
+# The branch-named half was missed originally, and it is the larger pile: on
+# 2026-09-09 liquibase-core had 830 versions, 229 branch-named, and 218 of those
+# pointed at branches that no longer exist. `firebird-sequence-start-value-SNAPSHOT`
+# among them, the fork-built artifact from TECHOPS-1101, still installable from
+# org.liquibase 22 days after it was reported. Per-PR publishing stopped at the
+# TECHOPS-1222 cutover so the pile no longer grows, but nothing removes it.
```
sayaliM0412: **High:** the pile can still grow. `build-branch.yml` is deleted from `main`, but its queued fork runs still execute the pre-cutover file on approval: [run 31821424949](https://github.com/liquibase/liquibase/actions/runs/31821424949) did exactly that today, 12:29-12:33 UTC on fork PR [#7888](https://github.com/liquibase/liquibase/pull/7888), with `Publish package` succeeding and `fix-postgres-jsonb-defaults-SNAPSHOT` created at 12:32:35Z, and [run 32256893441](https://github.com/liquibase/liquibase/actions/runs/32256893441) did the same on 2026-09-01.

52 more runs are still queued (45 `waiting`, 6 `pending`, 1 `action_required`), 14 of them created in the last 30 days, and two of those 14 are `firebird-sequence-start-value`: the exact artifact [TECHOPS-1101](https://datical.atlassian.net/browse/TECHOPS-1101) raised, which can be republished after this sweep deletes it. Could the comment say re-armable rather than stopped?

.github/workflows/cleanup-packages.yml
```diff
@@ -1,10 +1,18 @@
...
-# trigger ignored its branch filters and whose per-branch job had nothing left
-# to delete once per-PR SNAPSHOT publishing stopped. [TECHOPS-1099]
+# Weekly sweep of orphaned SNAPSHOT versions in GitHub Packages. Two shapes:
+#
+#   <sha>-SNAPSHOT      deleted when the commit no longer exists, or is not a
+#                       branch head, a tag target, or an ancestor of main/release
+#   <branch>-SNAPSHOT   deleted when no branch of that name exists any more
+#
+# The branch-named half was missed originally, and it is the larger pile: on
+# 2026-09-09 liquibase-core had 830 versions, 229 branch-named, and 218 of those
```
sayaliM0412: **High:** these figures are `liquibase-core` only. Replaying this selector and keep-check across all five packages gives 1,157 branch-named versions, 47 keeps and **1,110 deletions** (219 / 297 / 297 / 297 / 0), so the cron removes roughly 5x what a reviewer reading 218 thinks they are approving. Worth stating the all-package total here and in the PR body.

.github/workflows/cleanup-packages.yml
```diff
@@ -83,6 +100,25 @@ jobs:
...
+            while read -r id name; do
+              [ -z "$id" ] && continue
+              branch=${name%-SNAPSHOT}
+              if [ -n "$branch" ] && grep -qxF "$branch" <<<"$KEEP_BRANCHES"; then
+                echo "KEEP (live branch): $name"
+                continue
+              fi
+              if [ "$DRY_RUN" = true ]; then echo "ORPHAN (dry run): $name"
+              elif gh api -X DELETE "orgs/$ORG/packages/maven/$pkg/versions/$id" >/dev/null 2>&1; then echo "DELETED: $name"
+              else echo "::warning::Failed to delete $pkg $name (id $id)"; failed=1; fi
```
sayaliM0412: **High:** 1,110 deletions would leave no durable record. There is no `$GITHUB_STEP_SUMMARY` and no uploaded artifact, and this repo's Actions retention is 7 days (today's artifacts carry `expires_at` = run start + 7d), so within a week nothing says what was removed: the 2026-08-20 `cleanup-main-builds.yml` log already returns HTTP 410. Could the deleted names go to `$GITHUB_STEP_SUMMARY` as well as the log?

.github/workflows/cleanup-packages.yml
sayaliM0412: **Medium:** the bot's SHA-shaped-live-branch finding above is not reproducible today, 0 of the 177 live branches match `^[0-9a-f]{7,40}$`, so it is latent rather than blocking. Its one-line `KEEP_BRANCHES` check before SHA classification is cheap enough to just take.

.github/workflows/cleanup-packages.yml
```diff
@@ -83,6 +100,25 @@ jobs:
                 *) echo "${VERDICT[$sha]^^}: $name" ;;
               esac
             done <<< "$versions"
+
+            # Branch-named versions. The selector deliberately excludes the sha shape so
+            # the two passes stay disjoint: a version is classified by exactly one of them.
+            bversions=$(gh api --paginate "orgs/$ORG/packages/maven/$pkg/versions" \
```
sayaliM0412: **Medium:** this re-fetches the version list already pulled 20 lines above, and neither call sets `per_page`, so each pass paginates at the default 30 (about 124 pages across the five packages) and the two passes classify from two different snapshots of the same list. One fetch reused by both passes would match the one-fetch reasoning applied to `branches`.

.github/workflows/cleanup-packages.yml
```diff
@@ -38,11 +46,20 @@ jobs:
...
+            echo "::error::Failed to fetch branches; aborting to avoid incorrect deletions"; exit 1; }
+
           # SHAs to keep: every branch head and tag target (snapshot-branch.yml accepts tags).
-          KEEP_SHAS=$(gh api --paginate "repos/$REPO/branches?per_page=100" --jq '.[].commit.sha') || {
-            echo "::error::Failed to fetch branch heads; aborting to avoid incorrect deletions"; exit 1; }
+          KEEP_SHAS=$(jq -r '.[].commit.sha' <<<"$BRANCHES")
           KEEP_SHAS+=$'\n'$(gh api --paginate "repos/$REPO/tags?per_page=100" --jq '.[].commit.sha' || true)
 
+          # Branch names to keep, with / turned into _ because that is the munging
+          # snapshot-branch.yml applies when it names the artifact.
```
sayaliM0412: **Medium:** [`snapshot-branch.yml`](https://github.com/liquibase/liquibase/blob/main/.github/workflows/snapshot-branch.yml) publishes `<full-sha>-SNAPSHOT` and applies the `/`-to-`_` munging only to run-artifact names, never to a package version, so the real justification for munging here is the deleted `build.yml` / `build-branch.yml` that actually produced these branch-named versions.

.github/workflows/cleanup-packages.yml
```diff
@@ -1,10 +1,18 @@
...
+#                       branch head, a tag target, or an ancestor of main/release
+#   <branch>-SNAPSHOT   deleted when no branch of that name exists any more
+#
+# The branch-named half was missed originally, and it is the larger pile: on
+# 2026-09-09 liquibase-core had 830 versions, 229 branch-named, and 218 of those
+# pointed at branches that no longer exist. `firebird-sequence-start-value-SNAPSHOT`
+# among them, the fork-built artifact from TECHOPS-1101, still installable from
+# org.liquibase 22 days after it was reported. Per-PR publishing stopped at the
+# TECHOPS-1222 cutover so the pile no longer grows, but nothing removes it.
+# [TECHOPS-1224, replaces cleanup-branch-builds.yml under TECHOPS-1099]
```
sayaliM0412, line 28: **Medium:** [`cleanup-main-builds.yml`](https://github.com/liquibase/liquibase/blob/main/.github/workflows/cleanup-main-builds.yml) is still active on `main` and deletes package versions on every published release, so it is worth reconciling with this sweep. It omits `num-old-versions-to-delete`, whose [default is `1`](https://github.com/actions/delete-package-versions/blob/e5bc658cc4c965c472efe991f8beea3981499c55/action.yml), so its `min-versions-to-keep: 0` is inert and it quietly removes one oldest version per package per release, whatever that version is: only `^DAT-.*-SNAPSHOT$` is protected.

.github/workflows/cleanup-packages.yml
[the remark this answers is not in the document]
jandroav: Correct, and fixed in `2139c2333`.

Traced the path: a branch named `deadbeef` publishes `deadbeef-SNAPSHOT`, which matches the sha selector, and the branch pass excludes it with `test(...) | not`. It then reaches `classify`, fails to resolve as a commit, and is deleted while its branch is alive.

Took your fix, moved to just before the `VERDICT` lookup so the keep is recorded without spending an API call.

Not reproducible today: 0 of the 178 live branches match `^[0-9a-f]{7,40}$`, so this closes a latent case rather than an observed one. Worth closing anyway, since the cost is one `grep` and the failure mode is deleting a live branch's artifact.

.github/workflows/cleanup-packages.yml
jandroav: Verified, and you are right. Rewrote the comment.

`gh run view 31821424949` confirms it: workflow `Build Branch SNAPSHOT`, event `pull_request_target`, created 2026-08-14, **ran 2026-09-09 12:33:55Z**, `Build & Package` success. That is after `build-branch.yml` was deleted from `main`. A queued run executes the workflow file captured when it was queued, so deleting the file disarms new runs but does nothing to the backlog.

The queue is larger than the comment implied: 212 `action_required`, 91 `waiting`, 13 `pending` across the repo.

Corroborating detail: `liquibase-core` had 218 orphaned branch-named versions when I opened this PR and has 219 now. The new one is `fix-postgres-jsonb-defaults-SNAPSHOT`, from exactly that run.

The header now says re-armable, names the run, and states that approving a stale one can republish a name this sweep removed. Purging the queue is a separate operational action, not something this workflow can do, so I have flagged it rather than folded it in.

.github/workflows/cleanup-packages.yml
jandroav: Reproduced your figures exactly. Fixed in the header and I will update the PR body.

```
                              branch-named  keeps  delete
liquibase-core                     230       11     219
liquibase-extension-testing        309       12     297
liquibase-integration-tests        309       12     297
liquibase-maven-plugin             309       12     297
liquibase                            0        0       0
TOTAL                             1157       47    1110
```

You are right about the framing, and it is the part that matters: 218 was `liquibase-core` alone, so a reviewer reading it would approve five times less than the cron actually removes. The header now carries the all-package total with the per-package split.

.github/workflows/cleanup-packages.yml
jandroav: Agreed, and added. Every removal now goes to `$GITHUB_STEP_SUMMARY` as `- \`<package>\` \`<version>\``, plus a total line, and the summary states whether the run was a dry run.

The 7-day retention point is the deciding one: an irreversible deletion of 1,110 versions whose only record expires in a week is not reviewable after the fact, and your `cleanup-main-builds.yml` HTTP 410 example shows that is not hypothetical.

.github/workflows/cleanup-packages.yml
jandroav: Confirmed latent, not blocking: 0 of the 178 live branches match `^[0-9a-f]{7,40}$` today.

Took the fix anyway, in `2139c2333`. One `grep -qxF` against a list already in memory, against a failure mode that deletes a live branch's artifact.

.github/workflows/cleanup-packages.yml
jandroav: Both parts correct. Fixed.

There is now one fetch per package with an explicit `per_page=100`, and both passes derive from that same payload with `jq`, mirroring the one-fetch reasoning the branch list already used. The two-snapshot race goes with it.

Verified the derivation is lossless on `liquibase-core`: 832 versions fetched, 599 in the sha pass, 230 in the branch pass, sum 829, and 829 versions end in `-SNAPSHOT`. Disjoint, nothing dropped.

.github/workflows/cleanup-packages.yml
jandroav: Correct, the comment misattributed it. Fixed.

`snapshot-branch.yml:57` sets `BRANCH_FILE=$(tr '/' _ <<<"$REF")` and uses it only for run-artifact filenames (`artifacts-named/liquibase-${BRANCH_FILE}.tar.gz`). The version it publishes is `<full-sha>-SNAPSHOT`, which the sha pass handles.

The comment now says the munging matches what the deleted `build.yml` / `build-branch.yml` applied when naming the package version, which is what actually produced these.

.github/workflows/cleanup-packages.yml
jandroav: Verified, including the default. `actions/delete-package-versions` at the pinned `e5bc658` declares `num-old-versions-to-delete` with `default: "1"`, and our own comment in that file notes it takes precedence over `min-versions-to-keep`. So `min-versions-to-keep: 0` is inert and each published release removes one oldest version per package across all five, whatever it is, with only `^DAT-.*-SNAPSHOT$` protected.

Documented it in this file's header as a known interaction, but deliberately not changed here: it is a different trigger, a different failure mode, and changing deletion behaviour in a second workflow inside a PR about the sweep would make this harder to review, not easier. Opening it as its own ticket. Say the word if you would rather it rode along.

sayaliM0412:
The [SHA half](https://github.com/liquibase/liquibase/blob/42154efc83b21c99e093f5b235bbb52ad7f04a46/.github/workflows/cleanup-packages.yml#L124-L148) of the sweep is dead code: [`release`](https://github.com/liquibase/liquibase/blob/42154efc83b21c99e093f5b235bbb52ad7f04a46/.github/workflows/cleanup-packages.yml#L100-L101) does not exist in this repo, so every still-existing commit 404s on the second compare and is reported as kept, which is why the [dry run](https://github.com/liquibase/liquibase/actions/runs/34444949750) deleted 0 SHA-shaped versions while exiting green.

jandroav:
Correct, and taken in `53bead4c5`. Your finding and Sailee's turn out to be the same bug seen from two ends.

The mechanism you describe is what hid it: `skip (API error)` fell through `*)`, counted as kept, left `failed` untouched, and the run exited 0.

What it was hiding: `classify` compared each commit against `main` and `release`, and **there is no `release` branch in this repo**. So every sha that was not already a branch head or an ancestor of `main` got a 404 on that second compare, `|| return 2` marked it `skip (API error)`, and your case arm counted it as kept. The sha half of this sweep has never deleted anything since it was written.

Both halves fixed:

- `skip (API error)` now sets `failed=1` and does not increment `kept_total`, as you proposed. An unclassified version is not a kept version.
- The ancestry bases are derived from the branch list already fetched, so a base that does not exist cannot silently disable the pass. If none of the expected bases exist the run aborts rather than keeping everything.

Worth recording that the evidence was already in my own dry run: 634 versions came back as `SKIP` and I did not ask why. The finding-count check I was relying on cannot catch this, because the workflow was behaving exactly as written.

.github/workflows/cleanup-packages.yml
```diff
@@ -63,26 +107,65 @@ jobs:
...
                 orphan)
-                  if [ "$DRY_RUN" = true ]; then echo "ORPHAN (dry run): $name"
-                  elif gh api -X DELETE "orgs/$ORG/packages/maven/$pkg/versions/$id" >/dev/null 2>&1; then echo "DELETED: $name"
-                  else echo "::warning::Failed to delete $pkg $name (id $id)"; failed=1; fi ;;
-                *) echo "${VERDICT[$sha]^^}: $name" ;;
+                  if [ "$DRY_RUN" = true ]; then echo "ORPHAN (dry run): $name"; pkg_deleted=$((pkg_deleted + 1))
+                  elif gh api -X DELETE "orgs/$ORG/packages/maven/$pkg/versions/$id" >/dev/null 2>&1; then echo "DELETED: $name"; pkg_deleted=$((pkg_deleted + 1))
+                  else echo "::warning::Failed to delete $pkg $name (id $id)"; failed=1; continue; fi
+                  echo "- \`$pkg\` \`$name\`" >> "$GITHUB_STEP_SUMMARY" ;;
+                *) echo "${VERDICT[$sha]^^}: $name"; kept_total=$((kept_total + 1)) ;;
```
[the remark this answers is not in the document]
jandroav: Correct, and taken in `53bead4c5`. Your finding and Sailee's are the same bug seen from two ends.

The mechanism you describe is what hid it. What it was hiding: `classify` compared each commit against `main` and `release`, and there is **no `release` branch in this repo**. Every sha that was not already a branch head or an ancestor of `main` got a 404 on that second compare, `|| return 2` marked it `skip (API error)`, and this case arm counted it as kept. The sha half of the sweep has never deleted anything.

Both ends fixed: `skip (API error)` now sets `failed=1` and does not increment `kept_total`, as you proposed, and the ancestry bases are derived from the branch list already fetched so a base that does not exist cannot silently disable the pass.

The evidence was in my own dry run and I did not read it: 634 versions came back as `SKIP`. An actionlint finding-count check could never catch this, because the workflow was doing exactly what it said.

jandroav:
You were right, and it went deeper than the one bug. Fixed across `53bead4c5`, the retry commit, `b965b54e5` and `c33f596f4`.

**Your finding, confirmed.** There is no `release` branch in this repo, so the second compare in `classify` returned 404 for every commit that was not already a branch head or an ancestor of `main`. The error path marked those `skip (API error)`, the case arm counted them as kept, and the run exited 0. The commit-shaped half has never deleted anything.

The evidence was in my own dry run and I did not read it: **634 versions came back as SKIP** and I did not ask why, because the 1,110 orphans from the branch pass matched the figure I was checking against. I verified what I expected and ignored what was left over.

**Two more surfaced once that was fixed**, both worth recording:

`gh api repos/.../commits/f123ee6` returns **422 "No commit found for SHA"**, not 404. Treating that as absent would have been worse than the original bug: that seven-character prefix became ambiguous as the repo grew, and one of its candidates is commit `f123ee686`, "Fix uppercase acronym parameters being dropped in customChange (#7882)", which **is** an ancestor of `main` and compares `ahead`. Deleting on 422 would have removed a package version built from merged code. 422 now means keep, with its reason recorded.

And GitHub runs `run:` steps as `bash -e {0}`, which the `set -uo pipefail` line does not turn off. My first attempt at capturing classify's status used a bare assignment, which aborted the step on the first orphan. My local test could not reproduce it because my own shell has no `-e`.

**The number you reviewed has changed, and this is the part I most want in front of you.** You reviewed 1,110 deletions. With the commit-shaped pass working, it is **1,748**:

```
liquibase-core                360   (was 219)
liquibase-extension-testing   438   (was 297)
liquibase-integration-tests   438   (was 297)
liquibase-maven-plugin        438   (was 297)
liquibase                      74   (was   0)
```

That last row is the sharpest: a package the sweep never touched at all.

Verified on [run 34813322353](https://github.com/liquibase/liquibase/actions/runs/34813322353), green: 1,744 orphans, 4 keeps for the unresolvable commit, **0 unclassified**, nothing deleted. PR body updated with the new figure.

This is the same problem you raised when I quoted 218 instead of 1,110, so I would rather hand it back for a fresh look than carry the old number into a merge.

sayaliM0412 (approved):

PR #7990, opened 2026-09-10
[TECHOPS-1223] Build the release from a pinned commit, not from an operator-supplied run id

jandroav:
## Why

A real release today ships bytes whose only link to the version being released is an operator typing a run id into a form.

`create-release.yml` takes a `runId` input, downloads the `liquibase-artifacts` bundle from that run, signs it with GPG, and force-pushes the release tag. Nothing in the file checks that the run's commit is the commit being released: there is no `head_sha` comparison anywhere. The `reversion` job's `actions/checkout` declares no `ref:` either, so it checks out whatever ref the dispatch used, not the tag.

Two consequences, neither hypothetical:

- An operator pastes a stale run id and the pipeline tags current `main` while publishing binaries built from a different commit.
- `pr.yml` runs on plain `pull_request`, so workflow YAML comes from the PR head. A fork PR can add a job that uploads an artifact named `liquibase-artifacts` (artifact upload uses the per-job runtime token, so the workflow's `contents: read` does not prevent it). Turning that into a release still needs a maintainer to paste the run id, so this is social engineering rather than a direct exploit, but the pipeline offers no defence against it.

TECHOPS-1222 deleted the old *producers* (`run-tests.yml`, `build.yml`). This PR closes the *consumer* side.

## What changed

Trust now flows commit -> build -> tag, never tag -> artifact.

- `setup` resolves `inputs.branch` to an immutable 40-hex SHA, validates `inputs.version` against `^[A-Za-z0-9._-]+$` before it becomes a git ref or a filename, and rejects a pre-existing tag before any build minute is spent.
- `reversion` checks out that SHA and builds the reactor at the release version. Both checkouts pin `needs.setup.outputs.sha`.
- The tag is created last, from that SHA, with `gh api --method POST repos/{}/git/refs`. That call is atomic and has no `force` parameter, so "never repoint a released tag" becomes structural rather than a discipline. `git push -f` is gone.
- `re-version.sh` (188 lines), `ManifestReversion.java` (45) and `settings-releases.xml` (30) are deleted. The script existed only because `main.yml` builds `main-SNAPSHOT`; building at the release version makes its work native output. `settings-releases.xml` went with the long-lived `LIQUIBOT_PAT_GPM_ACCESS` org PAT it configured, replaced by `setup-java server-id: liquibase` plus the ephemeral job token, the same configuration `ci-test.yml` already uses.
- The `curl` block that fetched three executable files from `raw.githubusercontent.com` into a job holding `GPG_PASSPHRASE` is gone. `sign-artifacts.sh` now runs from the checked-out tree.
- `reversion` and `build-installers` gain `environment: ${{ inputs.dry_run && 'release-publish' || 'release' }}`, matching the pattern the four publishing workflows already use. No Terraform needed: `release-publish` already exists with 0 reviewers.
- Step order in `reversion` is build first, credentials second. Before this PR the job ran no Maven, so loading GPG early was harmless. Once it builds the reactor, that ordering would leave the GPG private key resident while dozens of third-party plugins execute.
- `reversion` drops to `contents: read`, `packages: read`, `id-token: write`. `contents: write` and `actions: read` are removed.

Two lines outside `.github/`: `liquibase-standard/src/main/resources/liquibase.build.properties` hardcoded the literal `build.version=DEV` while every sibling line used a `${...}` placeholder, and `re-version.sh` was the only thing that ever rewrote it. Without a fix, a natively built release would ship with `LiquibaseUtil.isDevVersion()` true, which makes `ChangeFactory` and `SnapshotGeneratorFactory` **throw** `UnexpectedLiquibaseException` instead of warning for third-party extensions, disables analytics, and resolves XSDs as `latest`. The resource now reads `${build.version}` with a `DEV` default in the root `pom.xml`, so dev, PR and `main` builds are unchanged.

## Evidence

Dispatched `create-release.yml` on this branch with `dry_run: true`, building this branch's source. [Run 34432707689](https://github.com/liquibase/liquibase/actions/runs/34432707689), all three jobs green. The produced `liquibase-core` jar contains:

```
build.version=dry-run-1223-051636
build.commit=131ef8bd3801df78900346acf285c952c559945e
```

`build.commit` is this branch's HEAD exactly. The shipped artifact now carries, verifiably, the commit it was built from.

Step order confirmed in that run: checkout(SHA), JDK, build, stage, AWS creds, vault, GPG, sign, upload, then `Get App credentials from vault`, `Get GitHub App token` and `Create release tag` all skipped under `dry_run`.

Other checks:

- `actionlint` findings on `create-release.yml` went from 5 to 4. The finding that disappeared is the always-empty `uber_jar_runId` output, whose reader used key `run_id` while the writer emitted `uber_jar_runId`.
- `./mvnw -B -pl liquibase-standard test`: 52/52 green, including `XMLChangeLogSAXParserTest`, `LiquibaseUtilTest` and `LiquibaseEntityResolverTest`, which assert on `DEV`.
- No `run-id:`, no `curl.*raw.githubusercontent`, no `git push`, no `git tag` remain in the file. The one `download-artifact` left is the same-run hand-off to `build-installers` and carries no `run-id:`.
- No out-of-repo workflow calls this file: a code search for `liquibase/liquibase/.github/workflows/create-release.yml` across the org returns only README hits. The extension repos each have their own `create-release.yml` calling a `build-logic` workflow of the same name.

## Behaviour change worth knowing

The `environment:` gate restricts which refs can launch a release. Both `release` and `release-publish` set `custom_branch_policies`, allowing only `branch: main` and `tag: v*`. Before this PR, `reversion` ran from any ref. A dispatch from a feature branch now fails at the job level with zero steps.

That matches practice: 38 of the last 40 `create-release.yml` dispatches came from `main`. Dispatch releases from `main`, as today.

## Not covered by this evidence

`Create release tag` and `Get GitHub App token` sit behind `dry_run == false`, so no dry run exercises them. They are first exercised by a real release. The tag call is a single `POST /git/refs` with an explicit refspec and SHA, and a failed release leaves nothing repointed because the call cannot force.

Also removed, both dead: the `build-installers` tag-existence guard, which ran `git show-ref --tags` against a checkout holding no tags AND runs after `reversion` already created the tag, so a working version of it would have failed every real release; and a `sign-artifacts.sh` call whose outputs were written into the download directory and never consumed.

TECHOPS-1223. Pairs with liquibase/build-logic#699.


.github/workflows/create-release.yml
```diff
@@ -60,33 +53,120 @@ on:
...
+          REPO: ${{ github.repository }}
+          REF: ${{ inputs.branch }}
+          VERSION: ${{ inputs.version }}
+        run: |
+          set -euo pipefail
+          # Injection sink: REF/VERSION are dispatch inputs. Passed via env, never
+          # interpolated into the script body, and VERSION is charset-validated
+          # because it becomes a git ref and a filename below.
+          [[ "$VERSION" =~ ^[A-Za-z0-9._-]+$ ]] || { echo "::error::bad version"; exit 1; }
+          sha=$(gh api "repos/$REPO/commits/$REF" --jq .sha)
```
sayaliM0412, line 88: **[HIGH]** `setup` resolves the branch to an immutable SHA but never asserts that the commit passed CI, and the build below runs [`-DskipTests` (L140)](https://github.com/liquibase/liquibase/blob/131ef8bd3801df78900346acf285c952c559945e/.github/workflows/create-release.yml#L140), so a release can be built from a red commit and nothing downstream would notice.

.github/workflows/create-release.yml
```diff
@@ -201,16 +239,60 @@ jobs:
...
+        if: ${{ inputs.standalone_zip == false && inputs.dry_run == false }}
+        uses: actions/create-github-app-token@bcd2ba49218906704ab6c1aa796996da409d3eb1 # v3.2.0
+        with:
+          app-id: ${{ env.LIQUIBASE_GITHUB_APP_ID }}
+          private-key: ${{ env.LIQUIBASE_GITHUB_APP_PRIVATE_KEY }}
+          owner: ${{ github.repository_owner }}
+          repositories: liquibase
+          permission-contents: write        # only scope; no permission-actions
+
+      - name: Create release tag
```
sayaliM0412: **[HIGH]** [`Get App credentials from vault` (L242)](https://github.com/liquibase/liquibase/blob/131ef8bd3801df78900346acf285c952c559945e/.github/workflows/create-release.yml#L242), [`Get GitHub App token` (L263)](https://github.com/liquibase/liquibase/blob/131ef8bd3801df78900346acf285c952c559945e/.github/workflows/create-release.yml#L263) and [`Create release tag` (L274)](https://github.com/liquibase/liquibase/blob/131ef8bd3801df78900346acf285c952c559945e/.github/workflows/create-release.yml#L274) all sit behind `dry_run == false`, so the green [evidence run](https://github.com/liquibase/liquibase/actions/runs/34432707689) exercised none of them and 5.0.5 is their first execution.

sayaliM0412: **[HIGH]** The tag is created at the end of `reversion`, before [`build-installers` (L289)](https://github.com/liquibase/liquibase/blob/131ef8bd3801df78900346acf285c952c559945e/.github/workflows/create-release.yml#L289) runs, so an installer failure leaves `v$VERSION` pointing at a commit with no release, and because `POST /git/refs` is deliberately non-forcing the retry then dies at the [pre-existing-tag check in `setup` (L84)](https://github.com/liquibase/liquibase/blob/131ef8bd3801df78900346acf285c952c559945e/.github/workflows/create-release.yml#L84) instead of recovering the way `git push -f` used to.

.github/workflows/create-release.yml
```diff
@@ -60,33 +53,120 @@ on:
...
+          REF: ${{ inputs.branch }}
+          RUN_NUMBER: ${{ github.run_number }}
+        run: |
+          set -euo pipefail
+          ./mvnw -B org.codehaus.mojo:versions-maven-plugin:2.21.0:set -DnewVersion="$VERSION"
+          ./mvnw -B -DskipTests \
+            -Dbuild.repository.owner=liquibase -Dbuild.repository.name=liquibase \
+            -Dbuild.branch="$REF" -Dbuild.number="$RUN_NUMBER" \
+            -Dbuild.commit="$SHA" -Dbuild.version="$VERSION" \
+            "-Dbuild.timestamp=$(date +'%Y-%m-%d %H:%M:%S %Z')" \
```
sayaliM0412: **[MEDIUM]** [`-Dbuild.timestamp=$(date ...)` (L144)](https://github.com/liquibase/liquibase/blob/131ef8bd3801df78900346acf285c952c559945e/.github/workflows/create-release.yml#L144) makes every build unique, so the released bytes still cannot be reproduced from the tag, which is the natural next step for a provenance change.

.github/workflows/create-release.yml
```diff
@@ -60,33 +53,120 @@ on:
...
     name: Re-version artifacts ${{ needs.setup.outputs.version }}
     runs-on: ubuntu-22.04
+    # [TECHOPS-1223] Same expression release-deploy-maven.yml uses, but keyed on dry_run
+    # instead of approved: this job is reached directly through workflow_dispatch (via
+    # dispatch-and-wait), never through this repo's own workflow_call chain, so dry_run
+    # is the input that actually distinguishes a rehearsal from a real release here.
+    # `release-publish` has 0 reviewers so the weekly cron is not blocked; `release` has
+    # 5. See .github/workflows/README.md for the broad-role OIDC trust this depends on.
+    environment:
+      name: ${{ inputs.dry_run && 'release-publish' || 'release' }}
```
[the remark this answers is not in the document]
sayaliM0412: @jandroav : could we check this ?

.github/workflows/create-release.yml
```diff
@@ -201,16 +239,60 @@ jobs:
...
+          private-key: ${{ env.LIQUIBASE_GITHUB_APP_PRIVATE_KEY }}
+          owner: ${{ github.repository_owner }}
+          repositories: liquibase
+          permission-contents: write        # only scope; no permission-actions
+
+      - name: Create release tag
+        # Atomic and non-forcing: POST /git/refs has no `force` parameter, so it 422s on a
+        # pre-existing ref instead of silently repointing an already-released tag. A re-run
+        # against the same version therefore fails fast here rather than repointing history.
+        if: ${{ inputs.standalone_zip == false && inputs.dry_run == false }}
```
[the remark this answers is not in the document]
sayaliM0412: @jandroav : could we check this ?

.github/workflows/create-release.yml
```diff
@@ -60,33 +53,120 @@ on:
...
+          GH_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+          REPO: ${{ github.repository }}
+          REF: ${{ inputs.branch }}
+          VERSION: ${{ inputs.version }}
+        run: |
+          set -euo pipefail
+          # Injection sink: REF/VERSION are dispatch inputs. Passed via env, never
+          # interpolated into the script body, and VERSION is charset-validated
+          # because it becomes a git ref and a filename below.
+          [[ "$VERSION" =~ ^[A-Za-z0-9._-]+$ ]] || { echo "::error::bad version"; exit 1; }
```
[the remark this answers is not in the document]
jandroav: Correct, and taken in `d600845a5`.

Reproduced the reasoning: `1..0` passes `^[A-Za-z0-9._-]+$` and `git check-ref-format refs/tags/v1..0` rejects it. The timing argument is the part that makes it worth fixing rather than noting: the tag is now created in the last job of the run, after the build, the signing and the installers, so an unusable name would burn the whole release before anything reported it.

Added `git check-ref-format "refs/tags/v$VERSION"` right after the charset check, in the same admission gate.

.github/workflows/create-release.yml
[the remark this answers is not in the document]
jandroav: The underlying concern is real and is fixed in `d600845a5`, but two parts of the framing do not hold, and the difference matters for anyone triaging this later.

**Corrected: reachability is not External.** `workflow_dispatch` requires write access on the repository. Both `release` and `release-publish` also carry `deployment_branch_policy` with `custom_branch_policies: true`, allowing only `branch: main` and `tag: v*`, so the workflow file itself cannot be run from an attacker-controlled ref. This is an insider or compromised-account path, not an external one.

**Valid, and the reason it still matters:** `branch` is a free-form dispatch input, and after this PR `reversion` builds it with Maven, which executes third-party plugins from that ref beside the GPG key. Before this PR the job downloaded a prebuilt artifact, so the arbitrary-code surface is genuinely wider now even though the change is a net improvement.

**Fix.** The reviewer-free environment is now reachable only for a dry run of `main`:

```yaml
name: ${{ (inputs.dry_run && inputs.branch == 'main') && 'release-publish' || 'release' }}
```

That keeps the weekly rehearsal unattended, which is why the conditional exists at all, and sends every other ref through `release` and its 5 reviewers. Applied to both `reversion` and `build-installers`, and the README guard table now says so.

.github/workflows/create-release.yml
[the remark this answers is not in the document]
jandroav: The fix is right and is in `d600845a5`, but the stated mechanism is not what happens, so I want to correct it on the record.

Both `action-gh-release` steps set `draft: true`, and a draft release creates no git ref. GitHub only creates the tag when the draft is published. Confirmed empirically rather than by reading: deleting a dry-run draft with `gh release delete --cleanup-tag` returns `HTTP 422: Reference does not exist`, because no tag was ever made.

What does survive is the second half of your point. `generate_release_notes` still needs a commitish to diff against, and without `target_commitish` that is the default branch rather than the commit the assets came from. Added it to both blocks with that rationale in the comment.

.github/workflows/create-release.yml
jandroav: Right, and this was the gap I liked least once you named it. Fixed in `d600845a5`.

`setup` now refuses a commit whose checks are not green, before a single build minute is spent:

```bash
runs=$(gh api "repos/$REPO/commits/$sha/check-runs?per_page=100" --jq '.check_runs')
bad=$(jq -r '[.[] | select(.conclusion=="failure" or .conclusion=="timed_out" or .conclusion=="cancelled") | .name] | join(", ")' <<<"$runs")
ok=$(jq -r '[.[] | select(.conclusion=="success")] | length' <<<"$runs")
[ -z "$bad" ] || { echo "::error::$sha has failing checks: $bad"; exit 1; }
[ "$ok" -gt 0 ] || { echo "::error::$sha has no successful check runs"; exit 1; }
```

The second assertion is deliberate: a commit with zero completed checks is refused too, because "never tested" and "tested green" must not look the same to this gate.

Kept `-DskipTests` rather than running the suite here. Re-running tests at release time would double a release to roughly 40 minutes and would still not prove the thing that matters, which is that the commit passed the full matrix. Asserting the recorded verdict is both cheaper and stronger.

.github/workflows/create-release.yml
jandroav: Correct, and I had flagged the same thing in the PR body and on the ticket, so let me say what I actually did about it rather than just agreeing.

Two changes in `d600845a5`:

1. The tag now lives in its own `tag-release` job rather than as three trailing steps of `reversion`. The unexercised surface is smaller and isolated, and a failure there cannot leave a half-finished `reversion` behind.
2. That job reads the tag back and fails the release if it is wrong:

```bash
got=$(gh api "repos/$REPO/git/ref/tags/v$VERSION" --jq .object.sha)
[ "$got" = "$SHA" ] || { echo "::error::tag v$VERSION points at $got, expected $SHA"; exit 1; }
```

That does not make the path exercised, and nothing can short of a real release. What it does is convert a silent wrong tag into a failed release, which is the difference that matters on 5.0.5.

jandroav: This is the best catch in the review, and you are right on both halves. Fixed in `d600845a5`.

I had removed `git push -f` as an unconditional win without noticing it was also the recovery path. Non-forcing plus a tag created too early is strictly worse than forcing: a failed `build-installers` left `v$VERSION` on a commit with no release, and the retry then died on setup's pre-existing-tag check with no way forward but deleting the tag by hand.

The fix is ordering, not weakening the guarantee. Tag creation moved out of `reversion` into a `tag-release` job with `needs: [setup, reversion, build-installers]`, so nothing is tagged until every asset exists. An installer failure now leaves no tag at all and the retry is clean.

It also puts the tag where it belongs conceptually: the tag is the release saying "this commit shipped", and it should be the last thing that happens, not the first.

.github/workflows/create-release.yml
jandroav: Correct. Fixed in `d600845a5`, with one caveat I want stated rather than implied.

The timestamp now comes from the commit instead of the clock:

```bash
committed=$(gh api "repos/$REPO/commits/$sha" --jq .commit.committer.date)
echo "build_timestamp=$(date -u -d "$committed" +'%Y-%m-%d %H:%M:%S UTC')" >> "$GITHUB_OUTPUT"
```

Re-running the same tag now yields the same value instead of a new one each time.

**The caveat:** this removes one source of non-determinism, it does not deliver reproducible builds. None of the 10 `pom.xml` files sets `project.build.outputTimestamp`, and the Maven wrapper is 3.9.16, so reproducible mode is off. Maven's own guide treats it as opt-in with a checklist. Calling this PR reproducible would be a claim I cannot back, so it does not.

Worth its own ticket: set `project.build.outputTimestamp`, run `mvn artifact:check-buildplan`, and then verify two builds of the same tag are byte-identical. Happy to raise it.

jandroav:
Rebased onto `main` and addressed all seven findings in `d600845a5`. Every one was reproduced before fixing; replies are on each thread.

**Two corrections to the bot, for the record:** the critical finding is not `External` (dispatch needs write access, and both environments restrict deployments to `main` and `v*`), and `action-gh-release` creates no tag here because both blocks set `draft: true`, proven by `gh release delete --cleanup-tag` returning `422 Reference does not exist`. The underlying concerns were real and are fixed; only the framing was off.

**Merge conflict** came from #7985 landing while this was open. We solved overlapping problems independently, so the resolution is a union rather than a pick: this PR keeps the atomic non-forcing `POST /git/refs` and the token staying out of `argv`, and takes #7985's two improvements, App credentials to step outputs rather than `GITHUB_ENV` and a derived `repositories` scope.

**Verified on a real dispatched run** ([34565860557](https://github.com/liquibase/liquibase/actions/runs/34565860557), green):

```
Setup                                   success
Re-version artifacts                    success
Build Installers                        success
Create release tag                      skipped   <- new job, dry_run guard holds
```

The new admission gate ran and reported `da28120ec...: 24 successful checks, none failing`, matching the count measured independently beforehand. The produced jar carries `build.timestamp=2026-09-10 16:26:08 UTC`, which is the commit's own date rather than the build clock, so re-running the same tag now yields the same value.

`PR Labels` was red because the PR carried no label. Added `skipReleaseNotes`, matching the other PRs in this epic.

**Still not covered by any evidence, and cannot be:** `Get GitHub App token` and `Create release tag` sit behind `dry_run == false`, so 5.0.5 remains their first execution. The new job reads the tag back and fails the release if it does not point at the built commit, which converts a silent wrong tag into a loud failure but does not exercise the path.

.github/workflows/create-release.yml
```diff
@@ -60,52 +53,163 @@ on:
...
+            echo "::error::tag v$VERSION already exists"; exit 1
+          fi
+
+          # Nothing else asserts the commit is releasable, and the build runs -DskipTests,
+          # so without this a red commit ships. A commit with no completed checks is refused
+          # as well: "never tested" and "tested green" must not look the same here.
+          runs=$(gh api "repos/$REPO/commits/$sha/check-runs?per_page=100" --jq '.check_runs')
+          bad=$(jq -r '[.[] | select(.conclusion=="failure" or .conclusion=="timed_out" or .conclusion=="cancelled") | .name] | join(", ")' <<<"$runs")
+          ok=$(jq -r '[.[] | select(.conclusion=="success")] | length' <<<"$runs")
+          [ -z "$bad" ] || { echo "::error::$sha has failing checks: $bad"; exit 1; }
```
sayaliM0412: One thing I hit while testing this gate: it treats every check run attached to the SHA as part of the verdict, including scheduled scans that land long after the commit.

Your own run is the example. At 05:24 setup printed `da28120ec...: 24 successful checks, none failing`, and the Trivy image scans failed on that same commit at 07:07. Same commit, opposite answer three hours later.

I looked at the last 12 commits on `main` and 6 of them carry failing check runs, mostly `Trivy Scan liquibase/liquibase-alpine`, plus one `Publish SNAPSHOT` and a `Run Test for (Java 21 windows-latest)`. So once the pagination point below is fixed, this would refuse most `main` commits, and the weekly dry run would start failing on the CVE feed rather than on anything about the build.

Could we scope it to the required contexts the ruleset already enforces, instead of everything attached to the SHA? Related: `conclusion == null` is not treated as bad today, so a commit whose checks are still running passes as long as one has already succeeded, which is "not yet red" rather than green.

.github/workflows/create-release.yml
```diff
@@ -60,52 +53,163 @@ on:
...
+    # dispatch-and-wait), never through this repo's own workflow_call chain, so dry_run
+    # is the input that actually distinguishes a rehearsal from a real release here.
+    # `release-publish` has 0 reviewers so the weekly cron is not blocked; `release` has
+    # 5. See .github/workflows/README.md for the broad-role OIDC trust this depends on.
+    environment:
+      # `release-publish` has 0 reviewers, which the weekly rehearsal needs so the cron
+      # does not hang. It is reachable only for a dry run of `main`: `branch` is a free-form
+      # dispatch input and this job builds it with Maven, so any other ref executing
+      # third-party plugins beside the GPG key goes through `release` and its 5 reviewers.
+      name: ${{ (inputs.dry_run && inputs.branch == 'main') && 'release-publish' || 'release' }}
```
sayaliM0412: This closes the `inputs.branch` half, but the environment is not holding the other half.

`release-publish` has `can_admins_bypass: true`, and all 7 deployments in its history came from `techops-1223-release-from-tag`, including run 34565860557, where `reversion` read `/vault/liquibase` and signed with the GPG key from this branch with no approval. `release` (`can_admins_bypass: false`) has had zero deployments ever, so the 5-reviewer path is still unexercised.

That makes the comment just above optimistic for anyone with admin: the workflow YAML comes from `github.ref` while the conditional keys on `inputs.branch`, so modified workflow code on any branch still lands in the 0-reviewer environment with vault and GPG. Either `can_admins_bypass: false` on `release-publish`, or a `github.ref == 'refs/heads/main'` term here, would make the comment true as written.

.github/workflows/create-release.yml
```diff
@@ -480,3 +521,90 @@ jobs:
...
+    # `build-installers` failed, and because the API call is deliberately non-forcing the
+    # retry then died on setup's pre-existing-tag check instead of recovering the way
+    # `git push -f` used to. Nothing is tagged now until every asset is built. [TECHOPS-1223]
+    if: ${{ inputs.standalone_zip == false && inputs.dry_run == false }}
+    runs-on: ubuntu-22.04
+    # `release` already gated `reversion` for this run, so asking again would be a second
+    # prompt for one release. `release-publish` is the codified "approved upstream" variant,
+    # unreachable on its own here: this job needs three jobs that cannot run without it.
+    environment:
+      name: release-publish
```
sayaliM0412: If the reason for `release-publish` here is avoiding a second prompt, I think `release` is free: approval is one per environment per run rather than per job (the pending-deployments API takes `environment_ids` and returns one object per environment), which is what we landed on in [TECHOPS-1214](https://datical.atlassian.net/browse/TECHOPS-1214). `reversion` and `build-installers` are already on `release` for a real release, so `tag-release` joining them should not add a prompt.

Worth the swap because this is the one irreversible step in the pipeline and it is the job holding the App token with `contents: write`. Happy to be wrong if 5.0.5 shows a second prompt, but it seems worth trying that way round first.

sayaliM0412 (commented):
Re-reviewed at `d600845a5`: the six points from the last pass are all genuinely addressed, and the new dry run on the fixed commit is green. Three things left, one of them about the environment rather than this file.

.github/workflows/create-release.yml
```diff
@@ -60,52 +53,163 @@ on:
...
+          fi
+
+          # Nothing else asserts the commit is releasable, and the build runs -DskipTests,
+          # so without this a red commit ships. A commit with no completed checks is refused
+          # as well: "never tested" and "tested green" must not look the same here.
+          runs=$(gh api "repos/$REPO/commits/$sha/check-runs?per_page=100" --jq '.check_runs')
+          bad=$(jq -r '[.[] | select(.conclusion=="failure" or .conclusion=="timed_out" or .conclusion=="cancelled") | .name] | join(", ")' <<<"$runs")
+          ok=$(jq -r '[.[] | select(.conclusion=="success")] | length' <<<"$runs")
+          [ -z "$bad" ] || { echo "::error::$sha has failing checks: $bad"; exit 1; }
+          [ "$ok" -gt 0 ] || { echo "::error::$sha has no successful check runs"; exit 1; }
```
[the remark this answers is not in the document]
sayaliM0412: @jandroav this one is real and worth picking up. `main` HEAD `da28120e` has 201 check runs, and recent commits have 235, 246 and 830, so the gate is reading the newest 100 of them. Right now the two failing Trivy scans on `da28120e` sit outside that page, so it would report "none failing" on a commit that has two failures.

One catch: the suggested diff does not run. `gh api --paginate --slurp ... --jq` errors with `the --slurp option is not supported with --jq`. `gh api --paginate "repos/$REPO/commits/$sha/check-runs?per_page=100" --jq '.check_runs[]'` works.

.github/workflows/README.md
```diff
@@ -112,26 +112,26 @@ You can check the `dry-run-release.yml` workflow, which is essentially composed
 [...]
 
   dry-run-create-release:
-    needs: [ setup ]
     uses: liquibase/liquibase/.github/workflows/create-release.yml@main
     with:
       version: "dry-run-${{ github.run_id }}"
-      runId: ${{ needs.setup.outputs.dry_run_id }}
+      branch: main
```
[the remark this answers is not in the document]
sayaliM0412: @jandroav could you take this one as well? L88-89 still say the artifacts come from the `main.yml` workflow, which L35 of the same file now contradicts. L99 is stale in the same way: it documents deleting the dry-run repository tag, and per your own `draft: true` finding that tag never exists.

.github/workflows/create-release.yml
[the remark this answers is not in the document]
jandroav: Confirmed, and it is worse than the sample suggested. I measured the last six `main` commits: 187, 421, 23, 235, 156 and 89 check runs. On `main` HEAD `7a9a3413` the first page of 100 contains **zero** failures while the full set contains three (two Trivy scans and a stale `test / Unit (Java 21, ubuntu-24.04)`). So the gate as written printed "none failing" on a commit with three failures.

You are also right about the diff. `gh api --paginate --slurp ... --jq` returns ``the `--slurp` option is not supported with `--jq` or `--template` ``. I went a different way entirely, see the reply on the thread above, but nothing in the new gate uses `--slurp`.

.github/workflows/create-release.yml
jandroav: Your diagnosis holds, the remedy needed changing. Three things were wrong, not one.

1. **Pagination**, per the thread below.
2. **Scheduled scans**, your point. The two Trivy failures on `7a9a3413` come from a `schedule` run of the `Docker` workflow.
3. **Attempt accumulation**, which neither of us named. `test / Unit (Java 21, ubuntu-24.04)` is on that commit **twice**: `failure` at 04:11 and `success` at 06:35. A job re-run green leaves its old failure in the list forever.

On scoping to the required contexts: I tried it and it does not work here. The ruleset requires `mergefreeze`, `PR Gate` and `PR Labels`, and all three are **absent** from every `main` commit, since they are PR-time checks that do not re-run on the merge. Scoping to them would refuse every commit.

What I did instead reaches your intent from the other side: judge **workflow runs** rather than check runs, drop `schedule` and `dynamic` events, keep only the newest run per workflow, and require all of them finished with none failing. That handles all three problems at once and gets pagination for free via `--paginate`.

Simulated old versus new over the last 12 `main` commits: 10 admitted. It flips `7a9a3413` from a false PASS to a correct REFUSE (`Main` really did fail on push), and `e4fbc2cf` from a false REFUSE to PASS.

And yes, `conclusion == null` is now refused explicitly, with its own message. "Not yet red" was passing as green.

.github/workflows/create-release.yml
jandroav: Verified every number and they are all right. `release-publish`: `can_admins_bypass: true`, no `required_reviewers` rule at all, only a branch policy, and all 7 deployments from `techops-1223-release-from-tag`. `release`: `can_admins_bypass: false`, 5 reviewers, and **0 deployments ever**.

Took your second option: `github.ref == refs/heads/main` is now part of the term, so the 0-reviewer path needs the dry run to be building `main` **and** running `main`s own workflow file. That makes the comment above it true for admins too, which it was not.

Your first option is still worth doing and I did not do it here, because it is a repo setting rather than a change to this file: `can_admins_bypass: false` on `release-publish` would be real defence in depth. Happy to raise it separately if you agree.

Worth flagging the other half of what you found: the 5-reviewer path has never executed once. Whatever 5.0.5 does, it will be the first time.

.github/workflows/create-release.yml
jandroav: Agreed, swapped to `release`.

Your reasoning on the prompt count is the deciding part: approval is per environment per run rather than per job, and `reversion` and `build-installers` are already on `release` for a real release, so `tag-release` joining them adds nothing to click. Referenced TECHOPS-1214 in the comment.

And the framing is right. This is the one irreversible step and the only job holding an App token with `contents: write`, so a 0-reviewer environment was the wrong place for it.

.github/workflows/README.md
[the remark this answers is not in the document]
jandroav: Fixed, and thank you for pushing on it, because the tag line was not only a docs problem.

Nothing creates `vdry-run-<run_id>` in this PR. The old `create-release.yml` made it at line 229 with `git tag -f "v${VERSION}"`, and this PR deletes that. But `dry-run-release.yml` still had a cleanup step doing `git push origin --delete refs/tags/vdry-run-${{ github.run_id }}`, which after this merges would fail on a ref that does not exist and take the weekly cron red. So your `draft: true` observation was pointing at a live regression, not stale prose.

Removed that step, plus the checkout and git identity that existed only to serve it. That also stops persisting a credential on disk in a job whose only remaining action is one API call.

README: the `main.yml` line now says the artifacts are built from the pinned commit SHA, and the tag-deletion line is gone.

PR #7999, opened 2026-09-15
[TECHOPS-1304] Apply Alpine security patches at build time

jnewton03:
## Why

The alpine image ships `libcrypto3` / `libssl3` at 3.5.7-r0, carrying 9 HIGH CVEs: CVE-2026-14456, -14457, -18798, -54874, -63072, -63073, -63075, -63076, -75803. All are fixed in 3.5.8-r0, which has been sitting in the Alpine v3.24 package index for months.

These are half of what has failed `docker.yml` on every scheduled run since 2026-09-03. Latest failure: https://github.com/liquibase/liquibase/actions/runs/34940240108

## Why not just bump the base image

I checked every candidate. They all still ship the vulnerable version:

| Image | openssl | Alpine |
|---|---|---|
| `alpine:3.24` (current) | 3.5.7-r0 | 3.24.1 |
| `alpine:3.24.1` | 3.5.7-r0 | 3.24.1 |
| `alpine:latest` | 3.5.7-r0 | 3.24.1 |
| `alpine:edge` | 3.5.7-r0 | 3.25.0_alpha |
| `eclipse-temurin:21-jre-alpine` | 3.5.7-r0 | 3.24.1 |

We are already on the newest Alpine release branch. There is no 3.25, and `latest` is just 3.24.1. The fix exists only in the package index, so `apk upgrade` is the only route.

## On `apk upgrade` being an anti-pattern

It used to be. That guidance was reversed:

- Docker's current `best-practices.md` does not mention `upgrade` at all. The old "avoid `apt-get upgrade`" line is gone.
- hadolint deleted **DL3017** (`apk upgrade`) in Aug 2021 and **DL3005** (`apt-get upgrade`) in Aug 2023. The PR removing DL3017 states "the rationale of not allowing container upgrades inside containers is outdated," driven by hadolint#562, "security updates should be in most Dockerfiles," which argues base images are not rebuilt often enough to deliver patches in time.

hadolint still rejects `zypper dist-upgrade` under DL3035, and that distinction is the point: this stays inside the pinned 3.24 release and moves patch versions only. It is not a dist-upgrade.

The real cost is slightly reduced build reproducibility, since the same Dockerfile can resolve different package versions over time. For an image gated on a vulnerability scan, that is the right side of the trade.

Not applied to the Debian community image, which has no OS findings today.

## Verified locally (linux/amd64)

| Build | HIGH/CRITICAL |
|---|---|
| Current `main` | 9: openssl pair + 8 lpm stdlib |
| This PR | 8: openssl row gone entirely |
| This PR + lpm 0.3.7 (#7998) | **0** |

Image still works: Liquibase 5.0.4, JRE 21.0.12, bash 5.3.9, `libcrypto3`/`libssl3` at 3.5.8-r0. Size 84.5 MB to 87.2 MB.

## Merge order

Independent of #7998, either order works. Both are needed for a fully green scan.

Jira: https://datical.atlassian.net/browse/TECHOPS-1304


docker/Dockerfile.alpine
```diff
@@ -7,8 +7,15 @@ RUN addgroup --gid 1001 liquibase && \
...
-# Install smaller JRE, if available and acceptable
-RUN apk add --no-cache openjdk21-jre-headless bash
+# Install smaller JRE, if available and acceptable.
+# `apk upgrade` first: the alpine:3.24 tag is rebuilt on Alpine's own cadence,
+# so the image lags security fixes that are already in the v3.24 package index.
+# openssl 3.5.8-r0 sat in the index for months while every published tag, edge
+# included, still shipped the vulnerable 3.5.7-r0. The scan gate fails on any
+# HIGH/CRITICAL, so we pull patches rather than wait for a rebuild. This stays
+# inside the pinned 3.24 release; it is not a dist-upgrade. [TECHOPS-1304]
+RUN apk upgrade --no-cache && \
```
filipelautert, line 17: The scan builds with cache-from: type=gha (build-logic reusable-docker-scan.yml), and this layer's key does not change when the index gets a new openssl. The next scheduled run restores it and the upgrade silently stops happening, which is the same staleness one level up.
Should not that build pass no-cache, or this RUN take a dated build-arg?

filipelautert (approved):

PR #8004, opened 2026-09-17
feat(javadocs): publish javadocs to Cloudflare R2 as well as S3 (TECHOPS-1318)

jandroav:
## Why?

`javadocs.liquibase.com` is served from **Cloudflare R2**, but this workflow only ever wrote to the **S3** origin. The only thing carrying those files across was **Sippy**, a lazy copy-on-first-read bridge that [liquibase-infrastructure#4391](https://github.com/liquibase/liquibase-infrastructure/pull/4391) removes.

With Sippy gone and no R2 write here, **the next release would land in S3 and never reach the edge** — javadocs would silently freeze at whatever was last copied. That is the TECHOPS-805 failure mode the Wave 3 epic exists to unwind, and it would not surface until someone noticed stale docs weeks later.

I found this while auditing something else, which is worth saying plainly: without it, #4391 would have shipped a silent regression.

## Evidence

Measured before the fix: the S3 origin held **3,673** objects against **3,265** in R2, so **408 had never been lazily copied** — whole trees including `liquibase-maven-plugin`, `liquibase-cdi-jakarta`, `liquibase-cdi` and `liquibase-cli`, plus the root `index.html`. A Super Slurper pass has since closed that gap; this change is what keeps it closed.

## What changed

Adds an R2 sync alongside the existing S3 one, reusing the pattern already proven in `liquibase-pro`'s own `publish-to-s3.yml`:

- Assume the org-level `DEVOPS_VAULT_OIDC_ROLE_ARN`, load `CLOUDFLARE_ACCOUNT_ID` plus the R2 S3-API key pair from the devops vault. Only those three keys, masked — no bulk import of the vault into the job.
- **All R2 work runs after the S3 sync and is `continue-on-error`.** Steps 1-5 are byte-identical to `main`; the S3 sync carries no `continue-on-error` and no `if:`, so the path that works today cannot regress.
- A broken R2 leg degrades to stale javadocs plus a `::warning::`, never a failed release.

### Why that ordering matters

Review caught that the first version put the R2 credential steps *ahead* of the S3 sync, which made the working S3 publish newly dependent on R2 credential availability. Worse, `release-published-orchestrator.yml` gates `deploy-maven` on this job's result, so a red javadocs job **skips the Maven Central deployment** for that release and the orchestrator marks the release failed. The blast radius of the fix was larger than the bug.

Two details that are easy to get wrong against R2, both learned the hard way today:

| | Why |
|---|---|
`env -u AWS_SESSION_TOKEN` | The step runs under an assumed role. R2 does not recognise that session token and SigV4 fails if it is left set. |
`--checksum-algorithm CRC32` | The newer AWS CLI default checksum trailer is not supported by R2. |

## S3 is kept on purpose

The origin bucket retires on its own ticket (TECHOPS-876). Writing to both means javadocs publish correctly **whether or not** the Sippy removal has applied yet, so this PR and #4391 are not order-dependent. Drop the S3 line together with the bucket.

## No companion PR

I originally opened liquibase-pro#4980 with the same change, on the reasoning that the byte-identical file meant two live publishers. That was wrong and it is now closed: `core/.github/workflows/` is not an executable path (Actions only reads the repo root, and liquibase-pro has no root copy), and `core/` is subtree-synced from here daily. This PR propagates on its own.

## Testing

- YAML parses; all 7 steps resolve in order.
- The upload step passes `bash -n`.
- The R2 credential path, endpoint form, `env -u` and `CRC32` flags are the same ones I used to list and Slurp these buckets successfully today, so they are verified against the live bucket rather than copied on faith.
- Not dry-runnable without cutting a release: the job is `if: inputs.dry_run == false` and gated behind the `release` environment. First real exercise will be the next release, so worth a second pair of eyes on the credential wiring.

Refs: TECHOPS-1318, TECHOPS-923, TECHOPS-876, TECHOPS-917


.github/workflows/release-deploy-javadocs.yml
```diff
@@ -104,3 +104,73 @@ jobs:
...
+          # newer AWS CLI default checksum trailer is not supported by R2.
+          env -u AWS_SESSION_TOKEN \
+            AWS_ACCESS_KEY_ID="$R2_AKID" AWS_SECRET_ACCESS_KEY="$R2_SECRET" \
+            aws s3 sync . "s3://javadocs-liquibase-com" \
+              --endpoint-url "$R2_ENDPOINT" \
+              --checksum-algorithm CRC32 \
+              --only-show-errors
+
+      - name: Warn if the R2 publish did not run
+        if: steps.r2-creds.outcome != 'success' || steps.r2-auth.outcome != 'success'
```
sayaliM0412, line 175: **[high]** This guard misses the failure it exists for: give the publish step an `id` and add it to the condition, otherwise a failed `aws s3 sync` goes green and silent.

[`Publish javadocs to Cloudflare R2`](https://github.com/liquibase/liquibase/blob/096673f62aaa2a5df9784d84c8fe8922fbe80f27/.github/workflows/release-deploy-javadocs.yml#L151) is `continue-on-error: true` and carries **no `id`**, so this `if:` can only ever catch credential problems. When the sync itself fails (wrong bucket, SigV4 rejection, R2 5xx, token scope) both `r2-auth` and `r2-creds` are still `success`, this step is skipped, and the job reports green with javadocs in S3 only.

That is the [TECHOPS-805](https://datical.atlassian.net/browse/TECHOPS-805) failure mode this PR exists to prevent, and it matters more than usual because [liquibase-infrastructure#4391](https://github.com/liquibase/liquibase-infrastructure/pull/4391) removes Sippy on the strength of this workflow working. The stated contract, "a broken R2 leg degrades to stale javadocs and a warning, never a failed release", holds for credentials but not for the publish.

```suggestion
      - name: Warn if the R2 publish did not run
        if: steps.r2-creds.outcome != 'success' || steps.r2-auth.outcome != 'success' || steps.r2-publish.outcome != 'success'
```

That needs `id: r2-publish` added to the publish step at line 151 to resolve.

jandroav: Confirmed and fixed in 0505a325f278e2026669f71a2f588482454c42c9.

You are right that the guard could only ever catch credential problems. Added `id: r2-publish` to the publish step and included its outcome in the condition, exactly as suggested.

Worth recording why the id was the missing piece: `continue-on-error` rewrites the step's *conclusion* to success but leaves its *outcome* as failure, so without an id there is nothing to read the real result from.

One note on the shape. A skipped publish also has a non-success outcome, so `steps.r2-publish.outcome != 'success'` on its own would subsume the other two clauses. I kept all three because they document what can actually go wrong, and added a comment saying so.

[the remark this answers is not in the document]
jandroav: Fixed in 0505a325f278e2026669f71a2f588482454c42c9. Same finding as https://github.com/liquibase/liquibase/pull/8004#discussion_r4039113962.

Added `id: r2-publish` and included its outcome in the warn condition.

You are right that `steps.r2-publish.outcome != 'success'` alone is sufficient, since a skipped publish is also non-success. I kept the two credential clauses alongside it for legibility about what can fail, which is behaviourally identical.

.github/workflows/release-deploy-javadocs.yml
```diff
@@ -104,3 +104,85 @@ jobs:
...
+        continue-on-error: true
+        if: steps.r2-auth.outcome == 'success'
+        # The R2 keys live in /vault/devops; every other secret stays in
+        # /vault/liquibase, read above. Pull ONLY these three and mask them rather
+        # than bulk-importing the vault. The mask escaping mirrors the vault step
+        # above: the runner unescapes %0D, %0A and %25 in command data, so a value
+        # carrying those literals would register a mask that never matches it.
+        run: |
+          set -euo pipefail
+          JSON=$(aws secretsmanager get-secret-value --secret-id /vault/devops --region us-east-1 --query SecretString --output text)
```
sayaliM0412, line 139: **[medium]** Worth confirming `jq` can parse `/vault/devops` as a whole before this merges, because a single raw control character anywhere in that secret fails this parse under `set -euo pipefail`, which skips `r2-publish` and silently degrades every release to an S3-only publish with nothing but a warning.

sayaliM0412 (commented):
Re-reviewed for TECHOPS-1318. The earlier finding is fixed: `r2-publish` now carries an `id` and the warn step checks its outcome, the R2 target matches `cloudflare_r2_bucket.javadocs_liquibase_com` (`download`/`javadocs-liquibase-com`), and `liquibase` is in `devops_vault_consumer_repos` with a subject pattern that matches the `environment:` form this job emits. One medium left.

sayaliM0412 (approved):

PR #8005, opened 2026-09-17
chore(release): publish OSS released assets to R2 instead of S3 [TECHOPS-1320]

jandroav:
## Why

TECHOPS-969 wants to decommission the S3 bucket `liquibaseorg-origin`, but it cannot be removed while OSS release assets are still being written into its `oss-released-assets/` prefix by `.github/workflows/release-publish-assets-s3.yml`. TECHOPS-1320 covers migrating all remaining `liquibaseorg-origin` writers off that bucket; this PR is the highest-risk slice, the live OSS release publish path, moved to the Cloudflare R2 bucket `liquibase-org-assets` while preserving the `oss-released-assets/` prefix exactly.

## What changed

- `release-publish-assets-s3.yml` now publishes to `s3://liquibase-org-assets/oss-released-assets/...` over the R2 S3-compatible API instead of `s3://liquibaseorg-origin/oss-released-assets/...`.
- Credential chain replaced: it no longer assumes `LIQUIBASE_RELEASE_VAULT_OIDC_ROLE_ARN` to read `/vault/liquibase` for the build-logic prod role (that role existed only to reach the old S3 bucket). It now assumes the org-wide `DEVOPS_VAULT_OIDC_ROLE_ARN` to read `/vault/devops` for the three Cloudflare R2 keys (`CLOUDFLARE_ACCOUNT_ID`, `CLOUDFLARE_R2_ACCESS_KEY_ID`, `CLOUDFLARE_R2_SECRET_ACCESS_KEY`), same pattern as `build-logic/package.yml` and `liquibase-pro/publish-to-s3.yml`.
- Display names updated from "S3" to "R2" (`name:`, `run-name:`, job `name:`) so Actions run history is not misleading.
- `.github/workflows/README.md` updated in the three spots that documented this job's target and credential scope, including the AWS-role classification (was `release-scoped`, now `broad`, see gotchas below).
- No error handling was weakened. The upload step still runs with `set -euo pipefail`, still has no `continue-on-error`, and still fails loudly if any R2 credential is missing or the sync fails.

## Consumer analysis for oss-released-assets

Searched both repos and did an org-wide `gh search code` for `oss-released-assets` and `liquibaseorg-origin`:

- The only writer of `oss-released-assets/` anywhere in the org is this workflow (and a dead vendored copy in liquibase-pro, see below).
- The only reader is the AWS SSO permission set `FullAccessToFossaOSSReport`, defined in `liquibase-infrastructure`'s `aws-control-tower/root-ou/admin-account/sso-policies.tf`, which grants `s3:GetObject/PutObject/DeleteObject/ListBucket` on `liquibaseorg-origin/oss-released-assets/*` to compliance staff via AWS console/CLI SSO. That is an internal, IAM-gated access path, not a public download.
- No public download URL, docs link, or other code path anywhere in the org references `oss-released-assets/`. The OSS releases themselves are downloaded by end users from GitHub Releases; this bucket is an internal archival copy read only by compliance staff. **No public-URL blocker found.**
- A parallel `liquibase-pro/core/.github/workflows/release-publish-assets-s3.yml` exists but is a byproduct of the `core/` git-subtree sync from `liquibase/liquibase`'s `main`. GitHub Actions only discovers workflows under a repo's top-level `.github/workflows/`, not nested paths like `core/.github/workflows/`, and nothing in liquibase-pro's own top-level workflows references it. **It is not live** and is left untouched; editing it would be pointless since the next subtree sync overwrites it from this repo's `main` anyway.

## R2 gotchas handled

- `--endpoint-url https://${CLOUDFLARE_ACCOUNT_ID}.r2.cloudflarestorage.com` on the sync call.
- `--checksum-algorithm CRC32` on the write, since newer aws-cli defaults to CRC64NVME, which R2 rejects.
- `unset AWS_SESSION_TOKEN AWS_SECURITY_TOKEN` before the R2 call, since R2 rejects requests carrying an OIDC session token.
- The original workflow used no tagging, ACLs, or presigned URLs (a plain `aws s3 sync`), so none of the R2-unsupported-operation traps apply here.

## What must land first

- The R2 bucket `liquibase-org-assets` itself. I could not find it in `liquibase-infrastructure`, on `main`, in any open PR, or via `gh search code`, as of this writing. If it has not actually been created yet, this workflow will fail loudly on the next real or dry-run release (which is the correct, loud failure mode, not a silent one), rather than writing anywhere unexpected. Please confirm the sibling infra PR has landed before relying on this in a real release.
- This PR does not touch `aws-control-tower/root-ou/engineering-ou/prod-account/github_oidc.tf` or `sso-policies.tf`. The `AllowXSDAndSchemaSync` / FOSSA-report IAM grants and the old `liquibaseorg-origin` write role are untouched and out of scope; they cover the separate XSD/JSON-schema and FOSSA-report migrations tracked elsewhere under TECHOPS-1320.

## Known limitation: broad OIDC role (CodeRabbit CWE-732, not fixed here)

CodeRabbit and a human reviewer both flagged that this workflow now assumes `DEVOPS_VAULT_OIDC_ROLE_ARN` (`liquibase-vault-oidc-role`), whose GitHub OIDC trust matches any workflow in the `liquibase`, `Datical` and `datical` orgs, in place of the release-scoped role it used before. Assessment:

- **Valid finding.** The `release` / `release-publish` environment on this job only gates who can start this specific workflow run; it does not narrow the AWS role's own trust policy. Any other workflow in those orgs that is already trusted by `liquibase-vault-oidc-role` can independently assume it and read `/vault/devops`, same as this job does.
- **Not something this workflow PR can fix.** The role and its trust policy are Terraform in `liquibase-infrastructure` (`vault-manager/oidc-roles.tf`), a different repo. Scoping it properly means adding a repository- and environment-scoped role there (mirroring `liquibase-release-vault-oidc-role`) with read access to the R2 keys in `/vault/devops`, then pointing this workflow at the new role's ARN. That is infra work, needs its own review, and needs live AWS access to verify, none of which is available from this PR or this session (AWS credentials here are expired).
- **No in-workflow mitigation closes the gap.** The job already uses the same `release`/`release-publish` environment gating as every other publishing job in this file (checked, this was not weakened), GitHub-token permissions are already minimal (`contents: read`, `id-token: write`), secrets are masked, and the OIDC session token is stripped before the R2 call. None of that affects the AWS-side trust policy, which is the actual control that's broad. Swapping to a narrower existing role wasn't done either: no repo/environment-scoped role with access to `/vault/devops` exists today, and guessing at a substitution without being able to test it against live AWS would risk silently breaking release publishing.
- **Recommendation:** file a follow-up TECHOPS ticket against `liquibase-infrastructure` to add a scoped OIDC role for R2 publishing and cut this workflow over to it. Not opened as part of this PR.

## Verification done and not done

Done:
- `python3 -c "import yaml; yaml.safe_load(...)"` parses the edited workflow cleanly.
- `bash -n` passes on all three `run:` blocks in the edited workflow.
- Confirmed `DEVOPS_VAULT_OIDC_ROLE_ARN` exists as an org-level GitHub secret (`gh secret list --org liquibase`) and is already relied on by `build-logic/package.yml`, which this repo's own `release-published-orchestrator.yml` calls with `secrets: inherit` in the same release run.
- Confirmed no remaining `liquibaseorg-origin` functional references in the edited workflow (one explanatory comment mentions it by name for context).
- Grepped this repo and did an org-wide `gh search code` for anything reading `oss-released-assets/` or linking to it publicly; found none.

Not done (needs a real or dry-run release to verify):
- Did not execute the workflow end to end (no access to trigger a real release from this environment).
- Did not independently verify the `liquibase-org-assets` R2 bucket exists or that `/vault/devops` currently holds valid Cloudflare R2 credentials; see "What must land first" above.

## Related

TECHOPS-1320, TECHOPS-969


.github/workflows/release-publish-assets-s3.yml
```diff
@@ -61,41 +61,49 @@ jobs:
         id: timestamp
         run: echo "timestamp=$(date +%Y%m%d-%H%M%S)" >> $GITHUB_ENV
 
-      - name: Configure AWS credentials for vault access
+      - name: Configure AWS credentials for devops vault access
         uses: aws-actions/configure-aws-credentials@e6de054238d6b7531b4efff3b6587d9aade6a06c # v6.2.3
         with:
-          role-to-assume: ${{ secrets.LIQUIBASE_RELEASE_VAULT_OIDC_ROLE_ARN }}
+          role-to-assume: ${{ secrets.DEVOPS_VAULT_OIDC_ROLE_ARN }}
```
sayaliM0412, line 67: This widens a release-gated job from the release-scoped `LIQUIBASE_RELEASE_VAULT_OIDC_ROLE_ARN` to a role whose trust policy admits `repo:liquibase/liquibase:*`, any ref and any event, which is the downgrade [the README table in this same PR](https://github.com/liquibase/liquibase/pull/8005/files) now records as `**broad**` and which undoes part of the [TECHOPS-1214](https://datical.atlassian.net/browse/TECHOPS-1214) split.

.github/workflows/release-publish-assets-s3.yml
```diff
@@ -61,41 +61,49 @@ jobs:
...
-            e=$v; e=${e//'%'/%25}; e=${e//$'\r'/%0D}; e=${e//$'\n'/%0A}
-            printf '::add-mask::%s\n' "$e"
-            D="VAULT_EOF_${RANDOM}${RANDOM}"
-            { echo "$k<<$D"; printf '%s\n' "$v"; echo "$D"; } >> "$GITHUB_ENV"
+          JSON=$(aws secretsmanager get-secret-value --secret-id /vault/devops --region us-east-1 --query SecretString --output text)
+          for k in CLOUDFLARE_ACCOUNT_ID CLOUDFLARE_R2_ACCESS_KEY_ID CLOUDFLARE_R2_SECRET_ACCESS_KEY; do
+            v=$(printf '%s' "$JSON" | jq -r --arg k "$k" '.[$k] // ""')
+            if [ -z "$v" ] || [ "$v" = "null" ]; then echo "::error::$k missing from /vault/devops"; exit 1; fi
+            echo "::add-mask::$v"
+            echo "$k=$v" >> "$GITHUB_ENV"
```
sayaliM0412, line 84: This replaces the hardened [TECHOPS-1107](https://datical.atlassian.net/browse/TECHOPS-1107) reader that still lives in [`release-deploy-xsd.yml`](https://github.com/liquibase/liquibase/blob/main/.github/workflows/release-deploy-xsd.yml#L74-L90), losing the sentinel that preserves trailing newlines, the `%`/`\r`/`\n` escaping that makes `add-mask` actually match, and the random heredoc delimiter, so a vault value containing a newline now becomes arbitrary `GITHUB_ENV` injection.

.github/workflows/release-publish-assets-s3.yml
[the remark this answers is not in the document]
jandroav: Valid finding, not fixed in this PR.

- `DEVOPS_VAULT_OIDC_ROLE_ARN` trust policy lives in liquibase-infrastructure (`vault-manager/oidc-roles.tf`), a different repo. This workflow cannot narrow it.
- The `release`/`release-publish` environment gates who can start this run, not who else can assume the role. That gap is real.
- No safe partial fix exists here: permissions are already minimal, no scoped role with `/vault/devops` access exists to swap in, and AWS creds are expired so nothing can be verified live.
- Recommendation: a repo- and environment-scoped OIDC role for R2 publishing, filed as its own liquibase-infrastructure ticket. Details in the PR description.

jandroav: Same underlying finding as CodeRabbit CWE-732, my earlier assessment stands: this is real, and the fix (a repo/environment-scoped role) belongs in liquibase-infrastructure, not this PR.

Fixed the self-contradiction you flagged though: the deployment-pipeline diagram in `.github/workflows/README.md` still marked `publish-assets-s3` as `:::scoped` (green) after the table above it already said `**broad**`. Diagram now says `:::broad` to match the table.

.github/workflows/release-publish-assets-s3.yml
jandroav: Agreed, this was a real regression. Ported the hardened TECHOPS-1107 reader back from release-deploy-xsd.yml:

- `jq -j` plus the printf `X` sentinel, so a trailing newline in the secret is preserved instead of stripped.
- Mask escaping in order: `%` to `%25` first, then `\r`/`\n`, so `::add-mask::` actually matches what the runner unescapes.
- Random heredoc delimiter into `GITHUB_ENV` instead of `k=v`, so a newline in a secret cannot inject extra env entries.

Only the secret id (`/vault/devops`) and key list (the three Cloudflare R2 keys) differ from the original. Comments kept.

sayaliM0412 (approved):

PR #8010, opened 2026-09-21
chore(javadocs): drop the S3 sync now that R2 is the only target (TECHOPS-921)

jandroav:
## Why

https://github.com/liquibase/liquibase-infrastructure/pull/4397 deletes the S3 bucket `javadocsliquibasecom-origin` and the IAM grants that let CI write to it. `release-deploy-javadocs.yml` still runs a plain `aws s3 sync` to that bucket with no `continue-on-error`, so once the bucket or grants go, this job fails outright.

That failure has a bigger blast radius than javadocs. In `release-published-orchestrator.yml`, `deploy-maven` only runs if `deploy-javadocs` is `success` or `skipped` (line 159); a `failure` is neither, so `deploy-maven` gets skipped. `release-docker` requires `deploy-maven` to be `success` (line 142), so it gets skipped too. A failed javadocs job silently takes Maven Central and Docker publishing down with it on the next release.

## What changed

- Removes the `aws s3 sync . s3://javadocsliquibasecom-origin` line.
- Removes the three steps that only existed to reach that bucket (`Configure AWS credentials for vault access`, `Get secrets from vault`, `Configure AWS credentials` / `AWS_PROD_GITHUB_OIDC_ROLE_ARN_BUILD_LOGIC`). Nothing else in this job used them.
- Cloudflare R2 (added in #8004) becomes the only publish target. Its `continue-on-error` behavior is unchanged: a broken R2 leg still degrades to a warning, never a failed release.
- Renamed the job and the unpack step so their names stop saying "S3", and updated the three `README.md` lines that described the old destination.

## Required merge order

1. https://github.com/liquibase/liquibase/pull/8004 (adds the R2 leg) — **must merge first**
2. This PR
3. https://github.com/liquibase/liquibase-infrastructure/pull/4397 (deletes the S3 bucket and grants) — **must merge after this one**

This PR is stacked on #8004's branch, so its diff will show as just this change once #8004 merges.

## What is still needed

- `README.md`'s credential-guard table and the deployment-pipeline mermaid diagram still describe `deploy-javadocs` as a `release-scoped` role reading `/vault/liquibase` through the build-logic role. After this PR it instead best-effort-reads `/vault/devops` via `DEVOPS_VAULT_OIDC_ROLE_ARN`, same as the `package` job. That reclassification touches the diagram's color legend and I did not want to guess at it inside a narrowly-scoped removal PR; it needs a deliberate follow-up doc pass (or can be folded into review here if a maintainer prefers).
- liquibase-pro needs no equivalent change. Its copy of this workflow lives only at `core/.github/workflows/release-deploy-javadocs.yml`, a git-subtree sync target that GitHub Actions never discovers (only top-level `.github/workflows/` is live), and it has no `LIQUIBASE_RELEASE_VAULT_OIDC_ROLE_ARN` secret either. liquibase-pro#4980 attempted the R2-leg companion PR there and was closed for this reason; this repo's changes propagate to `core/` on the next daily subtree sync with no action needed.


sayaliM0412 (approved):

jandroav:
Superseded by https://github.com/liquibase/liquibase/pull/8017. GitHub's native stack kept this PR pinned to the async merge endpoint after #8004 merged, and that endpoint would not clear the stale-review gate. #8017 is identical content cut from main and is merged.

